PR #247, opened 2021-09-14
Basic Framework for FIPS Service Indicator

samuel40791765:
### Issues:
Addresses `CryptoAlg-835` 

### Description of changes: 
Basic framework and tests of Service Indicator for FIPS 140-3 validation.

### Call-outs:
Indicators checks have only been added for the `aead` method for AES-GCM and one-shot function for AES-CBC in this PR so far. 
The main focus of this PR is to look at the correctness of the framework of the service indicator, but feel free to give advice on where the indicator checks that exist right now could be better placed.

### Testing:
Added new `service_indicator.cc` dimension:

- Basic function test
- Check for approval of AES-CBC one-shot function
- Check for approval of AES-GCM `aead` API usage


crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -0,0 +1,168 @@
...
+  if (!check_test(kPlaintext, output, sizeof(kPlaintext),
+                  "AES-CBC Decryption KAT")) {
+    return;
+  }
+
+  ASSERT_TRUE(awslc_fips_check_service_approved(counter, fips_approved_evp_aes_128_cbc));
+}
+
+TEST(ServiceIndicatorTest, AESGCM) {
+  EVP_AEAD_CTX aead_ctx;
```
torben-hansen, line 142: Should let C++ manage memory e.g. https://github.com/awslabs/aws-lc/blob/main/include/openssl/aead.h#L471

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -0,0 +1,92 @@
...
+      CRYPTO_get_thread_local(AWSLC_THREAD_LOCAL_FIPS_SERVICE_INDICATOR_STATE);
+  if (!indicator) {
+    awslc_fips_service_indicator_init_state();
+    indicator = CRYPTO_get_thread_local(AWSLC_THREAD_LOCAL_FIPS_SERVICE_INDICATOR_STATE);
+  }
+  indicator->counter = 0;
+  indicator->serviceID = fips_approved_no_state;
+}
+
+int awslc_fips_check_service_approved(int prev_counter, int service_id) {
```
nebeid, line 42: I was thinking we could have a variadic macro which calls the function and checks if it's approved, for example:
```
#define IS_FIPS_APPROVED_CALL_SERVICE(approved, func, ...)  \
  do { \
    approved = 0 \
    counter = awslc_fips_service_indicator_get_counter() \
    func(__VA_ARGS__);                                      \
    if (awslc_fips_service_indicator_get_counter() > counter) { \
        approved = 1                                             \
    }                                             \
 while(0)
```
So that the caller doesn't have to know the service_id to check for an approved call.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -0,0 +1,92 @@
...
+    return 0;
+  }
+
+  if(indicator->serviceID == service_id && indicator->counter > prev_counter){
+    return 1;
+  }
+  return 0;
+}
+
+void awslc_fips_service_indicator_init_state(void) {
```
nebeid, line 55: Don't we want to know if this call succeeded?

crypto/internal.h
```diff
@@ -894,6 +895,28 @@ static inline void CRYPTO_store_word_le(void *out, crypto_word_t v) {
 
 
 // FIPS functions.
+#if defined(AWSLC_FIPS)
+enum fips_approved_algorithm_t {
```
nebeid, line 899: I wonder if we should make them capital. Is there a convention for enums in the code?

nebeid (commented):
Will continue reviewing

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -0,0 +1,194 @@
...
+  state = bssl::UniquePtr<fips_service_indicator_state>(awslc_fips_service_indicator_get_state());
+
+  ASSERT_TRUE(state.get());
+  ASSERT_EQ(state.get()->counter,0);
+  ASSERT_EQ(state.get()->serviceID, fips_approved_no_state);
+
+  int counter = awslc_fips_service_indicator_get_counter();
+  ASSERT_EQ(counter, state.get()->counter);
+
+  // Call a random approved service.
```
nebeid, line 84:
```suggestion
  // Call an approved service.
```

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -0,0 +1,194 @@
...
+
+static const uint8_t kAESKey[16] = {
+    'B','o','r','i','n','g','C','r','y','p','t','o',' ','K', 'e','y'};
+static const uint8_t kPlaintext[64] = {
+    'B','o','r','i','n','g','C','r','y','p','t','o','M','o','d','u','l','e',
+    ' ','F','I','P','S',' ','K','A','T',' ','E','n','c','r','y','p','t','i',
+    'o','n',' ','a','n','d',' ','D','e','c','r','y','p','t','i','o','n',' ',
+    'P','l','a','i','n','t','e','x','t','!'};
+
+#if defined(AWSLC_FIPS)
```
nebeid, line 34: Why this is not inserted earlier?

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -0,0 +1,194 @@
...
+  bssl::ScopedEVP_AEAD_CTX aead_ctx;
+
+  uint8_t output[256];
+  int counter = awslc_fips_service_indicator_get_counter();
+
+  size_t out_len;
+  ASSERT_TRUE(EVP_AEAD_CTX_init(aead_ctx.get(), EVP_aead_aes_128_gcm_randnonce(), kAESKey,
+                         sizeof(kAESKey), 0, nullptr));
+
+  // AES-GCM Encryption KAT
```
nebeid, line 155: This is not a KAT (known-answer test), but can be consistency test if we use open to check that the plaintext is the same as the ciphertext and that it's approved as well (counter increased again).

crypto/internal.h
samuel40791765: Changed to capital letters

crypto/fipsmodule/service_indicator/service_indicator_test.cc
samuel40791765: Added a decryption test right after

crypto/fipsmodule/service_indicator/service_indicator.c
samuel40791765: Changed to a macro definition and let all the service indicator functions to return `int`

include/openssl/service_indicator.h
```diff
@@ -8,31 +8,26 @@
...
+#define IS_FIPS_APPROVED_CALL_SERVICE(approved, func, ...)                  \
+  do {                                                                      \
+    approved = 0;                                                           \
+    int counter = awslc_fips_service_indicator_get_counter();               \
+    func(__VA_ARGS__);                                                      \
+    if (awslc_fips_service_indicator_get_counter() > counter) {             \
+        approved = 1;                                                       \
+    }                                                                       \
+ }                                                                          \
+ while(0);                                                                  \
```
nebeid, line 25: The `do-while{0}` is used to put the code in a scope and to leave off the last `;` so that you can append it to the macro calls as if it were a function call, so I suggest you make this change, Samuel.

nebeid (commented):
Looking good, Samuel.

include/openssl/service_indicator.h
```diff
@@ -0,0 +1,33 @@
...
+OPENSSL_EXPORT int awslc_fips_service_indicator_get_counter(void);
+OPENSSL_EXPORT int awslc_fips_service_indicator_get_serviceID(void);
+OPENSSL_EXPORT int awslc_fips_service_indicator_reset_state(void);
+
+#if defined(AWSLC_FIPS)
+#define IS_FIPS_APPROVED_CALL_SERVICE(approved, func, ...)                  \
+  do {                                                                      \
+    approved = 0;                                                           \
+    int counter = awslc_fips_service_indicator_get_counter();               \
+    func(__VA_ARGS__);                                                      \
```
torben-hansen, line 20: Why use variadic arguments?

include/openssl/service_indicator.h
```diff
@@ -0,0 +1,33 @@
...
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
+OPENSSL_EXPORT int awslc_fips_service_indicator_get_counter(void);
+OPENSSL_EXPORT int awslc_fips_service_indicator_get_serviceID(void);
+OPENSSL_EXPORT int awslc_fips_service_indicator_reset_state(void);
+
+#if defined(AWSLC_FIPS)
+#define IS_FIPS_APPROVED_CALL_SERVICE(approved, func, ...)                  \
```
torben-hansen, line 16: There should also be a macro for non-fips, makes it much easier for apps to have a uniform codebase.

crypto/internal.h
```diff
@@ -894,6 +895,34 @@ static inline void CRYPTO_store_word_le(void *out, crypto_word_t v) {
...
+  FIPS_APPROVED_EVP_AES_192_CTR = 4,
+  FIPS_APPROVED_EVP_AES_256_CTR = 5,
+  FIPS_APPROVED_EVP_AES_128_CBC = 6,
+  FIPS_APPROVED_EVP_AES_192_CBC = 7,
+  FIPS_APPROVED_EVP_AES_256_CBC = 8,
+
+  FIPS_APPROVED_ALGORITHM_MAX = 8,
+};
+
+struct fips_service_indicator_state {
```
torben-hansen, line 914: Should not be an int types imo. either `uint32_t` or `uint64_t`. Pref the latter for the counter.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -0,0 +1,85 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include <openssl/crypto.h>
+#include <openssl/service_indicator.h>
+
+#if defined(AWSLC_FIPS)
+int awslc_fips_service_indicator_get_counter(void) {
+  struct fips_service_indicator_state *indicator =
+      CRYPTO_get_thread_local(AWSLC_THREAD_LOCAL_FIPS_SERVICE_INDICATOR_STATE);
+  if (!indicator) {
```
torben-hansen, line 11: I generally think `if(!ptr)` is harder to read than just `if(ptr == NULL)`. In the former you have to mentally do the logical not, and then what does if do in that case.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -0,0 +1,85 @@
...
+      return 0;
+    }
+    indicator = CRYPTO_get_thread_local(AWSLC_THREAD_LOCAL_FIPS_SERVICE_INDICATOR_STATE);
+  }
+  indicator->counter = 0;
+  indicator->serviceID = FIPS_APPROVED_NO_STATE;
+  return 1;
+}
+
+int awslc_fips_service_indicator_init_state(void) {
```
dkostic, line 46: you can make this function static since it's not supposed to be used outside of the file.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -0,0 +1,85 @@
...
+  struct fips_service_indicator_state *indicator =
+      CRYPTO_get_thread_local(AWSLC_THREAD_LOCAL_FIPS_SERVICE_INDICATOR_STATE);
+  if (!indicator) {
```
dkostic, line 49: if the function is static and only called when indicator is NULL then this is not needed.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -0,0 +1,85 @@
...
+    if(!awslc_fips_service_indicator_init_state()) {
+      return 0;
+    }
+    indicator = CRYPTO_get_thread_local(AWSLC_THREAD_LOCAL_FIPS_SERVICE_INDICATOR_STATE);
+  }
+  return indicator->serviceID;
+}
+
+int awslc_fips_service_indicator_reset_state(void) {
```
dkostic, line 32: you can make this function static since it's not supposed to be used outside of the file.

dkostic, line 32: why do we need the reset function?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -0,0 +1,85 @@
...
+  if (!indicator) {
+    if(!awslc_fips_service_indicator_init_state()) {
+      return 0;
+    }
+    indicator = CRYPTO_get_thread_local(AWSLC_THREAD_LOCAL_FIPS_SERVICE_INDICATOR_STATE);
+  }
+  return indicator->counter;
+}
+
+int awslc_fips_service_indicator_get_serviceID(void) {
```
dkostic, line 20: why not just return `FIPS_APPROVED_NO_STATE` and avoid creating the state?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -0,0 +1,85 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <openssl/crypto.h>
+#include <openssl/service_indicator.h>
+
+#if defined(AWSLC_FIPS)
+int awslc_fips_service_indicator_get_counter(void) {
```
dkostic, line 8: why not just return 0 and avoid creating the state?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -0,0 +1,85 @@
...
+void awslc_fips_service_indicator_update_state(enum fips_approved_algorithm_t service_id) {
+  struct fips_service_indicator_state *indicator =
+      CRYPTO_get_thread_local(AWSLC_THREAD_LOCAL_FIPS_SERVICE_INDICATOR_STATE);
+  if (!indicator) {
+    if(!awslc_fips_service_indicator_init_state()) {
+      return;
+    }
+    indicator = CRYPTO_get_thread_local(AWSLC_THREAD_LOCAL_FIPS_SERVICE_INDICATOR_STATE);
+  }
+  indicator->counter++;
```
dkostic, line 75: overflow protection?

crypto/internal.h
```diff
@@ -894,6 +895,34 @@ static inline void CRYPTO_store_word_le(void *out, crypto_word_t v) {
...
+    int counter;
+    int serviceID;
```
dkostic, line 916: I think `service_id` would be in line with our naming conventions?

dkostic (commented):
First pass, only reviewd the overall design of the solution.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -70,41 +70,33 @@ static const uint8_t kAESCBCCiphertext[64] = {
 
 TEST(ServiceIndicatorTest, BasicTest) {
```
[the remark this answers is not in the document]
samuel40791765: Added additional checks for the `service_id`

crypto/fipsmodule/service_indicator/service_indicator.c
samuel40791765: Added a `UINT64_MAX` check and an error output above.

crypto/fipsmodule/service_indicator/service_indicator.c
dkostic: I guess you are referring to the power-on self-tests being done in FIPS mode? We should definitely reset the indicator at the end of the power-on tests function. Otherwise we'd be indicating incorrectly that the module provided a FIPS approved service to the consumer even though the services are internally called and not asked for by the consumer.

If that's done then you won't need to reset the indicator in the unit test. Why do you think this function would be useful to a consumer? I'd prefer to keep the exported APIs minimal (only what's necessary).

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -62,6 +63,7 @@ void awslc_fips_service_indicator_update_state(enum fips_approved_algorithm_t se
   indicator->service_id = service_id;
   if(indicator->counter == UINT64_MAX) {
     OPENSSL_PUT_ERROR(EVP, ERR_R_OVERFLOW);
+    indicator->counter = 0;
     return;
   }
```
nebeid: I don't know if we need the check and the error. I think we can just wrap around quietly. Is that bad?

samuel40791765: I removed the error output. I don't think it makes sense to output an error message, if we've already wrapped around quietly ourselves either.

include/openssl/service_indicator.h
```diff
@@ -0,0 +1,40 @@
...
+    uint64_t counter = awslc_fips_service_indicator_get_counter();          \
+    func;                                                                   \
+    if (awslc_fips_service_indicator_get_counter() != counter) {            \
+        (approved) = 1;                                                     \
+    }                                                                       \
+ }                                                                          \
+ while(0)                                                                   \
+
+#else
+#define IS_FIPS_APPROVED_CALL_SERVICE(approved, func, ...)                  \
```
torben-hansen, line 27: variadic argument?

crypto/internal.h
```diff
@@ -893,7 +894,35 @@ static inline void CRYPTO_store_word_le(void *out, crypto_word_t v) {
 }
 
 
+enum fips_approved_algorithm_t {
```
torben-hansen, line 897: Should this be defined in `crypto/internal.h`? If consumers are attempting to write a uniform code-base with both fips and non-fips, they have to do conditionals every time not in fips mode because these enums would be hidden.

include/openssl/service_indicator.h
```diff
@@ -0,0 +1,40 @@
...
+
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
+OPENSSL_EXPORT uint64_t awslc_fips_service_indicator_get_counter(void);
+OPENSSL_EXPORT uint32_t awslc_fips_service_indicator_get_serviceID(void);
+
+#if defined(AWSLC_FIPS)
+#define IS_FIPS_APPROVED_CALL_SERVICE(approved, func)                       \
```
torben-hansen, line 15: why not include serviceid call in the macro scope? It hides more details and is easier to use.

include/openssl/service_indicator.h
```diff
@@ -0,0 +1,40 @@
...
+    if (awslc_fips_service_indicator_get_counter() != counter) {            \
+        (approved) = 1;                                                     \
+    }                                                                       \
+ }                                                                          \
+ while(0)                                                                   \
+
+#else
+#define IS_FIPS_APPROVED_CALL_SERVICE(approved, func, ...)                  \
+  do {                                                                      \
+    (approved) = 0;                                                         \
```
torben-hansen, line 29: You can argue what to return here... Consuming code would likely look for `approved != 0` or something (should we have a macro here, to lock into a value?).

So, for non-fips, this would fail all the time.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -0,0 +1,76 @@
...
+  struct fips_service_indicator_state *indicator =
+      CRYPTO_get_thread_local(AWSLC_THREAD_LOCAL_FIPS_SERVICE_INDICATOR_STATE);
+  if (indicator == NULL) {
+    if(!awslc_fips_service_indicator_init_state()) {
+      return;
+    }
+    indicator = CRYPTO_get_thread_local(AWSLC_THREAD_LOCAL_FIPS_SERVICE_INDICATOR_STATE);
+  }
+  indicator->service_id = service_id;
+  if(indicator->counter == UINT64_MAX) {
```
torben-hansen, line 64: We can just let it overflow?

crypto/internal.h
```diff
@@ -893,7 +894,35 @@ static inline void CRYPTO_store_word_le(void *out, crypto_word_t v) {
...
+  FIPS_APPROVED_EVP_AES_256_CBC = 8,
+
+  FIPS_APPROVED_ALGORITHM_MAX = 8,
+};
+
 // FIPS functions.
+#if defined(AWSLC_FIPS)
+
+struct fips_service_indicator_state {
+  uint64_t counter;
```
torben-hansen, line 915: We can statically assert that this field is at least a certain size in e.g. the unit tests.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  IS_FIPS_APPROVED_CALL_SERVICE(approved,AES_cbc_encrypt(kPlaintext, output,
+                              sizeof(kPlaintext), &aes_key, aes_iv, AES_ENCRYPT));
+  if (!check_test(kAESCBCCiphertext, output, sizeof(kAESCBCCiphertext),
+                  "AES-CBC Encryption KAT")) {
+    return;
+  }
+  ASSERT_TRUE(approved);
+  serviceID = awslc_fips_service_indicator_get_serviceID();
```
torben-hansen, line 110: So, consumer has to verify two values?

crypto/fipsmodule/service_indicator/service_indicator.c
torben-hansen: Not sure this is needed if the check is changed to `!=`.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -71,6 +71,8 @@ static const uint8_t kAESCBCCiphertext[64] = {
 TEST(ServiceIndicatorTest, BasicTest) {
   int approved = 0;
   uint32_t serviceID = 0;
+  uint64_t counter = awslc_fips_service_indicator_get_counter();
+  ASSERT_EQ(sizeof(counter), (uint64_t)8);
```
dkostic, line 75: how did you come up with 8? Shouldn't the counter be 0 since we haven't called any service yet?

include/openssl/service_indicator.h
```diff
@@ -8,25 +8,44 @@
...
  while(0)                                                                   \
 
 #else
-#define IS_FIPS_APPROVED_CALL_SERVICE(approved, func, ...)                  \
+// Assume true when FIPS is not on, for easier consumer compatibility that have
+// both FIPS and non-FIPS libraries.
+#define IS_FIPS_APPROVED_CALL_SERVICE(approved, func)                       \
   do {                                                                      \
-    (approved) = 0;                                                         \
+    (approved) = FIPS_APPROVED;                                             \
```
nebeid, line 48: I would worry to return a FIPS-specific indicator as approved when not in FIPS-mode. I'd rather return not approved. I think the BasicTest is assuming that too.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
samuel40791765: The test here is only asserting the field size of the counter, which is 8 bytes. I removed the test for the `counter = 0` because any previous tests running on the test thread still increment the service indicator counter. We also changed the logic of the counter to `!=`, so the initial value shouldn't really matter anymore.

crypto/fipsmodule/aes/mode_wrappers.c
```diff
@@ -57,23 +57,16 @@
 // hwaes_capable when on in ARM uses 10, 12, 14 for key rounds.
```
nebeid, line 57: Is this what you meant to say in the comment?
```suggestion
// hwaes_capable when not on ARM uses 10, 12, 14 for key rounds.
```

samuel40791765: reclarified 👍

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -0,0 +1,175 @@
...
+    0x34, 0xdd, 0x5a, 0x2d, 0xc4, 0xb4, 0xf5, 0xa8,
+#if !defined(BORINGSSL_FIPS_BREAK_AES_CBC)
+    0x35
+#else
+    0x00
+#endif
+};
+
+TEST(ServiceIndicatorTest, BasicTest) {
+  int approved = 0;
```
torben-hansen, line 72: Got macro's for this state. And below.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -0,0 +1,175 @@
...
+
+  // Call an approved service.
```
torben-hansen, line 74: should also try a non-approved service?

crypto/fipsmodule/service_indicator/internal.h
```diff
@@ -0,0 +1,28 @@
+//
+// Created by Chiang, Samuel on 9/20/21.
```
nebeid, line 2: Copyright header

nebeid (approved):

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -0,0 +1,174 @@
...
+TEST(ServiceIndicatorTest, AESCBC) {
+  int approved = AWSLC_NOT_APPROVED;
+  AES_KEY aes_key;
+  uint8_t aes_iv[16];
+  uint8_t output[256];
+
+  // AES-CBC Encryption KAT
+  memcpy(aes_iv, kAESIV, sizeof(kAESIV));
+  ASSERT_TRUE(AES_set_encrypt_key(kAESKey, 8 * sizeof(kAESKey), &aes_key) == 0);
+  CALL_SERVICE_AND_CHECK_APPROVED(approved,AES_cbc_encrypt(kPlaintext, output,
```
torben-hansen, line 107: Nit: space before `AES_cbc_encrypt`

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -0,0 +1,174 @@
...
+  uint8_t output[256];
+  size_t out_len;
+
+  OPENSSL_memset(nonce, 0, sizeof(nonce));
+  ASSERT_TRUE(EVP_AEAD_CTX_init(aead_ctx.get(), EVP_aead_aes_128_gcm(),
+                                kAESKey, sizeof(kAESKey), 0, nullptr));
+  CALL_SERVICE_AND_CHECK_APPROVED(approved, EVP_AEAD_CTX_seal(aead_ctx.get(),
+          output, &out_len, sizeof(output), nonce, EVP_AEAD_nonce_length(EVP_aead_aes_128_gcm()),
+          kPlaintext, sizeof(kPlaintext), nullptr, 0));
+  ASSERT_FALSE(approved);
```
torben-hansen, line 95: Safer to use `approved == AWSLC_NOT_APPROVED`?

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -0,0 +1,174 @@
...
+  bssl::ScopedEVP_AEAD_CTX aead_ctx;
+  uint8_t encrypt_output[256];
+  uint8_t decrypt_output[256];
+  size_t out_len;
+  size_t out2_len;
+
+  ASSERT_TRUE(EVP_AEAD_CTX_init(aead_ctx.get(), EVP_aead_aes_128_gcm_randnonce(),
+                                kAESKey, sizeof(kAESKey), 0, nullptr));
+
+  // AES-GCM Encryption
```
torben-hansen, line 134: Sorta implied by the function names.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -0,0 +1,174 @@
...
+  uint8_t encrypt_output[256];
+  size_t out_len;
+  ASSERT_TRUE(EVP_AEAD_CTX_init(aead_ctx.get(), EVP_aead_aes_128_gcm_randnonce(),
+                                kAESKey, sizeof(kAESKey), 0, nullptr));
+  CALL_SERVICE_AND_CHECK_APPROVED(approved, EVP_AEAD_CTX_seal(aead_ctx.get(),
+         encrypt_output, &out_len, sizeof(encrypt_output), nullptr, 0, kPlaintext, sizeof(kPlaintext), nullptr, 0));
+  // Macro should return true, to ensure FIPS/Non-FIPS compatibility.
+  ASSERT_TRUE(approved);
+
+  // Actual approval check should return false during non-FIPS.
```
torben-hansen, line 167: nice check

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -0,0 +1,174 @@
...
+TEST(ServiceIndicatorTest, NotApprovedTest) {
+  int approved = AWSLC_NOT_APPROVED;
+
+  // Call a not approved service.
+  bssl::ScopedEVP_AEAD_CTX aead_ctx;
+  uint8_t nonce[EVP_AEAD_MAX_NONCE_LENGTH];
+  uint8_t output[256];
+  size_t out_len;
+
+  OPENSSL_memset(nonce, 0, sizeof(nonce));
```
torben-hansen, line 89: Nit: you could just `uint8_t nonce[EVP_AEAD_MAX_NONCE_LENGTH] = {0};`

include/openssl/service_indicator.h
```diff
@@ -0,0 +1,67 @@
...
+//
+// |FIPS_service_indicator_before_call| is intended to be called right before
+// an approved service, while |FIPS_service_indicator_after_call| should be
+// called immediately after. If the values returned from these two functions are
+// not equal, this means that the service called in between is deemed to be
+// approved. If the values are still the same, this means the counter has not
+// been incremented, and the service called is otherwise not approved for FIPS.
+//
+// |FIPS_service_indicator_check_approved| is intended to take in the before and
+// after counter values. It will return 1 if the approval check was successful,
```
torben-hansen, line 26: It returns `AWSLC_APPROVED` if approved, `AWSLC_NOT_APPROVED` otherwise.

include/openssl/service_indicator.h
```diff
@@ -0,0 +1,67 @@
...
+// and 0 if otherwise.
+OPENSSL_EXPORT uint64_t FIPS_service_indicator_before_call(void);
+OPENSSL_EXPORT uint64_t FIPS_service_indicator_after_call(void);
+OPENSSL_EXPORT int FIPS_service_indicator_check_approved(int before, int after);
+
+#if defined(AWSLC_FIPS)
+
+// This macro provides a bundled way to do an approval check and run the service.
+// The |approved| value passed in will change to 1 and 0 accordingly to the
```
torben-hansen, line 35: `AWSLC_APPROVED`, and `AWSLC_NOT_APPROVED`

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -0,0 +1,215 @@
...
+  ASSERT_EQ(approved, AWSLC_APPROVED);
+
+  // Fail an approved service on purpose.
+  return_val = 0;
+  CALL_SERVICE_AND_CHECK_APPROVED(approved, return_val = EVP_AEAD_CTX_seal(aead_ctx.get(),
+          output, &out_len, 0, nullptr, 0, kPlaintext, sizeof(kPlaintext), nullptr, 0));
+  ASSERT_EQ(return_val, 0);
+  ASSERT_EQ(approved, AWSLC_NOT_APPROVED);
+
+  // Fail an approved service  on purpose while wrapped in an if statement.
```
torben-hansen, line 111: Giant nit, but white space after "service"

crypto/fipsmodule/service_indicator/internal.h
```diff
@@ -0,0 +1,42 @@
...
+OPENSSL_INLINE void FIPS_service_indicator_update_state(void) { }
+
+#endif // AWSLC_FIPS
+
+// Service indicator check for most AES algorithms.
+// hwaes_capable when enabled in x86 uses 9, 11, 13 for key rounds.
+// hwaes_capable when enabled in ARM uses 10, 12, 14 for key rounds.
+// When compiling with different ARM specific platforms, 9, 11, 13 are used for
+// key rounds.
+// TODO: narrow down when and which assembly/x86 ARM CPUs use [9,11,13] and [10,12,14]
```
torben-hansen, line 35: Should this be an open TODO?

torben-hansen (dismissed):

crypto/fipsmodule/service_indicator/internal.h
```diff
@@ -6,22 +6,35 @@
...
+// state. This function is used to update the service indicator state, if the
+// service is deemed to be approved.
 void FIPS_service_indicator_update_state(void);
 
 #else
-OPENSSL_INLINE int FIPS_service_indicator_reset_state(void) { return 0; }
+
+// Service indicator functions are not intended for use during non-FIPS mode.
+// If these functions are run during non-FIPS mode, they will return nothing or
+// 0 accordingly.
```
nebeid, line 24: Is this sentence still accurate?

crypto/fipsmodule/service_indicator/internal.h
```diff
@@ -6,22 +6,35 @@
...
 
+// Service indicator check for most AES algorithms.
+// hwaes_capable when enabled in x86 uses 9, 11, 13 for key rounds.
+// hwaes_capable when enabled in ARM uses 10, 12, 14 for key rounds.
+// When compiling with different ARM specific platforms, 9, 11, 13 are used for
+// key rounds.
 void AES_verify_service_indicator(unsigned key_rounds);
+
+// AEAD APIs work with different parameters. Only internal IV for AES-GCM is
+// meant to be approved.
```
nebeid, line 37:
```suggestion
// The service indicator checks use different parameters for AEAD APIs than those of other AES modes. 
// AES-GCM is approved only with an internal IV <IG reference and section>.
```

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -68,56 +68,89 @@ TEST(ServiceIndicatorTest, BasicTest) {
...
-  CALL_SERVICE_AND_CHECK_APPROVED(approved,AES_cbc_encrypt(kAESCBCCiphertext, output,
+  CALL_SERVICE_AND_CHECK_APPROVED(approved, AES_cbc_encrypt(kAESCBCCiphertext, output,
                         sizeof(kAESCBCCiphertext), &aes_key, aes_iv, AES_DECRYPT));
   ASSERT_TRUE(check_test(kPlaintext, output, sizeof(kPlaintext),
                   "AES-CBC Decryption KAT"));
-  ASSERT_TRUE(approved);
+  ASSERT_EQ(approved, AWSLC_APPROVED);
 }
 
 TEST(ServiceIndicatorTest, AESGCM) {
```
nebeid, line 156: I think it would be good to have a test here for the non-approved use of AES-GCM with an external IV here. May be also the one in BasicTest can be changed to something obviously non-approved, for example, AEAD with ChaCha20Poly1305 (is that possible?), or any other non-approved service and explain in the comment that/why it's not approved. This second ask can wait till later, but then on l.123, clarify why it's not approved.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -131,17 +164,15 @@ TEST(ServiceIndicatorTest, AESGCM) {
...
+
   CALL_SERVICE_AND_CHECK_APPROVED(approved,EVP_AEAD_CTX_open(aead_ctx.get(),
       decrypt_output, &out2_len, sizeof(decrypt_output), nullptr, 0, encrypt_output, out_len, nullptr, 0));
   ASSERT_TRUE(check_test(kPlaintext, decrypt_output, sizeof(kPlaintext),
                   "AES-GCM Decryption for Internal IVs"));
-  ASSERT_TRUE(approved);
+  ASSERT_EQ(approved, AWSLC_APPROVED);
 }
 
 #else
```
nebeid, line 178: I suggest you change the comment of the following line (feel free to reword it): 
"Service indicator calls should not be used in non-FIPS builds. However, in that case, if the macro |CALL_SERVICE_AND_CHECK_APPROVED| is used, it will return AWSLC_APPROVED, but the direct calls to |FIPS_service_indicator_xxx| will not indicate approved state."

torben-hansen (approved):

crypto/fipsmodule/service_indicator/internal.h
```diff
@@ -33,8 +32,9 @@ OPENSSL_INLINE void FIPS_service_indicator_update_state(void) { }
 // key rounds.
 void AES_verify_service_indicator(unsigned key_rounds);
 
-// AEAD APIs work with different parameters. Only internal IV for AES-GCM is
-// meant to be approved.
+// The service indicator checks use different parameters for AEAD APIs than
+// those of other AES modes. AES-GCM is approved only with an internal IV
+// <SP 800-38D  and 8.2>.
```
nebeid, line 37:
```suggestion
// , see SP 800-38D Sec 8.2.
```
Is Sec 8.2 the correct section?

nebeid: Also, I think we need to specify 8.2.2 not just 8.2, right?

samuel40791765: Updated

torben-hansen (approved):

nebeid (approved):

PR #248, opened 2021-09-17
Correctly check Jitter read_entropy return value

dkostic:
### Issues:
N/A

### Description of changes: 
Documentation for Jitter read_entropy functions states that the functions return either a negative integer which signifies an error, or the number of generated bytes of entropy. The implementation of the functions however checks if it the requested and the generated number of bytes is the same and returns an error if it's not. Nevertheless, to be on the safe side we now check that Jitter function returns the expected value.

### Call-outs:
N/A

### Testing:


torben-hansen (approved):
lgtm

SalusaSecondus (approved):
LGTM

PR #252, opened 2021-09-20
Import SIKE P434 Round-3 code and add basic unit test

dkostic:
### Issues:
CryptoAlg-838

### Description of changes: 
This PR imports the source code of SIKE from the official repository: https://github.com/microsoft/PQCrypto-SIDH. A basic unit test is also added to verify the functionality of SIKE implementation and the newly added PQ KEM API.

### Call-outs:
N/A

### Testing:
New unit test added.


pq-crypto/pq_kem.c
```diff
@@ -0,0 +1,114 @@
...
+
+int EVP_PQ_KEM_CTX_init(EVP_PQ_KEM_CTX *kem_ctx, const EVP_PQ_KEM *kem) {
+
+  if (kem_ctx == NULL || kem == NULL) {
+    return 0;
+  }
+
+  kem_ctx->kem = kem;
+
+  kem_ctx->public_key    = OPENSSL_malloc(kem->public_key_length);
```
torben-hansen, line 26: Dunno what is going on here with spacing, but it looks funny :)

torben-hansen (approved):

pq-crypto/pq_kem.c
dkostic: I aligned the equal signs, looks nicer to me that way :) if we don't do that in AWS-LC I'll remove the extra space?

torben-hansen: lol - I'm really indifferent

pq-crypto/pq_kem_params_size.h
```diff
@@ -0,0 +1,23 @@
+// -----------------------------------------------------------------------------
```
manastasova, line 1: Can this file be deleted? Probably we can only use the P434_api.h?

pq-crypto/EVP_kem.c
```diff
@@ -6,21 +6,20 @@
 // Abstract: Post-Quantum Algorithm API
 // -----------------------------------------------------------------------------
 
-#include <stdint.h>
-#include <stdlib.h>
 #include "EVP_kem.h"
-#include "sike_r3/sike_internal.h"
-#include "../include/openssl/mem.h"
+#include "pq_kem_params_size.h"
```
manastasova: Only P434_api.h should be enough?

manastasova (approved):

manastasova (approved):

pq-crypto/pq_kem_params_size.h
dkostic: Yes right now we don't need this file, but when we add another PQ KEM we'll need it because all the KEMs define the same macros with the same names so there would be a collision on the names if we don't define them separately in this file.

pq-crypto/EVP_kem.c
dkostic: same answer as on the other comment

PR #259, opened 2021-09-23
Always use SHA-256 for FIPS integrity check

dkostic:
### Issues:
CryptoAlg-854.

### Description of changes: 
FIPS 140-3 requires a Cryptographic Algorithm Self-Test (CAST)
for the approved integrity technique used for the integrity test.
The CAST must be performed to verify the integrity technique itself
and it must happen before the module performs the integrity test.

Since HMAC-SHA2-512 is not tested in our power-on self tests,
and HMAC-SHA2-256 is we change the code to always use SHA-256
based HMAC for integrity check. Also, we reorder the sequence
of function calls to perform the CAST of HMAC-SHA2-256 before
the integrity check.

### Call-outs:
N/A

### Testing:


darylmartin100 (dismissed):

crypto/fipsmodule/bcm.c
```diff
@@ -196,15 +196,10 @@ BORINGSSL_bcm_power_on_self_test(void) {
...
-#if defined(OPENSSL_AARCH64) || defined(OPENSSL_ANDROID)
   uint8_t result[SHA256_DIGEST_LENGTH];
   const EVP_MD *const kHashFunction = EVP_sha256();
-#else
-  uint8_t result[SHA512_DIGEST_LENGTH];
-  const EVP_MD *const kHashFunction = EVP_sha512();
-#endif
 
-  static const uint8_t kHMACKey[64] = {0};
+  static const uint8_t kHMACKey[32] = {0};
```
torben-hansen, line 202: magic number

dkostic: This is the desired length of the HMAC key, it can be any value and we choose 32. If the size is larger than the block size of the used hash function then the key will be hashed before use, if it's smaller then it will be padded with zeroes (https://csrc.nist.gov/csrc/media/publications/fips/198/1/final/documents/fips-198-1_final.pdf). But now I'm thinking to change it back to 64 since SHA-256 block size is 64 bytes and not 32 as I mistakenly assumed.

darylmartin100 (approved):

crypto/fipsmodule/bcm.c
```diff
@@ -237,13 +232,17 @@ BORINGSSL_bcm_power_on_self_test(void) {
 
   const uint8_t *expected = BORINGSSL_bcm_text_hash;
 
-  if (!check_test(expected, result, sizeof(result), "FIPS integrity test")) {
+  // Per FIPS 140-3 we have to perform the CAST of the HMAC used for integrity
```
torben-hansen, line 235: nit: prefer specific references (140-3 is sorta a big doc?)

crypto/fipsmodule/fips_shared_support.c
```diff
@@ -20,13 +20,10 @@
 // that must be replaced with the real value during the build process. This
 // value need only be distinct, i.e. so that we can safely search-and-replace it
 // in an object file.
-const uint8_t BORINGSSL_bcm_text_hash[64];
-const uint8_t BORINGSSL_bcm_text_hash[64] = {
+const uint8_t BORINGSSL_bcm_text_hash[32];
```
torben-hansen, line 23: nit: we have `SHA256_DIGEST_LENGTH`. Might not be available here.

torben-hansen (approved):

PR #260, opened 2021-09-24
[FIPS] Change the reseed interval to 2^24 when in FIPS mode

darylmartin100:
### Issues:
Addresses CryptoAlg-864

### Description of changes: 
When in FIPS mode we use the CPU Jitter entropy source to seed our DRBG. This entropy source is very slow and can incur a cost anywhere between 10-60ms depending on configuration and CPU.  Increasing to 2^24 will amortize the penalty over more requests.  The value was selected by reviewing OpenSSL and SP 800-90B for the max reseed interval (2^48).

### Call-outs:
N/A

### Testing:


crypto/fipsmodule/rand/rand.c
```diff
@@ -48,9 +48,22 @@
...
+// depending on configuration and CPU.  Increasing to 2^24 will amortize the 
+// penalty over more requests.  The value was selected by reviewing OpenSSL and 
+// SP 800-90B for the max reseed interval (2^48)
+//
+// CPU Jitter:  https://www.chronox.de/jent/doc/CPU-Jitter-NPTRNG.html
+// 
 // kReseedInterval is the number of generate calls made to CTR-DRBG before
 // reseeding.
+
+#if defined(BORINGSSL_FIPS)
```
torben-hansen, line 62: Nit: we should try and use `AWSLC_FIPS` in the future. But we will change this in bulk at a later stage.

torben-hansen (approved):

crypto/fipsmodule/rand/rand.c
```diff
@@ -48,9 +48,22 @@
 // This might be a bit of a leap of faith, esp on Windows, but there's nothing
 // that we can do about it.)
 
+// When in FIPS mode we use the CPU Jitter entropy source to seed our DRBG.  
+// This entropy source is very slow and can incur a cost anywhere between 10-60ms
+// depending on configuration and CPU.  Increasing to 2^24 will amortize the 
+// penalty over more requests.  The value was selected by reviewing OpenSSL and
```
nebeid, line 54: I think it would be good to add here that this is the value used in OpenSSL. Is it used in OpenSSL 3.0?

darylmartin100: Yes, this is the value used in OpenSSL 3.0.  I'll add a comment.

nebeid (approved):

torben-hansen (approved):

PR #265, opened 2021-09-30
Don't use Intel-specific intrinsic on the Arm path and force alignment of Arm vector type

torben-hansen:
### Description of changes: 

**Intel-specific intrinsic**

~~`OPENSSL_AARCH64` is used at compile-time to branch away from non-arm supported code. However, this is only enabled for Arm 64-bit platforms:
https://github.com/awslabs/aws-lc/blob/main/include/openssl/base.h#L96~~

~~The 32-bit Arm platform we are looking at makes the compiler define `__ARMEL__`, making us defining `OPENSSL_ARM`:
https://github.com/awslabs/aws-lc/blob/main/include/openssl/base.h#L99.~~

~~Fixed two instances of this.~~

https://github.com/awslabs/aws-lc/pull/165 introduced a breaking change. The CI didn't exercise this code path because it doesn't have an Arm 32-bit w/ Neon dimension. So it didn't pick up that it now uses Intel specific intrinsics.

Revert back to the original implementation for the Arm code path but use the fix in the PR for the x86 code path.

**alignment**

`struct poly` is aligned to 16-byte boundary:
https://github.com/awslabs/aws-lc/blob/main/crypto/hrss/hrss.c#L921

For SSE2 `vec_t` is automatically 16-byte boundary aligned because `__m128i` is automatically 16-byte aligned.

However, for Arm `vec_t` is typedef'ed to `uint16x8_t`, which I can't find any information about is automatically 16-byte aligned.

I can't say if this is required, but we statically assert that `vec_t` and `struct poly` should have the same alignment:
https://github.com/awslabs/aws-lc/blob/main/crypto/hrss/hrss.c#L1232

Force this alignment.

### Testing:
Tested manually that this builds on a Arm 32-bit system that has Arm NEON support.

### DIFF
--- a/crypto/hrss/hrss.c
+++ b/crypto/hrss/hrss.c
@@ -192,6 +192,10 @@ static inline vec_t vec_broadcast_bit(vec_t a) {
     (defined(__ARM_NEON__) || defined(__ARM_NEON))
 
 #define HRSS_HAVE_VECTOR_UNIT
+// The Intel vector type |__m128i| is automatically aligned to a 16-byte
+// boundary. But, I can't find anything information about |uint16x8_t| should
+// necessarily be aligned to a 16-byte boundary. We verify 16-byte alignment in
+// |poly_mul_vec()|, so try to force alignment here.
 typedef uint16x8_t vec_t __attribute__ ((aligned (16)));
 
 // These functions perform the same actions as the SSE2 function of the same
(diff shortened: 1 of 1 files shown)


kkoppolu1:
Looks good. Thanks!

nebeid (approved):

andrew-kaufman (approved):

PR #287, opened 2021-10-15
Make `EVP_PKEY_derive` use the shared secret computation within `fipsmodule`

nebeid:
### Issues:
Addresses #CryptoAlg-835

### Description of changes: 
Before this change, `EVP_PKEY_derive` calls `ECDH_compute_key` which is implemented in `ecdh_extra`, outside the `fipsmodule` folder, passing a NULL pointer as a KDF. With this change, the computation is brought within `fipsmodule`.
Now that the computation has been moved within the `fipsmodule` folder, a service indicator check and test is also added to `EVP_PKEY_derive` to say it's approved.

This is done by:
- factoring out the common code in `ECDH_compute_key_fips` and
  `ECDH_compute_key`, which is outside the module, in a new
  internal function, `ECDH_compute_shared_secret`
- changing the static function `pkey_ec_verify` to call the
  new function.

### Call-outs:
N/A


crypto/fipsmodule/ecdh/ecdh.c
```diff
@@ -77,33 +77,44 @@
...
+  size_t buflen = sizeof(buf);
+  int ret = 0;
+
+  if (!ECDH_compute_shared_secret(buf, &buflen, pub_key, priv_key)) {
     goto end;
   }
 
+  // Lock state here to avoid underlying |SHA*| functions updating the service
+  // indicator state unintentionally.
+  FIPS_service_indicator_lock_state();
```
andrewhop, line 116: Why was this moved down? In theory could ECDH_compute_shared_secret trigger something?

crypto/fipsmodule/evp/p_ec.c
```diff
@@ -155,6 +155,7 @@ static int pkey_ec_derive(EVP_PKEY_CTX *ctx, uint8_t *key,
   eckey = ctx->pkey->pkey.ec;
 
   if (!key) {
+    // This only outputs the expected |keylen| and does no actual crypto.
```
nebeid, line 158: l. 170: may be add a comment before the call to ECDH_compute_shared_secret stating that it's an internal function that does not set the service indicator.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -1153,6 +1153,32 @@ TEST_P(ECDH_ServiceIndicatorTest, ECDH) {
   CALL_SERVICE_AND_CHECK_APPROVED(approved, ASSERT_TRUE(ECDH_compute_key_fips(digest.data(),
                               digest.size(), EC_KEY_get0_public_key(peer_key.get()), our_key.get())));
   ASSERT_EQ(approved, ecdhTestVector.expect_approved);
+
```
nebeid, line 1156: There is a typo up there in ECDSAKeyCheck on l. 964, where it says:
// Test running EVP_PKEY_derive
which I think is supposed to be "keygen"

crypto/fipsmodule/ec/internal.h
```diff
@@ -81,6 +81,13 @@
 extern "C" {
 #endif
 
+// ECDH_compute_key calculates the shared key between |pub_key| and |priv_key|.
```
dkostic, line 84:
```suggestion
// |ECDH_compute_shared_secret| calculates the shared key between |pub_key| and |priv_key|.
```

nebeid: Done.

crypto/fipsmodule/evp/p_ec.c
nebeid: Done.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
nebeid: Done.

PR #299, opened 2021-11-01
Fix Android FIPS build on armv8

samuel40791765:
### Issues:
Resolves https://github.com/awslabs/aws-lc/issues/192

### Description of changes: 
* Clang with the Android NDK doesn't compile with `-Werror` on clang due to implicit addition of `-Wa,noexecstack`. Since `--Wa,--noexecstack` is not used during the preprocessor step (because assembler is not invoked), Clang reports that argument as unused. Added a removal the flag during FIPS builds for Android. 
Context: https://github.com/android/ndk/issues/171

* CMAKE inserts a `\` before whitespaces in arguments, which Android line does not recognize when using
  `add_custom_command`. This caused `${CMAKE_ASM_COMPILER}` to fail during the preprocessing step. Replacing whitespaces with `;` fixes this. 
  Context: https://stackoverflow.com/questions/8925396/why-does-cmake-prefixes-spaces-with-backslashes-when-executing-a-command

### Call-outs:
N/A

### Testing:
No Android CI yet, everything is tested on local with an android docker image.


crypto/fipsmodule/CMakeLists.txt
```diff
@@ -152,9 +159,12 @@ function(cpreprocess dest src)
     set(TARGET "--target=${CMAKE_ASM_COMPILER_TARGET}")
   endif()
 
+  # CMAKE inserts "\" before whitespaces in arguments, which the command line does not recognize when using
+  # |add_custom_command|. Replacing the whitespaces with ";" fixes this.
+  string(REGEX REPLACE "[ ]+" ";" CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS}")
```
bryce-shang, line 164: I think this may have something to do with Android/NDK. Does CMAKE not insert `\` on other platforms and then break other builds?

samuel40791765: I could move this under `if(ANDROID)`. I'll move this along with the other change you suggested.

bryce-shang (approved):

samuel40791765:
When using ${CMAKE_ASM_FLAGS} , `;` replacements are needed to fix `\` whitespace problems.
Right now Android doesn’t work with the original `$CMAKE_ASM_FLAGS` , and will get en error like this:
```
clang: error: no such file or directory: 'MAKE_ASM_FLAGS'
crypto/fipsmodule/CMakeFiles/bcm_hashunset.dir/build.make:130: recipe for target 'crypto/fipsmodule/vpaes-armv8.S.s' failed
make[2]: *** [crypto/fipsmodule/vpaes-armv8.S.s] Error 1
```
But if `${CMAKE_ASM_FLAGS}` is added and the `;` fix is applied to Android only, regular ARM builds will also run into the `\` whitespace problem  like this:
```
[21/480] Generating ghash-neon-armv8.S.s
FAILED: crypto/fipsmodule/ghash-neon-armv8.S.s 
cd /codebuild/output/src558511993/src/github.com/awslabs/aws-lc/test_build_dir/crypto/fipsmodule && /usr/bin/clang-7 \ -Wa,--noexecstack -E ghash-neon-armv8.S -I/codebuild/output/src558511993/src/github.com/awslabs/aws-lc/include > ghash-neon-armv8.S.s
clang: error: no such file or directory: ' -Wa,--noexecstack'
``` 
We’ll have to change `$CMAKE_ASM_FLAGS` to `${CMAKE_ASM_FLAGS}` in order to fix Android, but  `;`  has to be applied to all ARM builds because of this change.

bryce-shang (approved):

torben-hansen (approved):

PR #306, opened 2021-11-10
Removed jitterentropy static library file from install

DominicDams:
The jitterentropy static library file was previously getting installed
when running make install. Since this is just a file we link against and
not a shared library file or a static library file that is part of our
release we don't need to include this upon running make install.

This change removes the install logic from the jitterentropy CMake file
and modifies the crypto CMake file to do private linking against
jitterentropy. In order to use private linking I needed to explicitly
list all the other linking types as public.

### Issues:
Addresses CryptoAlg-882

### Description of changes: 
This change removes the install logic from the jitterentropy CMake file
and modifies the crypto CMake file to do private linking against
jitterentropy. In order to use private linking I needed to explicitly
list all the other linking types as public.

### Call-outs:
None

### Testing:
Locally built and tested. Verified that the `libjitterentropy.a` file was no longer present in the install directory after this change.


torben-hansen (dismissed):
An idea for the future would be to have a simple "link test" using a dependency chain and building an executable. But not sure it is necessary for this change.

crypto/CMakeLists.txt
```diff
@@ -490,14 +490,14 @@ endif()
...
+  target_link_libraries(crypto PUBLIC ws2_32)
 endif()
 if(NOT WIN32 AND NOT ANDROID)
-  target_link_libraries(crypto pthread)
+  target_link_libraries(crypto PUBLIC pthread)
 endif()
 
 if(FIPS)
-  target_link_libraries(crypto jitterentropy)
+    target_link_libraries(crypto PRIVATE jitterentropy)
```
bryce-shang, line 500: * Nit: extra space.
* Why `PRIVATE` has to be used here? Can the undefined(empty) be used?
https://stackoverflow.com/questions/51396608/what-is-default-target-link-libraries-privacy-setting

```
# aws-lc still supports old version CMake.
cmake_minimum_required(VERSION 2.8.12)
```

DominicDams: I'll remove the space there (my settings on VIM apparently modified that automatically). We need private because prior to specifying that this was a private dependency it was behaving as if it was public (complaining that there was no export of these build artifacts). In order to make this PRIVATE we needed to specify all the settings (since you need to either explicitly set the strategy on all of the libraries for a specific target or none of them).

torben-hansen (approved):

bryce-shang (approved):

PR #309, opened 2021-11-19
add `EVP_MD_CTX_set_pkey_ctx` function to support manual construction of `EVP_PKEY_CTX`

samuel40791765:
### Issues:
Resolves #CryptoAlg-899

### Description of changes: 
FIPS 140-3 requires that signing and verification methods hash and sign/verify the digest themselves. s2n currently calls one method to calculate the digest and then another method to generate or verify the signature. They will need to change their signing and verification methods to align with FIPS 140-3's new requirements.
AWS-LC recently went through a FIPS validation process for `EVP_DigestSign/Verify` which correctly implements the new FIPS 140-3 requirements signing and verification. However, after more research on s2n's side, `EVP_DigestSign/Verify` was found to need an additional `EVP_PKEY_CTX` argument during the context initialization process. This argument cannot be set when s2n is initializing the `EVP_MD_CTX` context because the private key is not available during that point in time.
Openssl has a function called `EVP_MD_CTX_set_pkey_ctx` available which allows setting of the private key in the context at a later time. This function is not available in AWS-LC and could fix s2n's issues if added.

* OpenSSL's `EVP_MD_CTX_set_pkey_ctx`
    * https://github.com/openssl/openssl/blob/ab547fc005307ecf48451638e947cdabca147159/crypto/evp/evp_lib.c#L1013-L1030
    * https://github.com/openssl/openssl/blob/bef9b48e5071cdd2b41a4f486d1bcb5e14b2a5c3/crypto/evp/digest.c#L70-L79
    * https://github.com/openssl/openssl/blob/051228353a9842eede597294603cc06a55e3a22c/include/crypto/evp.h#L19-L23
 * s2n's use case: https://gist.github.com/lrstewart/14f72fc7779749312ed65a1b22207d2a#file-test_2-c-L87-L115

### Call-outs:
In order to support use of this behavior, an additional `EVP_MD_CTX_FLAG_KEEP_PKEY_CTX` flag from OpenSSL was added. This flag locks freeing of `ctx->pctx` when the `pctx` is set with `EVP_MD_CTX_set_pkey_ctx`. In these cases, the actual freeing of `pctx` should be done by the user instead. 
However, adding an extra `flag` variable in `EVP_MD_CTX` could break ABI compatibility. The alternative is using `EVP_PKEY_CTX_dup` to duplicate the `pctx` pointer contents to `ctx->pctx` to allow safer freeing of `ctx->pctx` without effecting `pctx`. However, `EVP_PKEY_CTX_dup` would require more memory and sacrifice speed, which isn't ideal for s2n (or aws-lc).

### Testing:
New test dimension `PKEY_CTX_manual`, which tests the manual construction and signing/verifying behavior. 


include/openssl/digest.h
```diff
@@ -300,6 +300,11 @@ OPENSSL_EXPORT void EVP_MD_CTX_set_flags(EVP_MD_CTX *ctx, int flags);
 // EVP_MD_nid calls |EVP_MD_type|.
 OPENSSL_EXPORT int EVP_MD_nid(const EVP_MD *md);
 
+// |EVP_MD_CTX_set_pkey_ctx| sets a |EVP_PKEY_CTX| value to |EVP_MD_CTX|'s
+// |EVP_PKEY_CTX|. This function was deprecated from OpenSSL in BoringSSL, but
```
nebeid, line 304:
```suggestion
// |EVP_MD_CTX_set_pkey_ctx| sets |EVP_MD_CTX|'s |EVP_PKEY_CTX| to |pctx|. 
// This function was deprecated from OpenSSL in BoringSSL, but
```

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -1123,6 +1123,79 @@ TEST(EVPExtraTest, Ed25519Keygen) {
                                reinterpret_cast<const uint8_t *>("hello"), 5));
 }
 
+// Test that |EVP_DigestSignFinal| and |EVP_DigestSignVerify| works with a
```
nebeid, line 1126:
```suggestion
// Test that |EVP_DigestSignFinal| and |EVP_DigestSignVerify| work with a
```

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -1123,6 +1123,79 @@ TEST(EVPExtraTest, Ed25519Keygen) {
...
+
+  // Manual construction for verification.
+  ASSERT_TRUE(EVP_PKEY_verify_init(pctx.get()));
+  ASSERT_TRUE(EVP_PKEY_CTX_set_signature_md(pctx.get(), EVP_sha256()));
+  EVP_MD_CTX_set_pkey_ctx(ctx.get(), pctx.get());
+
+  ASSERT_TRUE(EVP_DigestVerifyFinal(ctx.get(), sig.data(), sig_len));
+
+
+  // Test the behavior with EC Keys.
```
nebeid, line 1162: I think a similar comment with RSA is needed above.

crypto/fipsmodule/digest/digest.c
```diff
@@ -97,7 +99,9 @@ int EVP_MD_CTX_cleanup(EVP_MD_CTX *ctx) {
   OPENSSL_free(ctx->md_data);
 
   assert(ctx->pctx == NULL || ctx->pctx_ops != NULL);
-  if (ctx->pctx_ops) {
+  // pctx should be freed by the user of EVP_MD_CTX if
```
nebeid, line 102: When is this behaviour needed: when you call EVP_MD_CTX_cleanup(), but you expect it to leave behind `ctx->pctx`? What is expected from the `EVP_MD_CTX_init()` call that follows? Is it only resetting the digest operation? Will the same hash algorithm stay in the ctx, or that's also wiped? I think at least we need comments about the behaviour.
We also need test cases for it based on what we're expecting to happen.

Do we want to add service indicator tests to ensure that this manual setting of EVP_PKEY_CTX doesn't affect the `Approved` status of EVP_DigestSign/VerifyFinal?

crypto/fipsmodule/evp/digestsign.c
```diff
@@ -275,3 +275,22 @@ int EVP_DigestVerify(EVP_MD_CTX *ctx, const uint8_t *sig, size_t sig_len,
   }
   return ret;
 }
+
+void EVP_MD_CTX_set_pkey_ctx(EVP_MD_CTX *ctx, EVP_PKEY_CTX *pctx)
+{
+    // pctx could be null, so we have to deal with the cleanup job here.
+    if (!(ctx->flags & EVP_MD_CTX_FLAG_KEEP_PKEY_CTX)) {
```
torben-hansen, line 282: I wonder how you will ever reach this code?

Let's say you execute `EVP_MD_CTX_set_pkey_ctx` with `pctx != NULL`. Okay, we now reference this through `ctx->pctx` and the `EVP_MD_CTX_FLAG_KEEP_PKEY_CTX` bit is set.  

I consider `env_md_ctx_st` nominally opaque (in practice it is not, because the definition is in `digest.h`), so there is no way to unset the `EVP_MD_CTX_FLAG_KEEP_PKEY_CTX` bit without calling this function with `pctx = NULL`. But then you would have had to free the memory pointed to by `ctx->pct` a priori to the function execution. How to do this without access to struct flag??

I guess the point is that the consumer has to manage a pointer to `pctx` in addition to `ctx->pctx`. Loosing this pointer, would mean you leak memory.

Having a function that mutates `EVP_MD_CTX_FLAG_KEEP_PKEY_CTX` can also be hazardous... It would be easy to end up with dangling pointers... In fact, if the consumer maintains a second reference, `ctx->pctx` will be dangling if that reference is used to free the memory.

You should explain this behaviour in the function description, as well as who has the responsibility of this memory life-cycle.

include/openssl/digest.h
```diff
@@ -300,6 +300,11 @@ OPENSSL_EXPORT void EVP_MD_CTX_set_flags(EVP_MD_CTX *ctx, int flags);
 // EVP_MD_nid calls |EVP_MD_type|.
 OPENSSL_EXPORT int EVP_MD_nid(const EVP_MD *md);
 
+// |EVP_MD_CTX_set_pkey_ctx| sets a |EVP_PKEY_CTX| value to |EVP_MD_CTX|'s
```
torben-hansen, line 303: The `EVP_PKEY_CTX` object needs to have been initialised with an operation before associating it with the `EVP_MD_CTX` object.

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -1123,6 +1123,79 @@ TEST(EVPExtraTest, Ed25519Keygen) {
...
+
+  bssl::UniquePtr<EVP_PKEY> pkey = LoadExampleRSAKey();
+  ASSERT_TRUE(pkey);
+  bssl::UniquePtr<EVP_PKEY_CTX> pctx(EVP_PKEY_CTX_new(pkey.get(), nullptr));
+  ASSERT_TRUE(pctx);
+
+  // Manual construction for signing.
+  ASSERT_TRUE(EVP_PKEY_sign_init(pctx.get()));
+  ASSERT_TRUE(EVP_PKEY_CTX_set_signature_md(pctx.get(), EVP_sha256()));
+  EVP_MD_CTX_set_pkey_ctx(ctx.get(), pctx.get());
```
torben-hansen, line 1144: In this call, should we verify that `ctx` and `pctx` has the same associated hash function?

In theory, you could associate different hash functions to each object. Might be nice if we can prevent that from happening. It would lead to an obvious error when trying to finalise the sign (at least for RSA, the EC implementation doesn't care).

crypto/fipsmodule/evp/digestsign.c
torben-hansen: And some tests for these edge cases...

samuel40791765: Added a small description in the comments, let me know if I need to add anything more.

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -1123,6 +1123,105 @@ TEST(EVPExtraTest, Ed25519Keygen) {
...
+  ASSERT_TRUE(pctx.get());
+
+  // Test assigning a NULL |pctx| with |EVP_MD_CTX_set_pkey_ctx|.
+  bssl::ScopedEVP_MD_CTX md_ctx;
+  bssl::UniquePtr<EVP_PKEY_CTX> null_pctx(nullptr);
+  ASSERT_FALSE(null_pctx.get());
+  // Assign a non-NULL pctx first.
+  EVP_MD_CTX_set_pkey_ctx(md_ctx.get(), pctx.get());
+  ASSERT_TRUE(md_ctx.get()->pctx);
+  // Set again with NUll |pctx|.
```
nebeid, line 1217: I think we need to call `cleanup` here and `ASSERT_TRUE(md_ctx.get()->pctx)`, if I understood correctly what will happen.
May be also put a comment that the flag `EVP_MD_CTX_FLAG_KEEP_PKEY_CTX` is set so `md_ctx.get()->pctx()` cannot be freed.

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -1123,6 +1123,105 @@ TEST(EVPExtraTest, Ed25519Keygen) {
...
+  EVP_MD_CTX_set_pkey_ctx(md_ctx.get(), null_pctx.get());
+  ASSERT_FALSE(md_ctx.get()->pctx);
+  ASSERT_FALSE(null_pctx.get());
+  // Check that clean up works.
+  ASSERT_TRUE(EVP_MD_CTX_cleanup(md_ctx.get()));
```
nebeid, line 1222: I think it would be good to make all the ASSERTS we expect from cleanup like the ones on L1203-L1208. This proves the flag was removed.

include/openssl/digest.h
```diff
@@ -300,6 +300,24 @@ OPENSSL_EXPORT void EVP_MD_CTX_set_flags(EVP_MD_CTX *ctx, int flags);
 // EVP_MD_nid calls |EVP_MD_type|.
 OPENSSL_EXPORT int EVP_MD_nid(const EVP_MD *md);
 
+// This function was deprecated from OpenSSL in BoringSSL, but was brought back
+// to AWS-LC.
+//
+// If |pctx| is not NULL, |EVP_MD_CTX_set_pkey_ctx| sets |ctx|'s |EVP_PKEY_CTX|
+// reference to |pctx|. The |EVP_PKEY_CTX| object |pctx| needs to have been
+// initialised before associating it with |ctx|. The hash functions associated
+// to |ctx| and |pctx| must be equal. Once |EVP_MD_CTX_set_pkey_ctx| is called,
```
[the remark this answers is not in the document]
samuel40791765: It's related to the discussion we were having here: https://github.com/awslabs/aws-lc/pull/309#discussion_r754711430  The actual signing/verifying enforces that the hash must be equal when we're using RSA. An error pops up when it's not aligned. 
I'm skeptical on adding an extra check in this function since it's from Openssl, and we may want to keep most of the behavior the same for backwards compatibility.

crypto/evp_extra/evp_extra_test.cc
samuel40791765: I think `md_ctx.get()->pctx` will be `FALSE` because the reference pointer from `md_ctx.get()` is still destroyed. What the `EVP_MD_CTX_FLAG_KEEP_PKEY_CTX` does is that it keeps the `pctx` that was originally assigned to it. If the flag isn't added, the externally assigned `pctx` will be inadvertently freed when `EVP_MD_CTX_cleanup` is called.
Based on that assumption, I added these two lines at L1218:
```
  ASSERT_TRUE(EVP_MD_CTX_cleanup(md_ctx.get()));
  ASSERT_TRUE(md_ctx.get()->pctx);
```
The test failed with:
```
/home/ubuntu/sachiang/workplace/repo/aws-lc/crypto/evp_extra/evp_extra_test.cc:1219: Failure
Value of: md_ctx.get()->pctx
  Actual: false
Expected: true
```

nebeid: Ah, so that's the case that Torben mentioned will leave a dangling pointer if `EVP_MD_CTX_cleanup` is called without freeing `pctx`, right?

include/openssl/digest.h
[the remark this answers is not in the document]
samuel40791765: After some discussion with @torben-hansen, we'll be adding an extra check to enforce the equal hash.

crypto/evp_extra/evp_extra_test.cc
samuel40791765: Yes, basically the `EVP_MD_CTX` will be destroyed safely. But `pctx` that was assigned will have to be taken care of by the caller.

include/openssl/digest.h
```diff
@@ -303,20 +303,19 @@ OPENSSL_EXPORT int EVP_MD_nid(const EVP_MD *md);
...
-// If |pctx| is not NULL, |EVP_MD_CTX_set_pkey_ctx| sets |ctx|'s |EVP_PKEY_CTX|
-// reference to |pctx|. The |EVP_PKEY_CTX| object |pctx| needs to have been
-// initialised before associating it with |ctx|. The hash functions associated
-// to |ctx| and |pctx| must be equal. Once |EVP_MD_CTX_set_pkey_ctx| is called,
-// the callee is responsible for freeing |pctx|. Calling |EVP_MD_CTX_cleanup|
-// will not free |pctx|.
+// |EVP_MD_CTX_set_pkey_ctx| sets |ctx|'s |EVP_PKEY_CTX| reference to |pctx|.
+// The |EVP_PKEY_CTX| object |pctx| needs to have been initialised before
+// associating it with |ctx|. The hash functions associated to |ctx| and |pctx|
+// must be equal. Once |EVP_MD_CTX_set_pkey_ctx| is called, the callee is
```
nebeid, line 309: I think "callee" here should be "caller".

include/openssl/digest.h
```diff
@@ -303,20 +303,19 @@ OPENSSL_EXPORT int EVP_MD_nid(const EVP_MD *md);
...
-// The callee is responsible for the memory associated to |ctx|'s |EVP_PKEY_CTX|
-// reference, until |EVP_MD_CTX_set_pkey_ctx| is called with |pctx| equal to
-// NULL.
+// The caller is responsible for the memory associated to |ctx|'s |EVP_PKEY_CTX|
+// reference, until |EVP_MD_CTX_set_pkey_ctx| is called again with |pctx| = NULL.
 //
 // |EVP_MD_CTX_set_pkey_ctx| will overwrite any |EVP_PKEY_CTX| object associated
-// to |ctx| if it was associated through a previous |EVP_MD_CTX_set_pkey_ctx|
-// call |ctx|.
+// to |ctx|. If it was associated through a previous |EVP_MD_CTX_set_pkey_ctx|
```
nebeid, line 317: If it was *not* associated.

include/openssl/digest.h
```diff
@@ -303,20 +303,19 @@ OPENSSL_EXPORT int EVP_MD_nid(const EVP_MD *md);
...
+// associating it with |ctx|. The hash functions associated to |ctx| and |pctx|
+// must be equal. Once |EVP_MD_CTX_set_pkey_ctx| is called, the callee is
+// responsible for freeing |pctx|. Calling |EVP_MD_CTX_cleanup| will not free
+// |pctx|.
 //
-// The callee is responsible for the memory associated to |ctx|'s |EVP_PKEY_CTX|
-// reference, until |EVP_MD_CTX_set_pkey_ctx| is called with |pctx| equal to
-// NULL.
+// The caller is responsible for the memory associated to |ctx|'s |EVP_PKEY_CTX|
+// reference, until |EVP_MD_CTX_set_pkey_ctx| is called again with |pctx| = NULL.
```
nebeid, line 314: Doesn't the first part here mean "The caller is responsible for the memory associated to |pctx|"?

samuel40791765:
Rolling back rsa check for `EVP_MD_CTX_set_pkey_ctx` because it's a `void` and it won't be possible for the caller to differentiate if the function has failed or not.

nebeid (dismissed):

samuel40791765:
>  Should we keep the check: it will put an error in the error log. May be in the service indicator check, we can make the check again and set the status to NOT_APPROVED
For RSA, the signing will not be done if the hash digest lengths are not equal, so the function will fail and the status will not be approved. That also means that the service indicator code will never be reached, because we insert the service indicator checks at the end of the function. 
That being said, would it still be reasonable to put the check there? There could be a bit of value for FIPS checking.

nebeid:
> > Should we keep the check: it will put an error in the error log. May be in the service indicator check, we can make the check again and set the status to NOT_APPROVED
> 
> For RSA, the signing will not be done if the hash digest lengths are not equal, so the function will fail and the status will not be approved. That also means that the service indicator code will never be reached, because we insert the service indicator checks at the end of the function. That being said, would it still be reasonable to put the check there? There could be a bit of value for FIPS checking.

I think it would be good to add the check, especially in case more digests are added with same length but not same algorithm. Also, I understand EC will not care, but I don't know if it will perform correctly or not.

nebeid (approved):

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -222,6 +222,12 @@ void EVP_PKEY_keygen_verify_service_indicator(const EVP_PKEY *pkey) {
 void DigestSign_verify_service_indicator(const EVP_MD_CTX *ctx) {
   if(ctx->pctx->pmeth->pkey_id == EVP_PKEY_RSA ||
       ctx->pctx->pmeth->pkey_id == EVP_PKEY_RSA_PSS) {
+    // Hash digest used in the private key should be of the same type.
+    const EVP_MD *pctx_md;
+    if(EVP_PKEY_CTX_get_signature_md(ctx->pctx, &pctx_md) &&
+        (EVP_MD_CTX_md(ctx)->type != pctx_md->type)) {
+      return;
```
torben-hansen, line 229: Is there a test case running over this code path? Same for `DigestVerify_verify_service_indicator`

torben-hansen (approved):

PR #312, opened 2021-11-19
Workaround gcc4 bug

torben-hansen:
### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 

### Call-outs:

### Testing:


DominicDams:
We should probably have a comment somewhere in the code explaining why we need to move all this code off of the stack.

DominicDams:
I also wonder if it would be worth having this compile to stack/head code depending on what the gcc version is. It would make the code potentially a little faster on modern compilers (probably not a large difference but some) but it would make this code more complicated. If we plan to stop supporting these old compilers at some point and will revert this change then I think it's probably worth eating a very minor performance degradation in the meantime.

torben-hansen:
> We should probably have a comment somewhere in the code explaining why we need to move all this code off of the stack.

You can follow the blame for context. But I should add a "todo text" saying should be removed as a reminder for the distant future.

> I also wonder if it would be worth having this compile to stack/heap code depending on what the gcc version is. It would make the code potentially a little faster on modern compilers (probably not a large difference but some) but it would make this code more complicated. If we plan to stop supporting these old compilers at some point and will revert this change then I think it's probably worth eating a very minor performance degradation in the meantime.

I tried addressing that point, maybe it didn't make it into the description (added retrospectively):
I prefer static allocation. However, how would we implement two different code-paths with one doing static allocation on the stack, while the other doing dynamic allocation on the heap? I can't think of an easy way to do this, when I have to define wrapper types for each array type I want to allocate on the heap and you have perform various dereferences. IMO it will be a mess of if/def's everywhere. With generics, maybe this could be done a "pretty" way. From a complexity point of view, I went with the simple solution to just move it all to a method that should work for all compilers.

I didn't detect any perf regressions (in my, arguably, limited measurement); see description

DominicDams:
Alright seems good then, if we get a TODO or something in the comments then I'll be happy to approve this PR :)

torben-hansen:
> Alright seems good then, if we get a TODO or something in the comments then I'll be happy to approve this PR :)

Added some TODO text. And cooked up some simple tests.

dkostic (dismissed):

crypto/mem.c
```diff
@@ -165,6 +169,42 @@ void *OPENSSL_malloc(size_t size) {
...
+    return NULL;
+  }
+
+  // Memory is organised as follows
+  // +--------+----------------+-------------+---------------+
+  // | poison | offset padding | offset size | memory region |
+  // +--------+----------------+-------------+---------------+
+  // poison: added by |OPENSSL_malloc|.
+  // offset padding: offset bytes to align memory at |alignment| boundary.
+  // offset size: number of offset bytes needed, plus |ALIGN_OFFSET_SIZE|.
```
nebeid, line 189: If I understood correctly, then I'd suggest:
```suggestion
  // offset size: number of bytes in offset padding. This location is of size |ALIGN_OFFSET_SIZE|.
```

include/openssl/mem.h
```diff
@@ -164,6 +164,27 @@ OPENSSL_EXPORT void CRYPTO_free(void *ptr, const char *file, int line);
...
+
+// Internal function.
+// Similar to |OPENSSL_malloc| but also aligns the resulting pointer to a
+// |alignment|-boundary. |alignment| must be a power of 2 with a maximum value
+// of 128 and stricly larger than 0. Memory must be free'd using
+// |OPENSSL_align_free_internal|.
+OPENSSL_EXPORT void * OPENSSL_malloc_align_internal(size_t size, size_t alignment);
+
+// Internal function.
+// OPENSSL_free does nothing if |aligned_ptr| is NULL. Otherwise it zeros out
```
nebeid, line 183:
```suggestion
// OPENSSL_free_align_internal does nothing if |aligned_ptr| is NULL. Otherwise it zeros out
```

PR #313, opened 2021-11-19
Fips version string

DominicDams:
### Issues:
Resolves CryptoAlg-896

### Description of changes: 
In order to be FIPS certified we need an API that provides a version string that will differ between fips and non fips builds. This introduces such an API

### Call-outs:

### Testing:
Manually tested the output and introduced a new unit test.


crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -2006,6 +2006,13 @@ TEST(ServiceIndicatorTest, DRBG) {
   CTR_DRBG_clear(&drbg);
 }
 
+// Verifies that the openssl_version_string is as expected.
+// Since this is running in FIPS mode it should end in FIPS
+// Update this when the OpenSSL version number is modified
+TEST(ServiceIndicatorTest, OpenSSLVersionString) {
+  ASSERT_STREQ(openssl_version_string(), "OpenSSL 1.1.1 (compatible; BoringSSL) FIPS");
```
torben-hansen, line 2013: boringssl?

DominicDams: > boringssl?

That's what we have in our OPENSSL_VERSION_TEXT at the moment. I could change it though.

torben-hansen: We should define something AWS-LC specific. It is confusing to keep using OpenSSL and BoringSSL names IMO.

There are no dependencies on this atm, so we can make a good choice up-front disconnecting from upstream.

Probably need to keep `OPENSSL_VERSION_TEXT` for compatibility though.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -7,6 +7,10 @@
 
 #if defined(AWSLC_FIPS)
 
+const char* openssl_version_string(void) {
```
torben-hansen, line 10: Propose re-factoring to a single function that returns a concatenation: `AWSLC_VERSION_TEXT FIPS_SUFFIX`.

You can define `FIPS_SUFFIX` to be empty if `AWSLC_FIPS` is not defined.

Just came up with some quick variable names.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
DominicDams: If we want an `AWSLC_VERSION_TEXT` should we stick with the same version number we're using for `OPENSSL_VERSION_TEXT` or should we change it to something else? We could pick our own numbering scheme but that might cause more confusion.

crypto/fipsmodule/service_indicator/service_indicator.c
DominicDams: I think either way this will end up with about the same level of complexity, but I can make the change since I don't feel super strongly about it anyways

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -2043,4 +2050,11 @@ TEST(ServiceIndicatorTest, BasicTest) {
           kPlaintext, sizeof(kPlaintext), nullptr, 0));
   ASSERT_EQ(approved, AWSLC_APPROVED);
 }
+
+// Verifies that the openssl_version_string is as expected.
```
torben-hansen, line 2054:
```suggestion
// Verifies that the awslc_version_string is as expected.
```

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -2006,6 +2006,13 @@ TEST(ServiceIndicatorTest, DRBG) {
   CTR_DRBG_clear(&drbg);
 }
 
+// Verifies that the openssl_version_string is as expected.
```
torben-hansen, line 2009:
```suggestion
// Verifies that the awslc_version_string is as expected.
```

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -2006,6 +2006,13 @@ TEST(ServiceIndicatorTest, DRBG) {
...
+// Since this is running in FIPS mode it should end in FIPS
+// Update this when the OpenSSL version number is modified
```
torben-hansen, line 2011:
```suggestion
// Update this when the AWS-LC version number is modified
```

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -2043,4 +2050,11 @@ TEST(ServiceIndicatorTest, BasicTest) {
           kPlaintext, sizeof(kPlaintext), nullptr, 0));
   ASSERT_EQ(approved, AWSLC_APPROVED);
 }
+
+// Verifies that the openssl_version_string is as expected.
+// Since this is not running in FIPS mode it shouldn't end in FIPS
+// Update this when the OpenSSL version number is modified
```
torben-hansen, line 2056:
```suggestion
// Update this when the AWS-LC version number is modified
```

include/openssl/base.h
```diff
@@ -203,6 +203,11 @@ extern "C" {
 
 #define AWSLC_API_VERSION 16
 
+// In order to keep track of any changes to AWLC we introduce a string that tracks with the public Github
+// release version at https://github.com/awslabs/aws-lc
+
+#define AWSLC_VERSION_NUMBER_STRING "0.0.2"
```
torben-hansen, line 209: Do we also need a version string that can be processed at compile-time? You can't really parse this string. Similar to `OPENSSL_VERSION`

DominicDams: I have this version string so that we only need to change this in one place to get the code change done (we also need to change the tests but I think this simplifies things)

torben-hansen: I don't mind having a version string and a version number.
How do you parse this string at compile-time?
With a version number you can branch at compile-time e.g. https://github.com/aws/s2n-tls/blob/main/crypto/s2n_openssl.h#L36
I dunno if you can do something similar with the string version. Or we need another mechanism.

darylmartin100, line 209: FYI:  The version string we sent over to NIST is 1.0.0.  Could you update the value?

DominicDams: Talked offline, planning to do a quick follow up PR to update this string on the fips branch

torben-hansen (approved):

darylmartin100 (approved):

PR #323, opened 2021-12-02
Implementation for alternative solution to gcc4.1 bug work-around

torben-hansen:
### Description of changes: 

Solution that was discarded: https://github.com/awslabs/aws-lc/pull/312

Ancient gcc compilers (up to (not including) version 4.6) contain a bug that cause wrong alignment for stack allocated memory even though it was tagged with an appropriate specifier `alignas`/`align` attribute. The bug is explained in more depth here: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=16660

To work around this bug, we use a temporary byte buffer. Say we want a pointer `aligned_ptr` of type `type_aligned_ptr`. We then allocate a buffer of bytes, find the correct alignment and cast the remaining bytes to the desired type.

```
stack_align_type bufffer[align_boundary + size of object(s)]
// align pointer |buffer| to `align_boundary` boundary.
// use align_pointer() for this.
// Aligned pointer is `aligned_buffer_ptr`.
type_aligned_ptr *aligned_ptr = (type_aligned_ptr *) aligned_buffer_ptr
```

`stack_align_tyoe` is a type of size 8 bits such that arithmetic in pointer-space moves 1 byte per basic increment. So for this, we choose `uint8_t`. `type_aligned_ptr` is the type the object the aligned pointer points to. The remaining memory, after aligning using `align_pointer`, should be sufficient to hold the object(s) since we allocated an extra `align_boundary` bytes in the buffer array (in theory we could do with `align_boundary` - 1).

Strongly recommend using modern compilers.

### Testing:
Try out this alignment for different `align_boundary` and object sizes. All test cases are generated using a Python script and split in two because the virtual functions produced by a test fixture becomes too big.


crypto/compiler_test.cc
```diff
@@ -174,3 +174,534 @@ TEST(CompilerTest, PointerRepresentation) {
...
+//  MACRO_NAME = 'CHECK_STACK_ALIGNMENT'
+//  type_names = ['uint8_t', 'char', 'struct_array_st', 'union_array_type']
+//  power_of_twos = [1, 2, 4, 8, 16, 32, 64]
+//  for type_name in type_names:
+//    for power_of_two in power_of_twos:
+//      for memory_size in range(1, power_of_two):
+//        print '{}({}, {}, {})'.format(MACRO_NAME, type_name, power_of_two, memory_size)
+
+// Windows doesn't like the big virtual functions produced. So, split this into
+// several test fixtures.
```
DominicDams, line 218: Nit: If you need to split this into multiple tests, you might as well split it into 4 tests with one type per test.

DominicDams (dismissed):

dkostic (dismissed):
Approving the PR with hope that one day we'll drop gcc4.1 support and be able to revert the change!

DominicDams (approved):

torben-hansen:
Performance sniff test. On the face, looks (marginally?) more stable than #312.

**With patch**
```
% for i in `seq 1 4`; do ./tool/bssl speed -filter "ECDH P-256"; done
Did 17000 ECDH P-256 operations in 1034097us (16439.5 ops/sec)
Did 17000 ECDH P-256 operations in 1021496us (16642.3 ops/sec)
Did 17000 ECDH P-256 operations in 1021955us (16634.8 ops/sec)
Did 17000 ECDH P-256 operations in 1021462us (16642.8 ops/sec)

% for i in `seq 1 4`; do ./tool/bssl speed -filter "ECDSA P-256"; done
Did 51000 ECDSA P-256 signing operations in 1005729us (50709.5 ops/sec)
Did 19000 ECDSA P-256 verify operations in 1047675us (18135.4 ops/sec)
Did 51000 ECDSA P-256 signing operations in 1009842us (50502.9 ops/sec)
Did 19000 ECDSA P-256 verify operations in 1043844us (18202.0 ops/sec)
Did 51000 ECDSA P-256 signing operations in 1005964us (50697.6 ops/sec)
Did 19000 ECDSA P-256 verify operations in 1049384us (18105.9 ops/sec)
Did 51000 ECDSA P-256 signing operations in 1005687us (50711.6 ops/sec)
Did 19000 ECDSA P-256 verify operations in 1046303us (18159.2 ops/sec)
```

**Without patch**
```
% for i in `seq 1 4`; do ./tool/bssl speed -filter "ECDH P-256"; done
Did 17000 ECDH P-256 operations in 1023951us (16602.4 ops/sec)
Did 17000 ECDH P-256 operations in 1024468us (16594.0 ops/sec)
Did 17000 ECDH P-256 operations in 1042906us (16300.6 ops/sec)
Did 17000 ECDH P-256 operations in 1024713us (16590.0 ops/sec)

% for i in `seq 1 4`; do ./tool/bssl speed -filter "ECDSA P-256"; done
Did 51000 ECDSA P-256 signing operations in 1008289us (50580.7 ops/sec)
Did 19000 ECDSA P-256 verify operations in 1046213us (18160.7 ops/sec)
Did 51000 ECDSA P-256 signing operations in 1007030us (50644.0 ops/sec)
Did 18000 ECDSA P-256 verify operations in 1015836us (17719.4 ops/sec)
Did 51000 ECDSA P-256 signing operations in 1009448us (50522.7 ops/sec)
Did 19000 ECDSA P-256 verify operations in 1048538us (18120.5 ops/sec)
Did 51000 ECDSA P-256 signing operations in 1008348us (50577.8 ops/sec)
Did 19000 ECDSA P-256 verify operations in 1047263us (18142.5 ops/sec)
```

crypto/fipsmodule/bn/bn_test.cc
```diff
@@ -2792,8 +2792,18 @@ TEST_F(BNTest, RSAZABI) {
     return;
   }
 
-  alignas(64) BN_ULONG table[32 * 18] = {0};
-  alignas(64) BN_ULONG rsaz1[40], rsaz2[40], rsaz3[40], n_rsaz[40];
+  stack_align_type buffer_table[64 + (sizeof(BN_ULONG) * (32*18))] = {0};
```
dkostic, line 2795: Isn't adding 63 bytes sufficient for the alignment?

crypto/fipsmodule/ec/p256-nistz.c
```diff
@@ -231,12 +232,13 @@ static void ecp_nistz256_windowed_mul(const EC_GROUP *group, P256_POINT *r,
   ecp_nistz256_point_double(&row[16 - 1], &row[8 - 1]);
 
   BN_ULONG tmp[P256_LIMBS];
-  alignas(32) P256_POINT h;
+  stack_align_type buffer_h[32 + sizeof(P256_POINT)];
```
dkostic, line 235: I know this code was here before, but just curious, do you know why is the table aligned to 64  bytes but `h` to 32 bytes?

dkostic (commented):
Looks good to me, thanks for the newly added tests and the performance sniff test!

crypto/fipsmodule/bn/bn_test.cc
torben-hansen: technically, yes. And practically, for that matter. When I wrote this, I guess I just want to make it obvious, this was the desired alignment. But you could drop it be 1.

crypto/fipsmodule/ec/p256-nistz.c
torben-hansen: Good question - I never looked into why.

DominicDams (approved):

dkostic (approved):

PR #324, opened 2021-12-02
Abort when RSA or EC key generation fail

nebeid:
### Issues:
Resolves #CryptoAlg-908

### Description of changes: 
It's required that the FIPS module aborts when PCT tests fail in `EC_KEY_generate_key_fips()` and `RSA_check_fips()`.
This change inserts the abort on failure after a certain number of attempts in the case of a FIPS build.

- In RSA, there existed a number of attempts already in the key generation.
There were also tests for the failure cases.
These tests are here repeated as part of the RSADeathTest suite when
built with -DFIPS=1 using ASSERT_DEATH.

- In EC, a number of attempts is introduced around the key generation and check.

### Call-outs:
There are currently no tests for the failure cases in EC.

### DIFF
--- a/crypto/internal.h
+++ b/crypto/internal.h
@@ -937,10 +937,13 @@ static inline uint64_t CRYPTO_rotr_u64(uint64_t value, int shift) {
 // FIPS functions.
 
 #if defined(BORINGSSL_FIPS)
+#define MAX_PCT_ATTEMPTS  5
 // BORINGSSL_FIPS_abort is called when a FIPS power-on or continuous test
 // fails. It prevents any further cryptographic operations by the current
 // process.
 void BORINGSSL_FIPS_abort(void) __attribute__((noreturn));
+#else
+#define MAX_PCT_ATTEMPTS  1
 #endif
 
 // boringssl_fips_self_test runs the FIPS KAT-based self tests. It returns one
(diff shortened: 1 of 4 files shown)


dkostic (approved):

crypto/internal.h
```diff
@@ -937,10 +937,13 @@ static inline uint64_t CRYPTO_rotr_u64(uint64_t value, int shift) {
 // FIPS functions.
 
 #if defined(BORINGSSL_FIPS)
+#define MAX_PCT_ATTEMPTS  5
```
samuel40791765, line 940: Is the number `5` a FIPS requirement or a number we decided for ourselves?

crypto/rsa_extra/rsa_test.cc
```diff
@@ -639,6 +639,7 @@ TEST(RSATest, BadExponent) {
   ERR_clear_error();
 }
 
+#if !defined(BORINGSSL_FIPS)
```
samuel40791765, line 642: We could think about using the `AWSLC_FIPS` macro we have? 
That might diverge from the FIPS ifdef in the `crypto/internal.h` file we're  changing in this PR though. But then again, the separate macros could also help keep track of which ifdefs AWS-LC created.

DominicDams (approved):
Looks good, I'd be interested to see if we can move to our own macro here, but I don't think that's a blocker.

crypto/rsa_extra/rsa_test.cc
nebeid: Done in PR#326

PR #327, opened 2021-12-06
Add Arm CPU capabilities environment variable override

dkostic:
### Issues:
N/A

### Description of changes: 
Add Arm CPU capabilities environment variable override.

### Call-outs:
The effect of the change can be checked for example by comparing the speed of SHA-256 when run with `bssl speed SHA-256` and `export OPENSSL_armcap_P=0 && bssl speed SHA-256`.

### Testing:

### DIFF
--- a/crypto/cpu-aarch64-linux.c
+++ b/crypto/cpu-aarch64-linux.c
@@ -96,12 +96,20 @@ void OPENSSL_cpuid_setup(void) {
     OPENSSL_armcap_P |= ARMV8_SHA256;
   }
 
+  // OPENSSL_armcap_P is a 32-bit, unsigned value which may start with "0x" to
+  // indicate a hex value. Prior to the 32-bit value, a '~' or '|' may be given.
+  //
+  // If the '~' prefix is present:
+  //   the value is inverted and ANDed with the probed CPUID result
+  // If the '|' prefix is present:
+  //   the value is ORed with the probed CPUID result
+  // Otherwise:
+  //   the value is taken as the result of the CPUID
   const char *env;
   env = getenv("OPENSSL_armcap_P");
-  if (env == NULL) {
-    return;
+  if (env != NULL) {
+    handle_cpu_env(&OPENSSL_armcap_P, env);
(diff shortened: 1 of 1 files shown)


crypto/cpu-aarch64-linux.c
```diff
@@ -57,6 +95,13 @@ void OPENSSL_cpuid_setup(void) {
   if (hwcap & kSHA256) {
     OPENSSL_armcap_P |= ARMV8_SHA256;
   }
+
+  const char *env;
+  env = getenv("OPENSSL_armcap_P");
+  if (env == NULL) {
+    return;
+  }
+  handle_cpu_env(&OPENSSL_armcap_P, env);
```
DominicDams, line 104: Maybe instead we should do:

```
if (env != NULL) {
    handle_cpu_env(&OPENSSL_armcap_P, env);
}
```

For additional clarity

DominicDams (approved):

crypto/cpu-aarch64-linux.c
```diff
@@ -19,13 +19,51 @@
...
+#include <string.h>
+
 #include <openssl/arm_arch.h>
 
 #include "internal.h"
 
 
 extern uint32_t OPENSSL_armcap_P;
 
+// handle_cpu_env applies the value from |in| to the CPUID values in |out[0]|
```
nebeid, line 38: Could this comment include example usage?
To set up 2 capabilities, how is it done? I'm guessing by putting directly the value that represents both.
How can `~` and `|` be used?

darylmartin100 (dismissed):

crypto/cpu-aarch64-linux.c
dkostic: done

crypto/cpu-aarch64-linux.c
dkostic: done

darylmartin100 (approved):

DominicDams (approved):

PR #330, opened 2021-12-07
Moved the CPU capability logic inside fipsmodule

dkostic:
### Issues:
N/A

### Description of changes: 
The change moves the CPU capability detection logic inside the FIPS module as required by FIPS.

### Call-outs:

### Testing:


tests/ci/run_fips_tests.sh
```diff
@@ -4,37 +4,23 @@
...
+fips_build_and_test -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=1
+
+# The AL2 version of Clang does not have all of the required artifacts for address sanitizer, see P45594051
+if [[ ("${CC}" == *clang*) && ($(uname -a) == *Ubuntu*) ]]; then
+  echo "Building with Clang and testing AWS-LC in FIPS RelWithDebInfo mode with address sanitizer."
+  fips_build_and_test -DASAN=1 -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_SHARED_LIBS=1
+else
+  echo "Not building with Clang and testing AWS-LC in FIPS RelWithDebInfo mode."
+  fips_build_and_test -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_SHARED_LIBS=1
+fi
```
bryce-shang, line 17: Nit: in CodeBuild batch script, env variable `AWSLC_NO_ASM_FIPS` is used to enable/disable the ASAM. If `Ubuntu` is used, maybe delete `AWSLC_NO_ASM_FIPS`, which is no longer used. See previous code.
```
if [[  "${AWSLC_NO_ASM_FIPS}" == "1" ]]; then
  # This dimension corresponds to boringssl CI 'linux_fips_noasm_asan'.
  # https://logs.chromium.org/logs/boringssl/buildbucket/cr-buildbucket.appspot.com/8852496158370398336/+/steps/cmake/0/logs/execution_details/0
  echo "Testing AWS-LC in FIPS OPENSSL_NO_ASM Release mode."
  fips_build_and_test -DASAN=1 -DOPENSSL_NO_ASM=1 -DCMAKE_BUILD_TYPE=Release
  # TODO(shang): add ASAN build with RelWithDebInfo if final FIPS build uses the option.
fi
```

crypto/internal.h
```diff
@@ -159,14 +159,6 @@ extern "C" {
 void OPENSSL_cpuid_setup(void);
 #endif
 
-#if (defined(OPENSSL_ARM) || defined(OPENSSL_AARCH64)) && \
-    !defined(OPENSSL_STATIC_ARMCAP)
-// OPENSSL_get_armcap_pointer_for_test returns a pointer to |OPENSSL_armcap_P|
-// for unit tests. Any modifications to the value must be made after
-// |CRYPTO_library_init| but before any other function call in BoringSSL.
-OPENSSL_EXPORT uint32_t *OPENSSL_get_armcap_pointer_for_test(void);
-#endif
```
DominicDams, line 168: It's unclear to me why we are removing this. Is there a rational for this/ reason this won't break anything?

crypto/internal.h
```diff
@@ -159,14 +159,6 @@ extern "C" {
 void OPENSSL_cpuid_setup(void);
 #endif
 
-#if (defined(OPENSSL_ARM) || defined(OPENSSL_AARCH64)) && \
```
nebeid, line 162: I wonder if we wanted to keep this capability for non-fips buids.

crypto/test/gtest_main.cc
```diff
@@ -43,33 +37,6 @@ int main(int argc, char **argv) {
     }
 #endif
 
-#if defined(TEST_ARM_CPUS)
```
nebeid, line 46: Related to my comment: I wonder if we wanted to keep this capability for non-fips buids.

include/openssl/cpu.h
```diff
@@ -92,15 +92,6 @@ extern "C" {
 // Note: the CPUID bits are pre-adjusted for the OSXSAVE bit and the YMM and XMM
 // bits in XCR0, so it is not necessary to check those.
 extern uint32_t OPENSSL_ia32cap_P[4];
-
-#if defined(BORINGSSL_FIPS) && !defined(BORINGSSL_SHARED_LIBRARY)
-const uint32_t *OPENSSL_ia32cap_get(void);
-#else
-OPENSSL_INLINE const uint32_t *OPENSSL_ia32cap_get(void) {
```
nebeid, line 99: Related to my comment: I wonder if we wanted to keep this capability for non-fips buids.

tests/ci/run_fips_tests.sh
```diff
@@ -4,37 +4,23 @@
...
 
+echo "Testing AWS-LC in FIPS Release mode."
+fips_build_and_test -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=1
+
+# The AL2 version of Clang does not have all of the required artifacts for address sanitizer, see P45594051
+if [[ "${AWSLC_NO_ASM_FIPS}" == "1" ]]; then
+  echo "Building with Clang and testing AWS-LC in FIPS RelWithDebInfo mode with address sanitizer."
+  fips_build_and_test -DASAN=1 -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_SHARED_LIBS=1
+else
+  echo "Not building with Clang and testing AWS-LC in FIPS RelWithDebInfo mode."
```
nebeid, line 15: What does this echo mean?

util/all_tests.go
```diff
@@ -49,13 +49,8 @@ var (
 	jsonOutput      = flag.String("json-output", "", "The file to output JSON results to.")
 	mallocTest      = flag.Int64("malloc-test", -1, "If non-negative, run each test with each malloc in turn failing from the given number onwards.")
 	mallocTestDebug = flag.Bool("malloc-test-debug", false, "If true, ask each test to abort rather than fail a malloc. This can be used with a specific value for --malloc-test to identity the malloc failing that is causing problems.")
-	simulateARMCPUs = flag.Bool("simulate-arm-cpus", simulateARMCPUsDefault(), "If true, runs tests simulating different ARM CPUs.")
```
nebeid, line 52: Related to my comment: I wonder if we wanted to keep this capability for non-fips buids.

util/all_tests.json
```diff
@@ -12,6 +12,16 @@
...
     "skip_valgrind": true
   },
+  {
+    "cmd": ["crypto/crypto_test"],
+    "env": ["OPENSSL_armcap=0x0"],
+    "target_arch": "arm"
+  },
+  {
+    "cmd": ["crypto/crypto_test"],
+    "env": ["OPENSSL_armcap=0x1"],
```
nebeid, line 22: There is also OPENSSL_armcap=0x7D that we're using with ACVP, it's equivalent to neon+crypto in the removed simulation. I see that the selections for x86 below doesn't cover all combinations we're testing for ACVP, but at least for Arm, I think 0x7D is important.
Note: testing OPENSSL_armcap=0 and 1 on Graviton2 EC2, the least significant bit which is for ARMV7_NEON doesn't make a difference, since, as per its name, it's for armv7.

nebeid (approved):
Comments for consideration later.

crypto/internal.h
andrewhop: I'm paraphrasing a bit but we can't have this logic in our FIPS build because the running application could accidentally change the running characteristics of AWS-LC into something that isn't approved. We could keep if for the regular build but we decided to move everything to environment variable for simplicity.

tests/ci/run_fips_tests.sh
andrewhop: The current CMake address sanitizer config only works with Clang, so in our CI for the GCC targets it doesn't test with ASAN.

crypto/internal.h
dkostic: Correct, due to FIPS we are not allowed to expose a function that returns a pointer to the capabilities variable.

crypto/test/gtest_main.cc
dkostic: No need for that because we added the same tests using `all_tests.go` configs where we can set the capability variable by exporting an env var.

include/openssl/cpu.h
dkostic: all calls to this function were inside the fipsmodule, these calls are now changed to direct access to the variable. So the function is not used anywhere anymore.

util/all_tests.go
dkostic: the same functionality is now done in a different way, with config file

util/all_tests.json
dkostic: 0x7D is covered by default -- the first test in the file. Assuming we are running on GV, but I think that's a safe assumption. Regarding 0x1, that means we can remove the test, right?

crypto/fipsmodule/bcm.c
```diff
@@ -174,7 +184,8 @@ static void BORINGSSL_maybe_set_module_text_permissions(int permission) {}
 
 static void __attribute__((constructor))
 BORINGSSL_bcm_power_on_self_test(void) {
-  CRYPTO_library_init();
+
+  OPENSSL_cpuid_setup();
```
torben-hansen, line 188: This is a change of behaviour. Consuming applications can call `CRYPTO_library_init()`, which, in turn, calls `OPENSSL_cpuid_setup()`, but only calls the latter function once.

Hence, if any of the functions `OPENSSL_cpuid_setup()` are not idempotent, (e.g. depends on external data) then the cpu capabilities might change after returning from `BORINGSSL_bcm_power_on_self_test()`.

DominicDams (approved):

torben-hansen (changes requested):
Think we should look into this https://github.com/awslabs/aws-lc/pull/330#discussion_r767025644

crypto/fipsmodule/bcm.c
torben-hansen: This should be fine though! Since `do_library_init()` no longer does anything in fips mode.
https://github.com/awslabs/aws-lc/blob/feae2ebf5c31e8fe357b80c69b2e4eccc896bfbb/crypto/crypto.c#L74

crypto/crypto.c
```diff
@@ -41,82 +41,6 @@
...
-
-uint32_t *OPENSSL_get_armcap_pointer_for_test(void) {
-  return &OPENSSL_armcap_P;
-}
-#endif
-
-#endif
-
 #if defined(BORINGSSL_FIPS)
 // In FIPS mode, the power-on self-test function calls |CRYPTO_library_init|
```
andrewhop, line 45: TODO update this say OPENSSL_cpuid_setup not CRYPTO_library_init.

torben-hansen (approved):

util/all_tests.json
nebeid: It doesn't hurt to keep it. I did some tests locally and it seems to be slower (??) than when set to 0. Needs more investigation.

nebeid:
This comment https://github.com/awslabs/aws-lc/pull/330#discussion_r764382755 (editorial in comment) and this one https://github.com/awslabs/aws-lc/pull/330#discussion_r767034542 (TODO: fixing function name in a comment) are addressed in PR https://github.com/awslabs/aws-lc/pull/342

PR #331, opened 2021-12-08
Include the RSA PCT check in the number of keygen attempts.

nebeid:
### Issues:
Continues CryptoAlg-908

### Description of changes: 
Based on feedback about the FIPS RSA PCT, it was only done once and not included in the number of attempts of key generation. This change moves it inside the `do-while` loop.

### DIFF
--- a/crypto/fipsmodule/rsa/rsa_impl.c
+++ b/crypto/fipsmodule/rsa/rsa_impl.c
@@ -1329,7 +1329,8 @@ static int RSA_generate_key_ex_maybe_fips(RSA *rsa, int bits,
   int ret = 0;
 
   // |rsa_generate_key_impl|'s 2^-20 failure probability is too high at scale,
-  // so we run the FIPS algorithm four times, bringing it down to 2^-80. We
+  // so we run the FIPS algorithm |MAX_RSA_KEYGEN_ATTEMPTS| times,
+  // bringing it down to 2^-80 when it's equal to 4. We
   // should just adjust the retry limit, but FIPS 186-4 prescribes that value
   // and thus results in unnecessary complexity.
   int failures = 0;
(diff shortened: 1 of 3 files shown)


crypto/fipsmodule/rsa/rsa_impl.c
```diff
@@ -1352,10 +1356,10 @@ static int RSA_generate_key_ex_maybe_fips(RSA *rsa, int bits,
 
     // Only retry on |RSA_R_TOO_MANY_ITERATIONS|. This is so a caller-induced
     // failure in |BN_GENCB_call| is still fatal.
-  } while (failures < 4 && ERR_GET_LIB(err) == ERR_LIB_RSA &&
+  } while (failures < MAX_KEYGEN_ATTEMPTS && ERR_GET_LIB(err) == ERR_LIB_RSA &&
```
torben-hansen, line 1359: This will always evaluate to false in non-fips mode:

If `rsa_generate_key_impl` succeeds then we break out of the do-while loop

If `rsa_generate_key_impl` fails we increment `failures` from 0 to 1. But since `MAX_KEYGEN_ATTEMPTS = 1` the condition will evaluate to false and we proceed without a retry.

This is prob screwing with the test fixture `KeygenInternalRetry`.

crypto/fipsmodule/rsa/rsa_impl.c
```diff
@@ -1342,7 +1342,11 @@ static int RSA_generate_key_ex_maybe_fips(RSA *rsa, int bits,
     }
 
     if (rsa_generate_key_impl(tmp, bits, e_value, cb)) {
-      break;
+      if (check_fips && !RSA_check_fips(tmp)) {
+        failures++;
```
torben-hansen, line 1346: you increment `failures` twice.

crypto/internal.h
```diff
@@ -937,13 +937,13 @@ static inline uint64_t CRYPTO_rotr_u64(uint64_t value, int shift) {
 // FIPS functions.
 
 #if defined(BORINGSSL_FIPS)
```
torben-hansen, line 939: Might as well switch to `AWSLC_FIPS`?

samuel40791765 (approved):

torben-hansen (approved):

PR #334, opened 2021-12-09
Reverted the inner loop of RSA keygen to retry when TOO_MANY_ITERATIONS

nebeid:
### Issues:
Addresses CryptoAlg-908

### Description of changes: 
In the FIPS feedback of the previous commit, it was pointed out a flaw in the logic:
Failing for other reasons than TOO_MANY_ITERATIONS would not allow more retries.
So the original loop in `RSA_generate_key_ex_maybe_fips()` in rsa_impl.c as in 6bdd4c3 is brought back to focus on only retrying keygen when TOO_MANY_ITERATIONS is the reason for failure.

An outer loop is added to test for PCT and have its own number of attempts which covers both keygen and PCT failures. This behaviour now matches that of `EC_KEY_generate_key_fips()`.

### DIFF
--- a/crypto/fipsmodule/rsa/rsa_impl.c
+++ b/crypto/fipsmodule/rsa/rsa_impl.c
@@ -1366,7 +1366,7 @@ static int RSA_generate_key_ex_maybe_fips(RSA *rsa, int bits,
       }
     }
     num_attempts++;
-  } while ((num_attempts < MAX_KEYGEN_ATTEMPTS) && (tmp == NULL));
+  } while ((tmp == NULL) && (num_attempts < MAX_KEYGEN_ATTEMPTS));
 
   if (tmp == NULL) {
     goto out;
--- a/crypto/fipsmodule/rsa/rsa_impl.c
+++ b/crypto/fipsmodule/rsa/rsa_impl.c
@@ -1327,37 +1327,46 @@ static int RSA_generate_key_ex_maybe_fips(RSA *rsa, int bits,
   RSA *tmp = NULL;
   uint32_t err;
   int ret = 0;
+  int failures;
+  int num_attempts = 0;
 
+  do {
   // |rsa_generate_key_impl|'s 2^-20 failure probability is too high at scale,
-  // so we run the FIPS algorithm |MAX_RSA_KEYGEN_ATTEMPTS| times,
-  // bringing it down to 2^-80 when it's equal to 4. We
+  // so we run the FIPS algorithm four times, bringing it down to 2^-80. We
   // should just adjust the retry limit, but FIPS 186-4 prescribes that value
   // and thus results in unnecessary complexity.
-  int failures = 0;
-  do {
-    ERR_clear_error();
-    // Generate into scratch space, to avoid leaving partial work on failure.
-    tmp = RSA_new();
-    if (tmp == NULL) {
-      goto out;
-    }
+  failures = 0;
+    do {
+      ERR_clear_error();
+      // Generate into scratch space, to avoid leaving partial work on failure.
+      tmp = RSA_new();
+      if (tmp == NULL) {
+        goto out;
+      }
 
-    if (rsa_generate_key_impl(tmp, bits, e_value, cb)) {
-      if (check_fips && !RSA_check_fips(tmp)) {
-        failures++;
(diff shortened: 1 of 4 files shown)


crypto/fipsmodule/rsa/rsa_impl.c
```diff
@@ -1327,37 +1327,46 @@ static int RSA_generate_key_ex_maybe_fips(RSA *rsa, int bits,
...
-           ERR_GET_REASON(err) == RSA_R_TOO_MANY_ITERATIONS);
+    // Perform PCT test in the case of FIPS
+    if (tmp) {
+      if (check_fips && !RSA_check_fips(tmp)) {
+        RSA_free(tmp);
+        tmp = NULL;
+      }
+    }
+    num_attempts++;
+  } while ((num_attempts < MAX_KEYGEN_ATTEMPTS) && (tmp == NULL));
```
torben-hansen, line 1369: `tmp != NULL` in almost-all cases. So check that first I guess

crypto/rsa_extra/rsa_test.cc
```diff
@@ -1082,7 +1086,7 @@ TEST(RSADeathTest, KeygenFailOnceAndDie) {
                [](int event, int n, BN_GENCB *cb_ptr) -> int {
                  bool *failed_ptr = static_cast<bool *>(cb_ptr->arg);
                  if (*failed_ptr) {
-                   ADD_FAILURE() << "Callback called multiple times.";
+                     //ADD_FAILURE() << "Callback called multiple times.";
```
torben-hansen, line 1089: Meant to remove?

crypto/rsa_extra/rsa_test.cc
```diff
@@ -1071,7 +1071,11 @@ TEST(RSADeathTest, KeygenFailAndDie) {
   EXPECT_NE(Bytes(der, der_len), Bytes(der3, der3_len));
 }
 
-TEST(RSADeathTest, KeygenFailOnceAndDie) {
+// This is not a death test. It's similar to the test in the non-FIPS case,
+// but because the callback gets called again in the outer loop,
+// this cause of failure is removed.
```
torben-hansen, line 1076: re-write sentence.

crypto/fipsmodule/rsa/rsa_impl.c
```diff
@@ -1327,37 +1327,46 @@ static int RSA_generate_key_ex_maybe_fips(RSA *rsa, int bits,
...
   // and thus results in unnecessary complexity.
-  int failures = 0;
-  do {
-    ERR_clear_error();
-    // Generate into scratch space, to avoid leaving partial work on failure.
-    tmp = RSA_new();
-    if (tmp == NULL) {
-      goto out;
-    }
+  failures = 0;
```
dkostic, line 1338: Nit: indentation

crypto/fipsmodule/rsa/rsa_impl.c
```diff
@@ -1327,37 +1327,46 @@ static int RSA_generate_key_ex_maybe_fips(RSA *rsa, int bits,
...
-    RSA_free(tmp);
-    tmp = NULL;
+      err = ERR_peek_error();
+      RSA_free(tmp);
+      tmp = NULL;
+      failures++;
+
+      // Only retry on |RSA_R_TOO_MANY_ITERATIONS|. This is so a caller-induced
+      // failure in |BN_GENCB_call| is still fatal.
+    } while (failures < 4 && ERR_GET_LIB(err) == ERR_LIB_RSA &&
```
dkostic, line 1358: any particular reason to replace MAX_RSA_KEYGEN_ATTEMPTS with 4?

dkostic (approved):

andrewhop (approved):

crypto/fipsmodule/rsa/rsa_impl.c
nebeid: 4 was originally there, and it makes the probability of failure of `rsa_generate_key_impl` down from 2^{-20} to 2^{-80} as per the original comment. The inner loop is exactly as it was before the PCT iterate-and-abort changes and I think of it as one step of RSA key generation.

PR #343, opened 2021-12-13
Add PQCrystals-Kyber512, Remove SIKE, relocate pq_kem interface.

brian-jarvis-aws:
### Description of changes: 
This PR imports the source code of Kyber512 from this commit of the official Kyber repository: https://github.com/pq-crystals/kyber/commit/faf5c3fe33e0b61c7c8a7888dd862bf5def17ad2

The keygen, encapsulate, and decapsulate methods are exposed using a newly created pq_kem interface which is based off of previous work from Dusan Kostic in this PR: https://github.com/awslabs/aws-lc/commit/5302e2d3ce5404ecac212804a77151378c81c847.

This PR also integrates the key generation and some associated methods into the EVP_PKEY APIs.

#### Creation of a new PQ KEM interface vs OpenSSL's KEM interface
Upstream, OpenSSL now includes an EVP KEM interface. This was investigated for use, but due to the number of additional dependencies it would bring in was decided against. Should AWS-LC decide to bring in this interface from OpenSSL, Kyber can then be integrated into it as a follow-on effort.

#### Addition to EVP interface
In addition to simply importing the Kyber512 source code and providing an interface to the KEM methods, this PR also makes Kyber512 keys available to the EVP PKEY interface including a subset of methods for key management.

#### Kyber512-Ref
The source code imported is for the Reference implementation of Kyber512. This was selected for initial import due to it being platform agnostic. The optimized implementation is for AVX2 instruction sets and so priority was given to the implementation which will work on all architectures. Importing optimized versions (and higher security levels) can be done as a follow-on effort.

#### SHA-3/SHAKE
Kyber requires SHA-3 and SHAKE methods which are not currently available in AWS-LC. For this PR, the needed SHA-3/SHAKE dependencies are included within the Kyber512 source tree directly. Separating out a separate SHA-3/SHAKE interface in AWS-LC is out of scope, but would be a good follow-on effort to do.

#### DRBG
To support KAT testing, a deterministic random number generator was needed. This PR introduces the same DRBG used by the Kyber source code and uses the same randombytes interface to seed and get random bytes from it. The DRBG only gets built when the testing flag is enabled. Otherwise the existing RNG is used.

### Testing:
Unit tests were added which exercise the use of Kyber512 in the pq_kem interface and the EVP interface. KAT tests from the Kyber submission to NIST's Round 3 Post-Quantum Cryptography selection process are also included. To support running the KAT tests, an additional random number generator was added which is deterministic and only built in test mode.

```
... other existing tests ...

[----------] 5 tests from Kyber512Test
[ RUN      ] Kyber512Test.KAT_tests
[       OK ] Kyber512Test.KAT_tests (71 ms)
[ RUN      ] Kyber512Test.EVP_PKEY_keygen
[       OK ] Kyber512Test.EVP_PKEY_keygen (0 ms)
[ RUN      ] Kyber512Test.EVP_PKEY_cmp
[       OK ] Kyber512Test.EVP_PKEY_cmp (1 ms)
[ RUN      ] Kyber512Test.EVP_PKEY_new_raw
[       OK ] Kyber512Test.EVP_PKEY_new_raw (0 ms)
[ RUN      ] Kyber512Test.EVP_PKEY_size
[       OK ] Kyber512Test.EVP_PKEY_size (0 ms)
[----------] 5 tests from Kyber512Test (72 ms total)

[----------] 1 test from PQKEMTest
[ RUN      ] PQKEMTest.Kyber512_Basic
[       OK ] PQKEMTest.Kyber512_Basic (1 ms)
[----------] 1 test from PQKEMTest (1 ms total)

... other existing tests ...

[==========] 1142 tests from 90 test suites ran. (89379 ms total)
[  PASSED  ] 1142 tests.
```


brian-jarvis-aws:
This is a cleaned up version of https://github.com/awslabs/aws-lc/pull/335, which was not caught up with main so had a lot of noise. This also removes SIKE since the plan for this branch is to only house Kyber until a NIST selection is made and then proceed with whichever KEM is the winner.

brian-jarvis-aws:
Working on recreating the CI failures locally so I can address them.

torben-hansen:
> [...] The DRBG only gets built when the testing flag is enabled. Otherwise the existing RNG is used.

I'm not sure this approach would work that well unfortunately. The test flag `BUILD_TESTING` is enabled by default for all build types: "debug", "release", etc. Hence, atm, all standard builds will induce the PQ algorithm to use a deterministic DRBG. Unless you explicitly disable `BUILD_TESTING` at build configuration time.

What about turning this into an internal runtime configuration? you can expose an internal function (a simple wrapper) that will, at runtime, use a deterministic backend for the `randombytes`. This backend could be implemented using the `CTR_DRBG` API, which is exported https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/rand/internal.h#L93. So, the `CTR_DRBG` API is only nominally opaque/hidden at build/link time. With a runtime configuration, you can target DRBG determinism for the single test fixture you have, that requires it.

This is similar to what we do when testing fork detection, using runtime configurations, clearly marked with "TESTING" i.e. https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/rand/fork_detect.c#L200.

The best way to implement this at compile-time would maybe be an extra dimension in the CI - just for this case. This can quickly lead to combinatorial explosion though.

DominicDams (approved):
I went over most of this code and it looks fine. We'll need to take a closer look again when we have something going to main but I have no qualms with this going to a feature branch.

PR #355, opened 2021-12-21
Documentation updates for Windows and building FIPS

andrewhop:
### Description of changes: 
* The Windows CI readme was out of date and would not be able to build or use our Docker containers.
* Add a safety check to our FIPS build to reduce confusion. Without the check the build fails with a delocate parse error.

### Call-outs:
We already don't support a FIPS static build, this check just helps users understand the issue.

### Testing:
Not passing in the required flag fails with:
```
CMake Error at CMakeLists.txt:501 (message):
  Building AWS-LC for FIPS only supports the shared build, you must pass in
  '-DBUILD_SHARED_LIBS=TRUE'
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -497,6 +497,10 @@ if(FIPS AND (NOT GO_EXECUTABLE OR NOT PERL_EXECUTABLE))
   message(FATAL_ERROR "Building AWS-LC for FIPS requires Go and Perl")
 endif()
 
+if(FIPS AND NOT BUILD_SHARED_LIBS)
+  message(FATAL_ERROR "Building AWS-LC for FIPS only supports the shared build, you must pass in '-DBUILD_SHARED_LIBS=TRUE'")
+endif()
+
 if(FIPS)
   add_definitions(-DBORINGSSL_FIPS)
   if(FIPS_BREAK_TEST)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -497,6 +497,10 @@ if(FIPS AND (NOT GO_EXECUTABLE OR NOT PERL_EXECUTABLE))
   message(FATAL_ERROR "Building AWS-LC for FIPS requires Go and Perl")
 endif()
 
+if(FIPS AND NOT BUILD_SHARED_LIBS)
+  message(FATAL_ERROR "Building AWS-LC for FIPS only supports the shared build, you must pass in '-DBUILD_SHARED_LIBS=TRUE'")
+endif()
```
bryce-shang, line 502: * FIPS shared build will be support soon, right? e.g. Android.
* Do we plan to just make this diff to `fips-2021-10-20` branch?

andrewhop: * Yes we need to add FIPS static support back soon. When we get to doing that work it will be very obvious we need to remove this change because there will be no way to test that change with this in place.
* I was not planning on moving this to the `fips-2021-10-20` branch. For that people should be following the build instructions in our security policy. I was treating this change as a nice to have sanity test for developers developing on `main`.

bryce-shang (approved):

darylmartin100 (approved):

PR #358, opened 2021-12-22
Partial public-key validation for peer key in ECDH

nebeid:
### Issues:
Addresses CryptoAlg-905

### Description of changes: 
Continuing on the requirements of public key validation in SP 800-56Ar3, this change satisfies Sec. 5.6.2.3.4, where a peer public key should undergo a partial key validation.
This change:
- satisfies performing the check in `EVP_PKEY_derive` and `ECDH_compute_key_fips`.
- ~also performs the check in `ECDH_compute_key` (non-FIPS-Approved), since the check is not expensive.~

### DIFF
--- a/crypto/ecdh_extra/ecdh_test.cc
+++ b/crypto/ecdh_extra/ecdh_test.cc
@@ -227,7 +227,7 @@ TEST(ECDHTest, InvalidPubKeyLargeCoord) {
     OPENSSL_memcpy(peer_key.get()->pub_key->raw.X.bytes, (const uint8_t *)xpp.get()->d, len);
     ret = ECDH_compute_key_fips(shared_key.data(), shared_key.size(),
                                     EC_KEY_get0_public_key(peer_key.get()), priv_key.get());
-    if (!awslc_fips() && (group.get()->curve_name == NID_secp224r1)) {
+    if (!awslc_fips() && has_uint128_and_not_small() && (group.get()->curve_name == NID_secp224r1)) {
       ASSERT_TRUE(ret);
     } else {
       ASSERT_FALSE(ret);
--- a/crypto/ecdh_extra/ecdh_test.cc
+++ b/crypto/ecdh_extra/ecdh_test.cc
@@ -36,7 +36,6 @@
 #include "../test/test_util.h"
 #include "../test/wycheproof_util.h"
 
-
 static bssl::UniquePtr<EC_GROUP> GetCurve(FileTest *t, const char *key) {
   std::string curve_name;
   if (!t->GetAttribute(&curve_name, key)) {
@@ -128,6 +127,22 @@ TEST(ECDHTest, TestVectors) {
   });
 }
 
+static int has_uint128_and_not_small() {
+#if defined(BORINGSSL_HAS_UINT128) && !defined(OPENSSL_SMALL)
+  return 1;
+#else
+  return 0;
+#endif
+}
+
+static int awslc_fips() {
+#if defined(AWSLC_FIPS)
+  return 1;
+#else
+  return 0;
+#endif
+}
+
 // The following test is adapted from ECTest.LargeXCoordinateVectors
 TEST(ECDHTest, InvalidPubKeyLargeCoord) {
   bssl::UniquePtr<BN_CTX> ctx(BN_CTX_new());
@@ -135,6 +150,7 @@ TEST(ECDHTest, InvalidPubKeyLargeCoord) {
 
   FileTestGTest("crypto/fipsmodule/ec/large_x_coordinate_points.txt",
                 [&](FileTest *t) {
+    int ret;
     bssl::UniquePtr<EC_GROUP> group = GetCurve(t, "Curve");
     ASSERT_TRUE(group);
     bssl::UniquePtr<BIGNUM> x = GetBIGNUM(t, "X");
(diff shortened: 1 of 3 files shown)


crypto/fipsmodule/ecdh/ecdh.c
```diff
@@ -79,26 +79,43 @@
...
   const EC_GROUP *const group = EC_KEY_get0_group(priv_key);
   if (EC_GROUP_cmp(group, pub_key->group, NULL) != 0) {
     OPENSSL_PUT_ERROR(EC, EC_R_INCOMPATIBLE_OBJECTS);
-    return 0;
+    goto end;
+  }
+
+  // |EC_KEY_check_fips| is not an expensive operation on an external
+  // public key.
+  key_pub_key = EC_KEY_new();
```
darylmartin100, line 97: Should you check for null here before using key_pub_key?

nebeid: Thanks. Done.

darylmartin100 (dismissed):

crypto/fipsmodule/ecdh/ecdh.c
```diff
@@ -79,26 +79,47 @@
...
   }
   const EC_SCALAR *const priv = &priv_key->priv_key->scalar;
   const EC_GROUP *const group = EC_KEY_get0_group(priv_key);
   if (EC_GROUP_cmp(group, pub_key->group, NULL) != 0) {
     OPENSSL_PUT_ERROR(EC, EC_R_INCOMPATIBLE_OBJECTS);
-    return 0;
+    goto end;
+  }
+
+  // |EC_KEY_check_fips| is not an expensive operation on an external
```
torben-hansen, line 95: Still extra work that can easily be scoped to fips mode only.

crypto/ecdh_extra/ecdh_test.cc
```diff
@@ -126,6 +128,82 @@ TEST(ECDHTest, TestVectors) {
...
+
+    ASSERT_TRUE(EC_KEY_set_group(peer_key.get(), group.get()));
+    // The following call converts the point to Montgomery form for P-256, 384 and 521.
+    // For P-224, when the functions from simple.c are used, i.e. when
+    // group->meth = EC_GFp_nistp224_method, the coordinate representation is not changed.
+    // This is determined based on compile flags in ec.c that are also used below.
+    ASSERT_TRUE(EC_POINT_set_affine_coordinates_GFp(
+                    group.get(), pub_key.get(), x.get(), y.get(), nullptr));
+    ASSERT_TRUE(EC_KEY_set_public_key(peer_key.get(), pub_key.get()));
+    //ASSERT_TRUE(EC_KEY_check_fips(peer_key.get()));
```
torben-hansen, line 169: ?

crypto/ecdh_extra/ecdh_test.cc
```diff
@@ -126,6 +128,82 @@ TEST(ECDHTest, TestVectors) {
...
+    // Note that both are in little-endian byte order.
+    OPENSSL_memcpy(peer_key.get()->pub_key->raw.X.bytes, (const uint8_t *)x.get()->d, len);
+    OPENSSL_memcpy(peer_key.get()->pub_key->raw.Y.bytes, (const uint8_t *)y.get()->d, len);
+    OPENSSL_memset(peer_key.get()->pub_key->raw.Z.bytes, 0, len);
+    peer_key.get()->pub_key->raw.Z.bytes[0] = 1;
+    // As mentioned, for P-224, setting the raw point directly with the coordinates
+    // still passes |EC_KEY_check_fips|.
+    // For P-256, 384 and 521, the failure is due to that the coordinates are
+    // not in Montgomery representation, and, hence, fail |EC_KEY_check_fips|.
+#if defined(BORINGSSL_HAS_UINT128) && !defined(OPENSSL_SMALL)
```
torben-hansen, line 183: IMO better/more readable to conditionally compile a boolean to either true or false, that you can use to branch i the if-else.

crypto/fipsmodule/ecdh/ecdh.c
```diff
@@ -79,26 +79,47 @@
 
 int ECDH_compute_shared_secret(uint8_t *buf, size_t *buflen, const EC_POINT *pub_key,
```
samuel40791765, line 80: We should insert a service indicator lock/unlock in this function. `EC_KEY_check_fips` is an approved function and will inadvertently update the state when called here.

torben-hansen: Only if the context has an associate private key no?

samuel40791765: Oh yeah, that's true. Didn't look at the condition closely enough.

crypto/fipsmodule/ecdh/ecdh.c
nebeid: Done. It didn't have an effect on performance, so it was faster to fail here than to fail during the following computation. I think we may widen the scope of the check once we have formal verification on the FIPS build. See TODO in the test.

crypto/ecdh_extra/ecdh_test.cc
nebeid: Done

torben-hansen (dismissed):

torben-hansen (dismissed):

samuel40791765 (dismissed):

samuel40791765 (approved):

crypto/fipsmodule/ecdh/ecdh.c
```diff
@@ -79,26 +79,52 @@
...
     OPENSSL_PUT_ERROR(ECDH, ECDH_R_NO_PRIVATE_VALUE);
-    return 0;
+    goto err;
   }
   const EC_SCALAR *const priv = &priv_key->priv_key->scalar;
   const EC_GROUP *const group = EC_KEY_get0_group(priv_key);
   if (EC_GROUP_cmp(group, pub_key->group, NULL) != 0) {
     OPENSSL_PUT_ERROR(EC, EC_R_INCOMPATIBLE_OBJECTS);
-    return 0;
+    goto err;
```
torben-hansen, line 91: nit: just return here instead of defining another label.

Looks like a fix in the FV is needed...

torben-hansen (approved):

PR #359, opened 2021-12-22
Move CRYPTO_tls1_prf to a public header

andrewhop:
### Description of changes: 
To let customers easily use the FIPS validated CRYPTO_tls1_prf it needs to be in a public header. Second only to naming the hardest problem in computer science is organization:
* I thought about hkdf.h but that is all focused on HKDF as defined in RFC 5869
* I thought about hmac.h because the TLS PRF uses a lot of HMAC but it didn't fit the vibe of other function
* I thought about ssl.h/tls.h but those are specific to libssl and this function exists in libcrypto
* I settled on moving the current header file as is and picking a reasonable name. This could be more generic like tls_crypto.h so it could house future libcrypto functions that are somewhat libssl specific.

I'm happy to move it to a better spot if someone knows of a good one.

### Testing:
libcrypto and libssl both build and pass all tests locally.


torben-hansen:
> I settled on moving the current header file as is and picking a reasonable name. This could be more generic like tls_crypto.h so it could house future libcrypto functions that are somewhat libssl specific.

I think we should pick a more generic name.

samuel40791765 (approved):

PR #364, opened 2022-01-05
Enable more Windows builds

andrewhop:
### Issues:
Addresses CryptoAlg-907

### Description of changes: 
* Turn on most windows build flavors:
    * Debug build
         * Only for MSVC 2017
    * Release small
    * Release no assembly
    * Release shared build
    * Regular Release is still on
* Update Window's PATH variable to search for our DLLs in the build directory
* Update CI script to support testing on local paths
* Update CI script to skip printing useless CMake logs
* Enable `GTEST_CREATE_SHARED_LIBRARY` so Windows can get the `dllexport`s it needs
* Only specify `BORINGSSL_IMPLEMENTATION` when building our code so Windows picks correctly between `__declspec(dllexport)` and `__declspec(dllimport)` in base.h

### Call-outs:
Without the changes to `BORINGSSL_IMPLEMENTATION` Windows tests would fail because in the unit tests `BORINGSSL_IMPLEMENTATION` would be set and all the `OPENSSL_EXPORT`s would still be `dllexport`ing code instead of importing it. This is due to `add_definitions(-DBORINGSSL_IMPLEMENTATION)` which turns it on for everything.

This does **not** fix the Windows FIPS build, more changes coming soon!

### Testing:
`run_windows_tests.bat` passes locally. These changes don't affect the linux behavior as far as I can tell.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -407,7 +407,7 @@ if(FUZZ)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address,fuzzer-no-link -fsanitize-coverage=edge,indirect-calls")
 endif()
 
-add_definitions(-DBORINGSSL_IMPLEMENTATION)
+#add_definitions(-DBORINGSSL_IMPLEMENTATION)
 
 if(BUILD_SHARED_LIBS)
   add_definitions(-DBORINGSSL_SHARED_LIBRARY)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -705,6 +705,10 @@ if(BUILD_TESTING)
   # Add minimal googletest targets. The provided one has many side-effects, and
   # googletest has a very straightforward build.
   add_library(boringssl_gtest third_party/googletest/src/gtest-all.cc)
+  if(BUILD_SHARED_LIBS)
+    # This is needed for the Windows build to correctly annotate GTest's API with __declspec(dllexport)
+    target_compile_options(boringssl_gtest PRIVATE -DGTEST_CREATE_SHARED_LIBRARY=1)
+  endif()
   target_include_directories(boringssl_gtest PRIVATE third_party/googletest)
 
   include_directories(third_party/googletest/include)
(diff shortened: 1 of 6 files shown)


DominicDams (approved):

tests/ci/run_windows_tests.bat
```diff
@@ -1,21 +1,26 @@
...
 @rem %1 contains the path to the setup batch file for the version of of visual studio that was passed in from the build spec file.
 @rem x64 comes from the architecture options https://docs.microsoft.com/en-us/cpp/build/building-on-the-command-line
-call %1 x64 || goto error
+set MSVC_PATH=%1
+call %MSVC_PATH% x64 || goto error
 SET
+
+@rem Run the same builds as run_posix_tests.sh
+@rem check which version of MSVC we're building with: remove 14.0 from the path to the compiler and check if it matches the
+@rem original string. MSVC 14 has an issue with a missing DLL that causes teh debug build to fail to start
```
DominicDams, line 14: nit: the

DominicDams (approved):

CMakeLists.txt
```diff
@@ -705,6 +703,10 @@ if(BUILD_TESTING)
   # Add minimal googletest targets. The provided one has many side-effects, and
   # googletest has a very straightforward build.
   add_library(boringssl_gtest third_party/googletest/src/gtest-all.cc)
+  if(BUILD_SHARED_LIBS)
+    # This is needed for the Windows build to correctly annotate GTest's API with __declspec(dllexport)
+    target_compile_options(boringssl_gtest PRIVATE -DGTEST_CREATE_SHARED_LIBRARY=1)
+  endif()
```
bryce-shang, line 709:
```suggestion
  if(BUILD_SHARED_LIBS AND MSVC)
    # This is needed for the Windows build to correctly annotate GTest's API with __declspec(dllexport)
    target_compile_options(boringssl_gtest PRIVATE -DGTEST_CREATE_SHARED_LIBRARY=1)
  endif()
```

bryce-shang: Nit: maybe we can avoid removing the global definition(and add the definition in place) by only call `remove_definitions` inside the if condition because the test seems the final one being built.
https://cmake.org/cmake/help/v3.1/command/remove_definitions.html

bryce-shang (approved):

CMakeLists.txt
andrewhop: I'll experiment, but I'm not sure that will work since the tests are in the same folder as some of the source code:
> Removes flags (added by add_definitions) from the compiler command line for sources in the current directory and below.

bryce-shang: you are right. the test is built before other directories. the function may not work.

PR #368, opened 2022-01-13
Optimized HMAC implementions

SalusaSecondus:
### Issues:
None

### Description of changes: 
Profiling the existing EVP-based HMAC implementations shows that much of their time is spent in memory management functions and EVP overhead. (In some cases it is more that 5%.) Additionally the EVP APIs do heap-based memory allocation internally.

The more involvement memory management can cause problems for some consumers, specifically those with their own memory management concerns:

* Applications using custom allocators (including zero-allocation)
* Applications with tight time constraints (as malloc/free and be expensive)
* Foreign Function Interface (FFI) applications such as Java
    * Allocating/Freeing/Copying items on the heap require expensive FFI calls
    * Stack-based items can commonly be stored in memory managed by the caller (such as Java) and so be allocated/freed/copy without FFI calls.

By refactoring the HMAC code so that all state lives locally in a single stack-allocatable object we can significantly improve performance.

I have attached three graphs.
1. Improvements: This shows the percentage improvement of the new implementation and APIs over the existing (measured in ops/sec)
  hmac_improvements2
2. HMAC Operation Time: Shows how long it takes to execute HMAC for various algorithms and data-sizes (measured in nanoseconds)
  hmac_time2
3. Where Time is Spent: For just HMAC-256, shows the percentage break-down of where CPU time is spent for the old implementation and new APIs.
  hmac_timespent2

### Call-outs:
* We need to ensure that these changes play nicely with FIPS requirements
* This PR changes the ABI and thus applications will need to re-compile to safely use it.

### Testing:
All existing tests pass and have been refactored to cover the new APIs.


crypto/cipher_extra/e_tls.c
```diff
@@ -45,19 +45,12 @@ typedef struct {
...
-OPENSSL_STATIC_ASSERT(alignof(union evp_aead_ctx_st_state) >=
-                          alignof(AEAD_TLS_CTX),
-                      AEAD_state_has_insufficient_alignment)
-#endif
-
 static void aead_tls_cleanup(EVP_AEAD_CTX *ctx) {
-  AEAD_TLS_CTX *tls_ctx = (AEAD_TLS_CTX *)&ctx->state;
+  AEAD_TLS_CTX *tls_ctx = (AEAD_TLS_CTX *)ctx->state.ptr;
   EVP_CIPHER_CTX_cleanup(&tls_ctx->cipher_ctx);
   HMAC_CTX_cleanup(&tls_ctx->hmac_ctx);
```
torben-hansen, line 51: Can't you do with making the `hmac_ctx` field a pointer? And only allocate it when you need it, according to the requested cipher? e.g. `aes-gcm` won't need it.

SalusaSecondus: The casting was getting annoying so I just added a `(void*)` pointer to the union. Since that's smaller than the existing union it doesn't actually change anything, just makes the casting much easier.

torben-hansen: I was thinking about avoiding dynamic memory allocation https://github.com/awslabs/aws-lc/pull/368/files#diff-544b0ce33ba4678dc2dcd3e8667b8473e4e5cf2e67291ea089e5ab3964d3cdb5R76

unless necessary depending on the cipher. Might not be technically possible, of course. Just thoughts.

SalusaSecondus: Actually, I think that this code-path is only encountered when HMAC is needed, so we may be fine.

torben-hansen: Ahh that's correct. I confused it with the APIs the layer above this one!

SalusaSecondus:
Newest version does not expose any new APIs (though still changes the ABI) and is slightly less performant that the original proposal (probably due to heavier use of function pointers). However, it removes essentially all macro programming and now only adds about 19.5KB to `libcrypto.a`

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -120,119 +175,114 @@ void HMAC_CTX_free(HMAC_CTX *ctx) {
...
       }
     } else {
-      assert(key_len <= sizeof(key_block));
       OPENSSL_memcpy(key_block, key, key_len);
-      key_block_len = (unsigned)key_len;
     }
-    // Keys are then padded with zeros.
-    if (key_block_len != EVP_MAX_MD_BLOCK_SIZE) {
-      OPENSSL_memset(&key_block[key_block_len], 0, sizeof(key_block) - key_block_len);
+    for (size_t i = 0; i < block_size / 8; i++) {
```
nebeid, line 212: The previous version of this function seems to allow key_len to not be divisible by 8 when padding with 0s. However, after that, I think both the previous and this version, due to XORing the ipad and opad with 0, will have those patterns show in key_block. Are we relying on the actual implementation to ignore those extra bytes?

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -120,119 +175,114 @@ void HMAC_CTX_free(HMAC_CTX *ctx) {
...
-
-    size_t block_size = EVP_MD_block_size(md);
-    assert(block_size <= sizeof(key_block));
+  // At this point we know we have valid methods and a context allocated.
+  const HmacMethods *methods = ctx->methods;
+  size_t block_size = EVP_MD_block_size(methods->evp_md);
+  assert(block_size % 8 == 0);
+  assert(block_size <= EVP_MAX_MD_BLOCK_SIZE);
+
+  FIPS_service_indicator_lock_state();
```
nebeid, line 198: I suggest keeping the comment: " 
// We have to avoid the underlying SHA services updating the indicator
// state, so we lock the state here."

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -120,119 +175,114 @@ void HMAC_CTX_free(HMAC_CTX *ctx) {
...
-      !EVP_DigestUpdate(&ctx->md_ctx, buf, i) ||
-      !EVP_DigestFinal_ex(&ctx->md_ctx, out, out_len)) {
-    *out_len = 0;
-    FIPS_service_indicator_unlock_state();
+  const HmacMethods *methods = ctx->methods;
+  if (methods == NULL || ctx->initialized != 1) {
     return 0;
   }
-
+  FIPS_service_indicator_lock_state();
```
nebeid, line 248: I suggest keeping the comment: "
  // We have to avoid the underlying SHA services updating the indicator
  // state, so we lock the state here.
"

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -64,34 +64,91 @@
 
 #include "../../internal.h"
 
+typedef int (*HashInit)(void *);
+typedef int (*HashUpdate)(void *, const void*, size_t);
+typedef int (*HashFinal)(uint8_t *, void*);
```
nebeid, line 69: Why this function doesn't take the ctx as the first parameter?

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -120,119 +175,114 @@ void HMAC_CTX_free(HMAC_CTX *ctx) {
...
-  HMAC_verify_service_indicator(ctx->md);
-  return 1;
+  if (result) {
+    HMAC_verify_service_indicator(evp_md);
+    if (out_len) {
+      *out_len = hmac_len;
+    }
+    return 1;
+  } else {
+    return 0;
```
nebeid, line 270: I think here `out_len` should be set to 0 to match the original behaviour.

crypto/cipher_extra/e_tls.c
```diff
@@ -45,19 +45,12 @@ typedef struct {
...
-                      AEAD_state_is_too_small)
-#if defined(__GNUC__) || defined(__clang__)
-OPENSSL_STATIC_ASSERT(alignof(union evp_aead_ctx_st_state) >=
-                          alignof(AEAD_TLS_CTX),
-                      AEAD_state_has_insufficient_alignment)
-#endif
-
 static void aead_tls_cleanup(EVP_AEAD_CTX *ctx) {
-  AEAD_TLS_CTX *tls_ctx = (AEAD_TLS_CTX *)&ctx->state;
+  AEAD_TLS_CTX *tls_ctx = (AEAD_TLS_CTX *)ctx->state.ptr;
```
nebeid, line 49: I am probably missing something obvious. Why is this change needed? why allocate memory when I think the opaque array in `evp_aead_ctx_st_state` was supposed to store what is being allocated here?

SalusaSecondus: The existing code reuses the space in `EVP_AEAD_CTX->state` but that was never really an intended use for it. (It's designed to have enough space for AES-GCM.) Previously this worked because the `AEAD_TLS_CTX` was relatively small (because it contained an `HMAC_CTX` which then had pointers to other objects). Now that we are flattening `HMAC_CTX` (by removing pointers) and making it larger, we've done the same to `AEAD_TLS_CTX` and it no longer fits in `EVP_AEAD_CTX->state`. We *could* make `EVP_AEAD_CTX->state` larger, but that means that all uses of AES-GCM will spend extra memory when it is rarely needed. Instead, we just store a pointer in `EVP_AEAD_CTX->state` to newly allocated memory.

(Remember, the old code still allocates memory implicitly because the old `HMAC_CTX` contains pointers which require memory allocation.)

crypto/fipsmodule/hmac/hmac.c
SalusaSecondus: Because that's how [libcrypto defines it](https://www.openssl.org/docs/man1.1.1/man3/SHA1_Final.html) :-(

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -64,34 +64,91 @@
...
+};
+
+#define DEFINE_IN_PLACE_METHODS(EVP_MD, HASH_NAME)               \
+  {                                                              \
+    out->methods[idx].evp_md = EVP_MD;                           \
+    out->methods[idx].init = (HashInit)(HASH_NAME##_Init);       \
+    out->methods[idx].update = (HashUpdate)(HASH_NAME##_Update); \
+    out->methods[idx].digest = (HashFinal)(HASH_NAME##_Final);   \
+    idx++;                                                       \
+    assert(idx < HMAC_METHOD_MAX);                               \
```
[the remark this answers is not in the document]
SalusaSecondus: Yup. I'll fix the assertion. Thank you.

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -64,34 +64,91 @@
...
+  result = HMAC_Init_ex(&ctx, key, key_len, evp_md, NULL) &&
+           HMAC_Update(&ctx, data, data_len) &&
+           HMAC_Final(&ctx, out, out_len);
   FIPS_service_indicator_unlock_state();
-  if(out) {
+  HMAC_CTX_cleanup(&ctx);
+  if (result) {
     HMAC_verify_service_indicator(evp_md);
+    return out;
+  } else {
```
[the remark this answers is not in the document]
SalusaSecondus: We can't set it to `NULL` (it's not a pointer to a pointer) but we can zeroize it. Is that good enough?

crypto/fipsmodule/hmac/hmac.c
SalusaSecondus: I'll restore it.

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -120,119 +175,114 @@ void HMAC_CTX_free(HMAC_CTX *ctx) {
...
-    assert(block_size <= sizeof(key_block));
+  // At this point we know we have valid methods and a context allocated.
+  const HmacMethods *methods = ctx->methods;
+  size_t block_size = EVP_MD_block_size(methods->evp_md);
+  assert(block_size % 8 == 0);
+  assert(block_size <= EVP_MAX_MD_BLOCK_SIZE);
+
+  FIPS_service_indicator_lock_state();
+  int result = 0;
+  if (key != NULL || ctx->initialized != 1) {
```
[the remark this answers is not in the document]
SalusaSecondus: I'll restore it.

Personally, I think that if you want to indicate the null key you pass a valid non-null pointer but `key_len -- 0`.

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -120,119 +175,114 @@ void HMAC_CTX_free(HMAC_CTX *ctx) {
...
+    uint64_t pad[EVP_MAX_MD_BLOCK_SIZE] = {0};
+    uint64_t key_block[EVP_MAX_MD_BLOCK_SIZE] = {0};
     if (block_size < key_len) {
-      // Long keys are hashed.
-      if (!EVP_DigestInit_ex(&ctx->md_ctx, md, impl) ||
-          !EVP_DigestUpdate(&ctx->md_ctx, key, key_len) ||
-          !EVP_DigestFinal_ex(&ctx->md_ctx, key_block, &key_block_len)) {
+      if (!methods->init(&ctx->md_ctx) ||
```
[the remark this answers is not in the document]
SalusaSecondus: Sure.

crypto/fipsmodule/hmac/hmac.c
SalusaSecondus: I don't understand this comment. Neither the old nor new implementation require that the key_len be divisible by 8. The new implementation requires that the `block_size` be divisible by 8. All currently supported hash functions meet this requirement.

crypto/fipsmodule/hmac/hmac.c
SalusaSecondus: Will restore.

crypto/fipsmodule/hmac/hmac.c
SalusaSecondus: Will restore.

crypto/fipsmodule/hmac/hmac.c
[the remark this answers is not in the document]
SalusaSecondus: Actually, we don't know how large the space is to wipe. We can make some reasonable guesses.

More importantly I see that we need to handle a case where `out == NULL` (though it doesn't need to be thread-safe).

SalusaSecondus:
As of this point, all CodeBuild tests **other than the HMAC formal proofs** appear to be passing successfully.

crypto/cipher_extra/e_tls.c
torben-hansen, line 49: You deleted the static asserts. What about the other static asserts in the code?
```
grep -r "evp_aead_ctx_st_state" ./                                                                                                                                                      
.//crypto/fipsmodule/cipher/e_aesccm.c:OPENSSL_STATIC_ASSERT(alignof(union evp_aead_ctx_st_state) >=                                                                                            
.//crypto/fipsmodule/cipher/e_aes.c:OPENSSL_STATIC_ASSERT(alignof(union evp_aead_ctx_st_state) >=                                                                                               
.//crypto/fipsmodule/cipher/e_aes.c:OPENSSL_STATIC_ASSERT(alignof(union evp_aead_ctx_st_state) >=                                                                                               
.//crypto/fipsmodule/cipher/e_aes.c:OPENSSL_STATIC_ASSERT(alignof(union evp_aead_ctx_st_state) >=                                                                                               
Binary file .//crypto/fipsmodule/rand/perf_drbg_changes matches                                                                                                                                 
.//crypto/cipher_extra/e_chacha20poly1305.c:OPENSSL_STATIC_ASSERT(alignof(union evp_aead_ctx_st_state) >=                                                                                       
.//crypto/cipher_extra/e_aesctrhmac.c:OPENSSL_STATIC_ASSERT(alignof(union evp_aead_ctx_st_state) >=                                                                                             
.//crypto/cipher_extra/e_aesgcmsiv.c:OPENSSL_STATIC_ASSERT(alignof(union evp_aead_ctx_st_state) >= 8,                                                                                           
.//crypto/cipher_extra/e_aesgcmsiv.c:OPENSSL_STATIC_ASSERT(alignof(union evp_aead_ctx_st_state) >=
```
I haven't looked into whether these can now fail.

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -64,34 +64,139 @@
...
-  HMAC_CTX_init(&ctx);
-  if (!HMAC_Init_ex(&ctx, key, key_len, evp_md, NULL) ||
-      !HMAC_Update(&ctx, data, data_len) ||
-      !HMAC_Final(&ctx, out, out_len)) {
-    out = NULL;
+  HMAC_CTX ctx = {.initialized = 0};
+  int result;
+  // If out is NULL, the digest is placed in a static array.
+  // Note: passing a NULL value for out to use the static array is not thread safe.
+  static uint8_t fallback_output[EVP_MAX_MD_SIZE] = {0};
```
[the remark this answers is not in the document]
torben-hansen: If this is not needed for JNI performance I propose to remove it.

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -64,34 +64,130 @@
...
   FIPS_service_indicator_lock_state();
-  HMAC_CTX ctx;
-  HMAC_CTX_init(&ctx);
-  if (!HMAC_Init_ex(&ctx, key, key_len, evp_md, NULL) ||
-      !HMAC_Update(&ctx, data, data_len) ||
-      !HMAC_Final(&ctx, out, out_len)) {
-    out = NULL;
-  }
-  // Unlock service indicator state and check if HMAC functions have succeeded.
+  HMAC_CTX ctx = {.state = 0};
```
torben-hansen, line 166: we wanna use `HMAC_STATE_UNINITIALIZED` here no?

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -64,34 +64,130 @@
...
+    return out;
+  } else {
+    OPENSSL_cleanse(out, EVP_MD_size(evp_md));
+    return NULL;
   }
-  HMAC_CTX_cleanup(&ctx);
-  return out;
 }
 
 void HMAC_CTX_init(HMAC_CTX *ctx) {
```
torben-hansen, line 189: Do we want to initialise `state` with `HMAC_STATE_UNINITIALIZED`? 

Redundant now - but future-prove it a bit.

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -135,104 +255,99 @@ int HMAC_Init_ex(HMAC_CTX *ctx, const void *key, size_t key_len,
...
-  return 1;
+  if (result) {
+    HMAC_verify_service_indicator(evp_md);
+    if (out_len) {
+      *out_len = hmac_len;
+    }
+    return 1;
+  } else {
+    if (out_len) {
+      *out_len = 0;
```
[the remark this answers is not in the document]
SalusaSecondus: Personally, I'd prefer to not modify `out_len` on failure at all but this change was requested earlier.

1. Should we make this change?
2. Should we commit in our documentation to maintaining this behavior?

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -118,121 +212,169 @@ void HMAC_CTX_free(HMAC_CTX *ctx) {
...
+
+  if (md && ctx->md != md) {
+    // The MD has changed
+    ctx->methods = GetInPlaceMethods(md);
+    if (ctx->methods == NULL) {
+      // Unsupported md
+      return 0;
+    }
     ctx->md = md;
+  } else if (!hmac_ctx_is_initialized(ctx)) {
```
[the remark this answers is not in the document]
SalusaSecondus: Blocklength isn't an input.

However, to avoid any risk of this, if the return value of this method is `0`, I'll reset it to uninitialized.

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -64,34 +64,130 @@
...
+
+struct hmac_methods_st {
+  const EVP_MD* evp_md;
+  HashInit init;
+  HashUpdate update;
+  HashFinal finalize; // Not named final to avoid keywords
+};
+
+// The maximum number of HMAC implementations
+#define HMAC_METHOD_MAX 9
```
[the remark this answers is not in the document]
SalusaSecondus: Fixing

crypto/fipsmodule/hmac/hmac.c
torben-hansen: Decided that `HMAC_STATE_UNINITIALIZED` must remain equal to `0` and added a comment about that.

crypto/fipsmodule/hmac/hmac.c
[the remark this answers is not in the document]
torben-hansen: Keeping as-is to be backwards compatible

torben-hansen:
Update: we are working to modify the formal verification such that it can handle the new code constructions.

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -64,34 +64,154 @@
...
               unsigned int *out_len) {
-  // We have to verify that all the HMAC services actually succeed before
-  // updating the indicator state, so we lock the state here.
+
+  // While returned new contexts are fully zeroed (and that is what we encourage our callers to do),
+  // we know enough about its inner workings to know that setting the state field to HMAC_STATE_UNINITIALIZED
+  // is sufficient for HMAC_Init_ex to get everything into a properly working state.
+  // So this is yet another micro-optimization of only setting a single byte rather than many
+  // (which will just be overwritten in a moment).
+  HMAC_CTX ctx = {.state = HMAC_STATE_UNINITIALIZED};
```
apetcher-amazon, line 189: What happens to all the fields that are not included in the initializer? Are they initialized to 0?

torben-hansen: This object has automatic storage duration. So, I think we want to look at C99 standard 6.7.8.19:
>  The initialization shall occur in initializer list order, each initializer provided for a
particular subobject overriding any previously listed initializer for the same subobject; all
subobjects that are not initialized explicitly shall be initialized implicitly the same as
objects that have static storage duration.

Okay, so back to 6.7.8.10:
> If an object that has automatic storage duration is not initialized explicitly, its value is
indeterminate. If an object that has static storage duration is not initialized explicitly,
then:
— if it has pointer type, it is initialized to a null pointer;
— if it has arithmetic type, it is initialized to (positive or unsigned) zero;
— if it is an aggregate, every member is initialized (recursively) according to these rules;
— if it is a union, the first named member is initialized (recursively) according to these
rules.

Hence, you need to recursively work backwards. But if `0` is `NULL`, then all should end up being 0. 

Would it be easier for you to just zeroise everything at initialisation?

apetcher-amazon: Thanks. That's what I thought was happening. I just wanted to confirm, and make sure the semantics that is used here matches what Clang and SAW are using. 

The code would probably be more clear if it explicitly initialized this struct to `0`, but the formal verification handles the code that is currently in the PR just fine. Leaving some fields uninitialized would work, too, but it looks like the `md` field must be initialized to `0` in order for `HMAC_INIT_ex` to work correctly. Otherwise `if (md && ctx->md != md)`  on line 262 may fail because `ctx->md`  happens to equal the supplied pointer.

SalusaSecondus: You are correct that if `ctx->md` equals the supplied pointer that `HMAC_INIT_ex` will fail (by returning `0` indicating failure). We should simplify this by just replacing it with `HMAC_CTX ctx = {0};`

torben-hansen: We already require this for `HMAC_STATE_UNINITIALIZED`:

> // HMAC_STATE_UNINITIALIZED *MUST* remain `0` so that callers can do `HMAC_CTX ctx = {0};` to get a usable context.

You could also statically assert it - if you are paranoid (:

torben-hansen (approved):

DominicDams (approved):

PR #379, opened 2022-02-01
Add reason codes for backwards compilation compatibility

torben-hansen:
### Issues:
CryptoAlg-954

### Description of changes: 
Add some reason codes from OpenSSL to ease compilation and integration against AWS-LC. Do this under its own "namespace"  for clarity and protect against any upstream merge conflict annoyance.

These will have a different reason code value compared to OpenSSL. But it all works fine if compilation happens against AWS-LC headers, of course!

None of these reason codes are actually returned by AWS-LC. Code paths that condition on these codes might break. Grep for `ERR_GET_REASON` and `ERR_GET_FUNC` to inspect.

### DIFF
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -5635,4 +5635,22 @@ BSSL_NAMESPACE_END
 #define SSL_R_TLSV1_ALERT_NO_APPLICATION_PROTOCOL 1120
 #define SSL_R_TLSV1_ALERT_ECH_REQUIRED 1121
 
+// Define some reason codes from OpenSSL to ease compilation against AWS-LC.
+// None of these reason codes are actually returned by AWS-LC.
+// Code paths that condition on these codes might break. Grep for
+// |ERR_GET_REASON| and |ERR_GET_FUNC| to inspect.
+
+// Per C99 standard 6.8.4.2.1 the controlling expression in the switch statement
+// has integer type. Hence it can handle quite large offset values assuming just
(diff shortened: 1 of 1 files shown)


bryce-shang (approved):

include/openssl/ssl.h
```diff
@@ -5635,4 +5635,22 @@ BSSL_NAMESPACE_END
...
+// Code paths that condition on these codes might break. Grep for
+// |ERR_GET_REASON| and |ERR_GET_FUNC| to inspect.
+
+// Per C99 standard 6.8.4.2.1 the controlling expression in the switch statement
+// has integer type. Hence it can handle quite large offset values assuming just
+// two bytes. Just pick a somewhat large random value to prevent collisions on
+// later macro definitions from upstream.
+#define SSL_R_BACKWARDS_COMPATABILITY_OFFSET 0x5D21
+
+// See CryptoAlg-954.
```
nebeid, line 5649: I suggest that instead of referring to CryptoAlg-954 in the code, since it's in the PR, we expand the comment above that something like:
these codes exist in OpenSSL with different values. So a user of AWS-LC can check to see that they're different. However, no part of AWS-LC depends on them nor returns them.

torben-hansen: Whether these macro's have different values, comparing AWS-LC and OpenSSL, is not important imo. It is important whether a consumer conditions on the codes i.e. makes assumptions based on the returned error code which could affect correctness.

nebeid (approved):

PR #386, opened 2022-02-09
Integrate s2n-bignum alternative functions for P-384

dkostic:
### Issues:
CryptoAlg-946

### Description of changes: 

On x86_64 platforms s2n-bignum uses bmi2 and adx instruction sets
for some of the P-384 functions. These instructions are not supported
by every x86 CPU. Previously, we checked in runtime if bmi2 and adx
are supported and if not we fallback to Fiat's implementation. This
solution had a drawbacks that we had to always compile both
s2n-bignum and Fiat code.

s2n-bignum implemented "alternative" implementations of functions
that don't require any special instructions. In this commit we integrate
those changes and instead of falling-back to Fiat's code we fall back
to s2n-bignum alternative implementations. The change has two
benefits:
 - s2n-bignum and Fiat are decoupled so we compile only one.
 - Performance improvement on x86 CPUs that don't support bmi2
 and adx instructions is x2.11, x1.58, x2.03 for ECDH, ECDSA sign,
 ECDSA verify, respectively.

The benchmark for a CPU without bmi2 and adx support, with
`bssl/speed -filter “P-384”`, before and after:
ECDH: 1594 —> 3372 (ops/s)
ECDSA Sign: 6332 —> 10000 (ops/s)
ECDSA Verify: 1981 —> 4019 (ops/s)

### Call-outs:

### Testing:

### DIFF
--- a/crypto/fipsmodule/CMakeLists.txt
+++ b/crypto/fipsmodule/CMakeLists.txt
@@ -163,10 +163,17 @@ if (S2N_BIGNUM_DIR AND UNIX AND NOT APPLE)
     bignum_tomont_p384.S
     bignum_deamont_p384.S
     bignum_montmul_p384.S
+    bignum_montmul_p384_alt.S
     bignum_montsqr_p384.S
+    bignum_montsqr_p384_alt.S
     bignum_nonzero_6.S
     bignum_littleendian_6.S
   )
+  if(ARCH STREQUAL "x86_64")
+    list(APPEND S2N_BIGNUM_ASM_SOURCES
+                bignum_tomont_p384_alt.S
+                bignum_deamont_p384_alt.S)
+  endif()
 
   # Copy s2n-bignum files to the current binary dir
   foreach(f ${S2N_BIGNUM_ASM_SOURCES})
--- a/crypto/fipsmodule/CMakeLists.txt
+++ b/crypto/fipsmodule/CMakeLists.txt
@@ -170,6 +170,10 @@ if (S2N_BIGNUM_DIR AND UNIX AND NOT APPLE)
     bignum_littleendian_6.S
   )
   if(ARCH STREQUAL "x86_64")
+    # The files below contain the alternative functions for x86_64.
+    # For AArch64, the alternative tomont/deamont functions are
+    # the same as the non-alternative ones, so they are defined
+    # in the corresponding "non-alt" files (bignum_<deamont|tomont>_p384.S)
     list(APPEND S2N_BIGNUM_ASM_SOURCES
                 bignum_tomont_p384_alt.S
                 bignum_deamont_p384_alt.S)
--- a/crypto/fipsmodule/ec/p384.c
+++ b/crypto/fipsmodule/ec/p384.c
@@ -5,9 +5,6 @@
 ------------------------------------------------------------------------------------
 */
 
-// Implementation of P-384 that uses Fiat-crypto for the field arithmetic
-// found in third_party/fiat, similarly to p256.c
-
 #include <openssl/bn.h>
 #include <openssl/ec.h>
 #include <openssl/err.h>
@@ -16,28 +13,6 @@
 #include "../bn/internal.h"
 #include "../delocate.h"
 #include "internal.h"
-
-#if defined(BORINGSSL_HAS_UINT128)
-#define BORINGSSL_NISTP384_64BIT 1
-#include "../../../third_party/fiat/p384_64.h"
-#else
-#include "../../../third_party/fiat/p384_32.h"
-#endif
-
-#if defined(BORINGSSL_NISTP384_64BIT)
-#define P384_NLIMBS (6)
-typedef uint64_t p384_limb_t;
-typedef uint64_t p384_felem[P384_NLIMBS];
-static const p384_felem p384_felem_one = {
-    0xffffffff00000001, 0xffffffff, 0x1, 0x0, 0x0, 0x0};
-#else  // 64BIT; else 32BIT
-#define P384_NLIMBS (12)
-typedef uint32_t p384_limb_t;
-typedef uint32_t p384_felem[P384_NLIMBS];
-static const p384_felem p384_felem_one = {
-    0x1, 0xffffffff, 0xffffffff, 0x0, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0};
-#endif  // 64BIT
-
 
 // We have two implementations of the field arithmetic for P-384 curve:
 //   - Fiat-crypto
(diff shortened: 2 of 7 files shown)


third_party/fiat/p384_32.h
```diff
@@ -29,6 +29,7 @@ OPENSSL_UNUSED static void fiat_p384_set_one(uint32_t out1[6]);
 OPENSSL_UNUSED static void fiat_p384_msat(uint32_t out1[7]);
 OPENSSL_UNUSED static void fiat_p384_divstep(uint32_t* out1, uint32_t out2[7], uint32_t out3[7], uint32_t out4[6], uint32_t out5[6], uint32_t arg1, const uint32_t arg2[7], const uint32_t arg3[7], const uint32_t arg4[6], const uint32_t arg5[6]);
 OPENSSL_UNUSED static void fiat_p384_divstep_precomp(uint32_t out1[6]);
+OPENSSL_UNUSED static void fiat_p384_selectznz(uint32_t out1[12], fiat_p384_uint1 arg1, const uint32_t arg2[12], const uint32_t arg3[12]);
```
nebeid, line 32: Do we need to add it to p384_64.h as well?

third_party/s2n-bignum/include/s2n-bignum_aws-lc.h
```diff
@@ -19,40 +19,57 @@
 
 // Add modulo p_384, z := (x + y) mod p_384, assuming x and y reduced
 // Inputs x[6], y[6]; output z[6]
-extern void bignum_add_p384 (uint64_t z[static 6], const uint64_t x[static 6], const uint64_t y[static 6]);
+extern void bignum_add_p384(uint64_t z[static 6], const uint64_t x[static 6], const uint64_t y[static 6]);
```
nebeid, line 22: I think we should add to this file something about the meaning of "_alt" in the function names as in https://github.com/awslabs/s2n-bignum/blob/main/include/s2n-bignum.h#L27-L35

crypto/fipsmodule/CMakeLists.txt
```diff
@@ -163,10 +163,17 @@ if (S2N_BIGNUM_DIR AND UNIX AND NOT APPLE)
     bignum_tomont_p384.S
     bignum_deamont_p384.S
     bignum_montmul_p384.S
+    bignum_montmul_p384_alt.S
     bignum_montsqr_p384.S
+    bignum_montsqr_p384_alt.S
```
nebeid, line 168: Does this mean these _alt functions are included in the arm build? If yes, are they needed now?

dkostic: yes they are included but not used right now, I explained in the comment here: https://github.com/awslabs/aws-lc/pull/386/files#diff-69efd8604b81a4adc44b8fa275c47b60ed2c615cc38fe1b7e345b08871a3b711R82-R87

Two reasons I included them: simplifies the code because if I exclude those we'd have to have ifdefs to differentiate x86 and arm, and second, I was hoping we soon find a way to make use of them (i.e., detect the platforms the alternatives are faster on).

Would you prefer I remove them?

third_party/fiat/p384_32.h
dkostic: it's there already, but there were also some unnecessary ones so I removed them.

third_party/s2n-bignum/include/s2n-bignum_aws-lc.h
dkostic: good idea, added.

crypto/fipsmodule/CMakeLists.txt
nebeid: Thanks for the explanation. May be then we can put a comment after l.172: 
"The following files contain functions of the same name for x86_64. However, on AArch64, these functions are defined as the same "non-alt" functions in their respective files, see third_party/s2n-bignum/arm/p384/bignum_<deamont|tomont>_p384.S"

dkostic: Done, thanks!

nebeid (approved):

crypto/fipsmodule/ec/p384.c
```diff
@@ -56,18 +31,60 @@ static const p384_felem p384_felem_one = {
...
 
 #if defined(OPENSSL_X86_64)
 // On x86_64 platforms s2n-bignum uses bmi2 and adx instruction sets
-// for some of the functions so we have to check if they are available.
-static inline uint8_t use_s2n_bignum(void) {
-  return ((OPENSSL_ia32cap_P[2] & (1u <<  8)) != 0) && // bmi2
-         ((OPENSSL_ia32cap_P[2] & (1u << 19)) != 0);   // adx
+// for some of the functions. These instructions are not supported by
+// every x86 CPU so we have to check if they are available and in case
+// they are not we fallback to slightly slower but generic implementation.
```
torben-hansen, line 76: Fiat? I dunno if it is even possible to switch between fiat and bignum at run-time anyway. It would both increase the binary size and complicate the code I suppose.

crypto/fipsmodule/CMakeLists.txt
```diff
@@ -163,10 +163,21 @@ if (S2N_BIGNUM_DIR AND UNIX AND NOT APPLE)
     bignum_tomont_p384.S
     bignum_deamont_p384.S
     bignum_montmul_p384.S
+    bignum_montmul_p384_alt.S
```
torben-hansen, line 166: Is "alt" and abbreviation for "alternative". Maybe not the best name if there is a third "alternative" solution.

crypto/fipsmodule/ec/p384.c
```diff
@@ -78,20 +95,20 @@ static inline uint8_t use_s2n_bignum(void) { return 1; }
 
 // The following four functions need bmi2 and adx support.
 #define p384_felem_mul(out, in0, in1) \
-  if (use_s2n_bignum()) bignum_montmul_p384(out, in0, in1); \
-  else fiat_p384_mul(out, in0, in1);
+  if (use_s2n_bignum_alt()) bignum_montmul_p384_alt(out, in0, in1); \
```
torben-hansen, line 98: I hope the cpu is good at branch prediction.

third_party/s2n-bignum/include/s2n-bignum_aws-lc.h
```diff
@@ -17,42 +17,69 @@
 // C prototypes for s2n-bignum functions used in AWS-LC
 // ----------------------------------------------------------------------------
 
+// For some functions there are additional variants with names ending in
+// "_alt". These have the same core mathematical functionality as their
+// non-"alt" versions, but can be better suited to some microarchitectures:
+//
+//      - On x86, the "_alt" forms avoid BMI and ADX instruction set
```
torben-hansen, line 24: Does this actually matter? I assume the set of "alt" function is not enough to implement all the feature we want on top of the p384 curve.

dkostic: Not sure I understand. I'll give an example of what's going on. s2n-bignum has `bignum_add_p384` function that is generic in a sense that it doesn't use any special x86 instruction. There is also `bignum_mul_p384` which uses bmi2 and adx instructions that are not available on all x86 CPUs, and the alternative version `bignum_mul_p384_alt` that is generic. So if aws-lc is run on a:
1. CPU with bmi2 and adx then we use `bignum_add_p384` and `bignum_mul_p384`,
2. CPU without bmi2 and adx then we use `bignum_add_p384` and `bignum_mul_p384_alt`.

crypto/fipsmodule/CMakeLists.txt
torben-hansen: Good thing we can influence s2n-bignum quite easily then ;)

third_party/s2n-bignum/include/s2n-bignum_aws-lc.h
torben-hansen: a'ha - I recall now!
Thanks for explaining it on the 5 years old level, needed that :)

crypto/fipsmodule/ec/p384.c
torben-hansen: Ai ai by generic I thought you meant portable C...

crypto/fipsmodule/ec/p384.c
torben-hansen: I recall, comment was mostly a jinx; when we hit the random CPU with bad branch predictor.

torben-hansen (approved):

PR #398, opened 2022-02-22
Support old assemblers that don't know about AVX

dkostic:
### Issues:
CryptoAlg-1011

### Description of changes: 
We assume that the assembler that is going to be used to compile the assembly code in the library is modern enough that it knows about AVX. However, this assumption is not always correct for our internal systems. With this patch we add logic to handle the case of ancient compilers that don't support AVX.  For this purpose we define a CMake flag called `MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX` that can be set to `TRUE` in case the assembler doesn't understand AVX instructions. In perl-assembly files we check if `MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX` flag is specified and don't generate the AVX code in case it is. Regarding pure assembly files that have AVX implementations -- we have to remove them from compilation all together.

Note that we also simplified and unified how our Perl-assembly scripts handle command line arguments. Previously, some scripts were expecting the `flavour` and `output` arguments to be the first two arguments (if given), while other scripts assumed `flavour` to be the first one and `output` to be the last argument. Now, all the scripts require at least two arguments, the first one being the `flavour`, the second one being the `output` file path, and optionally, additional flags starting from the third argument.

### Call-outs:
N/A

### Testing:
N/A

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,9 +193,9 @@ endmacro()
 # Some ancient assemblers don't know about AVX instructions, which is an
 # assumption we make for some of the assembly implementations. This flag
 # can be set to handle such cases.
+option(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX "Exclude AVX code from the build" OFF)
 if(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
-  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DMY_ASSEMBLER_IS_TOO_OLD_FOR_AVX")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DMY_ASSEMBLER_IS_TOO_OLD_FOR_AVX")
+  add_definitions(-DMY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
 endif()
 
 # Detect if memcmp is wrongly stripped like strcmp.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -190,6 +190,14 @@ macro(check_run file_to_test flag_to_set compile_flags)
   endif()
 endmacro()
 
+# Some ancient assemblers don't know about AVX instructions, which is an
+# assumption we make for some of the assembly implementations. This flag
+# can be set to handle such cases.
+if(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DMY_ASSEMBLER_IS_TOO_OLD_FOR_AVX")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DMY_ASSEMBLER_IS_TOO_OLD_FOR_AVX")
+endif()
+
 # Detect if memcmp is wrongly stripped like strcmp.
 # If exists, let CMake generate a warning.
 # memcmp bug link https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189.
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -74,10 +74,14 @@ if(PERL_EXECUTABLE)
       set(dir ".")
     endif()
 
+    if(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
+      set(PERLASM_FLAGS "${PERLASM_FLAGS} -DMY_ASSEMBLER_IS_TOO_OLD_FOR_AVX")
+    endif()
+
     add_custom_command(
       OUTPUT ${dest}
       COMMAND ${CMAKE_COMMAND} -E make_directory ${dir}
-      COMMAND ${PERL_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/${src} ${PERLASM_STYLE} ${PERLASM_FLAGS} ${ARGN} ${dest}
+      COMMAND ${PERL_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/${src} ${PERLASM_STYLE} ${dest} ${PERLASM_FLAGS}
       DEPENDS
       ${src}
       ${PROJECT_SOURCE_DIR}/crypto/perlasm/arm-xlate.pl
@@ -182,11 +186,14 @@ if(ARCH STREQUAL "x86_64")
     CRYPTO_ARCH_SOURCES
 
     chacha/chacha-x86_64.${ASM_EXT}
-    cipher_extra/aes128gcmsiv-x86_64.${ASM_EXT}
     cipher_extra/chacha20_poly1305_x86_64.${ASM_EXT}
-    hrss/asm/poly_rq_mul.S
     test/trampoline-x86_64.${ASM_EXT}
   )
+  if(NOT MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
+    list(APPEND CRYPTO_ARCH_SOURCES
+                cipher_extra/aes128gcmsiv-x86_64.${ASM_EXT}
+                hrss/asm/poly_rq_mul.S)
+  endif()
 endif()
 
 if(GO_EXECUTABLE)
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -187,13 +187,10 @@ if(ARCH STREQUAL "x86_64")
 
     chacha/chacha-x86_64.${ASM_EXT}
     cipher_extra/chacha20_poly1305_x86_64.${ASM_EXT}
+    cipher_extra/aes128gcmsiv-x86_64.${ASM_EXT}
     test/trampoline-x86_64.${ASM_EXT}
+    hrss/asm/poly_rq_mul.S
   )
-  if(NOT MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
-    list(APPEND CRYPTO_ARCH_SOURCES
-                cipher_extra/aes128gcmsiv-x86_64.${ASM_EXT}
-                hrss/asm/poly_rq_mul.S)
-  endif()
 endif()
 
 if(GO_EXECUTABLE)
--- a/crypto/chacha/asm/chacha-armv4.pl
+++ b/crypto/chacha/asm/chacha-armv4.pl
@@ -34,9 +34,12 @@
 #	but then Snapdragon S4 and Cortex-A8 results get
 #	20-25% worse;
 
+# The first two arguments should always be the flavour and output file path.
+if ($#ARGV < 1) { die "Not enough arguments provided.
+  Two arguments are necessary: the flavour and the output file path."; }
+
 $flavour = shift;
-if ($flavour=~/\w[\w\-]*\.\w+$/) { $output=$flavour; undef $flavour; }
-else { while (($output=shift) && ($output!~/\w[\w\-]*\.\w+$/)) {} }
+$output = shift;
 
 if ($flavour && $flavour ne "void") {
     $0 =~ m/(.*[\/\\])[^\/\\]+$/; $dir=$1;
--- a/crypto/chacha/asm/chacha-armv8.pl
+++ b/crypto/chacha/asm/chacha-armv8.pl
@@ -35,8 +35,12 @@
 #	higher instruction issue rate;
 # (**)	expected improvement was actually higher;
 
-$flavour=shift;
-$output=shift;
+# The first two arguments should always be the flavour and output file path.
+if ($#ARGV < 1) { die "Not enough arguments provided.
+  Two arguments are necessary: the flavour and the output file path."; }
+
+$flavour = shift;
+$output = shift;
 
 $0 =~ m/(.*[\/\\])[^\/\\]+$/; $dir=$1;
 ( $xlate="${dir}arm-xlate.pl" and -f $xlate ) or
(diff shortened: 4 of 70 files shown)


crypto/chacha/asm/chacha-x86_64.pl
```diff
@@ -65,7 +65,14 @@
...
 
-$avx = 2;
+if (`$ENV{CC} -Wa,-v -c -o /dev/null -x assembler /dev/null 2>&1`
+               =~ /GNU assembler version ([2-9]\.[0-9]+)/) {
+       $avx = ($1>=2.19) + ($1>=2.22);
+}
+
+if (!$avx && `$ENV{CC} -v 2>&1` =~ /((?:clang|LLVM) version|.*based on LLVM) ([0-9]+\.[0-9]+)/) {
+       $avx = ($2>=3.0) + ($2>3.0);
+}
```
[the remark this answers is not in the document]
dkostic: The code works provided that `CC` environment variable is set (which I assumed we always do when we build). For example,
```
export CC=gcc-7 CXX=g++-7
cmake ....
make
```
will make the Perl scripts correctly detect the compiler/assembler.

Is my assumption about how we build wrong?

torben-hansen:
If I as a user can declare `MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX`, then do we need the logic in e.g. `rsaz-avx2.pl`?

Maybe I don't understand what it is doing - I think it needs some explanation because it is not "obvious".

crypto/chacha/asm/chacha-x86_64.pl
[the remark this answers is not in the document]
bryce-shang: The assumption is correct. I found I forgot to preserve environment variables when using sudo access(which is needed during test to pull some golang pkgs).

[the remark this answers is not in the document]
bryce-shang: Besides, let's say `CC` env var is set, after these `if` logics, the `avx` can be `empty""`, `1` or `2`. The current code uses hardcoded value `2` for `avx`. Bssl might have removed some assembly code blocks when `avx` == 1 or empty.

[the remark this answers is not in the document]
dkostic: the `avx` flag determines the highest optimization level to be compiled. For example, if `avx=2` the generated assembly will contain: generic implementation (corresponding to `avx=0`), AVX implementation (corresponding to `avx=1`), and AVX2 implementation (corresponding to `avx=2`). The choice which of the three implementations to run is done in runtime based on CPU capabilities.

BoringSSL hardcodes `avx=2` because they think there is no reason not to _generate_ the AVX2 code (will it be run or not is a runtime decision). This thinking is right, provided that your assembler is modern enough to know about AVX2. This is not the case for us, and that's why we, unfortunately, have to check if the assembler supports AVX2 before generating the code.

CMakeLists.txt
```diff
@@ -190,6 +190,14 @@ macro(check_run file_to_test flag_to_set compile_flags)
   endif()
 endmacro()
 
+# Some ancient assemblers don't know about AVX instructions, which is an
+# assumption we make for some of the assembly implementations. This flag
+# can be set to handle such cases.
+if(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
```
torben-hansen, line 196: use cmake `option`?

CMakeLists.txt
```diff
@@ -190,6 +190,14 @@ macro(check_run file_to_test flag_to_set compile_flags)
...
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DMY_ASSEMBLER_IS_TOO_OLD_FOR_AVX")
```
torben-hansen, line 197: Use cmake `add_definitions`?

crypto/CMakeLists.txt
```diff
@@ -182,11 +186,14 @@ if(ARCH STREQUAL "x86_64")
     CRYPTO_ARCH_SOURCES
 
     chacha/chacha-x86_64.${ASM_EXT}
-    cipher_extra/aes128gcmsiv-x86_64.${ASM_EXT}
     cipher_extra/chacha20_poly1305_x86_64.${ASM_EXT}
-    hrss/asm/poly_rq_mul.S
     test/trampoline-x86_64.${ASM_EXT}
   )
+  if(NOT MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
```
torben-hansen, line 192: Instead of this, we can just guard everything in those files with `MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX`?

crypto/chacha/asm/chacha-x86_64.pl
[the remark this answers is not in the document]
dkostic: the latest changes remove the automatic detection of assembler version and use the compile flag instead.

CMakeLists.txt
dkostic: done

CMakeLists.txt
dkostic: done

crypto/CMakeLists.txt
dkostic: we can, but I thought this way is more clear. If you strongly prefer the other way I'll update the PR?

torben-hansen: My own preference is to exclude it in the file. My thought is that if we add a second condition `C` where we should only exclude file `hrss/asm/poly_rq_mul.S` from the build. Then that would be harder to implement. While if we do the exclusion in the file, we can simply do `&&`s.

On the other hand, this pattern cannot be used with third-party code we pull in e.g. s2n-bignum...

Thoughts?

dkostic: I agree, so I changed how we handle internal files. For the third-party files I think it's better to simply exclude them from the build

torben-hansen (dismissed):

crypto/chacha/asm/chacha-armv4.pl
```diff
@@ -34,9 +34,12 @@
...
 #	20-25% worse;
 
+# The first two arguments should always be the flavour and output file path.
+if ($#ARGV < 1) { die "Not enough arguments provided.
+  Two arguments are necessary: the flavour and the output file path."; }
+
 $flavour = shift;
-if ($flavour=~/\w[\w\-]*\.\w+$/) { $output=$flavour; undef $flavour; }
-else { while (($output=shift) && ($output!~/\w[\w\-]*\.\w+$/)) {} }
+$output = shift;
```
bryce-shang, line 42: This PR fix target is the old compiler and only `x86_64`. Can the fix be more scoped to avoid the change on arm file?

crypto/chacha/asm/chacha-armv8.pl
```diff
@@ -35,8 +35,12 @@
...
 # (**)	expected improvement was actually higher;
 
-$flavour=shift;
-$output=shift;
+# The first two arguments should always be the flavour and output file path.
+if ($#ARGV < 1) { die "Not enough arguments provided.
+  Two arguments are necessary: the flavour and the output file path."; }
+
+$flavour = shift;
+$output = shift;
```
bryce-shang, line 43: Same as above.

crypto/chacha/asm/chacha-x86.pl
```diff
@@ -38,14 +38,18 @@
...
 $0 =~ m/(.*[\/\\])[^\/\\]+$/; $dir=$1;
 push(@INC,"${dir}","${dir}../../perlasm");
 require "x86asm.pl";
 
-$output=pop;
+$output=$ARGV[1];
 open STDOUT,">$output";
 
-&asm_init($ARGV[0],$ARGV[$#ARGV] eq "386");
+&asm_init($ARGV[0],$ARGV[1] eq "386");
```
bryce-shang, line 52: Same as above. I think our support should only consider `x86_64`.

crypto/chacha/asm/chacha-x86_64.pl
```diff
@@ -54,9 +54,12 @@
 #
 # Modified from upstream OpenSSL to remove the XOP code.
 
+# The first two arguments should always be the flavour and output file path.
+if ($#ARGV < 1) { die "Not enough arguments provided.
+  Two arguments are necessary: the flavour and the output file path."; }
+
```
bryce-shang, line 60: Nit: We have control of this file and the input arguments. To reduce the merge conflicts from upstream, instead of `die`/`fail the build`, can the `die` code block be replaced with current logics?

crypto/fipsmodule/cpucap/cpu_intel.c
```diff
@@ -114,7 +114,9 @@ static uint64_t OPENSSL_xgetbv(uint32_t xcr) {
   return (uint64_t)_xgetbv(xcr);
 #else
   uint32_t eax, edx;
-  __asm__ volatile ("xgetbv" : "=a"(eax), "=d"(edx) : "c"(xcr));
+  // Some old assemblers don't support the xgetbv instruction so we emit
+  // the opcode of xgetbv directly.
+  __asm__ volatile (".byte 0x0f, 0x01, 0xd0" : "=a"(eax), "=d"(edx) : "c"(xcr));
```
bryce-shang, line 119: Is the related reference available? 
`xgetbv` to `".byte 0x0f, 0x01, 0xd0"`

Is this transfer always guaranteed(by what)? I think it's safer to enable this transfer only when `MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX` is defined.

bryce-shang (commented):
Overall, the fix on old assembler and `x86_64` should be more scoped. e.g. The `arm` assembly files change can be removed.

CMakeLists.txt
```diff
@@ -190,6 +190,14 @@ macro(check_run file_to_test flag_to_set compile_flags)
...
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DMY_ASSEMBLER_IS_TOO_OLD_FOR_AVX")
```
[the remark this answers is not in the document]
bryce-shang: This was addressed by Torben's comment.

crypto/CMakeLists.txt
```diff
@@ -74,10 +74,14 @@ if(PERL_EXECUTABLE)
...
 
+    if(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
+      set(PERLASM_FLAGS "${PERLASM_FLAGS} -DMY_ASSEMBLER_IS_TOO_OLD_FOR_AVX")
+    endif()
+
     add_custom_command(
       OUTPUT ${dest}
       COMMAND ${CMAKE_COMMAND} -E make_directory ${dir}
-      COMMAND ${PERL_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/${src} ${PERLASM_STYLE} ${PERLASM_FLAGS} ${ARGN} ${dest}
+      COMMAND ${PERL_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/${src} ${PERLASM_STYLE} ${dest} ${PERLASM_FLAGS}
```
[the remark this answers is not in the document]
dkostic: `ARGN` holds the list of arguments passed to a CMake function besides the ones that the function expects. In our case the function `perlasm` expects two arguments `dest` and `src`, if you give any additional argument it is going to end up in `ARGN`. We don't use this anywhere in our build scripts so I removed it to simplify the function. If need be in the future, we can add it back, but there is no reason to keep it right now.

crypto/chacha/asm/chacha-armv4.pl
dkostic: I explained the reason for changes in all Perl-asm files in a comment above.

crypto/chacha/asm/chacha-x86_64.pl
dkostic: I explained the reason for this change in all Perl-asm files in a comment above. It significantly simplifies command line parsing in our scripts precisely because we have control of the input arguments. The previous parsing was implemented for cases where we don't have full control and that's why it was complicated and involved even regular expressions matching.

crypto/CMakeLists.txt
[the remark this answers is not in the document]
bryce-shang: > We don't use this anywhere in our build scripts so I removed it to simplify the function. If need be in the future, we can add it back, but there is no reason to keep it right now.

Not an issue. But `ARGN` is not used in upstream either. After removing, how do we know it is used before.

CMakeLists.txt
```diff
@@ -190,6 +190,14 @@ macro(check_run file_to_test flag_to_set compile_flags)
   endif()
 endmacro()
 
+# Some ancient assemblers don't know about AVX instructions, which is an
+# assumption we make for some of the assembly implementations. This flag
+# can be set to handle such cases.
+option(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX "Exclude AVX code from the build" OFF)
+if(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
```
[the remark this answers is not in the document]
bryce-shang: Why this PR fix needs to handle the diff expectations? I think the current fix passes the argument(used to disable `avx`) through the interface(which is not extensible enough to support our cases). Maybe an env var can be emitted when the `MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX` flag is defined, and inside the perl file, `avx` is unset if the env is used.

bryce-shang (approved):

torben-hansen (approved):

PR #400, opened 2022-02-22
Integration of AES-XTS AArch64 implementation

nebeid:
### Issues:
Resolves CryptoAlg-929

### Description of changes:
This PR integrates the AArch64 implementation of AES-XTS into AWS-LC. The source is from this OpenSSL's commit: 
https://github.com/openssl/openssl/commit/9ce8e0d17e608de4f85f7543c52b146e3c6a2291:
* Some comments were added to the Perl Assembly implementation on `{prefix}_xts_encrypt`.
* BTI instructions were added at the beginning of `{prefix}_xts_encrypt` and `{prefix}_xts_decrypt`.
* Test vectors were added in order to test the different paths of the AArch64 implementation.
* Checks that the 2 keys are different and that the input size does not exceed 2^{20} are also added to the functions `aes_xts_init_key` and `aes_xts_cipher`, respectively.

### Call-outs
* `libdecrepit` was added as a dependency of the `bssl` tool in order to run the speed test.
* The assembly code added should work on Arm (other than Armv8) and would be chosen on those platforms but it was not tested other than on Armv8.

#### Init and Encrypt
|      Size   | Before (MB/s)(1) | After (MB/s) | Improvement |
--------------:|------------------:|--------------:|:--------------:
|    16 bytes |        88        |       91     |    1.03x    |
|   256 bytes |       647        |      883     |    1.36x    |
|  1350 bytes |       971        |     1804     |    1.86x    |
|  8192 bytes |      1097        |     2334     |    2.13x    |
| 16384 bytes |      1109        |     2398     |    2.16x    |

#### Init and Decrypt:
|      Size   | Before (MB/s)(1) | After (MB/s) | Improvement |
--------------:|------------------:|--------------:|:--------------:
|    16 bytes |        83        |       85     |    1.02x    |
|   256 bytes |       621        |      810     |    1.30x    |
|  1350 bytes |       948        |     1746     |    1.84x    |
|  8192 bytes |      1091        |     2289     |    2.10x    |
| 16384 bytes |      1105        |     2356     |    2.13x    |

(1) C implementation which includes AES block encryption in assembly.


decrepit/xts/xts.c
```diff
@@ -164,6 +165,27 @@ static int aes_xts_init_key(EVP_CIPHER_CTX *ctx, const uint8_t *key,
 
   if (key) {
     // key_len is two AES keys
+    /*
+     * Verify that the two keys are different.
```
torben-hansen, line 169: comment-style

decrepit/xts/xts_test.cc
```diff
@@ -266,3 +507,69 @@ TEST(XTSTest, TestVectors) {
...
+  bssl::ScopedEVP_CIPHER_CTX ctx;
+  ASSERT_FALSE(EVP_EncryptInit_ex(ctx.get(), cipher, nullptr, key.data(),
+                                 iv.data()));
+}
+
+// Negative test for input length
+TEST(XTSTest, InputTooLong) {
+
+  // The length of the input will be (wrongly) provided as larger than
+  // XTS_MAX_BLOCKS_PER_DATA_UNIT
```
torben-hansen, line 539:
```suggestion
  // XTS_MAX_BLOCKS_PER_DATA_UNIT.
```
:)

decrepit/xts/xts_test.cc
```diff
@@ -266,3 +507,69 @@ TEST(XTSTest, TestVectors) {
...
+  // The ciphertext does not correspond to the plaintext
+  // which does not matter since it will on length check.
```
torben-hansen, line 541:
```suggestion
  // which does not matter since it will fail on length check.
```

decrepit/xts/xts_test.cc
```diff
@@ -266,3 +507,69 @@ TEST(XTSTest, TestVectors) {
...
+  ASSERT_TRUE(DecodeHex(&ciphertext, kXTSWrongVector.ciphertext_hex));
+
+  bssl::ScopedEVP_CIPHER_CTX ctx;
+  ASSERT_TRUE(EVP_EncryptInit_ex(ctx.get(), cipher, nullptr, key.data(),
+                                 iv.data()));
+  int len;
+  std::vector<uint8_t> out(plaintext.size());
+  ASSERT_FALSE(
+    EVP_EncryptUpdate(ctx.get(), out.data(), &len, plaintext.data(),
+                      (XTS_MAX_BLOCKS_PER_DATA_UNIT * AES_BLOCK_SIZE) +1));
```
torben-hansen, line 565:
```suggestion
                      (XTS_MAX_BLOCKS_PER_DATA_UNIT * AES_BLOCK_SIZE) + 1));
```

decrepit/xts/xts_test.cc
```diff
@@ -266,3 +507,69 @@ TEST(XTSTest, TestVectors) {
...
+
+  // Test Decryption
+  ctx.Reset();
+  ASSERT_TRUE(EVP_DecryptInit_ex(ctx.get(), cipher, nullptr, key.data(),
+                                 iv.data()));
+  ASSERT_FALSE(
+    EVP_DecryptUpdate(ctx.get(), out.data(), &len, ciphertext.data(),
+                      (XTS_MAX_BLOCKS_PER_DATA_UNIT * AES_BLOCK_SIZE) +1));
```
torben-hansen, line 573:
```suggestion
                      (XTS_MAX_BLOCKS_PER_DATA_UNIT * AES_BLOCK_SIZE) + 1));
```

decrepit/xts/xts.c
```diff
@@ -193,10 +215,32 @@ static int aes_xts_cipher(EVP_CIPHER_CTX *ctx, uint8_t *out,
       !xctx->xts.key2 ||
       !out ||
       !in ||
-      len < AES_BLOCK_SIZE ||
-      !CRYPTO_xts128_encrypt(&xctx->xts, ctx->iv, in, out, len, ctx->encrypt)) {
+      len < AES_BLOCK_SIZE) {
```
torben-hansen, line 218: Haven't studied this scheme in depth. I see the underlying implementation has this limitation. Is it inherent to the XTS scheme or just an implementation choice?

decrepit/xts/xts.c
```diff
@@ -163,7 +164,28 @@ static int aes_xts_init_key(EVP_CIPHER_CTX *ctx, const uint8_t *key,
...
+    //
+    // FIPS 140-2 IG A.9 XTS-AES Key Generation Requirements states
+    // that:
+    //      "The check for Key_1 != Key_2 shall be done at any place
+    //       BEFORE using the keys in the XTS-AES algorithm to process
+    //       data with them."
+    //
     // key_len is two AES keys
+
+    if (OPENSSL_memcmp(key, key + ctx->key_len / 2, ctx->key_len / 2) == 0) {
```
torben-hansen, line 184: Compile should just replace this with a right-shift. Should we do it for it?

If we are only supporting a fixed key length - maybe this computation could even just be a constant?

include/openssl/cipher.h
```diff
@@ -643,6 +643,12 @@ BSSL_NAMESPACE_END
 
 #endif
 
+/*
+ * The maximum permitted number of cipher blocks per data unit in XTS mode.
+ * Reference IEEE Std 1619-2018.
+ */
+#define XTS_MAX_BLOCKS_PER_DATA_UNIT            (1<<20)
```
torben-hansen, line 650: Can we move to an internal header file?

decrepit/xts/xts.c
nebeid: I thought to keep the same values used in the rest of the code below.

It seems to me that both the C and assembly implementations can support a 128-bit key size if we define an EVP_CIPHER for it, and it's more consistent with the rest of the code to not put constants than around the block size and input size not being less than one block (16 bytes).

include/openssl/cipher.h
nebeid: I had first put it in an internal header file, then decided to move it here when I added the errors below, especially for the user to know the maximum allowed input size.

decrepit/xts/xts.c
nebeid: From IEEE 1619-2007:
> 4.3.1 Data unit: Within IEEE Std 1619, 128 or more bits of data within a key scope. 
> Sec 5.1: The data unit size shall be at least 128 bits.

crypto/fipsmodule/aes/internal.h
```diff
@@ -47,6 +47,7 @@ OPENSSL_INLINE int vpaes_capable(void) {
 
 #elif defined(OPENSSL_ARM) || defined(OPENSSL_AARCH64)
 #define HWAES
+#define HWAES_XTS
```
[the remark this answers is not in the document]
nebeid: Discussed offline: This is similar to `HWAES_ECB` above which exists only for x86, used to say that even when hwaes_capable() returns true, the platform may or may not have this extra implementation. In the case of XTS, this changes adds an assembly implementation only to Arm (not tested) and AArch64, hence the addition of this macro under this platform.

crypto/fipsmodule/aes/mode_wrappers.c
```diff
@@ -125,3 +125,15 @@ void AES_cfb128_encrypt(const uint8_t *in, uint8_t *out, size_t length,
   CRYPTO_cfb128_encrypt(in, out, length, key, ivec, &num_u, enc, AES_encrypt);
   *num = (int)num_u;
 }
+
+#if defined(HWAES_XTS)
```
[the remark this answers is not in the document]
nebeid: Discussed offline. I kept the macro here, but used a function in xts.c, `hwaes_xts_available()` defined in aes/internal.h.

decrepit/xts/xts.c
```diff
@@ -193,10 +215,32 @@ static int aes_xts_cipher(EVP_CIPHER_CTX *ctx, uint8_t *out,
...
+  // Impose a limit of 2^20 blocks per data unit as specified by
+  // IEEE Std 1619-2018.  The earlier and obsolete IEEE Std 1619-2007
+  // indicated that this was a SHOULD NOT rather than a MUST NOT.
+  // NIST SP 800-38E mandates the same limit.
+  if (len > XTS_MAX_BLOCKS_PER_DATA_UNIT * AES_BLOCK_SIZE) {
+    OPENSSL_PUT_ERROR(CIPHER, CIPHER_R_XTS_DATA_UNIT_IS_TOO_LARGE);
     return 0;
   }
+
+#if defined(HWAES_XTS)
```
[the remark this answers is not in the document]
nebeid: Replaced with a function here. Thanks. See comments above on aes/internal.h and mode_wrappers.c

include/openssl/cipher.h
torben-hansen: Sure, but the more defines we expose publicly the harder agility gets in the future. I.e. simply renaming it is breaking.

Leave it to you, but I lean towards not exposing it.

torben-hansen (dismissed):

decrepit/xts/xts_test.cc
```diff
@@ -206,6 +208,245 @@ static const XTSTestCase kXTSTestCases[] = {
...
+        "fffefdfcfbfaf9f8f7f6f5f4f3f2f1f0efeeedecebeae9e8e7e6e5e4e3e2e1e0"
+        "bfbebdbcbbbab9b8b7b6b5b4b3b2b1b0afaeadacabaaa9a8a7a6a5a4a3a2a1a0",
+        "9a785634120000000000000000000000",
+        "000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e",
+        "581ea1fee5516ad432ddebe75fd27c6fc30ca8f2ed57307edc87e544867ac8",
+    },
+    // https://github.com/BrianGladman/modes/blob/master/testvals/xts.12#L2611
+    // VEC 301, len = 32 bytes = 2 blocks (decryption vector)
+    // This vector did not pass, the output below is the same for
+    // the C implementation, |CRYPTO_xts128_encrypt|, and the AArch64 assembly.
```
dkostic, line 250: what does this mean?

decrepit/xts/xts_test.cc
```diff
@@ -206,6 +208,245 @@ static const XTSTestCase kXTSTestCases[] = {
         "dfce93b6f82ae22423274e58a0821cc5502e2d0ab4585e94de6975be5e0b4efce51cd3"
         "e70c25a1fbbbd609d273ad5b0d59631c531f6a0a57b9",
     },
+    // https://github.com/BrianGladman/modes/blob/master/testvals/xts.6#L313
```
dkostic, line 211: just curious, how did you choose the specific test vectors from that github repo to include here?

nebeid: I went through the many test vector files under `testvals` choosing the ones that had 256-bit key with a plaintext length that I was looking for.

dkostic: thanks!

decrepit/xts/xts_test.cc
nebeid: The ones that were added by the original author were covering different plaintext lengths but for key size 128 which we don't claim to support in the API (though I believe the C and Asm code do support it).

dkostic (dismissed):

include/openssl/cipher.h
nebeid: Good point. I think I opted to expose it as is done in OpenSSL in [include/crypto/modes.h](https://github.com/openssl/openssl/blob/8020d79b4033400d0ef659a361c05b6902944042/include/crypto/modes.h#L140-L144).

torben-hansen: I don't think header files in `include/crypto` are intended for application usage. The content is nominally opaque.

nebeid: Moved it inside in https://github.com/awslabs/aws-lc/pull/400/commits/f57cfb2d162155a584374a89b7f2bada09f4d3b2. Thanks.

dkostic (dismissed):

dkostic (approved):

dkostic (approved):

torben-hansen (approved):

PR #404, opened 2022-03-03
Enable the FIPS build for iOS (partially tested)

dkostic:
### Issues:
CryptoAlg-1018

### Description of changes: 
Enable the FIPS build for iOS. The library was built by:
```
cmake -DBUILD_SHARED_LIBS=1 -DCMAKE_OSX_ARCHITECTURES=arm64 -DCMAKE_OSX_SYSROOT=iphoneos -DFIPS=1
```
@samuel40791765 has tested the code locally on his iPhone (note that his PR #405 is required for the test to work). Samuel doesn't have the full test app developed yet so we can't execute the unit tests. However, his app is able to link with `libcrypto.dylib` and call functions from it. Successfully executing a function from `libcrypto.dylib` means that the library is loaded and the FIPS power-on self-tests are executed. The fact that the library doesn't abort means that all the self-tests pass and also that the integrity check passes.

### Call-outs:
See the note above about testing.

### Testing:
See the note above about testing.


samuel40791765:
I tried building this with https://github.com/awslabs/aws-lc/pull/405 and tested the build with a few hash functions on an iOS app built on an actual iphone. Outputs are correct with the experiments I've done so far.

dkostic:
> I tried building this with #405 and tested the build with a few hash functions on an iOS app built on an actual iphone. Outputs are correct with the experiments I've done so far.

Thank you Samuel! I'll change the PR title to remove "not tested" and add a note about testing in the description.

util/fipstools/inject_hash/inject_hash.go
```diff
@@ -357,7 +357,7 @@ func do(outPath, oInput string, arInput string, useSHA256 bool, macOS bool) erro
 
 	var moduleText, moduleROData []byte
 	var err error
-	if macOS == true {
+	if appleOS == true {
 		moduleText, moduleROData, err = doMacOS(objectBytes)
```
samuel40791765, line 361: minor nit: should we change to `doAppleOS`

samuel40791765 (dismissed):

util/fipstools/inject_hash/inject_hash.go
dkostic: yup, changed

samuel40791765 (approved):

crypto/rsa_extra/rsa_test.cc
```diff
@@ -653,7 +654,9 @@ TEST(RSATest, GenerateSmallKey) {
   EXPECT_EQ(ERR_LIB_RSA, ERR_GET_LIB(err));
   EXPECT_EQ(RSA_R_KEY_SIZE_TOO_SMALL, ERR_GET_REASON(err));
 }
-#else
+
+#elif !defined(OPENSSL_IOS) // ASSERT_DEATH is not defined for iOS.
```
DominicDams, line 658: We might want to consider using `ASSERT_DEATH_IF_SUPPORTED` to simplify this, but I can see not wanting to use that since it will just lead to some wasted cycles running tests on IOS. I only suggest it since it seems to be a way of dealing with these compatibility issues in the future (though I'm not sure there are any real issues like this we will hit). Certainly not a blocker just food for thought.

https://github.com/google/googletest/blob/main/googletest/include/gtest/gtest-death-test.h#L327

DominicDams (dismissed):

crypto/rsa_extra/rsa_test.cc
dkostic: good find! I changed the asserts to `ASSERT_DEATH_IF_SUPPORTED`

samuel40791765 (approved):

DominicDams (approved):

PR #413, opened 2022-03-08
Add new verification error for mismatched signatures

WillChilds-Klein:
Overview
========
OpenSSL and AWS-LC differ in that OpenSSL [**does not** set][1] an error
code when it fails to verify an otherwise valid† ECDSA signature while
[BoringSSL][4] and [AWS-LC][2] **do**. AWS-LC sets the
error code `BAD_SIGNATURE` in this case, but also in the case of
malformed signatures. This conflation of concerns makes it impossible
for callers to distinguish between malfored and invalid signatures
(though there are reasonable arguments for why this conflation is
desirable).

This commit differentiates between the two cases by introducing a new
error code (`MISMATCHED_SIGNATURE`) and setting it in the case of
well-formed signatuer mismatch. We still set `BAD_SIGNATURE` for
malformed signatures.

AWS-LC's use of `BAD_SIGNATURE` error in _DSA_ verification feailure is
consistent with [AWS-LC's treatment of _RSA_ signature verification
failures][3], so this commit retains that consistency by applying the
same changes to both RSA and ECDSA.

AmazonCorrettoCryptoProvider (ACCP)
===================================
When we (ACCP) see an error after attempting (non-RSA) verification, we
[throw an exception][8]. Unfortunately, because AWS-LC currently
conflates _malformed_ and _inauthentic_ signatures under the same
`BAD_SIGNATURE` error code, we can not differentiate the cases, and thus
will never return `false` on ECDSA verification -- we'll either return
`true` if the signature is valid and authentic, else throw. The
[javadocs][9] for `Signature.verify` seem to indicate that we should
indeed return boolean `false` on verification failure. In order to do
this, we'll need to make changes to AWS-LC to either set no error code
upon verification failure or to set a distinct code which we can key off
of.

Code Generation
===============
Error entries were added using this command:

```
$ go run ./util/make_errors.go ecdsa rsa
```
Auto-generated error code files were updated by invoking this script from
the root of the project:

```
$ python3 ./util/generate_build_files.py
```

Appendix
========
Relevant Links:

[1]: https://github.com/openssl/openssl/blob/5cbd2ea3f94aa8adec9b4486ac757d4d688e3f8c/crypto/ec/ecdsa_ossl.c#L458
[2]: https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/ecdsa/ecdsa.c#L194-L197
[3]: https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/rsa/rsa.c#L620-L621
[4]: https://boringssl.googlesource.com/boringssl/+/refs/heads/master/crypto/fipsmodule/ecdsa/ecdsa.c#194
[5]: https://github.com/corretto/amazon-corretto-crypto-provider/commit/dcbf8a7dc6fb5064b979739e2ffc590de61c39a2#diff-7fe3b81789f62d5bbf2ff0d8a6a8db06393630449ff914f9127f788084b06e49R481-R484
[6]: https://docs.oracle.com/en/java/javase/17/security/java-cryptography-architecture-jca-reference-guide.html
[7]: https://docs.oracle.com/javase/7/docs/api/java/security/SignatureException.html
[8]: https://github.com/corretto/amazon-corretto-crypto-provider/blob/develop/csrc/sign.cpp#L483
[9]: https://docs.oracle.com/javase/7/docs/api/java/security/Signature.html#verify(byte[])

This resolves SIM CryptoAlg-1027.

† in other words, the signature is well-formed, but it is inauthentic
per the specified public key

# Testing:

Tests pass locally on MacOS:

```
$ ninja run_tests
...
All tests passed!
```

Automated checks currently running for other platforms/test configurations.


SalusaSecondus (dismissed):
Seems reasonable to me.

WillChilds-Klein:
some topics i was hoping to discuss in review:

- name of new error code
- safety/backwards compatibility concerns around this change
  + this may break callers who expect and rely on `BAD_SIGNATURE` in the error case
- alternatives to this changes approach, including:
  + using `BAD_SIGNATURE` for the well-formed-but-mismatched case and a new error code (perhaps `MALFORMED_SIGNATURE`) for the malformed case
  + introduce new, granular error codes for each failure mode
  + modify ACCP's behavior to mirror the conflation of these cases, and either return `false` in all cases or throw in all cases
    - this is very much discouraged because it does not conform with expected JCA behavior
  + implement some sort of fallback in ACCP that will try to do the verification in java if the call to AWS-LC fails
    - this is undesirable because it's slow, vulnerable to DoS, and will introduce FIPS approval overhead for the library

include/openssl/ecdsa.h
```diff
@@ -232,5 +232,6 @@ BSSL_NAMESPACE_END
 #define ECDSA_R_NOT_IMPLEMENTED 103
 #define ECDSA_R_RANDOM_NUMBER_GENERATION_FAILED 104
 #define ECDSA_R_ENCODE_ERROR 105
+#define ECDSA_R_MISMATCHED_SIGNATURE 106
```
torben-hansen, line 235: Maybe we should reserve some space for openssl/boringssl defined values here. Offset of 100 reasonable?

include/openssl/rsa.h
```diff
@@ -859,5 +859,6 @@ BSSL_NAMESPACE_END
 #define RSA_R_PUBLIC_KEY_VALIDATION_FAILED 146
 #define RSA_R_D_OUT_OF_RANGE 147
 #define RSA_R_BLOCK_TYPE_IS_NOT_02 148
+#define RSA_R_MISMATCHED_SIGNATURE 149
```
torben-hansen, line 862: Maybe we should reserve some space for openssl/boringssl defined values here. Offset of 100 reasonable?

torben-hansen (commented):
I dunno why an application would want to distinguish between these two cases. But okay.

Feel free to add more tests - if you want to play around with it and can come up with some. But not a blocker.

torben-hansen (commented):
Github fail - didn't push this.

include/openssl/rsa.h
WillChilds-Klein: yep, sounds good.

> Feel free to add more tests - if you want to play around with it and can come up with some. But not a blocker.

i tried to identify new cases relevant to this change, but it looks like the existing test cases cover the "well-formed, but mismatched" cases pretty well.

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -615,13 +615,18 @@ int RSA_verify(int hash_nid, const uint8_t *digest, size_t digest_len,
...
-  if (len != signed_msg_len || OPENSSL_memcmp(buf, signed_msg, len) != 0) {
+  // Check that no other information follows the hash value (FIPS 186-4 Section 5.5)
+  if (len != signed_msg_len) {
     OPENSSL_PUT_ERROR(RSA, RSA_R_BAD_SIGNATURE);
     goto out;
   }
 
+  // Check that the computed hash matches the expected hash
+  if (OPENSSL_memcmp(buf, signed_msg, len) != 0) {
+    OPENSSL_PUT_ERROR(RSA, RSA_R_MISMATCHED_SIGNATURE);
```
torben-hansen, line 626: Please document new error distinction in the function description `include/openssl/rsa.h`

include/openssl/ecdsa.h
```diff
@@ -141,6 +141,10 @@ OPENSSL_EXPORT ECDSA_SIG *ECDSA_do_sign(const uint8_t *digest,
 // of |digest|. It returns one on success or zero if the signature is invalid
 // or on error.
 //
+// Error ECDSA_R_BAD_SIGNATURE is set if |sig| has been corrupted or is
```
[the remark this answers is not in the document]
WillChilds-Klein: updated.

include/openssl/rsa.h
```diff
@@ -357,6 +357,10 @@ OPENSSL_EXPORT int RSA_sign_raw(RSA *rsa, size_t *out_len, uint8_t *out,
 //
 // It returns one if the signature is valid and zero otherwise.
 //
+// Error RSA_R_BAD_SIGNATURE is set if |sig| has been corrupted or is otherwise
```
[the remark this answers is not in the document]
WillChilds-Klein: updated.

torben-hansen (approved):

torben-hansen (approved):

SalusaSecondus (approved):

dkostic (approved):

PR #414, opened 2022-03-08
Improve rdrand retries

torben-hansen:
### Description of changes: 

Make condition less strict, to make sure we also catch some weird situation where `tries` jumps past `RDRAND_MAX_RETRIES - 1`.

### DIFF
--- a/crypto/fipsmodule/rand/rand.c
+++ b/crypto/fipsmodule/rand/rand.c
@@ -167,7 +167,7 @@ OPENSSL_STATIC_ASSERT(RDRAND_MAX_RETRIES > 0, rdrand_max_retries_must_be_positiv
       if ((rdrand_func) == 1) {                                   \
         break;                                                    \
       }                                                           \
-      else if (tries == RDRAND_MAX_RETRIES - 1) {                 \
+      else if (tries >= RDRAND_MAX_RETRIES - 1) {                 \
         return fail_ret_value;                                    \
       }                                                           \
     }


crypto/fipsmodule/rand/rand.c
```diff
@@ -167,7 +167,7 @@ OPENSSL_STATIC_ASSERT(RDRAND_MAX_RETRIES > 0, rdrand_max_retries_must_be_positiv
       if ((rdrand_func) == 1) {                                   \
         break;                                                    \
       }                                                           \
-      else if (tries == RDRAND_MAX_RETRIES - 1) {                 \
+      else if (tries >= RDRAND_MAX_RETRIES - 1) {                 \
         return fail_ret_value;                                    \
       }                                                           \
     }
```
nebeid, line 173: May I suggest
```
#define CALL_RDRAND_WITH_RETRY(rdrand_func, fail_ret_value)       \
  do {                                                              \
    for (size_t tries = 0; tries < RDRAND_MAX_RETRIES; tries++) { \
      if ((rdrand_func) == 1) {                                   \
        break;                                                    \
      }                                                           \
    }                                                             \
    if (tries == RDRAND_MAX_RETRIES ) {                            \
       return fail_ret_value;                                    \
    }                                                           \
  } while(0)
```
The do-while is to be able to put a semi-column at the end when calling the macro and to keep statements in one context. We used it in `service_indicator.h`.

nebeid: (I forgot to declare `tries` outside the for-loop)

nebeid (approved):

dkostic (approved):

PR #415, opened 2022-03-09
Update P-256 AArch64 code to use PAuth as defined by BoringSSL

nebeid:
When we imported the OpenSSL P-256 AArch64 code, it contained instructions for pointer authentication. This commit https://github.com/openssl/openssl/commit/19e277dd19f2897f6a7b7eb236abe46655e575bf updated them to use the same macros defined in BoringSSL.
This PR applies the changes to the code in AWS-LC.

**Notes:**
- `scatter` and `gather` functions in OpenSSL are not used in AWS-LC.
- `ecp_niszt256_add` funtion was removed from assembly in https://github.com/awslabs/aws-lc/commit/d60b60e8e5f7eb8e5ca76190246f7fe17d95f9e1.
  It is not used in the critical path of computation and it's defined in p256-nistz.c in terms of the assembly function `ecp_nistz256_point_add`.
- The same is done to the code we added in `p256_beeu-armv8-asm.pl`.


dkostic (approved):

andrewhop (approved):
What is the recommended way to use this option? 
I tried passing it in on the command line and it failed:
```
cmake -E env CXXFLAGS="-mbranch-protection=standard" cmake -GNinja -DCMAKE_BUILD_TYPE=Release ../ && ninja
clang: error: argument unused during compilation: '-mbranch-protection=standard' [-Werror,-Wunused-command-line-argument]
[404/455] Linking CXX executable crypto/urandom_test
```

I was able to get it to work by modifying the crypto/fipsmodule/CMakeLists.txt with:
```
target_compile_options(fipsmodule PRIVATE -mbranch-protection=standard)
```

PR #422, opened 2022-03-14
Improve service indicator initialization and add tests for under/overflow

andrewhop:
### Issues:
Addresses CryptoAlg-907

### Description of changes: 
Add lazy initialization logic to all service indicator functions. This is *not* an issue for the current Linux FIPS branch. This issue only happens on Windows FIPS builds. The Windows issue is nothing was calling `FIPS_service_indicator_update_state` which created the state before calling `FIPS_service_indicator_lock_state`. When lock was called with a null state nothing happens and AWS-LC continues and eventually hits a call to update state which creates the state, then something calls unlock and underflows. My hypothesis is MSVC changes the order of self tests in the `BORINGSSL_bcm_power_on_self_test` to hit this unlock issue while Clang/GCC does not.

### Call-outs:
The CI does not test FIPS on Windows yet due to other issues. I have tested this part on a Windows machine manually and it addressed the issue. On Linux there is no behavior change.

### DIFF
--- a/crypto/fipsmodule/service_indicator/service_indicator.c
+++ b/crypto/fipsmodule/service_indicator/service_indicator.c
@@ -19,25 +19,29 @@ int is_fips_build(void) {
 
 #if defined(AWSLC_FIPS)
 
-// Should only be called once per thread. Only called when initializing the
-// state in |FIPS_service_indicator_before_call| (external call) or
-// in |FIPS_service_indicator_update_state| (internal call within every approved service).
-static int FIPS_service_indicator_init_state(void) {
-  struct fips_service_indicator_state *indicator;
-  indicator = OPENSSL_malloc(sizeof(struct fips_service_indicator_state));
-  if (indicator == NULL || !CRYPTO_set_thread_local(
-      AWSLC_THREAD_LOCAL_FIPS_SERVICE_INDICATOR_STATE, indicator, OPENSSL_free)) {
-    OPENSSL_PUT_ERROR(CRYPTO, ERR_R_MALLOC_FAILURE);
-    return 0;
+// FIPS 140-3 requires that the module should provide the service indicator
+// for approved services irrespective of whether the user queries it or not.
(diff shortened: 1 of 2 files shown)


crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -19,39 +19,36 @@ int is_fips_build(void) {
...
+    indicator = OPENSSL_malloc(sizeof(struct fips_service_indicator_state));
+    if (indicator == NULL) {
+      OPENSSL_PUT_ERROR(CRYPTO, ERR_R_MALLOC_FAILURE);
+      return NULL;
+    }
+    indicator->lock_state = STATE_UNLOCKED;
+    indicator->counter = 0;
+    if (!CRYPTO_set_thread_local(AWSLC_THREAD_LOCAL_FIPS_SERVICE_INDICATOR_STATE, indicator, OPENSSL_free)) {
+      OPENSSL_PUT_ERROR(CRYPTO, ERR_R_INTERNAL_ERROR);
+      return NULL;
```
torben-hansen, line 37: Memory leak here (`indicator`)?

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -664,7 +665,11 @@ struct AEADTestVector {
         kAESKey, 16, kAESCCMCiphertext, 64, AWSLC_APPROVED, false },
 };
 
-class AEAD_ServiceIndicatorTest : public testing::TestWithParam<AEADTestVector> {};
+class AEAD_ServiceIndicatorTest : public testing::TestWithParam<AEADTestVector> {
```
torben-hansen, line 668: Possible to move this destructor to a base class and derive all these other algorithm-specific service indicator classes from it?

torben-hansen (approved):

samuel40791765 (approved):

PR #429, opened 2022-03-16
Enable XTS optimisation for x86_64

nebeid:
### Issues:
Resolves CryptoAlg-929

### Description of changes: 
From the commit history of `aesni-x86_64.pl`, we gather that the xts assembly functions were disabled to save code space. They are re-enabled in this PR.

~This PR is to be rebased on PR #400.~
It was force-pushed on top of PR #445. PR #400 is now in main.

Init and Encrypt:
|      Size   | Before (MB/s)(1) | After (MB/s) | Improvement |
-------------:|-----------------:|-------------:|:------------:
|    16 bytes |       127        |      129     |    1.02x    |
|   256 bytes |       714        |     1473     |    2.06x    |
|  1350 bytes |       915        |     2816     |    3.08x    |
|  8192 bytes |      1006        |     3626     |    3.60x    |
| 16384 bytes |      1010        |     3748     |    3.71x    |

Init and Decrypt:
|      Size   | Before (MB/s)(1) | After (MB/s) | Improvement |
-------------:|-----------------:|-------------:|:------------:
|    16 bytes |       117        |      119     |    1.02x    |
|   256 bytes |       700        |     1373     |    1.96x    |
|  1350 bytes |       915        |     2663     |    2.91x    |
|  8192 bytes |      1004        |     3610     |    3.60x    |
| 16384 bytes |      1007        |     3735     |    3.71x    |


crypto/fipsmodule/aes/mode_wrappers.c
```diff
@@ -142,4 +142,11 @@ int aes_hw_xts_cipher(const uint8_t *in, uint8_t *out, size_t length,
   }
   return 1;
 }
-#endif
+
+#if defined(OPENSSL_X86_64)
+int CRYPTO_is_xts_AESNI_capable(void) {
```
torben-hansen, line 147: What's the motivation for defining this function here? Do we even need it now that it only depends on `CRYPTO_is_AESNI_capable()`?

crypto/fipsmodule/aes/internal.h
```diff
@@ -160,6 +170,9 @@ OPENSSL_INLINE int aes_hw_xts_cipher(const uint8_t *in, uint8_t *out, size_t len
                                       const uint8_t iv[16], int enc) {
   abort();
 }
+OPENSSL_INLINE int CRYPTO_is_xts_AESNI_capable(void) {
```
torben-hansen, line 173: FWIW (not sure we need this function), should prob return `0`?

crypto/fipsmodule/aes/mode_wrappers.c
nebeid: What I recall is that because `CRYPTO_is_AESNI_capable()` is inlined, I couldn't make it directly external [commits 1, 2], I got the same linker error that `OPENSSL_ia32cap_P` is not found in the build. I believe it can be removed once the code moves out of `libdecrepit` into `libcrypto`.
[1] https://github.com/awslabs/aws-lc/pull/429/commits/f4b9931109452b696660402445efc4ceed16b6f8
[2] https://github.com/awslabs/aws-lc/pull/429/commits/b87cf55d2eb860aa347c6549dced83c8853ab69b

crypto/fipsmodule/aes/internal.h
nebeid: I matched the behaviour of the other functions, where this code should not be executed if the defines are set properly for detecting `HWAES_XTS`.

torben-hansen (dismissed):

crypto/fipsmodule/aes/internal.h
torben-hansen: I think the `*_capable` functions returns 0 normally, can be changed in https://github.com/awslabs/aws-lc/pull/469 if need be.

torben-hansen (approved):

crypto/fipsmodule/aes/internal.h
nebeid: Thank you. I changed it to return 0 and removed it in #469.

dkostic (approved):

PR #430, opened 2022-03-17
Add a minimal analytics build that collects size, build, and test time of several build dimensions

andrewhop:
### Description of changes: 
We're interested in several metrics about AWS-LC that might change over time. This change adds the start of a system that could track several of them in cloudwatch:
* Size of source code and top level folders
* For the combination of small, no asm, static/shared:
    * Build time for various dimensions
    * Test time for various dimensions
    * Size of libcrypto and libssl 
* FIPS is treated special since only one dimension is actually FIPS certified
* It's 9 builds in total and CodeBuild takes about 30 minutes to finish

### Call-outs:
I like the metric names because it is very clear what is happening and line up with CMake. One drawback is the names are a little long and make it clunky in the console. For example by default the label in the graph is just the **value** of the dimension, so things look like `ON ON OFF x86_64` and you have to line up. The dimension could be updated to `Assembly=NO_ASM` and `Assembly=ASM` so the console (by default) would look like `ASM SHARED non-FIPS x64_64`. Naming things is still hard and I'm interested in other opinions on the user experience.

### Testing:
I've deployed this to my account and browsed the metrics in CloudWatch looks reasonable. Happy to give people a tour in my account before deploying it to the team account.


tests/ci/run_analytics.sh
```diff
@@ -0,0 +1,95 @@
...
+    linking="Shared"
+  else
+    linking="Static"
+  fi
+  if [[ "$fips" == "ON" ]]; then
+    fips_mode="FIPS"
+  else
+    fips_mode="NotFIPS"
+  fi
+  common_dimensions="Optimization=Release,BuildSize=${build_size},Assembly=${assembly},CPU=${PLATFORM},Linking=${linking},FIPS=${fips_mode}"
```
darylmartin100, line 57: I'm not sure if it makes sense, but what about capturing the size of all .obj files?  If we are tracking the size of the build output and want to figure out why it's gotten bigger or smaller then we could slice and dice on the individual. .obj files.

tests/ci/run_analytics.sh
```diff
@@ -0,0 +1,95 @@
...
+
+  build_start=$(date +%s)
+  run_build -DCMAKE_BUILD_TYPE=Release \
+      -DBUILD_SHARED_LIBS="$shared_library" \
+      -DOPENSSL_SMALL="$small" \
+      -DOPENSSL_NO_ASM="$no_assembly" \
+      -DFIPS="$fips"
+  build_end=$(date +%s)
+  build_time=$((build_end-build_start))
+  put_metric --metric-name BuildTime --value "$build_time" --unit Seconds --dimensions "$common_dimensions"
```
darylmartin100, line 67: We should put in the git commit used and branch this was built off of for complete traceability.  We don't want our long living branches (like FIPS) to pollute our data as it would have a very recent build time but based off of very old code.

tests/ci/cdk/cdk/codebuild/github_ci_analytics_omnibus.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Doc for batch https://docs.aws.amazon.com/codebuild/latest/userguide/batch-build-buildspec.html#build-spec.batch.build-list
+batch:
+  build-list:
+
+    - identifier: ubuntu2004_clang10_x86_64_analytics
+      buildspec: ./tests/ci/codebuild/common/run_analytics.yml
+      env:
+        type: LINUX_CONTAINER
+        privileged-mode: true
+        compute-type: BUILD_GENERAL1_LARGE
```
bryce-shang, line 15: Nit: metrics publish may just need `BUILD_GENERAL1_MEDIUM`?

tests/ci/cdk/cdk/codebuild/github_ci_analytics_omnibus.yaml
```diff
@@ -0,0 +1,24 @@
...
+        image: X86_ECR_REPO_PLACEHOLDER:ubuntu-20.04_clang-10x_latest
+
+    - identifier: ubuntu2004_clang10_arm_analytics
+      buildspec: ./tests/ci/codebuild/common/run_analytics.yml
+      env:
+        type: ARM_CONTAINER
+        privileged-mode: true
+        compute-type: BUILD_GENERAL1_LARGE
```
bryce-shang, line 23: Nit: arm `BUILD_GENERAL1_SMALL` can work for metrics publish?
https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-compute-types.html

tests/ci/run_analytics.sh
```diff
@@ -0,0 +1,111 @@
...
+SOURCE_CODE_SIZE=$(size "$SRC_ROOT")
+put_metric --metric-name FolderSize --value "$SOURCE_CODE_SIZE" --unit Bytes --dimensions "Folder=root,${common_dimensions}"
+
+for FOLDER_PATH in "$SRC_ROOT"/*/ ; do
+    FOLDER_SIZE=$(size "$FOLDER_PATH")
+    FOLDER=$(basename "$FOLDER_PATH")
+    put_metric --metric-name FolderSize --value "$FOLDER_SIZE" --unit Bytes --dimensions "Folder=${FOLDER},${common_dimensions}"
+done
+
+function run_build_and_collect_metrics {
```
bryce-shang, line 39: Nit: maybe we can move these metrics collection logics to `run_build` of `tests/ci/common_posix_setup.sh`.

tests/ci/cdk/cdk/codebuild/github_ci_analytics_omnibus.yaml
```diff
@@ -0,0 +1,24 @@
...
+batch:
+  build-list:
+
+    - identifier: ubuntu2004_clang10_x86_64_analytics
+      buildspec: ./tests/ci/codebuild/common/run_analytics.yml
+      env:
+        type: LINUX_CONTAINER
+        privileged-mode: true
+        compute-type: BUILD_GENERAL1_LARGE
+        image: X86_ECR_REPO_PLACEHOLDER:ubuntu-20.04_clang-10x_latest
```
bryce-shang, line 16: Nit: we are only interested in the metrics of build with `clang-10`?

bryce-shang (approved):

tests/ci/cdk/cdk/aws_lc_analytics_stack.py
```diff
@@ -0,0 +1,60 @@
...
+
+from aws_cdk import core, aws_codebuild as codebuild, aws_iam as iam, aws_ec2 as ec2, aws_efs as efs
+
+from util.ecr_util import ecr_arn
+from util.iam_policies import code_build_publish_metrics_in_json
+from util.metadata import AWS_ACCOUNT, AWS_REGION, GITHUB_REPO_OWNER, GITHUB_REPO_NAME
+from util.yml_loader import YmlLoader
+
+
+class AwsLcGitHubAnalkyticsStack(core.Stack):
```
dkostic, line 12:
```suggestion
class AwsLcGitHubAnalyticsStack(core.Stack):
```

tests/ci/run_analytics.sh
andrewhop: I thought about this but to maintain a more reasonable metrics history these builds are only triggered on push, not pr open/update. This is to prevent collecting weird data with PR's that add/remove a lot of code and then don't get merged in. Running this only on push makes sure it's what our branches actually looked like.

tests/ci/cdk/cdk/codebuild/github_ci_analytics_omnibus.yaml
andrewhop: For right now yes, do you think it makes sense to add compiler as a dimension to the metrics for the future? CloudWatch has a [limit of 10 dimensions](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_limits.html) per metric. For the library size metrics we're at 8 (library, release, assembly, small, static/shared, fips, CPU, branch). I can't think of that many more dimensions for size we might be interested in the future. It is making me think that instead of the metric LibrarySize and a dimension of Library (name) the metric could be CryptoSize and SSLSize. It doesn't really make sense to want to search/graph libcrypto and libssl size together with the name as a dimension.

bryce-shang (approved):

tests/ci/run_analytics.sh
```diff
@@ -0,0 +1,114 @@
...
+  build_time=$((build_end-build_start))
+  put_metric --metric-name BuildTime --value "$build_time" --unit Seconds --dimensions "$size_common_dimensions"
+
+  test_start=$(date +%s)
+  run_cmake_custom_target 'run_tests'
+  test_end=$(date +%s)
+  test_time=$((test_end-test_start))
+  put_metric --metric-name TestTime --value "$test_time" --unit Seconds --dimensions "$size_common_dimensions"
+
+  libcrypto_size=$(du --bytes --apparent-size "${BUILD_ROOT}/crypto/libcrypto.${lib_extension}" | cut -f1)
```
DominicDams, line 83: Nit: It might be worth having another function like `size` since you use this frequently

DominicDams (approved):

PR #435, opened 2022-03-23
Simplified the Benchmarking framework

DominicDams:
The benchmarking framework was complex and that led to difficulty
debugging issues, and eventually led to the discontinuation of it's
usage. This change simplifies the framework to provide easier debugging
as well as fixing some issues

### Description of changes: 
This updates the benchmarking framework in code to run directly on codebuild, stop looking for regressions compared to OpenSSL/BoringSSL and use a more accurate method for determining performance degradations.

### Call-outs:
This also enables the ci, but will not make it required on pull requests, so we can evaluate whether or not it fails too frequently before we start requiring it

### Testing:
This was manually tested in my developer account. I never saw a difference of above 5% from random chance with my changes so I think it's unlikely to cause spurious failures.


tests/ci/build_run_benchmarks.sh
```diff
@@ -7,49 +7,46 @@ set -x
...
+
+cd ..
+
 # run this from the bm_framework root directory!
-OPENSSL_ROOT=$(pwd)/openssl
-BORINGSSL_ROOT=$(pwd)/boringssl
 AWSLC_PR_ROOT=$(pwd)/"${PR_FOLDER_NAME}"
 AWSLC_PROD_ROOT=$(pwd)/aws-lc-prod
 
+NUM_CPU_THREADS=$(nproc)
```
andrewhop, line 21: NP: Can this script depend on common_posix_setup? It has a smarter NUM_CPU_THREADS variable, I know not all systems have nproc by default, but we probably wont run the benchmark on those systems anyway.

tests/ci/build_run_benchmarks.sh
```diff
@@ -7,49 +7,46 @@ set -x
...
 
 #build FIPS compliant version of AWSLC prod
-mkdir aws-lc-prod/fips_build
-cmake -Baws-lc-prod/fips_build -Haws-lc-prod -GNinja -DFIPS=1 -DCMAKE_BUILD_TYPE=Release -DAWSLC_INSTALL_DIR="${AWSLC_PROD_ROOT}"
+mkdir -p aws-lc-prod/fips_build
+mkdir -p aws-lc-prod/fips_install
+cmake -Baws-lc-prod/fips_build -Haws-lc-prod -GNinja -DFIPS=1 -DCMAKE_BUILD_TYPE=Release -DAWSLC_INSTALL_DIR="${AWSLC_PROD_ROOT}/fips_install" -DBUILD_SHARED_LIBS=TRUE -DCMAKE_INSTALL_PREFIX="${AWSLC_PROD_ROOT}/fips_install"
 ninja -C aws-lc-prod/fips_build
 
 # avoid cpus 0-3 since there are a lot of other things running on them
```
andrewhop, line 52: Is this comment still true for codebuild?

tests/ci/build_run_benchmarks.sh
```diff
@@ -7,49 +7,46 @@ set -x
...
 # build AWSLC pr
-mkdir "${PR_FOLDER_NAME}"/build
+mkdir -p "${PR_FOLDER_NAME}"/build
+mkdir -p "${PR_FOLDER_NAME}"/install
 cmake -B"${PR_FOLDER_NAME}"/build -H"${PR_FOLDER_NAME}" -GNinja -DCMAKE_BUILD_TYPE=Release \
-  -DAWSLC_INSTALL_DIR="${AWSLC_PR_ROOT}" \
-  -DBORINGSSL_INSTALL_DIR="${BORINGSSL_ROOT}" \
-  -DOPENSSL_INSTALL_DIR="${OPENSSL_ROOT}"
+  -DAWSLC_INSTALL_DIR="${AWSLC_PR_ROOT}"/install \
+  -DCMAKE_INSTALL_PREFIX="${AWSLC_PR_ROOT}"/install
```
andrewhop, line 31: To speed up all these build you could turn off testing and libssl.

tests/ci/benchmark_framework/compare_results.py
```diff
@@ -57,16 +57,18 @@ def main():
...
     df2_numCalls = df2['numCalls.2']
+    df1_time = df1['microseconds.1']
+    df2_time = df2['microseconds.2']
 
     # put both dataframes side by side for comparison
     dfs = pd.concat([df1, df2], axis=1)
 
     # we want things that have a +15% regression
-    compared = np.where(((df2_numCalls - df1_numCalls)/df1_numCalls*100 <= -15), df1.iloc[:, 0], np.nan)
+    compared = np.where( 100 *( 1 - df2_time*df1_numCalls/df2_numCalls/df1_time) >= 15, df1.iloc[:, 0], np.nan)
```
dkostic, line 67: can you please write this with parentheses because it's not obvious to me what we are trying to compute (for example, I don't know how a/b/c computes in python :)).

tests/ci/benchmark_framework/compare_results.py
```diff
@@ -57,16 +57,18 @@ def main():
...
 
     compared_df = dfs.loc[dfs.iloc[:, 0].isin(compared)]
-    compared_df["Percentage Difference"] = ((compared_df['numCalls.1'] - compared_df['numCalls.2'])
-                                            / compared_df['numCalls.1']*100)
+    compared_df["Percentage Difference"] = ( 100 *( 1 - compared_df['microseconds.2']*compared_df['numCalls.1']
```
dkostic, line 70: same as above

tests/ci/build_run_benchmarks.sh
dkostic: yeah, I don't think we should be binding processes to cores. Regarding the comment, I'm not sure where that claim comes from, it's weird.

tests/ci/build_run_benchmarks.sh
```diff
@@ -65,77 +62,55 @@ prod_pid=$!
 taskset -c 7 ./aws-lc-prod/fips_build/tool/awslc_bm -timeout 3 -json > aws-lc-prod_fips_bm.json &
 prod_fips_pid=$!
```
dkostic, line 63: what does "prod" stand for?

tests/ci/build_run_benchmarks.sh
```diff
@@ -65,77 +62,55 @@ prod_pid=$!
...
-python3 "${PR_FOLDER_NAME}"/tests/ci/benchmark_framework/convert_json_to_csv.py bssl_bm.json
 
 python3 "${PR_FOLDER_NAME}"/tests/ci/benchmark_framework/convert_json_to_csv.py aws-lc-pr_tt_bm.json
 python3 "${PR_FOLDER_NAME}"/tests/ci/benchmark_framework/convert_json_to_csv.py aws-lc-pr_tt_fips_bm.json
 python3 "${PR_FOLDER_NAME}"/tests/ci/benchmark_framework/convert_json_to_csv.py aws-lc-prod_tt_bm.json
 python3 "${PR_FOLDER_NAME}"/tests/ci/benchmark_framework/convert_json_to_csv.py aws-lc-prod_tt_fips_bm.json
-python3 "${PR_FOLDER_NAME}"/tests/ci/benchmark_framework/convert_json_to_csv.py ossl_tt_bm.json
-python3 "${PR_FOLDER_NAME}"/tests/ci/benchmark_framework/convert_json_to_csv.py bssl_tt_bm.json
 
 # once we have csvs, we want to update the main benchmark results files with the sequential trusttoken results
```
dkostic, line 90: do we need these trust token numbers at all? I feel like we never cared about them and it's just generating noise

tests/ci/cdk/run-cdk.sh
```diff
@@ -92,7 +92,7 @@ function create_github_ci_stack() {
 #  aws codebuild update-webhook --project-name aws-lc-ci-windows-x86 --build-type BUILD_BATCH
   aws codebuild update-webhook --project-name aws-lc-ci-fuzzing --build-type BUILD_BATCH
   # TODO: re-enable 'aws-lc-ci-bm-framework' when it's ready.
```
dkostic, line 94: this can probably be removed now?

tests/ci/build_run_benchmarks.sh
DominicDams: production

tests/ci/benchmark_framework/compare_results.py
```diff
@@ -57,16 +57,18 @@ def main():
...
     df2_numCalls = df2['numCalls.2']
+    df1_time = df1['microseconds.1']
+    df2_time = df2['microseconds.2']
 
     # put both dataframes side by side for comparison
     dfs = pd.concat([df1, df2], axis=1)
 
     # we want things that have a +15% regression
-    compared = np.where(((df2_numCalls - df1_numCalls)/df1_numCalls*100 <= -15), df1.iloc[:, 0], np.nan)
+    compared = np.where(100 * (1 - (df2_time * df1_numCalls) / (df2_numCalls * df1_time)) >= 15, df1.iloc[:, 0], np.nan)
```
dkostic, line 67: How about this (and similarly for the stuff below):
```suggestion
    df1_num_calls = df1['numCalls.1']
    df2_num_calls = df2['numCalls.2']
    df1_total_time = df1['microseconds.1']
    df2_total_time = df2['microseconds.2']
    
    # Filter out the entries where the performance difference is >= 15%
    df1_time = float(df1_total_time) / df1_num_calls
    df2_time = float(df2_total_time) / df2_num_calls
    compared = np.where(1 - df2_time/df1_time >= 0.15, df1.iloc[:, 0], np.nan)
```

tests/ci/benchmark_framework/compare_results.py
```diff
@@ -28,49 +28,60 @@ def main():
...
+    # Setup data for regressed entries
+    compared_df1_numCalls = compared_df['numCalls.1']
+    compared_df2_numCalls = compared_df['numCalls.2']
+    compared_df1_time = compared_df['microseconds.1']
+    compared_df2_time = compared_df['microseconds.2']
+    compared_df1_avg_time = float(compared_df1_time) / compared_df1_numcalls
+    compared_df2_avg_time = float(compared_df2_time) / compared_df2_numcalls
+
+    # Add regression data to the table
+    compared_df["Percentage Difference"] = 100 * (1 - (df2_avg_time / df1_avg_time))
```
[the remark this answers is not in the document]
DominicDams: :( I can't believe I missed that

tests/ci/build_run_benchmarks.sh
DominicDams: I don't know that we need them but I don't think it'll hurt to have and I doubt it'll cause too much noise since they should be easy enough to ignore

tests/ci/benchmark_framework/compare_results.py
```diff
@@ -28,47 +28,60 @@ def main():
...
 
     print(df1.columns)
 
-    # 2nd column of the dataframe represents the number of calls
+    # Setup data
     df1_numCalls = df1['numCalls.1']
     df2_numCalls = df2['numCalls.2']
+    df1_time = df1['microseconds.1']
+    df2_time = df2['microseconds.2']
+    df1_avg_time = float(df1_time) / df1_numcalls
```
dkostic, line 62: Have you tried running the script before you pushed the changes? I doubt it would work :)
`df1_numcalls` should be `df1_numCalls`.
Nit: while we are at it, what do you think about renaming `df1_numCalls` to `df1_num_calls` because in the whole file we use `_` instead of camel case...

dkostic (approved):
lgtm now. Before you merge the PR, please specify in the commit message the changes you've made to simplify the tool.

tests/ci/build_run_benchmarks.sh
DominicDams: Turning off LIBSSL causes it to not build the awslc_bm tool

samuel40791765 (approved):

PR #436, opened 2022-03-23
Detect ACCP keys when parsing RSA private keys from ASN.1/DER

WillChilds-Klein:
# Summary

Prior to this change, all RSA private keys were also validated as public
keys during ASN.1 decoding. That was all well and good for CRT-enabled
private keys, as they contained the public exponent `e`, but was not
suitable for stripped private keys (i.e. keys defined only by `n` and
`d`) as required by the JCA and ACCP.

The refactor is based directly on @htorben's sketch [here][1], and does
not change AWS-LC's external API at all. We intentionally attempt to
minimize behavior and validation changes in order to minimize risk of
accidentally impacting RSA key generation and validation. To that end,
we only perform ACCP-aware validation when parsing from DER, and not
when doing key generation or general validation. The relevant RFC
([RFC-8107][2]) does not seem to explicitly permit omitted (i.e. NULL
type [ASN.1 tag value][3] of 5) RSA parameters when parsing RSA private
keys, so we consider an INTEGER value of 0 to indicate that a parameter
is missing in cases pertaining to ACCP.

This change resolves i/CryptoAlg-1026.

[1]: https://github.com/torben-hansen/aws-lc/compare/main...thoughts_rsa_accp_key_check
[2]: https://datatracker.ietf.org/doc/html/rfc8017#appendix-A.1.2
[3]: https://en.wikipedia.org/wiki/X.690#Identifier_octets

# Testing

- local testing with ACCP (cf. [this PR](https://github.com/corretto/amazon-corretto-crypto-provider/pull/183))
- automated CI testing via GitHub

### DIFF
--- a/crypto/fipsmodule/rsa/internal.h
+++ b/crypto/fipsmodule/rsa/internal.h
@@ -66,6 +66,11 @@
 extern "C" {
 #endif
 
+typedef enum {
+    RSA_STRIPPED_KEY,
+    RSA_CRT_KEY,
+    RSA_PUBLIC_KEY
+} rsa_asn1_key_encoding_t;
 
 // Default implementations of RSA operations.
 
@@ -110,7 +115,7 @@ int RSA_padding_add_none(uint8_t *to, size_t to_len, const uint8_t *from,
 
 // rsa_check_public_key checks that |rsa|'s public modulus and exponent are
 // within DoS bounds.
-int rsa_check_public_key(const RSA *rsa);
+int rsa_check_public_key(const RSA *rsa, rsa_asn1_key_encoding_t key_enc_type);
 
 // RSA_private_transform calls either the method-specific |private_transform|
 // function (if given) or the generic one. See the comment for
@@ -123,6 +128,8 @@ int RSA_private_transform(RSA *rsa, uint8_t *out, const uint8_t *in,
 extern const BN_ULONG kBoringSSLRSASqrtTwo[];
 extern const size_t kBoringSSLRSASqrtTwoLen;
 
+int RSA_validate_key(const RSA *rsa,
+  rsa_asn1_key_encoding_t key_enc_type);
 
 // Functions that avoid self-tests.
 //
--- a/crypto/fipsmodule/rsa/internal.h
+++ b/crypto/fipsmodule/rsa/internal.h
@@ -128,8 +128,7 @@ int RSA_private_transform(RSA *rsa, uint8_t *out, const uint8_t *in,
 extern const BN_ULONG kBoringSSLRSASqrtTwo[];
 extern const size_t kBoringSSLRSASqrtTwoLen;
 
-int RSA_validate_key(const RSA *rsa,
-  rsa_asn1_key_encoding_t key_enc_type);
+int RSA_validate_key(const RSA *rsa, rsa_asn1_key_encoding_t key_enc_type);
 
 // Functions that avoid self-tests.
 //
--- a/crypto/fipsmodule/rsa/rsa.c
+++ b/crypto/fipsmodule/rsa/rsa.c
@@ -714,7 +714,7 @@ static int check_mod_inverse(int *out_ok, const BIGNUM *a, const BIGNUM *ainv,
   return ret;
 }
 
-int RSA_check_key(const RSA *key) {
+int RSA_validate_key(const RSA *key, rsa_asn1_key_encoding_t key_enc_type) {
   // TODO(davidben): RSA key initialization is spread across
   // |rsa_check_public_key|, |RSA_check_key|, |freeze_private_key|, and
   // |BN_MONT_CTX_set_locked| as a result of API issues. See
@@ -724,10 +724,6 @@ int RSA_check_key(const RSA *key) {
   if (RSA_is_opaque(key)) {
     // Opaque keys can't be checked.
     return 1;
-  }
-
-  if (!rsa_check_public_key(key)) {
-    return 0;
   }
 
   if ((key->p != NULL) != (key->q != NULL)) {
@@ -741,6 +737,15 @@ int RSA_check_key(const RSA *key) {
       (BN_is_negative(key->d) || BN_cmp(key->d, key->n) >= 0)) {
     OPENSSL_PUT_ERROR(RSA, RSA_R_D_OUT_OF_RANGE);
     return 0;
+  }
+
+  if (!rsa_check_public_key(key, key_enc_type)) {
+    return 0;
+  }
+
+  if (key_enc_type == RSA_STRIPPED_KEY) {
+    // stripped keys doesn't have more parameters we can verify.
+    return 1;
   }
 
   if (key->d == NULL || key->p == NULL) {
@@ -843,6 +848,10 @@ out:
   return ok;
 }
 
+int RSA_check_key(const RSA *key) {
+  return RSA_validate_key(key, RSA_CRT_KEY);
+}
+
 
 // This is the product of the 132 smallest odd primes, from 3 to 751.
 static const BN_ULONG kSmallFactorsLimbs[] = {
@@ -872,7 +881,7 @@ int RSA_check_fips(RSA *key) {
     return 0;
   }
 
-  if (!RSA_check_key(key)) {
+  if (!RSA_validate_key(key, RSA_CRT_KEY)) {
     return 0;
   }
 
--- a/crypto/fipsmodule/rsa/rsa_impl.c
+++ b/crypto/fipsmodule/rsa/rsa_impl.c
@@ -73,7 +73,7 @@
 #include "../rand/fork_detect.h"
 
 
-int rsa_check_public_key(const RSA *rsa) {
+int rsa_check_public_key(const RSA *rsa, rsa_asn1_key_encoding_t key_enc_type) {
   if (rsa->n == NULL || rsa->e == NULL) {
     OPENSSL_PUT_ERROR(RSA, RSA_R_VALUE_MISSING);
     return 0;
@@ -85,6 +85,24 @@ int rsa_check_public_key(const RSA *rsa) {
     return 0;
   }
 
+  // Verify |n > e|. Comparing |n_bits| to |kMaxExponentBits| is a small
+  // shortcut to comparing |n| and |e| directly. In reality, |kMaxExponentBits|
+  // is much smaller than the minimum RSA key size that any application should
+  // accept.
+  static const unsigned kMaxExponentBits = 33;
+  if (n_bits <= kMaxExponentBits) {
+    OPENSSL_PUT_ERROR(RSA, RSA_R_KEY_SIZE_TOO_SMALL);
+    return 0;
+  }
+
+  if (key_enc_type == RSA_STRIPPED_KEY || rsa->e == NULL || BN_is_zero(rsa->e)) {
+    // Stripped RSA key doesn't have |e| set. Nothing more to do here.
+    return 1;
+  }
+
+  // Slighty more thorough validation of |n > e|/
+  assert(BN_ucmp(rsa->n, rsa->e) > 0);
+
   // Mitigate DoS attacks by limiting the exponent size. 33 bits was chosen as
   // the limit based on the recommendations in [1] and [2]. Windows CryptoAPI
   // doesn't support values larger than 32 bits [3], so it is unlikely that
@@ -94,7 +112,6 @@ int rsa_check_public_key(const RSA *rsa) {
   // [1] https://www.imperialviolet.org/2012/03/16/rsae.html
   // [2] https://www.imperialviolet.org/2012/03/17/rsados.html
   // [3] https://msdn.microsoft.com/en-us/library/aa387685(VS.85).aspx
-  static const unsigned kMaxExponentBits = 33;
   unsigned e_bits = BN_num_bits(rsa->e);
   if (e_bits > kMaxExponentBits ||
       // Additionally reject e = 1 or even e. e must be odd to be relatively
@@ -104,16 +121,6 @@ int rsa_check_public_key(const RSA *rsa) {
     OPENSSL_PUT_ERROR(RSA, RSA_R_BAD_E_VALUE);
     return 0;
   }
-
-  // Verify |n > e|. Comparing |n_bits| to |kMaxExponentBits| is a small
-  // shortcut to comparing |n| and |e| directly. In reality, |kMaxExponentBits|
-  // is much smaller than the minimum RSA key size that any application should
-  // accept.
-  if (n_bits <= kMaxExponentBits) {
-    OPENSSL_PUT_ERROR(RSA, RSA_R_KEY_SIZE_TOO_SMALL);
-    return 0;
-  }
-  assert(BN_ucmp(rsa->n, rsa->e) > 0);
 
   return 1;
 }
@@ -263,7 +270,7 @@ int RSA_encrypt(RSA *rsa, size_t *out_len, uint8_t *out, size_t max_out,
                 const uint8_t *in, size_t in_len, int padding) {
   boringssl_ensure_rsa_self_test();
 
-  if (!rsa_check_public_key(rsa)) {
+  if (!rsa_check_public_key(rsa, RSA_PUBLIC_KEY)) {
     return 0;
   }
 
@@ -600,7 +607,7 @@ static int mod_exp(BIGNUM *r0, const BIGNUM *I, RSA *rsa, BN_CTX *ctx);
 int rsa_verify_raw_no_self_test(RSA *rsa, size_t *out_len, uint8_t *out,
                                 size_t max_out, const uint8_t *in,
                                 size_t in_len, int padding) {
-  if (!rsa_check_public_key(rsa)) {
+  if (!rsa_check_public_key(rsa, RSA_PUBLIC_KEY)) {
     return 0;
   }
 
@@ -787,7 +794,7 @@ int rsa_default_private_transform(RSA *rsa, uint8_t *out, const uint8_t *in,
   // than the CRT attack, but there have likely been improvements since 1997.
   //
   // This check is cheap assuming |e| is small; it almost always is.
-  if (rsa->e != NULL) {
+  if (rsa->e != NULL && !BN_is_zero(rsa->e)) {
     BIGNUM *vrfy = BN_CTX_get(ctx);
     if (vrfy == NULL ||
         !BN_mod_exp_mont(vrfy, result, rsa->e, rsa->n, ctx, rsa->mont_n) ||
@@ -1303,7 +1310,7 @@ static int rsa_generate_key_impl(RSA *rsa, int bits, const BIGNUM *e_value,
   // The key generation process is complex and thus error-prone. It could be
   // disastrous to generate and then use a bad key so double-check that the key
   // makes sense.
-  if (!RSA_check_key(rsa)) {
+  if (!RSA_validate_key(rsa, RSA_CRT_KEY)) {
     OPENSSL_PUT_ERROR(RSA, RSA_R_INTERNAL_ERROR);
     goto err;
   }
--- a/crypto/fipsmodule/rsa/rsa_impl.c
+++ b/crypto/fipsmodule/rsa/rsa_impl.c
@@ -74,7 +74,9 @@
 
 
 int rsa_check_public_key(const RSA *rsa, rsa_asn1_key_encoding_t key_enc_type) {
-  if (rsa->n == NULL || rsa->e == NULL) {
+  // Despite its name, this function is used for validating all RSA keys. |n|
+  // is required for all keys, but stripped private keys do not have |e|.
+  if (rsa->n == NULL || (key_enc_type != RSA_STRIPPED_KEY && rsa->e == NULL)) {
     OPENSSL_PUT_ERROR(RSA, RSA_R_VALUE_MISSING);
     return 0;
   }
@@ -95,7 +97,7 @@ int rsa_check_public_key(const RSA *rsa, rsa_asn1_key_encoding_t key_enc_type) {
     return 0;
   }
 
-  if (key_enc_type == RSA_STRIPPED_KEY || rsa->e == NULL || BN_is_zero(rsa->e)) {
+  if (key_enc_type == RSA_STRIPPED_KEY) {
     // Stripped RSA key doesn't have |e| set. Nothing more to do here.
     return 1;
   }
@@ -794,7 +796,7 @@ int rsa_default_private_transform(RSA *rsa, uint8_t *out, const uint8_t *in,
   // than the CRT attack, but there have likely been improvements since 1997.
   //
   // This check is cheap assuming |e| is small; it almost always is.
-  if (rsa->e != NULL && !BN_is_zero(rsa->e)) {
+  if (rsa->e != NULL) {
     BIGNUM *vrfy = BN_CTX_get(ctx);
     if (vrfy == NULL ||
         !BN_mod_exp_mont(vrfy, result, rsa->e, rsa->n, ctx, rsa->mont_n) ||
--- a/crypto/rsa_extra/rsa_asn1.c
+++ b/crypto/rsa_extra/rsa_asn1.c
@@ -102,7 +102,7 @@ RSA *RSA_parse_public_key(CBS *cbs) {
     return NULL;
   }
 
-  if (!RSA_check_key(ret)) {
+  if (!RSA_validate_key(ret, RSA_PUBLIC_KEY)) {
     OPENSSL_PUT_ERROR(RSA, RSA_R_BAD_RSA_PARAMETERS);
     RSA_free(ret);
     return NULL;
@@ -153,6 +153,15 @@ int RSA_public_key_to_bytes(uint8_t **out_bytes, size_t *out_len,
 // RSAPrivateKey structure (RFC 3447).
 static const uint64_t kVersionTwoPrime = 0;
 
+// Distinguisher for stripped ACCP RSA private keys.
+// Return 1 if ACCP stripped private key.
+// Return 0 otherwise.
+static int detect_stripped_accp_private_key(const RSA *key) {
+  return (key->e == NULL || BN_is_zero(key->e)) &&
+         (key->p == NULL || BN_is_zero(key->p)) &&
+         (key->q == NULL || BN_is_zero(key->q));
+}
+
 RSA *RSA_parse_private_key(CBS *cbs) {
   RSA *ret = RSA_new();
   if (ret == NULL) {
@@ -188,7 +197,12 @@ RSA *RSA_parse_private_key(CBS *cbs) {
     goto err;
   }
 
-  if (!RSA_check_key(ret)) {
+  rsa_asn1_key_encoding_t rsa_enc_key_type = RSA_CRT_KEY;
+  if (detect_stripped_accp_private_key(ret) == 1) {
+    rsa_enc_key_type = RSA_STRIPPED_KEY;
+  }
+
+  if (!RSA_validate_key(ret, rsa_enc_key_type)) {
     OPENSSL_PUT_ERROR(RSA, RSA_R_BAD_RSA_PARAMETERS);
     goto err;
   }
--- a/crypto/rsa_extra/rsa_asn1.c
+++ b/crypto/rsa_extra/rsa_asn1.c
@@ -153,13 +153,30 @@ int RSA_public_key_to_bytes(uint8_t **out_bytes, size_t *out_len,
 // RSAPrivateKey structure (RFC 3447).
 static const uint64_t kVersionTwoPrime = 0;
 
-// Distinguisher for stripped ACCP RSA private keys.
-// Return 1 if ACCP stripped private key.
-// Return 0 otherwise.
-static int detect_stripped_accp_private_key(const RSA *key) {
-  return (key->e == NULL || BN_is_zero(key->e)) &&
-         (key->p == NULL || BN_is_zero(key->p)) &&
-         (key->q == NULL || BN_is_zero(key->q));
+// Distinguisher for stripped ACCP RSA private keys, sets zeroed values to NULL
+// because ASN.1 treats absent values as 0, but post-parsing validation logic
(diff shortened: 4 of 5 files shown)


crypto/fipsmodule/rsa/rsa_impl.c
```diff
@@ -113,6 +108,27 @@ int rsa_check_public_key(const RSA *rsa) {
     OPENSSL_PUT_ERROR(RSA, RSA_R_KEY_SIZE_TOO_SMALL);
     return 0;
   }
+
+  return 1;
+}
+
+rsa_key_type_t detect_rsa_key_type(const RSA *rsa) {
+  if (rsa->d == NULL || BN_is_zero(rsa->d)) {
```
torben-hansen, line 116: I'm not sure this is OK. Technically, you don't need `d` for RSA-CRT, you just need the intermediate values to compute the CRT.

It at least looks like the default signing implementation wouldn't allow this though:
```
int rsa_default_private_transform(RSA *rsa, uint8_t *out, const uint8_t *in,
                                  size_t len) {
  if (rsa->n == NULL || rsa->d == NULL) {
    OPENSSL_PUT_ERROR(RSA, RSA_R_VALUE_MISSING);
    return 0;
  }
```

But who knows - maybe in the future this will not be the case and we will infer the wrong key type here.

We might want to rethink this function. It is also a behaviour change no? Since non-private keys before were only identified by looking at `p/q`.

crypto/fipsmodule/rsa/rsa_impl.c
```diff
@@ -113,6 +108,27 @@ int rsa_check_public_key(const RSA *rsa) {
...
+  return RSA_KEY_PRIVATE_CRT;
+}
+
+int rsa_check_public_key(const RSA *rsa) {
+  if (rsa->n == NULL || rsa->e == NULL) {
+    OPENSSL_PUT_ERROR(RSA, RSA_R_VALUE_MISSING);
+    return 0;
+  }
+
+  rsa_check_key_common(rsa, RSA_KEY_PUBLIC);
```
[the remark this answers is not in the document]
WillChilds-Klein: ah, shoot. yes i missed the ret val check here. tests still passed because the second call to `rsa_check_key_common` in `rsa_check_key_internal` caught issues with such keys, because as you point out we're calling `common` twice for public keys.

i'm going to rework this such that we call `common` exactly once for each key type.

crypto/fipsmodule/rsa/internal.h
```diff
@@ -112,6 +117,12 @@ int RSA_padding_add_none(uint8_t *to, size_t to_len, const uint8_t *from,
 // within DoS bounds.
 int rsa_check_public_key(const RSA *rsa);
 
+int rsa_check_key_internal(const RSA *rsa, const rsa_key_type_t key_type);
```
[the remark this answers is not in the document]
WillChilds-Klein: will do.

crypto/fipsmodule/rsa/rsa_impl.c
WillChilds-Klein: yeah, i agree that we should try to "future-proof" the key type detection logic to the best of our ability.

while there is the check on `d` there in the default private transform, i don't believe `d` is actually used when the [CRT params are present](https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/rsa/rsa_impl.c#L752) (i.e. function `mod_exp` is called, and i don't see any uses of `d` in [that function](https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/rsa/rsa_impl.c#L849-L930), only `d mod (p-1)`). when the CRT params are omitted, `d` [is used](https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/rsa/rsa_impl.c#L763). so, i could see a world in which we relax the NULL check on `d` at the top of `rsa_default_private_transform`.

how about these criteria for determining key type?

- PUBLIC: `e` is present and non-zero, and any of `p`, `q`, or `d` are absent or zero
- PRIVATE_STRIPPED: `d` is present and non-zero, and any of `p`, `q`, or `iqmp` are absent or zero, or `dmp1` or `dmq1` are absent
- PRIVATE_CRT: all of `p`, `q`, and `iqmp` are present and non-zero, and `dmp1` and `dmq1` are both present
- UNKNOWN: none of the above conditions met. we return an error immediately when we detect an unknown key type.

dkostic: `d` is not used in decryption because `d mod (p-1)` and `d mod (q-1)` are already stored as part of the key. So I think classifying a key as PRIVATE_CRT should be done when: `p` and `q` are present and `d` or `(d mod (p-1), d mod (q-1))` is present. Since the existing functions already have the check on `d` and with `d` one can compute `(d mod (p-1), d mod (q-1))`, I would rather leave it in the new function as well.

WillChilds-Klein: ah, that makes sense. thanks dusan, i'll implement that and update the tests accordingly.

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -704,7 +704,17 @@ int RSA_check_key(const RSA *key) {
...
 
-  if (!rsa_check_public_key(key)) {
+  const rsa_key_type_t key_type = rsa_detect_key_type(key);
+  if (key_type == RSA_KEY_UNKNOWN) {
+    OPENSSL_PUT_ERROR(RSA, RSA_R_UNKOWN_KEY_TYPE);
+    return 0;
+  }
+
+  if (key_type == RSA_KEY_PUBLIC) {
+    return rsa_check_public_key(key);
```
SalusaSecondus, line 714: Why do we short-circuit before calling `rsa_check_key_common`? (Should that call come before anything else as it is "common"?

crypto/fipsmodule/rsa/rsa_impl.c
```diff
@@ -72,13 +72,7 @@
...
 #include "../rand/fork_detect.h"
 
-
-int rsa_check_public_key(const RSA *rsa) {
-  if (rsa->n == NULL || rsa->e == NULL) {
-    OPENSSL_PUT_ERROR(RSA, RSA_R_VALUE_MISSING);
-    return 0;
-  }
-
+int rsa_check_key_common(const RSA *rsa, const rsa_key_type_t key_type) {
```
SalusaSecondus, line 75: Should this also check for the modulus being too small?

crypto/fipsmodule/rsa/rsa_impl.c
```diff
@@ -113,6 +108,42 @@ int rsa_check_public_key(const RSA *rsa) {
...
+      (rsa->d == NULL || BN_is_zero(rsa->d))) {
+      return RSA_KEY_PUBLIC;
+  }
+  // Stripped keys must have |d| but do not have |p| or |q|.
+  if (rsa->d != NULL && !BN_is_zero(rsa->d) &&
+      (rsa->p == NULL || BN_is_zero(rsa->p) ||
+       rsa->q == NULL || BN_is_zero(rsa->q))) {
+      return RSA_KEY_PRIVATE_STRIPPED;
+  }
+  // CRT keys must have |p| and |q|, but needn't have |d|.
```
SalusaSecondus, line 130: This should also check the other values associated with CRT keys as well.

crypto/fipsmodule/rsa/rsa.c
WillChilds-Klein: this is because, after the refactor, `rsa_check_public_key` [calls](https://github.com/awslabs/aws-lc/pull/436/files#diff-25b26f92c168808944b0db449431b2f1f8bc5cb61b99a9a7aedce57811512f48R144) `rsa_check_key_common` and we want to avoid calling it twice.

why call `rsa_check_key_common` in `rsa_check_public_key` if we already call it in `RSA_check_key`? because there are [other callers](https://github.com/awslabs/aws-lc/search?q=rsa_check_public_key) outside of `RSA_check_key` and we want them to benefit from the validation around `n` and `e` as well.

crypto/fipsmodule/rsa/rsa_impl.c
WillChilds-Klein: perhaps. i tried to hew as closely as possible to existing logic and code paths (while still allowing us to use stripped keys, which motivated this PR), so i did not add additional validation.

if AWS-LC folks would like such validation in this code path and can point me towards some guidance on modulus size (i'd guess that this is mostly an issue w/r/t key generation?), i'd be happy to take a stab at adding it, perhaps in another PR to keep the scope of this one tight.

dkostic: what would be the threshold for too small modulus? I'm afraid that if we set the threshold to something reasonable like 1024 bits we'll break something because I bet there are 512/768 bit moduli still in use.

If you are referring to the check "is n > e" this check is done further below in the function.

crypto/fipsmodule/rsa/rsa_impl.c
```diff
@@ -113,6 +108,42 @@ int rsa_check_public_key(const RSA *rsa) {
     OPENSSL_PUT_ERROR(RSA, RSA_R_KEY_SIZE_TOO_SMALL);
     return 0;
   }
+
+  return 1;
+}
+
+rsa_key_type_t rsa_detect_key_type(const RSA *rsa) {
```
torben-hansen, line 115: This could be a static function in `rsa.c`.

crypto/fipsmodule/rsa/internal.h
```diff
@@ -112,6 +111,22 @@ int RSA_padding_add_none(uint8_t *to, size_t to_len, const uint8_t *from,
 // within DoS bounds.
 int rsa_check_public_key(const RSA *rsa);
 
+// An enum used internally to designate what type of RSA key we detect upon
+// validation. Used to determine which validation rules to enforce.
+typedef enum {
+    RSA_KEY_PRIVATE_STRIPPED,
```
torben-hansen, line 117: "Stripped keys" does not have a well-defined semantic (AFAIK).
We should be clear why we introduce this classification and where it comes from.

crypto/fipsmodule/rsa/rsa_impl.c
torben-hansen: The private key operation implementations support not having CRT values in the RSA object. So, not sure this would be a valid classification.

crypto/fipsmodule/rsa/internal.h
```diff
@@ -112,6 +111,22 @@ int RSA_padding_add_none(uint8_t *to, size_t to_len, const uint8_t *from,
...
+    RSA_KEY_PRIVATE_CRT,
+    RSA_KEY_PUBLIC,
+    RSA_KEY_UNKNOWN,
+} rsa_key_type_t;
+
+// An internal mthod used to determine which validation rules to enforce on an
```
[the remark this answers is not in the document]
WillChilds-Klein: sure.

crypto/fipsmodule/rsa/internal.h
```diff
@@ -112,6 +111,22 @@ int RSA_padding_add_none(uint8_t *to, size_t to_len, const uint8_t *from,
...
+// arbitrary RSA key.
```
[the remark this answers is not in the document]
WillChilds-Klein: sure.

crypto/fipsmodule/rsa/internal.h
```diff
@@ -112,6 +111,22 @@ int RSA_padding_add_none(uint8_t *to, size_t to_len, const uint8_t *from,
...
+rsa_key_type_t rsa_detect_key_type(const RSA *rsa);
+
+// Common validation logic shared between _all_ types of RSA keys.
```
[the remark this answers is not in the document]
WillChilds-Klein: sure.

crypto/fipsmodule/rsa/rsa_impl.c
WillChilds-Klein: sure.

crypto/fipsmodule/rsa/internal.h
WillChilds-Klein: we discussed this offline and arrived at the consensus that this change should keep its scope as narrow as possible while still supporting keys passed from JCA/ACCP. the net effect of that the mos recent revision ([`8acfe39`](https://github.com/awslabs/aws-lc/commit/8acfe39fe5059a65414d795c50f4c9ff7d9e2182)) hews much more closely to torben's [initial sketch](https://github.com/torben-hansen/aws-lc/compare/main...thoughts_rsa_accp_key_check), with a few modifications by me to get it to compile, account for some ACCP quirks, add test coverage, etc.

crypto/fipsmodule/rsa/rsa_impl.c
WillChilds-Klein: reached consensus offline on this, no new tests added.

WillChilds-Klein:
hm, formal verification checks appear to currently be failing when verifying some of our RSA PSS signing proofs:

```
[18:10:52.009] Subgoal failed: RSA_sign_pss_mgf1 safety assertion:
/codebuild/output/src807045921/src/github.com/awslabs/aws-lc-verification-build/SAW/proof/RSA/RSA.saw:646:1: error: in _SAW_verify_prestate
Literal equality postcondition
Expected term: Cryptol.ecNumber (Cryptol.TCNum 1) (Prelude.Vec 32 Prelude.Bool)
  (Cryptol.PLiteralSeqBool (Cryptol.TCNum 32))
Actual term:   let { x@1 = Prelude.Vec 32 Prelude.Bool
```

which appears to be [this SAW routine](https://github.com/awslabs/aws-lc-verification/blob/master/SAW/proof/RSA/RSA.saw#L646). i wonder if this commits [one-line change](https://github.com/awslabs/aws-lc/pull/436/files#diff-25b26f92c168808944b0db449431b2f1f8bc5cb61b99a9a7aedce57811512f48R784) adding an additional restriction ("non-null and non-zero" instead of just "non-null") to [a conditional in the private transform method](https://github.com/awslabs/aws-lc/pull/436/files#diff-25b26f92c168808944b0db449431b2f1f8bc5cb61b99a9a7aedce57811512f48R784) (i.e. the common code used for signing and encrypting) is the culprit...

dkostic (approved):
lgtm (modulo the FV CI failure)

WillChilds-Klein:
thanks @dkostic. it looks like the most [recent upstream merge](https://github.com/awslabs/aws-lc/pull/445) caused a merge conflict with this change, so i've resolved that conflict and re-pushed to this branch. if the next CI formal verification run fails again, i'll follow up with apetcher.

crypto/fipsmodule/rsa/rsa_impl.c
```diff
@@ -787,7 +794,7 @@ int rsa_default_private_transform(RSA *rsa, uint8_t *out, const uint8_t *in,
   // than the CRT attack, but there have likely been improvements since 1997.
   //
   // This check is cheap assuming |e| is small; it almost always is.
-  if (rsa->e != NULL) {
+  if (rsa->e != NULL && !BN_is_zero(rsa->e)) {
```
apetcher-amazon, line 797: Is it necessary to add this additional representation of "e is missing" to the RSA struct? Why not check for e==0 when parsing and then set e to null in the struct? The danger of requiring these additional checks for e==0 here is we would need to make sure we are not missing any, and we need to make sure they are added to any upstream code that is incorporated.

apetcher-amazon: Related: do we need to check for e==0 in any of the other places above where we check whether the pointer is null? For example around line 750?

apetcher-amazon (commented):
Formal verification fails because the spec doesn't account for the new way to represent "e is missing" in this code. Should be relatively easy to fix, but I want to make sure this is actually the change we want before I fix the spec.

crypto/fipsmodule/rsa/rsa_impl.c
WillChilds-Klein: I agree that adding an additional notion of "missing" to this logic is a bit of a code smell. In an earlier iteration of this changeset, I tried doing as you suggest by modifying `rsa_asn1.c` to `NULL`ify zero-valued BNs, but there were many places in that portion of the code that assumed RSA struct members to be non-`NULL`. It felt like the change was quickly increasing in scope, so i backtracked.

We should probably examine that approach more comprehensively, though, so I'll work on a branch implementing that approach so we can compare it side-by-side with the current changeset. Will circle back here when that's ready.

WillChilds-Klein: ok, turns out the changes were pretty straightforward (commit [here](https://github.com/awslabs/aws-lc/compare/main...WillChilds-Klein:null-e-asn1), full diff with `main` [here](https://github.com/awslabs/aws-lc/commit/e50bee977e699327d1414b3b30345af87194a515)). i'm going to clean up some comments/tests and post another revision.

WillChilds-Klein: ok, new revision posted that doesn't introduce the notion of "0 means absent" to  `rsa.c` or `rsa_impl.c`. formal verification appears to be happy now.

crypto/rsa_extra/rsa_asn1.c
```diff
@@ -153,6 +153,32 @@ int RSA_public_key_to_bytes(uint8_t **out_bytes, size_t *out_len,
...
 
+// Distinguisher for stripped ACCP RSA private keys, sets zeroed values to NULL
+// because ASN.1 treats absent values as 0, but post-parsing validation logic
+// expects absent values to be NULL. Returns 1 if ACCP stripped private key, 0
+// otherwise.
+static int detect_stripped_accp_private_key(RSA *key) {
+  if (!BN_is_zero(key->d) && !BN_is_zero(key->n)
+          && BN_is_zero(key->e) && BN_is_zero(key->iqmp)
+          && BN_is_zero(key->p) && BN_is_zero(key->q)
+          && BN_is_zero(key->dmp1) && BN_is_zero(key->dmq1)) {
```
dkostic, line 164: Nit:
```suggestion
  if (!BN_is_zero(key->d) && !BN_is_zero(key->n) &&
       BN_is_zero(key->e) && BN_is_zero(key->iqmp) &&
       BN_is_zero(key->p) && BN_is_zero(key->q) &&
       BN_is_zero(key->dmp1) && BN_is_zero(key->dmq1)) {
```

dkostic (approved):

dkostic (approved):

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -743,6 +739,15 @@ int RSA_check_key(const RSA *key) {
...
+  if (!rsa_check_public_key(key, key_enc_type)) {
+    return 0;
+  }
+
+  if (key_enc_type == RSA_STRIPPED_KEY) {
+    // stripped keys doesn't have more parameters we can verify.
+    return 1;
+  }
+
   if (key->d == NULL || key->p == NULL) {
```
SalusaSecondus, line 751: Why doesn't this look at `RSA_PUBLIC_KEY` instead of looking at the contained parameters?

crypto/rsa_extra/rsa_asn1.c
```diff
@@ -153,6 +153,32 @@ int RSA_public_key_to_bytes(uint8_t **out_bytes, size_t *out_len,
 // RSAPrivateKey structure (RFC 3447).
 static const uint64_t kVersionTwoPrime = 0;
 
+// Distinguisher for stripped ACCP RSA private keys, sets zeroed values to NULL
+// because ASN.1 treats absent values as 0, but post-parsing validation logic
+// expects absent values to be NULL. Returns 1 if ACCP stripped private key, 0
+// otherwise.
+static int detect_stripped_accp_private_key(RSA *key) {
```
torben-hansen, line 160:
```suggestion
static int detect_stripped_jca_private_key(RSA *key) {
```

crypto/rsa_extra/rsa_asn1.c
```diff
@@ -188,7 +214,12 @@ RSA *RSA_parse_private_key(CBS *cbs) {
     goto err;
   }
 
-  if (!RSA_check_key(ret)) {
+  rsa_asn1_key_encoding_t rsa_enc_key_type = RSA_CRT_KEY;
+  if (detect_stripped_accp_private_key(ret) == 1) {
```
torben-hansen, line 218:
```suggestion
  if (detect_stripped_jca_private_key(ret) == 1) {
```

crypto/rsa_extra/rsa_test.cc
```diff
@@ -578,13 +578,71 @@ TEST(RSATest, OnlyDGiven) {
...
+
   ASSERT_LE(RSA_size(key2.get()), sizeof(buf));
   EXPECT_TRUE(RSA_sign(NID_sha256, kDummyHash, sizeof(kDummyHash), buf,
                        &buf_len, key2.get()));
 
   // Verify the signature with |key|. |key2| has no public exponent.
   EXPECT_TRUE(RSA_verify(NID_sha256, kDummyHash, sizeof(kDummyHash), buf,
                          buf_len, key.get()));
+
+  // Perform the same test case as above ("ACCP-style" key specified in terms
```
torben-hansen, line 593:
```suggestion
  // Perform the same test case as above ("JCA-style" key specified in terms
```

crypto/rsa_extra/rsa_test.cc
```diff
@@ -578,13 +578,71 @@ TEST(RSATest, OnlyDGiven) {
...
+  // ASN.1 first. This most closely replicates ACCP's use-case of asking AWS-LC
+  // to encode to and decode from DER. While already-parsed keys tend to be
+  // validated and used with "NULL"ness indicating absence of various RSA
+  // parameters, the actual parsing logic assumes that these NULL'd values are
+  // present and zero-valued when being un/marshalled. So, we zero them out
+  // before marshalling, and assert that they were NULLed when parsing back
+  // from DER.
+  //
+  // Also, note that here and in the previous test we expect RSA_check_key to
+  // return false for ACCP-style keys. The current implementation does not
```
torben-hansen, line 604:
```suggestion
  // return false for JCA-style keys. The current implementation does not
```

crypto/rsa_extra/rsa_test.cc
```diff
@@ -578,13 +578,71 @@ TEST(RSATest, OnlyDGiven) {
...
+  // consider ACCP-style keys to be valid. We deliberately made this decision
```
torben-hansen, line 605:
```suggestion
  // consider JCA-style keys to be valid. We deliberately made this decision
```

crypto/rsa_extra/rsa_test.cc
```diff
@@ -578,13 +578,71 @@ TEST(RSATest, OnlyDGiven) {
...
+  // to avoid introducing potentially dangerous modifications to key validation
+  // and generation logic. Instead, we detect ACCP-style keys only when parsing
```
torben-hansen, line 607:
```suggestion
  // and generation logic. Instead, we detect JCA-style keys only when parsing
```

crypto/rsa_extra/rsa_asn1.c
SalusaSecondus, line 160: Let's pull `ACCP` out of this. This is a result of how the JCA is defined and ACCP just meets its requirements.

crypto/rsa_extra/rsa_test.cc
```diff
@@ -578,13 +578,71 @@ TEST(RSATest, OnlyDGiven) {
   ASSERT_TRUE(BN_hex2bn(&key2->d, kD));
   key2->flags |= RSA_FLAG_NO_BLINDING;
 
+  // While keys defined only in terms of |n| and |d| must be functional, our
+  // validation logic doesn't consider them "valid".
+  EXPECT_FALSE(RSA_check_key(key2.get()));
```
torben-hansen, line 583: Nice

crypto/rsa_extra/rsa_test.cc
SalusaSecondus, line 593: Remove "ACCP".

crypto/rsa_extra/rsa_test.cc
```diff
@@ -578,13 +578,71 @@ TEST(RSATest, OnlyDGiven) {
...
+  // Also, note that here and in the previous test we expect RSA_check_key to
+  // return false for ACCP-style keys. The current implementation does not
+  // consider ACCP-style keys to be valid. We deliberately made this decision
+  // to avoid introducing potentially dangerous modifications to key validation
+  // and generation logic. Instead, we detect ACCP-style keys only when parsing
+  // from DER and special case that.
+  //
+  // At some point in the future, we will likely want to standardize on one of
+  // of NULL/0 for indicating parameter absence across the codebase, as well as
+  // fix up RSA_check_key to accurately account for all the different types of
```
torben-hansen, line 612:
```suggestion
  // fix up |RSA_check_key| to accurately account for all the different types of
```

crypto/rsa_extra/rsa_test.cc
```diff
@@ -578,13 +578,71 @@ TEST(RSATest, OnlyDGiven) {
...
   EXPECT_TRUE(RSA_sign(NID_sha256, kDummyHash, sizeof(kDummyHash), buf,
                        &buf_len, key2.get()));
 
   // Verify the signature with |key|. |key2| has no public exponent.
   EXPECT_TRUE(RSA_verify(NID_sha256, kDummyHash, sizeof(kDummyHash), buf,
                          buf_len, key.get()));
+
+  // Perform the same test case as above ("ACCP-style" key specified in terms
+  // of |d| and |n|, turn off blinding), but this time do a round trip to DER +
+  // ASN.1 first. This most closely replicates ACCP's use-case of asking AWS-LC
```
torben-hansen, line 595: Link to ACCP code, maybe?

crypto/rsa_extra/rsa_asn1.c
WillChilds-Klein: sure. will look to you and @torben-hansen for guidance on naming here.

crypto/rsa_extra/rsa_test.cc
WillChilds-Klein: sure. will look to you and @torben-hansen for guidance on naming here.

crypto/rsa_extra/rsa_test.cc
WillChilds-Klein: sure, can do.

crypto/rsa_extra/rsa_test.cc
WillChilds-Klein: sure.

crypto/fipsmodule/rsa/rsa.c
WillChilds-Klein: one goal of this change (per guidance from the AWS-LC team earlier in this PR) is to make this change as minimal as possible. so, we tried to preserve original logic inasmuch as possible to avoid accidentally modifying the behavior beyond ACCP's immediate need. prior iterations of this PR did attempt to use the new "key type" enum more liberally, but we walked that back a bit. in the current revision of this PR, the ACCP detector logic doesn't determine whether  to set `RSA_PUBLIC_KEY` as the key type -- that's only done explicitly in certain parts of the code.

this is kind of a long-winded way of saying "we could do that, but it's out of scope because consolidating our key validation logic needs to be done as a standalone paroject with proper design process". I'm currently working on a SIM with some suggestions for how AWS-LC might be able to do that.

crypto/rsa_extra/rsa_asn1.c
WillChilds-Klein: i have no issue with that. perhaps we can consolidate discussion with @SalusaSecondus around the naming issue on [this thread](https://github.com/awslabs/aws-lc/pull/436#discussion_r850827317)?

crypto/rsa_extra/rsa_asn1.c
WillChilds-Klein: i have no issue with that. perhaps we can consolidate discussion with @SalusaSecondus around the naming issue on [this thread](https://github.com/awslabs/aws-lc/pull/436#discussion_r850827317)?

crypto/rsa_extra/rsa_test.cc
WillChilds-Klein: i have no issue with that. perhaps we can consolidate discussion with @SalusaSecondus around the naming issue on [this thread](https://github.com/awslabs/aws-lc/pull/436#discussion_r850827317)?

crypto/rsa_extra/rsa_test.cc
WillChilds-Klein: i have no issue with that. perhaps we can consolidate discussion with @SalusaSecondus around the naming issue on [this thread](https://github.com/awslabs/aws-lc/pull/436#discussion_r850827317)?

crypto/rsa_extra/rsa_test.cc
WillChilds-Klein: i have no issue with that. perhaps we can consolidate discussion with @SalusaSecondus around the naming issue on [this thread](https://github.com/awslabs/aws-lc/pull/436#discussion_r850827317)?

crypto/rsa_extra/rsa_test.cc
WillChilds-Klein: i have no issue with that. perhaps we can consolidate discussion with @SalusaSecondus around the naming issue on [this thread](https://github.com/awslabs/aws-lc/pull/436#discussion_r850827317)?

crypto/rsa_extra/rsa_test.cc
WillChilds-Klein: i have no issue with that. perhaps we can consolidate discussion around the naming issue on [this thread](https://github.com/awslabs/aws-lc/pull/436#discussion_r850827317)?

crypto/rsa_extra/rsa_asn1.c
WillChilds-Klein: @SalusaSecondus -- are OK with referring to "JCA" instead of "ACCP"?

SalusaSecondus: Works for me!

torben-hansen (approved):

dkostic (approved):

PR #444, opened 2022-03-28
Update the analytics to only run on the main branch

andrewhop:
### Description of changes: 
Add a filter to the analytics webhook so it only runs on pushes to main. This is useful because not all branches have the analytics config yet and would create failing builds. 

### Testing:
I deployed this change to my local CI and made several pushes to branches and verified it only ran when I pushed to my main.

### DIFF
--- a/tests/ci/cdk/cdk/aws_lc_analytics_stack.py
+++ b/tests/ci/cdk/cdk/aws_lc_analytics_stack.py
@@ -28,6 +28,9 @@ class AwsLcGitHubAnalyticsStack(core.Stack):
             webhook=True,
             webhook_filters=[
                 codebuild.FilterGroup.in_event_of(codebuild.EventAction.PUSH)
+                # The current FIPS branch does not have the configuration needed to run the analytics, once we update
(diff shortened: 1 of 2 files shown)


tests/ci/cdk/cdk/aws_lc_analytics_stack.py
```diff
@@ -28,6 +28,9 @@ def __init__(self,
             webhook=True,
             webhook_filters=[
                 codebuild.FilterGroup.in_event_of(codebuild.EventAction.PUSH)
+                # The current FIPS branch does not have the configuration needed to run the analytics, once we update
+                # the branch or create a new FIPS branch it should be updated to '(main)|(fips.*)'
+                .and_branch_is("main")
```
bryce-shang, line 33: NIT: maybe `fips-*` can be added earlier before cutting the FIPS branch to avoid the deployment.

bryce-shang (approved):

bryce-shang (commented):
Do we plan to change the instance type? The larger instance type can avoid the go timeout but gives different metrics of test time.

bryce-shang (approved):

samuel40791765 (approved):

PR #445, opened 2022-03-28
Merge BoringSSL commits staged in `current-upstream-merge-20220318`

nebeid:
Merge BoringSSL commits between https://github.com/google/boringssl/commit/f7975703204c738144990efb8406eee0668fbf2e (Jan 13) and https://github.com/google/boringssl/commit/c7a3c46574e7fc32357b2cc68f961c56c72b0ca4 (Mar 15)

See PR #439 for details on the changes. 


crypto/fipsmodule/sha/asm/sha512-x86_64.pl
```diff
@@ -129,9 +129,11 @@
...
 # output, so this isn't useful anyway.
 #
-# TODO(davidben): Enable AVX2 code after testing by setting $avx to 2. Is it
-# necessary to disable AVX2 code when SHA Extensions code is disabled? Upstream
-# did not tie them together until after $shaext was added.
+# This file also has an AVX2 implementation, controlled by setting $avx to 2.
+# For now, we intentionally disable it. While it gives a 13-16% perf boost, the
+# CFI annotations are wrong. It allocates stack in a loop and should be
+# rewritten to avoid this.
+$avx = 1;
```
dkostic, line 136: `avx=1` is done twice (here, and in line 138). Also I think we can remove the `set to zero if compiling for 1.0.1` comment because it came from OpenSSL and it's not relevant for aws-lc

crypto/internal.h
```diff
@@ -934,24 +938,73 @@ static inline uint64_t CRYPTO_rotr_u64(uint64_t value, int shift) {
...
 
 #if defined(BORINGSSL_FIPS_COUNTERS)
 void boringssl_fips_inc_counter(enum fips_counter_t counter);
 #else
 OPENSSL_INLINE void boringssl_fips_inc_counter(enum fips_counter_t counter) {}
 #endif
 
+#if defined(BORINGSSL_FIPS_BREAK_TESTS)
+OPENSSL_INLINE int boringssl_fips_break_test(const char *test) {
+  const char *const value = getenv("BORINGSSL_FIPS_BREAK_TEST");
```
dkostic, line 999: I like this new method for break tests, but will this work on Windows, iOS, macOS?

nebeid: That's a good question. I guess if it doesn't work there, you can always revert back to flipping an input byte in the KAT. Also this getenv is now only for 
1. RSA_PWCT
1. ECDSA_PWCT
1. CRNG
as per FIPS.md
The other break tests are using a script that searches for the input value in the binary to flip them, which also is sensitive to the build; it doesn't seem to find the values in the shared build.

crypto/fipsmodule/sha/asm/sha512-x86_64.pl
nebeid: Done. Thanks.

samuel40791765:
Verified merge with runbook and commit doc.

samuel40791765 (approved):

dkostic (dismissed):

dkostic (approved):

samuel40791765 (approved):

PR #447, opened 2022-03-29
Add AWS CLI to ARM AmazonLinux2 base image

andrewhop:
### Description of changes: 
The ARM AmazonLinux2 base image is out of sync with the [X86 image which includes the AWS CLI](https://github.com/awslabs/aws-lc/blob/main/tests/ci/docker_images/linux-x86/amazonlinux-2_base/Dockerfile#L29-L33). This change adds it to get them both in sync.

### Testing:
Built this image on an ARM instance and verified the `aws` command is found.

### DIFF
--- a/tests/ci/docker_images/linux-aarch/amazonlinux-2_base/Dockerfile
+++ b/tests/ci/docker_images/linux-aarch/amazonlinux-2_base/Dockerfile
@@ -24,7 +24,13 @@ RUN set -ex && \
     wget \
     llvm \
     llvm-devel \
-    valgrind && \
+    valgrind \
+    unzip && \
+    # Based on https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html \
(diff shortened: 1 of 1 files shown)


bryce-shang (commented):
The CLI was added in x86 base image because `bcm.o` was planned to get from s3. After that, more CLI commands were used to call some AWS services. Do we need this synch to support analytics?

If we are going to deploy this Docker image change, try to avoid the fuzz image rebuild(see `CryptoAlg-1043?selectedConversation=2bb0d654-a668-4516-bc3c-fcc829e2730b`). The Fuzz related issue is not investigated yet, but I guess it may have something to do with below code. In other words, I am thinking if one code commit/release tag of the repo should be used to lock/minimize the behavior change after Docker image rebuild.
```
git clone --depth 1 https://github.com/guidovranken/cryptofuzz.git
cd cryptofuzz
```

samuel40791765 (approved):

PR #448, opened 2022-03-29
Add Android cross-compilation CI dimension with `AWSLCAndroidTestRunner` 

samuel40791765:
### Issues:
Addresses `CryptoAlg-895` 

### Description of changes: 
This change adds a new codebuild batch for android builds `Static`, `Shared`, `Debug`, `Release`, and `FIPS`. It also adds the `ANDROIDTESTRUNNER` flag mentioned in https://github.com/awslabs/aws-lc/pull/389, which is a new flag that compiles our test executables into `so` executables when it's on. Only Shared Object files can be automatically packaged into an Android app with the right execution permissions, so test executable outputs need to be outputted as an `so` for our `AWSLCAndroidTestRunner` app to be able to execute the tests within the app dimension.

### Call-outs:
The files under `tests/ci/codebuild/android/` were split into 5 dimensions for each build on purpose, so we can parallelize the Device Farm runs once the Device Farm functionality is turned on. 
 
### Testing:
Tested with proof of concept in a PR on my fork: https://github.com/samuel40791765/aws-lc/pull/10 (The status check with the name: `aws-lc-devicefarm-android`)


DominicDams (dismissed):

crypto/fipsmodule/ec/ec_test.cc
```diff
@@ -1163,6 +1163,8 @@ TEST(ECTest, SmallGroupOrder) {
 }
 
 #else
+// AWSLCAndroidTestRunner does not take tests that do |ASSERT_DEATH| very well.
+#if !defined(OPENSSL_ANDROID)
```
bryce-shang, line 1167: What does this mean? What behavior is observed?

tests/ci/cdk/app.py
```diff
@@ -50,5 +51,7 @@
 # TODO: re-enable 'aws-lc-ci-bm-framework' when it's ready.
 # bm_framework_build_spec_file = "./cdk/codebuild/bm_framework_omnibus.yaml"
 # BmFrameworkStack(app, "aws-lc-ci-bm-framework", LINUX_X86_ECR_REPO, bm_framework_build_spec_file, env=env)
+android_build_spec_file = "cdk/codebuild/github_ci_android_omnibus.yaml"
+AwsLcAndroidCIStack(app, "aws-lc-devicefarm-android", LINUX_X86_ECR_REPO, android_build_spec_file, env=env)
```
bryce-shang, line 55: Can this dimension get merged to x86 omnibus? GitHub has 20 webhook limits. New webhook/batch build stack should be avoided to leave spaces for future. Other stacks are separated due to various reasons(e.g. Fuzz one require EFS, ARM one requires ARM instance).

tests/ci/cdk/cdk/aws_lc_android_ci_stack.py
```diff
@@ -0,0 +1,73 @@
...
+
+        # Define a IAM role for this stack.
+        code_build_batch_policy = iam.PolicyDocument.from_json(
+            code_build_batch_policy_in_json([id])
+        )
+        inline_policies = {"code_build_batch_policy": code_build_batch_policy}
+        role = iam.Role(scope=self,
+                        id="{}-role".format(id),
+                        assumed_by=iam.ServicePrincipal("codebuild.amazonaws.com"),
+                        inline_policies=inline_policies)
```
bryce-shang, line 45: If this file is moved to x86 omnibus dimension, `device-farm` related permission/policy may be needed.

tests/ci/codebuild/android/run_android_fips.yml
```diff
@@ -0,0 +1,8 @@
+version: 0.2
+
+phases:
+  build:
+    commands:
+      - chmod +x ./tests/ci/android/AWSLCAndroidTestRunner/gradlew
+      - cd tests/ci/android/AWSLCAndroidTestRunner
+      - ./gradlew assembleDebug assembleAndroidTest -PFIPS
```
bryce-shang, line 8: * I may miss something. We don't need to call any AWS CLI(device farm) to start the project? These seems to run the logics in the Docker image.
* nit: Move these logics to some bash script like other yml files?

tests/ci/cdk/cdk/codebuild/github_ci_android_omnibus.yaml
```diff
@@ -0,0 +1,43 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+version: 0.2
+
+# Doc for batch https://docs.aws.amazon.com/codebuild/latest/userguide/batch-build-buildspec.html#build-spec.batch.build-list
```
bryce-shang, line 6: * Try to reuse existing omnibus.yml to avoid a new webhook creation?
* How long does each dimension take?
* How's the memory/CPU usage of `BUILD_GENERAL1_LARGE`? This info is available through metrics tab of the build job. If the usage is low, smaller instance type can be used but remember some timeout cases.

samuel40791765: update: the android apk packaging at the end of the build downloads and runs some jar files which `BUILD_GENERAL1_SMALL` seems to not be able to handle well, and fails. I'll be using `BUILD_GENERAL1_MEDIUM` instead.

tests/ci/cdk/run-cdk.sh
```diff
@@ -224,6 +224,11 @@ function setup_ci() {
   build_win_docker_images
 
   create_github_ci_stack
+  create_android_ci_stack
+}
+
+function create_android_ci_stack() {
+  cdk deploy aws-lc-ci-devicefarm-android --require-approval never
```
bryce-shang, line 231: Still need `create_android_ci_stack`?

```
function create_github_ci_stack() {
  cdk deploy aws-lc-ci-* --require-approval never
```

tests/ci/cdk/cdk/codebuild/github_ci_android_omnibus.yaml
bryce-shang: > Try to reuse existing omnibus.yml to avoid a new webhook creation?

I revisited this thought and think it makes sense to have a separated stack/webhook which will have device farm relevant resource definition.

tests/ci/cdk/app.py
bryce-shang: > I think we can merge it in with x86 omnibus.

https://github.com/awslabs/aws-lc/pull/448#discussion_r841955279

tests/ci/cdk/run-cdk.sh
samuel40791765: removed 👍

DominicDams (approved):

PR #449, opened 2022-03-30
CAVP should only be built if we have testing

DominicDams:
When we try to build with NO_TESTING currently it fails because CAVP
can't be built. We don't need CAVP when testing is off so this disables
CAVP if we don't have testing so we can successfully build without
testing.

### Issues:
Resolves CryptoAlg-934

### Testing:
Built locally


bryce-shang:
> cavp's going to removed in the new upstream merge #445 with this commit: [ea9fb94](https://github.com/awslabs/aws-lc/commit/ea9fb94c355afd24e5a877d67a7970c64005ad6f). Should we just wait until that's merged in, or do we need a quick fix right now?

Agree. We should just wait the PR merge.

DominicDams:
Seeing those comments I'm closing this

PR #450, opened 2022-03-31
Refactoring and fixing small issues in service indicator

dkostic:
### Issues:
N/A

### Description of changes: 
Refactoring and fixing small issues in service indicator

### Call-outs:

### Testing:


crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -57,7 +65,8 @@ uint64_t FIPS_service_indicator_after_call(void) {
 }
 
 int FIPS_service_indicator_check_approved(uint64_t before, uint64_t after) {
-  if(before != after) {
+
+  if (before < after) {
```
samuel40791765, line 69: I remember that we did `(before != after)` to avoid `after` overflowing and becoming less than `before`

dkostic: can we then remove the check in line 88, in update_state function? If the check is `before != after` why do we have a counter then, instead of a `{0, 1}` flag?

samuel40791765: Yeah, now that you point that out, line 88 was a recent change when we were trying to fix underflow/overflow errors with the service indicator locks. After our internal discussion, I think it's alright to do `before < after` if we're checking for overflowing there.

samuel40791765 (dismissed):

DominicDams (approved):

crypto/fipsmodule/service_indicator/service_indicator.c
nebeid: I'm now worried about a service that has used more than 2^64 FIPS operations since its start, wouldn't the check on l.87 fail when the counter = 2^{64} - 1? I think that's why we didn't have that check and we were only checking if the value changed. (Having that check on the lock makes sense because our call graph is few levels deep.)

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -149,39 +163,37 @@ void AES_verify_service_indicator(const EVP_CIPHER_CTX *ctx, const unsigned key_
...
   // We only have support for 128 bit and 256 bit keys for AES-GCM. AES-GCM is
   // approved only with an internal IV, see SP 800-38D Sec 8.2.2.
-  // Not the best way to write this, but the delocate parser for ARM/clang can't
-  // recognize || if statements, or switch statements for this.
-  // TODO: Update the delocate parser to be able to recognize a more readable
-  // version of this.
-   if(EVP_AEAD_key_length(ctx->aead) == 16) {
-    FIPS_service_indicator_update_state();
-   } else if(EVP_AEAD_key_length(ctx->aead) == 32) {
+  // Note: |EVP_AEAD_key_length| returns the length in bytes.
```
nebeid, line 169: I'm afraid to make those changes now before we work on the static build.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -149,39 +163,37 @@ void AES_verify_service_indicator(const EVP_CIPHER_CTX *ctx, const unsigned key_
...
+
   // Only 128 and 256 bit keys are approved for AES-CMAC.
-  switch (ctx->cipher_ctx.key_len) {
-    case 16:
-    case 32:
-      FIPS_service_indicator_update_state();
-      break;
-    default:
-      break;
+  // Note: |key_len| is the length in bytes.
```
nebeid, line 188: same comment as related to the delocator.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -196,177 +208,187 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
+      return 1;
+    default:
+      return 0;
+  }
+}
+
+// Updates the service indicator if the elliptic curve contained in |eckey|
+// is FIPS approved, otherwise does nothing.
+void EC_KEY_keygen_verify_service_indicator(EC_KEY *eckey) {
+
```
nebeid, line 229: Nit: remove the blank line?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -196,177 +208,187 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
-       default:
-         break;
-     }
-   }
-   if(ret) {
-     FIPS_service_indicator_update_state();
-   }
+
+  if (pkey->type == EVP_PKEY_RSA || pkey->type== EVP_PKEY_RSA_PSS) {
+
```
nebeid, line 239: Nit: remove the blank line?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -196,177 +208,187 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
+      case 384:
+      case 512:
+        FIPS_service_indicator_update_state();
+        break;
+      default:
+        break;
+    }
+
+  } else if (pkey->type == EVP_PKEY_EC) {
+
```
nebeid, line 255: Nit: remove the blank line?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -196,177 +208,187 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
+    if (is_ec_fips_approved(curve_nid)) {
+      FIPS_service_indicator_update_state();
+    }
+  }
+}
+
+// Returns 1 if the given message digest type is FIPS approved for signing, and
+// 0 otherwise. Only SHA2-224, SHA2-256, SHA2-384, and SHA2-512 are approved.
+static int is_md_fips_approved_for_signing(int md_type) {
+
```
nebeid, line 268: Nit: remove the blank line?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -196,177 +208,187 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
-          break;
+
+    // The approved RSA key sizes for signing are 2048, 3072 and 4096 bits.
+    // Note: |EVP_PKEY_size| returns the size in bytes.
+    size_t pkey_size = EVP_PKEY_size(ctx->pctx->pkey);
+
+    // Check if the MD type and the RSA key size are approved.
+    if (is_md_fips_approved_for_signing(md_type) &&
+        (pkey_size == 256 || pkey_size == 384 || pkey_size == 512)) {
+
```
nebeid, line 306: Nit: remove the blank line?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -196,177 +208,187 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
+      FIPS_service_indicator_update_state();
+    }
+
+  } else if (pkey_type == EVP_PKEY_EC) {
+
```
nebeid, line 311: Nit: remove the blank line?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -196,177 +208,187 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
 void DigestVerify_verify_service_indicator(const EVP_MD_CTX *ctx) {
-  if(ctx->pctx->pmeth->pkey_id == EVP_PKEY_RSA ||
-      ctx->pctx->pmeth->pkey_id == EVP_PKEY_RSA_PSS) {
-    // Hash digest used in the private key should be of the same type.
+
+  int pkey_type = ctx->pctx->pmeth->pkey_id;
+
+  if (pkey_type == EVP_PKEY_RSA || pkey_type == EVP_PKEY_RSA_PSS) {
+
```
nebeid, line 347: Nit: remove the blank line?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -196,177 +208,187 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
 
 void DigestSign_verify_service_indicator(const EVP_MD_CTX *ctx) {
-  if(ctx->pctx->pmeth->pkey_id == EVP_PKEY_RSA ||
-      ctx->pctx->pmeth->pkey_id == EVP_PKEY_RSA_PSS) {
-    // Hash digest used in the private key should be of the same type.
+
+  int pkey_type = ctx->pctx->pmeth->pkey_id;
+
+  if (pkey_type == EVP_PKEY_RSA || pkey_type == EVP_PKEY_RSA_PSS) {
+
```
nebeid, line 285: Nit: remove the blank line?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -196,177 +208,187 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
+      FIPS_service_indicator_update_state();
     }
   }
 }
 
+// Returns 1 if the given message digest type is FIPS approved for verifying,
+// and 0 otherwise. Only SHA1, SHA2-224, SHA2-256, SHA2-384, and SHA2-512
+// are approved.
+static int is_md_fips_approved_for_verifying(int md_type) {
+
```
nebeid, line 329: Nit: remove the blank line?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -196,177 +208,187 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
-    FIPS_service_indicator_update_state();
-  }
-  else if( curve == NID_X9_62_prime256v1) {
-    FIPS_service_indicator_update_state();
-  }
-  else if(curve == NID_secp384r1) {
-    FIPS_service_indicator_update_state();
-  }
-  else if(curve == NID_secp521r1) {
+
```
nebeid, line 389: Nit: remove the blank line?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -196,177 +208,187 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
+  int curve_nid = ec_key->group->curve_name;
```
nebeid, line 390: I hope the delocator will not complain.

crypto/fipsmodule/ec/ec_key.c
```diff
@@ -421,7 +421,7 @@ int EC_KEY_check_fips(const EC_KEY *key) {
 end:
   FIPS_service_indicator_unlock_state();
   if(ret){
-    FIPS_service_indicator_update_state();
+    EC_KEY_keygen_verify_service_indicator((EC_KEY*)key);
```
nebeid, line 424: Just curious: why was the cast needed here and not in the other change?

nebeid (commented):
Mostly "nits" about blank lines and continuing the discussion on updating/checking the counter.

crypto/fipsmodule/service_indicator/service_indicator.c
dkostic: yes, our current design allows only 2^64 FIPS approved operations, after that, it would always indicate "not approved". That's why my suggestion is to remove the check in line 87 and keep the `before != after` condition. This way we don't have the artificial limit on the number of FIPS approved operations and we don't have to care about overflowing the counter.

p.s. the counters are thread local so assuming that one FIPS approved operation takes 1 ns, it will take almost 600 years to overflow the counter :)

crypto/fipsmodule/service_indicator/service_indicator.c
dkostic: don't worry, I'll start working on the static build right after this so I'll change it back if need be :) The thing is, I really don't see how this has anything to do with delocator, so I'm hoping it was some other issue causing delocation to fail.

crypto/fipsmodule/service_indicator/service_indicator.c
dkostic: I added all the blank lines because to me it's more readable that way. If for you it's better without the blank lines I'll remove them?

crypto/fipsmodule/ec/ec_key.c
dkostic: the function argument in this function is declared as `const` so the compiler complains that we are dropping the const qualifier.

crypto/fipsmodule/service_indicator/service_indicator.c
nebeid: I agree with you, Dusan in keeping the condition as `!=` and removing the check in l.87.
Good observation and the calculation :)

crypto/fipsmodule/service_indicator/service_indicator.c
nebeid: It's just different from the style in all other files and I would personally prefer the no line after the function signature. Also here, some of the `if` statements have their first line followed by a blank lines and some don't.

crypto/fipsmodule/service_indicator/service_indicator.c
dkostic: done

crypto/fipsmodule/service_indicator/service_indicator.c
dkostic: ok, I removed the blank lines after function signatures and after if statements.

nebeid (approved):

samuel40791765 (approved):

PR #456, opened 2022-04-08
Add support for overriding OPENSSL_realloc

andrewhop:
### Issues:
Addresses CryptoAlg-953

### Description of changes: 
Customers can already control most heap memory allocation inside AWS-LC. `OPENSSL_realloc` uses functions that can be overwritten, but it is still opinionated about how realloc should work. This change adds support to define a custom realloc function. This also adds a test for using these weak symbols which is a little cumbersome from C++.

### Callouts
I had to make a new test executable since the way the weak symbols work it overrides it for the entire process. We may want to consider adding support to set these as callback functions like [s2n-tls](https://github.com/aws/s2n-tls/blob/main/utils/s2n_mem.c#L133-L149) which would be easier to test and debug.

### Testing:
Existing unit tests pass and new unit test works and is run by default.


crypto/mem.c
```diff
@@ -191,7 +192,9 @@ void *OPENSSL_realloc(void *orig_ptr, size_t new_size) {
   if (orig_ptr == NULL) {
     return OPENSSL_malloc(new_size);
   }
-
+  if (OPENSSL_memory_realloc != NULL) {
+    return OPENSSL_memory_realloc(orig_ptr, new_size);
```
dkostic, line 196: should we
```suggestion
    assert(OPENSSL_memory_free != NULL);
    assert(OPENSSL_memory_get_size != NULL);
    return OPENSSL_memory_realloc(orig_ptr, new_size);
```
as in `OPENSSL_malloc`?

andrewhop: I don't think it's necessary. I believe the reason OPENSSL_malloc checks for the other 2 is they are required. If a customer is taking control of alloc we can't use our free as it expects the size in the prefix. So If you want to control the memory you need free and size. OPENSSL_realloc can play nicely with our alloc or a custom one. The only people that might be interested in providing a realloc would do it to try to save on allocations (and take on a lot of risk and potential bugs) but might not care about alloc.

crypto/mem_test.cc
```diff
@@ -0,0 +1,101 @@
...
+
+  void *malloc_ptr = nullptr;
+  // Check setup conditions
+  ASSERT_EQ(starting_alloc, alloc_count);
+  ASSERT_EQ(starting_free, free_count);
+  ASSERT_EQ(starting_realloc, realloc_count);
+  ASSERT_EQ(0, size_count);
+  ASSERT_EQ(malloc_ptr, nullptr);
+
+  // Verify malloc calls OPENSSL_memory_alloc and doesn't do anything else
```
torben-hansen, line 61:
```suggestion
  // Verify malloc calls |OPENSSL_memory_alloc| and doesn't do anything else
```
I think this is somewhat standard comment style throughout.

crypto/mem_test.cc
```diff
@@ -0,0 +1,101 @@
...
+  void *realloc_ptr_1 = OPENSSL_realloc(null_realloc_ptr, test_size);
+  ASSERT_NE(realloc_ptr_1, nullptr);
+  ASSERT_EQ(starting_alloc + 2, alloc_count);
+  ASSERT_EQ(starting_free, free_count);
+  ASSERT_EQ(starting_realloc, realloc_count);
+  ASSERT_EQ(0, size_count);
+  ASSERT_EQ(nullptr, null_realloc_ptr);
+
+  // Check realloc with a not null pointer calls OPENSSL_memory_realloc
+  ASSERT_NE(realloc_ptr_1, nullptr);
```
torben-hansen, line 83: Checked this on line 75

crypto/mem_test.cc
```diff
@@ -0,0 +1,101 @@
...
+  test_size = 300;
+  void *realloc_ptr_2 = OPENSSL_realloc(malloc_ptr, test_size);
+  ASSERT_NE(realloc_ptr_2, realloc_ptr_1);
+  ASSERT_NE(realloc_ptr_2, malloc_ptr);
```
torben-hansen, line 87: Can't these two addresses be the same? Does the kernel _have_ to move the memory?

crypto/mem_test.cc
```diff
@@ -0,0 +1,101 @@
...
+  ASSERT_EQ(0, size_count);
+
+  // Check OPENSSL_memory_free
+  OPENSSL_free(realloc_ptr_1);
+  ASSERT_EQ(starting_alloc + 2, alloc_count);
+  ASSERT_EQ(starting_free + 1, free_count);
+  ASSERT_EQ(starting_realloc + 1, realloc_count);
+  ASSERT_EQ(0, size_count);
+  OPENSSL_free(realloc_ptr_2);
+}
```
torben-hansen, line 100: Call `OPENSSL_memory_get_size()` at some points as well?

crypto/mem_test.cc
```diff
@@ -0,0 +1,101 @@
...
+#endif
```
torben-hansen, line 101: Missing new-line above imo.

crypto/mem.c
andrewhop: That is a good point. So defining a custom realloc requires a custom malloc, but not vice versa.

crypto/CMakeLists.txt
```diff
@@ -535,10 +541,10 @@ if(BUILD_TESTING)
   # and handle it in urandom_test.cc
   if(JITTER_ENTROPY)
     target_compile_options(${RANDOM_TEST_EXEC} PUBLIC -DJITTER_ENTROPY)
+    target_compile_options(${MEM_TEST} PUBLIC -DJITTER_ENTROPY)
```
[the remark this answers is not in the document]
andrewhop: Correct, I was messing with the build earlier and ran into an issue missing Jitter but that's gone now.

crypto/mem_test.cc
```diff
@@ -0,0 +1,101 @@
...
+#include <gtest/gtest.h>
+#include <openssl/mem.h>
+
+int alloc_count = 0;
+size_t test_size = 0;
+int free_count = 0;
+int size_count = 0;
+int realloc_count = 0;
+
+extern "C" {
```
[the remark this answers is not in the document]
andrewhop: Yepp, that was a hold over from another experiment.

crypto/mem_test.cc
andrewhop: `OPENSSL_memory_get_size` is only used in `OPENSSL_realloc` if you don't define `OPENSSL_memory_realloc`. This test has a custom realloc so it never hits the code path that needs `OPENSSL_memory_get_size`. Do you think testing that path warrants another test program without realloc defined?

torben-hansen (approved):

PR #457, opened 2022-04-11
Add support for secp256k1 elliptic curve

dkostic:
### Issues:
CryptoAlg-1024

### Description of changes: 

The change adds support for secp256k1 elliptic curve (required by ACCP).
We use the generic |EC_GFp_mont_method| for the new curve.

The largest part of the change is adding tests to cover as much as
possible the potential use of the curve. We have added EC, ECDH, and
ECDSA tests. In all three cases we generate the test vectors by newly
added Go scripts. The standard Go library doesn't have support for
secp256k1, and moreover, we can't instantiate the curve on our own
because Go's implementation assumes a curve given by y^2 = x^3 + ax + b
has a = -3, which is not the case for secp256k1. We work around this
issue by using the most widely used secp256k1 Go implementation,
Ethereum's go-ethereum/crypto/secp256k1 module. Also, EVP Wycheproof
tests specific to secp256k1 were added.

A small issue with FIPS service indicator test for ECDH is also fixed.
The test assumed in one case that the indicator should always return
APPROVED, when in reality the indicator status should depened on which
elliptic curve is used in the protocol. For example, when a non FIPS
approved curve like secp256k1 is used, we expect the indicator to
return NOT_APPROVED.

### Call-outs:

### Testing:


torben-hansen (commented):
Nice.

I see some missing tests in `crypto/fipsmodule/ec/ec_test.cc`.

Also check out `crypto/evp_extra/evp_test.cc`. Could potentially also add some p256k1-specific tests there?

dkostic (commented):
The changes are ported from this feature branch: https://github.com/awslabs/aws-lc/tree/add_secp192r1_secp256k1. Since we don't need to support P-192 curve I ported only changes relevant to secp256k1.

crypto/fipsmodule/ecdsa/ecdsa_sign_tests.txt
```diff
@@ -2273,3 +2273,678 @@ Digest = fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
 K = 00938d2f6550a46fb07b058e6287f428f0ff12aa6732a666d4a6cf2dd7cd8023ca76d0ce4e16b62830d0ff9e2fab9987261f3f3ffe0749ff70950d91b897d57007b2
 R = 00ec0b91fa4386a8acdc0e46dd9c1d1775abbe0da8ead424aa4ace58e284a5be00e2c1ef95b6f4d861615564e1e7305656567f95275ce63b534420eae77ec37492c2
 S = 01e1099fb389db498ab4cf23b4f06a74b9326878ae3c76ea13832e50702b30fe8303093a59cc9a0995f1dfc15e6f7dabca8a2acaf03ec005447d29fb429a252064ec
+
```
nebeid, line 2276: Should there be a reference here to where the tests were generated from?

crypto/fipsmodule/ecdsa/ecdsa_test.cc
```diff
@@ -192,6 +192,7 @@ TEST(ECDSATest, BuiltinCurves) {
       { NID_secp384r1, "secp384r1" },
       { NID_secp521r1, "secp521r1" },
       { NID_secp160r1, "secp160r1" },
+      { NID_secp256k1, "secp256k1" },
```
nebeid, line 195: Nit: In the order below, secp256k1 is checked for before secp160r1.

crypto/fipsmodule/ecdsa/ecdsa_verify_tests.txt
```diff
@@ -2774,3 +2774,589 @@ Digest = 09ca7e4eaa6e8ae9c7d261167129184883644d07dfba7cbfbc4c8a2e08360d5b
 R = 00ffffffffffffffffffffffffffffffff80000003
 S = 0100000000000000000001f4c8f927aed3ca752254
 Invalid =
+
```
nebeid, line 2777: Same comment as in ecdsa_sign_tests.txt about a reference.

nebeid (commented):
Just curious: why `EC_curve_name2nid` and `EC_curve_nid2name` are no longer needed?

crypto/ecdh_extra/make_secp256k1_test_vectors.go
```diff
@@ -0,0 +1,80 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
dkostic: agreed

crypto/fipsmodule/ecdsa/ecdsa_test.cc
dkostic: Forgot to change this before I pushed the update. If you don't mind I'd let it be like this?

dkostic:
> I see some missing tests in crypto/fipsmodule/ec/ec_test.cc.
Also check out crypto/evp_extra/evp_test.cc. Could potentially also add some p256k1-specific tests there?

@torben-hansen  Added more tests both in `ec_test.cc` and `evp_test.cc`. Let me know if I missed anything.

> Just curious: why EC_curve_name2nid and EC_curve_nid2name are no longer needed?

@nebeid They were not needed in the first place, I realized that only now.

crypto/fipsmodule/ec/ec_test.cc
```diff
@@ -378,6 +378,48 @@ static const uint8_t kP521PublicKey_compressed_0x03[] = {
   0x9f, 0x5f, 0xb4, 0xf8, 0xe7, 0x7b
 };
 
+static const uint8_t ksecp256k1PublicKey_uncompressed_0x02[] = {
```
nebeid, line 381: How were the new tests generated?

crypto/fipsmodule/ecdsa/ecdsa_test.cc
nebeid: Sure

crypto/fipsmodule/ec/ec_test.cc
dkostic: I used Sage to generate a random point :)

nebeid (approved):

torben-hansen (approved):

crypto/fipsmodule/ec/ec_test.cc
dkostic: For posterity, the Sage code to generate a random point on secp256k1. The setup of the curve is taken from this very useful webpage (that covers many other curves as well): https://neuromancer.sk/std/secg/secp256k1#

```
p = 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f
K = GF(p)
a = K(0x0000000000000000000000000000000000000000000000000000000000000000)
b = K(0x0000000000000000000000000000000000000000000000000000000000000007)
E = EllipticCurve(K, (a, b))
G = E(0x79be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798, 0x483ada7726a3c4655da4fbfc0e1108a8fd17b448a68554199c47d08ffb10d4b8)
E.set_order(0xfffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141 * 0x1)

(x, y, z) = E.random_element()
```

PR #471, opened 2022-04-19
Fix benchmark criteria

DominicDams:
The benchmarking criteria was accidentally set to be a 15% improvement
as a failure criteria rather than a 15% degradation. This change fixes
that.


tests/ci/benchmark_framework/compare_results.py
```diff
@@ -66,7 +66,7 @@ def main():
     dfs = pd.concat([df1, df2], axis=1)
 
     # Filter out entries with a +15% regression
-    compared = np.where((1 - (df2_avg_time / df1_avg_time)) >= 0.15, df1.iloc[:, 0], np.nan)
+    compared = np.where(((df2_avg_time / df1_avg_time) - 1) >= 0.15, df1.iloc[:, 0], np.nan)
```
torben-hansen, line 69: Should we try `5%` just to see how sensitive it is?

samuel40791765 (approved):

tests/ci/benchmark_framework/compare_results.py
DominicDams: I'm fine with dialing it back if we really want to but I suspect that will be noisier than we want. That said, this isn't a required reviewer so we can always adjust it based on the noise generated

torben-hansen: It would be an experiment

torben-hansen (dismissed):

torben-hansen (approved):

samuel40791765 (approved):

PR #477, opened 2022-04-27
Add legacy build test to our CI

samuel40791765:
### Issues:
Resolves `CryptoAlg-767`

### Description of changes: 
We recently added support to maintain compatibility with gcc 4.1.2, but we don't have a CI build that uses that specific compiler version. This new CI support will help us find if changes break the gcc 4.1.2 build before we find out that the build broke when doing internal imports.

### Call-outs:
It's nearly impossible to find a rhel5 docker image, so I've leaned into the closest OS environment I can find. Ubuntu 7.04 supports `gcc4.1.2`, but the gnu assembly does not support SSE4 instructions, while rhel5's gnu assembly does. Ubuntu 7.04 constantly fails without the SSE4 assembly support. 
Ubuntu 7.10 uses `gcc4.1.3` with gnu assembly 2.18, but it is the closest thing to rhel5 I can find at the moment. When testing on commits before we added rhel5 support, the Ubuntu 7.10 environment produces identical build errors to what rhel5 produces. 
More details can be found in  `CryptoAlg-767`

### Testing:
Locally tested. 
CI test: https://github.com/samuel40791765/aws-lc/pull/16 (android-device-farm fails because the PR points to team account's device farm projects)


tests/ci/docker_images/linux-x86/build_legacy_images.sh
```diff
@@ -0,0 +1,26 @@
...
+
+if [ -n "$1" ]; then
+  docker_name="$1"
+else
+  docker_name='ubuntu-7.10_gcc-4.1x'
+fi
+
+# `ubuntu-7.10-gutsy_gcc-4.1x` was rebuilt from an iso image
+# from: https://old-releases.ubuntu.com/releases/gutsy/
+# The docker tar file is checked into our team account.
```
bryce-shang, line 13: In which s3 bucket?

tests/ci/docker_images/linux-x86/build_legacy_images.sh
```diff
@@ -0,0 +1,26 @@
...
+# Must be logged into team account to access docker tar file.
+mkdir -p ${docker_name}/dependencies
+cd ${docker_name}/dependencies
+aws s3 cp s3://ubuntu-7.10-gutsy-gcc-4.1x/ubuntu-7.10-gutsy_gcc-4.1x.tar ${docker_name}.tar
+docker load --input ${docker_name}.tar
+# The wget version that comes with this docker image is too old to
+# download the cmake version from source, so we predownload the 
+# dependencies outside of the image and pull it in the container.
+wget -O cmake-2.8.12.tar.gz https://cmake.org/files/v2.8/cmake-2.8.12.tar.gz
```
bryce-shang, line 22: Nit: why use `2.8.12` instead of other higher version CMake?

```
cmake_minimum_required(VERSION 2.8.12)
```

tests/ci/run_legacy_build.sh
```diff
@@ -0,0 +1,19 @@
...
+# `GNU assembler 2.18` resembles rhel5's binutils the most. SSE4
+# assembly instructions are not supported in ubuntu 7.04, while rhel5
+# and ubuntu 7.10 do.
+echo "Testing gcc 4.1.2 / 4.1.3 support."
+
+# Go is disabled because go tries to fetch from the server when building,
+# but the certificates on the ancient ubuntu are expired and the ancient 
+# tools are insufficient to update the certificates.
+# The build flags used are the same as our ancient internal build.
+run_build  -DBUILD_TESTING="OFF" -DBUILD_LIBSSL="OFF" -DCMAKE_BUILD_TYPE="Debug" "-DMY_ASSEMBLER_IS_TOO_OLD_FOR_AVX=ON -std=gnu99 -fgnu89-inline"
```
bryce-shang, line 19: The docker image provides c++ compiler, should `-DBUILD_TESTING="OFF"` be removed?
Some tests may fail. If that happens, we can have separated SIM to track and enable the build flag.

tests/ci/cdk/cdk/codebuild/github_ci_linux_x86_omnibus.yaml
```diff
@@ -355,3 +355,11 @@ batch:
         privileged-mode: true
         compute-type: BUILD_GENERAL1_SMALL
         image: ECR_REPO_PLACEHOLDER:ubuntu-20.04_clang-9x_latest
+
+    - identifier: ubuntu710_gcc4.1x_x86_64_build
```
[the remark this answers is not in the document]
samuel40791765: Got caught that this hadn't been tested yet, just tested on my fork and changed this :)

tests/ci/docker_images/linux-x86/build_legacy_images.sh
samuel40791765: I've gone with a different method now. The new steps are what I originally did to create the image. Now the script downloads the iso image and builds a docker image from that. This should be more sustainable than fixating the `build-linux-docker-img` Codebuild project to a specific s3 bucket.
I've tested the `build-linux-img` cdk command on my account and the new steps works.

tests/ci/docker_images/linux-x86/build_legacy_images.sh
samuel40791765: My thought process was that this was the oldest build that we support, so I went with the oldest cmake version that we support to test that along with it. I don't have a strong preference though, I can update it if we want to test another version.

tests/ci/run_legacy_build.sh
samuel40791765: These build flags are taken from our internal build. I've asked around and the main reason why it's off is because gcc 4.1.2 does not support c++11 and all our tests (gtest) use c++11 features. I've also tried removing the c++ compiler from the docker image, but the build still complains about needing c++ support.

tests/ci/docker_images/linux-x86/build_legacy_images.sh
bryce-shang: The internal build uses oldest compiler but `CMake 3.9.x`. If `CMake 2.8.12` is used here, the Docker image will get changed when we need to bump the number.

bryce-shang (dismissed):

bryce-shang (dismissed):

tests/ci/docker_images/linux-x86/ubuntu-7.10_gcc-4.1x/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+ENV CC=gcc-4.1
+ENV CXX=g++-4.1
+
+# Pull cmake as an external source since the wget version
+# on this image is too old to access the cmake repo.
+COPY dependencies/cmake-3.9.6.tar.gz /tmp/cmake-3.9.6.tar.gz
+RUN cd /tmp && \
+    tar -xvf cmake-3.9.6.tar.gz && \
+    cd cmake-3.9.6 && \
+    ./configure && make && make install
```
andrewhop, line 26:
```suggestion
    ./configure && make && make install && \
    rm -rf /tmp/cmake-3.9.6.tar.gz && \
    rm -rf cmake-3.9.6
```
Without looking at the tmp folder I think this is everything you want to remove.

tests/ci/run_legacy_build.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -exo pipefail
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+source tests/ci/common_posix_setup.sh
+
+
```
andrewhop, line 8:
```suggestion
```

tests/ci/run_legacy_build.sh
andrewhop, line 19:
```suggestion
run_build  -DBUILD_TESTING=OFF -DBUILD_LIBSSL=OFF -DCMAKE_BUILD_TYPE=Debug -DMY_ASSEMBLER_IS_TOO_OLD_FOR_AVX=ON -std=gnu99 -fgnu89-inline
```
You don't need all this extra quoting, I think that is a carry over from the other builds where on/off are bash variables we don't want to split on spaces.

tests/ci/docker_images/linux-x86/ubuntu-7.10_gcc-4.1x/Dockerfile
samuel40791765: removed all the contents of `/tmp/*` like we do with the rest of our docker images

bryce-shang (approved):

PR #486, opened 2022-05-10
Small fixes to build with mysql

samuel40791765:
### Issues:
Resolves `CryptoAlg-1076`

### Description of changes: 
These changes are necessary to build with [mySql8.0](https://github.com/mysql/mysql-server/tree/8.0).

* Update `BIO_get_mem_data` to take `const char**`. Should not break anything because `char` is convertible to `const char**`.
* Add `DH_get_2048_256` (https://www.openssl.org/docs/man1.1.1/man3/DH_get_2048_256.html)
   *  This function has been deprecated in OpenSSL 3.0 and there does not seem to be any functionality anymore in OpenSSL for this. 
   * Using a no-op function for now since mySQL still requires it. There doesn't seem to be anything done directly with the values, but we'll have to verify this is the case when we run against mysql tests.
   * https://github.com/mysql/mysql-server/blob/8.0/router/src/harness/src/tls_server_context.cc#L172-L174
 * Add `DH_bits` function that points to `BN_num_bits`

### Call-outs:
N/A

### Testing:
Tested by commenting out certain X509 and TLS1.3 related functions in MySql and building against AWS-LC. Rest of the build passes through.


crypto/dh_extra/dh_asn1.c
```diff
@@ -158,3 +158,5 @@ int i2d_DHparams(const DH *in, uint8_t **outp) {
   }
   return CBB_finish_i2d(&cbb, outp);
 }
+
+DH *DH_get_2048_256(void) { return NULL; }
```
andrewhop, line 162: I don't think this will work, there code calls [uniqueptr reset](https://www.cplusplus.com/reference/memory/unique_ptr/reset/) with the value from `DH_get_2048_256` which is now null and sets it to dh2048. Which means they will pass null to `SSL_CTX_set_tmp_dh` on line 197.

crypto/fipsmodule/dh/dh.c
```diff
@@ -415,6 +415,8 @@ int DH_compute_key_hashed(DH *dh, uint8_t *out, size_t *out_len,
 
 int DH_size(const DH *dh) { return BN_num_bytes(dh->p); }
 
+int DH_bits(const DH *dh) { return DH_num_bits(dh); }
```
andrewhop, line 418: Is this always a safe cast, this could cast the number bits to a negative size right? I realize this is very unlikely since no one should (or even can?) make a 2^31 + 1 sized DH number.

andrewhop: I just noticed this is following the same pattern as DH_size.

crypto/dh_extra/dh_asn1.c
samuel40791765: Added the RFC value function

include/openssl/dh.h
```diff
@@ -301,6 +306,13 @@ OPENSSL_EXPORT int i2d_DHparams(const DH *in, unsigned char **outp);
 OPENSSL_EXPORT int DH_compute_key(uint8_t *out, const BIGNUM *peers_key,
                                   DH *dh);
 
+// Standard parameters. These parameters are taken from RFC 5114.
+// This function returns a new DH object with standard parameters. It returns
+// NULL on allocation failure.
+// 
+// WARNING: Don't use this, this function has been deprecated and is only 
+// provided because to support MySQL.
+OPENSSL_EXPORT DH *DH_get_2048_256(void);
```
andrewhop, line 315: Can you move this down to the deprecated section below line 917.
```suggestion
// Standard parameters. These parameters are taken from RFC 5114.
// This function returns a new DH object with standard parameters. It returns
// NULL on allocation failure.
// 
// WARNING: deprecated: Use .......
OPENSSL_EXPORT DH *DH_get_2048_256(void);
```
What is the alternative we should recommend customers to use? That is our standard pattern in header giles.

include/openssl/dh.h
```diff
@@ -301,6 +306,13 @@ OPENSSL_EXPORT int i2d_DHparams(const DH *in, unsigned char **outp);
 OPENSSL_EXPORT int DH_compute_key(uint8_t *out, const BIGNUM *peers_key,
                                   DH *dh);
 
+// Standard parameters. These parameters are taken from RFC 5114.
+// This function returns a new DH object with standard parameters. It returns
+// NULL on allocation failure.
+// 
+// WARNING: Don't use this, this function has been deprecated and is only 
+// provided because to support MySQL.
```
torben-hansen, line 314: nitty: "because to support"

torben-hansen (approved):

andrewhop (approved):

PR #490, opened 2022-05-13
CMAKE_ASM_COMPILER_ID not always set

justsmth:
### Description of changes: 
Clang-7's assembler doesn't support the generation of debug symbols. Our current cmake logic depends on the value of `CMAKE_ASM_COMPILER_ID` to indicate whether Clang's assembler is being used. However, this variable is not always set. When not set, `-Wa,-g` gets added to `CMAKE_ASM_FLAGS` causing the build to fail.

### Call-outs:
If it's unable to determine which assembler is being used, the logic assumes that debug symbols are supported by the assembler.

### Testing:
Successfully completed a debug build using clang-7:
```
cmake . -DBUILD_SHARED_LIBS=TRUE "-DCMAKE_C_COMPILER=clang-7" "-DCMAKE_CXX_COMPILER=clang++-7" "-DCMAKE_ASM_COMPILER=clang-7" "-DFIPS=1" "-DCMAKE_INSTALL_PREFIX=/tmp/fips-build"
```

```
cmake --build . --target ssl --config Debug
```

### DIFF
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -33,7 +33,7 @@ if(NOT OPENSSL_NO_ASM)
     set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,--noexecstack")
 
     # Clang's integerated assembler does not support debug symbols.
-    if (CMAKE_ASM_COMPILER_ID MATCHES "Clang" || CMAKE_ASM_COMPILER MATCHES "clang")
+    if (CMAKE_ASM_COMPILER_ID MATCHES "Clang" OR CMAKE_ASM_COMPILER MATCHES "clang")
       message(STATUS "Disabling debug symbols for Clang internal assembler")
     else()
       set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,-g")
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -33,8 +33,16 @@ if(NOT OPENSSL_NO_ASM)
     set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,--noexecstack")
 
     # Clang's integerated assembler does not support debug symbols.
-    if(NOT CMAKE_ASM_COMPILER_ID MATCHES "Clang")
-      set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,-g")
+    if(CMAKE_ASM_COMPILER_ID)
+      if(NOT CMAKE_ASM_COMPILER_ID MATCHES "Clang")
+        set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,-g")
+      endif()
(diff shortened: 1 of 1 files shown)


crypto/CMakeLists.txt
```diff
@@ -33,8 +33,16 @@ if(NOT OPENSSL_NO_ASM)
     set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,--noexecstack")
 
     # Clang's integerated assembler does not support debug symbols.
-    if(NOT CMAKE_ASM_COMPILER_ID MATCHES "Clang")
-      set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,-g")
+    if(CMAKE_ASM_COMPILER_ID)
```
torben-hansen, line 36: Proposal for simplifying logic. In pseudo-code:
```
if (CMAKE_ASM_COMPILER_ID MATCHES "Clang" || CMAKE_ASM_COMPILER MATCHES "clang")
  message(STATUS "Disabling debug symbols for Clang internal assembler)
else ()
 set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,-g")
endif()
``

Comment (torben-hansen) [reply crypto/CMakeLists.txt:36]: This logic also doesn't capture when user specifies an external assembler - but this was not captured previously as well.

File: crypto/CMakeLists.txt [hunk 2]
```diff
@@ -33,8 +33,16 @@ if(NOT OPENSSL_NO_ASM)
...
+      if(NOT CMAKE_ASM_COMPILER_ID MATCHES "Clang")
+        set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,-g")
+      endif()
+    elseif(CMAKE_ASM_COMPILER)
+      if(NOT CMAKE_ASM_COMPILER MATCHES "clang")
```
bryce-shang, line 41: Nit: `Clang` vs `clang`? Maybe the `MATCHES` can ignore the uppper and lowercase difference?

crypto/CMakeLists.txt
```diff
@@ -33,8 +33,16 @@ if(NOT OPENSSL_NO_ASM)
...
+        set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,-g")
+      endif()
+    else()
+        set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wa,-g")
```
bryce-shang, line 45: Nit: indentation?

crypto/CMakeLists.txt
justsmth: The command `clang` is lower case, the value in `CMAKE_ASM_COMPILER_ID` is upper-case.

crypto/CMakeLists.txt
justsmth: Thanks. That is much better.

torben-hansen (approved):

bryce-shang (approved):

PR #491, opened 2022-05-17
Fix prefix build

torben-hansen:
### Description of changes: 

https://github.com/awslabs/aws-lc/pull/368 broke the prefix build. For example, the symbol list will contain the public symbol `MD5`. But then the object type [expanded here](https://github.com/awslabs/aws-lc/blob/e14e153cda6c0921a10209c78762de37707d48c4/crypto/fipsmodule/hmac/hmac.c#L90) will be
`<prefix>_MD5_CTX`. But the prefixing tool will only match exact, so the [corresponding declaration](https://github.com/awslabs/aws-lc/blob/e14e153cda6c0921a10209c78762de37707d48c4/include/openssl/base.h#L441) is not prefixed. 

The result are undefined identifiers.

### Call-outs:

Need to be more specific and avoid general tokens e.g. `MD5` to avoided them being prefixed. Unfortunately, this involves moving away from the more general `MD_TRAMPOLINES` to using only the explicit version and making it even more explicit.

We need to add the prefix build to the CI...

### Testing:

@justsmth verified that this fixes the prefix build

### DIFF
--- a/crypto/fipsmodule/hmac/hmac.c
+++ b/crypto/fipsmodule/hmac/hmac.c
@@ -104,8 +104,6 @@ struct hmac_methods_st {
       sizeof(HASH_CTX) <= sizeof(union md_ctx_union),                         \
       HASH_NAME##_has_overlarge_context_t)
 
-#define MD_TRAMPOLINES(HASH_NAME) MD_TRAMPOLINES_EXPLICIT(HASH_NAME, HASH_NAME)
-
 // The maximum number of HMAC implementations
 #define HMAC_METHOD_MAX 7
 
--- a/crypto/fipsmodule/hmac/hmac.c
+++ b/crypto/fipsmodule/hmac/hmac.c
@@ -81,27 +81,27 @@ struct hmac_methods_st {
 // while the destination functions have specific pointer types for the relevant contexts.
 //
 // This also includes hash-specific static assertions as they can be added.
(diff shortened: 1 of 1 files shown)


crypto/fipsmodule/hmac/hmac.c
```diff
@@ -81,41 +81,39 @@ struct hmac_methods_st {
...
   }                                                                           \
-  OPENSSL_STATIC_ASSERT(CTX_PREFIX##_CBLOCK % 8 == 0,                         \
+  OPENSSL_STATIC_ASSERT(HASH_CBLOCK% 8 == 0,                                  \
                         HASH_NAME##_has_blocksize_not_divisible_by_eight_t)   \
-  OPENSSL_STATIC_ASSERT(CTX_PREFIX##_CBLOCK <= EVP_MAX_MD_BLOCK_SIZE,         \
+  OPENSSL_STATIC_ASSERT(HASH_CBLOCK <= EVP_MAX_MD_BLOCK_SIZE,                 \
                         HASH_NAME##_has_overlarge_blocksize_t)                \
   OPENSSL_STATIC_ASSERT(                                                      \
-      sizeof(CTX_PREFIX##_CTX) <= sizeof(union md_ctx_union),                 \
+      sizeof(HASH_CTX) <= sizeof(union md_ctx_union),                         \
```
justsmth, line 104: Nitpick -- there appears to be trailing whitespace on the changes lines.

justsmth (approved):
I verified that a symbol-prefix build succeeds with this change.

PR #493, opened 2022-05-18
Avoid externalizing the s2n-bignum symbols in shared library.

justsmth:
### Description of changes: 
The shared library produced by the current build process externalizes the s2n-bignum symbols (`bignum_*`).

### Call-outs:
The s2n-bignum symbols are still exposed from our *static* library.

### Testing:
I built the shared library on Mac (x86) and Linux (x86 and aarch64). Verified that the s2n-bignum symbols were not being externalized.


torben-hansen:
This looks like an upstream change is better? It breaks as soon as we include more s2n-bignum code.

Does this work with Git subtree? Windows?

justsmth:
I'll submit this PR upstream.

PR #502, opened 2022-05-31
Update `EVP_MD_CTX_set_pkey_ctx` comments about responsibility

samuel40791765:
### Issues:
Addresses `P64861145`

### Description of changes: 
Some of the comments in `EVP_MD_CTX_set_pkey_ctx` may have been unclear on who's responsibility it is to free the `pctx` pointer. This tries to clarify that the caller is responsible.  

### Call-outs:
N/A

### Testing:
Tests already exist. When doing the incorrect behavior, the test will fail against address sanitizer tests due to a leaked pointer.

### DIFF
--- a/include/openssl/digest.h
+++ b/include/openssl/digest.h
@@ -317,9 +317,10 @@ OPENSSL_EXPORT int EVP_MD_nid(const EVP_MD *md);
 // responsible for freeing |pctx|. Calling |EVP_MD_CTX_cleanup| will not free
 // |pctx|.
 //
-// A NULL |pctx| pointer is also allowed to clear the |EVP_PKEY_CTX| assigned to
-// |ctx|. However, even when doing so, the caller is still responsible for
-// freeing the |pctx| pointer that had originally been associated.
+// A NULL |pctx| pointer is also allowed to set the |EVP_PKEY_CTX| reference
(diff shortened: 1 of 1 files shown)


include/openssl/digest.h
```diff
@@ -317,9 +317,9 @@ OPENSSL_EXPORT int EVP_MD_nid(const EVP_MD *md);
 // responsible for freeing |pctx|. Calling |EVP_MD_CTX_cleanup| will not free
 // |pctx|.
 //
-// The caller is responsible for the memory associated to |ctx|'s |EVP_PKEY_CTX|
-// reference (the passed in |pctx| parameter), until |EVP_MD_CTX_set_pkey_ctx|
-// is called again with |pctx| = NULL.
+// A NULL |pctx| pointer is also allowed to clear the |EVP_PKEY_CTX| assigned to
+// |ctx|. However, even when doing so, the caller is still responsible for
```
nebeid, line 321:
```suggestion
// A NULL |pctx| pointer is also allowed to set the |EVP_PKEY_CTX| reference inside
// |ctx| to NULL. However, even when doing so, the caller is still responsible for
```

samuel40791765: Applied the new wording

torben-hansen (approved):

PR #508, opened 2022-06-08
Fix docker build - Use Ubuntu 22.04 with gcc-11x

justsmth:
### Description of changes: 
Docker build was failing for Ubuntu 20.04 with gcc-11x.  Updated to use Ubuntu 22.04 for gcc-11x.

### Testing:
I successfully built the docker images locally.


bryce-shang (approved):

samuel40791765 (approved):
Just a reminder that we'll have to update the omnibus names and placeholders to use the new docker image in the ECR. That would need an actual CI update to propagate the changes, so we could look to do in that in a separate PR.
* https://github.com/awslabs/aws-lc/blob/main/tests/ci/cdk/cdk/codebuild/github_ci_linux_x86_omnibus.yaml#L59-L65
* https://github.com/awslabs/aws-lc/blob/main/tests/ci/cdk/cdk/codebuild/github_ci_linux_arm_omnibus.yaml#L25-L31

PR #509, opened 2022-06-08
Fix Unused statements for fiat

DominicDams:
We have `OPENSSL_UNUSED` statements for our fiat header files in order
to prevent errors for functions that we don't use. There was a mistake
that used the 64-bit function definitions for the 32-bit header file. On
some platforms this would cause a compilation error.

### Issues:
Potential fix for https://github.com/awslabs/aws-lc/issues/487


samuel40791765 (approved):

DominicDams:
We see a small regression in `TrustToken-Exp2PMB-Batch1 finish_redemption` but there is no change here that would cause that. Ignoring as spurious.

torben-hansen (approved):

PR #514, opened 2022-06-13
Ci prefix build

justsmth:
### Description of changes: 
Add prefix builds to our CI.

### Call-outs:
Also fixed a few related issues:
* CDK Readme showed commands that were incomplete, and wouldn't even work.
* Move Ubuntu's gcc-11x build to Ubuntu 22.04 to avoid Ubuntu glibc/kernel install issue: https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1962225.
* Patched Ubuntu-7.10/gcc-4.1x docker image to avoid same glibc/kernel install issue.
* Explicitly invokes the `boringssl_prefix_symbols` target prior to prefix build.  AL2/CMake2 hosts seemed to ignore this target dependency when building target `ssl`
* With certain build flags (like `OPENSSL_NO_ASM=1`) some object files would be w/o a "Symbols" section causing the `read_symbols.go` script to fail. Updated script to ignore object files not containing symbols.

### Testing:
* I synthesized the docker images and CDK into my personal account, and successfully tested the CI for both x86 and ARM.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,5 @@ util/bot/yasm-win32.exe
 test_build_dir/
 cmake-build-debug/
 symbols.txt
-/.idea
 
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,9 @@ util/bot/sde-win32.tar.xz
 util/bot/win_toolchain.json
 util/bot/yasm-win32.exe
 
-test_build_dir
+test_build_dir/
+cmake-build-debug/
+symbols.txt
+/.idea
 
 .DS_Store
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,7 +144,7 @@ if(BORINGSSL_PREFIX AND BORINGSSL_PREFIX_SYMBOLS AND GO_EXECUTABLE)
            symbol_prefix_include/boringssl_prefix_symbols_asm.h
            symbol_prefix_include/boringssl_prefix_symbols_nasm.inc
     COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include
-    COMMAND ${GO_EXECUTABLE} run ${CMAKE_CURRENT_SOURCE_DIR}/util/make_prefix_headers.go -out ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include ${BORINGSSL_PREFIX_SYMBOLS}
+    COMMAND ${GO_EXECUTABLE} run ${CMAKE_CURRENT_SOURCE_DIR}/util/make_prefix_headers.go -out ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include ${BORINGSSL_PREFIX_SYMBOLS_PATH}
     DEPENDS util/make_prefix_headers.go
             ${BORINGSSL_PREFIX_SYMBOLS_PATH})
 
--- a/tests/ci/cdk/cdk/codebuild/github_ci_linux_arm_omnibus.yaml
+++ b/tests/ci/cdk/cdk/codebuild/github_ci_linux_arm_omnibus.yaml
@@ -148,8 +148,24 @@ batch:
         compute-type: BUILD_GENERAL1_LARGE
         image: ECR_REPO_PLACEHOLDER:ubuntu-20.04_clang-10x_latest
 
+    - identifier: ubuntu2004_clang10x_aarch_prefix
+      buildspec: ./tests/ci/codebuild/linux-aarch/run_prefix_tests.yml
+      env:
+        type: ARM_CONTAINER
+        privileged-mode: false
+        compute-type: BUILD_GENERAL1_LARGE
+        image: ECR_REPO_PLACEHOLDER:ubuntu-20.04_clang-10x_latest
+
     - identifier: amazonlinux2_gcc7x_aarch
       buildspec: ./tests/ci/codebuild/linux-aarch/run_posix_tests.yml
+      env:
+        type: ARM_CONTAINER
+        privileged-mode: false
+        compute-type: BUILD_GENERAL1_LARGE
+        image: ECR_REPO_PLACEHOLDER:amazonlinux-2_gcc-7x_latest
+
+    - identifier: amazonlinux2_gcc7x_aarch_prefix
+      buildspec: ./tests/ci/codebuild/linux-aarch/run_prefix_tests.yml
       env:
         type: ARM_CONTAINER
         privileged-mode: false
--- a/tests/ci/common_posix_setup.sh
+++ b/tests/ci/common_posix_setup.sh
@@ -85,16 +85,22 @@ function generate_symbols_file {
 
 
 function verify_symbols_prefixed {
-  nm -g "$BUILD_ROOT"/crypto/libcrypto.a | grep "aws_lc_1_1_0"
-  nm -g "$BUILD_ROOT"/ssl/libssl.a | grep "aws_lc_1_1_0"
+  go run "$SRC_ROOT"/util/read_symbols.go -out "$BUILD_ROOT"/symbols_final_crypto.txt "$BUILD_ROOT"/crypto/libcrypto.a
+  go run "$SRC_ROOT"/util/read_symbols.go -out "$BUILD_ROOT"/symbols_final_ssl.txt "$BUILD_ROOT"/ssl/libssl.a
+  cat "$BUILD_ROOT"/symbols_final_crypto.txt  "$BUILD_ROOT"/symbols_final_ssl.txt | grep -v -e '^_\?bignum' >  "$SRC_ROOT"/symbols_final.txt
+  if [ $(grep -c -v ${CUSTOM_PREFIX}  "$SRC_ROOT"/symbols_final.txt) -ne 0 ]; then
+    echo "Symbol(s) missing prefix!"
+    exit 1
+  fi
 }
 
 
 function build_prefix_and_test {
+  CUSTOM_PREFIX=aws_lc_1_1_0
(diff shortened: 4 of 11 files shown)


.gitignore
```diff
@@ -31,6 +31,9 @@ util/bot/sde-win32.tar.xz
 util/bot/win_toolchain.json
 util/bot/yasm-win32.exe
 
-test_build_dir
+test_build_dir/
+cmake-build-debug/
+symbols.txt
+/.idea
```
bryce-shang, line 37: Nit: keep these in the global ignore `~/.gitignore`?

CMakeLists.txt
```diff
@@ -144,7 +144,7 @@ if(BORINGSSL_PREFIX AND BORINGSSL_PREFIX_SYMBOLS AND GO_EXECUTABLE)
            symbol_prefix_include/boringssl_prefix_symbols_asm.h
            symbol_prefix_include/boringssl_prefix_symbols_nasm.inc
     COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include
-    COMMAND ${GO_EXECUTABLE} run ${CMAKE_CURRENT_SOURCE_DIR}/util/make_prefix_headers.go -out ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include ${BORINGSSL_PREFIX_SYMBOLS}
+    COMMAND ${GO_EXECUTABLE} run ${CMAKE_CURRENT_SOURCE_DIR}/util/make_prefix_headers.go -out ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include ${BORINGSSL_PREFIX_SYMBOLS_PATH}
```
bryce-shang, line 147: Just a note.

```
  if(IS_ABSOLUTE ${BORINGSSL_PREFIX_SYMBOLS})
    set(BORINGSSL_PREFIX_SYMBOLS_PATH ${BORINGSSL_PREFIX_SYMBOLS})
  else()
    set(BORINGSSL_PREFIX_SYMBOLS_PATH ${CMAKE_BINARY_DIR}/${BORINGSSL_PREFIX_SYMBOLS})
  endif()
```

tests/ci/cdk/cdk/codebuild/github_ci_linux_arm_omnibus.yaml
```diff
@@ -148,6 +148,14 @@ batch:
         compute-type: BUILD_GENERAL1_LARGE
         image: ECR_REPO_PLACEHOLDER:ubuntu-20.04_clang-10x_latest
 
+    - identifier: ubuntu2004_clang10x_aarch_prefix
+      buildspec: ./tests/ci/codebuild/linux-aarch/run_prefix_tests.yml
```
bryce-shang, line 152: Nit: maybe add a comment with a rust pkg/repo link(which this dimension try to protect)?

tests/ci/common_posix_setup.sh
```diff
@@ -61,6 +68,38 @@ function build_and_test {
...
+function generate_symbols_file {
+  # read_symbols.go currently only support static libraries
+  if [ ! -f  "$BUILD_ROOT"/crypto/libcrypto.a ]; then
+    echo "Static library not found: ${BUILD_ROOT}/crypto/libcrypto.a"
+    print_system_and_dependency_information
+    exit 1
+  fi
+
+  go run "$SRC_ROOT"/util/read_symbols.go -out "$BUILD_ROOT"/symbols_crypto.txt "$BUILD_ROOT"/crypto/libcrypto.a
+  go run "$SRC_ROOT"/util/read_symbols.go -out "$BUILD_ROOT"/symbols_ssl.txt "$BUILD_ROOT"/ssl/libssl.a
```
bryce-shang, line 80: Nit: Is `libdecrepit.a` needed?

tests/ci/common_posix_setup.sh
```diff
@@ -61,6 +68,38 @@ function build_and_test {
...
+
+  # The $BUILD_ROOT gets deleted on each run. symbols.txt must be placed elsewhere.
+  cat "$BUILD_ROOT"/symbols_crypto.txt  "$BUILD_ROOT"/symbols_ssl.txt | grep -v -e '^_\?bignum' >  "$SRC_ROOT"/symbols.txt
+}
+
+
+function verify_symbols_prefixed {
+  nm -g "$BUILD_ROOT"/crypto/libcrypto.a | grep "aws_lc_1_1_0"
+  nm -g "$BUILD_ROOT"/ssl/libssl.a | grep "aws_lc_1_1_0"
+}
```
bryce-shang, line 90: Nit: this verification may only indicate the `aws_lc_1_1_0` exists. May we can have a `check_symbols.go` by extending the code of `read_symbols.go` to ensure each symbol this prefixed? My comment is a little over and should not block this PR.

tests/ci/run_prefix_tests.sh
```diff
@@ -0,0 +1,20 @@
...
+build_prefix_and_test
+
+echo "Testing a prefix build of AWS-LC in release mode."
+build_prefix_and_test -DCMAKE_BUILD_TYPE=Release
+
+echo "Testing a prefix build of AWS-LC small compilation."
+build_prefix_and_test -DOPENSSL_SMALL=1 -DCMAKE_BUILD_TYPE=Release
+
+echo "Testing a prefix build of AWS-LC in no asm mode."
+build_prefix_and_test -DOPENSSL_NO_ASM=1 -DCMAKE_BUILD_TYPE=Release
```
bryce-shang, line 20: Nit: Add one more dimension `-DASAN=1`? `ASAN`(`run_posix_sanitizer.sh`) maybe the most helpful dimension IMO. However, with that enabled, the ssl runner tests(8k) are very slow on arm when ASAN is enabled.

.gitignore
justsmth: I'll move`.idea` to my global `~/.gitignore`.  The other 3 lines are very specific to the artifacts created by this projects CI build scripts. Any objections to keeping those 3 lines?

tests/ci/cdk/cdk/codebuild/github_ci_linux_arm_omnibus.yaml
justsmth: I don't think Rust is directly relevant --  we want to protect the prefix build regardless of Rust.  

Also, the Rust build utilizing the prefix build is still just a proof-of-concept, so there isn't really any link to provide that would be viable long-term.

tests/ci/common_posix_setup.sh
justsmth: It's not currently supported, but it's worth considering. We should revisit supporting the prefixing of symbols in `libdecrepit.a` once we make decisions regarding how we want to handle this library long-term.

tests/ci/common_posix_setup.sh
justsmth: We could probably just re-use `read_symbols.go` to verify this.  I'll try updating the script to do that instead.

bryce-shang: I just point out this but I actually don't know how the verification can be enhanced.

util/read_symbols.go
```diff
@@ -175,17 +175,19 @@ func listSymbolsELF(contents []byte) ([]string, error) {
...
+	if err == nil {
+		for _, sym := range syms {
+			// Only include exported, defined symbols
+			if elf.ST_BIND(sym.Info) != elf.STB_LOCAL && sym.Section != elf.SHN_UNDEF {
+				names = append(names, sym.Name)
+			}
 		}
+	} else if err != elf.ErrNoSymbols {
+	// When `OPENSSL_NO_ASM` build flag is set, some assembly files will produce object files w/o a Symbols section
+		return nil, err
```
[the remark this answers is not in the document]
justsmth: I'll increase the indent for the comment.

tests/ci/common_posix_setup.sh
samuel40791765: I was going to ask something similar to this. Is there a way to verify that the symbols have been correctly prefixed? Maybe a script that checks if the "to-prefix" symbol list generated in `symbols_crypto.txt` and `symbols_ssl.txt` align with the prefixed symbols in `libcrypto.a` and `libssl.a`?

samuel40791765 (commented):
Just a question about `verify_symbols_prefixed` verification methods

util/read_symbols.go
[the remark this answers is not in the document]
justsmth: I'm not sure that that's the *only* place it occurs, but that is the file that the process broke on.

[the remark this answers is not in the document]
bryce-shang: should the fix be scoped? the current fix opens a case where no symbol in all files is allowed?

tests/ci/run_prefix_tests.sh
justsmth: I think we should consider a (prefix + ASAN) build test for a subsequent PR.

tests/ci/common_posix_setup.sh
```diff
@@ -61,6 +68,38 @@ function build_and_test {
...
+
+
+function build_prefix_and_test {
+  run_build "$@"
+  generate_symbols_file
+  PREBUILD_CUSTOM_TARGET="boringssl_prefix_symbols"
```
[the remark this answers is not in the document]
justsmth: The need for `PREBUILD_CUSTOM_TARGET` is related to Cmake2 not recognizing that the `global_target` depends on `boringssl_prefix_symbols`.  (Is that what you're asking about?) This is not needed for platforms using Cmake3.

I tried making changes to the build script, but couldn't find a way to get Cmake2 to automatically build the `boringssl_prefix_symbols` target.

samuel40791765 (approved):

PR #532, opened 2022-06-27
Adjusted the sensativity of the benchmarking CI

DominicDams:
The benchmarking CI has been giving spurious failures frequently because
the sensitivity was set too high. We expected to have to make
adjustments to this, and it seems like 10% was too low a cutoff for
detecting regressions. Adjusting that number to 15% so we will see fewer
spurious failures.


torben-hansen (approved):
That the benchmark dimension fails in this PR, is just a little bit hilarious.

bryce-shang (approved):
Nit: do we want higher threshold?

PR #536, opened 2022-06-28
Add support for GCC-12 in CI

samuel40791765:
### Issues:
Resolves `CryptoAlg-1100`

### Description of changes: 
We've recently fixed and added support for gcc-12: https://github.com/awslabs/aws-lc/issues/487
This PR adds a gcc-12 dimension to our CI to ensure we don't break support.

### Call-outs:
N/A

### Testing:
Tested in local fork: https://github.com/samuel40791765/aws-lc/pull/19 (android is expected to fail, because the targeted arn is for the team account)


bryce-shang (approved):

bryce-shang:
The cryptofuzz failure is tracked in `CryptoAlg-1175`. This PR should not cause the failure. Not sure if the failure is transient.

PR #543, opened 2022-06-30
Remove known 2.8 cmake hack

torben-hansen:
### Issues:

CryptoAlg-1137

### Description of changes: 

https://github.com/awslabs/aws-lc/pull/514 introduced a work-around for some weird cmake behaviour. We identified this as an issue for cmake version < 3.0. Since we now require cmake version at least 3.0, we can remove this hack.

### Testing:

CI has prefix-build tests. Hence, should be captured through CI.


justsmth (dismissed):

crypto/CMakeLists.txt
```diff
@@ -485,6 +485,10 @@ add_library(
 
 target_compile_definitions(crypto_objects PRIVATE BORINGSSL_IMPLEMENTATION)
 
+if(BORINGSSL_PREFIX AND BORINGSSL_PREFIX_SYMBOLS AND GO_EXECUTABLE)
+  add_dependencies(crypto_objects boringssl_prefix_symbols)
+endif()
```
justsmth, line 490: Why not place an unconditional dependency on `global_target`?  There might be other problem that arise due to missing transitive dependencies.

```
add_dependencies(crypto_objects global_target)
```

samuel40791765 (approved):

crypto/CMakeLists.txt
torben-hansen: That makes sense I think.

justsmth (dismissed):

justsmth (approved):

samuel40791765 (approved):

PR #548, opened 2022-07-08
Generation of crate for Rust-bindings 

justsmth:
### Description of changes: 
The change provides scripts that generate a crate (`aws-lc-sys`) for Rust bindings to AWS-LC.

### Call-outs:
These are low-level binding.  These would generally not be usable by the Rust community. 

### Testing:
* The scripts automatically test the resulting crate on Mac (x86) and Linux, both Ubuntu and AL2, via docker images.
* I've manually tested the resulting crate on ARM64 AL2 host.
* The crate does a sanity test of the its bindings, but more testing could be done.


bindings/rust/generate/_test_supported_builds.sh
```diff
@@ -0,0 +1,83 @@
...
+docker run -v `pwd`:`pwd` -w `pwd` --rm ${DOCKER_IMAGE_HOST}/${DOCKER_IMAGE_REPO_NAME}${DI_ARCH}:${DI_TAG} /bin/bash "${SCRIPT_DIR}"/_crate_test_build.sh
+
+###
+###
+DI_ARCH=aarch
+DI_TAG=ubuntu-20.04_clang-10x_latest
+###
+###
+# TODO: The ARM64 images fail the Rust build with a strange message. I suspect it's due to Docker emulation on my Mac.
+#docker run -v `pwd`:`pwd` -w `pwd` --rm ${DOCKER_IMAGE_HOST}/${DOCKER_IMAGE_REPO_NAME}${DI_ARCH}:${DI_TAG} /bin/bash "${SCRIPT_DIR}"/_crate_test_build.sh
```
justsmth, line 72: The crate build fails when run within a docker image.  However, when I copy the create to my AL2 (ARM) host, the crate builds and the tests pass.

bindings/rust/generate/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+        .generate_comments(true)
+        .fit_macro_constants(false)
+        .size_t_is_usize(true)
+        .layout_tests(true)
+        .prepend_enum_name(true)
+        .rustfmt_bindings(true)
+        .clang_args(clang_args)
+        .raw_line(COPYRIGHT)
+        .raw_line(PRELUDE)
+        .header(find_include_path(out_dir).join("rust_wrapper.h").display().to_string());
```
justsmth, line 122: Since `bindgen` assumes the bindings are for its current platform, I think this logic will be moved into the `build.rs` of the crate.

bindings/rust/aws-lc-sys-template/build.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    cmake_cfg
+}
+
+fn build_aws_lc() -> Result<PathBuf, String> {
+    let mut cmake_cfg = prepare_cmake_build(false, Some(&prefix_string()));
+    let output_dir = cmake_cfg.build_target("ssl").build();
+
+    Ok(output_dir)
+}
```
DominicDams, line 258: Why is this using a result? This never returns an error

bindings/rust/generate/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+            None
+        }
+    }
+}
+*/
+
+fn modify_bindings(bindings_path: &PathBuf, prefix: &str) -> io::Result<()>{
+    // Needed until this issue is resolved: https://github.com/rust-lang/rust-bindgen/issues/1375
+
+    let prefix_func_detector = Regex::new(&format!("(^\\s+)pub\\s+fn\\s+{}_(\\w*)(.*)", prefix)).unwrap();
```
DominicDams: I think this regex is complicated enough to warrant a comment briefly explaining it, especially since it's also doing string formatting.

bindings/rust/aws-lc-sys-template/build.rs
justsmth: A previous iteration of this function had potential for errors. I will change the function signature.

bindings/rust/generate/src/main.rs
justsmth: Good point. Will do.

bindings/rust/aws-lc-sys-template/LICENSE
```diff
@@ -0,0 +1,456 @@
+AWS Lib Crypto is a fork of BoringSSL, which is itself a fork of OpenSSL. New
```
andrewhop, line 1: Can this reference our top level license file to avoid this duplicate?

bindings/rust/aws-lc-sys-template/build.rs
```diff
@@ -0,0 +1,318 @@
...
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY
+ * SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION
+ * OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+ * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE. */
+
```
andrewhop, line 16: This doesn't seem right, is this an original file with code from AWS or from Google?

bindings/rust/aws-lc-sys-template/build.rs
```diff
@@ -0,0 +1,318 @@
...
+use std::fs::File;
+use std::io::{BufRead, BufReader, BufWriter, Write};
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use std::{env, fs, io};
+
+use regex::Regex;
+
+fn modify_bindings(bindings_path: &PathBuf, prefix: &str) -> io::Result<()> {
```
andrewhop, line 25: Is it possible to split up this review into two reviews:
1. review your changes
2. moving these bindings from rust/ to bindings/rust

Also is there a particular need to move this code? Does this help/hurt our merge-ability?

bindings/rust/aws-lc-sys-template/build.rs
```diff
@@ -0,0 +1,318 @@
...
+    // Into lines like this:
+    //
+    //     #[link_name="aws_lc_0_1_0_ERR_load_BIO_strings"]
+    //     pub fn ERR_load_BIO_strings();
+    //
+    // This allows the function to appear with its original name (e.g., ERR_load_BIO_strings)
+    // in our bindings, while still being linked to the prefixed (e.g., aws_lc_0_1_0_ERR_load_BIO_strings)
+    // function name.
+
+    // The RE expression here has 3 capture groups.
```
andrewhop, line 43: Is this [RAS (redundant acronym syndrome) syndrome](https://en.wikipedia.org/wiki/RAS_syndrome)? RE expression = Regular Expression expression?
```suggestion
    // The regular expression here has 3 capture groups.
```

bindings/rust/generate/_collect_symbols_build.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
andrewhop, line 1: What is the plan for windows?

bindings/rust/generate/_run_supported_symbol_builds.sh
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env bash
```
andrewhop, line 1: What is the intended purpose of this script? Can we update our existing CI to run these builds?

bindings/rust/aws-lc-sys-template/src/lib.rs
```diff
@@ -1,22 +1,24 @@
...
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
 
-// populated by cmake
-${INCLUDES}
+mod bindings;
+pub use bindings::*;
 
+#[allow(non_snake_case)]
 pub fn ERR_GET_LIB(packed_error: u32) -> i32 {
```
andrewhop, line 8: What makes these three functions special compared to the rest of libcrypto?

CMakeLists.txt
```diff
@@ -808,14 +808,6 @@ if(FUZZ)
   add_subdirectory(fuzz)
 endif()
 
-if(RUST_BINDINGS)
```
andrewhop, line 811: Why remove this from CMake? Some of the CMake logic seems to be moved to bash scripts.

bindings/rust/aws-lc-sys-template/build.rs
justsmth: I started from the original [build.rs](https://github.com/awslabs/aws-lc/blob/main/rust/build.rs) and just never removed the original copyright notice.. Most of the code is either original or copied from [here](https://github.com/cloudflare/boring/blob/master/boring-sys/build.rs).

CMakeLists.txt
justsmth: We might be able to reintroduce a CMake mechanism to produce Rust bindings for someone's local use, but that would have little value for consumers. The Rust community needs a way to utilize AWS-LC by simply listing a crate/version in their Cargo.toml. The goal of this PR is to generate an "aws-lc-sys" Rust crate that can be published for that purpose.

bindings/rust/aws-lc-sys-template/src/lib.rs
justsmth: These "functions" are actually preprocessor macros. An alternative way would be to code the operations of the Macro in Rust (as is done [here](https://github.com/cloudflare/boring/blob/master/boring-sys/src/lib.rs#L47)), but if the macro was ever changed we might forget to update that logic in Rust. The way it's done in this PR (utilizing Boring's `rust_wrapper.c` above) helps ensure that the logic used in Rust is consistent with the C macro.

justsmth: Honestly, I don't really see those macros ever changing their logic. There might be a slight performance improvement by implementing it in Rust directly? I don't have a strong opinion on this either way.

bindings/rust/generate/_collect_symbols_build.sh
justsmth: I'd like to add Windows support to this crate as well, but this first iteration doesn't provide that. We can prioritize that next if needed.

bindings/rust/aws-lc-sys-template/LICENSE
justsmth: This file will be published as part of the crate. I could copy the LICENSE from the top-level during crate generation instead of placing it in this "template" -- it's just more convenient having it in the template.

bindings/rust/aws-lc-sys-template/src/lib.rs
```diff
@@ -1,22 +1,24 @@
...
 
+#[allow(non_snake_case)]
 pub fn ERR_GET_FUNC(packed_error: u32) -> i32 {
     unsafe { ERR_GET_FUNC_RUST(packed_error) }
 }
 
 pub fn init() {
-    unsafe { CRYPTO_library_init(); }
+    unsafe { CRYPTO_library_init() }
 }
```
camshaft, line 24: Is this just an alias? Can this function be called multiple times? On multiple concurrent threads? It's probably best to keep it unsafe if it's not truly idempotent and synchronized.

bindings/rust/aws-lc-sys-template/build.rs
```diff
@@ -0,0 +1,318 @@
...
+        "cargo:rustc-link-lib={}={}",
+        Static.rust_lib_type(),
+        Crypto.libname(None)
+    );
+    println!(
+        "cargo:rustc-link-lib={}={}",
+        Static.rust_lib_type(),
+        Ssl.libname(None)
+    );
+
```
camshaft, line 316: You should export the location of your C headers in case consumers want to do some additional preprocessing. s2n-tls does this:
https://github.com/aws/s2n-tls/blob/e9906015bbc19ba17dd26846b84210c93dcc6e13/bindings/rust/s2n-tls-sys/build.rs#L162

bindings/rust/aws-lc-sys-template/build.rs
```diff
@@ -0,0 +1,318 @@
...
+    let mut cmake_cfg = prepare_cmake_build(false, Some(&prefix_string()));
+
+    cmake_cfg.build_target("ssl").build()
+}
+
+fn main() -> Result<(), String> {
+    use crate::OutputLib::{Crypto, Ssl};
+    use crate::OutputLibType::Static;
+
+    let out_dir = env::current_dir().unwrap();
```
camshaft, line 287: you should pull this from the `OUT_DIR` environment variable. I'm not sure it's guaranteed to be the cwd

camshaft: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts

bindings/rust/aws-lc-sys-template/build.rs
```diff
@@ -0,0 +1,318 @@
...
+                cc_version, REQUIRED_CLANG_VERSION
+            );
+        } else if !cc_version.is_empty() {
+            eprintln!(
+                "warning: FIPS requires clang version {}, skipping incompatible version \"{}\"",
+                REQUIRED_CLANG_VERSION, cc_version
+            );
+        }
+    }
+    unreachable!()
```
camshaft, line 237: Might be better to provide an error message here indicating that it couldn't find a c compiler?

bindings/rust/aws-lc-sys-template/src/lib.rs
justsmth: According to [its documentation](https://github.com/awslabs/aws-lc/blob/610e2f58a8596a84be3207d433a2ca3b0c2cb63b/include/openssl/crypto.h#L41):
> // CRYPTO_library_init initializes the crypto library. It must be called if the
> // library is built with BORINGSSL_NO_STATIC_INITIALIZER. Otherwise, it does
> // nothing and a static initializer is used instead. It is safe to call this
> // function multiple times and concurrently from multiple threads.

justsmth: We don't build AWS-LC with that flag, so this function could probably be removed altogether.

bindings/rust/generate/_collect_symbols_build.sh
justsmth: The scripts in the "generate" directory currently must execute on a Mac that has access to our Linux (x64/aarch) Docker images.  A Mac can utilize its local (Mac-specific) dev tools as well as run scripts on the docker images to build and test the crate, all from a single host. I'm open to suggestions on better ways to do this. 

To support Windows, just the `_collect_symbols_build.sh` and `_crate_test_build.sh` scripts would either need to rewritten (in Rust?), or we'd need to setup a Windows docker image so it could run these bash scripts.

bindings/rust/aws-lc-sys-template/build.rs
```diff
@@ -0,0 +1,318 @@
...
+        .write_to_file(&bindings_file)
+        .expect("Unable to write bindings to file.");
+    modify_bindings(&bindings_file, &prefix).map_err(|err| err.to_string())?;
+
+    let aws_lc_dir = build_aws_lc();
+
+    let libcrypto_dir = Crypto.locate_dir(&aws_lc_dir);
+    let libssl_dir = Ssl.locate_dir(&aws_lc_dir);
+    println!("cargo:rustc-link-search=native={}", libcrypto_dir.display());
+    println!("cargo:rustc-link-search=native={}", libssl_dir.display());
```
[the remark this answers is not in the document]
justsmth: You have a good point. Since there are 2 libs (libcrypto and libssl), should there be 2 crates?  Or make "ssl" a feature of the `aws-lc-sys` crate?

The include files for libcrypto and libssl are all in the [same directory](https://github.com/awslabs/aws-lc/tree/main/include/openssl), so we'd need to split those when invoking `rust-bindgen` in a way that's consistent with which lib the corresponding functions are placed in.

bindings/rust/generate/generate.sh
```diff
@@ -0,0 +1,170 @@
...
+  fi
+}
+
+function prepare_crate_dir {
+  echo Preparing crate directory: ${CRATE_DIR}
+  mkdir -p "${CRATE_DIR}"
+  mkdir -p "${CRATE_AWS_LC_DIR}"/
+
+  cp -r "${CRATE_TEMPLATE_DIR}"/* "${CRATE_DIR}"/
+  perl -pi -e "s/__AWS_LC_SYS_VERSION__/${AWS_LC_SYS_VERSION}/g" "${CRATE_DIR}"/Cargo.toml
```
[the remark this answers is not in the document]
justsmth: In general, `sed` would be more portable than `perl`. However, in-place substitution in files differs slightly between [Linux and MacOS](https://stackoverflow.com/questions/12696125/sed-edit-file-in-place).  Since `perl` is already required by the AWS-LC build, I felt that using it here wouldn't be problematic.

bindings/rust/aws-lc-sys-template/LICENSE
justsmth: I agree.  I'll remove this license, and have the script copy from the top-level.

bindings/rust/aws-lc-sys-template/src/lib.rs
```diff
@@ -1,22 +1,24 @@
...
+mod bindings;
+pub use bindings::*;
 
+#[allow(non_snake_case)]
 pub fn ERR_GET_LIB(packed_error: u32) -> i32 {
     unsafe { ERR_GET_LIB_RUST(packed_error) }
 }
 
+#[allow(non_snake_case)]
 pub fn ERR_GET_REASON(packed_error: u32) -> i32 {
```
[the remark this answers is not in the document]
justsmth: I agree. I'll mark the functions `unsafe`.

bindings/rust/aws-lc-sys-template/tests/sanity-tests.rs
```diff
@@ -0,0 +1,32 @@
...
+use Vec;
+use aws_lc_sys;
+
+use sha1::{Sha1, Digest};
+
+fn sha1_tester(input: &[u8]) -> [u8; 20] {
+    let mut hash = MaybeUninit::<[u8; 20]>::uninit();
+
+    unsafe {
+        aws_lc_sys::SHA1(input.as_ptr(), input.len(), hash.as_mut_ptr().cast());
```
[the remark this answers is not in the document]
justsmth: No. I also removed the initialization function since it's not needed given the build flags we're using.

bindings/rust/aws-lc-sys-template/tests/sanity-tests.rs
```diff
@@ -0,0 +1,32 @@
...
+        hash.assume_init()
+    }
+}
+
+fn compare(result: &[u8], expected_result: &[u8]) {
+    //println!("Comparing: {:?} to {:?}", result, expected_result);
+    assert_eq!(Vec::from(result), Vec::from(expected_result));
```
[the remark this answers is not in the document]
justsmth: I'm not sure why I wrapped these in a Vec. I'll remove it.

bindings/rust/aws-lc-sys-template/build.rs
justsmth: Actually, this "output" directory relates to the where the generated bindings will be placed. I'll change the name to avoid confusion.

> CARGO_MANIFEST_DIR — The directory containing the manifest for the package being built (the package containing the build script). Also note that this is the value of the current working directory of the build script when it starts.

bindings/rust/generate/_collect_symbols_build.sh
```diff
@@ -0,0 +1,40 @@
...
+
+function usage {
+  echo "Usage: $(basename $0)"
+}
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+AWS_LC_DIR=$( cd -- "${SCRIPT_DIR}/../../../" &> /dev/null && pwd)
+TMP_DIR="${AWS_LC_DIR}"/bindings/rust/tmp
+RAND_NAME=$(od -vN 8 -An -tx1 /dev/urandom | tr -d " \n" )
+BUILD_DIR=${TMP_DIR}/BUILD-${RAND_NAME}
```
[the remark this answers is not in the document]
justsmth: I just wanted to control where this directory gets created. I'll look into how hard this would be to refactor to use mktemp.

bindings/rust/generate/_crate_test_build.sh
```diff
@@ -0,0 +1,58 @@
...
+  if [[ $(lsb_release -a |grep -i ubuntu) ]]; then
+    apt-get -y update
+    apt-get -y install make
+  else
+    yum -y update
+    yum -y install make
+  fi
+
+  rm -f /usr/bin/cc
+  rm -f /usr/bin/c++
```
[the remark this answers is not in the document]
justsmth: Yeah, the right solution is for me to create/update the docker images so such insanity like this isn't needed. I thought I could delay updating the docker images to a subsequent PR, but maybe I should do that first.

bindings/rust/generate/_crate_test_build.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
justsmth: The purpose is to indicate that these scripts shouldn't be executed directly by a user, they're invoked by "generate.sh". Maybe I should just move all of this logic into a single script?

bindings/rust/generate/generate.sh
```diff
@@ -0,0 +1,170 @@
...
+set -e
+
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+
+function usage {
+  echo
+  echo "Usage: $(basename $0) [-d] [-b] [-u] [AWS_ACCOUNT_ID]"
+  echo
+  echo "    AWS_ACCOUNT_ID - The account providing docker images for the build."
```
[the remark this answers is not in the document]
justsmth: Anyone can build the docker images from the code in our Testing/CI directory, but I don't think any of our team's AWS accounts are open for them to be retrieved from.

bindings/rust/aws-lc-sys-template/build.rs
justsmth: Good point. Will fix...

bindings/rust/aws-lc-sys-template/build.rs
justsmth: I removed the code I had adapted from the other (MIT licensed) source.  All of the code in that file now is either original to that file or code I had written on my own.

bindings/rust/aws-lc-sys-template/build.rs
justsmth: I removed the function entirely.  It relates to a FIPS build, which we don't yet support anyways.

bindings/rust/generate/_test_supported_builds.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -e -x
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+AWS_LC_DIR=$( cd -- "${SCRIPT_DIR}/../../../" &> /dev/null && pwd)
+TMP_DIR="${AWS_LC_DIR}"/bindings/rust/tmp
+CRATE_DIR="${TMP_DIR}"/aws-lc-sys
```
torben-hansen, line 10: Unused?

bindings/rust/generate/_run_supported_symbol_builds.sh
```diff
@@ -0,0 +1,35 @@
...
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+set -e
+
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+AWS_ACCOUNT_ID="${1}"
```
torben-hansen, line 8: Unused?

bindings/rust/generate/generate.sh
```diff
@@ -0,0 +1,165 @@
...
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+AWS_LC_DIR=$( cd -- "${SCRIPT_DIR}/../../../" &> /dev/null && pwd)
+CRATE_TEMPLATE_DIR="${AWS_LC_DIR}"/bindings/rust/aws-lc-sys-template
+TMP_DIR="${AWS_LC_DIR}"/bindings/rust/tmp
+SYMBOLS_FILE="${TMP_DIR}"/symbols.txt
+CRATE_DIR="${TMP_DIR}"/aws-lc-sys
+CRATE_AWS_LC_DIR="${CRATE_DIR}"/deps/aws-lc
+PREFIX_HEADERS_FILE="${CRATE_AWS_LC_DIR}"/include/boringssl_prefix_symbols.h
+BINDINGS_FILE="${CRATE_DIR}"/src/bindings.rs
```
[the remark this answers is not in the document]
justsmth: Yes. Will remove.

bindings/rust/generate/_test_supported_builds.sh
justsmth: Yes. Will remove.

bindings/rust/generate/_run_supported_symbol_builds.sh
justsmth: Yes. Will remove.

DominicDams (approved):

bindings/rust/aws-lc-sys-template/build.rs
```diff
@@ -0,0 +1,271 @@
...
+        "cargo:rustc-link-lib={}={}",
+        Static.rust_lib_type(),
+        Crypto.libname(None)
+    );
+
+    println!(
+        "cargo:include={}",
+        get_include_path(&manifest_dir).display()
+    );
+    println!("cargo:rerun-if-changed=build.rs");
```
camshaft, line 268: nit: this already works out of the box so it isn't really needed

camshaft (approved):

bindings/rust/aws-lc-sys-template/build.rs
justsmth: The default behavior seems slightly more likely to [trigger an unnecessary build](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed). However, in practice the behavior our dependencies see should be the same whether this line is present or not.

> If the build script inherently does not need to re-run under any circumstance, then emitting cargo:rerun-if-changed=build.rs is a simple way to prevent it from being re-run (otherwise, the default if no rerun-if instructions are emitted is to scan the entire package directory for changes).

bindings/rust/aws-lc-sys-template/build.rs
```diff
@@ -0,0 +1,271 @@
...
+// Modifications Copyright Amazon.com, Inc. or its affiliates. See GitHub history for details.
+
+use regex::Regex;
+use std::{env, fs, io};
+use std::fs::File;
+use std::io::{BufRead, BufReader, BufWriter, Write};
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+fn modify_bindings(bindings_path: &PathBuf, prefix: &str) -> io::Result<()> {
```
justsmth, line 25: The current logic here is not sufficient -- it only updates the function symbols, but there are other symbols that still need to be modified. For example, the following can currently be found in bindings.rs:
```
...
extern "C" {
    pub static aws_lc_0_1_0_ASN1_BMPSTRING_it: ASN1_ITEM;
}
extern "C" {
    pub static aws_lc_0_1_0_ASN1_GENERALSTRING_it: ASN1_ITEM;
}
extern "C" {
    pub static aws_lc_0_1_0_ASN1_IA5STRING_it: ASN1_ITEM;
}
...
```

DominicDams (approved):

camshaft (approved):

torben-hansen (approved):

PR #551, opened 2022-07-13
Install bssl as well

torben-hansen:
### Description of changes: 

`bssl` tool is not installed. Hence, we don't distribute it.

This change adds an install target to install `bssl` on (e.g.) `ninja-build install`, in turn, making it easier to distribute.

### Call-outs:

Right now, just installed in root of install directory. That is:

```
% pwd
[...]/aws-lc/awslc-install/lib64

% ls
bssl  /crypto  libcrypto.a  libssl.a  /ssl
```

### Testing:

Ran `ninja-build install` on AL2 instance. Got the output you see in "Call-outs".

### DIFF
--- a/tool/CMakeLists.txt
+++ b/tool/CMakeLists.txt
@@ -43,7 +43,7 @@ else()
 endif()
 
 install(TARGETS bssl
-        RUNTIME DESTINATION ${CMAKE_INSTALL_LIBDIR})
+        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
 
 function(build_benchmark target_name additional_include_dir lib_crypto)
   message("-- Building ${target_name} benchmark using header files from ${additional_include_dir} and libcrypto from ${lib_crypto}.")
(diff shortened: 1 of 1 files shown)


DominicDams (dismissed):
We might want to consider adding logic to install the benchmarking tools as well, but this works for now at the very least

torben-hansen:
> We might want to consider adding logic to install the benchmarking tools as well, but this works for now at the very least

I wouldn't do that. The "other libcrypto version" benchmarking tools are not very useful. And `bssl` already contains `speed`.

tool/CMakeLists.txt
```diff
@@ -42,6 +42,9 @@ else()
   endif()
 endif()
 
+install(TARGETS bssl
+        RUNTIME DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
andrewhop, line 46: Our top level CMake script defines CMAKE_INSTALL_BINDIR or it includes GNUInstallDirs which also defines it. 
```suggestion
        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```

DominicDams (approved):

andrewhop (approved):

PR #546, opened 2022-07-07
Benchmark build testing

DominicDams:
### Issues:
CryptoAlg-1093

### Description of changes: 
Adds CI dimensions for OpenSSL 1.1, 3.0 and aws-lc benchmarking tools

### Testing:
Built and tested the CI in my test account

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #552, opened 2022-07-14
Open ssl 1.0 benchmark

DominicDams:
### Description of changes: 
Adds support for OpenSSL1.0 benchmarking and makes the benchmarking framework more uniform for different versions of OpenSSL

### Testing:
Verified that all currently working versions of the benchmark framework still work, and the new OpenSSL 1.0 version worked by running the benchmarks locally:
After installing all 3 versions of OpenSSL in the appropriate directories I ran
```
cmake .. -GNinja -DOPENSSL_3_0_INSTALL_DIR=/home/ec2-user/libcrypto_installs/openssl-3.0 -DAWSLC_INSTALL_DIR=/home/ec2-user/libcrypto_installs/aws-lc -DOPENSSL_1_0_INSTALL_DIR=/home/ec2-user/libcrypto_installs/openssl-1.0 -DOPENSSL_1_1_INSTALL_DIR=/home/ec2-user/libcrypto_installs/openssl-1.1
```
then 
```
ninja-build
```
then in the nu terminal I ran
```
for x in ["ossl_1_1_bm" "ossl_1_0_bm" "ossl_3_0_bm" "awslc_bm"] { nu -c $"./tool/($x)"}
```
to verify that all the tools ran properly without crashing


DominicDams (commented):
I put this in Draft since it will conflict with #546

tool/README.md
```diff
@@ -26,8 +26,9 @@ In order to build the above-mentioned benchmarking tools, absolute paths to each
 | ------------- | ------------- |
 | awslc_bm | -DAWSLC_INSTALL_DIR |
 | bssl_bm | -DBORINGSSL_INSTALL_DIR |
-| ossl_bm | -DOPENSSL_INSTALL_DIR |
-| ossl3_bm | -DOPENSSL3_INSTALL_DIR |
+| ossl_1_0_bm | -DOPENSSL_1_0_INSTALL_DIR |
```
torben-hansen, line 29: Use full version numbers? There are multiple other versions than `1.0.2` and `1.1.1`.

DominicDams: I think this should be compatible with any 1.0.x/1.1.x but I'll do some work to double check that

DominicDams: Verified that the `1.0` version works with the stable releases of OpenSSL `1.0.0` `1.0.1` and `1.0.2` and the `1.1` version works with both `1.1.0` and `1.1.1`. I haven't looked into any of the `0.x` releases but I doubt anyone is going to need to compare those versions performance to aws-lc's performance.

samuel40791765 (dismissed):

tool/speed.cc
```diff
@@ -627,10 +635,18 @@ static bool SpeedAES256XTS(const std::string &name, //const size_t in_len,
 
   return true;
 }
+#endif
 
 static bool SpeedHashChunk(const EVP_MD *md, std::string name,
                            size_t chunk_len) {
+  // OpenSSL 1.0 doesn't use ptr types for EVP_MD_CTX
```
torben-hansen, line 642: Could you elaborate? Not sure what is meant by
> doesn't use ptr types for EVP_MD_CTX

Is `EVP_MD_CTX_create()` what you are looking for?

tool/speed.cc
```diff
@@ -673,7 +689,11 @@ static bool SpeedHash(const EVP_MD *md, const std::string &name,
 
 static bool SpeedHmacChunk(const EVP_MD *md, std::string name,
                            size_t chunk_len) {
+#if defined(OPENSSL_1_0_BENCHMARK)
+  std::unique_ptr<HMAC_CTX> ctx = std::unique_ptr<HMAC_CTX>(new HMAC_CTX);
```
torben-hansen, line 693: I theory I think you are supposed to call `HMAC_CTX_init()` on the object. But in practice, meh, not sure it makes a difference.

tool/ossl_bm.h
```diff
@@ -48,9 +48,13 @@ OSSL_MAKE_DELETER(BIGNUM, BN_free)
 OSSL_MAKE_DELETER(EC_KEY, EC_KEY_free)
 OSSL_MAKE_DELETER(EC_POINT, EC_POINT_free)
 OSSL_MAKE_DELETER(BN_CTX, BN_CTX_free)
-OSSL_MAKE_DELETER(EVP_MD_CTX, EVP_MD_CTX_free)
 OSSL_MAKE_DELETER(EVP_CIPHER_CTX, EVP_CIPHER_CTX_free)
+
+// OpenSSL 1.0 uses structs for EVP_MD/HMAC so they use the default deleter
```
torben-hansen, line 53: What do you mean by?
> OpenSSL 1.0 uses structs for EVP_MD/HMAC

We probably don't even need this pointers, since their object structure is public in OpenSSL 1.0.2. So, we can just put an object on the stack and get a pointer to it. But might be slightly easier what you have done, with less openssl-version conditional code.

tool/CMakeLists.txt
```diff
@@ -69,28 +69,40 @@ if(AWSLC_INSTALL_DIR)
 endif()
 
 # This expects a directory which contains the includes in include/openssl/ and the OpenSSL artifacts in lib/
-if(OPENSSL_INSTALL_DIR)
-  build_benchmark(ossl_bm ${OPENSSL_INSTALL_DIR}/include ${OPENSSL_INSTALL_DIR}/lib/libcrypto.a)
+# Currently this is the default OpenSSL build we target so the "OPENSSL_1_1_BENCHMARK" flag isn't used, 
+# but we include this to maintain uniformity across OpenSSL versions
```
torben-hansen, line 73: Makes sense.

tool/ossl_bm.h
DominicDams: That was why I did it this way :)

tool/speed.cc
DominicDams: I swear I looked for this and couldn't find it, but you're right

tool/ossl_bm.h
torben-hansen: Makes sense :)

samuel40791765 (dismissed):

torben-hansen (approved):

samuel40791765 (approved):

PR #556, opened 2022-07-18
Prettify printing system and dependency information

torben-hansen:
### Description of changes: 

Low-handing fruit. But can re-factor printing system and dependency information in the CI making it slightly easier to add new ones.

Looking at adding Rust dependency information as well.

### Testing:

Tested locally


bryce-shang (dismissed):

tests/ci/common_posix_setup.sh
```diff
@@ -143,63 +143,31 @@ function build_and_run_minimal_test {
...
+  LABEL=${3}
+
+  if command -v ${EXE_NAME} &> /dev/null
   then
     echo ""
-    echo "Make version:"
-    make --version
+    echo "${LABEL}:"
+    ${EXE_NAME} ${EXE_ARGUMENT}
   fi
```
justsmth, line 156: Should there be an `else` block to log the fact that the command is not found -- i.e., be explicit instead of implicit about missing tools?

tests/ci/common_posix_setup.sh
```diff
@@ -143,63 +143,31 @@ function build_and_run_minimal_test {
...
+  print_executable_information "cmake" "--version" "CMake version"
+  print_executable_information "cmake3" "--version" "CMake version (cmake3 executable)"
+  print_executable_information "go" "version" "Go version"
+  print_executable_information "perl" "--version" "Perl version"
+  # Ninja executable names are not uniform over operating systems
+  print_executable_information "ninja-build" "--version" "Ninja version (ninja-build executable)"
+  print_executable_information "ninja" "--version" "Ninja version (ninja executable)"
+  print_executable_information "gcc" "--version" "GCC version"
+  print_executable_information "g++" "--version" "G++ version"
+  print_executable_information "clang" "--version" "Clang version"
```
justsmth, line 169: You specifically look for `gcc`, `g++` and `clang`, but what about verifying the version of `clang++`? Should we check the whether/which of system defaults: `cc` and `c++`?

bryce-shang: Maybe we can skip `gcc/++` or `clang/++` because CMake build logs tell which compilers are used
```
+ cmake -DBUILD_SHARED_LIBS=1 -DCMAKE_BUILD_TYPE=Release -GNinja ../
-- The C compiler identification is GNU 7.5.0
-- Check for working C compiler: /usr/bin/gcc-7
-- Check for working C compiler: /usr/bin/gcc-7 -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- The CXX compiler identification is GNU 7.5.0
-- Check for working CXX compiler: /usr/bin/g++-7
-- Check for working CXX compiler: /usr/bin/g++-7 -- works
```

tests/ci/common_posix_setup.sh
torben-hansen: I guess that doesn't hurt

tests/ci/common_posix_setup.sh
torben-hansen: I think we should just add all the stuff we can think about to see what environment we are running in.

justsmth (approved):

PR #559, opened 2022-07-21
Upstream merge 2022-07-21

dkostic:
### Issues:
N/A

### Description of changes: 
Upstream merge 2022-07-21.

### Call-outs:
Details in internal doc.

### Testing:


DominicDams (approved):

justsmth (approved):

crypto/x509/x_crl.c
```diff
@@ -243,14 +208,31 @@ static int crl_cb(int operation, ASN1_VALUE **pval, const ASN1_ITEM *it,
...
-    case ASN1_OP_D2I_POST:
+    case ASN1_OP_D2I_POST: {
+        /* The version must be one of v1(0) or v2(1). */
+        long version = X509_CRL_VERSION_1;
+        if (crl->crl->version != NULL) {
+            version = ASN1_INTEGER_get(crl->crl->version);
+            /* TODO(https://crbug.com/boringssl/364): |X509_CRL_VERSION_1|
+             * should also be rejected. This means an explicitly-encoded X.509v1
+             * version. v1 is DEFAULT, so DER requires it be omitted. */
+            if (version < X509_CRL_VERSION_1 || version > X509_CRL_VERSION_2) {
```
sdkellyAWS, line 224: This block of code is duplicated in two different files. One with a comment and another without the comment. Is there a way to turn this logic into a helper function so they do not go out of sync?

crypto/x509v3/v3_utl.c
```diff
@@ -222,7 +222,8 @@ static char *bignum_to_string(const BIGNUM *bn)
     return ret;
 }
 
-char *i2s_ASN1_ENUMERATED(X509V3_EXT_METHOD *method, const ASN1_ENUMERATED *a)
+char *i2s_ASN1_ENUMERATED(const X509V3_EXT_METHOD *method,
```
sdkellyAWS, line 225: There seem to be quite a few functions that don't use all the parameters. Am I missing the reason for this?

include/openssl/x509v3.h
```diff
@@ -79,23 +79,24 @@ struct v3_ext_ctx;
...
-    const struct v3_ext_method *method, void *ext,
-    STACK_OF(CONF_VALUE) *extlist);
-typedef void *(*X509V3_EXT_V2I)(const struct v3_ext_method *method,
-                                struct v3_ext_ctx *ctx,
-                                STACK_OF(CONF_VALUE) *values);
-typedef char *(*X509V3_EXT_I2S)(const struct v3_ext_method *method, void *ext);
-typedef void *(*X509V3_EXT_S2I)(const struct v3_ext_method *method,
-                                struct v3_ext_ctx *ctx, const char *str);
-typedef int (*X509V3_EXT_I2R)(const struct v3_ext_method *method, void *ext,
+typedef STACK_OF(CONF_VALUE) *(*X509V3_EXT_I2V)(const X509V3_EXT_METHOD *method,
```
sdkellyAWS, line 88: Some functions/definitions seem to define parameters as:
```
foo (type a,
        int b,
        thing c)
```

While others as:
```foo(type a, int b, thing c)```

Is there a style guide that indicates which is correct, and when, for me to look at?

File: ssl/tls13_server.cc [hunk 4]
```diff
@@ -131,15 +131,12 @@ static bool add_new_session_tickets(SSL_HANDSHAKE *hs, bool *out_sent_tickets) {
...
-  // TLS 1.3 recommends single-use tickets, so issue multiple tickets in case
-  // the client makes several connections before getting a renewal.
-  static const int kNumTickets = 2;
-
   // Rebase the session timestamp so that it is measured from ticket
   // issuance.
   ssl_session_rebase_time(ssl, hs->new_session.get());
 
-  for (int i = 0; i < kNumTickets; i++) {
+  assert(ssl->session_ctx->num_tickets <= kMaxTickets);
```
sdkellyAWS, line 138: Is `kMaxTickets` initialized somewhere else now?

crypto/x509v3/v3_skey.c
```diff
@@ -67,23 +67,13 @@
...
-    0, 0, 0, 0,
-    NULL
-};
-
-char *i2s_ASN1_OCTET_STRING(X509V3_EXT_METHOD *method, const ASN1_OCTET_STRING *oct)
+char *i2s_ASN1_OCTET_STRING(const X509V3_EXT_METHOD *method,
+                            const ASN1_OCTET_STRING *oct)
 {
     return x509v3_bytes_to_hex(oct->data, oct->length);
 }
```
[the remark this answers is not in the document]
justsmth: This function is declared in [x509v3.h](https://github.com/awslabs/aws-lc/blob/752ee755f10de080c055d87cea9b7d667342a528/include/openssl/x509v3.h#L530) so it's exposed externally. I assume such parameters are being kept around to avoid breaking existing client code.(?)

include/openssl/x509v3.h
justsmth: There's a mention parameters style [here](https://github.com/awslabs/aws-lc/blob/main/STYLE.md#parameters), but I don't see specific guidance for line returns between them.

I suspect the convention is to use a single line for the function declaration unless that line gets too long, otherwise place parameters (after the first) each on a separate line.

justsmth: Whatever the style conventions may be, they're not consistently followed nor enforced.

ssl/tls13_server.cc
justsmth: I see initialized in `ssl/internal.h`:

```
// kMaxTickets is the maximum number of tickets to send immediately after the
// handshake. We use a one-byte ticket nonce, and there is no point in sending
// so many tickets.
constexpr size_t kMaxTickets = 16;
```

PR #564, opened 2022-07-25
Take in BoringSSL service indicator improvements

samuel40791765:
### Issues:
Resolves `CryptoAlg-1209`

### Description of changes: 
This PR takes in the service indicator improvements BoringSSL made in https://github.com/google/boringssl/commit/118a892d2da8c78b46ed549454b3b62ded8c84b7#diff-ce7ffb6368fc741735e07a58ef82c4e0bb9b31d389299f8d0d1e961c76dd9f38

**Specific change details can be found within the SIM and internal document: [BoringSSL-service-indicator-differences].
Any new behavior within the tests/functions will most likely be brought up within the internal documentation.**

Most of the changes were taken except: 
* AWS-LC returns `NOT_APPROVED` for SHA-512/256. This is because we have not gone through FIPS validation for algorithms involving SHA-512/256, but we can expect to do so within the next round of FIPS.
* `kEVPKeyGenShouldCallFIPSFunctions`, `kCurveSecp256k1Supported`, `kEVPDeriveSetsServiceIndicator` are set to `true`. BoringSSL was nice enough to use these `bool` to help us keep some of our existing AWS-LC specific tests. These are set to `false` for Boring.

### Call-outs:
N/A

### Testing:
Some test cases were rearranged and optimized according to their improvements. 


crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -256,50 +181,9 @@ static int is_md_fips_approved_for_signing(int md_type) {
...
-    }
-  }
-}
-
-// Returns 1 if the given message digest type is FIPS approved for verifying,
-// and 0 otherwise. Only SHA1, SHA2-224, SHA2-256, SHA2-384, and SHA2-512
-// are approved.
+// is_md_fips_approved_for_verifying returns one if the given message digest
+// type is FIPS approved for verifying, and zero otherwise.
+// TODO: FIPS validate SHA512/256 for verifying.
```
dkostic, line 186: relevant for aws-lc?

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -313,56 +197,153 @@ static int is_md_fips_approved_for_verifying(int md_type) {
...
   } else if (pkey_type == EVP_PKEY_EC) {
-    int md_type = EVP_MD_CTX_type(ctx);
-    // Note: even though the function is called |EC_GROUP_get_curve_name|
-    // it actually returns the NID of the curve rather than the name.
-    int curve_nid = EC_GROUP_get_curve_name(ctx->pctx->pkey->pkey.ec->group);
-
     // Check if the MD type and the elliptic curve are approved.
-    if (is_md_fips_approved_for_verifying(md_type) &&
-        is_ec_fips_approved(curve_nid)) {
+    if (md_ok(md_type) && is_ec_fips_approved(EC_GROUP_get_curve_name(
```
dkostic, line 255: nit: I like more the previous impl:
```
int curve_nid = EC_GROUP_get_curve_name(ctx->pctx->pkey->pkey.ec->group);
is_ec_fips_approved(curve_nid)
```

include/openssl/service_indicator.h
```diff
@@ -28,56 +28,71 @@ OPENSSL_EXPORT uint64_t FIPS_service_indicator_after_call(void);
...
-    (approved) = AWSLC_NOT_APPROVED;                                \
-    int before = FIPS_service_indicator_before_call();              \
-    func;                                                           \
-    int after = FIPS_service_indicator_after_call();                \
-    if (FIPS_service_indicator_check_approved(before, after)) {     \
-        (approved) = AWSLC_APPROVED;                                \
-    }                                                               \
- }                                                                  \
- while(0)                                                           \
+#define AWSLC_MODE_STRING "AWS-LC FIPS "
```
dkostic, line 33: is the space at the end of the string a typo? the same for the non-FIPS string?

include/openssl/service_indicator.h
```diff
@@ -28,56 +28,71 @@ OPENSSL_EXPORT uint64_t FIPS_service_indicator_after_call(void);
...
 
 #define AWSLC_VERSION_STRING AWSLC_MODE_STRING AWSLC_VERSION_NUMBER_STRING
 
 #if defined(__cplusplus)
 }  // extern C
-#endif
+
+#if !defined(BORINGSSL_NO_CXX)
+
+extern "C++" {
```
dkostic, line 48: probably stupid question but nevertheless: defining `CALL_SERVICE_AND_CHECK_APPROVED` inside `extern "C++"` -- are we still going to be able to invoke the macro in C code?

include/openssl/service_indicator.h
samuel40791765: No I believe it's actually intentional. It's used in `AWSLC_VERSION_STRING` and printing it along with the `AWSLC_VERSION_NUMBER_STRING` appended behind it will give "AWS-LC FIPS 1.1.0" and "AWS-LC 1.1.0".

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -208,42 +166,9 @@ static int is_ec_fips_approved(int curve_nid) {
...
-// is FIPS approved, otherwise does nothing.
-void EC_KEY_keygen_verify_service_indicator(EC_KEY *eckey) {
-  int curve_nid = EC_GROUP_get_curve_name(eckey->group);
-  if (is_ec_fips_approved(curve_nid)) {
-    FIPS_service_indicator_update_state();
-  }
-}
-
-void EVP_PKEY_keygen_verify_service_indicator(const EVP_PKEY *pkey) {
-  if (pkey->type == EVP_PKEY_RSA || pkey->type== EVP_PKEY_RSA_PSS) {
```
nebeid, line 221: Why was EVP_PKEY_RSA_PSS removed? I understand we support it: https://github.com/awslabs/aws-lc/pull/140 and ACVP-tested it.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -313,56 +197,153 @@ static int is_md_fips_approved_for_verifying(int md_type) {
...
+    goto err;
+  }
+
+  EVP_PKEY_CTX *const pctx = ctx->pctx;
+  const EVP_PKEY *const pkey = EVP_PKEY_CTX_get0_pkey(pctx);
+  const int pkey_type = EVP_PKEY_id(pkey);
+  const int md_type = EVP_MD_CTX_type(ctx);
 
-  if (pkey_type == EVP_PKEY_RSA || pkey_type == EVP_PKEY_RSA_PSS) {
+  // EVP_PKEY_RSA_PSS SPKIs aren't supported.
```
nebeid, line 213: Related to my question about removing EVP_PKEY_RSA_PSS above.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -694,161 +618,358 @@ TEST_P(AEAD_ServiceIndicatorTest, EVP_AEAD) {
...
+        sizeof(kTDES_EDE3_CBCCipherText),
+        false,
+        AWSLC_NOT_APPROVED,
+    },
 };
 
-class EVP_ServiceIndicatorTest : public awslc::TestWithNoErrors<CipherTestVector> {};
+class EVPServiceIndicatorTest : public TestWithNoErrors<CipherTestVector> {};
+
+static void TestOperation(const EVP_CIPHER *cipher, bool encrypt,
```
nebeid, line 782: This function seems to have gotten smaller when moved here. Is that intended? The comments seem to have been kept, but I don't know if we're testing 
> |EVP_EncryptFinal_ex| and |EVP_DecryptFinal_ex| for approval at the end.

Also I had documented internally:
> [Commit from Boringssl](https://github.com/google/boringssl/commit/16a94930ac6edb19bc647a626f58a7b3fdc9dddf) (https://github.com/google/boringssl/commit/16a94930ac6edb19bc647a626f58a7b3fdc9dddf) adds new functions EVP_CipherFinal, EVP_EncryptFinal, and EVP_DecryptFinal . These functions are wrapper functions to the respective EVP_*Final_ex functions, and provide no extra logic. All three of these functions should be Service Indicator approved like the functions they call.

samuel40791765: It was intentional, and they're deduction was as follows:
> You've both [DoEncryptFinal](https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/service_indicator/service_indicator_test.cc#L120) and [DoCipherFinal](https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/service_indicator/service_indicator_test.cc#L146), but if you look at the definition of EVP_CipherFinal_ex you'll see that they can be collapsed.

It aligns with what you documented internally. Looking into the function definition, it is true the original `DoEncryptFinal` test was a bit redundant: 
https://github.com/awslabs/aws-lc/blob/752ee755f10de080c055d87cea9b7d667342a528/crypto/fipsmodule/cipher/cipher.c#L518-L524

crypto/fipsmodule/service_indicator/service_indicator.c
nebeid: Thank you, Samuel. So I'm curious, how were the ACVP tests passing for PSS if we didn't have this check in the `if` condition.

samuel40791765: Oh we already had the check, I just forgot to add it back in when porting over BoringSSL's changes. 
But on a side note, I think this only checks if the RSA key is a PSS type in the service indicator, the ACVP functionality shouldn't be effected by it either way.

samuel40791765:
The AES key round check for the service indicator was removed in this PR. BoringSSL removed the key round check because the validity of the key size is already checked within the key initialization process. Any invalid size will throw an error. From normal use of AWS-LC/BoringSSL there is no way this value could be wrong.

https://github.com/awslabs/aws-lc/blob/752ee755f10de080c055d87cea9b7d667342a528/crypto/fipsmodule/aes/aes.c#L82-L106

nebeid (approved):

dkostic (approved):

crypto/fipsmodule/service_indicator/service_indicator.c
nebeid: We discussed offline. The ACVP tests were added in https://github.com/awslabs/aws-lc/pull/564/commits/56917817a32e5f0012a0f9e29fbdd7e403f940d5 when the check was added back, see https://github.com/awslabs/aws-lc/blob/56917817a32e5f0012a0f9e29fbdd7e403f940d5/crypto/fipsmodule/service_indicator/service_indicator_test.cc#L1196-L1225

PR #568, opened 2022-07-28
Improve our legacy CI by removing need for squashfs

samuel40791765:
### Issues:
Addresses `CryptoAlg-1166`

### Description of changes: 
While figuring out a legacy build CI for ACCP, I stumbled upon gcc-4.1.3 when using `ubuntu:10.04` and updating its source repos. `ubuntu:10.04` is the oldest ubuntu docker image available from [the official docker repo](https://hub.docker.com/_/ubuntu?tab=tags&page=1). 
Updating this will let us use a more official docker image, and remove the need for `squashfs` to unbundle an ancient ubuntu 7.04 iso file like we were doing. This will also make our CI docker image building faster and more stable. 
Previous issues like the certificates on the docker image being too old still apply. This mean we still have to use `tests/ci/docker_images/linux-x86/build_legacy_image.sh` to pull a cmake  source externally, then pull and build it within the docker image.

### Call-outs:
* `#include <time.h>` has to be included before `#include <fcntl.h>`, or an error will be thrown. 
```
[ 29%] Building C object crypto/CMakeFiles/crypto_objects.dir/bio/socket_helper.c.o                                                                           
[ 29%] Building C object crypto/CMakeFiles/crypto_objects.dir/blake2/blake2.c.o
[ 30%] Building C object crypto/CMakeFiles/crypto_objects.dir/bn_extra/bn_asn1.c.o                                                                            
[ 30%] Building C object crypto/CMakeFiles/crypto_objects.dir/bn_extra/convert.c.o                                                                            
In file included from /usr/include/fcntl.h:38,                                                  
from /home/ubuntu/workplace/aws-lc/crypto/bio/socket_helper.c:23:  
/usr/include/sys/stat.h:372: error: array type has incomplete element type     
[ 30%] Building C object crypto/CMakeFiles/crypto_objects.dir/buf/buf.c.o      
[ 31%] Building C object crypto/CMakeFiles/crypto_objects.dir/bytestring/asn1_compat.c.o                                                                      
[ 31%] Building C object crypto/CMakeFiles/crypto_objects.dir/bytestring/ber.c.o                                                                              
[ 31%] Building C object crypto/CMakeFiles/crypto_objects.dir/bytestring/cbb.c.o 
make[2]: *** [crypto/CMakeFiles/crypto_objects.dir/bio/socket_helper.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....  
make[1]: *** [crypto/CMakeFiles/crypto_objects.dir/all] Error 2                
make[1]: *** Waiting for unfinished jobs....                                   
[ 31%] Built target fipsmodule                                                 
make: *** [all] Error 2     
```
This issue had been found in BoringSSL: https://github.com/tensorflow/tensorflow/issues/12579, and the proposed fix was referenced here: https://github.com/tensorflow/tensorflow/issues/12579#issuecomment-363874684
This fix should be worth taking as it seems to potentially resolve the same issue with other versions of gcc on different platforms. Applying the fix successfully passes the build.

### Testing:
Still need to do an actual CI propogation/test on my local fork, but a local build with the docker image passes successfully.
Run: `tests/ci/run_legacy_build.sh`


crypto/bio/socket_helper.c
```diff
@@ -20,6 +20,7 @@
 
 #if !defined(OPENSSL_TRUSTY)
 
+#include <time.h>
```
bryce-shang, line 23: Build in internal system does not need this, right?

bryce-shang (approved):

crypto/bio/socket_helper.c
samuel40791765: Yeah, I was confused as to why this wasn't needed in the internal system. But it's been brought up as a past issue for BoringSSL in other similar platforms, so I thought the change would be worth pushing forward.

andrewhop (dismissed):

bryce-shang (approved):

DominicDams (approved):

PR #570, opened 2022-07-29
Fix parse issue in delocate.peg

manastasova:
### Issues:
Resolves parsing issues for ARMv8 assembly with clang7 on ubuntu 20.04 (Note: found through PR https://github.com/awslabs/aws-lc/pull/566 CI test fails).

### Description of changes: 

- Fix parsing issue in `delocate.peg` for ARM assembly.
- Edit rule `RegisterOrConstant` to allow shifting a register/constant by two digit value (e.g., the case of ARMv8 mask for  SHA3 hardware support) instead of just one digit.
- Add a new rule for allowing addition, subtraction and multiplication in the offset. (Note: useful for looping address accesses, e.g., `#8*($i+2)`). Add a set of `OffsetOperator` to define the operations allowed in the offset. Add a new set of `Offset` rule operations interpreted depending on parenthesis location, if added. 

### Call-outs:
The parenthesis in the `Offset` rule should be included both or should not be included at all, i.e., the parenthesis set should be closed. The `OffsetOperator` includes addition, subtraction and multiplication only.

### Testing:
Testing is added by PR https://github.com/awslabs/aws-lc/pull/566 which adds a new assembly file for ARMv8. The changes fix the parsing compile error generated when clang7 is used on ubuntu 20.04 in the fips static build. 

### DIFF
--- a/util/fipstools/delocate/delocate.peg
+++ b/util/fipstools/delocate/delocate.peg
@@ -100,7 +100,7 @@ ARMGOTLow12 <- ":got_lo12:" SymbolName
 ARMPostincrement <- '!'
 BaseIndexScale <- '(' RegisterOrConstant? WS? (',' WS? RegisterOrConstant WS? (',' [0-9]+)? )? ')'
 Operator <- [+\-]
-OperatorOffset <- '+' / '-' / '*'
-Offset <- '+'? '-'? (("0b" [01]+) / ("0x" [[0-9A-F]]+) / [0-9]+ ( OperatorOffset '(' [0-9]+ OperatorOffset [0-9]+ ')' )? / [0-9]+ ( OperatorOffset [0-9]+ OperatorOffset [0-9]+ )? / [0-9]+ ( OperatorOffset [0-9]+ )? )
+OffsetOperator <- '+' / '-' / '*'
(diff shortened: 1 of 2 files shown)


nebeid:
The looping address accesses is expanded when processed by Perl, right? i.e. I understand `$i` won't show in the resulting assembly.

util/fipstools/delocate/delocate.peg
```diff
@@ -100,6 +100,7 @@ ARMGOTLow12 <- ":got_lo12:" SymbolName
 ARMPostincrement <- '!'
 BaseIndexScale <- '(' RegisterOrConstant? WS? (',' WS? RegisterOrConstant WS? (',' [0-9]+)? )? ')'
 Operator <- [+\-]
-Offset <- '+'? '-'? (("0b" [01]+) / ("0x" [[0-9A-F]]+) / [0-9]+)
+OperatorOffset <- '+' / '-' / '*'
```
nebeid, line 103: Nit: If this is the operator acting on the offset, should it be called `OffsetOperator`?

nebeid:
> The looping address accesses is expanded when processed by Perl, right? i.e. I understand `$i` won't show in the resulting assembly.

manastasova:
> The looping address accesses is expanded when processed by Perl, right? i.e. I understand `$i` won't show in the resulting assembly.



> Yes, the variable is extended back to its actual value (e.g., 0, 1, 2 ...).

nebeid (approved):

torben-hansen (approved):

util/fipstools/delocate/delocate.peg
manastasova: Fixed. Thanks.

PR #571, opened 2022-07-29
Removing the string literal from within the macro.

nebeid:
To cater for gcc version 4.8.5 (Ubuntu 4.8.5-4ubuntu8).
Note that gcc version 4.8.5 20150623 (Red Hat 4.8.5-44)
doesn't complain.

### Issues:
A problem was reported with the usage of gcc version 4.8.5 (Ubuntu 4.8.5-4ubuntu8):
```
crypto/x509/x509_test.cc:4364:20: error: unterminated raw string
   EXPECT_EQ(print, R"(Certificate:
                    ^
```
### Description of changes: 
The string literal was put in a constant in order to avoid having it in a macro causing a problem like the following:
https://godbolt.org/z/Wc8TYaqbf

### Call-outs:
The version used in our CI in Centos-7 Docker image is gcc version 4.8.5 20150623 (Red Hat 4.8.5-44) which does not incur this error.

### Testing:
It's not tested with our CI so pending customer testing for now.

### DIFF
--- a/crypto/x509/x509_test.cc
+++ b/crypto/x509/x509_test.cc
@@ -4361,7 +4361,7 @@ TEST(X509Test, Print) {
   size_t data_len;
   ASSERT_TRUE(BIO_mem_contents(bio.get(), &data, &data_len));
   std::string print(reinterpret_cast<const char*>(data), data_len);
-  EXPECT_EQ(print, R"(Certificate:
+  static const char expected_certificate_string[] = R"(Certificate:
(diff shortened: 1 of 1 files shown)


DominicDams (approved):

torben-hansen:
Is the tested?

bryce-shang (approved):

bryce-shang:
`V649762450` verified this fix works.

PR #567, opened 2022-07-26
Added endianess checks

DominicDams:
### Issues:
Resolves #557 
### Description of changes: 
Currently we have a list of allowed architectures to prevent people from trying to build on Big Endian systems since we don't support them. It's simpler to instead explicitly disallow Big Endian systems and have unknown architectures compile as generic builds.

### Call-outs:
Point out areas that need special attention or support during the review process. Discuss architecture or design changes.

### Testing:
Used https://github.com/dockcross/dockcross to verify that this fails for Big Endian mips. Also checked that mipsel builds despite not being explicitly allowed

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #577, opened 2022-08-03
Added sanitizer specific images

DominicDams:
### Issues:
The Sanitizer tests need access to `LLVMgold.so` in some situations in
order to properly perform the endianess checks.

### Description of changes: 
This change adds a new Docker image that will have access to `LLVMgold.so` and is a prerequisite to adding our endianess checks. It also updates the sanitizer tests to use this image

### Call-outs:
This shouldn't change the requirements for user builds of AWS-LC in any way.

### Testing:
The docker images were build locally and the CI test that was failing before was verified to succeed on these new images.


dkostic:
I'm not sure that we need to add a new Docker image. As far as I understand, the problem with #567 is that when we build aws-lc with clang sanitizers enabled then CMake fails to perform its builtin endianness check. The reason it fails is because it can't find system headers `sys/types.h, stdint.h, ...`. To fix this you install `llvm-dev` on the system, but you can probably fix it by installing `libc6-dev` because those headers are originally from libc (I don't think the issue has anything to do with `llvm-gold` linker). Anyway, can we just not perform the CMake endianness check when sanitizers are used and avoid adding a new Docker image?

DominicDams:
> I'm not sure that we need to add a new Docker image. As far as I understand, the problem with #567 is that when we build aws-lc with clang sanitizers enabled then CMake fails to perform its builtin endianness check. The reason it fails is because it can't find system headers `sys/types.h, stdint.h, ...`. To fix this you install `llvm-dev` on the system, but you can probably fix it by installing `libc6-dev` because those headers are originally from libc (I don't think the issue has anything to do with `llvm-gold` linker). Anyway, can we just not perform the CMake endianness check when sanitizers are used and avoid adding a new Docker image?

The Cmake builds are failing with the following error
```
/usr/bin/ld: /usr/lib/llvm-9/bin/../lib/LLVMgold.so: error loading plugin: /usr/lib/llvm-9/bin/../lib/LLVMgold.so: cannot open shared object file: No such file or directory
```
when trying to find the system headers. This seems difficult to recreate, as installing the same versions of clang used by this didn't do anything to recreate this issue, I only was able to recreate the issue by using the docker images. I don't know that it really matters one way or another if we enable the endianess checks when we run the sanitzer tests, I just worry that disabling them seems to be putting something in the codebase in order to work around an issue with one of our docker images.

samuel40791765 (approved):

samuel40791765 (commented):
We add some address sanitizer stuff in the base docker image already, maybe adding this `llvm-9-dev` to this docker image would prevent the need of an extra docker image?
https://github.com/awslabs/aws-lc/blob/dc5ad88a1f131aa39c2b19fb3134bd2c49f41629/tests/ci/docker_images/linux-aarch/ubuntu-20.04_base/Dockerfile#L30-L31

https://github.com/awslabs/aws-lc/blob/817cd4e1dee86f0738db0f090607b7a757937e1c/tests/ci/docker_images/linux-x86/ubuntu-20.04_base/Dockerfile#L30-L31

PR #592, opened 2022-08-18
Aes xts into fips

Taffer:
### Issues:

Resolves CryptoAlg-1223

### Description of changes: 

Moves AES-XTS into the FIPS boundary, adds a service indicator test for the AES-XTS cipher. Extends the cipher tests in the service indicator to let each cipher specify its IV, plaintext, and expected ciphertext. Extends service indicator checks into `EVP_EncryptUpdate()` and `EVP_DecryptUpdate()`.

### Testing:

`run_tests` pass in non-FIPS and FIPS builds, I look forward to seeing what CI has to say.


crypto/fipsmodule/cipher/cipher.c
```diff
@@ -424,6 +427,8 @@ int EVP_DecryptUpdate(EVP_CIPHER_CTX *ctx, uint8_t *out, int *out_len,
     *out_len += b;
   }
 
+out:
+  EVP_Cipher_verify_service_indicator(ctx);
```
samuel40791765, line 431: It's my fault for not testing this, but `EVP_EncryptUpdate` and `EVP_DecryptUpdate` shouldn't be able to update the state of the service indicator. Our discussions behind this were that the "service" hasn't finished during `EVP_*Update`, and one could only know if the entire "service" operation had finished when the user called `EVP_*Final`. 
That being said, maybe we should add a test similar to the code below in `DoCipherFinal` (`crypto/fipsmodule/service_indicator/service_indicator_test.cc`)
https://github.com/awslabs/aws-lc/blob/07f270f995e6ba1dac141d5349d07bfa81ccb623/crypto/fipsmodule/service_indicator/service_indicator_test.cc#L825-L826

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -259,6 +259,35 @@ static const uint8_t kAESCMACOutput[16] = {0xe7, 0x32, 0x43, 0xb4, 0xae, 0x79,
...
+    0xd6, 0x19, 0xb7, 0xe2, 0xb9, 0x14, 0x1e, 0x58, 0x31, 0x8e, 0xea, 0x39,
+    0x2c, 0xf4, 0x1b, 0x08
+};
+
+static const uint8_t kAESXTSIV_256[16] = {
+    0xad, 0xf8, 0xd9, 0x26, 0x27, 0x46, 0x4a, 0xd2, 0xf0, 0x42, 0x8e, 0x84,
+    0xa9, 0xf8, 0x75, 0x64
+};
+
+static const uint8_t kAESXTSPlaintext_256[32] = {
```
samuel40791765, line 279: I'm not too familiar with the semantics of `XTS`, but is AES-XTS only able to take in 32 bit test lengths?

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -259,6 +259,35 @@ static const uint8_t kAESCMACOutput[16] = {0xe7, 0x32, 0x43, 0xb4, 0xae, 0x79,
                                            0x08, 0x86, 0xe7, 0x9f, 0x0d, 0x3f,
                                            0x88, 0x3f, 0x1a, 0xfd};
 
+// AES-XTS test vector from
+// https://csrc.nist.gov/projects/cryptographic-algorithm-validation-program
```
samuel40791765, line 263: We can use the same test inputs in `CipherTestVector` if possible (`kAESKey_256`, `kPlaintext`, etc.). In my mind, the service indicator tests were mainly used for testing correctness of the service indicator. 
Using test vectors from Nist are nice though, we could add those to `xts_test.cc`?

crypto/fipsmodule/service_indicator/service_indicator_test.cc
Taffer: I also have no idea, I'm just moving code around. I'll look at [NIST SP800-38e](https://csrc.nist.gov/publications/detail/sp/800-38e/final) and see if I can find out.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
Taffer: Good idea, I'll add some NIST test vectors to `xts_test.cc`.

It's probably also a good idea to use NIST KATs in `service_indicator_test.cc` as well, I'll open a new ticket for that.

Taffer: CryptoAlg-1265 for FIPS KATs in the service indicator.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
Taffer: XTS mode operates on fixed size blocks (<= 2^20^ AES blocks); this size doesn't need to be a multiple of the AES block size, the incomplete block at the end will be padded out with ciphertext stealing. The output ciphertext will be the same size as the input plaintext.

Keys must be 128 or 256 bits. I'll add some additional test cases.

Taffer: Looks like the AWS-LC implementation requires 256-bit keys and a multiple of the AES block size for the data (tested with 256, 384, 140, 250, 384-bit data from the NIST KATs).

crypto/fipsmodule/cipher/cipher.c
Taffer: I've backed out the `EncryptUpdate`/`DecryptUpdate` changes and (I think) modified `DoCipherFinal()` as you've suggested.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
samuel40791765: Thanks for all the details, that was really enlightening.

DominicDams (approved):

samuel40791765 (approved):
Nit: If we add NIST KATs to `service_indicator_test.cc`, I think we could add them to a separate parameterized test suite. It would be cleaner to see `EVPServiceIndicatorTest` test upon the same `kPlaintext`, `kAESIV`, `kAESkey_256`, etc. IMO. But this isn't a blocker and more of a test styling nitpick on my end.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -662,117 +693,168 @@ static const struct CipherTestVector {
...
+        kAESXTSKey_256,
+        sizeof(kAESXTSKey_256),
+        kAESXTSIV_256,
+        sizeof(kAESXTSIV_256),
+        kAESXTSPlaintext_256,
+        sizeof(kAESXTSPlaintext_256),
+        kAESXTSCiphertext_256,
+        sizeof(kAESXTSCiphertext_256),
+        AWSLC_APPROVED,
+    },
```
nebeid, line 835: Can we add a test vector with AWSLC_NOT_APPROVED as the expected result? For example, this negative test case: https://github.com/Taffer/aws-lc/blob/4e1fd6c8f10d21a374a6539b8f4d93c26b0e88ad/crypto/fipsmodule/modes/xts_test.cc#L543

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -314,6 +314,8 @@ void EVP_Cipher_verify_service_indicator(const EVP_CIPHER_CTX *ctx) {
     case NID_aes_128_ctr:
     case NID_aes_192_ctr:
     case NID_aes_256_ctr:
+
+    case NID_aes_256_xts:
```
nebeid, line 318: `EVP_Cipher_verify_service_indicator()` is called within `EVP_Cipher()`, `EVP_EncryptFinal_ex()` and `EVP_DecryptFinal_ex()`. However, in the case of XTS, the `Final` functions are NOP, will we check in `Update` whether the cipher is xts to call `EVP_Cipher_verify_service_indicator`? Note also that in the case of xts, update can be called only once after init, it cannot be called repeatedly; it does the whole encryption in one shot to do the cipher stealing at the end.
I don't know if all the other ciphers listed here need to call final or not, but I guess yes.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -799,11 +882,19 @@ static void TestOperation(const EVP_CIPHER *cipher, bool encrypt,
   // |EVP_EncryptFinal_ex| and |EVP_DecryptFinal_ex| for approval at the end.
```
nebeid, line 882: Sorry, this may have come up previously. The comment says we test 
> |EVP_EncryptFinal_ex| and |EVP_DecryptFinal_ex| for approval at the end.
But I don't see we're calling Init and Update to test that they won't set Approved, then call Final to test that it does set it. Is that done elsewhere?
Also, see my comment above in the case of xts particularly where update should return approved, unless we force the FIPS user to call final which would be a NOP but just to return Approved (to be tested that this is what will happen).

crypto/fipsmodule/service_indicator/service_indicator.c
samuel40791765: In my opinion, I would keep the original service indicator implementation of leaving the checks within the `*Final` operations. `*Final` might be a no-op for XTS, but it was within our general service indicator usage guidance that this "*Final" step was our way of differentiating if the user had finished the "FIPS service". This general guidance goes for all the crypto functions within AWS-LC.
If we tried to add a check within `EVP_Update`, it would add additional complexity for us to maintain this outlier and our service indicator usage guidance would have to be updated as well.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
samuel40791765: Yes this test was when of the first versions of the service indicator tests, so we hadn't checked for non-approval for `Init` and `Update`. "Taffer" has added a new non-approval check for `Update` on Line 116, but we should add one for `init` as well.

PR #594, opened 2022-08-19
Fix aws_lc_sys::ERR_xxx; support optimized build

justsmth:
### Description of changes: 
* `rust_wrapper.c` was not being built. The file is only present with being built for the aws-lc-sys crate.
* Added test to verify the usability of `aws_lc_sys::ERR_get_error()`
* Support a "release" (optimized) build of the crate.

### Call-outs:
N/A

### Testing:
I generated and tested the `aws_lc_sys` crate locally.


samuel40791765 (approved):

bindings/rust/aws-lc-sys-template/build.rs
```diff
@@ -217,6 +217,15 @@ fn prepare_cmake_build(build_prefix: Option<&str>) -> cmake::Config {
...
     let mut cmake_cfg = get_cmake_config();
 
+    let opt_level = env::var("OPT_LEVEL").unwrap_or_else(|_| "0".to_string());
+    if opt_level.ne("0") {
+        if opt_level.eq("1") || opt_level.eq("2") {
+            cmake_cfg.define("CMAKE_BUILD_TYPE", "relwithdebinfo");
+        } else {
+            cmake_cfg.define("CMAKE_BUILD_TYPE", "release");
+        }
+    }
```
bryce-shang, line 227: - What if `OPT_LEVEL` is not visible?
- Is the mapping from `1/2` to `relwithdebinfo` defined by us or defined by some docs?

bindings/rust/aws-lc-sys-template/tests/sanity-tests.rs
```diff
@@ -26,3 +26,16 @@ fn sha1() {
...
+#[test]
+fn error_checking() {
+    unsafe {
+        let error = aws_lc_sys::ERR_get_error();
+        let err_lib = aws_lc_sys::ERR_GET_LIB(error);
+        let err_reason = aws_lc_sys::ERR_GET_REASON(error);
+        let err_func = aws_lc_sys::ERR_GET_FUNC(error);
+        assert_eq!(err_lib, 0);
+        assert_eq!(err_reason, 0);
+        assert_eq!(err_func, 0);
```
bryce-shang, line 39: Not sure why not equal 0.
```c
// ERR_get_error gets the packed error code for the least recent error and
// removes that error from the queue. If there are no errors in the queue then
// it returns zero.
OPENSSL_EXPORT uint32_t ERR_get_error(void);
```

bryce-shang (approved):

bindings/rust/aws-lc-sys-template/build.rs
justsmth: * Cargo documents the environment variable here: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts
* The possible values for opt-level are documented here: https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level
    * Its values are 0, 1, 2, 3, "s", or "z".

bindings/rust/aws-lc-sys-template/tests/sanity-tests.rs
justsmth: It should be 0 b/c no errors have been generated. I mostly included this test to ensure that the function is being linked properly.  (Prior to my changes to CMakeLists.txt, this test would fail to compile due to a linker error.)

PR #598, opened 2022-08-22
Fix for the static library constructor

dkostic:
When linking with a static library, if no symbols in an object file are
referenced then the object file is discarded, even if it has a constructor
function. For example, if an application is linking with libcrypto.a and
not referencing any symbol from crypto.o file, then crypto.o will be
discarded. This is an issue because we define the library constructor
in crypto.o so if the file is discarded then the library is not initialized.
Note that this is not a problem for the FIPS build because when building the
FIPS mode we have to ensure the initialization is done before the power-on
self-tests so the test function itself calls |OPENSSL_cpuid_setup|.

So, the issue manifests only in the static non-FIPS build. The work around
is we add a dummy function |dummy_func_for_constructor| in |bcm.c| that
calls |CRYPTO_library_init| function defined in |crypto.c|. This will ensure
that, when linking with libcrypto.a, crypto.o will not be discarded as long
as bcm.o is not discarded.

This workaround is partial in a sense that if the application that's linking
to libcrypto.a is not using any of the symbols from bcm.o or crypto.o then
both object files will be discarded. But this would be an edge case that we
don't expect to happen with significant probability. In case it happens, the
application would have to call the |CRYPTO_library_init| function itself to
ensure the initialization is done.

### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 

### Call-outs:

### Testing:


DominicDams (approved):

crypto/fipsmodule/bcm.c
```diff
@@ -136,6 +136,36 @@
...
+//
+// This workaround is partial in a sense that if the application that's linking
+// to libcrypto.a is not using any of the symbols from bcm.o then both bcm.o
+// and crypto.o will be discarded. But this would be an edge case that we don't
+// expect to happen with significant probability. In case it happens, the
+// application would have to call the |CRYPTO_library_init| function itself to
+// ensure the initialization is done.
+extern uint8_t dummy_var_for_constructor;
+
+void dummy_func_for_constructor() {
```
DominicDams, line 164: I think you need
```suggestion
void dummy_func_for_constructor(void) {
```

nebeid (commented):
I suggest keeping the comment in one file and referring to it in other files.

bryce-shang:
Nit: There is related bug link(35) in upstream. Not sure what the original concern is and if it's still worth.

torben-hansen:
Instead of using a dummy variable, just call `CRYPTO_library_init()` in bcm.c? Is that enough?

It is only called once per process.

nebeid:
> Nit: There is related bug link(35 ?) in upstream. Not sure what the original concern is and if it's still worth.

https://bugs.chromium.org/p/boringssl/issues/detail?id=35
As far as I understand, It seems to be similar concern of forgetting to initialise the library and not being sure who owns the initialisation.

nebeid:
Related to Torben's comment, do we need to keep `do_library_init()`in `crypto.c` as the constructor when not in FIPS build, or we can make the constructor in `bcm.c` be always the constructor which initialises the library and, only in FIPS build, run `jent_entropy_init()` and the integrity and self-test.

dkostic:
following the discussions we had offline, I moved the constructor to bcm in the latest commit (we'll see if the CI is happy with that).

bryce-shang:
> we'll see if the CI is happy with that

Not sure what this means. CI relies on new dimensions(e.g. tests) added by each PR.

dkostic:
> > we'll see if the CI is happy with that
> 
> Not sure what this means. The `happy` may need new dimensions(e.g. tests) added in each PR. The testing section in this PR provides no context.

what I meant is, I first need to make the existing tests in the CI happy with the change (windows/android/apple,...). I'm still thinking about how to add a new test that can test if the constructor is always executed. Not sure how to do it best. If you have ideas let me know

crypto/crypto.c
```diff
@@ -18,70 +18,7 @@
...
-
-void CRYPTO_library_init(void) {
-  // TODO(davidben): It would be tidier if this build knob could be replaced
-  // with an internal lazy-init mechanism that would handle things correctly
-  // in-library. https://crbug.com/542879
-#if defined(BORINGSSL_NO_STATIC_INITIALIZER)
-  CRYPTO_once(&once, do_library_init);
-#endif
-}
+void CRYPTO_library_init(void) { }
```
nebeid, line 21: Do we need to keep this here?

nebeid (commented):
I was wondering whether we need to remove `CRYPTO_library_init()` from `/fipstools/test_fips.c`, `test/gtest_main.h`, `test/bssl_shim.cc`, as well as from `SSL_library_init()` in `ssl/ssl_lib.cc`. We would also need to run the speed test to make sure our benchmarks didn't change if we remove it from `test/gtest_main.h`.

dkostic:
> I was wondering whether we need to remove `CRYPTO_library_init()` from `/fipstools/test_fips.c`, `test/gtest_main.h`, `test/bssl_shim.cc`, as well as from `SSL_library_init()` in `ssl/ssl_lib.cc`. We would also need to run the speed test to make sure our benchmarks didn't change if we remove it from `test/gtest_main.h`.

We can remove those calls, but we don't need to. `CRYPTO_library_init` becomes a no-op with this change so no harm calling it.

crypto/crypto.c
dkostic: we need define it not to break customers who might be calling the function.

crypto/fipsmodule/bcm.c
```diff
@@ -342,3 +329,45 @@ void BORINGSSL_FIPS_abort(void) {
...
+
+#if defined(OPENSSL_WINDOWS)
+#define OPENSSL_CDECL __cdecl
+#else
+#define OPENSSL_CDECL
+#endif
+
+static void OPENSSL_CDECL do_library_init(void) {
+
+  OPENSSL_cpuid_setup();
```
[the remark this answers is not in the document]
dkostic: good point, I'll take a look

DominicDams (dismissed):

dkostic:
As discussed offline, I updated the PR with the simple solution that Torben suggested (dummy function in bcm.c calling a function from crypto.c). Also added a CI test.

DominicDams (dismissed):

crypto/fipsmodule/bcm.c
```diff
@@ -342,3 +342,31 @@ void BORINGSSL_FIPS_abort(void) {
...
+// not referencing any symbol from crypto.o file, then crypto.o will be
+// discarded. This is an issue because we define the library constructor
+// in crypto.o so if the file is discarded then the library is not initialized.
+// Note that this is not a problem for the FIPS build because when building the
+// FIPS mode we have to ensure the initialization is done before the power-on
+// self-tests so the test function itself calls |OPENSSL_cpuid_setup|.
+//
+// So, the issue manifests only in the static non-FIPS build. The work around
+// it we add a dummy function |dummy_func_for_constructor| in |bcm.c| that
+// references a symbol, |CRYPTO_library_init| from |crypto.c|. This will ensure
```
nebeid, line 359:
```suggestion
// calls the initialisation function |CRYPTO_library_init| defined in |crypto.c|. This will ensure
```
Same suggestion applies to the comment in internal.h, where I suggest to keep the comment and reference it here and in cpucap.c or wherever you prefer, but I think one copy is sufficient.

crypto/fipsmodule/cpucap/cpucap.c
```diff
@@ -82,3 +82,4 @@ HIDDEN uint32_t OPENSSL_armcap_P = 0;
 HIDDEN uint8_t BORINGSSL_function_hit[7] = {0};
 #endif
 
+HIDDEN uint8_t OPENSSL_cpucap_initialized = 0;
```
nebeid, line 85: Is this variable used only in testing? If so, may be we can document that here.

crypto/fipsmodule/bcm.c
dkostic: fixed this. Removed the comment in internal.h and added a reference to the one in bcm.c

crypto/fipsmodule/cpucap/cpucap.c
dkostic: done

nebeid (dismissed):
I think the PR description can be updated similarly to the comment. Thanks, dkostic.

crypto/fipsmodule/bcm.c
```diff
@@ -342,3 +342,31 @@ void BORINGSSL_FIPS_abort(void) {
 }
 
 #endif  // BORINGSSL_FIPS
+
+#if !defined(BORINGSSL_FIPS) && !defined(BORINGSSL_SHARED_LIBRARY)
+// When linking with a static library, if no symbols in an object file are
+// referenced then the object file is discarded, even if it has constructor
```
torben-hansen, line 348:
```suggestion
// referenced then the object file is discarded, even if it has a constructor
```

crypto/fipsmodule/bcm.c
```diff
@@ -342,3 +342,31 @@ void BORINGSSL_FIPS_abort(void) {
...
+// function. For example, if an application is linking with libcrypto.a and
+// not referencing any symbol from crypto.o file, then crypto.o will be
+// discarded. This is an issue because we define the library constructor
+// in crypto.o so if the file is discarded then the library is not initialized.
+// Note that this is not a problem for the FIPS build because when building the
+// FIPS mode we have to ensure the initialization is done before the power-on
+// self-tests so the test function itself calls |OPENSSL_cpuid_setup|.
+//
+// So, the issue manifests only in the static non-FIPS build. The work around
+// it we add a dummy function |dummy_func_for_constructor| in |bcm.c| that
```
torben-hansen, line 358:
```suggestion
// is we add a dummy function |dummy_func_for_constructor| in |bcm.c| that
```

crypto/fipsmodule/cpucap/cpucap.c
```diff
@@ -82,3 +82,6 @@ HIDDEN uint32_t OPENSSL_armcap_P = 0;
 HIDDEN uint8_t BORINGSSL_function_hit[7] = {0};
 #endif
 
+// This variable is used only for testing purposes to ensure that the library
+// constructor is executed and the capability variable is initialized.
+HIDDEN uint8_t OPENSSL_cpucap_initialized = 0;
```
torben-hansen, line 87: Shouldn't we turn this into a public assertion function? Ala
```
int AWSLC_is_initialized()
```
Keep the variable stuff in `crypto.o` instead of spreading it around in multiple files.

Organisationally, I think `crypto.o` should handle the init logic, so is it possible to bound it to that object?

torben-hansen: Also prevent us from having to add `OPENSSL_cpucap_initialized` to every new (rare, I know) cpu capability implementation.

dkostic: I thought about that and I like the idea of adding it to `crypto.o`. The problem is that then the initialization test I added in this PR would always return success, because if you are calling `AWSLC_is_initialized` which is in `crypto.o` then `crypto.o` will not be discarded. So yeah, I agree it's better to have `AWSLC_is_initialized` in `crypto.o` but I don't see a way to test this in the CI :)

crypto/internal.h
```diff
@@ -1026,6 +1026,10 @@ OPENSSL_INLINE int boringssl_fips_break_test(const char *test) {
 extern uint8_t BORINGSSL_function_hit[7];
 #endif  // BORINGSSL_DISPATCH_TEST
 
+#if !defined(AWSLC_FIPS) && !defined(BORINGSSL_SHARED_LIBRARY)
```
torben-hansen, line 1029: You can declare this in `bcm.c` if you want to.

crypto/fipsmodule/cpucap/cpucap.c
torben-hansen, line 87: Not sure I like all these variables. But I can't come up with a better way to do it myself... so...

torben-hansen (approved):

PR #601, opened 2022-08-24
allow static or shared libs to be used

graebm:
### Issues:
Partially addresses https://github.com/aws4embeddedlinux/meta-aws/pull/191 (many other repos need a similar fix)

### Description of changes: 

Currently, a static lib that uses AWS-LC must use a static build of AWS-LC, it cannot use a shared build.

And vice-versa, a shared lib that uses AWS-LC must use a shared build of AWS-LC, it cannot use a static build.

With this change, either a static or shared build of AWS-LC can be used. If both are installed, then BUILD_SHARED_LIBS serves as a tie-breaker between the two.

### Call-outs:
n/a

### Testing:
Tested locally mixing static and shared build of AWS-LC with static and shared builds of a lib that uses AWS-LC


crypto/cmake/crypto-config.cmake
```diff
@@ -2,11 +2,22 @@ if(WIN32 OR UNIX OR APPLE)
...
+# Allow static or shared lib to be used.
+# If both are installed, choose based on BUILD_SHARED_LIBS.
 if (BUILD_SHARED_LIBS)
-    message(STATUS "FOUND AWS-LC CRYPTO cmake config - shared")
-    include(${CMAKE_CURRENT_LIST_DIR}/shared/crypto-targets.cmake)
+    if (EXISTS "${CMAKE_CURRENT_LIST_DIR}/shared/crypto-targets.cmake")
+        message(STATUS "FOUND AWS-LC CRYPTO cmake config - shared")
+        include(${CMAKE_CURRENT_LIST_DIR}/shared/crypto-targets.cmake)
+    else()
+        message(STATUS "FOUND AWS-LC CRYPTO cmake config - static")
```
torben-hansen, line 12: This was present in existing code, but is this message accurate? We don't actually know yet...

crypto/cmake/crypto-config.cmake
```diff
@@ -2,11 +2,22 @@ if(WIN32 OR UNIX OR APPLE)
     find_package(Threads REQUIRED)
 endif()
 
+# Allow static or shared lib to be used.
+# If both are installed, choose based on BUILD_SHARED_LIBS.
 if (BUILD_SHARED_LIBS)
```
torben-hansen, line 7: Another way I could think of doing this is adding an extra "force"/"override" condition:
```suggestion
if (BUILD_SHARED_LIBS OR FORCE_USING_SHARED_LIBS)
```

But, you prefer it being implicit?

crypto/cmake/crypto-config.cmake
graebm: Good point!
I moved the `message()` after the `include()`

crypto/cmake/crypto-config.cmake
graebm: The existing logic prevents a user from building their own library as static when `libcrypto.so` is preinstalled from a package manager (see ticket in PR description).

With my change, whichever version of libcrypto is installed, that's what gets picked. That's what users expect.

But this does introduce ambiguity when both `libcrypto.so` and `libcrypto.a` are installed, so in that case I'm keeping the old behavior where BUILD_SHARED_LIBS drives the decision, to maintain backwards compatibility with how things were before.

I'm hesitant to add an explicit option. I spent a long time trying things out but I can see downsides to all approaches. We can add an explicit option later when we've really thought through all the issues. Nothing I'm doing here prevents that

torben-hansen: Makes sense.

torben-hansen:
I think we need a test to verify this doesn't break in the future.

samuel40791765:
Could we add an integration script or test to help us ensure this works? It'll also help us continue maintaining this behavior in the future.
An example of a customer static lib that builds with AWS-LC's shared library and a customer shared lib that builds with AWS-LC's static library would do. We have [an integration test](https://github.com/awslabs/aws-lc/blob/main/tests/ci/run_s2n_integration.sh) that builds with s2n here, perhaps we could leverage that as a test example.

graebm:
yeah, I got similar feedback from S2N (have a similar PR there): https://github.com/aws/s2n-tls/pull/3467

I'll work on something...

graebm:
I added a new test similar to the s2n test

Some of the checks are failing, but I don't have permissions to see the logs. These checks were failing even before I added the new test, so no idea if it's related

samuel40791765 (dismissed):

tests/ci/cdk/cdk/codebuild/github_ci_linux_x86_omnibus.yaml
```diff
@@ -232,6 +232,14 @@ batch:
         compute-type: BUILD_GENERAL1_LARGE
         image: 620771051181.dkr.ecr.us-west-2.amazonaws.com/aws-lc-docker-images-linux-x86:ubuntu-20.04_clang-9x_latest
 
+    - identifier: install_shared_and_static
```
torben-hansen, line 235: Okay, I thought we might just tag it on to some other script that ran. But I guess we can use this dimension for something in the future as well.

tests/ci/run_install_shared_and_static.sh
```diff
@@ -0,0 +1,133 @@
...
+# Assumes script is executed from the root of aws-lc directory
+AWS_LC_DIR=$(pwd)
+ROOT=$(realpath ${AWS_LC_DIR}/..)
+
+SCRATCH_DIR=${ROOT}/SCRATCH_AWSLC_INSTALL_SHARED_AND_STATIC
+mkdir -p ${SCRATCH_DIR}
+rm -rf ${SCRATCH_DIR}/*
+
+function fail() {
+    echo "testf failure: $1"
```
torben-hansen, line 34:
```suggestion
    echo "test failure: $1"
```
?

tests/ci/run_install_shared_and_static.sh
```diff
@@ -0,0 +1,133 @@
...
+    local INSTALL_DIR=${SCRATCH_DIR}/$1
+    local BUILD_SHARED_LIBS=$2
+
+    local BUILD_DIR=${SCRATCH_DIR}/build
+    rm -rf ${BUILD_DIR}
+    ${CMAKE_COMMAND} -H${AWS_LC_DIR} -B${BUILD_DIR} -GNinja -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR} -DBUILD_TESTING=OFF -D${BUILD_SHARED_LIBS}
+    ${CMAKE_COMMAND} --build ${BUILD_DIR} --target install
+}
+
+# create installation with libssl.so
```
torben-hansen, line 48: You mainly depends on `libcrypto`, no? :)

torben-hansen: But `librypto` is a dependency of `libssl`, so maybe that is the motivation?

graebm: Yes, I personally am only interested in libcrypto

But this PR tweaks both both ssl-config.cmake and crypto-config.cmake.

So I'm testing both here.

I tweaked the comments to always mention both libssl and libcrypto

tests/ci/cdk/cdk/codebuild/github_ci_linux_x86_omnibus.yaml
torben-hansen: I think we can keep as-is :)

torben-hansen (approved):

samuel40791765 (approved):

PR #603, opened 2022-08-26
Pbkdf2 service indicator

Taffer:
### Issues:
Resolves CryptoAlg-1270

### Description of changes: 
Adds service indicator tests and self_check tests for PBKDF2.

### Call-outs:
The PBKDF2 service indicator now enforces FIPS parameter requirements. Also adjusts the benchmarking code to use FIPS-valid password/salt values so we can directly compare FIPS-mode vs normal speeds in scrypt.

### Testing:
Adds 30 tests to service indicator and one to self_check.


crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -346,24 +346,26 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
-  // * key_len >= 14 bytes (112 bits)
+  // * password_len >= 14 bytes (112 bits)
   // * salt_len >= 16 bytes (128 bits), assuming its randomly generated
   // * iterations "as large as possible, as long as the time required to
   //   generate the key using the entered password is acceptable for the users."
-  //   (clearly we can't test for "as large as possible")
+  //   (clearly we can't test for "as large as possible"); NIST SP800-132
+  //   suggests >= 1000, but it's still not a requirement.
   switch (evp_md->type) {
     case NID_sha1:
```
andrewhop, line 361: Is SHA1 still approved? SP 800-131A section 9 doesn't make it clear KDF is an approved use, does PBKDF2 need collision resistance?

Taffer: AFAIK SHA-1 is still in the approved list, presumably for "legacy" uses. I think the weaker collision resistance is mitigated somewhat by password/salt/iterations being in the mix.

Taffer: According to the NIST ACVP testing data, SHA-1 is approved: https://pages.nist.gov/ACVP/draft-celi-acvp-pbkdf.html#section-7.3.1

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -1967,6 +1967,451 @@ TEST_P(KDF_ServiceIndicatorTest, TLSKDF) {
   EXPECT_EQ(approved, test.expect_approved);
 }
 
+// PBKDF2 test data from RFC 6070.
```
nebeid, line 1970: maybe put a comment here or in the tests that Salt1 and Password1 are too short and will cause the service indicator to be NOT_APPROVED.
Also a comment about Salt3 and Password3 would be good as I don't know what they add. Aren't they short too?

Taffer: They are; the only difference is that they include `NUL` characters, presumably to ensure the password/salt are being handled as byte buffers instead of C strings. I'll remove set 3 as it doesn't really add anything to AWS-LC.

nebeid (approved):

DominicDams (approved):

PR #607, opened 2022-08-30
Add HKDF service indicator tests and self check

Taffer:
### Issues:
Resolves CryptoAlg-1280
Addresses #ISSUE-NUMBER2

### Description of changes: 
Adds HKDF tests to the service indicator and self check for FIPS. Also marked made a couple of self-test functions static, as they did not need to be exposed.

### Call-outs:
n/a

### Testing:
n/a


crypto/fipsmodule/hkdf/hkdf.c
```diff
@@ -44,16 +45,26 @@ int HKDF_extract(uint8_t *out_key, size_t *out_len, const EVP_MD *digest,
                  size_t salt_len) {
   // https://tools.ietf.org/html/rfc5869#section-2.2
 
+  // We have to avoid the underlying SHA services updating the indicator
```
samuel40791765, line 48: Small nit, would "HMAC services" be more accurate? Although SHA still is being used under the hood to be fair.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -346,6 +346,54 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
   }
 }
 
+void HKDF_expand_verify_service_indicator(const EVP_MD *evp_md) {
```
samuel40791765, line 349: Can we combine `HKDF_expand_verify_service_indicator` and `HKDF_extract_verify_service_indicator` into one function called `HKDF_verify_service_indicator`? It seems like they're checking against the same parameters.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -1117,6 +1118,491 @@ TEST_P(HMACServiceIndicatorTest, HMACTest) {
...
+    const FIPSStatus expect_approved;
+} kHKDFTestVectors[] = {
+    // RFC 5869 Test Case 1
+    {
+        EVP_sha1,
+        kHKDF_ikm_tc1, sizeof(kHKDF_ikm_tc1),
+        kHKDF_salt_tc1, sizeof(kHKDF_salt_tc1),
+        kHKDF_info_tc1, sizeof(kHKDF_info_tc1),
+        kHKDF_okm_tc1_sha1, sizeof(kHKDF_okm_tc1_sha1),
+        AWSLC_APPROVED,
```
samuel40791765, line 1373: Maybe we should add another test that checks for `AWSLC_NOT_APPROVED` if the `HKDF` function fails

samuel40791765 (commented):
We've recently pulled in the OpenSSL's original EVP_PKEY HKDF APIs from BoringSSL.
https://github.com/awslabs/aws-lc/commit/28ba57ba1f8335ca66d1a53e08fa4c6916fe9ef4
Would it be worth FIPS validating the EVP versions of HKDF as well?

Taffer:
If we add the EVP versions of HKDF as well, it should be in a separate story/PR. See CryptoAlg-1288 for details.

crypto/fipsmodule/service_indicator/service_indicator.c
Taffer: Certainly; I was originally thinking they'd end up testing different things (`extract()` doesn't know about your Info, `expand()` doesn't know about the IKM or Salt), but I couldn't find any specific requirements for sizes on any of them.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
Taffer: Good idea, I'll throw in one using MD5.

crypto/fipsmodule/self_check/self_check.c
```diff
@@ -1026,7 +1056,8 @@ static int boringssl_self_test_fast(void) {
 
   if (!boringssl_self_test_sha256() ||
       !boringssl_self_test_sha512() ||
-      !boringssl_self_test_hmac_sha256()) {
+      !boringssl_self_test_hmac_sha256() ||
```
nebeid, line 1059: The HMAC-256 self test is called first from within the integrity test in bcm.c in https://github.com/awslabs/aws-lc/blame/08c30300a33f9000a0b11c674a8f66e4f0b21c2b/crypto/fipsmodule/bcm.c#L286, so no need to include it here.
Additional side notes:
- The sha256 test is called there as well. We can remove it, but it's ok to leave it for now.
- The hkdf test could have been embedded in this big function because it can be static, but it's ok with me as standalone.

crypto/fipsmodule/hkdf/hkdf.c
```diff
@@ -44,16 +45,26 @@ int HKDF_extract(uint8_t *out_key, size_t *out_len, const EVP_MD *digest,
                  size_t salt_len) {
   // https://tools.ietf.org/html/rfc5869#section-2.2
 
+  // We have to avoid the underlying HMAC services updating the indicator
```
nebeid, line 48: I think we will be able to insert the service indicator checks only in HKDF which does both Expansion and Extraction, and not in each step individually (and when we add it to the EVP interface, it will be only for the default setting of [EVP_PKEY_HKDEF_MODE_EXTRACT_AND_EXPAND] (https://commondatastorage.googleapis.com/chromium-boringssl-docs/kdf.h.html#EVP_PKEY_HKDEF_MODE_EXTRACT_AND_EXPAND))
I'm under that impression from the example ACVP tests in https://pages.nist.gov/ACVP/draft-hammett-acvp-kas-kdf-hkdf.html#name-example-test-vectors-json, which have a final output only.
See also [FIPS 140**-2** IG, Sec G.13, search for "KDA1 (Cert. #A25)"](https://csrc.nist.gov/csrc/media/projects/cryptographic-module-validation-program/documents/fips140-2/fips1402ig.pdf#%5B%7B%22num%22%3A94%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C88%2C415%2C0%5D).

samuel40791765 (dismissed):

crypto/fipsmodule/hkdf/hkdf.c
Taffer: The way CryptoAlg-1198 was written, it seemed that someone was using `HKDF_expand()` stand-alone, so I wanted to cover both the one-shot and the stand-alone entry points.

csosto-pk: It will not suffice to cover just the full HKDF(). 

It is possible to test and validate AWS-LC's HKDF_Expand() as a [KBKDF](https://pages.nist.gov/ACVP/draft-celi-acvp-kbkdf.html ) compliant with SP 800-108 (KDF in Feedback mode) by using `"fixedDataOrder"="after fixed data"` 

Just HKDF() will not suffice because it is not approved for key derivation from an imported key in SP 800-108.

Taffer: I'll add a link to this discussion in the HKDF ACVP story.

samuel40791765 (dismissed):

samuel40791765 (approved):

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -1117,6 +1118,512 @@ TEST_P(HMACServiceIndicatorTest, HMACTest) {
...
+    // RFC Test Case 5 repeats the inputs from RFC 5869 Test Case 2; RFC Test
+    // Case 6 repeats the inputs from RFC 5869 Test Case 3.
+
+    // RFC 5869 Test Case 7
+    //
+    // Info is a zero-length array, thus the hard-coded length.
+    {
+        EVP_sha1,
+        kHKDF_ikm_tc7, sizeof(kHKDF_ikm_tc7),
+        NULL, 0,
```
nebeid, line 1566: Should we not comment out the NULL salt for now with a TODO pointing to CryptoAlg-1281?

Taffer: I'm going to not touch this anymore until we hear back from atsec; I'll convert this PR into a draft so people can ignore it until I've incorporated atsec's suggestions.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -357,6 +358,14 @@ void HKDF_verify_service_indicator(const EVP_MD *evp_md) {
   //
   // We can't test for that, as we don't know what the HKDF output will be
   // used for.
+  //
+  // Per SP800-56Crev2, the salt cannot be empty in FIPS mode; a NULL salt
+  // is replaced in the HMAC with a buffer filled with NUL bytes (0x00), so
+  // that's OK.
+  if (salt != NULL && salt_len == 0) {
```
nebeid, line 365: If I understand correctly, this function is once called with salt and once with info, does the same check apply to both? i.e. info cannot be empty in FIPS mode?
Also, HMAC with key=NULL (salt = NULL) is ok if md = NULL as in https://github.com/awslabs/aws-lc/blame/3f87d09e2b99e4523d60bce18b4fbfb0784f55e0/crypto/fipsmodule/hmac/hmac.c#L255. Can this occur in the calls in HMAC? I guess not. In that case, we may want to add to the comment that `evp_md` is always non-NULL.

Taffer: Based on what atsec said, the `info` argument in `HKDF_expand()` is the incoming salt when it's used in a stand-alone method. It's possible that this isn't quite right and will have to be adjusted once we start testing `HKDF_expand()` via ACVP.

Wow, nothing before this is checking for a `NULL` `md` so the `case` statement just under this would segfault...

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -346,24 +346,60 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
+  // large as the targeted security strength (in bits) of any application that
+  // will be supported by the use of the derived keying material.
+  //
+  // We can't test for that, as we don't know what the HKDF output will be
+  // used for.
+  //
+  // Per SP800-56Crev2, the salt cannot be empty in FIPS mode; a NULL salt
+  // is replaced in the HMAC with a buffer filled with NUL bytes (0x00), so
+  // that's OK.
+  if ((salt != NULL && salt_len == 0) || (evp_md == NULL)) {
```
nebeid, line 365: I think when we reach this point in HKDF(), evp_md will not be null or it would have failed in `HMAC_Init_ex()`, because it wouldn't have been able to initialise the HMAC_CTX ctx.
So we can remove that extra check.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -346,24 +346,60 @@ void HMAC_verify_service_indicator(const EVP_MD *evp_md) {
...
+  // It is recommended that the length of a KDK used by a KDF be at least as
+  // large as the targeted security strength (in bits) of any application that
+  // will be supported by the use of the derived keying material.
+  //
+  // We can't test for that, as we don't know what the HKDF output will be
+  // used for.
+  //
+  // Per SP800-56Crev2, the salt cannot be empty in FIPS mode; a NULL salt
+  // is replaced in the HMAC with a buffer filled with NUL bytes (0x00), so
+  // that's OK.
```
nebeid, line 364:
```suggestion
  // is replaced in the HMAC with a buffer filled with NUL bytes (0x00), 
  // when |evp_md| is not NULL, so that's OK.
  // Note that we reach this check if |HMAC| succeeds which cannot be 
  // the case if |evp_md| = NULL; it would have then failed in |HMAC_Init_ex|.
```

nebeid (approved):

samuel40791765 (approved):

PR #610, opened 2022-09-01
Add PQCrystals Dilithium3 to AWS-LC

jakemas:
### Description of changes: 
This PR introduces a new signature algorithm Crystals Dilithium that has been selected for standardization by NIST as part of the PQC project. We import the Dilithium source code from this commit of the official Dilithium repository https://github.com/pq-crystals/dilithium/commit/3e9b9f1412f6c7435dbeb4e10692ea58f181ee51.

Dilithium has been added into the EVP_PKEY APIs utilizing key generation, sign and verify. This introduction follows a similar pattern as the import of Crystals Kyber in https://github.com/awslabs/aws-lc/pull/343.

### Call-outs:

**Addition to EVP interface**
In addition to simply importing the Dilithium source code and providing an interface to the sign/verify methods, this PR also makes Dilithium3 keys available to the EVP PKEY interface including a subset of methods for key management. As the Dilithium signing procedure is a single-shot signing algorithm that does not use a pre-hash, we follow much of the same structure as the EVP interface for Ed25519. For example, to sign and verify we use the EVP operations `EVP_DigestSign` and `EVP_DigestVerify` with hash `|type|` set to NULL.

**Dilithium3-Ref**
The source code imported is for the Reference implementation of Dilithium3. This was selected for initial import due to it being platform agnostic. The optimized implementation is for AVX2 instruction sets and so priority was given to the implementation which will work on all architectures. Importing optimized versions (and higher security levels) can be done as a follow-on effort.

**SHA-3/SHAKE**
Dilithium requires SHA-3 and SHAKE methods which are in the process of being added to AWS-LC. For this PR, the needed SHA-3/SHAKE dependencies are included within the Dilithium3 source tree directly. We are currently working on separating out a separate SHA-3/SHAKE interface in AWS-LC and utilizing this across both Kyber and Dilithium.

**DRBG**

### Testing:

```
... other existing tests ...

[----------] 5 tests from Dilithium3Test
[       OK ] Dilithium3Test.KeyGeneration (0 ms)
[       OK ] Dilithium3Test.KeySize (0 ms)
[       OK ] Dilithium3Test.KAT (114 ms)
[----------] 5 tests from Dilithium3Test (115 ms total)
... other existing tests ...

[==========] 1224 tests from 96 test suites ran. (21495 ms total)
[  PASSED  ] 1224 tests.
```


torben-hansen:
Can we please split this PR into several logically separated PRs?

jakemas:
> Can we please split this PR into several logically separated PRs?
> 
> I.e.
> 
> 1. import dilithium
> 2. higher level API integration

Okay, splitting up into 2 PRs. Please leave this PR open while I spilt so I can refer back to it in the meantime.

PR #614, opened 2022-09-07
Import Reference PQCrystals Dilithium3 Implementation

jakemas:
### Description of changes: 
This PR introduces a new signature algorithm CRYSTALS Dilithium that has been selected for standardization by NIST as part of the PQC project. We import the Dilithium source code from this commit of the official Dilithium repository https://github.com/pq-crystals/dilithium/commit/3e9b9f1412f6c7435dbeb4e10692ea58f181ee51 as of May 2022.

This PR includes only the Dilithium source implementation and does not include higher level API integration. The interface to the EVP APIs will be included within a following PR.

### Call-outs:

**Dilithium3-Ref**
The source code imported is for the Reference implementation of Dilithium3. This was selected for initial import due to it being platform agnostic. The optimized implementation is for AVX2 instruction sets and so priority was given to the implementation which will work on all architectures. Importing optimized versions (and higher security levels) can be done as a follow-on effort.

**SHA-3/SHAKE**
Dilithium requires SHA-3 and SHAKE methods which are in the process of being added to AWS-LC. For this PR, the needed SHA-3/SHAKE dependencies are included within the Dilithium3 source tree directly. We are currently working on separating out a separate SHA-3/SHAKE interface in AWS-LC and utilizing this across both Kyber and Dilithium.

### Testing:
The unit tests and known answer tests (KATs) will be added once the higher level API integration has been added. This will come in a following PR.


dkostic:
@jakemas could you please check and fix the CI failures?

jakemas:
> @jakemas could you please check and fix the CI failures?
Yes, I am currently working through them

crypto/dilithium/pqcrystals-dilithium_dilithium3_ref/sign.c
```diff
@@ -0,0 +1,334 @@
...
+  else {
+    /* All good, copy msg, return 0 */
+    for(i = 0; i < *mlen; ++i)
+      m[i] = sm[CRYPTO_BYTES + i];
+    return 0;
+  }
+
+badsig:
+  /* Signature verification failed */
+  *mlen = 0; /* Deviation from CRYSTALS (this is set as -1 i.e. not a size_t) */
```
andrewhop, line 329: Should we open an issue with crystals? Does setting this to zero instead of negative one break any of their assumptions or ours?

crypto/dilithium/pqcrystals-dilithium_dilithium3_ref/sign.c
```diff
@@ -0,0 +1,334 @@
...
+  for(i = 0; i < smlen; ++i)
+    m[i] = 0;
```
andrewhop, line 331: Are we planning any deeper cleanup/integration of this code? In this case we could use OPENSSL_cleanse which might be faster and attempts to ensure the compiler doesn't optimize out unused writes.

crypto/dilithium/pqcrystals-dilithium_dilithium3_ref/symmetric-aes.c
```diff
@@ -0,0 +1,16 @@
+#include <stdint.h>
+#include "symmetric.h"
+#include "aes256ctr.h"
+
+
+#ifdef DILITHIUM_USE_AES
```
andrewhop, line 6: Would it make more sense to just not include this file? It seems to be if/def'd out completely and this isn't in the upstream code.

crypto/dilithium/pqcrystals-dilithium_dilithium3_ref/symmetric-shake.c
```diff
@@ -0,0 +1,30 @@
+#include <stdint.h>
+#include "params.h"
+#include "symmetric.h"
+#include "fips202.h"
+
+#ifndef DILITHIUM_USE_AES
```
andrewhop, line 6: Why this change from upstream?

andrewhop (commented):
I reviewed:
* the changes to AWS-LC's CMakeList.txt and common_posix_setup.sh
* the differences between this code and the upstream repository

I did not review the Dilithium code for correctness or safety, only that it matches the repository at [3e9b9f1](https://github.com/pq-crystals/dilithium/commit/3e9b9f1412f6c7435dbeb4e10692ea58f181ee51)

crypto/dilithium/pqcrystals-dilithium_dilithium3_ref/symmetric-aes.c
jakemas: Removed it in 65221e3. I had included the AES modes so that the flag `DILITHIUM_USE_AES` could be turned on if desired, as I wanted to test the AES KATs also pass. But I can remove this for now and we can decide if the AES modes are something we want to offer later.

crypto/dilithium/pqcrystals-dilithium_dilithium3_ref/symmetric-shake.c
jakemas: Removed it in https://github.com/awslabs/aws-lc/commit/65221e3f3983d4e00e635bb4f61fe471e00cf611. If the flag `DILITHIUM_USE_AES` was turned on this file would fail to compile, as it uses SHAKE instead of AES. I added this `#ifndef` to fix this issue. However, if we are removing the support for AES modes then the issue will not arise.

crypto/dilithium/pqcrystals-dilithium_dilithium3_ref/sign.c
jakemas: I've opened an issue upstream with CRYSTALS https://github.com/pq-crystals/dilithium/issues/65. 

> Does setting this to zero instead of negative one break any of their assumptions or ours?

The returned value of `mlen` does not appear to be checked anywhere at the return of this function, so to my understanding there is no further impact. The function documentation states that failure is indicated by the function itself returning value of `-1`, not individual arguments. We can see if the CRYSTALS team responds and can pull from upstream should they modify the code.

crypto/dilithium/pqcrystals-dilithium_dilithium3_ref/sign.c
jakemas: This may fall under a deeper strategy decision as to if we want to maintain the code ourself and make cleanup/optimizations of our own, or if we want to stay aligned with the original source to the best of our ability. I'm reluctant to make any modifications to the source at this stage.

dkostic: I think the decision was to not modify the code for now, unless necessary to make it work with our build. So I agree with Jake to leave everything as is.

andrewhop (dismissed):

dkostic:
Could you please add a Readme file in dilithium folder specifying where the code was taken from, when, and at which commit, and list of all the changes you had to make to the code? (btw, I have to do the same for Kyber, it's on my todo list).

jakemas:
> added README to document the source of Dilithium and the version

Great idea, thank you. I've added a [README](https://github.com/awslabs/aws-lc/pull/614/commits/f2a8ba7b7e0e7ac2137ac0cc0d7dc30c45e0058c). Let me know if there's any further detail you would like to see in there, or if you have suggestions for the formatting.

andrewhop (approved):

dkostic (approved):

dkostic (approved):

andrewhop (approved):

PR #628, opened 2022-09-21
Merge integrate-pq branch into main

dkostic:
### Issues:
CryptoAlg-1260

### Description of changes: 
Merge integrate-pq branch into main.

### Call-outs:

### Testing:


CMakeLists.txt
```diff
@@ -149,6 +149,11 @@ if(BORINGSSL_PREFIX AND BORINGSSL_PREFIX_SYMBOLS AND GO_EXECUTABLE)
            symbol_prefix_include/boringssl_prefix_symbols_nasm.inc
     COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include
     COMMAND ${GO_EXECUTABLE} run ${CMAKE_CURRENT_SOURCE_DIR}/util/make_prefix_headers.go -out ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include ${BORINGSSL_PREFIX_SYMBOLS_PATH}
+    # Temporary exclude all Kyber symbols.
+    # TODO(dkostic): fix the prefix build to work with Kyber source code.
```
torben-hansen, line 153: I think we should fix this before including it into main.

dkostic: I would leave that for later because:
1. we already do something similar for s2n-bignum files in the main branch, see [here](https://github.com/awslabs/aws-lc/blob/main/tests/ci/common_posix_setup.sh#L86) and [here](https://github.com/awslabs/aws-lc/blob/main/tests/ci/common_posix_setup.sh#L93). Not sure which work-around is better, the one I did for Kyber or the one for s2n-bignum. The reason Justin excluded the s2n-bignum functions from prefixing is that we already do some kind of "prefixing" for those function in p384.c and p521.c (same for Kyber, there is already prefixing happening in the code), so adding one prefix on top of the other doesn't work. Not sure why, but I don't see an easy and non-invasive way to fix this. 
2. All the Kyber symbols are internal so they don't have to be prefixed, according to Justin.

torben-hansen (approved):

torben-hansen (dismissed):

torben-hansen (approved):

andrewhop (approved):

PR #633, opened 2022-09-29
Remove the 32 consecutive KeyUpdate limit for TLS 1.3

andrewhop:
### Issues:
Resolves V728984594 and D52012438

### Description of changes: 
AWS-LC currently only allows 32 consecutive key updates before closing the connection. This is a reasonable behavior because a client and server should be sending application data and not just doing key updates for fun. However, OpenSSL and s2n-tls do not restrict consecutive key updates and this breaks some clients when migrating to AWS-LC. 

### Call-outs:
Reviewed internally in V728984594. I decided not to remove `SSL_R_TOO_MANY_KEY_UPDATES` from ssl.h because some customers might be using the define and we may continue to tweak this restriction in the future.

### Testing:
Updated the lots of key updates tests to expect it to pass and not fail.

### DIFF
--- a/ssl/test/runner/runner.go
+++ b/ssl/test/runner/runner.go
@@ -3045,14 +3045,13 @@ read alert 1 0
 			expectedLocalError: "remote error: illegal parameter",
 		},
 		{
-			name: "TooManyKeyUpdates",
+			name: "LotsOfKeyUpdates",
 			config: Config{
 				MaxVersion: VersionTLS13,
 			},
-			sendKeyUpdates:   33,
+			sendKeyUpdates:   100,
 			keyUpdateRequest: keyUpdateNotRequested,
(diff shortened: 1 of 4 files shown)


samuel40791765 (approved):

ssl/test/runner/runner.go
```diff
@@ -3045,14 +3045,13 @@ read alert 1 0
 			expectedLocalError: "remote error: illegal parameter",
 		},
 		{
-			name: "TooManyKeyUpdates",
+			name: "LotsOfKeyUpdates",
```
samuel40791765, line 3048: Should we leave a comment above here, describing that this isn't expected behavior when using AWS-LC, but we still allow it anyways and are testing it?

torben-hansen (approved):

PR #636, opened 2022-10-04
Approve SHA-512/256 for HMAC and Digest signing/verifying

samuel40791765:
### Issues:
Resolves `CryptoAlg-1212`

### Description of changes: 
Recently we've pulled in service indicator improvements from BoringSSL. Within these changes, a noticeable gap was BoringSSL has SHA-512/256 approved for digest signing/verifying and HMAC while we don't. This updates the service indicator to update that.

### Call-outs:
`EVP_sha512_256` is not supported for `PKCS#1v1.5` and will throw an unknown algorithm type error.

### Testing:
`crypto/fipsmodule/service_indicator/service_indicator_test.cc`


crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -1727,33 +1727,39 @@ struct RSATestVector kRSATestVectors[] = {
     { 2048, &EVP_sha256, false, AWSLC_APPROVED, AWSLC_APPROVED },
     { 2048, &EVP_sha384, false, AWSLC_APPROVED, AWSLC_APPROVED },
     { 2048, &EVP_sha512, false, AWSLC_APPROVED, AWSLC_APPROVED },
+    // SHA-512/256 is not supported for PKCS#1v1.5.
```
andrewhop, line 1730: What is the relation between this comment and the test? It seems to be saying this isn't approved, but the test is appears to be saying it is approved.

nebeid: May be leaving a blank line under the comment would make it clearer that it belongs to the test vectors above it not the ones below it.
For my understanding, where in the code do we see that PKCS#1 v1.5 doesn't support this MD? Is it the case in upstream as well?

samuel40791765: Sure, I can do that.
The RSA function throws an `UNKNOWN_ALGORITHM_TYPE` error in the function below. Our upstream doesn't support this either, in fact, they were the one that had pointed out this gap in our service indicator conversions..
https://github.com/awslabs/aws-lc/blob/c950ac8cb890a54458e526f6ccaa2b9ef9c50ffb/crypto/fipsmodule/rsa/rsa.c#L426-L476

Taffer (approved):
IMHO we should (eventually) remove legacy support (so SHA512/256, SHA1, MD5, etc. etc. etc.) but I have no idea how much legacy API use is happening in AWS.

PR #637, opened 2022-10-04
Add Dilithium3 higher level API integration

jakemas:
### Description of changes: 
This PR introduces the EVP_PKEY APIs for Dilithium3, utilizing key generation, sign and verify. This introduction follows a similar pattern as the import of CRYSTALS Kyber in https://github.com/awslabs/aws-lc/pull/343.

### Call-outs:

**Addition to EVP interface**
This PR also makes Dilithium3 keys available to the EVP PKEY interface including a subset of methods for key management. As the Dilithium signing procedure is a single-shot signing algorithm that does not use a pre-hash, we follow much of the same structure as the EVP interface for Ed25519. For example, to sign and verify we use the EVP operations `EVP_DigestSign` and `EVP_DigestVerify` with hash `|type|` set to NULL.

**DRBG**

### Testing:

```
... other existing tests ...

[----------] 6 tests from Dilithium3Test
[       OK ] Dilithium3Test.KeyGeneration (0 ms)
[       OK ] Dilithium3Test.KeySize (0 ms)
[ RUN      ] Dilithium3Test.SIGOperations
[       OK ] Dilithium3Test.SIGOperations (7 ms)
[       OK ] Dilithium3Test.KAT (114 ms)
[----------] 5 tests from Dilithium3Test (115 ms total)
... other existing tests ...

[==========] 1225 tests from 96 test suites ran. (21495 ms total)
[  PASSED  ] 1225 tests.
```


jakemas:
Working through CI issues, once these are sorted we can start the review process.

crypto/dilithium/p_dilithium3_asn1.c
```diff
@@ -0,0 +1,206 @@
...
+
+  return 1;
+}
+
+static int dilithium3_size(const EVP_PKEY *pkey) {
+  return DILITHIUM3_PUBLIC_KEY_BYTES + DILITHIUM3_PRIVATE_KEY_BYTES;
+}
+
+static int dilithium3_bits(const EVP_PKEY *pkey) {
+  return 8 * (DILITHIUM3_PUBLIC_KEY_BYTES + DILITHIUM3_PRIVATE_KEY_BYTES);
```
andrewhop, line 179: What is your rationale for this? The comment in evp.h for EVP_PKEY_bits is hard to map to dilithium3_bits.

crypto/dilithium/p_dilithium3_asn1.c
```diff
@@ -0,0 +1,206 @@
...
+  const DILITHIUM3_KEY *key = pkey->pkey.ptr;
+
+  // See https://datatracker.ietf.org/doc/draft-massimo-lamps-pq-sig-certificates/00/ section 4.
+  CBB spki, algorithm, oid, key_bitstring;
+  if (!CBB_add_asn1(out, &spki, CBS_ASN1_SEQUENCE) ||
+      !CBB_add_asn1(&spki, &algorithm, CBS_ASN1_SEQUENCE) ||
+      !CBB_add_asn1(&algorithm, &oid, CBS_ASN1_OBJECT) ||
+      !CBB_add_bytes(&oid, dilithium3_asn1_meth.oid, dilithium3_asn1_meth.oid_len) ||
+      !CBB_add_asn1(&spki, &key_bitstring, CBS_ASN1_BITSTRING) ||
+      !CBB_add_u8(&key_bitstring, 0 /* padding */) ||
```
andrewhop, line 123: What is this padding for?

crypto/dilithium/p_dilithium3_asn1.c
```diff
@@ -0,0 +1,206 @@
...
+    return 0;
+  }
+
+  return dilithium3_set_pub_raw(out, CBS_data(key), CBS_len(key));
+}
+
+static int dilithium3_pub_encode(CBB *out, const EVP_PKEY *pkey) {
+  const DILITHIUM3_KEY *key = pkey->pkey.ptr;
+
+  // See https://datatracker.ietf.org/doc/draft-massimo-lamps-pq-sig-certificates/00/ section 4.
```
andrewhop, line 116: Does the specification have any KATs to ensure the encoding and decoding logic here lines up with the specification?

crypto/dilithium/p_dilithium_test.cc
```diff
@@ -0,0 +1,152 @@
...
+
+  EXPECT_TRUE(EVP_PKEY_keygen_init(dilithium_pkey_ctx));
+  EXPECT_TRUE(EVP_PKEY_keygen(dilithium_pkey_ctx, &dilithium_pkey));
+  ASSERT_NE(dilithium_pkey->pkey.ptr, nullptr);
+  const DILITHIUM3_KEY *dilithium3Key = (DILITHIUM3_KEY *)(dilithium_pkey->pkey.ptr);
+
+  // New raw public key
+  EVP_PKEY *new_public = EVP_PKEY_new_raw_public_key(EVP_PKEY_DILITHIUM3,
+                                                     NULL,
+                                                     dilithium3Key->pub,
```
andrewhop, line 109: Instead of directly accessing the raw key through the pointer should this test the ASN1 encoding and decoding with EVP_marshal_public_key and EVP_parse_public_key?

crypto/dilithium/p_dilithium3_asn1.c
```diff
@@ -0,0 +1,206 @@
...
+  return DILITHIUM3_PUBLIC_KEY_BYTES + DILITHIUM3_PRIVATE_KEY_BYTES;
+}
+
+static int dilithium3_bits(const EVP_PKEY *pkey) {
+  return 8 * (DILITHIUM3_PUBLIC_KEY_BYTES + DILITHIUM3_PRIVATE_KEY_BYTES);
+}
+
+const EVP_PKEY_ASN1_METHOD dilithium3_asn1_meth = {
+    EVP_PKEY_DILITHIUM3,
+    //1.3.6.1.4.1.2.267.7 = Dilithium SIG Round-3
```
andrewhop, line 184: Where does this come from?

crypto/evp_extra/pq_sig_kat_tests_dilithium3.cc
```diff
@@ -0,0 +1,85 @@
...
+  //the KAT files generated by the dilithium team use the optional APIs that create a
+  //signature for a message m and append the message to the end of the signature.
+  //We only want to bring the APIs that create and verify just the signature,
+  //therefore each signature is a constant DILITHIUM3_SIGNATURE_BYTES and we truncate
+  //the signed message down to a constant DILITHIUM3_SIGNATURE_BYTES.
+  uint8_t signature[DILITHIUM3_SIGNATURE_BYTES];
+  size_t signature_len = DILITHIUM3_SIGNATURE_BYTES;
+  sm.resize(signature_len);
+
+  //convert string read from KAT to int (is there a cleaner way?)
```
andrewhop, line 37: You want `stoi()`.

crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,91 @@
...
+    OPENSSL_PUT_ERROR(EVP, ERR_R_MALLOC_FAILURE);
+    return 0;
+  }
+
+  if (!EVP_PKEY_set_type(pkey, EVP_PKEY_DILITHIUM3)) {
+    OPENSSL_free(key);
+    return 0;
+  }
+
+  DILITHIUM3_keypair(key->pub, key->priv);
```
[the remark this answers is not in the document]
jakemas: I'm trying to be consistent with [p_kyber.c](https://github.com/awslabs/aws-lc/blob/main/crypto/kyber/p_kyber.c#L22) [p_ed25519.c](https://github.com/awslabs/aws-lc/blob/main/crypto/evp_extra/p_ed25519.c#L41) and [p_x25519.c](https://github.com/awslabs/aws-lc/blob/main/crypto/evp_extra/p_x25519.c#L40) which all do it this way...

crypto/dilithium/p_dilithium3_asn1.c
```diff
@@ -0,0 +1,206 @@
...
+      !CBB_flush(out)) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_ENCODE_ERROR);
+    return 0;
+  }
+
+  return 1;
+}
+
+static int dilithium3_size(const EVP_PKEY *pkey) {
+  return DILITHIUM3_PUBLIC_KEY_BYTES + DILITHIUM3_PRIVATE_KEY_BYTES;
```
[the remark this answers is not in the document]
jakemas: Fixed, thanks!

crypto/dilithium/p_dilithium3_asn1.c
jakemas: Agreed, I got it wrong here. Looking at [OQS](https://github.com/open-quantum-safe/openssl/blob/OQS-OpenSSL_1_1_1-stable/crypto/ec/oqs_meth.c#L1082) they use the length of the public key - so i've changed to be consistent with this.

crypto/dilithium/p_dilithium3_asn1.c
jakemas: These are temp values as we wait for NIST to release OIDs, but as with [kyber](https://github.com/awslabs/aws-lc/blob/main/crypto/kyber/p_kyber512_asn1.c#L121), I'm using these ones from http://oid-info.com/get/1.3.6.1.4.1.2.267.7

crypto/dilithium/p_dilithium3.c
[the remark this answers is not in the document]
andrewhop: Is that a bug in those spots as well? @dkostic thoughts?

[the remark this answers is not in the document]
dkostic: I don't think it's a bug, those functions always return successfully so I don't see a reason to check the return value.

dkostic:
I see the CI is still failing, I prefer to wait until the CI is passing before I review a PR. But if you want I'll start the review even now?

crypto/dilithium/p_dilithium3.c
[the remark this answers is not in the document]
dkostic: discussed this with Torben a bit -- it's probably a good idea to always check the return values in case the functions change at some point and start returning something meaningful. So let's add the checks.

jakemas:
> I see the CI is still failing, I prefer to wait until the CI is passing before I review a PR. But if you want I'll start the review even now?

The CI that are failing are an s2n prefixing issue and the cryptofuzz undefined references that I cannot understand. The only other failures are due to the formal verification that are currently being made.

crypto/dilithium/p_dilithium_test.cc
```diff
@@ -0,0 +1,152 @@
...
+  EVP_PKEY *dilithium_pkey2 = EVP_PKEY_new();
+  ASSERT_NE(dilithium_pkey2, nullptr);
+
+  EXPECT_TRUE(EVP_PKEY_keygen_init(dilithium_pkey_ctx2));
+  EXPECT_TRUE(EVP_PKEY_keygen(dilithium_pkey_ctx2, &dilithium_pkey2));
+  ASSERT_NE(dilithium_pkey2->pkey.ptr, nullptr);
+
+  EXPECT_EQ(0, EVP_PKEY_cmp(dilithium_pkey1, dilithium_pkey2));
+  EXPECT_EQ(1, EVP_PKEY_cmp(dilithium_pkey1, dilithium_pkey1));
+  EXPECT_EQ(1, EVP_PKEY_cmp(dilithium_pkey2, dilithium_pkey2));
```
dkostic, line 87: not sure what's the value of these two tests :)

crypto/dilithium/p_dilithium_test.cc
```diff
@@ -0,0 +1,152 @@
...
+  EVP_PKEY_CTX *dilithium_pkey_ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_DILITHIUM3, nullptr);
+  ASSERT_NE(dilithium_pkey_ctx, nullptr);
+
+  EVP_PKEY *dilithium_pkey = EVP_PKEY_new();
+  ASSERT_NE(dilithium_pkey, nullptr);
+
+  EXPECT_TRUE(EVP_PKEY_keygen_init(dilithium_pkey_ctx));
+  EXPECT_TRUE(EVP_PKEY_keygen(dilithium_pkey_ctx, &dilithium_pkey));
+
+  EXPECT_EQ(DILITHIUM3_PUBLIC_KEY_BYTES + DILITHIUM3_PRIVATE_KEY_BYTES, EVP_PKEY_size(dilithium_pkey));
```
dkostic, line 148: doesn't `EVP_PKEY_size` return the signature size?

crypto/evp_extra/pq_sig_kat_tests_dilithium3.cc
```diff
@@ -0,0 +1,79 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <gtest/gtest.h>
+#include <openssl/evp.h>
+#include "../rand_extra/pq_custom_randombytes.h"
```
dkostic, line 6: nit: move this include after vector

crypto/evp_extra/pq_sig_kat_tests_dilithium3.cc
```diff
@@ -0,0 +1,79 @@
...
+  ASSERT_TRUE(t->GetAttribute(&count, "count"));
+  ASSERT_TRUE(t->GetBytes(&seed, "seed"));
+  ASSERT_TRUE(t->GetAttribute(&mlen, "mlen"));
+  ASSERT_TRUE(t->GetBytes(&msg, "msg"));
+  ASSERT_TRUE(t->GetBytes(&pk, "pk"));
+  ASSERT_TRUE(t->GetBytes(&sk, "sk"));
+  ASSERT_TRUE(t->GetAttribute(&smlen, "smlen"));
+  ASSERT_TRUE(t->GetBytes(&sm, "sm"));
+
+  //the KAT files generated by the dilithium team use the optional APIs that
```
dkostic, line 28: what does it mean `optional API`?

crypto/evp_extra/pq_sig_kat_tests_dilithium3.cc
```diff
@@ -0,0 +1,79 @@
...
+  //create a signature for a message m and append the message to the end of the
+  //signature. We only want to bring the APIs that create and verify just the
```
dkostic, line 30: > We only want to bring the APIs that create and verify just the signature.

Why is that?

crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,91 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
dkostic, line 1: Regarding this: 
```
Dilithium offers both deterministic and randomized signing.
```

How do we handle this? In Dilithium's source code, the choice between deterministic and randomized signing is made based on a compile time flag which means we support only one of  those two.
- Can we support only one signing method?
- If so, which one? Should we specify and document this?
- If we have to support both signing methods then we need to implement them both.
- Does the signing method implie the verifying method or is the verification always the same regadless of the signing method?

crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,91 @@
...
+// SPDX-License-Identifier: Apache-2.0
```
dkostic, line 2: another general comments: it'd be nice to add a short example how to use dilithium to the readme file (similarly to https://github.com/awslabs/aws-lc/blob/main/crypto/kyber/README.md).

crypto/dilithium/p_dilithium3_asn1.c
```diff
@@ -0,0 +1,207 @@
...
+      !CBB_add_bytes(&private_key, key->priv, DILITHIUM3_PRIVATE_KEY_BYTES) ||
+      !CBB_flush(out)) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_ENCODE_ERROR);
+    return 0;
+  }
+
+  return 1;
+}
+
+static int dilithium3_size(const EVP_PKEY *pkey) {
```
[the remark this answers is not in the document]
jakemas: Thanks, fixed in https://github.com/jakemas/aws-lc/commit/945faef337b75019d1afbb782864312397ce8ca6

crypto/dilithium/p_dilithium3_asn1.c
```diff
@@ -0,0 +1,207 @@
...
+  return DILITHIUM3_SIGNATURE_BYTES;
+}
+
+static int dilithium3_bits(const EVP_PKEY *pkey) {
```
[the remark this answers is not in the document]
jakemas: Thanks, fixed in https://github.com/jakemas/aws-lc/commit/945faef337b75019d1afbb782864312397ce8ca6

crypto/dilithium/p_dilithium_test.cc
jakemas: Thanks, fixed in https://github.com/jakemas/aws-lc/commit/945faef337b75019d1afbb782864312397ce8ca6

crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,91 @@
...
+    return 1;
+  }
+
+  if (*siglen < DILITHIUM3_SIGNATURE_BYTES) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_BUFFER_TOO_SMALL);
+    return 0;
+  }
+
+  if (DILITHIUM3_sign(sig, siglen, tbs, tbslen, key->priv)!= 0) {
+    return 0;
```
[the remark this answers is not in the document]
jakemas: I'm following the president set by [p_ed25519.c ](https://github.com/awslabs/aws-lc/blob/main/crypto/evp_extra/p_ed25519.c#L68-L70), but yes, if an error message would be helpful we can perhaps add:
```suggestion
    OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_SIGNATURE);
    return 0;
```
if you agree that `EVP_R_INVALID_SIGNATURE` seems the best fit from `evp_errors.h`.

crypto/dilithium/p_dilithium_test.cc
jakemas: Agree, [was just following the same tests in kyber](https://github.com/jakemas/aws-lc/blob/main/crypto/kyber/p_kyber_test.cc#L83-L84).

crypto/evp_extra/pq_sig_kat_tests_dilithium3.cc
jakemas: Thanks, fixed in https://github.com/awslabs/aws-lc/pull/637/commits/ad38c62541c964623f04455185f6b40639e80eeb.

crypto/dilithium/p_dilithium_test.cc
```diff
@@ -0,0 +1,152 @@
...
+  EXPECT_TRUE(EVP_PKEY_keygen_init(dilithium_pkey_ctx));
+  EXPECT_TRUE(EVP_PKEY_keygen(dilithium_pkey_ctx, &dilithium_pkey));
+  ASSERT_NE(dilithium_pkey->pkey.ptr, nullptr);
+
+  const DILITHIUM3_KEY *dilithium3Key = (DILITHIUM3_KEY *)(dilithium_pkey->pkey.ptr);
+  EXPECT_TRUE(dilithium3Key->has_private);
+
+  uint8_t *buf = nullptr;
+  size_t buf_size;
+  EXPECT_TRUE(EVP_PKEY_get_raw_public_key(dilithium_pkey, buf, &buf_size));
```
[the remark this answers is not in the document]
jakemas: Done in https://github.com/awslabs/aws-lc/pull/637/commits/4a421b9c151e0ee6ba2b6c626be32d609ab761e2.

crypto/dilithium/p_dilithium3.c
jakemas: Thanks, I've followed the structure of your README and updated in https://github.com/awslabs/aws-lc/pull/637/commits/5d44b6ed8f3a495d1bca9daa293bace87e1c8871.

crypto/evp_extra/pq_sig_kat_tests_dilithium3.cc
jakemas: fixed in https://github.com/awslabs/aws-lc/pull/637/commits/d4aaf17d57b9852b74b3eee841e87e4ad2cbd339

crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,91 @@
...
+
+  OPENSSL_free(pkey->pkey.ptr);
+  pkey->pkey.ptr = key;
+  return 1;
+}
+
+static int pkey_dilithium3_sign_message(EVP_PKEY_CTX *ctx, uint8_t *sig,
+                                        size_t *siglen, const uint8_t *tbs,
+                                        size_t tbslen) {
+  DILITHIUM3_KEY *key = ctx->pkey->pkey.ptr;
```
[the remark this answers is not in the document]
jakemas: added in https://github.com/awslabs/aws-lc/pull/637/commits/9923bfc9b80af49593139688a17bdd6b1057e155

crypto/dilithium/p_dilithium3.c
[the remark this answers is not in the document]
jakemas: https://github.com/awslabs/aws-lc/pull/637/commits/9923bfc9b80af49593139688a17bdd6b1057e155

crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,91 @@
...
+
+  if (*siglen < DILITHIUM3_SIGNATURE_BYTES) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_BUFFER_TOO_SMALL);
+    return 0;
+  }
+
+  if (DILITHIUM3_sign(sig, siglen, tbs, tbslen, key->priv)!= 0) {
+    return 0;
+  }
+
```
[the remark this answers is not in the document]
jakemas: added in https://github.com/awslabs/aws-lc/pull/637/commits/9923bfc9b80af49593139688a17bdd6b1057e155

crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,91 @@
...
+  *siglen = DILITHIUM3_SIGNATURE_BYTES;
+  return 1;
+}
+
+static int pkey_dilithium3_verify_message(EVP_PKEY_CTX *ctx, const uint8_t *sig,
+                                          size_t siglen, const uint8_t *tbs,
+                                          size_t tbslen) {
+  DILITHIUM3_KEY *key = ctx->pkey->pkey.ptr;
```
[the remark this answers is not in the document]
jakemas: added in https://github.com/awslabs/aws-lc/pull/637/commits/9923bfc9b80af49593139688a17bdd6b1057e155

jakemas:
cherry picked https://github.com/awslabs/aws-lc/pull/637/commits/60ea1509bd7e5123b462b9259d85b8d09642ab8d to apply the fix to prefixing - as prefixing issues are preventing the CI from successfully passing.

PR #638, opened 2022-10-05
Minor service indicator test improvements

samuel40791765:
### Issues:
Resolves `CryptoAlg-1273`

### Description of changes: 
Minor service indicator test improvements called out in https://github.com/awslabs/aws-lc/pull/592

### Call-outs:
N/A

### Testing:
`crypto/fipsmodule/service_indicator/service_indicator_test.cc`


crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -275,6 +275,16 @@ static const uint8_t kAESXTSKey_256[64] = {
     0x2c, 0xf4, 0x1b, 0x08
 };
 
+// From crypto/fipsmodule/modes/xts_test.cc
```
andrewhop, line 278: Is it possible to avoid this duplication?

samuel40791765: I'm still syncing up with @Taffer if we should remove this test. We think it might be a bit redundant. The `Init` fails, which makes it `NOT_APPROVED`, but the init function was never supposed to touch the service indicator in the first place.

andrewhop: I meant the duplicate test data, can it be shared? I like the test in general

samuel40791765: I don't think the data can be shared unless we do a major test refactor in either file. `crypto/fipsmodule/modes/xts_test.cc` consumes the data via a struct and `char` strings throughout the entire test suite, while all of the service indicator tests through `uint8_t` arrays. 
Chris and I have discussed that this specific test isn't adding any additional value to the service indicator tests though, its just hitting the already tested behavior in [`crypto/fipsmodule/modes/xts_test.cc`](https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/modes/xts_test.cc#L562). I'll be adding another test that should be more robust.

Taffer (dismissed):
This is better, thanks!

andrewhop (approved):

Taffer (approved):

PR #641, opened 2022-10-06
Upstream merge 10 05

DominicDams:
### Issues:
Regular Upstream merge

### Description of changes: 
Pull in commits from BSSL

### Call-outs:
See internal docs

### Testing:
`ninja-build run_tests`


andrewhop (dismissed):

CMakeLists.txt
```diff
@@ -233,6 +233,8 @@ if (GCC)
   # https://cmake.org/cmake/help/latest/command/try_run.html
   set(MEMCMP_INVALID_STRIPPED "")
   set(MEMCMP_INVALID_STRIPPED__TRYRUN_OUTPUT "")
+  # See https://gitlab.kitware.com/cmake/cmake/-/issues/16920 for why we set this to 2
+  set(THREADS_PTHREAD_ARG "2" CACHE STRING "Forcibly set by CMakeLists.txt." FORCE)
```
dkostic, line 237: which command in our cmake is creating the issue that is solved by this line?

DominicDams: The issue is the try run command here https://github.com/awslabs/aws-lc/blob/main/CMakeLists.txt#L207 . If we try to find-threads then it will run a test program for the threading library and the test program apparently needs the `THREADS_PTHREAD_ARG` set to 2 otherwise it will fail.

dkostic: ok, can you please add that to the comment and open a ticket for fixing this?

DominicDams: I don't really understand what we would fix here, the try_run command is part of cmake, and this appears to be undocumented (as far as I know) behavior for try_run. Our options are to either set the argument here in order to make sure this works for cross-compiled code, or stop using try_run entirely. The info I put there was from the issue I linked above and the comments on it.

dkostic: yeah, I meant we can maybe stop using try_run and find another solution for that.

CMakeLists.txt
```diff
@@ -233,6 +233,15 @@ if (GCC)
   # https://cmake.org/cmake/help/latest/command/try_run.html
   set(MEMCMP_INVALID_STRIPPED "")
   set(MEMCMP_INVALID_STRIPPED__TRYRUN_OUTPUT "")
+
+  # See https://gitlab.kitware.com/cmake/cmake/-/issues/16920 for why we set this to 2
+  # During the `try_run` commands invoked by `check_run` a test program is run for the threading library.
+  # This test is run expecting two threads, and if this variable isn't set, for some systems
```
dkostic, line 239: I think something is missing from this sentence, no?
```
# This test is run expecting two threads, and if this variable isn't set, for some systems
# (particularly for systems using cross-compilation).
```

dkostic (approved):

andrewhop (approved):

PR #655, opened 2022-10-19
Support Rust Address Sanitizer; Prefix KYBER symbols

justsmth:
### Issues:
N/A

### Description of changes: 
* AWS-LC supports an address sanitizer (ASAN) build.  This change allows ASAN to be used when building the `aws-lc-sys` crate.
* The prefix build on Mac was broken.  (On Mac, "sed" requires an extension when performing in-place file modifications.)
* Update headers for Kyber to support prefixing.
* Update to the latest Rust-bindgen version.

### Call-outs:
N/A

### Testing:
I manually ran the `aws-lc-sys` crate generation scripts since these are not yet integrated with our CI.


dkostic:
in `kyber` folder we have a readme file where we keep track of all changes that we make to the Kyber source code. Could you please update it with the changes in this PR?

bindings/rust/generate/_run_supported_symbol_builds.sh
```diff
@@ -27,7 +27,7 @@ docker run -v "$(pwd)":"$(pwd)" -w "$(pwd)" --rm --platform linux/amd64 rust:lin
 ## arm64 build
 docker run -v "$(pwd)":"$(pwd)" -w "$(pwd)" --rm --platform linux/arm64 rust:linux-arm64 /bin/bash "${SCRIPT_DIR}"/_collect_symbols_build.sh
 
-sort "${SYMBOLS_COLLECT_FILE}" | uniq > "${SYMBOLS_FILE}"
+sort "${SYMBOLS_COLLECT_FILE}" | uniq | grep -v "^_\?bignum_" | grep -v "pqcrystals" > "${SYMBOLS_FILE}"
```
dkostic, line 30: so why do we need to exclude bignum and pqcrystal symbols when in this PR we fixed their prefixing?

justsmth: During the Rust build, "rust-bindgen" parses our header files directly -- including the parsing of an *unfiltered* `boringssl_prefix_symbols.h` (b/c it's not going through CMAKE). We need to ensure that the contents of `boringssl_prefix_symbols.h` generated here is consistent with what is used during the actual CMAKE build.

dkostic: ok, I think I don't understand how the whole prefixing for Rust works. But anyway, I thought this PR makes the bignum and Kyber symbols the same as all other symbols in aws-lc. Is that not the case?

CMakeLists.txt
```diff
@@ -151,9 +151,10 @@ if(BORINGSSL_PREFIX AND BORINGSSL_PREFIX_SYMBOLS AND GO_EXECUTABLE)
     COMMAND ${GO_EXECUTABLE} run ${CMAKE_CURRENT_SOURCE_DIR}/util/make_prefix_headers.go -out ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include ${BORINGSSL_PREFIX_SYMBOLS_PATH}
     # Temporary exclude all Kyber symbols.
     # TODO(dkostic): fix the prefix build to work with Kyber source code.
-    COMMAND sed -i '/kyber/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/boringssl_prefix_symbols.h
-    COMMAND sed -i '/kyber/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/boringssl_prefix_symbols_asm.h
-    COMMAND sed -i '/kyber/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/boringssl_prefix_symbols_nasm.inc
+    COMMAND sed -i.bak '/pqcrystals/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/boringssl_prefix_symbols.h
```
dkostic, line 154: I missed that this line was added. I'm totally confused now. The PR description says:
```
- Update headers for Kyber to support prefixing.
```
So if we support prefixing Kyber symbols now, why do we have to exclude them with this `sed`?

bindings/rust/generate/_run_supported_symbol_builds.sh
justsmth: Ok. The "bignum" symbols are removed from `boringssl_prefix_symbols.h` by CMake because they are not prefixed. The "pqcrystals" symbols are removed because we are handling the prefixing through a different mechanism. We are effectively doing the same thing for the same reasons, but we're doing it at this point because it's easier.

CMakeLists.txt
justsmth: After a prefix build, all "kyber" symbols are prefixed -- however, "bignum" symbols are not prefixed.  So they both are being removed from `boringssl_prefix_symbols.h` but for different reasons. 

I didn't remove the "pqcrystals" from the assembly headers b/c I assumed prefix-adding macros would still be needed there if the functions were being called from assembly code.

DominicDams (approved):

dkostic (approved):

PR #656, opened 2022-10-19
Add SSH KDF to FIPS boundary

Taffer:
### Issues:
Resolves CryptoAlg-1226

### Description of changes: 
SSH KDF is a special-purpose KDF that's FIPS Approved for use only in ssh. OpenSSL 3 supports it.

### Call-outs:
Lots more unit testing than OpenSSL's version.

### Testing:
Unit tests, service indicator, built FIPS and non-FIPS.


include/openssl/kdf.h
```diff
@@ -83,6 +83,29 @@ OPENSSL_EXPORT int EVP_PKEY_CTX_add1_hkdf_info(EVP_PKEY_CTX *ctx,
                                                const uint8_t *info,
                                                size_t info_len);
 
+// SSH-specific KDF
+//
+// This KDF should only be called from SSH client/server code; it's not a
```
amirhosv, line 88: Does exposing this KDF function via `EVP_KDF_XYZ` APIs violate FIPS?

include/openssl/kdf.h
```diff
@@ -83,6 +83,29 @@ OPENSSL_EXPORT int EVP_PKEY_CTX_add1_hkdf_info(EVP_PKEY_CTX *ctx,
...
+#define EVP_KDF_SSHKDF_TYPE_ENCRYPTION_KEY_SRV_TO_CLI 68
+#define EVP_KDF_SSHKDF_TYPE_INTEGRITY_KEY_CLI_TO_SRV  69
+#define EVP_KDF_SSHKDF_TYPE_INTEGRITY_KEY_SRV_TO_CLI  70
+
+// SSHKDF calculates |out_len| bytes of the SSH KDF, using |digest|, and
+// writes them to |out|. It returns one on success and zero on error.
+OPENSSL_EXPORT int SSHKDF(const EVP_MD *evp_md,
+                          const uint8_t *key, size_t key_len,
+                          const uint8_t *xcghash, size_t xcghash_len,
+                          const uint8_t *session_id, size_t session_id_len,
```
amirhosv, line 105: Maybe it's worth introducing a new struct that has `uint8_t*` and `size_t` and then use that to reduce the number of arguments. That would increase readability of the API as well.

crypto/fipsmodule/ssh/kdf.c
```diff
@@ -0,0 +1,105 @@
...
+int SSHKDF(const EVP_MD *evp_md,
+           const uint8_t *key, size_t key_len,
+           const uint8_t *xcghash, size_t xcghash_len,
+           const uint8_t *session_id, size_t session_id_len,
+           char type,
+           uint8_t *out, size_t out_len)
+{
+  EVP_MD_CTX *md = NULL;
+  uint8_t digest[EVP_MAX_MD_SIZE];
+  unsigned int digest_size = 0;
```
amirhosv, line 20: `size_t` instead of `unsigned int`?

crypto/fipsmodule/ssh/kdf.c
```diff
@@ -0,0 +1,105 @@
...
+  int ret = 0;
+
+  FIPS_service_indicator_lock_state();
+
+  md = EVP_MD_CTX_new();
+  if (md == NULL) {
+    return 0;
+  }
+
+  if (!EVP_DigestInit_ex(md, evp_md, NULL)) {
```
amirhosv, line 31: Are branches corresponding to `EVP_DigestXYZ` APIs failures have tests?

crypto/fipsmodule/ssh/kdf.c
```diff
@@ -0,0 +1,105 @@
...
+#include <openssl/mem.h>
+
+#include "../service_indicator/internal.h"
+#include "../../internal.h"
+
+int SSHKDF(const EVP_MD *evp_md,
+           const uint8_t *key, size_t key_len,
+           const uint8_t *xcghash, size_t xcghash_len,
+           const uint8_t *session_id, size_t session_id_len,
+           char type,
```
amirhosv, line 15: Shouldn't this method do some checks to ensure `type` is one of the `EVP_KDF_SSHKDF_TYPE`?

include/openssl/kdf.h
andrewhop: We already have that, it's a crypto byte string [cbs_st](https://github.com/awslabs/aws-lc/blob/main/include/openssl/bytestring.h#L40), it is primarily used for internal functions/parsing in libssl. The convention for libcrypto's public API is unfortunately lots of pointers and lengths which offers the greatest flexibility for other languages and consumers at the cost of complexity and easy to misuse API.

What API does OpenSSL offer for this KDF? Is that the API OpenSSH expects?

include/openssl/kdf.h
Taffer: FIPS doesn't care about APIs, just proper crypto (and evidence of that). SSH KDF isn't available via the `EVP` APIs, just directly as `SSHKDF()`.

include/openssl/kdf.h
Taffer: This is the pattern used by the rest of the APIs... we generally do what BoringSSL/OpenSSL do, so we're more compatible with existing Boring/Open SSL code, and users of things we've added have a familiar API.

crypto/fipsmodule/ssh/kdf.c
Taffer: Good catch!

crypto/fipsmodule/ssh/kdf.c
Taffer: That's a good idea; I wasn't sure those were the only valid choices, but they do seem to be the only choices available in OpenSSL. I'll add checks and an additional unit test.

crypto/fipsmodule/ssh/kdf.c
Taffer: They should, in the digest unit tests.

include/openssl/kdf.h
Taffer: OpenSSL has an EVP API for this; the `SSHKDF()` function there is `static`. This seemed like the least intrusive way of adding this functionality.

crypto/fipsmodule/ssh/kdf.c
amirhosv: Is there any unit tests for `SSHKDF` that would cause `!EVP_DIgestInit_ex(md, evp_md, NULL)` fail and cover that branch in this file?

include/openssl/kdf.h
amirhosv: @andrewhop Similar to symmetric encryption EVP APIs, OpenSSL offers EVP_KDF_ APIs and as a matter of fact, OpenSSL supports SSHKDF: https://www.openssl.org/docs/man3.0/man7/EVP_KDF-SSHKDF.html. I think it's best to introduce this API via the existing framework.

include/openssl/kdf.h
amirhosv: https://www.openssl.org/docs/man3.0/man7/EVP_KDF-SSHKDF.html

include/openssl/kdf.h
Taffer: The EVP_KDF API is so terrible to use though. I suppose we'll have to though, if our goal is to remain compatible with OpenSSL.

crypto/fipsmodule/ssh/kdf.c
Taffer: The unit tests in AWS-LC seem to cover the API being tested, not the APIs they're calling; the PBKDF and HKDF unit tests (for example) don't test the underlying HMAC calls.

If we want to improve code coverage by triggering all (or any? more?) branches, we'll have a lot of additional tests to add. I'm not saying that would be a bad thing, just that we don't currently have it.

include/openssl/kdf.h
torben-hansen: The public EVP_KDF API and internals is not compatible with AWS-LC for many reasons and we are not interested in pulling in all that code. Especially not for something as esoteric as an SSH KDF.

I think current approach is good.

nebeid: I second Torben. See my response to the above comment: https://github.com/awslabs/aws-lc/pull/656#discussion_r1001040100

include/openssl/kdf.h
Taffer: Surely. I was trying to find a good spot for it without creating a new file; IMHO we should put all the KDFs together in one logical spot, but that moves us away from Boring/Open.

amirhosv: @nebeid Thanks for the pointer. I was under the assumption that AWS-LC's API is suppose to be 100% compatible with OpenSSL so it can be a drop-in replacement for OpenSSL, which is not true. Now that AWS-LC is providing some APIs that are not in OpenSSL (and possibly not in BoringSSL), it's better to have their corresponding header files in a separate directory like `include/aws-lc` rather than `include/openssl`. In this way, consumers of AWS-LC can easily distinguish APIs that are specific to AWS-LC that are not supported in OpenSSL.

nebeid (commented):
Sorry, I'm still working on the review, but I'll post this comment for now.

crypto/fipsmodule/bcm.c
```diff
@@ -136,6 +136,7 @@
 #include "sha/sha256.c"
 #include "sha/sha3.c"
 #include "sha/sha512.c"
+#include "ssh/kdf.c"
```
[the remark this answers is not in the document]
Taffer: Sure. I did it this way originally to mirror the TLS KDF, the other protocol-specific KDF we've got.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -407,6 +407,25 @@ void PBKDF2_verify_service_indicator(const EVP_MD *evp_md, size_t password_len,
   }
 }
 
+void SSHKDF_verify_service_indicator(const EVP_MD *evp_md) {
```
nebeid, line 410: I think that since [the ACVP tests will register a certain cipher](https://pages.nist.gov/ACVP/draft-celi-acvp-kdf-ssh.html#name-property-registration), then there is an expected output of the derived key, `out_len`. So I think the service indicator should check out_len when the request is for keys and not IVs.
Moreover, based on condition 3 at the end of Sec. 5.2 of [SP 800 135 r1](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-135r1.pdf) for SSHKDF:
> The hash function deployed in the KDFs meets the security strength(s)
required by the cryptographic function(s) for which the keying material is
being generated. The security strengths of approved hash functions used in
KDFs can be found in SP 800-57

we may even have to match the length checks with the evp_md checks, e.g. sha256 for length 128. I am not sure, but we can ask the lab.

Looking again, this is showing in the unit tests that the IVs have a certain length independent of the MD and the encryption and integrity keys have different lengths from each other and different for each MD. We may want to dig into this more to see how they're determined, or if they're arbitrary.

crypto/fipsmodule/sshkdf/kdf.c
```diff
@@ -0,0 +1,112 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+// Based on OpenSSL 3.x SSHKDF().
```
samuel40791765, line 3: If the code was mostly taken from OpenSSL, we should put their licenses in the corresponding files. We should be able to use the license header they have here:
https://github.com/openssl/openssl/blob/ad4910fad22d57c6d685b0ca83bdb7b2bf69d8fd/providers/implementations/kdfs/sshkdf.c#L1-L8
It might also help to specify which test cases were taken from OpenSSL as well.

crypto/fipsmodule/sshkdf/sshkdf_test.cc
```diff
@@ -0,0 +1,522 @@
...
+        session_id, session_id_len, type,
+        out, out_len));
+    EXPECT_EQ(Bytes(expected_output, sizeof(out_len)),
+        Bytes(out, sizeof(out_len)));
+}
+
+TEST(SSHKDFTest, SSHKDF_TYPE_INSANITY) {
+    // The sanity check for the |type| param happens before anything else, so
+    // we can check it without any valid info.
+    ASSERT_FALSE(SSHKDF(nullptr, nullptr, 0, nullptr, 0, nullptr, 0,
```
nebeid, line 33: Don't we need to test each parameter being wrong separately?

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -2960,6 +2961,101 @@ TEST_P(PBKDF2_ServiceIndicatorTest, PBKDF2) {
...
+    const size_t xcghash_len;
+    const uint8_t *session_id;
+    const size_t session_id_len;
+    const char type;
+    const size_t output_len;
+    const uint8_t *expected_output;
+    const FIPSStatus expect_approved;
+} kSSHKDFTestVectors[] = {
+    {
+        EVP_sha256,
```
nebeid, line 3014: I understand our trend in the service indicator is to test all possible inputs that are approved and one non-approved (or more, if there are different combinations). So I suggest adding tests for the other approved MDs, with one type of key, not all types as in the unit tests.

crypto/fipsmodule/sshkdf/kdf.c
Taffer: Good idea, I wasn't sure what to do in this case.

include/openssl/sshkdf.h
```diff
@@ -0,0 +1,42 @@
...
+// |type| values for SSHKDF().
+#define EVP_KDF_SSHKDF_TYPE_INITIAL_IV_CLI_TO_SRV     65
+#define EVP_KDF_SSHKDF_TYPE_INITIAL_IV_SRV_TO_CLI     66
+#define EVP_KDF_SSHKDF_TYPE_ENCRYPTION_KEY_CLI_TO_SRV 67
+#define EVP_KDF_SSHKDF_TYPE_ENCRYPTION_KEY_SRV_TO_CLI 68
+#define EVP_KDF_SSHKDF_TYPE_INTEGRITY_KEY_CLI_TO_SRV  69
+#define EVP_KDF_SSHKDF_TYPE_INTEGRITY_KEY_SRV_TO_CLI  70
+
+// SSHKDF calculates |out_len| bytes of the SSH KDF, using |digest|, and
+// writes them to |out|. It returns one on success and zero on error.
```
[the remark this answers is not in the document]
Taffer: I think this is too much detail (the only people who should be using this are just calling this instead of the one built in to OpenSSH), but I've gone with something similar.

crypto/fipsmodule/bcm.c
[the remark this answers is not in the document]
Taffer: It's currently following the other single-purpose KDF in `tls/`... 🤷

I really love to go through this code base and make it consistent, but that'd make merges with Boring painful.

crypto/fipsmodule/sshkdf/kdf.c
```diff
@@ -0,0 +1,112 @@
...
+
+  // Sanity-check.
+  if (type < EVP_KDF_SSHKDF_TYPE_INITIAL_IV_CLI_TO_SRV ||
+      type > EVP_KDF_SSHKDF_TYPE_INTEGRITY_KEY_SRV_TO_CLI) {
+    return 0;
+  }
+
+  FIPS_service_indicator_lock_state();
+
+  md = EVP_MD_CTX_new();
```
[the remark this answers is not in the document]
Taffer: Good idea!

crypto/fipsmodule/sshkdf/sshkdf_test.cc
Taffer: At the time this was written, no, but after adding more sanity checking from https://github.com/openssl/openssl/blob/ad4910fad22d57c6d685b0ca83bdb7b2bf69d8fd/providers/implementations/kdfs/sshkdf.c#L99, yes.

crypto/fipsmodule/sshkdf/kdf.c
```diff
@@ -0,0 +1,112 @@
...
+  if (md == NULL) {
+    return 0;
```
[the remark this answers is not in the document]
Taffer: Changed `return 0` to `goto out` for consistency. If `EVP_MD_CTX_free(NULL)` segfaults, I'll have to adjust again (and then I'll be annoyed at `EVP_MD_CTX_free`)...

crypto/fipsmodule/service_indicator/service_indicator.c
Taffer: New sanity checking makes sure the `out_len` is `> 0`, and I considered making it `>= the hash length`, but I didn't think there was enough context in `SSHKDF()` to know what the purpose of the output was.

The test vectors are all straight from NIST, so I'm assuming they should be considered valid.

crypto/fipsmodule/sshkdf/sshkdf.c
```diff
@@ -23,6 +28,18 @@ int SSHKDF(const EVP_MD *evp_md,
...
+  }
+  if (xcghash == NULL || xcghash_len == 0) {
+    return 0;
+  }
+  if (session_id == NULL || session_id_len == 0) {
+    return 0;
+  }
   if (type < EVP_KDF_SSHKDF_TYPE_INITIAL_IV_CLI_TO_SRV ||
       type > EVP_KDF_SSHKDF_TYPE_INTEGRITY_KEY_SRV_TO_CLI) {
     return 0;
```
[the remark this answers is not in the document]
Taffer: That ends up being harder to read/maintain; the compiler will probably coalesce these into one giant Boolean with one return.

nebeid (approved):

PR #658, opened 2022-10-21
Add HKDF to ACVP Testing Harness

billbo-yang:
### Issues:
Resolves [CryptoAlg-1281](https://sim.amazon.com/issues/CryptoAlg-1281)

### Description of changes: 
Currently AWS-LC has HKDF within the FIPS boundary, but doesn't support testing it with it's ACVP tool. This PR adds support for HKDF to the ACVP tool. Note that this does not add `HKDF_expand()` into the testing harness, as that requires it to be wrapped in an API (see [CryptoAlg-1367](https://sim.amazon.com/issues/CryptoAlg-1367)).

### Call-outs:
No architecture or design changes.

### Testing:
Change tested by running the ACVP tool on HKDF in online mode to fetch the appropriate vectors, compute the result, then upload said results to be compared to the expected results by NIST. Also tested by adding HKDF support for the ACVP tool testing scripts and doing the same process as above locally.


Taffer (approved):
Looks good to me!

DominicDams (approved):

PR #663, opened 2022-10-27
Add SSH KDF to AWS-LC ACVP Tool

billbo-yang:
### Issues:
Resolves [CryptoAlg-1359](https://sim.amazon.com/issues/CryptoAlg-1359)

### Description of changes: 
Currently AWS-LC has SSH KDF within the FIPS boundary, but doesn't support testing it with its ACVP tool. This PR adds support for SSH KDF to the ACVP tool.

### Call-outs:
Due to the architecture of the tool, I needed to put the logic for SSH KDF in the same file that originally only handled TLS KDF instead of in its own dedicated file. There is some precedent for this, which can be seen in how the tool handles ECDSA.

### Testing:
Changes tested by running the ACVP tool on SSH KDF in online mode to fetch the appropriate vectors, compute the result, then upload said results to be compared to the expected results by NIST. Also tested by adding SSH KDF support to the ACVP tool testing script and doing the same process locally. I also tested TLS KDF to make sure that the changes made to the file didn't break it.

the ISC license.


util/fipstools/acvp/acvptool/subprocess/kdf-components.go
```diff
@@ -0,0 +1,349 @@
+// Copyright (c) 2020, Google Inc.
```
Taffer, line 1: If this wasn't actually copied from Google, you'll have to change this.

Taffer (approved):

util/fipstools/acvp/acvptool/subprocess/kdf-components.go
andrewhop, line 1: NP: can you make 2 reviews (in any order):
1. Move/rename files with no changes
2. Update logic

As is it's hard to tell what you changed and we would have to review the entire file as if it is a new file

util/fipstools/acvp/modulewrapper/modulewrapper.cc
```diff
@@ -2155,6 +2175,34 @@ static bool HKDF(const Span<const uint8_t> args[], ReplyCallback write_reply) {
...
+  const Span<const uint8_t> session_id = args[2];
+  const Span<const uint8_t> type = args[3];
+  const Span<const uint8_t> out_bytes = args[4];
+  const EVP_MD *md = MDFunc();
+
+
+  char type_char;
+  memcpy(&type_char, type.data(), sizeof(type_char));
+  unsigned int out_bytes_uint;
+  memcpy(&out_bytes_uint, out_bytes.data(), sizeof(out_bytes_uint));
```
andrewhop, line 2191: Why did you copy just these two? Can you just use type.data and out_bytes.data down below, or should you copy all of them?

util/fipstools/acvp/acvptool/subprocess/kdf-components.go
billbo-yang: Ok, I'll make sure to do that in the future. Did you want me to close this PR and resubmit it as well?

util/fipstools/acvp/modulewrapper/modulewrapper.cc
billbo-yang: The function in `openssl/sshkdf.h` requires those types, and implicit casting as well as using `reinterpret_cast` causes the build to fail, so this was my solution. I believe there is precedent in other parts of the file where something similar is done as well.

util/fipstools/acvp/acvptool/subprocess/kdf-components.go
andrewhop: Yes please, I'd prefer not to review the whole 349 line file.

billbo-yang:
Closing and reopening as two separate PRs because of renamed file.

PR #664, opened 2022-10-28
Added Benchmarks for EC Key generation

DominicDams:
### Issues:
Resolves #CryptoAlg-1376

### Description of changes: 
Adds benchmarks for Key Generation to the NIST EC Curves

### Call-outs:
None

### Testing:
Manually ran benchmarks.

the ISC license.


Taffer (dismissed):

tool/speed.cc
```diff
@@ -1029,8 +1029,25 @@ static bool SpeedECDSACurve(const std::string &name, int nid,
...
+#endif
+      })) {
+    return false;
+  }
+
+  results.Print(name + " key generation");
+
+  if (!EC_KEY_generate_key(key.get())) {
     return false;
   }
```
WillChilds-Klein, line 1052: is this check this make redundant by the null check + return on L1045?

DominicDams: Kinda? I'd say there's not really a reason we should expect this to fail if our key generation benchmark succeeded, but it doesn't hurt and if this failed we'd probably want the benchmarking run to fail.

WillChilds-Klein: sure, it's not inside a timed block so i suppose it doesn't hurt.

WillChilds-Klein (dismissed):

tool/speed.cc
```diff
@@ -1029,8 +1029,25 @@ static bool SpeedECDSACurve(const std::string &name, int nid,
...
+#if defined(BORINGSSL_FIPS)
+        return EC_KEY_generate_key_fips(key.get()) == 1;
+#else
+        return EC_KEY_generate_key(key.get()) == 1;
+#endif
+      })) {
+    return false;
+  }
+
+  results.Print(name + " key generation");
```
andrewhop, line 1048: What is the final name in the results? Is it "ECDSA {curve} key generation" instead of just "{curve} key generation"? If so should this be in a separate function that just does key generation since ECDH and ECDSA use the same keygen and this performance is not tied to ECDSA specifically?

andrewhop (commented):
Can you include a snippet of the speed output for the new benchmarks for text and json output? It doesn't need to be anything scientific just looking to make sure the output looks sane.

tool/speed.cc
DominicDams:
```
 Did 177557 ECDSA P-256 key generation operations in 20000036us (8877.8 ops/sec)
```

tool/speed.cc
```diff
@@ -1022,15 +1022,42 @@ static bool SpeedECDHCurve(const std::string &name, int nid,
...
                             const std::string &selected) {
   if (!selected.empty() && name.find(selected) == std::string::npos) {
     return true;
   }
 
   BM_NAMESPACE::UniquePtr<EC_KEY> key(EC_KEY_new_by_curve_name(nid));
-  if (!key ||
-      !EC_KEY_generate_key(key.get())) {
+  if(!key ||
+     !EC_KEY_generate_key(key.get())) {
```
andrewhop, line 1060:
```suggestion
  if (!key ||
      !EC_KEY_generate_key(key.get())) {
```

andrewhop (dismissed):

tool/speed.cc
```diff
@@ -1022,6 +1022,33 @@ static bool SpeedECDHCurve(const std::string &name, int nid,
   return true;
 }
 
+static bool SpeedECGenCurve(const std::string &name, int nid,
```
dkostic, line 1025: We are generating a key, not a curve:
```suggestion
static bool SpeedECKeyGenCurve(const std::string &name, int nid,
```

tool/speed.cc
```diff
@@ -1072,6 +1099,14 @@ static bool SpeedECDH(const std::string &selected) {
          SpeedECDHCurve("ECDH secp256k1", NID_secp256k1, selected);
 }
 
+static bool SpeedECGen(const std::string &selected) {
+  return SpeedECGenCurve("Generate P-224", NID_secp224r1, selected) &&
```
dkostic, line 1103:
```suggestion
  return SpeedECGenCurve("Key generation P-224", NID_secp224r1, selected) &&
```

tool/speed.cc
```diff
@@ -1022,6 +1022,33 @@ static bool SpeedECDHCurve(const std::string &name, int nid,
...
+  }
+
+  TimeResults results;
+  if (!TimeFunction(&results, [&key]() -> bool {
+// We run extra checks on the key in FIPS mode
+#if defined(BORINGSSL_FIPS)
+        return EC_KEY_generate_key_fips(key.get()) == 1;
+#else
+        return EC_KEY_generate_key(key.get()) == 1;
+#endif
```
dkostic, line 1043: both functions are always available if I'm not mistaken (regardless of the FIPS flag), so why not measure and report performance for both?

tool/speed.cc
DominicDams: This should be read as EC Key-Gen Curve (i.e. Generate a key for the specific curve, rather than generate a curve). It's a bit confusing but I was trying to stay consistent with how our other EC tests behave/are named

tool/speed.cc
DominicDams: We could do that, I just figured we'd want fips for fips, and non-fips for non-fips. If you think this is valuable I'll make a quick change for it

tool/speed.cc
dkostic: not sure I get it, I'm suggesting to to rename the function to `SpeedECKeyGenCurve` which would mean exactly what you are saying: EC KeyGen for a Curve.

tool/speed.cc
```diff
@@ -1022,6 +1022,38 @@ static bool SpeedECDHCurve(const std::string &name, int nid,
   return true;
 }
 
+static bool SpeedECKeyGenCurve(const std::string &name, int nid,
```
dkostic, line 1025: Calling the higher level EVP function `EVP_PKEY_keygen` ([link](https://github.com/awslabs/aws-lc/blob/8569259849f6397431e0941ec53d701fef241fed/include/openssl/evp.h#L692)) with appropriate context should take care of calling the underlying FIPS/non-FIPS implementation by itself ([link](https://github.com/awslabs/aws-lc/blob/7792c41f96d9b943cc4006ea6d7dfe733df815f5/crypto/fipsmodule/evp/p_ec.c#L228)) and you wouldn't have to worry about that here.

tool/speed.cc
```diff
@@ -1073,6 +1104,14 @@ static bool SpeedECDH(const std::string &selected) {
          SpeedECDHCurve("ECDH secp256k1", NID_secp256k1, selected);
 }
 
+static bool SpeedECGen(const std::string &selected) {
```
dkostic, line 1107:
```suggestion
static bool SpeedECKeyGen(const std::string &selected) {
```

tool/speed.cc
```diff
@@ -1867,6 +1906,8 @@ bool Speed(const std::vector<std::string> &args) {
      !SpeedRandom(selected) ||
      !SpeedECDH(selected) ||
      !SpeedECDSA(selected) ||
+     !SpeedECGen(selected) ||
```
dkostic, line 1909:
```suggestion
     !SpeedECKeyGen(selected) ||
```

dkostic (approved):

tool/speed.cc
```diff
@@ -1867,6 +1906,8 @@ bool Speed(const std::vector<std::string> &args) {
      !SpeedRandom(selected) ||
      !SpeedECDH(selected) ||
      !SpeedECDSA(selected) ||
+     !SpeedECKeyGen(selected) ||
+     // OpenSSL 1.0 doesn't support Scrypt
```
andrewhop, line 1910: Typo?
```suggestion
```

andrewhop (approved):

dkostic (approved):

PR #668, opened 2022-11-01
Add SSHKDF to ACVP Tool

billbo-yang:
Member
### Issues:
Resolves CryptoAlg-1359

### Description of changes:
Currently AWS-LC has SSH KDF within the FIPS boundary, but doesn't support testing it with its ACVP tool. This PR adds support for SSH KDF to the ACVP tool.

### Call-outs:
Due to the architecture of the tool, I needed to put the logic for SSH KDF in the same file that originally only handled TLS KDF instead of in its own dedicated file. There is some precedent for this, which can be seen in how the tool handles ECDSA.

### Testing:
Changes tested by running the ACVP tool on SSH KDF in online mode to fetch the appropriate vectors, compute the result, then upload said results to be compared to the expected results by NIST. Also tested by adding SSH KDF support to the ACVP tool testing script and doing the same process locally. I also tested TLS KDF to make sure that the changes made to the file didn't break it.

the ISC license.


Taffer (approved):

util/fipstools/acvp/acvptool/subprocess/kdf-components.go
```diff
@@ -55,134 +65,282 @@ type kdfCompTestGroupResponse struct {
...
-func (k *kdfComp) Process(vectorSet []byte, m Transactable) (interface{}, error) {
-	var parsed kdfCompVectorSet
-	if err := json.Unmarshal(vectorSet, &parsed); err != nil {
-		return nil, err
+var ivLenMap = map[string]uint32{
+	"AES-128": 16,
+	"AES-192": 16,
+	"AES-256": 16,
+	"TDES":    8,
+}
```
andrewhop, line 90: These constants will never change but is there somewhere else we can pull them from that are already defined?

util/fipstools/acvp/acvptool/subprocess/kdf-components.go
```diff
@@ -55,134 +65,282 @@ type kdfCompTestGroupResponse struct {
...
+	}
+
+	sessionId, err := hex.DecodeString(test.SessionIdHex)
+	if err != nil {
+		return err
+	}
+
+	// NIST SP 800-135 says that the lengths of the IVs and encryption keys are determined by the supported block ciphers
+	// We can look at https://github.com/usnistgov/ACVP-Server/blob/master/gen-val/json-files/kdf-components-ssh-1.0/internalProjection.json
+	// to determine what these lengths are
```
andrewhop, line 255: If we can look at the data why do we need the hard coded maps?

util/fipstools/acvp/acvptool/subprocess/kdf-components.go
```diff
@@ -55,134 +65,282 @@ type kdfCompTestGroupResponse struct {
...
+	ivLen := ivLenMap[group.Cipher]
+	encyrptionKeyLen := encyrptionKeyLenMap[group.Cipher]
+	integrityKeyLen := integrityKeyLenMap[group.Hash]
+
+	// Definitions for last argument of Transact() can be found in aws-lc/include/openssl/sshkdf.h
+	initialIvClient, err := m.Transact(method, 1,
+		secretVal, hashVal, sessionId,
+		uint32le(65), uint32le(ivLen))
```
andrewhop, line 263: What do these 65/66/67 mean?

util/fipstools/acvp/acvptool/subprocess/kdf-components.go
```diff
@@ -55,134 +65,282 @@ type kdfCompTestGroupResponse struct {
...
+		if err != nil {
+			return nil, err
+		}
+
+		for _, test := range group.Tests {
+			if parsed.Mode == "ssh" {
+				err = HandleSSH(test, k, m, method, group, &response)
+			} else if parsed.Mode == "tls" || parsed.Mode == "KDF" {
+				err = HandleTLS(test, k, m, method, group, &response)
 			}
```
andrewhop, line 339: Does this need to handle the case mode != ssh or tls?

billbo-yang: So as is if the script gets something where the case mode isn't described, it'll just output an empty json, which should fail any test. But it'd probably be a good idea to handle it more elegantly so I'll do that 👍

util/fipstools/acvp/acvptool/subprocess/kdf-components.go
billbo-yang: These are the `type` values that the SSHKDF C function uses, defined [here](https://github.com/awslabs/aws-lc/blob/d998125d62c8bb009e18f4a95757c3c05f5709f9/include/openssl/sshkdf.h#L26). Will pull them into dedicated variables like I did the maps.

util/fipstools/acvp/acvptool/subprocess/kdf-components.go
billbo-yang: So the data for the hardcoded maps were found by looking through the NIST test vectors and determining the lengths of the IVs and keys for each cipher. For example if you look at the test vectors, you'll notice that the IV lengths for all vectors using AES-128 will be the same length, and the same will be the case for the encryption key and session ID for the other ciphers.

Getting that information is a matter of looking through the file, and since it never changes I don't think its worth finding a way to do that programmatically.

util/fipstools/acvp/acvptool/subprocess/kdf-components.go
billbo-yang: Not that I'm aware of in AWS-LC. I've looked in [`sshkdf.h`](https://github.com/awslabs/aws-lc/blob/d998125d62c8bb009e18f4a95757c3c05f5709f9/include/openssl/sshkdf.h#L43), Chris' PR for [SSHKDF](https://github.com/awslabs/aws-lc/pull/656/files), and all over the golang files for the ACVP tool but didn't really find anything.

Taffer: `cipher.h` has `EVP_MAX_IV_LENGTH` = 16, but that's not particularly better, and there's nothing IV-related = 8.

util/fipstools/acvp/acvptool/subprocess/kdf-components.go
Taffer: `strlen(iv_buff) / 2` or something? Then you don't have to hard-code it for the cipher and it'll still  work if they add ChaCha20 or "AES II" in the future.

Taffer (approved):

andrewhop (approved):

PR #670, opened 2022-11-03
Refactoring EC key check for FIPS

dkostic:
### Issues:
N/A

### Description of changes: 
Refactoring EC key check for FIPS.

### Call-outs:

### Testing:

the ISC license.


crypto/fipsmodule/ec/ec_key.c
```diff
@@ -329,92 +329,77 @@ int EC_KEY_check_key(const EC_KEY *eckey) {
...
-  sig_der = OPENSSL_malloc(sign_len);
-  if (!sig_der ||
-      !EVP_DigestSign(ctx, sig_der, &sign_len, msg, msg_len)) {
-    goto err;
-  }
-  if (boringssl_fips_break_test("ECDSA_PWCT")) {
-    msg[0] = ~msg[0];
+ 
+  BIGNUM *x = BN_new();
+  BIGNUM *y = BN_new();
```
darylmartin100, line 339: For completeness sake, I would add null checks for both x,y here.  It looks like the subsequent functions (ec_felem_to_bignum & BN_free & others) handle null correctly it's still better to add the checks here.

dkostic: done

crypto/fipsmodule/ec/ec_key.c
```diff
@@ -329,92 +329,81 @@ int EC_KEY_check_key(const EC_KEY *eckey) {
...
 int EC_KEY_check_fips(const EC_KEY *key) {
-  // We have to avoid the underlying |EVP_DigestSign| and |EVP_DigestVerify|
-  // services in |EVP_EC_KEY_check_fips| updating the indicator state, so we
-  // lock the state here.
-  FIPS_service_indicator_lock_state();
   int ret = 0;
+  FIPS_service_indicator_lock_state();
 
-  if (EC_KEY_is_opaque(key)) {
+  if (EC_KEY_is_opaque(key) ||
```
nebeid, line 397: I think we can separate again these 3 checks because the last 2 have their own return error so that it's clearer to the caller and to the tests.

crypto/fipsmodule/ec/ec_key.c
```diff
@@ -329,92 +329,81 @@ int EC_KEY_check_key(const EC_KEY *eckey) {
...
-    }
-    BN_free(x);
-    BN_free(y);
-    if (check_ret == 0) {
-      goto end;
-    }
-  }
-
   if (key->priv_key) {
-    if (!EVP_EC_KEY_check_fips((EC_KEY*)key)) {
```
nebeid, line 417: I should have probably noticed and asked this question much earlier; when `key->priv_key` is true, we do this pairwise consistency test (PCT) which uses signing and verifying. But in l.398, where `EC_KEY_check_key()` was called, another PCT was performed using multiplying the private key by the base point and comparing the public key. The latter agrees with [SP 800-56Ar3, Sec 5.6.2.1.4 Owner Assurance of Pair-wise Consistency](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf#%5B%7B%22num%22%3A108%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C96%2C289%2C0%5D). Do we still need the sign/verify PCT? Am I missing something?

dkostic:
Closing the PR until we clear up some things with the lab.

PR #142, opened 2021-04-30
Implements an additional fork detection mechanism: pthread_atfork

torben-hansen:
### Issues:

CryptoAlg-727

### Description of changes: 

AWS-LC currently has one fork detection mechanisms: `WIPEONFORK`. This PR implements an additional fork detection mechanism using the pthread library function `pthread_atfork`. Using this function, we can register a handler that is executed in the child after a fork call `fork()` that is used to flip a sentinel fork detection value. This change is intended to add defense-in-depth to the fork detection.

The code flow is roughly as follows:

* The fork detection is lazily initialised
* On first call: `init_fork_detect` is only called by the first thread entering it
* mmap a page that is acting as the sentinel informing whether a fork has occurred or not
* Initialise the two fork detection mechanisms
* If initialisation is successful, set the sentinel value to 1 (indicating no fork has occurred). The value 0 indicates that a fork has occurred.

### Call-outs:

Emphasis has been given to keeping the generation number logic in the function `CRYPTO_get_fork_generation` as is.

Had to move things around to be able to inject ignore flags into the code path for testing. This probably makes merging harder, but beneficial for possible future extensions.

### Testing:

Added extra dimensions that test each fork detection mechanism in isolation. This ensures that each mechanism is correctly initialised and can properly detect a fork. Existing tests cover the full combination of the mechanisms.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #673, opened 2022-11-04
Remove pthread_atfork fork detection mechanism

torben-hansen:
### Description of changes: 

#142 added `pthread_atfork()` as a secondary fork detection mechanism. However, we have observed several issues consuming AWS-LC artifacts due to the symbol `pthread_atfork` being handled differently in distributions (and in a non-portable way). This affects our ability to distribute AWS-LC to our intended places. See internal document for details.

Therefore, completely remove `pthread_atfork()` from the code-base.

### Call-outs:

This change is OK for two reasons:
* `MADV_WIPEONFORK` captures forks more broadly than `pthread_atfork()` (if supported). For example, the former captures forks through `clone()` while the latter does not.
* The decisional logic that determines whether prediction resistance should be disabled or not, depends on `MADV_WIPEONFORK` being supported before making the "disable" decision. That is, even if `phtread_atfork()` is supported, prediction resistance is only disabled if `MADV_WIPEONFORK` is also supported (and they both initialise without errors at run-time).
* No `pthread_atfork()` behaviour or functionality was ever exposed publicly, so we can remove this safely without consumer incompatibility issues. Neither does this PR change behaviour in how `RAND_bytes()` behaves from a public perspective.

### Testing:

Remove `pthread_atfork()` specific tests and replace with tests that forcefully disables `MADV_WIPEONFORK` to test those code-paths (which in the future could potentially be different from `--fork_unsafe_buffering`).

the ISC license.

### DIFF
--- a/util/all_tests.json
+++ b/util/all_tests.json
@@ -55,14 +55,14 @@
   {
     "comment": "No RDRAND and without fork detection",
     "cmd": ["crypto/urandom_test"],
-    "env": ["OPENSSL_ia32cap=~0x4000000000000000", "BORINGSSL_IGNORE_MADV_WIPEONFORK=1", "BORINGSSL_IGNORE_PTHREAD_ATFORK=1"],
+    "env": ["OPENSSL_ia32cap=~0x4000000000000000", "BORINGSSL_IGNORE_MADV_WIPEONFORK=1"],
     "skip_valgrind": true,
     "target_arch": "x86"
   },
   {
     "comment": "Potentially with RDRAND, but not Intel, and without fork detection",
     "cmd": ["crypto/urandom_test"],
-    "env": ["OPENSSL_ia32cap=~0x0000000040000000", "BORINGSSL_IGNORE_MADV_WIPEONFORK=1", "BORINGSSL_IGNORE_PTHREAD_ATFORK=1"],
+    "env": ["OPENSSL_ia32cap=~0x0000000040000000", "BORINGSSL_IGNORE_MADV_WIPEONFORK=1"],
     "skip_valgrind": true,
     "target_arch": "x86"
   },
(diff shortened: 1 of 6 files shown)


sdkellyAWS (approved):

skmcgrail (approved):

util/all_tests.json
```diff
@@ -72,25 +72,13 @@
...
-    "cmd": ["crypto/crypto_test", "--gtest_filter=ForkDetect.*"],
-    "env": ["BORINGSSL_IGNORE_PTHREAD_ATFORK=1"],
-    "skip_valgrind": true
-  },
-  {
-    "comment": "Run fork detection test suite with only pthread_atfork enabled",
+    "comment": "Run fork detection test suite without MADV_WIPEONFORK enabled",
     "cmd": ["crypto/crypto_test", "--gtest_filter=ForkDetect.*"],
     "env": ["BORINGSSL_IGNORE_MADV_WIPEONFORK=1"],
     "skip_valgrind": true
```
andrewhop, line 84: What is is this test doing if the only fork detection logic is disabled?

torben-hansen: It tests the fall-back logic to prediction resistance. That is, the tests performs some forking and other uniqueness breaking events that could cause state duplication. We must still protect against this even though fork detection mechanism are not supported.

Right now, this is somewhat indistinguishable from the test case with `--fork_unsafe_buffering`. But I touched a bit on this in the PR description:
> Remove pthread_atfork() specific tests and replace with tests that forcefully disables MADV_WIPEONFORK to test those code-paths (which in the future could potentially be different from --fork_unsafe_buffering).

dkostic (approved):

andrewhop (approved):

darylmartin100 (approved):

PR #674, opened 2022-11-04
Add Dilithium3 higher level API integration

jakemas:
### Description of changes: 
This PR introduces the EVP_PKEY APIs for Dilithium3, utilizing key generation, sign and verify. This introduction follows a similar pattern as the import of CRYSTALS Kyber in https://github.com/awslabs/aws-lc/pull/343.

A code review was conducted in the original PR https://github.com/awslabs/aws-lc/pull/637 that was lost due to a merge from main. :(

### Call-outs:

**Addition to EVP interface**
This PR also makes Dilithium3 keys available to the EVP PKEY interface including a subset of methods for key management. As the Dilithium signing procedure is a single-shot signing algorithm that does not use a pre-hash, we follow much of the same structure as the EVP interface for Ed25519. For example, to sign and verify we use the EVP operations `EVP_DigestSign` and `EVP_DigestVerify` with hash `|type|` set to NULL.

**DRBG**

### Testing:

```
... other existing tests ...

[----------] 7 tests from Dilithium3Test
[       OK ] Dilithium3Test.KeyGeneration (1 ms)
[       OK ] Dilithium3Test.KeySize (1 ms)
[ RUN      ] Dilithium3Test.DilithiumMarshal
[       OK ] Dilithium3Test.DilithiumMarshal (0 ms)
[ RUN      ] Dilithium3Test.SIGOperations
[       OK ] Dilithium3Test.SIGOperations (5 ms)
[       OK ] Dilithium3Test.KAT (403 ms)
[----------] 7 tests from Dilithium3Test (411 ms total)
... other existing tests ...

[==========] 1281 tests from 99 test suites ran. (59364 ms total)
[  PASSED  ] 1281 tests.
```

the ISC license.


crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,109 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jakemas, line 1: @[dkostic](https://github.com/dkostic) dkostic [27 days ago](https://github.com/awslabs/aws-lc/pull/637/files#r1002083002)
Regarding this:

> Dilithium offers both deterministic and randomized signing.

How do we handle this? In Dilithium's source code, the choice between deterministic and randomized signing is made based on a compile time flag which means we support only one of those two.

- Can we support only one signing method?
- If so, which one? Should we specify and document this?
- If we have to support both signing methods then we need to implement them both.
- Does the signing method implie the verifying method or is the verification always the same regadless of the signing method?

@[jakemas](https://github.com/jakemas) jakemas [16 days ago](https://github.com/awslabs/aws-lc/pull/637/files#r1010808091)
For now we will use deterministic signing.

Dilithium offers both deterministic and randomized modes of producing signatures. While the deterministic version may be better suited in constrained environments, or in devices that lack good sources of entropy, it can facilitate differential fault attacks that can be used to recover the private key. Randomized signing can mitigate against such attacks, but forces the security of the signature scheme to be dependent on a good source of randomness – as nonce reuse can lead to trivial secret key recovery.

While discussions are still underway about the best path forward, [a middle ground between the two methodologies known as hedging is being considered for use in Dilithium](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/1mQjngj_2Po/m/-p4RKXGQAwAJ), whereby the signature scheme produces randomized signatures that are seeded in part by the message and the key along side a source of randomness. We expect this method of hedged signing in all applications for which there is not a specific need for deterministic signing.

For interop we note that [liboqs](https://github.com/open-quantum-safe/liboqs/blob/55bc8cb38dbc01fadb324a95e00abb880cb03873/src/sig/dilithium/pqcrystals-dilithium_dilithium3_ref/config.h#L6) also uses deterministic signing.

crypto/evp_extra/pq_sig_kat_tests_dilithium3.cc
```diff
@@ -0,0 +1,77 @@
...
+  ASSERT_TRUE(t->GetAttribute(&count, "count"));
+  ASSERT_TRUE(t->GetBytes(&seed, "seed"));
+  ASSERT_TRUE(t->GetAttribute(&mlen, "mlen"));
+  ASSERT_TRUE(t->GetBytes(&msg, "msg"));
+  ASSERT_TRUE(t->GetBytes(&pk, "pk"));
+  ASSERT_TRUE(t->GetBytes(&sk, "sk"));
+  ASSERT_TRUE(t->GetAttribute(&smlen, "smlen"));
+  ASSERT_TRUE(t->GetBytes(&sm, "sm"));
+
+  //the KAT files generated by the dilithium team use the optional APIs that
```
jakemas, line 28: @[dkostic](https://github.com/dkostic) dkostic [27 days ago](https://github.com/awslabs/aws-lc/pull/637/files#r1002063413)
what does it mean optional API?

crypto/evp_extra/pq_sig_kat_tests_dilithium3.cc
```diff
@@ -0,0 +1,77 @@
...
+  //create a signature for a message m and append the message to the end of the
+  //signature. We only want to bring the APIs that create and verify just the
```
jakemas, line 30: @[dkostic](https://github.com/dkostic) dkostic [27 days ago](https://github.com/awslabs/aws-lc/pull/637/files#r1002064009)
We only want to bring the APIs that create and verify just the signature.

Why is that?

crypto/dilithium/p_dilithium3_asn1.c
```diff
@@ -0,0 +1,208 @@
...
+    return 0;
+  }
+
+  OPENSSL_memcpy(out, key->pub, DILITHIUM3_PUBLIC_KEY_BYTES);
+  *out_len = DILITHIUM3_PUBLIC_KEY_BYTES;
+  return 1;
+}
+
+static int dilithium3_pub_decode(EVP_PKEY *out, CBS *params, CBS *key) {
+  // See https://datatracker.ietf.org/doc/draft-ietf-lamps-dilithium-certificates/ section 4.
```
jakemas, line 103: @[dkostic](https://github.com/dkostic) dkostic [27 days ago](https://github.com/awslabs/aws-lc/pull/637/files#r1001987606)
In Section 4 in the draft there is this:

> ...The algorithms defined in this document always encode the public key as TODO.

Does our implementation depend on the definition of the `TODO`?

jakemas: Yes, this PoC is being built to support the draft, so that we can test interop as the draft develops.

crypto/dilithium/p_dilithium3_asn1.c
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  return dilithium3_set_pub_raw(out, CBS_data(key), CBS_len(key));
+}
+
+static int dilithium3_pub_encode(CBB *out, const EVP_PKEY *pkey) {
+  const DILITHIUM3_KEY *key = pkey->pkey.ptr;
+
+  // See https://datatracker.ietf.org/doc/draft-ietf-lamps-dilithium-certificates/ section 4.
+  // TODO: finalize this definition - OCTETSTRING to BITSTRING conversion.
```
jakemas, line 117: @[dkostic](https://github.com/dkostic) dkostic [27 days ago](https://github.com/awslabs/aws-lc/pull/637/files#r1001988322)
is this TODO for aws-lc or for the draft? :)

jakemas: Kind of both, the draft will come first, but it's to clarify in this code that these are not final.

crypto/dilithium/p_dilithium3_asn1.c
```diff
@@ -0,0 +1,208 @@
...
+  CBB spki, algorithm, oid, key_bitstring;
+  if (!CBB_add_asn1(out, &spki, CBS_ASN1_SEQUENCE) ||
+      !CBB_add_asn1(&spki, &algorithm, CBS_ASN1_SEQUENCE) ||
+      !CBB_add_asn1(&algorithm, &oid, CBS_ASN1_OBJECT) ||
+      !CBB_add_bytes(&oid, dilithium3_asn1_meth.oid, dilithium3_asn1_meth.oid_len) ||
+      !CBB_add_asn1(&spki, &key_bitstring, CBS_ASN1_BITSTRING) ||
+      !CBB_add_u8(&key_bitstring, 0 /* padding */) ||
```
jakemas, line 124: @[andrewhop](https://github.com/andrewhop) andrewhop [on Oct 4](https://github.com/awslabs/aws-lc/pull/637/files#r987427443)
What is this padding for?

jakemas: Still working details out like this in the draft spec, but essentially, the padding is required to make the bit string total size to 15616 bit -> 1952 bytes as per the [Dilithium spec](https://pq-crystals.org/dilithium/).

crypto/dilithium/p_dilithium_test.cc
```diff
@@ -0,0 +1,289 @@
...
+  EVP_PKEY *dilithium_pkey = EVP_PKEY_new();
+  ASSERT_NE(dilithium_pkey, nullptr);
+
+  EXPECT_TRUE(EVP_PKEY_keygen_init(dilithium_pkey_ctx));
+  EXPECT_TRUE(EVP_PKEY_keygen(dilithium_pkey_ctx, &dilithium_pkey));
+  ASSERT_NE(dilithium_pkey->pkey.ptr, nullptr);
+  const DILITHIUM3_KEY *dilithium3Key = (DILITHIUM3_KEY *)(dilithium_pkey->pkey.ptr);
+
+  // New raw public key
+  EVP_PKEY *new_public = EVP_PKEY_new_raw_public_key(EVP_PKEY_DILITHIUM3,
```
jakemas, line 116: @[andrewhop](https://github.com/andrewhop) andrewhop [on Oct 4](https://github.com/awslabs/aws-lc/pull/637/files#r987430623)
Instead of directly accessing the raw key through the pointer should this test the ASN1 encoding and decoding with EVP_marshal_public_key and EVP_parse_public_key?

jakemas: This test is specifically targeting `EVP_PKEY_new_raw_public_key `and `EVP_PKEY_new_raw_private_key`. But I agree that we should test `EVP_marshal_public_key` and `EVP_parse_public_key` as well. [See the newly added marshal/parse tests here](https://github.com/awslabs/aws-lc/blob/47415745b8e74604e182b38ee24fdd01d3d8ad5e/crypto/dilithium/p_dilithium_test.cc#L194-L222).

crypto/evp_extra/pq_sig_kat_tests_dilithium3.cc
jakemas: The Dilithium authors added extra APIs than what the NIST call for proposals required.

crypto/evp_extra/pq_sig_kat_tests_dilithium3.cc
jakemas: We have chosen to select the two APIs `pqcrystals_dilithium3_ref_signature` and `pqcrystals_dilithium3_ref_verify` to be consistent with existing EVP signature functionality. This is also consistent with other implementations, such as within [LibOQS](https://github.com/open-quantum-safe/liboqs/blob/main/src/sig/dilithium/sig_dilithium_3.c#L33-L34).

If we have requirements for the other option (append message) these can be added in later.

crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,109 @@
...
+
+#include <openssl/evp.h>
+#include <openssl/err.h>
+#include <openssl/mem.h>
+
+#include "../fipsmodule/evp/internal.h"
+#include "../evp_extra/internal.h"
+#include "sig_dilithium.h"
+
+static int pkey_dilithium3_keygen(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey) {
```
jakemas, line 12: @[andrewhop](https://github.com/andrewhop) andrewhop [on Oct 4](https://github.com/awslabs/aws-lc/pull/637/files#r987414686)
NP: missing return check, looking at sign.c this function can't fail, but we should be consistent.

jakemas: fixed in [46bdba8](https://github.com/awslabs/aws-lc/pull/674/commits/46bdba84a5986bb71374e0f3e82805b0ece09124)

crypto/dilithium/p_dilithium_test.cc
```diff
@@ -0,0 +1,289 @@
...
+  EXPECT_EQ(priv_len, 4000u);
+  ASSERT_TRUE(EVP_PKEY_get_raw_private_key(privkey.get(), priv_buf, &priv_len));
+
+  // The private key must encode properly.
+  ASSERT_TRUE(CBB_init(cbb.get(), 0));
+  ASSERT_TRUE(EVP_marshal_private_key(cbb.get(), privkey.get()));
+  ASSERT_TRUE(CBB_finish(cbb.get(), &der, &der_len));
+  free_der.reset(der);
+
+  // The private key must parse properly.
```
andrewhop, line 227: Missing a comparison that dilithium_priv_from_der is equal to the original key.

crypto/dilithium/p_dilithium_test.cc
```diff
@@ -0,0 +1,289 @@
...
+  // sign a message
+  bssl::ScopedEVP_MD_CTX md_ctx;
+  uint8_t signature[DILITHIUM3_SIGNATURE_BYTES];
+  size_t signature_len = DILITHIUM3_SIGNATURE_BYTES;
+  std::vector<uint8_t> msg = {
+      0x4a, 0x41, 0x4b, 0x45, 0x20, 0x4d, 0x41, 0x53, 0x53, 0x49,
+      0x4d, 0x4f, 0x20, 0x41, 0x57, 0x53, 0x32, 0x30, 0x32, 0x32, 0x2e};
+  std::vector<uint8_t> badmsg = {
+      0x4a, 0x41, 0x4b, 0x45, 0x20, 0x4d, 0x41, 0x53, 0x53, 0x49,
+      0x4d, 0x4f, 0x20, 0x41, 0x57, 0x53, 0x32, 0x30, 0x32, 0x31, 0x2e};
```
andrewhop, line 259: Any reason you flipped the second to last byte? It took me a second to figure out if anything was different.

jakemas: Yes, I am testing that verification fails on a message that is different to that which was signed, so it was flipped on purpose.

crypto/dilithium/p_dilithium3_asn1.c
```diff
@@ -0,0 +1,208 @@
...
+      !CBB_add_bytes(&oid, dilithium3_asn1_meth.oid, dilithium3_asn1_meth.oid_len) ||
+      !CBB_add_asn1(&pkcs8, &private_key, CBS_ASN1_OCTETSTRING) ||
+      !CBB_add_bytes(&private_key, key->priv, DILITHIUM3_PRIVATE_KEY_BYTES) ||
+      !CBB_flush(out)) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_ENCODE_ERROR);
+    return 0;
+  }
+
+  return 1;
+}
```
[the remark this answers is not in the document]
jakemas: Just as it does with other PKEY structures, see [p_ed25519_asn1.c](https://github.com/awslabs/aws-lc/blob/main/crypto/evp_extra/p_ed25519_asn1.c#L157-L198). All other parameters are omitted with the exception of the private key: 
`return dilithium3_set_priv_raw(out, CBS_data(key), CBS_len(key));`

This is infact simpler than the ed25519 example, as that has an extra layer of the 32-byte seed value in another OCTET string layer that has to be extracted, where as we don't have that for Dilithium.

The tests in `p_dilithium_test.cc` verify that the private key is extracted as desired.

crypto/dilithium/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+**Supported versions.** AWS-LC supports only Dilithium3 algorithm at this point. The NID assigned to Dilithium3 is `NID_DILITHIUM3` and the corresponding `PKEY` identifier is `EVP_PKEY_DILITHIUM3`.
+
+**Source code origin and modifications.** The source code was taken from the primary source of Diltihium at [link](https://github.com/pq-crystals/dilithium), at [commit](https://github.com/pq-crystals/dilithium/commit/3e9b9f1412f6c7435dbeb4e10692ea58f181ee51) as of 9th May 2022.
+The `api.h`, `fips202.h` and `params.h` header files were modified [in this PR](https://github.com/awslabs/aws-lc/pull/655) to support our [prefixed symbols build](https://github.com/awslabs/aws-lc/blob/main/BUILDING.md#building-with-prefixed-symbols).
```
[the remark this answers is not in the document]
jakemas: This fix names kyber explicitly, but in fact the fix works for Dilithium too, as the prefix targets all `pqcrystals` files, of which Dilithium is included.

include/openssl/evp.h
```diff
@@ -124,13 +124,15 @@ OPENSSL_EXPORT int EVP_PKEY_missing_parameters(const EVP_PKEY *pkey);
...
+// ECDSA signature. For a Dilithium key, this returns the signature byte size.
+// For a KEM key, this returns the sum of the size of the public key and the
+// secret key.
 OPENSSL_EXPORT int EVP_PKEY_size(const EVP_PKEY *pkey);
 
 // EVP_PKEY_bits returns the "size", in bits, of |pkey|. For an RSA key, this
 // returns the bit length of the modulus. For an EC key, this returns the bit
-// length of the group order.
+// length of the group order. For a Dilithium key, this returns the bit length
+// of the signature size.
```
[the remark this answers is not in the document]
jakemas: Nope, it really is the signature size. `static int dilithium3_bits(const EVP_PKEY *pkey) {
  return 8 * (DILITHIUM3_SIGNATURE_BYTES);
}`

crypto/dilithium/p_dilithium_test.cc
jakemas: added in [2d9d0b1](https://github.com/awslabs/aws-lc/pull/674/commits/2d9d0b18a8f6c51a7b6d036e23fba0457e758e11)

crypto/dilithium/p_dilithium3_asn1.c
```diff
@@ -0,0 +1,208 @@
...
+    9,
+    dilithium3_pub_decode,
+    dilithium3_pub_encode,
+    dilithium3_pub_cmp,
+    dilithium3_priv_decode,
+    dilithium3_priv_encode,
+    dilithium3_set_priv_raw,
+    dilithium3_set_pub_raw,
+    dilithium3_get_priv_raw,
+    dilithium3_get_pub_raw,
```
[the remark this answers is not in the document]
jakemas: added in [2d9d0b1](https://github.com/awslabs/aws-lc/pull/674/commits/2d9d0b18a8f6c51a7b6d036e23fba0457e758e11)

crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,112 @@
...
+    OPENSSL_PUT_ERROR(EVP, EVP_R_MISSING_PARAMETERS);
+    return 0;
+  }
+
+  if (!EVP_PKEY_set_type(pkey, EVP_PKEY_DILITHIUM3)) {
+    OPENSSL_free(key);
+    return 0;
+  }
+
+  if (DILITHIUM3_keypair(key->pub, key->priv) != 0) {
```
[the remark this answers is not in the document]
jakemas: Added documentation for each of these functions to describe this in [2d9d0b1](https://github.com/awslabs/aws-lc/pull/674/commits/2d9d0b18a8f6c51a7b6d036e23fba0457e758e11). In the process of doing this I also noticed an error in the Dilithium source - opened a PR https://github.com/pq-crystals/dilithium/pull/66.

crypto/dilithium/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Only the reference C implementation of Dilithium3 is currently integrated. The code is in the `pqcrystals-dilithium_dilithium3_ref` folder. The following changes were made to the code.
+
+* `randombytes.{h|c}` are deleted because we are using the randomness generation functions provided by AWS-LC.
+* `rng.{h|c}` are deleted because we are using the randomness generation functions provided by AWS-LC.
+* `symmetric-aes.c` are removed because we are using only the SHA3 based Dilithium.
+* `sign.c`: calls to `randombytes` function is replaced with calls to `pq_custom_randombytes` and the appropriate header file is included (`crypto/rand_extra/pq_custom_randombytes.h`).
+* `sign.c` was modified from `*mlen = -1;` to `*mlen = 0;` within the `crypto_sign_open` function. This was modified as `*mlen = -1;` is attempting to assign `-1` to an unsigned integer, giving the warning: ` warning C4245: '=': conversion from 'int' to 'std::size_t'`. This warning would cause Windows x86 builds to fail.
```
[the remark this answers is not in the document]
jakemas: [2d9d0b1](https://github.com/awslabs/aws-lc/pull/674/commits/2d9d0b18a8f6c51a7b6d036e23fba0457e758e11)

crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,112 @@
...
+
+static int pkey_dilithium3_sign_message(EVP_PKEY_CTX *ctx, uint8_t *sig,
+                                        size_t *siglen, const uint8_t *tbs,
+                                        size_t tbslen) {
+  if (ctx == NULL || ctx->pkey->pkey.ptr == NULL) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_MISSING_PARAMETERS);
+    return 0;
+  }
+
+  DILITHIUM3_KEY *key = ctx->pkey->pkey.ptr;
```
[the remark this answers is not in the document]
jakemas: Okay, we can do this. But this can come later as it will with kyber. Right now I want continue to the X.509 code to get the PoC finished. Should we need it during the X.509 part I'll address it sooner than planned.

crypto/dilithium/sig_dilithium.h
```diff
@@ -0,0 +1,32 @@
...
+#include <openssl/base.h>
+#include <openssl/evp.h>
+
+// The values below are taken from the |api.h| file in the
+// |crypto/dilithium/pqcrystals-dilithium_dilithium_common| directory.
+#define DILITHIUM3_PUBLIC_KEY_BYTES 1952
+#define DILITHIUM3_PRIVATE_KEY_BYTES 4000
+#define DILITHIUM3_SIGNATURE_BYTES 3293
+
+int DILITHIUM3_keypair(uint8_t *public_key,
```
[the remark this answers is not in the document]
jakemas: added in [2d9d0b1](https://github.com/awslabs/aws-lc/pull/674/commits/2d9d0b18a8f6c51a7b6d036e23fba0457e758e11)

crypto/dilithium/p_dilithium_test.cc
jakemas: had to remove in [5e8bfc6](https://github.com/awslabs/aws-lc/pull/674/commits/5e8bfc6d4fecceb8f928c9180d92d3ef96db1e87) as `EXPECT_EQ(1, EVP_PKEY_cmp(dilithium_priv_from_der, privkey.get()));` would fail even though the DER files were exactly the same. This comparison must not be the right one.

crypto/dilithium/p_dilithium3_asn1.c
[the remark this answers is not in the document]
jakemas: Just as it does with other PKEY structures, see [p_ed25519_asn1.c](https://github.com/awslabs/aws-lc/blob/main/crypto/evp_extra/p_ed25519_asn1.c#L157-L198). All other parameters are omitted with the exception of the private key: 
`return dilithium3_set_priv_raw(out, CBS_data(key), CBS_len(key));`

which is then extracted by `dilithium3_set_priv_raw`

This is infact simpler than the ed25519 example, as that has an extra layer of the 32-byte seed value in another OCTET string layer that has to be extracted, where as we don't have that for Dilithium.

The tests in `p_dilithium_test.cc` verify that the private key is extracted as desired.

crypto/dilithium/p_dilithium_test.cc
jakemas: We'd have to [follow what they do in ed25519 ](https://github.com/awslabs/aws-lc/blob/main/crypto/evp_extra/evp_extra_test.cc#L1299-L1304). i.e. `EXPECT_EQ(Bytes(kPrivateKeyPKCS8), Bytes(der, der_len));` but this would mean hard coding in a private key/seed.

andrewhop: Why does EVP_PKEY_cmp fail? Is it comparing the wrong thing? Or comparing past the end of the private key buffer?

include/openssl/evp.h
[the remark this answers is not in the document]
andrewhop: From the existing comment:
* RSA -> bit length of the modulus
* EC -> bit length of the group order
EVP_PKEY_bits returns the "size", in bits, of |pkey| which is talking more about the size of the key not the signature.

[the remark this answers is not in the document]
jakemas: Right, but there are no defining qualities of Dilithium like modulus size or group order.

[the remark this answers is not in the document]
jakemas: I'll keep a note of what is most natural here (this comes up in the x509 part). Public key len is the obvious choice.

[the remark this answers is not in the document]
andrewhop: For signature size that seems to fit with `EVP_PKEY_size`:
> EVP_PKEY_size returns the maximum size, in bytes, of a signature signed by |pkey|.
And it's just a coincidence that for `RSA EVP_PKEY_size == EVP_PKEY_bits`

crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,112 @@
...
+    OPENSSL_PUT_ERROR(EVP, EVP_R_BUFFER_TOO_SMALL);
+    return 0;
+  }
+
+  if (DILITHIUM3_sign(sig, siglen, tbs, tbslen, key->priv) != 0) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_SIGNATURE);
+    return 0;
+  }
+  // The size of the signature that has been written to the output buffer.
+  *siglen = DILITHIUM3_SIGNATURE_BYTES;
```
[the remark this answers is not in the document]
jakemas: fixed in [6319f20](https://github.com/awslabs/aws-lc/pull/674/commits/6319f20aebbebf4df15b02040363465d0d20cebf)

crypto/dilithium/sig_dilithium.h
```diff
@@ -0,0 +1,32 @@
...
+                       uint8_t *secret_key);
+
+int DILITHIUM3_sign(uint8_t *sig, size_t *sig_len,
+                    const uint8_t *message,
+                    size_t message_len,
+                    const uint8_t *secret_key);
```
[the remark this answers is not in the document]
jakemas: fixed in [6319f20](https://github.com/awslabs/aws-lc/pull/674/commits/6319f20aebbebf4df15b02040363465d0d20cebf)

crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,112 @@
...
+    return 1;
+  }
+
+  if (*siglen < DILITHIUM3_SIGNATURE_BYTES) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_BUFFER_TOO_SMALL);
+    return 0;
+  }
+
+  if (DILITHIUM3_sign(sig, siglen, tbs, tbslen, key->priv) != 0) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_SIGNATURE);
```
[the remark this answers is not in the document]
jakemas: fixed in [6319f20](https://github.com/awslabs/aws-lc/pull/674/commits/6319f20aebbebf4df15b02040363465d0d20cebf)

crypto/dilithium/p_dilithium3.c
```diff
@@ -0,0 +1,112 @@
...
+#include <openssl/evp.h>
+#include <openssl/err.h>
+#include <openssl/mem.h>
+
+#include "../fipsmodule/evp/internal.h"
+#include "../evp_extra/internal.h"
+#include "sig_dilithium.h"
+
+static int pkey_dilithium3_keygen(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey) {
+  DILITHIUM3_KEY *key = OPENSSL_malloc(sizeof(DILITHIUM3_KEY));
```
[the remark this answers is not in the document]
jakemas: fixed in [6319f20](https://github.com/awslabs/aws-lc/pull/674/commits/6319f20aebbebf4df15b02040363465d0d20cebf)

include/openssl/evp.h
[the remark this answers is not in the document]
jakemas: Fair enough, it's pub keylen in [OQS](https://github.com/open-quantum-safe/openssl/blob/OQS-OpenSSL_1_1_1-stable/crypto/ec/oqs_meth.c#L990) so we will match here and see what happens. fixed in [037822b](https://github.com/awslabs/aws-lc/pull/674/commits/037822bc91319867c0ab8b8003bcab8d591692eb).

I guess this is all up for interpretation. From OpenSSL manpages: 

> Note that the definition of cryptographic length is specific to the key cryptosystem.

crypto/dilithium/p_dilithium_test.cc
```diff
@@ -0,0 +1,294 @@
...
+  free_der.reset(der);
+
+  // The private key must parse properly.
+  CBS_init(&cbs, der, der_len);
+  EVP_PKEY *dilithium_priv_from_der = EVP_parse_private_key(&cbs);
+  ASSERT_TRUE(dilithium_priv_from_der);
+  const DILITHIUM3_KEY *dilithium3Key_from_der = (DILITHIUM3_KEY *)(dilithium_priv_from_der->pkey.ptr);
+  // The private key dilithium3Key_from_der must be equal to the original key
+  EXPECT_EQ(Bytes(dilithium3Key->priv), Bytes(dilithium3Key_from_der->priv,
+                                              DILITHIUM3_PRIVATE_KEY_BYTES));
```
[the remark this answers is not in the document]
jakemas: added in [7dfab72](https://github.com/awslabs/aws-lc/pull/674/commits/7dfab72128c65390d186c9127bcb165d929d2707)

crypto/dilithium/p_dilithium_test.cc
```diff
@@ -0,0 +1,294 @@
...
+  ASSERT_FALSE(EVP_DigestVerify(md_ctx.get(), signature, signature_len,
+                                badmsg.data(), badmsg.size()));
+
+  // sign the bad message
+  uint8_t signature1[DILITHIUM3_SIGNATURE_BYTES];
+  ASSERT_TRUE(EVP_DigestSign(md_ctx.get(), signature1, &signature_len,
+                             badmsg.data(), badmsg.size()));
+
+  // check that the two signatures are not equal
+  EXPECT_NE(0, memcmp(signature, signature1, signature_len));
```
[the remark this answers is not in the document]
jakemas: added in [6438293](https://github.com/awslabs/aws-lc/pull/674/commits/6438293b9627e80d78da13a9f82cd941f4310292)

crypto/evp_extra/internal.h
```diff
@@ -26,18 +26,25 @@ typedef struct {
...
 extern const EVP_PKEY_ASN1_METHOD rsa_pss_asn1_meth;
 extern const EVP_PKEY_ASN1_METHOD ed25519_asn1_meth;
 extern const EVP_PKEY_ASN1_METHOD x25519_asn1_meth;
 extern const EVP_PKEY_ASN1_METHOD kyber512_asn1_meth;
+extern const EVP_PKEY_ASN1_METHOD dilithium3_asn1_meth;
 
 extern const EVP_PKEY_METHOD ed25519_pkey_meth;
 extern const EVP_PKEY_METHOD x25519_pkey_meth;
 extern const EVP_PKEY_METHOD kyber512_pkey_meth;
-extern const EVP_PKEY_METHOD hkdf_pkey_meth;
```
dkostic, line 40: why is this line removed?

jakemas: This likely came from when I merged main. Why is it included in this list, if it isn't in [p_methods.c](https://github.com/awslabs/aws-lc/blob/main/crypto/evp_extra/p_methods.c#L10-L14)?

jakemas: I guess it comes in to play [here](https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/evp/internal.h#L270)

andrewhop (approved):

dkostic (approved):

PR #682, opened 2022-11-09
Kyber768 integration

dkostic:
### Issues:
CryptoAlg-1306

### Description of changes: 
Adding support for Kyber768 in the exact same way as for Kyber512.

### Call-outs:

### Testing:

the ISC license.


jakemas:
watching this

crypto/kyber/p_kyber_test.cc
```diff
@@ -4,280 +4,463 @@
...
 
-  // Invalidate the ciphertext and run decapsulate on it.
-  ciphertext_alice[0] ^= 1;
+TEST(KyberTest, KEMOperations) {
+  for (auto pkey_id : pkeyIDsToTest) {
+    // Basic functional test for Kyber that simulates two sides
+    // of the key exchange mechanism.
+    size_t shared_secret_len = kyber_shared_secret_size(pkey_id);
+    size_t ciphertext_len = kyber_ciphertext_size(pkey_id);
+    size_t public_key_len = kyber_public_key_size(pkey_id);
```
andrewhop, line 241: Why not call EVP_PKEY_encapsulate/EVP_PKEY_decapsulate with null to get these sizes instead of having to make a custom test mapping?

crypto/kyber/p_kyber_test.cc
```diff
@@ -4,280 +4,463 @@
...
+
+    EVP_PKEY *kyber_pkey = EVP_PKEY_new();
+    ASSERT_NE(kyber_pkey, nullptr);
+
+    EXPECT_TRUE(EVP_PKEY_keygen_init(kyber_pkey_ctx));
+    EXPECT_TRUE(EVP_PKEY_keygen(kyber_pkey_ctx, &kyber_pkey));
+    ASSERT_NE(kyber_pkey->pkey.ptr, nullptr);
+
+    uint8_t *pub  = get_pub_from_kyber_key(kyber_pkey);
+    uint8_t *priv = get_priv_from_kyber_key(kyber_pkey);
```
andrewhop, line 182: Does this need to be the exact same memory address or can it use EVP_PKEY_get_raw_public/private_key to get a copy and get rid of these functions?

crypto/kyber/p_kyber_test.cc
```diff
@@ -4,280 +4,463 @@
...
-  EVP_PKEY *kyber_pkey = EVP_PKEY_new();
-  ASSERT_NE(kyber_pkey, nullptr);
+static const int pkeyIDsToTest[2] = { EVP_PKEY_KYBER512,
+                                      EVP_PKEY_KYBER768 };
 
-  EXPECT_TRUE(EVP_PKEY_keygen_init(kyber_pkey_ctx));
-  EXPECT_TRUE(EVP_PKEY_keygen(kyber_pkey_ctx, &kyber_pkey));
-  ASSERT_NE(kyber_pkey->pkey.ptr, nullptr);
+TEST(KyberTest, KeyGeneration) {
+  for (auto pkey_id : pkeyIDsToTest) {
```
andrewhop, line 88: Did you consider using a [parameterized gtest](https://github.com/google/googletest/blob/main/docs/advanced.md#value-parameterized-tests) like aead_test.c? It accomplishes basically the same thing, but you get nicer test output:
* As is we would see "KeyGeneration failed assert on line 90", and not be able to tell which kyber level failed
* With paramertized tests we would get "KeyGeneration/Kyber512 failed assert on line 90"

crypto/kyber/p_kyber_test.cc
```diff
@@ -4,280 +4,463 @@
...
 }
 
-TEST(Kyber512Test, KEMFailureModes) {
-  size_t shared_secret_len = KYBER512_SHARED_SECRET_BYTES;
-  size_t ciphertext_len = KYBER512_CIPHERTEXT_BYTES;
-  uint8_t shared_secret[KYBER512_SHARED_SECRET_BYTES];
-  uint8_t ciphertext[KYBER512_CIPHERTEXT_BYTES];
+TEST(KyberTest, Kyber512KAT) {
+  RunKAT("crypto/kyber/pqcrystals_kyber_kats/kyber512_kat.txt", EVP_PKEY_KYBER512);
+}
```
andrewhop, line 461: This is basically a parameterized test but not using gtest's features.

crypto/kyber/p_kyber_test.cc
```diff
@@ -4,280 +4,463 @@
 #include <openssl/evp.h>
```
andrewhop, line 4: Is it possible to make this the generic kem_test.cc that can test any kem's basic functionality (key generation, encap/decaps consistency) and kats?

crypto/kyber/p_kyber_test.cc
```diff
@@ -4,280 +4,463 @@
...
+    size_t expected_ct_len = kyber_ciphertext_size(pkey_id);
+
+    EVP_PKEY_CTX *kyber_pkey_ctx = EVP_PKEY_CTX_new_id(pkey_id, nullptr);
+    EVP_PKEY *kyber_pkey = EVP_PKEY_new();
+    EXPECT_TRUE(EVP_PKEY_keygen_init(kyber_pkey_ctx));
+    EXPECT_TRUE(EVP_PKEY_keygen(kyber_pkey_ctx, &kyber_pkey));
+
+    ASSERT_TRUE(EVP_PKEY_encapsulate(kyber_pkey_ctx, NULL, &ciphertext_len, NULL, &shared_secret_len));
+    EXPECT_EQ(shared_secret_len, expected_ss_len);
+    EXPECT_EQ(ciphertext_len, expected_ct_len);
```
andrewhop, line 317: Ahhh I see this test needs a second source of truth to compare EVP_PKEY_encapsulate with null pointers returns the expected size. Do you think it's worth having this test but need to create a second kem -> size mapping up in `kyber_ciphertext_size`?

Maybe if we go the full parametrized route it gets encoded into the test like the `KnownAEAD` struct, it would include the kem, the kat file, and the expected sizes.

crypto/kyber/README.md
```diff
@@ -8,26 +8,25 @@ The AWS-LC team considers the official repository of [Kyber](https://github.com/
...
+and `KYBER_K 3` for Kyber512 and Kyber768, respectively, so that the appropriate code paths are chosen for compilation.
 
 **Source code origin and modifications.** The source code was taken from the primary source of Kyber at [link](https://github.com/pq-crystals/kyber), at [commit](https://github.com/pq-crystals/kyber/tree/faf5c3fe33e0b61c7c8a7888dd862bf5def17ad2) as of September 13th 2021.
-The `api.h`, `fips202.h` and `params.h` header files were modified [in this PR](https://github.com/awslabs/aws-lc/pull/655) to support our [prefixed symbols build](https://github.com/awslabs/aws-lc/blob/main/BUILDING.md#building-with-prefixed-symbols).
 
-Only the reference C implementation of Kyber512 is currently integrated. The code is in the `pqcrystals-kyber_kyber512_ref` folder. The following changes were made to the code.
+Only the reference C implementation of Kyber512 is currently integrated. The code is in the `pqcrystals_kyber_ref_common` folder. The following changes were made to the code.
```
andrewhop, line 18:
```suggestion
Only the reference C implementation of Kyber512 and Kyber768 are currently integrated. The code is in the `pqcrystals_kyber_ref_common` folder. The following changes were made to the code.
```

crypto/kyber/kyber512_ref.c
```diff
@@ -0,0 +1,25 @@
+
```
andrewhop, line 1: Missing copyrights

crypto/kyber/p_kyber.c
```diff
@@ -109,6 +109,108 @@ static int pkey_kyber512_decapsulate(EVP_PKEY_CTX *ctx,
   return 1;
 }
 
+static int pkey_kyber768_keygen(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey) {
```
andrewhop, line 112: I thought a lot about ways to combine these methods. Do you think it's worth the effort to put this size specific data into the data pointer in EVP_PKEY_CTX like EC_PKEY_CTX which lets the EC code be generic across curves? I could be convinced the risk of combining this code and introducing a bug is higher than the chance we have a bug (or one of the "copies" gets of of date) is higher.

I think it could be worth it if we think there's a future where we have 4+ kems in AWS-LC and they could all use the same generic pkey_kem_keygen function instead of having 4 almost identical copies of these functions.

dkostic: I agree that redesigning the implementation as you suggest is worth only if we are going to support more KEMs. As long as we have only Kyber, I think it's ok to keep it simple and straightforward (on account of some code duplication). So let's discuss internally about this.

crypto/kyber/p_kyber_test.cc
dkostic: neat :) didn't know about this type of tests, I'll reimplement.

crypto/kyber/p_kyber.c
jakemas, line 112: [NP: missing return check ... we should be consistent.](https://github.com/awslabs/aws-lc/pull/637/files#r987414686)

dkostic:
Following the internal discussion we had, I am closing this PR in favor of a series of PRs that will bring us to Kyber768/1024 support. First in the series is #705 that only removes Kyber files duplication.

PR #686, opened 2022-11-16
Update benchmark to support externally built aws-lc

andrewhop:
### Description of changes: 
This fixes an issue when building the benchmark with libcrypto that happens to be AWS-LC. This is useful to test AWS-LC FIPS when new benchmarks are added to mainline that we want to run against an existing FIPS branch. Previously when building the `awslc_bm` we passed in just `crypto` which results in CMake using the `crypto` defined in crypto/CMakeLists.txt, not the libcrypto in `AWSLC_INSTALL_DIR`. This change uses the same pattern as OpenSSL which uses find_library to only search in `AWSLC_INSTALL_DIR`.

### Call-outs:
Two of the existing benchmarks (trust token and hash to curve) require internal header files. These were already disabled when building with OpenSSL. Currently we don't care a ton about these two functions so I disabled them for external AWS-LC benchmark builds.

Also update run_benchmark_build_tests.sh to suppress some of OpenSSL's build output, previously it was 27,000 lines long and kills the CodeBuild output.

### Testing:
tests/ci/run_benchmark_build_tests.sh is run by https://github.com/awslabs/aws-lc/blob/main/tests/ci/codebuild/linux-x86/run_bm_build_test.yml as a part of the normal linux x86 builds. 

Before our CI included:
```
-- Building awslc_bm benchmark using header files from /codebuild/output/src747165175/src/github.com/awslabs/aws-lc/../libcrypto_install_dir/aws_lc/include and libcrypto from crypto.
135
```

After:
```
-- Building awslc_bm benchmark using header files from /codebuild/output/src925441200/src/github.com/awslabs/aws-lc/../libcrypto_install_dir/aws-lc-fips/include and libcrypto from /codebuild/output/src925441200/src/github.com/awslabs/libcrypto_install_dir/aws-lc-fips/lib/libcrypto.a.
```
the ISC license.

### DIFF
--- a/tests/ci/run_benchmark_build_tests.sh
+++ b/tests/ci/run_benchmark_build_tests.sh
@@ -15,7 +15,7 @@ openssl_3_0_branch='openssl-3.0.5'
 
 function build_aws_lc_fips {
   echo "building aws-lc in FIPS mode"
-  run_build -DCMAKE_INSTALL_PREFIX="${install_dir}/aws-lc-fips" -DFIPS=1
+  run_build -DCMAKE_INSTALL_PREFIX="${install_dir}/aws-lc-fips" -DFIPS=1 -DCMAKE_BUILD_TYPE=Release
   pushd "$BUILD_ROOT"
   ninja install
   popd
@@ -58,6 +58,12 @@ function build_openssl_3_0 {
 }
 
 # We build each tool individually so we can have more insight into what is failing
+
+# Building AWS-LC always builds bssl (which includes the speed tool) with the "local" libcrypto. We
+# also support building speed.cc with an "external" aws-lc libcrypto (and openssl). This is useful
+# when we want to compare the performance of a particular FIPS release against mainline if mainline
+# has changes in speed.cc that could affect the comparison of the FIPS to non-FIPS, or if new
+# algorithms have been added to speed.cc
 build_aws_lc_fips
 echo "Testing awslc_bm with AWS-LC FIPS"
 run_build -DAWSLC_INSTALL_DIR="${install_dir}/aws-lc-fips"
--- a/tests/ci/run_benchmark_build_tests.sh
+++ b/tests/ci/run_benchmark_build_tests.sh
@@ -13,47 +13,54 @@ openssl_1_0_branch='OpenSSL_1_0_2-stable'
 # failing our CI.
 openssl_3_0_branch='openssl-3.0.5'
 
+function build_aws_lc_fips {
+  echo "building aws-lc in FIPS mode"
+  run_build -DCMAKE_INSTALL_PREFIX="${install_dir}/aws-lc-fips" -DFIPS=1
+  pushd "$BUILD_ROOT"
+  ninja install
+  popd
+}
+
 function build_openssl_1_0 {
     echo "building OpenSSL 1.0"
-    git clone --branch "${openssl_1_0_branch}" "${openssl_url}" "../openssl-1.0"
+    git clone --depth 1 --branch "${openssl_1_0_branch}" "${openssl_url}" "../openssl-1.0"
     pushd "../openssl-1.0"
     mkdir -p "${install_dir}/openssl-1.0"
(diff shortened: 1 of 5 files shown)


tests/ci/run_benchmark_build_tests.sh
```diff
@@ -13,47 +13,54 @@ openssl_1_0_branch='OpenSSL_1_0_2-stable'
...
+    make install_sw
     popd
     rm -rf "../openssl-3.0"
 }
 
 # We build each tool individually so we can have more insight into what is failing
-mkdir -p "${install_dir}"
-echo "Testing awslc_bm"
-mkdir -p "${install_dir}/aws_lc"
-run_build -DAWSLC_INSTALL_DIR="${install_dir}/aws_lc"
```
DominicDams, line 56: We probably want to keep the test for aws-lc building in non-fips mode

andrewhop: I don't think this test was doing anything, it was just building the benchmark with the "local" aws-lc: there was nothing in `${install_dir}/aws_lc`. We already build the normal benchmark on every build. What I think the spirit of this test was supposed to be is:
1. build aws-lc's libcrypto somewhere somehow
2. copy that libcrypto
3. build the latest and greatest benchmark with that aws-lc libcrypto

DominicDams (dismissed):

tests/ci/run_benchmark_build_tests.sh
```diff
@@ -13,47 +13,54 @@ openssl_1_0_branch='OpenSSL_1_0_2-stable'
 # failing our CI.
 openssl_3_0_branch='openssl-3.0.5'
 
+function build_aws_lc_fips {
+  echo "building aws-lc in FIPS mode"
+  run_build -DCMAKE_INSTALL_PREFIX="${install_dir}/aws-lc-fips" -DFIPS=1
```
torben-hansen, line 18: FYI: This doesn't build the release build

tests/ci/run_benchmark_build_tests.sh
```diff
@@ -13,47 +13,54 @@ openssl_1_0_branch='OpenSSL_1_0_2-stable'
...
 }
 
 # We build each tool individually so we can have more insight into what is failing
-mkdir -p "${install_dir}"
-echo "Testing awslc_bm"
-mkdir -p "${install_dir}/aws_lc"
-run_build -DAWSLC_INSTALL_DIR="${install_dir}/aws_lc"
+build_aws_lc_fips
+echo "Testing awslc_bm with AWS-LC FIPS"
+run_build -DAWSLC_INSTALL_DIR="${install_dir}/aws-lc-fips"
```
torben-hansen, line 63: Why is this needed when we just did `build_aws_lc_fips` 2 lines above?

andrewhop: `build_aws_lc_fips` builds aws-lc and installs the public header files and libcrypto artifact into `${install_dir}/aws-lc-fips`, this line then builds aws-lc again and uses those public artifacts to make the `awslc_bm` executable. This is attempting to replicate building the fips branch with the latest speed.cc with any changes/additions to speed.cc.

torben-hansen: Please add a comment on what is going on here :)

tool/CMakeLists.txt
```diff
@@ -58,14 +59,15 @@ function(build_benchmark target_name additional_include_dir lib_crypto)
   # include directory
   target_include_directories(${target_name} PUBLIC ${additional_include_dir} ${AWSLC_INSTALL_DIR}/include/internal/tool)
   target_link_libraries(${target_name} ${lib_crypto} ${LIBRT_FLAG})
+  if(NOT MSVC AND NOT ANDROID)
+    target_link_libraries(${target_name} pthread dl)
+  endif()
+
 endfunction()
 
 if(AWSLC_INSTALL_DIR)
```
torben-hansen, line 68: these if-conditional stanza's are sorta similar. Can we simplify?

torben-hansen (approved):

DominicDams (approved):

PR #687, opened 2022-11-16
Rust - include generated-src in crate

justsmth:
### Issues:
N/A

### Description of changes: 
* Added `generated-src` to the crate, removing need for Perl and Golang.
* Removed `ssl` and other unnecessary files. Verified crate size is now ~5MB.
* Removed dependence on `cmake-rs` so that build options can be passed prior to build path.
* Updated `aws-lc-sys` crate version to `0.2.0`
* Added verification of backwards compatible public api to the `publish.sh` script.
* Set minimum supported rust version to 1.57.0.

### Call-outs:

### Testing:
* Ran the generate/publish scripts (but didn't publish).  Verified crate size if ~4.8MB:
```
➜ du -h tmp/aws-lc-sys/target/package/aws-lc-sys-0.2.0.crate 
4.8M	tmp/aws-lc-sys/target/package/aws-lc-sys-0.2.0.crate
```

the ISC license.

### DIFF
--- a/bindings/rust/aws-lc-sys-template/build.rs
+++ b/bindings/rust/aws-lc-sys-template/build.rs
@@ -245,7 +245,11 @@ fn load_env_var(var_name: &str) -> String {
 }
 
 fn build_aws_lc() -> Result<PathBuf, &'static str> {
-    let Some(cmake_cmd) = find_cmake_command() else {
+    let cmake_cmd: &OsStr;
+    if let Some(cmd) = find_cmake_command() {
+        // This can be a let-else as-of Rust 1.65
+        cmake_cmd = cmd;
+    } else {
         return Err("Missing dependency: cmake");
     };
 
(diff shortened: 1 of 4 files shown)


bindings/rust/aws-lc-sys-template/build.rs
```diff
@@ -210,59 +196,115 @@ fn find_cmake_command() -> Option<&'static OsStr> {
...
+
+    Ok(out_dir)
 }
 
 fn main() -> Result<(), String> {
     use crate::OutputLib::Crypto;
     use crate::OutputLibType::Static;
 
     let mut missing_dependency = false;
-    if !test_go_command() {
```
samuel40791765, line 257: just at datapoint: Go is still required when building for FIPS, so we might need to add this back for the FIPS build. But we can remove it for now

samuel40791765 (approved):

skmcgrail (approved):

PR #690, opened 2022-11-21
KBKDF in Feedback Mode for FIPS boundary

Taffer:
### Issues:
Resolves CryptoAlg-1343
Addresses CryptoAlg-1198

### Description of changes: 
The HKDF_expand() function is *almost* what FIPS wants for a KBKDF in Feedback Mode; this is the real deal, suitable for FIPS 140-3 certification. The service indicator documentation has also been updated.

### Call-outs:
As documented in the code, I *think* this deviates slightly from the NIST SP800-108r1 specification, but it works as-is with the NIST-supplied test vectors. This disturbs me.

### Testing:
Unit tests using NIST's [CAVP](https://csrc.nist.gov/Projects/cryptographic-algorithm-validation-program/key-derivation) test vectors. Unit tests for input sanity. Service indicator test for good hash vs bad hash.


crypto/fipsmodule/kbkdf/kbkdf.c
```diff
@@ -0,0 +1,178 @@
...
+
+#include <openssl/hmac.h>
+#include <openssl/kbkdf.h>
+
+#include "../../internal.h"
+#include "../service_indicator/internal.h"
+
+// Convert a 32-bit little-endian value into a 32-bit big-endian value.
+// AWS-LC only supports little-endian hosts.
+static uint32_t to_be32(uint32_t host)
```
torben-hansen, line 17: Endian conversion functions: https://github.com/awslabs/aws-lc/blob/main/crypto/internal.h#L715

crypto/fipsmodule/kbkdf/kbkdf.c
```diff
@@ -0,0 +1,178 @@
...
+
+    value |= (host & 0xff000000) >> 24;
+    value |= (host & 0x00ff0000) >> 8;
+    value |= (host & 0x0000ff00) << 8;
+    value |= (host & 0x000000ff) << 24;
+
+    return value;
+}
+
+// Surprised min32/min64/etc. didn't end up in stdint.h...
```
torben-hansen, line 29: Redundant comment?

crypto/fipsmodule/kbkdf/kbkdf.c
```diff
@@ -0,0 +1,178 @@
...
+        out_len < 1 || key_in_len < 1 ||
+        (key_in == NULL && key_in_len > 0) ||
+        (label == NULL && label_len > 0) ||
+        (context == NULL && context_len > 0) ||
+        (iv == NULL && iv_len > 0)) {
+        return 0;
+    }
+
+    int retval = 0;
+    uint8_t *ki = NULL;
```
torben-hansen, line 67: More descriptive name maybe, or you following the spec?

torben-hansen (commented):
Quick glance for the first iteration.

crypto/fipsmodule/kbkdf/kbkdf.c
Taffer: Following the spec, I'll add a comment about the terrible variable names.

crypto/fipsmodule/kbkdf/kbkdf.c
Taffer: I'm still surprised. Suppose nobody cares about my editorial comments though...

crypto/fipsmodule/kbkdf/kbkdf.c
```diff
@@ -0,0 +1,169 @@
...
+        //
+        // ki = PRF(key_in, previous ki || i || Label || 0x00 || Context || L)
+        //
+        // The test data from the CAVP Testing: SP 800-108 Key Derivation Using
+        // Pseudorandom Functions - Key-Based (KBKDF) page
+        // https://csrc.nist.gov/Projects/cryptographic-algorithm-validation-program/key-derivation
+        // seems to require us to ignore L.
+        //
+        // The spec also doesn't indicate whether L should be appended in
+        // big-endian format, or how large (in bits) it should be.
```
andrewhop, line 129: Can we ask our lab about this? Are you sure you have the right test vectors?

crypto/fipsmodule/kbkdf/kbkdf.c
```diff
@@ -0,0 +1,169 @@
...
+        }
+        */
+
+        ki_len = hmac_size;
+        if (!HMAC_Final(&hmac, ki, (unsigned int *)&ki_len)) {
+            goto out;
+        }
+
+        size_t this_write = ct_min(ki_len, to_write);
+        memcpy(out_key + written, ki, this_write);
```
andrewhop, line 142: For safety and consistency: 
```suggestion
        OPENSSL_memcpy(out_key + written, ki, this_write);
```

crypto/fipsmodule/kbkdf/kbkdf.c
```diff
@@ -0,0 +1,169 @@
...
+    uint8_t zero = 0x00;
+
+    size_t n = 1 + ((L - 1) / h);  // ceil(L/h)
+
+    if (n > UINT32_MAX) {
+        // SP 800-108r1 section 4.0, n <= 2^r - 1
+        goto out;
+    }
+
+    ki = OPENSSL_malloc(ct_max(iv_len, hmac_size));
```
andrewhop, line 85: For consistency and safety:
```suggestion
    ki = OPENSSL_malloc(ct_max(iv_len, hmac_size));
    if (ki == NULL) {
        OPENSSL_PUT_ERROR(CRYPTO, ERR_R_MALLOC_FAILURE);
        goto out;
  }
```

crypto/fipsmodule/kbkdf/kbkdf.c
```diff
@@ -0,0 +1,169 @@
...
+    size_t ki_len = key_in_len;
+    if (iv != NULL && iv_len > 0) {
+        // Set k(0) to the IV.
+        memcpy(ki, iv, iv_len);
```
andrewhop, line 89:
```suggestion
        OPENSSL_memcpy(ki, iv, iv_len);
```

crypto/fipsmodule/kbkdf/kbkdf.c
andrewhop, line 142: Is there a safer way to write this? I'm convinced `written` will always be less than `out_len` but this is a little tricky accounting.

crypto/fipsmodule/kbkdf/kbkdf.c
```diff
@@ -0,0 +1,169 @@
...
+out:
+    FIPS_service_indicator_unlock_state();
+    if (retval) {
+        KBKDF_verify_service_indicator(digest);
+    }
+
+    if (ki != NULL) {
+        OPENSSL_cleanse(ki, ct_max(iv_len, hmac_size));
+        OPENSSL_free(ki);
+    }
```
andrewhop, line 165: OPENSSL_free always cleanses memory before freeing it and handles null.
```suggestion
    OPENSSL_free(ki);
```

crypto/fipsmodule/kbkdf/kbkdf.c
```diff
@@ -0,0 +1,169 @@
...
+    HMAC_CTX_cleanse(&hmac);
+
+    return retval;
```
andrewhop, line 168: In the case of the above loop failing part way through should we wipe any partial output in `out_key`? I don't think it's needed like in the case of partial decryption before verifying a tag, just wondering if you thought about it.

crypto/fipsmodule/kbkdf/kbkdf_test.cc
```diff
@@ -0,0 +1,641 @@
...
+
+#include <openssl/digest.h>
+#include <openssl/err.h>
+#include <openssl/kbkdf.h>
+
+#include <gtest/gtest.h>
+#include "../../test/test_util.h"
+
+// Test vectors from NIST CAVP for SP 800-108
+// https://csrc.nist.gov/Projects/cryptographic-algorithm-validation-program/key-derivation
```
andrewhop, line 12: Overall A+ testing here, I like the paramaetrized tests and good coverage. However, I'd like to simplify these a bit: our `FileTest` can parse and read the files as is from NIST. This makes it easier to review (make sure there were no types or conversion errors) and makes this test file smaller. `aead_test.cc` is a good example of testing different parameters with different files/algorithms.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -2961,9 +2962,264 @@ TEST_P(PBKDF2_ServiceIndicatorTest, PBKDF2) {
...
+  delete[] output;
+}
+
+INSTANTIATE_TEST_SUITE_P(KBKDF_ServiceIndicatorTest, KBKDFTestSI,
+                         testing::ValuesIn(nonzero_iv_vectors));
+
 // SSHKDF test vector from NIST CAVS 14.1 test vectors
 // https://csrc.nist.gov/Projects/Cryptographic-Algorithm-Validation-Program/Component-Testing#KDF135
-static const char kSSHKDFtype = EVP_KDF_SSHKDF_TYPE_INITIAL_IV_CLI_TO_SRV;
+const char kSSHKDFtype = EVP_KDF_SSHKDF_TYPE_INITIAL_IV_CLI_TO_SRV;
```
andrewhop, line 3222: Any reason for all these static changes?

crypto/fipsmodule/kbkdf/kbkdf.c
```diff
@@ -0,0 +1,169 @@
...
+//
+// The |use_counter| option enables the "KDF in Feedback Mode with Counter"
+// mode.
+OPENSSL_EXPORT int KBKDF_feedback(uint8_t *out_key, size_t out_len,
+                                  const EVP_MD *digest,
+                                  const uint8_t *key_in, size_t key_in_len,
+                                  const uint8_t *label, size_t label_len,
+                                  const uint8_t *context, size_t context_len,
+                                  const uint8_t *iv, size_t iv_len,
+                                  bool use_counter)
```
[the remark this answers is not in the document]
Taffer: `stdbool.h` is a C99 feature, and even the ancient platforms in our CI passed; do we still support anything that doesn't support C99?

`digest.h` also includes `stdbool.h`, I searched before I used this. ;-)

crypto/fipsmodule/kbkdf/kbkdf.c
Taffer: Positive they're the right test vectors, straight from NIST. I'll ask the lab for confirmation though.

crypto/fipsmodule/kbkdf/kbkdf.c
Taffer: Added an `if` to double-check that we're in range for the `out_len`.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
Taffer: Just making them consistent with the other constants in the file.

crypto/fipsmodule/kbkdf/kbkdf_test.cc
Taffer: I'll look into that, thanks. Without knowing anything about it yet, I'd be concerned that the test vectors from the CAVP page have way too much data in them... if `FileTest` can filter to only run certain vectors that'll be fine, if not, the majority of the CAVP data isn't going to test properly. The spec for KBKDF is stupidly flexible for some reason.

Taffer: If it's OK with you, I'll do this as a separate task, and also create one to adapt the SSHKDF tests to use this... they also have CAVP test data avilable but don't currently use it.

Taffer: CryptoAlg-1436 created for KBKDF

crypto/fipsmodule/kbkdf/kbkdf.c
Taffer: I did not, I was assuming the user would discard the partial data rather than use it as a key when the function returns an error. Everyone always checks their return values, right?

PR #695, opened 2022-11-23
Fips 2022 macos signing

Taffer:
### Issues:
CryptoAlg-1426

### Description of changes: 

On macOS 11 and higher on Apple Silicon, codesigning is mandatory for
binaries to run. This applies to both executables and dylibs. An ad-hoc
signature is sufficient, and the linker will automatically apply one when
a binary is created (see https://github.com/Homebrew/brew/issues/9082).

When we build libcrypto.dylib the linker automatically signs it. But then
we inject the FIPS integrity hash into libcrypto.dylib which changes the
binary so the signature applied by the linker is obviously not valid
anymore. So when an application, like crypto_test, tries to load
libcrypto.dylib it crashes because the signature is not valid. To work
round this we add an ad-hoc signature to `libcrypto.dylib` after the
FIPS integrity hash is injected.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and  the ISC license.

### DIFF
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -583,6 +583,27 @@ if(FIPS_SHARED)
       DEPENDS ../util/fipstools/inject_hash/inject_hash.go
       WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
     )
+
+    # On macOS 11 and higher on Apple Silicon, codesigning is mandatory for
+    # binaries to run. This applies to both executables and dylibs. An ad-hoc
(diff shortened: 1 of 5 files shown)


Taffer:
Not sure why it's including `9c2b0b9`, `573620e`, and `c5b4987` in here, they're already present in the `fips-2022-11-02` branch (as you can tell from the changed files in this PR). I need more `git` learning!

dkostic (approved):

samuel40791765 (approved):

PR #696, opened 2022-11-24
Create a variable per benchmark built to avoid CMake caching the result of find_library

andrewhop:
### Description of changes: 
In https://github.com/awslabs/aws-lc/pull/686 we changed the tool CMakeList to find the libcrypto library as a part of the build_benchmark function. However, by default CMake caches the result of the find library call and if multiple libcryptos are passed in only the first is used. To get around that create a new variable per libcrypto so CMake doesn't cache the result, CMake 3.21 add a [NO_CACHE](https://cmake.org/cmake/help/latest/command/find_library.html) option to avoid this behavior that we can use in the future.

Also specify the [message as status so it gets printed to stdout](https://cmake.org/cmake/help/latest/command/message.html) and captured by the logs.

Fix build_run_benchmarks.sh so it uses the "locally" built `bssl` executable that gets built as a part of the overall AWS-LC build. There is no need to build the local code a second time to make a targeted awslc_bm.

### Call-outs:
This wasn't caught in the `run_benchmark_build_tests.sh` integ test because it only builds one benchmark at a time instead of multiple.

### Testing:
Canary is is happy now:
```
-- Building awslc_bm benchmark using header files from /home/ec2-user/canary/install-AWS-LC-FIPS-2022-11-02/include and libcrypto from /home/ec2-user/canary/install-AWS-LC-FIPS-2022-11-02/lib64/libcrypto.a.
-- Building ossl_1_1_bm benchmark using header files from /home/ec2-user/canary/install-OpenSSL-1-1-1/include and libcrypto from /home/ec2-user/canary/install-OpenSSL-1-1-1/lib/libcrypto.so.
-- Building ossl_3_0_bm benchmark using header files from /home/ec2-user/canary/install-OpenSSL-3-0/include and libcrypto from /home/ec2-user/canary/install-OpenSSL-3-0/lib/libcrypto.so.

```

the ISC license.

### DIFF
--- a/tests/ci/build_run_benchmarks.sh
+++ b/tests/ci/build_run_benchmarks.sh
@@ -25,29 +25,23 @@ git clone https://github.com/awslabs/aws-lc.git aws-lc-prod
 
 # build AWSLC pr
 mkdir -p "${PR_FOLDER_NAME}"/build
-mkdir -p "${PR_FOLDER_NAME}"/install
 ${CMAKE_COMMAND} -B"${PR_FOLDER_NAME}"/build -H"${PR_FOLDER_NAME}" -GNinja -DCMAKE_BUILD_TYPE=Release \
-  -DAWSLC_INSTALL_DIR="${AWSLC_PR_ROOT}"/install \
-  -DCMAKE_INSTALL_PREFIX="${AWSLC_PR_ROOT}"/install \
   -DBUILD_TESTING=OFF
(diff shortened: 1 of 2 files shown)


dkostic (dismissed):

torben-hansen (commented):
Looking at the log for the BM dimension, it seems to be broken even with this change?

torben-hansen (approved):

dkostic (approved):

PR #697, opened 2022-11-26
Fix the code signing issue on M1 Macs when NO_ASM=1

dkostic:
### Issues:
N/A

### Description of changes:

PR #693 added a `codesign` post-build command to be executed when
building the FIPS module on an M1 mac. The command was done only
if `ARCH` is "aarch64". However, when `NO_ASM` build flag is defined
the `ARCH` variable is set to "generic" instead and the `codesign`
command is not called. This commit fixes the issue.

### Call-outs:

### Testing:

the ISC license.


torben-hansen (approved):

Taffer (approved):

Taffer:
We'll want to pull this one into both FIPS branches.

PR #703, opened 2022-11-30
RSA_check_fips param _should_ be const

justsmth:
### Issues:
N/A

### Description of changes: 
The `RSA *key` parameter to `RSA_check_fips` should be `const`.

### Call-outs:
N/A

### Testing:
N/A

the ISC license.

### DIFF
--- a/crypto/fipsmodule/rsa/rsa.c
+++ b/crypto/fipsmodule/rsa/rsa.c
@@ -966,7 +966,7 @@ int RSA_check_fips(const RSA *key) {
     return 0;
   }
 
-  if (!RSA_sign(NID_sha256, data, sizeof(data), sig, &sig_len, key)) {
+  if (!RSA_sign(NID_sha256, data, sizeof(data), sig, &sig_len, (RSA*) key)) {
     OPENSSL_PUT_ERROR(RSA, ERR_R_INTERNAL_ERROR);
     ret = 0;
     goto cleanup;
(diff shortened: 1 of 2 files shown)


crypto/fipsmodule/rsa/rsa.c
```diff
@@ -902,7 +902,7 @@ DEFINE_LOCAL_DATA(BIGNUM, g_small_factors) {
   out->flags = BN_FLG_STATIC_DATA;
 }
 
-int RSA_check_fips(RSA *key) {
+int RSA_check_fips(const RSA *key) {
```
torben-hansen, line 905: But `key` is an argument to `RSA_sign()` which would then drop const'ness? https://github.com/awslabs/aws-lc/blob/main/crypto/fipsmodule/rsa/rsa.c#L969

justsmth: The documentation for `RSA_check_fips` claims:
> This function does not mutate |rsa| for thread-safety purposes and may be used concurrently.

Should the parameter to `RSA_sign` also be `const RSA*`?  Or does `RSA_sign` modify the parameter?

torben-hansen: I'm just speaking from a function signature compatibility perspective. The signature doesn't match.

These functions doesn't mutate the `key` and you should be able to change them. But that would likely make a lot of applications angry.

include/openssl/rsa.h
```diff
@@ -473,7 +473,7 @@ OPENSSL_EXPORT int RSA_check_key(const RSA *rsa);
 // RSA_check_fips performs public key validity tests on |key|. It returns one if
 // they pass and zero otherwise. Opaque keys always fail. This function does not
 // mutate |rsa| for thread-safety purposes and may be used concurrently.
-OPENSSL_EXPORT int RSA_check_fips(RSA *key);
+OPENSSL_EXPORT int RSA_check_fips(const RSA *key);
```
justsmth, line 476: Based upon the assertion made in the documentation here -- "_This function does not mutate |rsa| and for thread-safety purposes may be used concurrently_" -- shouldn't the parameter be `const`?  (If our implementation calls a mutating function, then we should remove this claim from the documentation.)

crypto/fipsmodule/rsa/rsa.c
justsmth: Ok. So for historical reasons, the function signature allows for mutation of the parameter even though the implementation currently doesn't.

PR #705, opened 2022-11-30
Deduplicate Kyber files

dkostic:
### Issues:
N/A

### Description of changes: 
The commit removes the file duplication issue with Kyber implementation.
Previously, Kyber512 and Kyber768 implementations resided in separate
folders, both with a copy of all the source files. In this change we:
 - create a single _common_ folder with Kyber source code files,
 - add `kyber512_ref.c` file that includes the required C files
   from the common folder and selects the Kyber512 implementation
   by defining a macro (`KYBER_K 2`) specified in Kyber as the selector
   of the variant to be compiled.

In this way, when we add Kyber768 support we will only have to add the
corresponding `kyber768_ref.c` file, which will avoid duplicating the
Kyber source code files.

### Call-outs:

### Testing:

the ISC license.


crypto/kyber/kyber512_ref.c
```diff
@@ -0,0 +1,31 @@
...
+#include "./pqcrystals_kyber_ref_common/cbd.h"
+#include "./pqcrystals_kyber_ref_common/fips202.h"
+#include "./pqcrystals_kyber_ref_common/indcpa.h"
+#include "./pqcrystals_kyber_ref_common/kem.h"
+#include "./pqcrystals_kyber_ref_common/ntt.h"
+#include "./pqcrystals_kyber_ref_common/poly.h"
+#include "./pqcrystals_kyber_ref_common/polyvec.h"
+#include "./pqcrystals_kyber_ref_common/reduce.h"
+#include "./pqcrystals_kyber_ref_common/symmetric.h"
+#include "./pqcrystals_kyber_ref_common/verify.h"
```
andrewhop, line 20: Do we need to include all the header files? Won't each c file include what it needs?

crypto/kyber/kem_kyber.h
```diff
@@ -7,7 +7,7 @@
 #include <openssl/evp.h>
```
andrewhop, line 7: Missing copyright, is this our file or kyber's originally?

andrewhop (commented):
Also missing a copyright in p_kyber.c, p_kyber512_asn1.c

crypto/kyber/kyber512_ref.c
dkostic: you are right, I removed the unnecessary includes.

crypto/kyber/kem_kyber.h
dkostic: added. it's our file

andrewhop (dismissed):

crypto/kyber/kyber512_ref.c
```diff
@@ -0,0 +1,23 @@
+/*
```
torben-hansen, line 1: okay, this does avoid dynamic and/or over-allocating stack memory etc in each function. But still "funny" macrofied.

torben-hansen (dismissed):

torben-hansen (approved):

andrewhop (approved):

PR #707, opened 2022-11-30
Enable host name and email length detection with strlen

andrewhop:
### Issues:
Resolves CryptoAlg-1285

### Description of changes: 
All internal usage of `X509_check_host`, `X509_check_email` and `X509_check_ip` pass in the correct length of the data to check. These three functions all call the common `do_x509_check` to preform that check. The logic in `do_x509_check` to determine the length was removed in https://github.com/awslabs/aws-lc/commit/c64ccb51b0daa44abe715c00862fa16ade79e14b since it is considered an error for BoringSSL. AWS-LC wants to maintain the OpenSSL behavior and this PR adds it back to  `X509_check_host` and `X509_check_email`.

### Call-outs:
I decided not to put the logic in `do_x509_check` like it was originally because it is an error to call strlen on a byte array that is not null terminated such as the binary IP address from `X509_check_ip`. This is in line with the refactor in https://github.com/awslabs/aws-lc/commit/1902d818ac4fef9497dfe5d0ce6f2c99f585bdff wanting to avoid calling strlen on the binary IP address byte array.
### Testing:
Updated existing unit tests and added a new test for the `peername` behavior that didn't seem to be covered directly before.

the ISC license.


Taffer (dismissed):

crypto/x509/x509_test.cc
```diff
@@ -1425,21 +1425,39 @@ TEST(X509Test, ZeroLengthsWithCheckFunctions) {
...
+  EXPECT_NE(1, X509_check_host(leaf.get(), kHostname, strlen(kHostname) - 1, 0, nullptr));
+
+  EXPECT_EQ(-2, X509_check_email(leaf.get(), kEmail, strlen(kEmail) + 1, 0));
+  EXPECT_NE(1, X509_check_email(leaf.get(), kEmail, strlen(kEmail) - 1, 0));
+}
+
+TEST(X509Test, MatchFoundSetsPeername) {
+  bssl::UniquePtr<X509> leaf(CertFromPEM(kSANTypesLeaf));
+  char *peername;
+  EXPECT_EQ(1, X509_check_host(leaf.get(), kHostname, strlen(kHostname), 0, &peername));
```
torben-hansen, line 1457: `peername` is allocated in `X509_check_host` I think, so must be free'd before returning.

torben-hansen (approved):

Taffer (approved):

PR #708, opened 2022-12-01
Key Encapsulation Mechanism (KEM) API

dkostic:
### Issues:
N/A

### Description of changes: 
Add support for a new primitive: Key Encapsulation Mechanism (KEM).

The primary use-case for a KEM are the post-quantum crypto algorithms
specified and evaluated in the NIST PQC Standardization project.

To support KEM operations, the following public APIs are implemented:
  - EVP_PKEY_KEM - new PKEY identifier for KEMs.
  - EVP_PKEY_keygen - key generation (already exist),
  - EVP_PKEY_encapsulate - encapsulation,
  - EVP_PKEY_decapsulate - decapsulation,
  - EVP_PKEY_CTX_kem_set_params - set specific KEM parameters.

Additionaly, creating a key from raw data and extracting raw data from
a key can be done with:
  - EVP_PKEY_kem_new_raw_key - new PKEY from (public, secret) raw data,
  - EVP_PKEY_kem_new_raw_public_key - new PKEY from public key raw dat,
  - EVP_PKEY_kem_new_raw_secret_key - new PKEY from secret key raw dat,
  - EVP_PKEY_get_raw_public_key - get raw public key from PKEY,
  - EVP_PKEY_get_raw_private_key - get raw private key from PKEY.

Internally, we define the following new structures:
  - KEM_METHOD - table of function pointers defined for a KEM,
  - KEM - holds information for a specific KEM,
  - built_in_kems - list of supported built-in KEMs,
  - KEM_KEY - PKEY for KEM,

Unit and KAT tests are added to crypto/evp_extra/evp_extra_test.cc.

### Call-outs:
Only a draft for now, full test suite is still missing.

### Testing:

the ISC license.


include/openssl/evp.h
```diff
@@ -1121,6 +1152,7 @@ struct evp_pkey_st {
     DSA *dsa;
     DH *dh;
     EC_KEY *ec;
+    KEM_KEY *kem;
```
torben-hansen, line 1155: I vote for using void pointer.

crypto/kem/internal.h
```diff
@@ -0,0 +1,70 @@
...
+#endif
+
+// KEM_METHOD structure and helper functions.
+typedef struct {
+  int (*keygen)(uint8_t *public_key,
+                uint8_t *secret_key);
+
+  int (*encaps)(uint8_t *ciphertext,
+                uint8_t *shared_secret,
+                const uint8_t *public_key);
```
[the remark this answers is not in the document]
andrewhop: Are we sure all callers of these functions are doing the length checks and that we won't miss one in the future?

andrewhop (commented):
You need to update the Kyber Readme example code with the new KEM API changes.

torben-hansen (commented):
Passed through the kem_pkey_meth functions and that layer.

include/openssl/nid.h
```diff
@@ -4286,6 +4286,10 @@ extern "C" {
 #define SN_KYBER512 "KYBER512"
 #define NID_KYBER512 970
 
+#define SN_kem "KEM"
+#define LN_kem "kem"
+#define NID_kem 972
```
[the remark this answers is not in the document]
dkostic: made a guess in a comment above. Anyway, it's "fixed" now.

crypto/evp_extra/p_kem.c
```diff
@@ -0,0 +1,233 @@
...
+                                uint8_t *shared_secret,
+                                size_t  *shared_secret_len) {
+  KEM_PKEY_CTX *dctx = ctx->data;
+  const KEM *kem = dctx->kem;
+  if (kem == NULL) {
+    if (ctx->pkey == NULL) {
+      OPENSSL_PUT_ERROR(EVP, EVP_R_NO_PARAMETERS_SET);
+      return 0;
+    }
+    kem = KEM_KEY_get0_kem((KEM_KEY*)ctx->pkey->pkey.ptr);
```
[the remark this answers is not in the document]
andrewhop: Is 2 a valid state to end up in? Can you call pkey_kem_keygen if there is no specific kem configured?

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -1695,3 +1699,434 @@ TEST_P(EVPRsaPssBadKeyTest, InvalidSaltLength) {
...
+  // ---- 4. Test calling encapsulate with different lengths ----
+  // Set ct length to be less than expected -- should fail.
+  ct_len = GetParam().ciphertext_len - 1;
+  ASSERT_FALSE(EVP_PKEY_encapsulate(ctx.get(), ct.data(), &ct_len, ss.data(), &ss_len));
+
+  // Set ct length to be greater than expected -- should succeed because
+  // it's ok to provide buffer that's larger than needed.
+  ct_len = GetParam().ciphertext_len + 1;
+  ASSERT_TRUE(EVP_PKEY_encapsulate(ctx.get(), ct.data(), &ct_len, ss.data(), &ss_len));
+
```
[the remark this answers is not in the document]
andrewhop: This test is narrowly avoid a memory corruption bug:
```
  std::vector<uint8_t> ct(ct_len);
  ct_len = GetParam().ciphertext_len + 1;
  ASSERT_TRUE(EVP_PKEY_encapsulate(ctx.get(), ct.data(), &ct_len, ss.data(), &ss_len));
```
The buffer is still only `ct_len` long but you're telling EVP_PKEY_encapsulate it can use up to `ct_len + 1` which we don't and address sanitizer is ensuring that.

crypto/evp_extra/p_kem.c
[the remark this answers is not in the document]
dkostic: It is a valid state because the KEM can be configured in the PKEY object associated with the context. For example when creating a new raw key (see `RawKeyOperations` test in `evp_extra_tests.cc` for reference).
```
EVP_PKEY *pkey = EVP_PKEY_kem_new_raw_public_key(nid, pk, pk_len);
EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new(pkey, NULL);

// you can use the context now for keygen/encaps/decaps.
```

BUT, to get back to your original question:
> Should this also assign it to ctx->data to avoid having to call get0 every time?

I just realized that it does actually assign it :)
```
const KEM *kem = ctx->data->kem;
  if (kem == NULL) {
    // ...
    kem = KEM_KEY_get0_kem(ctx->pkey->pkey.kem_key);
  }
```

crypto/kem/kem.c
```diff
@@ -0,0 +1,133 @@
...
+void KEM_KEY_free(KEM_KEY *key) {
+  if (key == NULL) {
+    return;
+  }
+  key->kem = NULL;
+  OPENSSL_free(key->public_key);
+  OPENSSL_free(key->secret_key);
+}
+
+const KEM *KEM_KEY_get0_kem(KEM_KEY* key) {
```
torben-hansen, line 96: Why call it `get0`? :D

crypto/evp_extra/p_kem_asn1.c
```diff
@@ -0,0 +1,148 @@
...
+  return kem->public_key_len + kem->secret_key_len;
+}
+
+static int kem_bits(const EVP_PKEY *pkey) {
+  return 8 * kem_size(pkey);
+}
+
+const EVP_PKEY_ASN1_METHOD kem_asn1_meth = {
+  EVP_PKEY_KEM,
+  // TODO(awslc): this is a placeholder OID. Do we need OID for KEM at all?
```
torben-hansen, line 129: Well, here lies a danger: if the different KEM instantiations implements something that is not compatible with each other e.g. RSA padding.

crypto/evp_extra/p_kem_asn1.c
```diff
@@ -0,0 +1,148 @@
...
+static int kem_pub_cmp(const EVP_PKEY *a, const EVP_PKEY *b) {
+  int ret;
+  ret = kem_cmp_parameters(a, b);
+  if (ret <= 0) {
+    return ret;
+  }
+
+  const KEM_KEY *a_key = a->pkey.kem_key;
+  const KEM_KEY *b_key = b->pkey.kem_key;
+  return OPENSSL_memcmp(a_key->public_key, b_key->public_key,
```
torben-hansen, line 103: check `a_key->kem->public_key_len = b_key->kem->public_key_len` for good measure before indexing?

crypto/evp_extra/p_kem_asn1.c
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  KEM_KEY *key = pkey->pkey.kem_key;
+  const KEM *kem = key->kem;
+  if (kem == NULL) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_NO_PARAMETERS_SET);
+    return 0;
+  }
+
+  return kem->public_key_len + kem->secret_key_len;
```
torben-hansen, line 120: Why the sum? Not that I think it is wrong, just don't know.

torben-hansen (commented):
Still need to review tests.

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -1695,3 +1699,512 @@ TEST_P(EVPRsaPssBadKeyTest, InvalidSaltLength) {
...
+  // Generate a key pair.
+  EVP_PKEY *raw = nullptr;
+  ASSERT_TRUE(EVP_PKEY_keygen_init(ctx.get()));
+  ASSERT_TRUE(EVP_PKEY_keygen(ctx.get(), &raw));
+  ASSERT_TRUE(raw);
+  bssl::UniquePtr<EVP_PKEY> pkey(raw);
+
+  // ---- 2. Test key generation with PKEY as a template ----
+  ctx.reset(EVP_PKEY_CTX_new(pkey.get(), nullptr));
```
torben-hansen, line 1745: Sorta annoying having to re-initialise a new object...

crypto/kem/kem.c
```diff
@@ -0,0 +1,133 @@
...
+// TODO(awslc): placeholder OID, replace with the real one when available.
+static const uint8_t kOIDKyber512[] = {0xff, 0xff, 0xff, 0xff};
+
+static const KEM built_in_kems[AWSLC_NUM_BUILT_IN_KEMS] = {
+  {
+    NID_KYBER512,         // kem.nid
+    kOIDKyber512,         // kem.oid
+    sizeof(kOIDKyber512), // kem.oid_len
+    "Kyber512",           // kem.comment
+    800,                  // kem.public_key_len
```
torben-hansen, line 24: Should also reference where these come from.

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -1695,3 +1699,512 @@ TEST_P(EVPRsaPssBadKeyTest, InvalidSaltLength) {
...
+  ASSERT_TRUE(ctx);
+
+  // Generate a key pair.
+  raw = nullptr;
+  ASSERT_TRUE(EVP_PKEY_keygen_init(ctx.get()));
+  ASSERT_TRUE(EVP_PKEY_keygen(ctx.get(), &raw));
+  ASSERT_TRUE(raw);
+  pkey.reset(raw);
```
torben-hansen, line 1753: why reset? Should be unnecessary now at this point.

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -1695,3 +1699,512 @@ TEST_P(EVPRsaPssBadKeyTest, InvalidSaltLength) {
...
+
+#define CMP_VEC_AND_PKEY_SECRET(vec, pkey, len) \
+          CMP_VEC_AND_PTR(vec, pkey->pkey.kem_key->secret_key, len)
+
+TEST_P(PerKEMTest, RawKeyOperations) {
+  
+  // ---- 1. Setup phase: generate a context and a key ----
+  // Create context of KEM type.
+  bssl::UniquePtr<EVP_PKEY_CTX> ctx(EVP_PKEY_CTX_new_id(EVP_PKEY_KEM, nullptr));
+  EXPECT_TRUE(ctx);
```
torben-hansen, line 1979: Use assert's here, since it doesn't make sense to continue if `ctx` is `NULL`.

torben-hansen: Few of these here.

crypto/kem/kem.c
dkostic: done

crypto/kem/kem.c
dkostic: From `evp.h`:
```
// The |get1| functions return a fresh reference to the
// underlying object or NULL if |pkey| is not of the correct type. The |get0|
// functions behave the same but return a non-owning pointer.
```
I assumed this is a general rule for the library. But maybe it's not relevant in our case?

crypto/kem/kem.c
```diff
@@ -0,0 +1,133 @@
...
+  OPENSSL_free(key->public_key);
+  OPENSSL_free(key->secret_key);
+}
+
+const KEM *KEM_KEY_get0_kem(KEM_KEY* key) {
+  return key->kem;
+}
+
+int KEM_KEY_set_raw_public_key(KEM_KEY *key, const uint8_t *in) {
+  key->public_key = OPENSSL_memdup(in, key->kem->public_key_len);
```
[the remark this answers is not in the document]
dkostic: I think this is solved now based on our offline discussion.

crypto/kem/kem.c
```diff
@@ -0,0 +1,133 @@
...
+  return 1;
+}
+
+int KEM_KEY_set_raw_key(KEM_KEY *key, const uint8_t *in_public,
+                                      const uint8_t *in_secret) {
+  key->public_key = OPENSSL_memdup(in_public, key->kem->public_key_len);
+  key->secret_key = OPENSSL_memdup(in_secret, key->kem->secret_key_len);
+  if (key->public_key == NULL || key->secret_key == NULL) {
+    OPENSSL_PUT_ERROR(EVP, ERR_R_MALLOC_FAILURE);
+    KEM_KEY_free(key);
```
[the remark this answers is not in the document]
dkostic: I think this is solved now based on our offline discussion.

include/openssl/evp.h
```diff
@@ -866,7 +869,36 @@ OPENSSL_EXPORT int EVP_PKEY_CTX_get0_rsa_oaep_label(EVP_PKEY_CTX *ctx,
 OPENSSL_EXPORT int EVP_PKEY_CTX_set_ec_paramgen_curve_nid(EVP_PKEY_CTX *ctx,
                                                           int nid);
 
-
+// KEM specific functions.
+
+// EVP_PKEY_CTX_kem_set_params sets in |ctx| the parameters associated with
+// KEM defined by the given |nid|. It returns one on success and zero on error.
+OPENSSL_EXPORT int EVP_PKEY_CTX_kem_set_params(EVP_PKEY_CTX *ctx, int nid);
```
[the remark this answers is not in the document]
dkostic: not anymore. Now, it's not allowed to call the function with `ctx` that has an associated `pkey`. Do you think we should specify that in the comment or leave it out?

crypto/evp_extra/evp_extra_test.cc
dkostic: added back. The `raw` pointer is a regular C pointer so we have to free in the end, otherwise the test leaks the allocated memory. Instead of freeing directly, I just added back the `reset` -- the `UniquePtr` class takes care of properly freeing the object.

andrewhop (approved):

crypto/evp_extra/p_kem.c
```diff
@@ -0,0 +1,339 @@
...
+
+  return 1;
+}
+
+static void pkey_kem_cleanup(EVP_PKEY_CTX *ctx) {
+  OPENSSL_free(ctx->data);
+}
+
+static int pkey_kem_keygen(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey) {
+  KEM_PKEY_CTX *dctx = ctx->data;
```
darylmartin100, line 37: This variable doesn't server a purpose.  I would just replace the next line with `const KEM *kem = ctx->data->kem`

crypto/evp_extra/p_kem_asn1.c
torben-hansen: Yeah, that's right. We have used `EVP_PKEY_*` values for different versions that needs incompatible encoding/decoding (e.g. RSA padding). We now need another method to branch in the underlying code. As long as we can do that.

crypto/evp_extra/evp_extra_test.cc
torben-hansen: That's what I am annoyed about :D

torben-hansen (approved):

PR #714, opened 2022-12-07
Add crate generation for aws-lc-fips-sys

samuel40791765:
### Issues:
Resolves `CryptoAlg-1398`

### Description of changes: 
* Took out `openssl` as a dev dependency from `aws-lc-fips-sys` because it was taking up too much time during the crate testing phase. I've added a FIPS check test assertion instead.
* FIPS static build only builds for linux x86_64 (gcc/clang) and arm (clang). I've only added support for these two platforms. We can look into using the shared FIPS build (which supports more OEs) as a subsequent task.
* Fips generation script has quite a few differences between the `aws-lc-sys` generation process, so I've created a separate template and generation directory.
* `OPENSSL_armcap_P` is intentionally removed from the generated symbol list, to support the prefix ARM FIPS static build.

### Call-outs:
N/A

### Testing:
`./generate-fips.sh` script

the ISC license.


bindings/rust/publish/publish_fips.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
justsmth: Nit-pick: We use dash (`-`) for `generate-fips.sh`, but underscore (`_`) for `publish_fips.sh`.

bindings/rust/aws-lc-fips-sys-template/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+version = "__AWS_LC_FIPS_SYS_VERSION__"
+authors = ["AWS-LC"]
+edition = "2021"
+repository = "https://github.com/awslabs/aws-lc"
+license = "Apache-2.0 OR ISC"
+rust-version = "1.57.0"
+include = [
+    "build.rs",
+    "Cargo.toml",
+    "deps/aws-lc/**/*.c",
```
justsmth, line 13: Missing `build_bindgen.rs`?

justsmth (changes requested):

bindings/rust/aws-lc-fips-sys-template/Cargo.toml
samuel40791765: This one was intentional. We don't expose the `generate_bindings` in `aws-lc-fips-sys` since FIPS is only supported on the platforms we offer pre-generated bindings for, so the `build_bindgen` file isn't needed in the official crate.

bindings/rust/publish/publish_fips.sh
samuel40791765: Changed :)

justsmth (approved):

bindings/rust/aws-lc-fips-sys-template/build_bindgen.rs
```diff
@@ -0,0 +1,103 @@
+// SPDX-License-Identifier: Apache-2.0 OR ISC
```
skmcgrail, line 1: Minor comment, not blocking on the PR. But we should migrate this to a build folder similar to [openssl-sys](https://github.com/sfackler/rust-openssl/tree/master/openssl-sys/build). That way we don't have have multiple build related files at the root of the crate. You can specify this separate location in the [`Cargo.toml`](https://github.com/sfackler/rust-openssl/blob/master/openssl-sys/Cargo.toml#L14). Just keeps the crate root cleaner.

skmcgrail (approved):

bindings/rust/aws-lc-fips-sys-template/build_bindgen.rs
samuel40791765: That's a good point. I'm planning to add `README.md`s for both sys crates in the next PR. I'll apply the recommended directory shuffling along with that change then.

PR #720, opened 2022-12-08
Speed up CI

andrewhop:
### Description of changes: 
Speed up the CI by tweaking the slowest parts:
* Formal verification is almost perfectly parallelizable: the instances are at 100% CPU utilization for the entire build
* Shorten the duration of the fuzz tests per CI run, with the larger corpus in EFS it takes longer to merge in the final results
* Use a larger instance for the Intel SDE build, like the formal verification the instances are at 100% utilization for the entire build

Using more/larger instances doesn't actually cost us more money if the build finishes faster. The CodeBuild pricing is basically the same 70 minutes of a xlarge == 35 minutes of a 2xlarge, or 35 minutes of 2 xlarge instance.

### Testing:
Tested in my local account's CI.

the ISC license.


tests/ci/cdk/cdk/codebuild/github_ci_linux_x86_omnibus.yaml
```diff
@@ -277,7 +277,7 @@ batch:
     # When 'AES_GCM_SELECTCHECK' is defined, AES-GCM formal verification is executed against more parameters.
     # Each dimension only checks |evp_cipher_update_len| from AES_GCM_SELECTCHECK_START to AES_GCM_SELECTCHECK_END.
     # https://github.com/awslabs/aws-lc-verification/blob/master/SAW/proof/AES/verify-AES-GCM-selectcheck-template.txt
-    - identifier: ubuntu2004_clang10x_formal_verification_aes_gcm_selectcheck_1_172
+    - identifier: ubuntu2004_clang10x_formal_verification_aes_gcm_selectcheck_1_140
```
torben-hansen, line 280: Almost at the point where we should just write a script that generates these stanza's.

tests/ci/run_cryptofuzz.sh
```diff
@@ -65,9 +65,9 @@ SRC_CORPUS="$CRYPTOFUZZ_SEED_CORPUS"
 
 # Perform the actual fuzzing. We want the total build time to be about an 45 minutes:
```
torben-hansen, line 66:
```suggestion
# Perform the actual fuzzing. We want the total build time to be about 45 minutes:
```

tests/ci/cdk/cdk/codebuild/github_ci_linux_x86_omnibus.yaml
andrewhop: I thought about it, but this will go away in the near future with the unbounded proof that is faster and won't need to be split up like this.

dkostic (approved):

PR #722, opened 2022-12-12
Fix Jitter time measurement on Apple platforms

dkostic:
### Issues:
CryptoAlg-1441

### Description of changes: 

On Apple platforms, Jitter RNG implementation uses `mach_absolute_time`
function to get the current value of the clock. However, this function
deals in "tick" units, not in time units. On Intel based Apple
platforms 1 tick = 1 ns, which is good enough timer resolution for
Jitter. On M1 macbooks 1 tick ~ 41.67 ns. This seems to be a problem
for Jitter since it causes frequent failures to produce enough entropy
on M1 macbooks. Another potential cause for the issue might be that
`mach_absolute_time` clock doesn't increment while they system is
asleep. I am not sure what is the definition of "asleep" on M1 macs,
how it correlates to performance and efficiency cores and different
power saving states, etc. The upstream Jitter RNG repository has
recently added support for using the system counter on `aarch64`
which seems like way to go for all 64-bit Arm platforms.

So this change:
  - Adds the system counter calls on `aarch64` platforms,
  - Removes the `__MACH__` special case.

### Call-outs:

### Testing:

the ISC license.


third_party/jitterentropy/jitterentropy-base-user.h
```diff
@@ -112,14 +112,24 @@ static inline void jent_get_nstime(uint64_t *out)
...
+
+#else /* (__x86_64__) || (__i386__) || (__aarch64__) */
 
 static inline void jent_get_nstime(uint64_t *out)
 {
-	/* OSX does not have clock_gettime -- taken from
-	 * http://developer.apple.com/library/mac/qa/qa1398/_index.html */
 # ifdef __MACH__
-	*out = mach_absolute_time();
+	*out = clock_gettime_nsec_np(CLOCK_MONOTONIC_RAW);
```
Taffer, line 132: Is there a URL explaining `mach_absolute_time()` vs `clock_gettime_nsec_np()` we could link to? Or just `clock_gettime_nsec_np()` docs would probably be fine too.

Taffer:
This is required for the `fips-2021-*-1MU` branch.

third_party/jitterentropy/jitterentropy-base-user.h
Taffer: Adding this to a comment there is enough IMHO: https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time

dkostic: done

third_party/jitterentropy/jitterentropy-base-user.h
```diff
@@ -112,14 +112,28 @@ static inline void jent_get_nstime(uint64_t *out)
...
+
+#else /* (__x86_64__) || (__i386__) || (__aarch64__) */
 
 static inline void jent_get_nstime(uint64_t *out)
 {
-	/* OSX does not have clock_gettime -- taken from
-	 * http://developer.apple.com/library/mac/qa/qa1398/_index.html */
 # ifdef __MACH__
-	*out = mach_absolute_time();
+    /* Apple suggests using clock_gettime_nsec_np instead of mach_* functions:
```
torben-hansen, line 132: Indentation... :(

third_party/jitterentropy/jitterentropy-base-user.h
```diff
@@ -112,14 +112,28 @@ static inline void jent_get_nstime(uint64_t *out)
...
+        *out = ctr_val;
+}
+
+#else /* (__x86_64__) || (__i386__) || (__aarch64__) */
 
 static inline void jent_get_nstime(uint64_t *out)
 {
-	/* OSX does not have clock_gettime -- taken from
-	 * http://developer.apple.com/library/mac/qa/qa1398/_index.html */
 # ifdef __MACH__
```
torben-hansen, line 131: Major nit: Line 148 should be `(__MACH__) || (_AIX)`

third_party/jitterentropy/jitterentropy-base-user.h
```diff
@@ -112,14 +112,28 @@ static inline void jent_get_nstime(uint64_t *out)
...
-	*out = mach_absolute_time();
+    /* Apple suggests using clock_gettime_nsec_np instead of mach_* functions:
+	 * https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time
+	 * https://developer.apple.com/documentation/kernel/1646199-mach_continuous_time
+	 */
+	*out = clock_gettime_nsec_np(CLOCK_MONOTONIC_RAW);
```
torben-hansen, line 136: `man clock_gettime_nsec_np ` suggests this function call can fail:
```
For clock_gettime_nsec_np() a return value of non-0 indicates success.  A 0
     return value indicates an error occurred and an error code is stored in errno.
```
Should this be handled? At least should be documented why it is ignored.

dkostic: Now when I think about it, should we just delete the whole `__MACH__` branch? Every Apple platform is either `aarch64` or `x86`, both covered with the system counters above.

torben-hansen: Unless, we hit an Apple 32-bit Arm somewhere? Laptop not. But what's the story on IOS for example?

torben-hansen: Okay, I don't know the distribution history for iOS. If it doesn't impact anything, I don't see why not.

Taffer (dismissed):

third_party/jitterentropy/jitterentropy-base-user.h
dkostic: I verified offline that we can remove the special case (32-bit Apple platforms requiring FIPS) and removed it.

torben-hansen (approved):

Taffer (approved):

PR #724, opened 2022-12-13
Add Dilithium to bssl speed benchmarks

jakemas:
### Description of changes: 
Adds benchmarks for Dilithium3 key generation, sign, and verify.

### Call-outs:
Followed recent benchmark formatting as done in https://github.com/awslabs/aws-lc/pull/664.

### Testing:
Manually ran benchmarks and sanity checked against [liboqs](https://openquantumsafe.org/benchmarking/visualization/speed_sig.html).

the ISC license.


tool/speed.cc
```diff
@@ -1907,6 +1966,7 @@ bool Speed(const std::vector<std::string> &args) {
      !SpeedECDH(selected) ||
      !SpeedECDSA(selected) ||
      !SpeedECKeyGen(selected) ||
+     !SpeedDilithium(selected) ||
```
andrewhop, line 1969: This is going to have to go in the !OpenSSL block down by SpeedKEM, and the above methods should go inside the if/def next to the KEM benchmark.

crypto/dilithium/README.md
```diff
@@ -40,4 +43,34 @@ EVP_DigestSign(md_ctx.get(), signature, &signature_len, msg, msg_len);
...
++-------------------------------+-----------+-----------------+---------+--------------+----------+----------------+
+| Dilithium3 AWS-LC(x86_64-ref) |    4762.5 |          524934 |  2529.6 |       988298 |   5053.2 |         494736 |
+| Dilithium3 OQS (x86_64-ref)   |    5315.0 |          470266 |  1358.3 |      1840581 |   5690.7 |         439257 |
++-------------------------------+-----------+-----------------+---------+--------------+----------+----------------+
+```
+Dilithium3 AWS-LC(x86_64-ref) benchmarks were taken with the following configuration:
+```
+CPU	        Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz 
+Platform 	Linux ip-172-31-41-0 5.15.0-1026-aws
+Instance	t3.medium
```
andrewhop, line 74: You shouldn't benchmark on T* instances which have variable performance as a feature:
> Amazon EC2 allows you to choose between Fixed Performance instance families (e.g. M6, C6, and R6) and Burstable Performance Instance families (e.g. T3). Burstable Performance Instances provide a baseline level of CPU performance with the ability to burst above the baseline.
From https://aws.amazon.com/ec2/instance-types/

crypto/dilithium/README.md
```diff
@@ -40,4 +43,34 @@ EVP_DigestSign(md_ctx.get(), signature, &signature_len, msg, msg_len);
...
+
+Open Quantum Safe (OQS) provides benchmarks for the liboqs implementation of Dilithium based upon the same reference implementation. Full OQS benchmarks are available at [OQS benchmarking](https://openquantumsafe.org/benchmarking/visualization/speed_sig.html). For comparison we include the OQS Dilithium3 benchmarks from 2022-12-07.
+
+``` 
++-------------------------------+-----------+-----------------+---------+--------------+----------+----------------+
+|           Algorithm           | keypair/s | keypair(cycles) | sign/s  | sign(cycles) | verify/s | verify(cycles) |
++-------------------------------+-----------+-----------------+---------+--------------+----------+----------------+
+| Dilithium3 AWS-LC(x86_64-ref) |    4762.5 |          524934 |  2529.6 |       988298 |   5053.2 |         494736 |
+| Dilithium3 OQS (x86_64-ref)   |    5315.0 |          470266 |  1358.3 |      1840581 |   5690.7 |         439257 |
++-------------------------------+-----------+-----------------+---------+--------------+----------+----------------+
```
andrewhop, line 68: Historically we haven't published performance numbers in our readme's @dkostic what do you think? This isn't offering a performance guarantee, but is this data valuable on it's own? Should we include the gen/sign/verify time of all algorithms?

andrewhop (commented):
Can you include the output of both:
```
./bssl speed -filter Dilithium3
./bssl speed -filter Dilithium3 -json
```
To verify the names and output will work with internal tooling.

crypto/dilithium/README.md
jakemas: Thanks, I was going to ask you before I spun one up if there was a good choice.  I wanted to target the same processor as OQS so the benchmarks were like-for-like. Looks like m5d.xlarge uses Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz, what do you think of m5d?

crypto/dilithium/README.md
dkostic: While the numbers we present are not a guarantee, it would be bad customer experience if the numbers didn't match the reality, so we'd have to updated the numbers every time there is a change. But not only changes of the core primitive can affect the performance. For example, Dilithium3 goes through layers of EVP signature implementation so any change in those may impact the performance. Granted, I don't expect this to happen often, but because it can happen we'd have to check and update the numbers accordingly. Then the next set of problems arises -- we need to ensure that every time we measure the performance in exactly the same conditions (CPU, OS, compiler, build flags, etc.), that the measured performance is not an outlier, etc. While this might be doable for a single algorithm, I'm concerned about scalability and maintainability. I think a much better solution would be to make the data from the internal performance dashboards publicly available in some way.

The comparison with libOQS for Dilithium is more problematic because:
 - we can't ensure that the numbers libOQS presents were obtained in the same conditions as our numbers (CPU is not the only variable in the system, especially when we are dealing with C code),
 - the code flow in libOQS (OpenSSL) and AWS-LC is not the same which means we are not comparing apples to apples. For example, the numbers Jake added show that signing is almost twice slower in AWS-LC, while theoretically they should be the same because we use exactly the same core implementation of Dilithium. How can we explain this? And why is there no such difference in other operations?
We've already been dealing with these problems in the benchmarking tool and we have a solution, so again, I'd rather add Dilithium to the general benchmarking tool. And if we really want to compare with libOQS we should add it to the tool as well.

A bit off-topic: @jakemas how did you get the number of cycles?

jakemas: Agree with Dusan here.
> We've already been dealing with these problems in the benchmarking tool and we have a solution, so again, I'd rather add Dilithium to the general benchmarking tool. And if we really want to compare with libOQS we should add it to the tool as well.
What tool is this? bssl speed? If so, this PR is doing exactly that - adding dilithium to the general benchmarking tool. I would just remove OQS and the table to be consistent with other benchmarks.

tool/speed.cc
jakemas: Addressed in [096f384](https://github.com/awslabs/aws-lc/pull/724/commits/096f3840d3a43a74655b995a630e0544ae9b9afd), let me know if I misunderstood.

crypto/dilithium/README.md
jakemas: removed benchmark data in [096f384](https://github.com/awslabs/aws-lc/pull/724/commits/096f3840d3a43a74655b995a630e0544ae9b9afd)

tool/speed.cc
```diff
@@ -660,10 +660,75 @@ static bool SpeedKEM(const std::string &name, int nid, const std::string &select
...
+  // each algorithm) at run time by attempting a sign with a NULL signature.
+  // The sign algorithm must support calling NULL to obtain the signature length
+  size_t sig_len = 0;
+  EVP_DigestSignInit(md_ctx.get(), NULL, NULL, NULL, key);
+  EVP_DigestSign(md_ctx.get(), NULL, &sig_len, msg, msg_len);
+  std::unique_ptr<uint8_t[]> signature(new uint8_t[sig_len]);
+
+
+  if (!TimeFunction(&results, [&md_ctx, &key, &signature, &sig_len, msg_len ]() -> bool {
+        return EVP_DigestSignInit(md_ctx.get(), NULL, NULL, NULL, key) &&
```
andrewhop, line 708: Do we need to call sign init each iteration?

tool/speed.cc
```diff
@@ -660,10 +660,75 @@ static bool SpeedKEM(const std::string &name, int nid, const std::string &select
...
+               EVP_DigestSign(md_ctx.get(), signature.get(), &sig_len, msg, msg_len);
+      })) {
+    return false;
+  }
+  results.Print(name + " signing");
+
+  // Verify
```
andrewhop, line 715: You never call EVP_DigestVerifyInit, do we need to call SignInit? Looking at at Dilithium sign and verify init are always null and a noop https://github.com/awslabs/aws-lc/blob/integrate-pq/crypto/dilithium/p_dilithium3.c#L100. I could see value in calling it for future compatibility, but we should do both or neither.

tool/speed.cc
jakemas: you are correct, removed in [f865b54](https://github.com/awslabs/aws-lc/pull/724/commits/f865b5454bc50474974894a87307ec1621b09a3c)

tool/speed.cc
jakemas: I agree, I opt to do neither as a better representation of sign/verify [f865b54](https://github.com/awslabs/aws-lc/pull/724/commits/f865b5454bc50474974894a87307ec1621b09a3c)

andrewhop: You still call EVP_DigestSignInit up on line 702 once.

dkostic (approved):

andrewhop (approved):

PR #727, opened 2022-12-14
Fips 2021 10 20 1 mu jitter

Taffer:
### Issues:
Resolves CryptoAlg-1442

### Description of changes: 
Use `clock_gettime_nsec_np(CLOCK_MONOTONIC_RAW);` for Apple Silicon macOS, the previous technique didn't have a high-enough resolution on ARM.

### Call-outs:
The fix in `main` (which removes the `__MACH__` branch) doesn't work in this old FIPS branch. I have no idea why, hopefully just because the version of JitterEntropy is old here.

### Testing:
`ninja run_test` and all is happy in FIPS and non-FIPS mode on Apple Silicon.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and  the ISC license.

### DIFF
--- a/third_party/jitterentropy/jitterentropy-base-user.h
+++ b/third_party/jitterentropy/jitterentropy-base-user.h
@@ -109,15 +109,23 @@ static inline void jent_get_nstime(uint64_t *out)
 	*out = EAX_EDX_VAL(val, low, high);
 }
 
-#else /* __x86_64__ */
+#elif defined(__aarch64__)
 
 static inline void jent_get_nstime(uint64_t *out)
 {
-	/* OSX does not have clock_gettime -- taken from
-	 * http://developer.apple.com/library/mac/qa/qa1398/_index.html */
-# ifdef __MACH__
-	*out = mach_absolute_time();
-# elif _AIX
+        uint64_t ctr_val;
+        /*
+         * Use the system counter for aarch64 (64 bit ARM).
+         */
+        __asm__ volatile("mrs %0, cntvct_el0" : "=r" (ctr_val));
+        *out = ctr_val;
+}
+
+#else /* (__x86_64__) || (__i386__) || (__aarch64__) */
+
+static inline void jent_get_nstime(uint64_t *out)
+{
+#ifdef _AIX
 	/* clock_gettime() on AIX returns a timer value that increments in
 	 * steps of 1000
 	 */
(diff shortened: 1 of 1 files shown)


torben-hansen:
Can you upgrade jitter entropy to a newer version?

dkostic:
> The fix in main (which removes the __MACH__ branch) doesn't work in this old FIPS branch
Why it doesn't work? The fix in main should not have any impact on the CI because in the CI we are testing only x86 macOS which would be covered by the existing `#ifdef __x86_64__`, which means the new code path should not even be reachable (in the CI tests).

Taffer:
> > The fix in main (which removes the **MACH** branch) doesn't work in this old FIPS branch
> 
> Why it doesn't work? The fix in main should not have any impact on the CI because in the CI we are testing only x86 macOS which would be covered by the existing `#ifdef __x86_64__`, which means the new code path should not even be reachable (in the CI tests).

Running locally, the unit tests fail with the same jitter errors as before; works fine with the `__MACH__` bit added back in. If it didn't take 650+ days to get through NIST approval, we could drop this in favour of the new FIPS branch...

dkostic:
> Running locally, the unit tests fail with the same jitter errors as before;

Is this on an M1 mac? I just tried it on my M1 and don't see a failure.

Taffer:
> > Running locally, the unit tests fail with the same jitter errors as before;
> 
> Is this on an M1 mac? I just tried it on my M1 and don't see a failure.

Yeah; let me try again.

Taffer:
> > Running locally, the unit tests fail with the same jitter errors as before;
> 
> Is this on an M1 mac? I just tried it on my M1 and don't see a failure.

Maybe my system is "too fast" or something?

```
[419/423] cd /Users/herborth/git/aws-l...-lc-main/build-fips/ssl/test/bssl_shim
FAILED (TLS-TLS1-RSA_WITH_3DES_EDE_CBC_SHA-LargeRecord)
unexpected failure: local error 'child exited early: signal: abort trap', child error 'signal: abort trap', stdout:

stderr:
CPU Jitter entropy RNG initialization failed.
```

Running macOS 12.6.1 with current (updated yesterday, IIRC) command-line tools. Building with `cmake -G Ninja -DFIPS=TRUE -DBUILD_SHARED_LIBS=TRUE -DCMAKE_BUILD_TYPE=RelWithDebInfo ..` from a sub-directory.

dkostic:
> 

amazing, exact same commands on my mac -- doesn't fail.

Taffer:
> > 
> 
> amazing, exact same commands on my mac -- doesn't fail.

But like this are great. :-P

PR #729, opened 2022-12-16
OpenSSL Crate Compatibility Support

skmcgrail:
### Description of changes: 
Updates `aws-lc-sys` and `aws-lc-fips-sys` crates to build libssl in addition to libcrypto. Adds in some missing components needed for openssl crate compatibility.

* Adds `ssl` feature which will toggle on generation / selection of bindnings which include the libssl components.
* The crate build script now generates bindings for non pre-generated platforms to `${OUT_DIR}/bindings.rs` which will be included by the `src/lib.rs` automatically. Using `${OUT_DIR}` ensures we are not modifying the actual crate files during a build. If a target platform is not per-generated, then the bindings will be generated and written to that location if `bindgen` is enabled.


samuel40791765 (approved):

bindings/rust/aws-lc-fips-sys-template/deps/aws-lc/include/rust_wrapper.h
```diff
@@ -100,6 +100,7 @@ int ERR_GET_FUNC_RUST(uint32_t packed_error);
 #include "openssl/sha.h"
 #include "openssl/siphash.h"
 #include "openssl/span.h"
+#include "openssl/ssl.h"
```
samuel40791765, line 103: If we add `#include "openssl/ssl3.h"` here, do we still need to create `rust_ssl_wrapper.h`?

samuel40791765 (commented):
Is it possible to directly include the `ssl` feature as part of our default bindings? `generate_src_bindings` currently generates two files, `${platform}_crypto` and `${platform}_crypto_ssl`, and both are included part of our sys crates. 
Can we just use`${platform}_crypto_ssl` directly as our "pre-generated" bindings file going forward to minimize some complexity?

bindings/rust/aws-lc-fips-sys-template/build/main.rs
```diff
@@ -178,17 +182,69 @@ fn prepare_cmake_build(build_prefix: Option<&str>) -> cmake::Config {
...
-    cmake_cfg.build_target("crypto").build()
+    // cmake supports passing multiple arguments to target, but this is broken in the cmake crate
+    // ssl requires crypto so we can get away with just picking the top-most requried one.
+    let target = if cfg!(feature = "ssl") {
+        Some("ssl")
+    } else {
+        Some("crypto")
+    };
+
+    cmake_cfg.build_target(target.unwrap()).build()
```
justsmth, line 193: I don't see a need for `Some(...)` here since we unconditionally unwrap it anyways:
```
    let target = if cfg!(feature = "ssl") {
        "ssl"
    } else {
        "crypto"
    };

    cmake_cfg.build_target(target).build()
```

justsmth (approved):

bindings/rust/aws-lc-fips-sys-template/deps/aws-lc/include/rust_wrapper.h
justsmth: I don't think we want `ssl.h` included here -- it will cause bindings for the ssl library to be generated when we're not building the ssl library. 

I'm curious why this line was added.

bindings/rust/aws-lc-sys-template/src/lib.rs
```diff
@@ -1,6 +1,9 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR ISC
 
+use paste::paste;
+use std::os::raw::{c_long, c_char, c_void};
```
justsmth, line 5: The bindings generated use primitive types from `std::os::raw` (not from `std::ffi`).

justsmth (approved):

samuel40791765 (approved):

PR #731, opened 2022-12-17
Add back ssl stat counters

samuel40791765:
### Issues:
Resolves `CryptoAlg-1462`

### Description of changes: 
This is an attempt to add back the ssl stat counters that were taken out from BoringSSL in these commits:
* https://github.com/awslabs/aws-lc/commit/da881e9a15f673e862eeeb30c3112aa88c747ede
* https://github.com/awslabs/aws-lc/commit/b6d0c6db5e132f600eae9e70ed97144ad5fab221

### Call-outs:
Details can be found in the doc under the thread of the SIM

### Testing:
WIP

the ISC license.


ssl/handshake_server.cc
```diff
@@ -399,6 +399,8 @@ static const SSL_CIPHER *choose_cipher(
 static enum ssl_hs_wait_t do_start_accept(SSL_HANDSHAKE *hs) {
   ssl_do_info_callback(hs->ssl, SSL_CB_HANDSHAKE_START, 1);
   hs->state = state12_read_client_hello;
+  ssl_update_counter(hs->ssl->session_ctx.get(),
+                     hs->ssl->session_ctx->stats.sess_accept, true);
```
andrewhop, line 403: What are these info callbacks used for? In some places they're called before the counter is updated, and others after. Could these info callbacks use the counter and expect to see (or not see) the up to date counter?

ssl/internal.h
```diff
@@ -3533,6 +3537,25 @@ struct ssl_ctx_st {
   SSL_SESSION *(*get_session_cb)(SSL *ssl, const uint8_t *data, int len,
                                  int *copy) = nullptr;
 
+  struct {
+    int sess_connect = 0;             // SSL new conn - started
+    int sess_connect_renegotiate = 0; // SSL reneg - requested
+    int sess_connect_good = 0;        // SSL new conne/reneg - finished
+    int sess_accept = 0;              // SSL new accept - started
+    int sess_accept_renegotiate = 0;  // SSL reneg - requested
```
andrewhop, line 3545: Can we save a few bytes and not include sess_accept_renegotiate which is always zero because we don't support server side renegotiation?

ssl/ssl_lib.cc
```diff
@@ -480,6 +480,20 @@ bool SSL_get_traffic_secrets(const SSL *ssl,
   return true;
 }
 
+void ssl_update_counter(SSL_CTX *ctx, int &counter, bool locked) {
+  if (locked) {
+    MutexWriteLock lock(&ctx->lock);
```
andrewhop, line 485: This locks a SSL_CTX lock.

ssl/ssl_session.cc
```diff
@@ -661,11 +661,15 @@ static enum ssl_hs_wait_t ssl_lookup_session(
           MakeConstSpan(sess->session_id, sess->session_id_length);
       return key_id == sess_id ? 0 : 1;
     };
-    MutexReadLock lock(&ssl->session_ctx->lock);
+    MutexWriteLock lock(&ssl->session_ctx->lock);
```
andrewhop, line 664: This is locking the SSL_CTX's session context lock, should this stay the same and we let the ssl_update_counter get the SSL_CTX lock in that funciton?

ssl/handshake_client.cc
```diff
@@ -1863,6 +1865,8 @@ static enum ssl_hs_wait_t do_finish_client_handshake(SSL_HANDSHAKE *hs) {
   ssl->s3->initial_handshake_complete = true;
   if (has_new_session) {
     ssl_update_cache(ssl);
+    ssl_update_counter(ssl->session_ctx.get(),
+                        ssl->session_ctx->stats.sess_hit, true);
```
andrewhop, line 1869: If this is a new session it's not reused right? Are you sure this should be updated here?

ssl/ssl_session.cc
```diff
@@ -661,11 +661,15 @@ static enum ssl_hs_wait_t ssl_lookup_session(
...
     // |lh_SSL_SESSION_retrieve_key| returns a non-owning pointer.
     session = UpRef(lh_SSL_SESSION_retrieve_key(ssl->session_ctx->sessions,
                                                 &session_id, hash, cmp));
     // TODO(davidben): This should probably move it to the front of the list.
+    if (session == nullptr) {
+      ssl_update_counter(ssl->session_ctx.get(),
+                          ssl->session_ctx->stats.sess_hit, /*lock=*/ false);
+    }
```
andrewhop, line 672: Here too, if the session is null does that mean this is a new session?

ssl/ssl_session.cc
```diff
@@ -698,11 +704,15 @@ static enum ssl_hs_wait_t ssl_lookup_session(
...
   if (session && !ssl_session_is_time_valid(ssl, session.get())) {
+    ssl_update_counter(ssl->session_ctx.get(),
+                       ssl->session_ctx->stats.sess_timeout, true);
     // The session was from the cache, so remove it.
     SSL_CTX_remove_session(ssl->session_ctx.get(), session.get());
     session.reset();
   }
 
+  ssl_update_counter(ssl->session_ctx.get(),
+                     ssl->session_ctx->stats.sess_hit, true);
```
andrewhop, line 715: This sess_hit makes sense.

ssl/internal.h
```diff
@@ -3533,6 +3537,25 @@ struct ssl_ctx_st {
...
+    int sess_accept_good = 0;         // SSL accept/reneg - finished
+    int sess_miss = 0;                // session lookup misses
```
andrewhop, line 3547: sess_miss is never updated, but I think we can down where we check the session cache.

ssl/ssl_session.cc
samuel40791765: The function is going to be using the same `SSL_CTX` lock, since the stats are within the same `ssl->session_ctx`. If we let `ssl_update_counter` update the lock, the tests processes will crash (which is expected since we're trying to lock the mutex twice).

ssl/ssl_lib.cc
samuel40791765: Yes, this is because the `stats` structure is nested in `SSL_CTX` so we use the lock that was already with it.

ssl/ssl_test.cc
```diff
@@ -6074,6 +6107,8 @@ TEST_P(SSLVersionTest, SessionCacheThreads) {
       thread.join();
     }
   }
+  EXPECT_EQ(SSL_CTX_sess_misses(server_ctx_.get()), 255);
```
andrewhop, line 6110: Can you explain what this test is doing and why there is one less miss than timeout?

ssl/ssl_test.cc
```diff
@@ -6074,6 +6107,8 @@ TEST_P(SSLVersionTest, SessionCacheThreads) {
...
+  EXPECT_EQ(SSL_CTX_sess_timeouts(server_ctx_.get()), 256);
```
andrewhop, line 6111: Also can these both use a shared variable for clarity?

ssl/ssl_test.cc
```diff
@@ -2962,6 +2962,31 @@ class SSLVersionTest : public ::testing::TestWithParam<VersionParam> {
...
+    EXPECT_EQ(SSL_CTX_sess_accept(client_ctx_.get()), 0);
+    EXPECT_EQ(SSL_CTX_sess_accept(server_ctx_.get()), 0);
+    EXPECT_EQ(SSL_CTX_sess_accept_renegotiate(client_ctx_.get()), 0);
+    EXPECT_EQ(SSL_CTX_sess_accept_renegotiate(server_ctx_.get()), 0);
+    EXPECT_EQ(SSL_CTX_sess_accept_good(client_ctx_.get()), 0);
+    EXPECT_EQ(SSL_CTX_sess_accept_good(server_ctx_.get()), 0);
+    EXPECT_EQ(SSL_CTX_sess_hits(client_ctx_.get()), 0);
+    EXPECT_EQ(SSL_CTX_sess_hits(server_ctx_.get()), 0);
+    EXPECT_EQ(SSL_CTX_sess_cb_hits(client_ctx_.get()), 0);
+    EXPECT_EQ(SSL_CTX_sess_cb_hits(server_ctx_.get()), 0);
```
andrewhop, line 2981: We're missing a test when SSL_CTX_sess_cb_hits is not 0.

samuel40791765: Added a test called `SessionExternalCacheHit`

andrewhop (dismissed):

torben-hansen:
Does this take into account the session transfer feature?

samuel40791765:
> Does this take into account the session transfer feature?

No, but good question. These were based on a customer ask from the same team, the counters provide valuable insight for them into SSL usage and interactions. 
`SSL_CTX` is a shared state management configuration, so the data inside it won't be transferred around like the `SSL` structures for the session transfer feature.

ssl/ssl_session.cc
```diff
@@ -879,6 +892,7 @@ static bool add_session_locked(SSL_CTX *ctx, UniquePtr<SSL_SESSION> session) {
                           /*lock=*/false)) {
         break;
       }
+      ssl_update_counter(ctx, ctx->stats.sess_cache_full, /*lock=*/ false);
```
dkostic, line 895: this is the only place where we provide `false` for the lock argument, why is that?

samuel40791765: This placement is within the function `add_session_locked`. `add_session_locked` in this context means that the mutex has already been locked before this function is called. 
I added a "lock" argument to `ssl_update_counter` to indicate whether the mutex has been locked already, to prevent the mutex from being locked twice. There's also a similar function above called `remove_session` which applies the same concept.

dkostic: can we update the variable name (`locked` to `lock`) and function description then to make this more clear, right now it says:
```
// ssl_update_counter updates the stat counters in |SSL_CTX|. `locked` should be
// set to false if the mutex has already been locked.
void ssl_update_counter(SSL_CTX *ctx, int &counter, bool locked);
```

samuel40791765: Updated :)

dkostic (approved):

andrewhop (approved):

PR #732, opened 2022-12-19
Add hkdf_expand() as a KDF in feedback mode to the ACVP tool

billbo-yang:
### Issues:
Resolves CryptoAlg-1344

### Description of changes: 
AWS-LC currently has `hkdf_expand()` in the FIPS boundary as a KBKDF in feedback mode (8 bit counter after fixed data, requires zero iv length). However, the ACVP tool does not support testing it. This PR adds support for `hkdf_expand()` as a KBKDF in feedback mode to the ACVP tool.

### Call-outs:
The way NIST implemented this particular ACVP test is interesting. Namely, the fixeddata variable is generated by the IUT and provided to NIST, who will then use the provided fixed data to check if the implementation output is correct. Due to this, testing with input/expected vectors locally is not possible because the fixeddata is generated randomly by the implementation.

As such, the local test for KDF simply runs the parser on an input vector to ensure that no errors arise when parsing, but does not check the correctness of the output.

### Testing:
Change was tested by running the ACVP test for KDF in online mode and ensuring that it passed the NIST demo server test.

the ISC license.


util/fipstools/acvp/acvptool/subprocess/kdf.go
```diff
@@ -74,54 +73,46 @@ func (k *kdfPrimitive) Process(vectorSet []byte, m Transactable) (interface{}, e
...
-			break
-		default:
-			return nil, fmt.Errorf("Label location %q not supported", group.CounterLocation)
+		if group.CounterLocation != "after fixed data" {
+			// We only support the counter location being after fixed data
+			return nil, fmt.Errorf("label location %q not supported", group.CounterLocation)
+		}
+
+		if group.CounterBits != 8 {
+			// We only support counter lengths of 32
```
Taffer, line 87: I think you mean 8. ;-)

Taffer (approved):
1 comment to fix, otherwise looks good to me

nebeid (approved):

samuel40791765 (approved):

PR #734, opened 2022-12-20
remove kas-kdf from ACVP tool

billbo-yang:
### Issues:
As discussed in [this BoringSSL commit](https://github.com/google/boringssl/commit/726585c58ec89f65ab669187665232790cbf3d02), NIST has changed how they test HKDF from using KAS-KDF to KDA. We've already implemented this change in [PR658](https://github.com/awslabs/aws-lc/pull/658), but there is still some leftover code that references it that we now know is out of date. This PR removes the remnants of KAS-KDF from our codebase.

### Description of changes: 
Currently AWS-LC's ACVP tool still has some logic for handling KAS-KDF for HKDF. This PR removes them to reduce confusion as to what part of the codebase tests HKDF. Additionally, it renames our current parsing implementation for HKDF from `kda.go` to `hkdf.go`. With these changes, both AWS-LC and BoringSSL will have files named `hkdf.go` that are functionally the same.

### Call-outs:
N/A

### Testing:
Local ACVP testing script was run and online mode for HKDF was run in order to ensure that functionality for ACVP tool remains unaffected.

the ISC license.


Taffer (dismissed):

util/fipstools/acvp/acvptool/subprocess/subprocess.go
```diff
@@ -106,7 +106,6 @@ func NewWithIO(cmd *exec.Cmd, in io.WriteCloser, out io.ReadCloser) *Subprocess
 		"ctrDRBG":           &drbg{"ctrDRBG", map[string]bool{"AES-128": true, "AES-192": true, "AES-256": true}},
 		"hmacDRBG":          &drbg{"hmacDRBG", map[string]bool{"SHA-1": true, "SHA2-224": true, "SHA2-256": true, "SHA2-384": true, "SHA2-512": true}},
 		"KDF":               &kdfPrimitive{},
-		"KAS-KDF":           &hkdf{},
```
nebeid, line 109: In BoringSSL's commit, the line below has `&hkdf{},` do we want to have the same? See https://github.com/google/boringssl/commit/726585c58ec89f65ab669187665232790cbf3d02#diff-6fae1d42d6f42e954a91feb6e08d8802c80eb49863f45a76ed5d1ae84414398cR109

util/fipstools/acvp/acvptool/testmodulewrapper/testmodulewrapper.go
```diff
@@ -90,39 +90,6 @@ func getConfig(args [][]byte) error {
...
-				"after fixed data"
-			],
-			"counterLength": [
-				8
-			],
-			"requiresEmptyIv": true,
-			"supportsEmptyIv": true
-		}],
-		"l": 256,
-		"z": [256, 384]
```
nebeid, line 125: Why we don't have the entry for? 
```
"algorithm": "KDA",
"mode": "HKDF",
```
See
https://github.com/google/boringssl/commit/726585c58ec89f65ab669187665232790cbf3d02#diff-6fae1d42d6f42e954a91feb6e08d8802c80eb49863f45a76ed5d1ae84414398cR109

billbo-yang: We do, it's just not in the `testmodulewrapper.go` file since that file isn't used for testing our crypto implementation (It's just a golang implementation of an algorithm to ensure the parser works). You can find the entry you want [here](https://github.com/awslabs/aws-lc/blob/f6f2eb0f94add6cfecbbe528b87009b8a658e86c/util/fipstools/acvp/modulewrapper/modulewrapper.cc#L921).

util/fipstools/acvp/acvptool/subprocess/subprocess.go
billbo-yang: I think at the very least the change will reduce confusion as to what is actually being tested with the KDF algorithm. The current name `kdfPrimitive` could infer any one of the different KDFs that ACVP supports when it actually is just `HKDF`. Will change.

util/fipstools/acvp/acvptool/testmodulewrapper/testmodulewrapper.go
nebeid: Thanks, Bill. Do we want to add this feature to `testmodulewrapper.go` or we have another way to ensure the parser works for hkdf?

util/fipstools/acvp/acvptool/subprocess/subprocess.go
nebeid: Does this bring the file `hkdf.go` to be similar to BoringSSL's?, if so, may be we should update the PR description and maybe mention in in the commit message at the time of merge.

billbo-yang: I would argue that BoringSSL's change brought them in line with _us_ :) but yes their version and ours are almost the same at this point. I'll make a note in the description and make sure to mention it in the commit message at time of merge as well.

util/fipstools/acvp/acvptool/testmodulewrapper/testmodulewrapper.go
billbo-yang: No need! the ACVP tool's offline tests test the parser (and our implementation) for both `testmodulewrapper.go` and `modulewrapper.cc`, so we're already ensuring that the parser for HKDF (and all the other algorithms in our ACVP tool) are being tested offline. You can find the file telling us what to check [here](https://github.com/awslabs/aws-lc/blob/main/util/fipstools/acvp/acvptool/test/tests.json).

nebeid (approved):

Taffer (approved):

PR #739, opened 2022-12-29
Add readme file for KEM primitive

dkostic:
### Issues:
N/A

### Description of changes: 
Add readme file for KEM primitive explaining how to use the API.

### Call-outs:

### Testing:

the ISC license.


crypto/kem/README.md
```diff
@@ -0,0 +1,159 @@
...
+// Decapsulate.
+EVP_PKEY_decapsulate(decaps_ctx, ss, &ss_len, ct, ct_len);
+
+// END                          Alice's shared secret is equal to Bob's shared secret if the protocol succeeded.
+
+
+```
+Notes about the code snipet above:
+* The functions used in the example can fail. For brevity, we omit the success/failure checks, but they should definitely be added to any code using the API.
+* Memory allocated in several places in the example is not freed. The memory should definitely be freed in any production code to avoid memory leaks.
```
andrewhop, line 158: Should we just show the complete example that cleans everything up? It might be useful to identify what memory needs to be explicitly cleaned up and "owned" by the caller to AWS-LC.

crypto/kem/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+// Decapsulate.
+EVP_PKEY_decapsulate(decaps_ctx, ss, &ss_len, ct, ct_len);
+
+// END                          Alice's shared secret is equal to Bob's shared secret if the protocol succeeded.
+
+
+```
+Notes about the code snipet above:
+* The functions used in the example can fail. For brevity, we omit the success/failure checks, but they should definitely be added to any code using the API.
```
andrewhop, line 157: How bad is adding all the error checks? If I was a totally new developer could I copy/paste this code and get in trouble down the line?

crypto/kem/README.md
```diff
@@ -0,0 +1,159 @@
...
+                          NULL, &shared_secret_len);
+```
+
+will store the size of the ciphertext and the shared secret (of the KEM specified in `ctx`) in `ciphertext_len` and `shared_secret_len`, respectively.
+
+## Full example
+
+Simulation of one KEM protocol execution using Kyber512:
+
+```
```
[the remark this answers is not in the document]
dkostic: Not sure what you are suggesting here? To create a script that takes the Readme file, extracts the code, builds it and runs it? It's an awesome idea, but I'd consider it out of scope for this PR. Maybe it's something we can work on in the future.

crypto/kem/README.md
```diff
@@ -0,0 +1,159 @@
...
+   This produces a ciphertext. Sends ciphertext to Alice.
+3. Alice decapsulates the ciphertext with her key pair and obtains the shared secret.
+   Alices and Bobs shared secrets are the same if the protocol succeeded.
+```
+
+The KEM design in AWS-LC is generic and not specific to any particular KEM. The primary use-case for the API are the post-quantum cryptographic algorithms that were specified and evaluated in NIST's Post-Quantum Crypto Standardization Project ([link](https://csrc.nist.gov/Projects/post-quantum-cryptography/post-quantum-cryptography-standardization)).
+
+## The public KEM API
+
+**Core functions.** To use a KEM, the customer needs to have access to three core functions: key generation, encapsulation, decapsulation. Following the general design in AWS-LC, these these three functions are offered in the form of standard AWS-LC EVP APIs:
```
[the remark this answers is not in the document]
dkostic: definitely! I basically copied some of the internal documentation that we did previously. It's fixed now throughout the file.

crypto/kem/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+The second way to create a context is with an already existing key (`EVP_PKEY`):
+
+```
+2. EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new(pkey, NULL);
+```
+
+This creates a fresh context of type `EVP_PKEY_KEM`, sets the specific KEM parameters defined in the `pkey` object itself, and associates `pkey` to the context. The context is now ready to encapsulate/decapsulate because it contains the encapsulation/decapsulation key. The key is usually either generated by ourselves via `EVP_PKEY_keygen`, or we received the public key of the other party as a raw byte array and created the key as shown below.
+
+**EVP_PKEY from/to raw data.** The customer sometimes needs to store/load the key-pair (`public_key, secret_key`) to/from memory. The customer certainly needs to be able to send/receive the `public_key` as part of the protocol. Therefore, we define the following methods:
```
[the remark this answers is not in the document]
dkostic: done

crypto/kem/README.md
```diff
@@ -0,0 +1,159 @@
...
+```
+2. Extract raw data from EVP_PKEY object:
+
+   OPENSSL_EXPORT int EVP_PKEY_get_raw_private_key(const EVP_PKEY *pkey,
+                                                   uint8_t *out, size_t *out_len);
+   OPENSSL_EXPORT int EVP_PKEY_get_raw_public_key(const EVP_PKEY *pkey,
+                                                  uint8_t *out, size_t *out_len);
+```
+
+**KEM parameters size.** The customer needs to know the size of they keys, the ciphertext, and the shared secret. All the sizes can be retrieved by calling the above listed APIs with `NULL` parameters in place of the actual output pointers. For example, calling:
```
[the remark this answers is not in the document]
dkostic: not really, because the exact location of the declaration/definition can change and we'd always have to remember ti update it here. I'm afraid it'll be out of sync quickly.

crypto/kem/README.md
dkostic: done

crypto/kem/README.md
dkostic: done

crypto/kem/README.md
```diff
@@ -0,0 +1,229 @@
...
+                          shared_secret, &shared_secret_len);
+
+// KEM Decapsulation
+EVP_PKEY_decapsulate(ctx, shared_secret, &shared_secret_len,
+                          ciphertext, ciphertext_len);
+```
+
+The `ctx` variable above is a pointer to a “context” object of type `EVP_PKEY_CTX`. The `pkey` variable listed as an argument of `EVP_PKEY_keygen` function is a pointer to an object of type `EVP_PKEY`. The `ciphertext` and `shared_secret` are byte arrays of length `ciphertext_len` and `shared_secret_len`, respectively.
+
+**The context.** You to create the context. Depending on the use case, this can be done in two ways:
```
andrewhop, line 36: I'm not sure what you're trying to say here with "You to create", should it be "How to create"?

dkostic: yep, typo, fixed.

crypto/kem/README.md
```diff
@@ -0,0 +1,227 @@
...
+                                                                                  // Get the required buffer lengths and allocate the buffers.
+                                                                                  if (!EVP_PKEY_encapsulate(encaps_ctx, NULL, &ct_len, NULL, &ss_len)) {
+                                                                                    goto end;
+                                                                                  }
+                                                                                  ct = (uint8_t*) OPENSSL_malloc(ct_len);
+                                                                                  ss = (uint8_t*) OPENSSL_malloc(ss_len);
+                                                                                  if (!ct || !ss) {
+                                                                                    goto end;
+                                                                                  }
+
```
nebeid, line 168: Maybe add a comment here about generating the `ss`, or a call to RAND_bytes if you want.

crypto/kem/README.md
```diff
@@ -0,0 +1,227 @@
...
+                                                                                  OPENSSL_free(ss);
+                                                                                  OPENSSL_free(ct);
+                                                                                  EVP_PKEY_free(key);
+                                                                                  EVP_PKEY_CTX_free(encaps_ctx);
+
+                                                                                  // We don't free the shared secret buffer because the whole point
+                                                                                  // of the protocol is to generate it, so we consider it as the output.
+
+                                                                                  return ret;
+                                                                    ct
```
nebeid, line 187: I'm confused a bit whether `ct` is freed or not because it's sent. And what does the comment refer to as the `shared secret buffer`?

crypto/kem/README.md
```diff
@@ -0,0 +1,227 @@
...
+The `ctx` variable above is a pointer to a “context” object of type `EVP_PKEY_CTX`. The `pkey` variable listed as an argument of `EVP_PKEY_keygen` function is a pointer to an object of type `EVP_PKEY`. The `ciphertext` and `shared_secret` are byte arrays of length `ciphertext_len` and `shared_secret_len`, respectively.
+
+**The context.** How to create the context? Depending on the use case, this can be done in two ways:
+
+```
+1. EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_KEM, NULL);
+   EVP_PKEY_CTX_kem_set_params(ctx, NID_KYBER512);
+```
+
+This creates a fresh context of type `EVP_PKEY_KEM` and sets the specific KEM parameters (Kyber512 in this example). The context is now ready for key generation (`EVP_PKEY_keygen`). However, the context created in this way doesn’t have an associated key (`EVP_PKEY`), so obviously, we can’t encapsulate/decapsulate with it. Therefore, this is useful for key generation only, i.e. before we have a key.
```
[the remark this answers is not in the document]
dkostic: so we have `EVP_PKEY_CTX` of type `EVP_PKEY_KEM/EVP_PKEY_EC/...`. That's what I meant, but I can reword if it's confusing

crypto/kem/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+The second way to create a context is with an already existing key (`EVP_PKEY`):
+
+```
+2. EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new(pkey, NULL);
+```
+
+This creates a fresh context of type `EVP_PKEY_KEM`, sets the specific KEM parameters defined in the `pkey` object itself, and associates `pkey` to the context. The context is now ready to encapsulate/decapsulate because it contains the encapsulation/decapsulation key. The key is usually either generated by ourselves via `EVP_PKEY_keygen`, or we received the public key of the other party as a raw byte array and created the key as shown below.
```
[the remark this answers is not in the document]
dkostic: same as above

crypto/kem/README.md
dkostic: the shared secret (ss) is generated inside `EVP_PKEY_encapsulate`, no need for users to do anything.

crypto/kem/README.md
dkostic: right, I don't know exactly how customers are going to implement the protocol in terms of sending, waiting for response, etc. Anyway, I'll update the example to show three separate full functions. Hopefully, it'll be more clear then

crypto/kem/README.md
nebeid: I don't think this was mentioned, though, in the readme text. I think it's worth mentioning. I see it mentioned at the end of the example.

crypto/kem/README.md
```diff
@@ -90,137 +90,198 @@ Simulation of one KEM protocol execution using Kyber512:
...
+
+                                                                                    EVP_PKEY_free(key);
+                                                                                    EVP_PKEY_CTX_free(encaps_ctx);
+
+                                                                                    return ret;
+                                                                                  }
 
+                                                                                  // With the helper function above and upon receiving
+                                                                                  // Alice's public key you can encapsulate to get the
+                                                                                  // shared secret and the ciphertext:
```
nebeid, line 221:
```suggestion
                                                                                  // Alice's public key you can encapsulate to generate the
                                                                                  // shared secret and the corresponding ciphertext:
```

crypto/kem/README.md
```diff
@@ -90,137 +90,198 @@ Simulation of one KEM protocol execution using Kyber512:
...
 
-// Generate the key.
-if (!EVP_PKEY_keygen_init(keygen_ctx) ||
-    !EVP_PKEY_keygen(keygen_ctx, &key)) {
-  goto end;
+int get_raw_public_key(/* IN  */ EVP_PKEY *key,
+                       /* OUT */ uint8_t **pub_key,
+                       /* OUT */ size_t *pub_key_len) {
+
+  // We call the function with NULL as the out argument
```
nebeid, line 120:
```suggestion
  // We call the function with NULL as the |out| argument
```

crypto/kem/README.md
```diff
@@ -90,137 +90,198 @@ Simulation of one KEM protocol execution using Kyber512:
...
-                                                                                  OPENSSL_free(ss);
-                                                                                  OPENSSL_free(ct);
-                                                                                  EVP_PKEY_free(key);
-                                                                                  EVP_PKEY_CTX_free(encaps_ctx);
-
-                                                                                  // We don't free the shared secret buffer because the whole point
-                                                                                  // of the protocol is to generate it, so we consider it as the output.
-
-                                                                                  return ret;
+                                                                                  // On |ret| success, |ss| is the generated shared secret you can use,
```
nebeid, line 228:
```suggestion
                                                                                  // On |ret| being SUCCESS, |ss| is the generated shared secret you can use,
```

crypto/kem/README.md
```diff
@@ -90,137 +90,198 @@ Simulation of one KEM protocol execution using Kyber512:
...
+uint8_t *ss = NULL;   // shared secret,
+size_t ss_len;        // and its length.
 
-// We don't free the shared secret buffer because the whole point
-// of the protocol is to generate it, so we consider it as the output.
+int ret = decapsulate(key, ct, ct_len, &ss, &ss_len);
+EVP_PKEY_free(key);  // free if |key| not needed any more.
 
-return ret;
+// On |ret| success, |ss| is the generated shared secret you can use.
```
nebeid, line 284:
```suggestion
// On |ret| being SUCCESS, |ss| is the generated shared secret you can use.
```

nebeid (commented):
Looks very good.

crypto/kem/README.md
dkostic: done, added to lines 25 and 29.

nebeid (approved):

andrewhop (approved):

PR #744, opened 2023-01-05
Fix FIPS static release build on ARM for gcc

samuel40791765:
### Issues:
Resolves `CryptoAlg-1399`

### Description of changes: 
This PR fixes the gcc release build for static FIPS on ARM. This had been working on other compiler combinations for ARM, like gcc debug or clang release/debug.
The following steps were done to fix this:
1. Fix three `delocate.peg` parsing issues.
2. An offset exists at the end of `:lo12:.LC9+8`, which `delocate.go` was choking on. Removing the check gets
around this.
3. The gcc release build does not use `.comm` to define common symbols. This creates a gap in the bss accessor functions, so we use `bl` to catch these accessor function symbols instead.
4. Older versions of the gcc assembler maps variables of the same value to each other, and skips label creation. `delocate.go` relies on these labels to determine if a symbol is "known"; if a label doesn't exist, it's addressed
at the end of `bcm-delocated.S`. Rearranging the variable definition forces a label creation and gets around the
unwanted assembler optimization.

More details can be found in the doc under the SIM.

### Call-outs:
N/A

### Testing:

Turned on tests for gcc release build for static FIPS on ARM.

the ISC license.

### DIFF
--- a/util/fipstools/delocate/delocate.go
+++ b/util/fipstools/delocate/delocate.go
@@ -574,6 +574,7 @@ func (d *delocation) processAarch64Instruction(statement, instruction *node32) (
 						}
 
 						if skipWS(parts.next) != nil || parts.up.next != nil {
+							fmt.Print("\n\n", parts, parts.next, parts.up, parts.up.next, "\n\n")
 							panic("can't handle tweak or post-increment with symbol references")
 						}
 
--- a/util/fipstools/delocate/delocate.go
+++ b/util/fipstools/delocate/delocate.go
@@ -78,6 +78,8 @@ type delocation struct {
 	// bssAccessorsNeeded maps from a BSS symbol name to the symbol that
 	// should be used to reference it. E.g. “P384_data_storage” ->
 	// “P384_data_storage”.
+	// Symbols saved in bssAccessorsNeeded will be rewritten with a "_bss_get"
+	// accessor appended.
 	bssAccessorsNeeded map[string]string
 	// tocLoaders is a set of symbol names for which TOC helper functions
 	// are required. (ppc64le only.)
@@ -488,6 +490,19 @@ func (d *delocation) processAarch64Instruction(statement, instruction *node32) (
 		}
 
 		return d.loadAarch64Address(statement, targetReg, symbol, offset)
+	case "bl":
+		// We were relying on symbols defined with ".comm" to populate bssAccessorsNeeded,
+		// but gcc release does not use ".comm" to define common symbols. The accessor
+		// functions are defined with a ".type $symbol %object" followed with a
+		// ".size $symbol $symbol_size" instead. These definition methods are generic and
+		// do not only apply to symbols that need accessors. Thus we attempt to reverse
+		// engineer the accessor symbols by populating bssAccessorsNeeded with branch labels
+		// that have the accessor "_bss_get" at the suffix.
+		bss_get_symbol := d.contents(argNodes[0])
+		if strings.HasSuffix(bss_get_symbol, "_bss_get") {
+			trimmed_symbol := strings.TrimSuffix(bss_get_symbol, "_bss_get")
+			d.bssAccessorsNeeded[trimmed_symbol] = trimmed_symbol
+		}
 	}
 
 	var args []string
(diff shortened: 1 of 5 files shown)


util/fipstools/delocate/delocate.go
```diff
@@ -488,6 +490,19 @@ func (d *delocation) processAarch64Instruction(statement, instruction *node32) (
 		}
 
 		return d.loadAarch64Address(statement, targetReg, symbol, offset)
+	case "bl":
+		// We were relying on symbols defined with ".comm" to populate bssAccessorsNeeded,
+		// but gcc release does not use ".comm" to define common symbols. The accessor
+		// functions are defined with a ".type $symbol %object" followed with a
+		// ".size $symbol $symbol_size" instead. These definition methods are generic and
+		// do not only apply to symbols that need accessors. Thus we attempt to reverse
+		// engineer the accessor symbols by populating bssAccessorsNeeded with branch labels
```
nebeid, line 499: What does branch labels here mean?

util/fipstools/delocate/delocate.go
```diff
@@ -488,6 +490,19 @@ func (d *delocation) processAarch64Instruction(statement, instruction *node32) (
 		}
 
 		return d.loadAarch64Address(statement, targetReg, symbol, offset)
+	case "bl":
+		// We were relying on symbols defined with ".comm" to populate bssAccessorsNeeded,
+		// but gcc release does not use ".comm" to define common symbols. The accessor
+		// functions are defined with a ".type $symbol %object" followed with a
```
nebeid, line 496: So `%object` was used for data objects and functions that needed `bss_get()`? (If I understand correctly, `DEFINE_METHOD_FUNCTION` adds `bss_get` to function pointers, but I didn't dig into it.)

util/fipstools/delocate/delocate.go
```diff
@@ -488,6 +490,19 @@ func (d *delocation) processAarch64Instruction(statement, instruction *node32) (
 		}
 
 		return d.loadAarch64Address(statement, targetReg, symbol, offset)
+	case "bl":
```
nebeid, line 493: If I understand the solution here:
* we found some `bl` instructions that jumped to accessor functions that have the suffix `_bss_get` as a suffix.
* but these functions were not found because they would have been formed by delocate.go at https://github.com/awslabs/aws-lc/blob/d3cbac9bd889592e83d94c3c16282c320b4d8df4/util/fipstools/delocate/delocate.go#L205 when a `.comm` directive was found and they were added to `bssAccessorsNeeded`.
* so we now find them where they're called in `bl` instructions and add them to that same array.

If this understanding is correct, is `bl` the only instruction they're found with?

util/fipstools/delocate/delocate.go
```diff
@@ -573,8 +588,10 @@ func (d *delocation) processAarch64Instruction(statement, instruction *node32) (
 							panic("Symbol reference outside of ldr instruction")
 						}
 
-						if skipWS(parts.next) != nil || parts.up.next != nil {
-							panic("can't handle tweak or post-increment with symbol references")
+						// The "parts.up.next != nil" check was removed because gcc/release appends an
```
nebeid, line 591:
```suggestion
						// The check for "parts.up.next != nil" check was removed because gcc/release appends an
```

util/fipstools/delocate/delocate.go
```diff
@@ -488,6 +490,19 @@ func (d *delocation) processAarch64Instruction(statement, instruction *node32) (
 		}
 
 		return d.loadAarch64Address(statement, targetReg, symbol, offset)
+	case "bl":
+		// We were relying on symbols defined with ".comm" to populate bssAccessorsNeeded,
+		// but gcc release does not use ".comm" to define common symbols. The accessor
```
nebeid, line 495: * In this comment and the following one, we say `gcc release`, do we need to be more specific about gcc-7, or this was also the case with gcc-11 (or other versions)?
* Would a `release build by gcc` be clearer than `gcc release`?
* I think the PR description and the commit message should mention that these issues were not occurring in the debug build with gcc.
* Is it correct to modify "The accessor functions" to "The symbols requiring accessor functions (i.e. with a suffix`_bss_get`)"

util/fipstools/delocate/delocate.go
```diff
@@ -492,12 +492,13 @@ func (d *delocation) processAarch64Instruction(statement, instruction *node32) (
...
 	case "bl":
 		// We were relying on symbols defined with ".comm" to populate bssAccessorsNeeded,
-		// but gcc release does not use ".comm" to define common symbols. The accessor
-		// functions are defined with a ".type $symbol %object" followed with a
-		// ".size $symbol $symbol_size" instead. These definition methods are generic and
-		// do not only apply to symbols that need accessors. Thus we attempt to reverse
-		// engineer the accessor symbols by populating bssAccessorsNeeded with branch labels
-		// that have the accessor "_bss_get" at the suffix.
+		// but the gcc release build does not use ".comm" to define common symbols. The
+		// symbols requiring accessor functions (i.e. with a suffix_bss_get) are defined
```
nebeid, line 496: I think there is a space missing between "suffix" and "_bss_get"

nebeid (approved):

util/fipstools/delocate/delocate.go
```diff
@@ -573,8 +589,10 @@ func (d *delocation) processAarch64Instruction(statement, instruction *node32) (
 							panic("Symbol reference outside of ldr instruction")
 						}
 
-						if skipWS(parts.next) != nil || parts.up.next != nil {
-							panic("can't handle tweak or post-increment with symbol references")
```
dkostic, line 577: why was it that post-increment couldn't be handled? did you need to change something else as well to make it work?

samuel40791765: This was the only place where I changed to get this aspect working. It's unclear why the check was there in the first place, but it's been there since BoringSSL's first implementation of the delocate.go for ARM.
@nebeid determined that treating this offset as we normally would is safe to do, more details about that analysis can be found in the internal ARM debugging doc.

* https://boringssl-review.googlesource.com/c/boringssl/+/44848/8/util/fipstools/delocate/delocate.go

dkostic (approved):

nebeid (approved):

PR #745, opened 2023-01-06
Add support to delocate.go to parse output from GCC 11 and test with AL2022

andrewhop:
### Issues:
Resolves CryptoAlg-1482

### Description of changes: 
1. Add support for the GCC 11 FIPS static build by handling `pinsrq` instructions
2. Update error message to make it a bit easier to figure out if similar new instructions get misclassified in the future
3. Add AL2022 base image
4. Add AL2022 images with gcc and clang
5. Update CDK to run the regular build and fips build with each compiler (4 new builds)

### Call-outs:
None

### Testing:
Built images locally and ran tests. Working on deploying the CI to my account next.

the ISC license.


ottok:
I have not reviewed the code nor tested it but I +1 that the bug is relevant and it is important for AWS-LC to test every commit on Amazon Linux 2022 and so that downstream projects/users of AWS-LC on any version of Amazon Linux can "just use it" without extra work.

tests/ci/docker_images/linux-x86/amazonlinux-2022_gcc-11x/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+FROM amazonlinux-2022:base
+
+SHELL ["/bin/bash", "-c"]
+
+# gcc 11.3.1 is the latest version versions `yum --showduplicates list gcc`
```
samuel40791765, line 8: Could the gcc version change when we update the docker image?

tests/ci/docker_images/linux-x86/amazonlinux-2022_clang-14x/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+FROM amazonlinux-2022:base
+
+SHELL ["/bin/bash", "-c"]
+
+# clang 14.0.5 is the latest version versions `yum --showduplicates list clang`
```
samuel40791765, line 8: Same version comment here

samuel40791765 (dismissed):

tests/ci/docker_images/linux-x86/amazonlinux-2022_gcc-11x/Dockerfile
andrewhop: Yeah it could, that's why the overall image is 11x. I think the original intention of this comment was to help people keep track of what was being tested. It could be better to just check the image directly, or look at old build output.

samuel40791765: Ah sorry, I should have been more clearer. I'm just curious about the possibility of AL2022 upgrading it's `gcc/clang` default to something like gcc-12x or clang-15x. Or are the default C versions locked down once the OS has been released?

andrewhop: Historically it seems locked down, they might add other versions, but we would need to install them specifically. For Ubuntu they provide `gcc` and various `gcc-X` versions. Right now it looks like AL2022 just has the one of each:
```
bash-5.2# yum --showduplicates list gcc
Last metadata expiration check: 0:01:43 ago on Fri Jan  6 18:34:56 2023.
Installed Packages
gcc.x86_64               11.3.1-2.amzn2022.0.8               @amazonlinux
Available Packages
gcc.x86_64               11.3.1-2.amzn2022.0.8               amazonlinux 
bash-5.2# yum --showduplicates list clang
Last metadata expiration check: 0:01:44 ago on Fri Jan  6 18:34:56 2023.
Available Packages
clang.x86_64              14.0.5-2.amzn2022.0.5               amazonlinux
bash-5.2# 

```

tests/ci/docker_images/linux-x86/amazonlinux-2022_base/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    valgrind-devel \
+    unzip && \
+    # Based on https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-linux.html
+    curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
+    unzip awscliv2.zip && \
+    ./aws/install --bin-dir /usr/bin && \
+    rm -rf awscliv2.zip aws/ && \
+    mkdir -p ${DEPENDENCIES_DIR} && \
+    cd ${DEPENDENCIES_DIR} && \
+    git clone https://github.com/llvm/llvm-project.git --branch llvmorg-15.0.6  --depth 1 && \
```
sdkellyAWS, line 41: Why does this require llvm? Also why version 15 when the al2022 version of clang is 14?

andrewhop: To run the memory sanitizer without false positives from inside the C standard functions we need to compile libc as a [part of our build](https://github.com/awslabs/aws-lc/blob/main/CMakeLists.txt#L705-L772). I didn't think about the clang 14 vs llvm 15. This only needs the libcxx and libcxxabi parts, but it makes more sense to use the matching set.

andrewhop: Turns out nothing was using this and trying to add the sanitizer results in C++11 vs 17 version issues. I'll tackle this all later.

sdkellyAWS (approved):

samuel40791765 (approved):

PR #755, opened 2023-01-13
Fix FIPS static release build on x86_64 for gcc-8,11,12 on ubuntu

samuel40791765:
### Issues:
Resolves `CryptoAlg-1489`

### Description of changes: 
This PR fixes the gcc release build for static FIPS on Ubuntu 20.04/22.04. gcc-8, 11, and 12 were failing due to some unrecognized instructions.

### Call-outs:
1. `notrack jmp	*%rax` was failing, because there were no commas between the `InstructionArg`.
2. `.long	 1f - 0f` was failing because `delocate.go` was incorrectly recognizing `1f` as an offset, when it should be a `LocalLabelRef`. Our `Offset` rule was too lax and recognized all number references as "offsets", regardless of what was appended at the end. I've added `![[A-Z]` to the end of the `Offset` rule to reject any additional charactors.
3. `gcc-8` The gcc-8 assembler will attempt to optimize function pointers used in multiple places under a `.data.rel.ro.local` section, but `delocate.go` does not have the ability to handle `.data` sections. It’s unclear why this is specific to only gcc-8, but we speculate that this was unwanted behavior and was removed in subsequent versions of the gcc compiler. We get around this by defining `pkey_pss_init` as two separate functions for signing and verifying.

### Testing:
New CI for all these dimensions

the ISC license.


samuel40791765:
Formal verification seems to be failing due to the static function naming change.
I'll work with the formal verification team to resolve this after the contents of this PR are approved.

dkostic (approved):

nebeid (approved):

PR #756, opened 2023-01-17
Add a way to allow parsing RSA keys with d > n

dkostic:
Some services are dealing with RSA keys that have been inappropriately generated -- the private exponent |d| is greater than the modulus |n|. Until such keys are eradicated, we _temporarly_ add a way to relax the requirements for checking RSA keys such that the condition |d < n| can be skipped. This "relaxed" behavior has to be explicitly enabled by the user by calling |allow_rsa_keys_d_gt_n(true)|. The default behavior is still to check if |d < n| and fail if not true.

### Issues:
N/A

### Description of changes: 

### Call-outs:

### Testing:

the ISC license.


include/openssl/rsa.h
```diff
@@ -800,6 +800,18 @@ struct rsa_st {
   unsigned private_key_frozen:1;
 };
 
+#if !defined(AWSLC_FIPS)
+#include <stdbool.h>
+// Some services are dealing with RSA keys that have been inappropriately
```
torben-hansen, line 805:
```suggestion
// Some RSA keys are inappropriately
```

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -742,6 +742,43 @@ static int check_mod_inverse(int *out_ok, const BIGNUM *a, const BIGNUM *ainv,
   return ret;
 }
 
+#if !defined(AWSLC_FIPS)
+// Some services are dealing with RSA keys that have been inappropriately
```
torben-hansen, line 746: No need to state this twice.

include/openssl/rsa.h
```diff
@@ -800,6 +800,18 @@ struct rsa_st {
...
+// generated -- the private exponent |d| is greater than the modulus |n|.
+// Until such keys are eradicated, we _temporarly_ add a way to relax the
+// requirements for validating RSA keys such that the condition |d < n|
+// can be skipped. This "relaxed" behavior has to be explicitly enabled by
+// the user by calling |allow_rsa_keys_d_gt_n(true)|. The default behavior
+// is still to check if |d < n| and fail if not true.
+OPENSSL_EXPORT void allow_rsa_keys_d_gt_n(bool enable);
```
torben-hansen, line 812: IMO this should already by marked as deprecated.

torben-hansen (commented):
I don't think this is a thread-safe implementation:
* Thread 1: wants to allow the verification and enables it explicitly. 
* Thread 2: doesn't want to allow the verification and disables it explicitly.

Consider the linear execution:
Thread 1 enables -> Thread 2 disables -> Thread 1 parses RSA key -> Thread 1 parsing RSA key fails.

include/openssl/rsa.h
dkostic: done

crypto/fipsmodule/rsa/rsa.c
dkostic: done

include/openssl/rsa.h
dkostic: done

dkostic:
> I don't think this is a thread-safe implementation:

you are right. I simplified the implementation as discussed offline.

torben-hansen (approved):

PR #762, opened 2023-01-18
Zeroise private random data on the stack before returning

torben-hansen:
CryptoAlg-1524

### Description of changes: 

Zeroise private random data immediately after use. This is just good secret value hygiene.

Most is stack-allocated. So, good changes it would be overwritten fairly quickly anyway. Looked in both `/crypto` and `/ssl`, grepping my way through.

### Testing:

CI

the ISC license.

### DIFF
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -1996,7 +1996,7 @@ void ED25519_keypair_from_seed(uint8_t out_public_key[32],
   ge_p3_tobytes(out_public_key, &A);
 
   OPENSSL_memcpy(out_private_key, seed, ED25519_SEED_LEN);
-  OPENSSL_memcpy(out_private_key + 32, out_public_key, 32);
+  OPENSSL_memcpy(out_private_key + ED25519_SEED_LEN, out_public_key, 32);
 }
 
 
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -1874,10 +1874,13 @@ static void sc_muladd(uint8_t *s, const uint8_t *a, const uint8_t *b,
   s[31] = s11 >> 17;
 }
 
-void ED25519_keypair(uint8_t out_public_key[32], uint8_t out_private_key[64]) {
-  uint8_t seed[32];
-  RAND_bytes(seed, 32);
+#define ED25519_SEED_LEN 32
+
+void ED25519_keypair(uint8_t out_public_key[ED25519_SEED_LEN], uint8_t out_private_key[64]) {
+  uint8_t seed[ED25519_SEED_LEN];
+  RAND_bytes(seed, ED25519_SEED_LEN);
   ED25519_keypair_from_seed(out_public_key, out_private_key, seed);
+  OPENSSL_cleanse(seed, ED25519_SEED_LEN);
 }
 
 int ED25519_sign(uint8_t out_sig[64], const uint8_t *message,
@@ -1982,9 +1985,9 @@ int ED25519_verify(const uint8_t *message, size_t message_len,
 
 void ED25519_keypair_from_seed(uint8_t out_public_key[32],
                                uint8_t out_private_key[64],
-                               const uint8_t seed[32]) {
+                               const uint8_t seed[ED25519_SEED_LEN]) {
   uint8_t az[SHA512_DIGEST_LENGTH];
-  SHA512(seed, 32, az);
+  SHA512(seed, ED25519_SEED_LEN, az);
 
   az[0] &= 248;
   az[31] &= 127;
@@ -1994,7 +1997,7 @@ void ED25519_keypair_from_seed(uint8_t out_public_key[32],
   x25519_ge_scalarmult_base(&A, az);
   ge_p3_tobytes(out_public_key, &A);
 
-  OPENSSL_memcpy(out_private_key, seed, 32);
+  OPENSSL_memcpy(out_private_key, seed, ED25519_SEED_LEN);
   OPENSSL_memcpy(out_private_key + 32, out_public_key, 32);
 }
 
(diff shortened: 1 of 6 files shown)


crypto/curve25519/curve25519.c
```diff
@@ -1874,10 +1874,13 @@ static void sc_muladd(uint8_t *s, const uint8_t *a, const uint8_t *b,
   s[31] = s11 >> 17;
 }
 
-void ED25519_keypair(uint8_t out_public_key[32], uint8_t out_private_key[64]) {
-  uint8_t seed[32];
-  RAND_bytes(seed, 32);
+#define ED25519_SEED_LEN 32
```
justsmth, line 1877: Should this be in the header file alongside the other related constants? https://github.com/awslabs/aws-lc/blob/736fce4b11417ea07ae0df06913d7e87a3cd58ca/include/openssl/curve25519.h#L71

torben-hansen: That's better.

justsmth (dismissed):

crypto/curve25519/curve25519.c
```diff
@@ -1982,9 +1983,9 @@ int ED25519_verify(const uint8_t *message, size_t message_len,
 
 void ED25519_keypair_from_seed(uint8_t out_public_key[32],
                                uint8_t out_private_key[64],
-                               const uint8_t seed[32]) {
+                               const uint8_t seed[ED25519_SEED_LEN]) {
   uint8_t az[SHA512_DIGEST_LENGTH];
-  SHA512(seed, 32, az);
+  SHA512(seed, ED25519_SEED_LEN, az);
```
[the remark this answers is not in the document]
torben-hansen: Agree - but I didn't want to change those at the same time. Another time.

ssl/ssl_key_share.cc
```diff
@@ -203,11 +203,16 @@ class CECPQ2KeyShare : public SSLKeyShare {
...
+    }
+
     HRSS_public_key hrss_public_key;
-    RAND_bytes(hrss_entropy, sizeof(hrss_entropy));
+    RAND_bytes(hrss_entropy.data(), hrss_entropy.size());
     if (!HRSS_generate_key(&hrss_public_key, &hrss_private_key_,
-                           hrss_entropy)) {
+                           hrss_entropy.data())) {
       return false;
     }
```
justsmth, line 217: Did you intend to include this change in this PR?  It seems unrelated to zeroize'ing the ED25519 seed.

torben-hansen: Yes, I repurposed this PR :)

andrewhop (commented):
Did you consider using sizeof in the calls to OPENSSL_cleanse to ensure all the data is cleansed as expected? If a future change made one of these arrays smaller it would trigger an address sanitizer error as cleanse tries to cleanse past the end, but if the array gets bigger the cleanse would only cleanse the first part.

Also this change introduces an inconsistencies: some places use a new constant (ED25519_SEED_LEN) while others continue to use magic numbers.

crypto/curve25519/curve25519.c
```diff
@@ -1994,7 +1995,7 @@ void ED25519_keypair_from_seed(uint8_t out_public_key[32],
   x25519_ge_scalarmult_base(&A, az);
   ge_p3_tobytes(out_public_key, &A);
 
-  OPENSSL_memcpy(out_private_key, seed, 32);
+  OPENSSL_memcpy(out_private_key, seed, ED25519_SEED_LEN);
   OPENSSL_memcpy(out_private_key + 32, out_public_key, 32);
```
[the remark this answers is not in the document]
justsmth: There's a "slippery slope" concern here in that it's not clear which/how-many of the *magic numbers* (32 and 64) in the logic here should be replaced by their corresponding constants (ED25519_SEED_LEN, ED25519_PUBLIC_KEY_LEN, ED25519_PRIVATE_KEY_LEN and ED25519_SIGNATURE_LEN) for the current PR.

I think a reasonable line to draw is to replace "32" with "ED25519_SEED_LEN" wherever appropriate, but otherwise leave the other magic numbers alone for now.
```
  OPENSSL_memcpy(out_private_key, seed, ED25519_SEED_LEN);
  OPENSSL_memcpy(out_private_key + ED25519_SEED_LEN, out_public_key, 32);
```

crypto/fipsmodule/rand/rand.c
```diff
@@ -392,6 +392,8 @@ void RAND_bytes_with_additional_data(uint8_t *out, size_t out_len,
       CRYPTO_STATIC_MUTEX_unlock_write(thread_states_list_lock_bss_get());
     }
 #endif
+    OPENSSL_cleanse(seed, CTR_DRBG_ENTROPY_LEN);
+    OPENSSL_cleanse(personalization, CTR_DRBG_ENTROPY_LEN);
```
justsmth, line 396: Are we concerned at all about the performance implications of these additional operations? How much will this impact the latency of functions calling `RAND_bytes`?

justsmth (changes requested):

crypto/fipsmodule/rand/rand.c
torben-hansen: No issue, these are all (except one) during seeding.

torben-hansen:
> Did you consider using sizeof in the calls to OPENSSL_cleanse to ensure all the data is cleansed as expected?

There is an equal issue here if the array is made a pointer that is dynamically allocated. Then `sizeof` would be bad. So, I think just sticking to current implementation. Let me know if you have other concerns.

justsmth (approved):

andrewhop (approved):

torben-hansen:
Zeroise private random data immediately after use. This is just good secret value hygiene.

PR #763, opened 2023-01-19
Fuzzing test cases for PKCS#8 v2 support

skmcgrail:
### Description of changes: 
Adds fuzzing coverage for the PKCS#8 v2 support. The inputs were seeded from the associated EVP_KEY tests cases added in the prior PKCS#8 pull request. I have broken up the fuzzing between pkcs8 and pkcs8_v2 to ensure the fuzzer can reach `EVP_marshal_private_key_v2` rather then it potentially being unreachable if it was in a combined function testing both `EVP_marshal_private_key` and `EVP_marshal_private_key_v2`.

the ISC license.


fuzz/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ fuzzer(privkey)
 fuzzer(cert)
 fuzzer(spki)
 fuzzer(pkcs8)
```
samuel40791765, line 23: Not sure if we should rename this to `pkcs8_v1`, but the current version does correspond with `EVP_marshal_private_key`/`EVP_marshal_private_key_v2`

samuel40791765 (approved):

andrewhop (approved):

fuzz/CMakeLists.txt
skmcgrail: I'm going to bias towards leaving it the same for now to stay consistent with BoringSSL upstream. Can revisit later if needed.

PR #764, opened 2023-01-23
Fix/build on openbsd

knightjoel:
### Issues:
n/a

### Description of changes: 
This PR enables successful builds on OpenBSD. 

### Call-outs:
- Building googletest with `-Wvla` on OpenBSD throws an error. I'm not sure what an appropriate fix is here or if it's even worth it (how critical is getting the thread count?). As a starting point, I've just removed the offending code.

### Testing:
Test builds done on OpenBSD 7.2/amd64. Also verified build is still successful on FreeBSD 13/amd64.


include/openssl/thread.h
```diff
@@ -59,6 +59,10 @@
 
 #include <sys/types.h>
 
+#if defined(__OpenBSD__)
+#   include <pthread.h>
+#endif
```
PiotrSikora, line 64: Nit: This should be moved inside the `!defined(__GLIBC__)` block to respect `OPENSSL_THREADS`.

include/openssl/thread.h
```diff
@@ -59,6 +59,10 @@
 
 #include <sys/types.h>
 
+#if defined(__OpenBSD__)
```
PiotrSikora, line 62: Nit: This should be moved to `include/openssl/base.h` and defined as `OPENSSL_OPENBSD`, which should be used here instead. This is how it's done for other operating systems (e.g. FreeBSD).

torben-hansen:
What's the entire log output for?
```
unknown type name 'pthread_rwlock_t
```

knightjoel:
> What's the entire log output for?
> 
> ```
> unknown type name 'pthread_rwlock_t
> ```

```
In file included from /home/joel/tmp/aws-lc/crypto/fipsmodule/cpucap/cpucap.c:12:
In file included from /home/joel/tmp/aws-lc/crypto/../include/openssl/cpu.h:18:
In file included from /home/joel/tmp/aws-lc/crypto/../include/openssl/crypto.h:27:
/home/joel/tmp/aws-lc/crypto/../include/openssl/thread.h:81:9: error: unknown type name 'pthread_rwlock_t'
typedef pthread_rwlock_t CRYPTO_MUTEX;
        ^
1 error generated.
```

torben-hansen:
I not sure `GetThreadCount()` is that valuable, it checks whether there is actually more threads. Newer versions don't have this functions I think.

include/openssl/thread.h
torben-hansen: Agree.

This leaks `<pthread.h>` everywhere since `thread.h` is a public header... Not sure I like that...

knightjoel: Oof, I hadn't considered that. Thanks for the pointer. How does the updated diff look to you?

include/openssl/thread.h
knightjoel: Updated the diff. Please let me know how close/far I am to what you were expecting.

PiotrSikora (commented):
Updated diff LGTM for the `include/`.

As for `third_party/googletest`, it looks that it wasn't updated in 2 years, perhaps it's easier to update it than patch locally (assuming it's still used in AWS-LC)?

knightjoel:
> Updated diff LGTM for the `include/`.

Thanks 👍 
 
> As for `third_party/googletest`, it looks that it wasn't updated in 2 years, perhaps it's easier to update it than patch locally (assuming it's still used in AWS-LC)?

Upstream, that bit of code is the same as what's in aws-lc. Upstream doesn't compile with `-Wvla` and so it's unaffected. I considered sending a patch upstream, but they don't have much incentive to fix this given it doesn't directly affect them.

torben-hansen (dismissed):

torben-hansen:
Note that there is no CI dimension for OpenBSD, so this is **experimental**.

knightjoel:
> Note that there is no CI dimension for OpenBSD, so this is **experimental**.

Would having CI for OpenBSD be of interest? It would have to run as a GitHub action given there is no support for OpenBSD in CodeBuild.

torben-hansen (approved):

andrewhop (approved):

knightjoel:
Hi folks, would someone be able to merge this? Please and thank you.

PR #769, opened 2023-01-24
Update p_kyber512_asn1.c OID

jakemas:
### Issues:
Incorrect encoding of the OID used for kyber.

### Description of changes: 
`0x06, 0x09` should not be included in the encoding as 0x06 is used to signify that this is an object identifier and the second byte to indicate the total length. Since the next value `11` was used to describe the total length of this OID, this has also been modified to `9`.

### Call-outs:
See 8.19 Encoding of an object identifier value in [X.690 : Information technology - ASN.1 encoding rules: Specification of Basic Encoding Rules (BER), Canonical Encoding Rules (CER) and Distinguished Encoding Rules (DER)](https://www.itu.int/rec/T-REC-X.690-202102-I/en).

the ISC license.


torben-hansen (approved):

dkostic:
this is changing a file that is part of the obsolete KEM implementation and will be deleted soon. The new place for OIDs is here: https://github.com/awslabs/aws-lc/blob/main/crypto/kem/kem.c#L22. Although, I'm not sure why do we need an OID for Kyber at all? Is the OID used for anything except key encoding/decoding (which we don't support for KEMs)?

andrewhop:
This was deleted in https://github.com/awslabs/aws-lc/pull/738.

PR #785, opened 2023-02-06
Update ASN.1 privkey encode/decode based upon IETF spec

jakemas:
### Description of changes: 

Modifying the ASN.1 encoding/decoding structure to match upcoming [new release -01 of IETF draft](https://datatracker.ietf.org/doc/draft-ietf-lamps-dilithium-certificates/).

### Testing:
A full test of ASN.1 encode/decode/extraction is performed within the test [Dilithium3Test, Encoding.](https://github.com/jakemas/aws-lc/blob/integrate-pq/crypto/dilithium/p_dilithium_test.cc#L167-L241)

the ISC license.


dkostic:
how come this change doesn't affect any test? Don't we test the encoding/decoding?

PR #787, opened 2023-02-07
Update ssm document to handle s3 files

samuel40791765:
### Issues:
Addresses `CryptoAlg-1423`

### Description of changes: 
We need our Codebuild dimension to be able to handle source code from S3. This lets us easily test internal fixes.

### Call-outs:
N/A

### Testing:
Tested by following the internal runbook and running against the source code from s3. 

the ISC license.


billbo-yang (approved):
Offline correspondence copied here for reference:

**Bill Yang**
For your PR, how exactly is the codebuild instance started up by the S3?
Is it still part of the GitHub CI or is there another version hosted somewhere else?

**Samuel Chiang**
Ah yes, we have this runbook to test internal fixes on our CI, by uploading artifacts to test from S3 -> Codebuild.
It’s made to address Security fixes we can only test internally, and can’t make public on our Github repo.

**Bill Yang**
Oh ok neat - and I assume in this case the {SOURCE} env variable will point to the S3 bucket where the code is held? (edited) 

**Samuel Chiang**
Yeah I replace {SOURCE} with `CODEBUILD_SOURCE_REPO_URL`([link](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html)). The variable is different, depending on whether the source is from S3 or Github.

skmcgrail (approved):

PR #800, opened 2023-02-10
HKDF_expand CAVP tests

Taffer:
### Issues:
Resolves CryptoAlg-1436

### Description of changes: 
Adds CAVP KATs for using `HKDF_expand()` as a "KDF in Feedback mode" for NIST SP 800-108r1.

### Call-outs:
This code was previously being unit tested as part of the HKDF unit tests.

### Testing:
This adds about 150ms to the HKDF tests on my laptop.

### DIFF
--- a/crypto/fipsmodule/hkdf/hkdf_test.cc
+++ b/crypto/fipsmodule/hkdf/hkdf_test.cc
@@ -437,7 +437,7 @@ static void RunTest(FileTest *t)
   // attributes.
   bool run_test = true;
 
-  if ((ctrlocation != "AFTER_ITER") && (rlen != "32_BITS")) {
+  if ((ctrlocation != "AFTER_FIXED") || (rlen != "8_BITS")) {
     // HKDF_expand() only works for "Counter After" and an r length of 32-bits.
     // Skip this test.
     run_test = false;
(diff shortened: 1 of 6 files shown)


Taffer:
This looks like it's coming from the wrong branch on my end, but it's the right patch, soooooo...

samuel40791765 (approved):

crypto/fipsmodule/hkdf/hkdf_test.cc
```diff
@@ -416,3 +416,162 @@ TEST(HKDFTest, WycheproofSHA512) {
...
+  ASSERT_TRUE(t->GetInstruction(&rlen, "RLEN"));
+
+  // Each test passing through RunTest() needs to consume all of its attributes
+  // or it generates an error. There is no IgnoreAllAttributes() or SkipTest()
+  // that we could take advantage of to return early instead of eating all the
+  // attributes.
+  bool run_test = true;
+
+  if ((ctrlocation != "AFTER_FIXED") || (rlen != "8_BITS")) {
+    // HKDF_expand() only works for "Counter After" and an r length of 32-bits.
```
dkostic, line 441: the comment says 32 bits, the code checks `rlen != 8_BITS`, what's the catch?

crypto/fipsmodule/hkdf/hkdf_test.cc
```diff
@@ -416,3 +416,162 @@ TEST(HKDFTest, WycheproofSHA512) {
...
+  t->IgnoreAttribute("K(i)/35");
+  t->IgnoreAttribute("K(i)/36");
+  t->IgnoreAttribute("K(i)/37");
+  t->IgnoreAttribute("K(i)/38");
+
+  unsigned long l_len = std::stoul(l_str) / 8;
+
+  // HKDF_expand being used as a KDF in Feedback Mode
+  if (run_test) {
+    uint8_t *output = static_cast<uint8_t *>(new uint8_t[l_len]);
```
dkostic, line 565: if we use `std::vector` instead of `uint8_t*` do we have to explicitly free the memory in the end like we do here with `delete`?

crypto/fipsmodule/hkdf/hkdf_test.cc
```diff
@@ -416,3 +416,162 @@ TEST(HKDFTest, WycheproofSHA512) {
...
+
+    ASSERT_TRUE(HKDF_expand(output, l_len, md, ki.data(), ki.size(), fixed.data(), fixed.size()));
+    EXPECT_EQ(Bytes(ko.data(), ko.size()), Bytes(output, l_len));
+
+    delete[] output;
+    output = NULL;
```
dkostic, line 571: maybe remove this because it generates warnings "unused variable" warnings (output is assigned a value but never used, annoying but better to avoid the warnings...)

Taffer:
I think I created this PR in error, it's not working properly yet.

PR #802, opened 2023-02-11
Add .pdb debug symbols to Windows build

billbo-yang:
### Issues:
N/A

### Description of changes: 
When evaluating the Windows FIPS build, uses debug symbols as part of their process. However, AWS-LC currently doesn't output the necessary debug symbols for them to conduct their work. This change enables the debug linker option that allows debug symbols to be created with the build.

### Call-outs:
N/A

### Testing:
Tested by ensuring the build works and produces the correct output on Windows, and that it still builds correctly on MacOS and Ubuntu.

the ISC license.


torben-hansen:
But https://github.com/awslabs/aws-lc/blob/main/CMakeLists.txt#L1
and `target_link_options` is only available from cmake version 3.13.

andrewhop (commented):
Should we only add this option if the build type is debug or release with debug info? Is it possible to move this configuration to a central location?

billbo-yang:
Closing again to address comments and do further testing

PR #803, opened 2023-02-11
Add basic OCSP functions and structures needed for OCSP requests

samuel40791765:
### Issues:
Addresses `CryptoAlg-1536`

### Description of changes: 
Import the OCSP structures and ASN.1 functions needed to support OCSP request sending. Also add two basic functions 
for `OCSP_REQ_CTX`, `OCSP_REQ_CTX_new` and `OCSP_REQ_CTX_free`.

Taken from: 
* https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/ocsp/ocsp_asn.c

### Call-outs:
This is the basis for importing OCSP functions needed to handle OCSP requests.

### Testing:
These ASN.1 functions are built on existing ASN.1 functionality. We will add fuzz testing on some of these APIs soon.

the ISC license.


crypto/ocsp/internal.h
```diff
@@ -24,6 +24,28 @@ struct ocsp_cert_id_st {
   ASN1_INTEGER *serialNumber;
 };
 
+//   Request ::=     SEQUENCE {
```
andrewhop, line 27: Please add a link to the RFC that define these structures.

crypto/ocsp/internal.h
```diff
@@ -24,6 +24,28 @@ struct ocsp_cert_id_st {
...
+//       reqCert                    CertID,
+//       singleRequestExtensions    [0] EXPLICIT Extensions OPTIONAL }
```
andrewhop, line 29: What does the EXPLICIT and OPTIONAL mean in this definition?

crypto/ocsp/internal.h
```diff
@@ -35,6 +57,25 @@ struct ocsp_signature_st {
   STACK_OF(X509) *certs;
 };
 
+//   OCSPRequest     ::=     SEQUENCE {
+//       tbsRequest                  TBSRequest,
+//       optionalSignature   [0]     EXPLICIT Signature OPTIONAL }
+//
+struct ocsp_request_st {
+    OCSP_REQINFO tbsRequest;
+    OCSP_SIGNATURE *optionalSignature; /* OPTIONAL */
```
andrewhop, line 66: Is there something I should check with this `/* OPTIONAL */` that's different than requestExtensions on line 46?

include/openssl/ocsp.h
```diff
@@ -48,8 +53,15 @@ DECLARE_ASN1_FUNCTIONS(OCSP_RESPDATA)
...
 DECLARE_ASN1_FUNCTIONS(OCSP_RESPBYTES)
+DECLARE_ASN1_FUNCTIONS(OCSP_ONEREQ)
 DECLARE_ASN1_FUNCTIONS(OCSP_CERTID)
+DECLARE_ASN1_FUNCTIONS(OCSP_REQUEST)
 DECLARE_ASN1_FUNCTIONS(OCSP_SIGNATURE)
+DECLARE_ASN1_FUNCTIONS(OCSP_REQINFO)
+
+OCSP_REQ_CTX *OCSP_REQ_CTX_new(BIO *io, int maxline);
+
+void OCSP_REQ_CTX_free(OCSP_REQ_CTX *rctx);
```
andrewhop, line 64: Missing API documentation

crypto/ocsp/ocsp_asn.c
```diff
@@ -22,6 +29,30 @@ ASN1_SEQUENCE(OCSP_CERTID) = {
...
+                             X509_EXTENSION, 0)
+} ASN1_SEQUENCE_END(OCSP_ONEREQ)
+
+IMPLEMENT_ASN1_FUNCTIONS(OCSP_ONEREQ)
+
+ASN1_SEQUENCE(OCSP_REQINFO) = {
+    ASN1_EXP_OPT(OCSP_REQINFO, version, ASN1_INTEGER, 0),
+    ASN1_EXP_OPT(OCSP_REQINFO, requestorName, GENERAL_NAME, 1),
+    ASN1_SEQUENCE_OF(OCSP_REQINFO, requestList, OCSP_ONEREQ),
+    ASN1_EXP_SEQUENCE_OF_OPT(OCSP_REQINFO, requestExtensions, X509_EXTENSION, 2)
```
andrewhop, line 44: From the ASN.1 definition what do the 0 and 2 do here vs line 35?
* singleRequestExtensions    [0] EXPLICIT Extensions OPTIONAL
* requestExtensions   [2] EXPLICIT Extensions OPTIONAL

samuel40791765: These are context-specific ASN.1 TAGS to unambiguously identify certain components in an encoded stream. These context specific tags are used to eliminate ambiguity in a `ASN.1 SEQUENCE` which contains `OPTIONAL` elements.
"If you know you are receiving two items that are not optional, one after the other, you know which is which even if their tags are the same. However, if the first one is optional, the two must have different tags, or you would not be able to tell which one you had received if only one was present in the encoding."

* reference: https://stackoverflow.com/a/15071901

skmcgrail: https://luca.ntop.org/Teaching/Appunti/asn1.html is a helpful resource that I liked when working on the pkcs#8 pieces.

The `0` and `1` get encoded into the lower 6 bits of the (T)ag portion of the TLV of the DER encoding. Then as indicated earlier the explicit controls whether the underlying type's tag information is then encoded into the V portion as a nested TLV pair.

skmcgrail (approved):

samuel40791765:
Closing in favor of: https://github.com/aws/aws-lc/pull/810

PR #813, opened 2023-02-16
Add set up of OCSP_REQ_CTX with OCSP_REQUEST

samuel40791765:
### Issues:
Addresses `CryptoAlg-1536`

### Description of changes: 
Add functionality for setting up `OCSP_REQ_CTX` with an `OCSP_REQUEST`. `OCSP_REQ_CTX` sets up the OCSP request to be sent over HTTP.

Taken from: 

### Call-outs:
N/A

### Testing:
Generated OCSP Request DER files with the openssl ocsp tool, and added generation method to the existing README. We mostly test if valid request files are parsable, but we have an intentional failing case at the end.
Also added a test grabbing the internal contents of the |OCSP_REQ_CTX|, and verify that both construction methods have the same output. There is also a check that the constructed http header is there as expected, along with the length of the OCSP Request contents.

the ISC license.


crypto/ocsp/ocsp_test.cc
```diff
@@ -680,3 +688,54 @@ TEST_P(OCSPTest, VerifyOCSPResponse) {
...
+
+  std::string data = GetTestData(std::string("crypto/ocsp/test/aws/" +
+                                std::string(t.ocsp_request) + ".der").c_str());
+  std::vector<uint8_t> ocsp_request_data(data.begin(), data.end());
+
+  bssl::UniquePtr<BIO> bio(BIO_new(BIO_s_mem()));
+  bssl::UniquePtr<OCSP_REQUEST> ocspRequest =
+      LoadOCSP_REQUEST(ocsp_request_data);
+
+  if(t.expected_status) {
```
andrewhop, line 722: To help readability of which status is 0 vs 1:
```suggestion
  if(t.expected_status == OCSP_RESQUEST_PARSE_SUCCESS) {
```

crypto/ocsp/ocsp_test.cc
```diff
@@ -680,3 +688,54 @@ TEST_P(OCSPTest, VerifyOCSPResponse) {
...
+                                                nullptr, ocspRequest.get(), 0));
+    ASSERT_TRUE(ocspReqCtx);
+
+    // Set up |OCSP_REQ_CTX| without a |OCSP_REQUEST|. Then finalize the context
+    // later.
+    bssl::UniquePtr<OCSP_REQ_CTX> ocspReqCtxLater(OCSP_sendreq_new(bio.get(),
+                                                nullptr, nullptr, 0));
+    ASSERT_TRUE(ocspReqCtx);
+    ASSERT_TRUE(OCSP_REQ_CTX_set1_req(ocspReqCtxLater.get(),
+                                      ocspRequest.get()));
```
andrewhop, line 736: Is it possible to test that this finalized request context is finalized in the way we expect? Or that the content is sane?

crypto/ocsp/ocsp_http.c
```diff
@@ -57,3 +61,64 @@ void OCSP_REQ_CTX_free(OCSP_REQ_CTX *rctx)
...
+    return rctx;
+
+ err:
+    OCSP_REQ_CTX_free(rctx);
+    return NULL;
+}
+
+int OCSP_REQ_CTX_http(OCSP_REQ_CTX *rctx, const char *op, const char *path)
+{
+    static const char http_hdr[] = "%s %s HTTP/1.0\r\n";
```
andrewhop, line 90: This is a windows line ending, will this impact linux clients/servers?

crypto/ocsp/ocsp_http.c
```diff
@@ -16,12 +16,16 @@
...
+// OCSP_REQ_CTX states
 
 // If set no reading should be performed
 #define OHS_NOREAD              0x1000
 // Error condition
 #define OHS_ERROR               (0 | OHS_NOREAD)
+// First call: ready to start I/O
+#define OHS_ASN1_WRITE_INIT     (5 | OHS_NOREAD)
+// Headers set, no final \r\n included
+#define OHS_HTTP_HEADER         (9 | OHS_NOREAD)
```
andrewhop, line 28: What do each bit of this flag mean? It looks a little weird in binary:
* OHS_NOREAD -> 0b1000000000000
* OHS_ERROR -> 0b1000000000000
* OHS_ASN1_WRITE_INIT -> 0b1000000000101
* OHS_HTTP_HEADER -> 0b1000000001001

Why skip 1 and 2? Why set 5 and 9 which is two bits instead of 4 and 8? Why is noread 4096?

andrewhop (commented):
General comment for all the error cases in this PR: do they deserve new error codes?

crypto/ocsp/ocsp_http.c
samuel40791765: These flags were taken from [here](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/ocsp/ocsp_ht.c#L39-L60). The list should provide a better picture of everything. I'm skipping the ones that aren't needed for the time being, but all of them are going to eventually be pulled. 
`OCSP_REQ_CTX_nbio` (next PR) uses these flags to determine the read/write state of |OCSP_REQ_CTX|.  `OHS_NOREAD` is a shared state across flags, which `OCSP_REQ_CTX_nbio` uses to check when to start writing to the memory BIO, so it has a more significant bit .

samuel40791765 (commented):
> General comment for all the error cases in this PR: do they deserve new error codes?

The error cases that occur in this PR are either related to basic BIO writing or ASN.1 parsing. Both of these throw a clear error code if anything goes wrong, so they might not warrant a new one on top.

crypto/ocsp/test/aws/OCSP-TEST.md
```diff
@@ -28,6 +28,26 @@ The leaf cert/key. OCSP responses will be generated for this cert.
 
 DER formatted OCSP response for the Server Cert.
 
+## Generating a OCSP request for the leaf cert
+```
+openssl ocsp -issuer ca_cert.pem -cert server_cert.pem -reqout ocsp_request.der
```
andrewhop, line 33: I tried running this in `aws-lc/crypto/ocsp/test` and it failed with:
> Could not open file or uri for loading issuer certificate from ca_cert.pem

I see the readme mention ca_cert.pem, but it's not checked in and I can't find instructions to generate it

andrewhop (approved):

crypto/ocsp/ocsp_http.c
skmcgrail: The `'\r\n` is part of the HTTP standard, this looks like this is writing out the initial request line. https://www.rfc-editor.org/rfc/rfc2616#section-5.1

skmcgrail: I still question why we need to be doing anything with HTTP in our library though? Shouldn't it be enough to provide serialization / de-serialization of the ASN1 messages?

There are two layers of OCSP:
1. There is the OCSP "protocol messages" which are request/response structure messages defined using ASN.1 syntax that are DER-encoded.
2. There are transports (for example HTTP) for which these protocol messages can be sent over. When using those transports there is defined behavior of how the message should be sent. (GET vs POST behavior etc). The transport mechanism for HTTP is defined in the Appendix of the RFC, but as the RFC states technically any transport could be used. [OCSP over HTTP](https://datatracker.ietf.org/doc/html/rfc6960#page-30)

I certainly think part 1 is the most difficult item to get right, and the one we should be providing APIs for. I question why we need to be doing item number two though. It seems crazy to me we are also synthesizing the raw http request bytes. Feels like that should be up to the application higher up in the stack to handle? Number two isn't what is difficult to implement OCSP for, the difficulty is in bullet point number one. It's a weird surface area to be adding to our library.

crypto/ocsp/test/aws/OCSP-TEST.md
samuel40791765: Ah the file `ca_cert.pem` is under `aws-lc/crypto/ocsp/test/aws`. I placed it under the same directory as this `README` :). It's from our first iteration of OCSP work last year.

andrewhop (approved):

skmcgrail (approved):

crypto/ocsp/ocsp_http.c
samuel40791765: Yeah I completely agree. I was hoping we could be focusing on part 1 for AWS-LC's OCSP implementation, and it initially was during our previous iterations of OCSP. Sadly though, OpenSSL implements the full package of OCSP inside their libcrypto layer, and teams moving over from OpenSSL to AWS-LC have continuously expected us to support OCSP for their use-case as well.
A good number of potential downstream consumers, like nginx or haproxy, implement their own HTTP behavior around OpenSSL's OCSP request APIs. But there are also a great deal of customers asks for us to support the OpenSSL1.1.1's original OCSP/HTTP behavior because they've been relying on it. A couple examples of customers that are already asking for the support right now are s2n-tls and the Datasync team. There are also findings of 7-8 internal packages relying on this behavior as well, which I suspect we're going to need to support as AWS-LC gains more adoption. It's a weird surface area to add, but this has been a part of a continuously growing ask from customers asking for parity with OpenSSL1.1.1.

PR #814, opened 2023-02-17
Allow keys without e in RSA_set0_key

WillChilds-Klein:
### Issues:
Resolves t/V835962727
Addresses t/V835962727

### Description of changes: 

In a [previous pull request][1], we added support for JCA-style "stripped" RSA private keys defined only in terms of modulus `n` and private exponent `d`. When that change was made, ACCP only gave these keys to AWS-LC in the form of a DER-encoded ASN.1 structure. Consequently, we only adjusted the key-decoding code path to account for these stripped keys.

This changed, however, when ACCP implemented native KeyFactory support. That change [uses][3] AWS-LC's `RSA_set0_key` function to create RSA private keys. Because the key factory branch predated ACCP's migration from OpenSSL to AWS-LC, it [relied][4] on OpenSSL's behavior of treating 0-valued public exponents as absent in order to specify JCA-style stripped keys. Because these stripped keys don't supply the public exponent `e`, the caller must turn off blinding. Some historical commits ([here][5], [here][6]) reference this need for |e| to perform blinding; apparently the `RSA_set0_key` behavior assumed that blinding would always be performed when that function was used to construct the key.

Previously, the `RSA_set*` methods didn't have any coverage, so we add a few test cases to cover the source code affected by this change.

[1]: https://github.com/aws/aws-lc/pull/436
[2]: https://github.com/corretto/amazon-corretto-crypto-provider/pull/132
[3]: https://github.com/corretto/amazon-corretto-crypto-provider/blob/8d32fe3fb685875c9e7418fdb30f2d70a8c4d80b/csrc/java_evp_keys.cpp#L588
[4]: https://github.com/corretto/amazon-corretto-crypto-provider/blob/8d32fe3fb685875c9e7418fdb30f2d70a8c4d80b/csrc/java_evp_keys.cpp#L578-L580
[5]: https://github.com/aws/aws-lc/commit/86361a391062e02012c92d1eefa20df3deb897f3
[6]: https://github.com/aws/aws-lc/commit/598e55a795fa206571f94a7cb2e5834728b85c98

### Testing:
- CI checks

the ISC license.


WillChilds-Klein:
lcov shows that this `RSA_set0_key` is now covered.

Screen Shot 2023-02-16 at 8 12 04 PM

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -230,7 +230,7 @@ void RSA_get0_crt_params(const RSA *rsa, const BIGNUM **out_dmp1,
 
 int RSA_set0_key(RSA *rsa, BIGNUM *n, BIGNUM *e, BIGNUM *d) {
   if ((rsa->n == NULL && n == NULL) ||
-      (rsa->e == NULL && e == NULL)) {
+      (rsa->e == NULL && e == NULL && rsa->d == NULL && d == NULL)) {
```
torben-hansen, line 233: Let's do as before and define a static`detect_stripped_jca_key_something()`, which makes this non-obvious condition more clear...

include/openssl/rsa.h
```diff
@@ -152,8 +152,9 @@ OPENSSL_EXPORT void RSA_get0_crt_params(const RSA *rsa, const BIGNUM **out_dmp1,
 // |n|, |e|, and |d| respectively, if non-NULL. On success, it takes ownership
 // of each argument and returns one. Otherwise, it returns zero.
 //
-// |d| may be NULL, but |n| and |e| must either be non-NULL or already
-// configured on |rsa|.
+// For a public key, |d| may be NULL, but |n| and |e| must either be non-NULL
```
torben-hansen, line 155: Call out the special missing `e` thing is for JCA only...

crypto/rsa_extra/rsa_test.cc
```diff
@@ -648,6 +648,57 @@ TEST(RSATest, OnlyDGiven) {
...
+  // Allocate an empty key to imitate JCA keys via calls to RSA_set0_key
+  bssl::UniquePtr<RSA> jcaKey(RSA_new());
+  ASSERT_TRUE(jcaKey);
+
+  // FULL KEY, BLINDING => OK
+  rsa.reset(RSA_private_key_from_bytes(kKey1, sizeof(kKey1) - 1));
+  ASSERT_TRUE(rsa);
+  jcaKey.reset(RSA_new());
+  ASSERT_TRUE(jcaKey);
+  EXPECT_TRUE(RSA_set0_key(jcaKey.get(), rsa->n, rsa->e, rsa->d));
```
torben-hansen, line 670: After this step you are double-referencing memory. Pointers in both `jcaKey` and `rsa` points to the same memory, and the former will contain dangling pointers after L677.

include/openssl/rsa.h
WillChilds-Klein: Are we sure that JCA is the only entity (or language ecosystem) that allows private keys to be specified without `e`? I'm a little hesitant to over-specify the comment if it's a more general usage pattern. Although, for what it's worth, I did to a quick spot-check of [golang's](https://pkg.go.dev/crypto/rsa#PrivateKey) and [pyca's](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateNumbers) analogous structures, and both seem to expect `e`. so maybe it is just a Java thing...

crypto/rsa_extra/rsa_test.cc
WillChilds-Klein: ah, that makes sense. for some reason, I don't hit the segfault locally that the CI is hitting. I wrapped these pointers in `BN_dup`. Still passes locally, hopefully the CI agrees.

crypto/fipsmodule/rsa/rsa.c
WillChilds-Klein: This is an art-not-science opinion, but it seems to me that pulling this out into a separate function would over-complicate things, particularly because we can use the function's doc comment to clarify that lack of `e` is indicative of a JCA key.

Also, in the DER-decoding change from last april, we not only had to detect keys but had to modify them. This was because when we deserialized them, the value of absent fields was `0` instead of `NULL`. Here, the caller has full control of the invocation and can set parameters to `NULL`, so we don't need to (and shouldn't) modify the parameters.

WillChilds-Klein: i'll update the comment.

include/openssl/rsa.h
WillChilds-Klein: i'll update the comment.

andrewhop (approved):

dkostic (approved):

PR #815, opened 2023-02-17
Move all fuzzing to al2022 with clang 14

andrewhop:
### Description of changes: 
1. Move all CI fuzzing from Ubuntu Clang 9 to AL2022 with Clang14
2. Remove undefined behavior sanitizer
3. Update seed corpus

### Call-outs:
We still run the undefined behavior sanitizer with our regular tests in the CI. This just removes it from Cryptofuzz since it doesn't fail the build (missing `-fno-sanitize-recover`) and raises too many false positives from code we don't own.

This gets the total CI run time down to 51 minutes for this PR, down from ~100 minutes. We still fuzz for the same amount of time, just the reporting is so much faster without the undefined behavior sanitizer.

### Testing:
Built and ran images locally.

the ISC license.


samuel40791765 (dismissed):
Awesome speed up and everything lgtm. Should we remove the [ubuntu-20.04_cryptofuzz](https://github.com/aws/aws-lc/tree/main/tests/ci/docker_images/linux-x86/ubuntu-20.04_cryptofuzz) docker images, since they won't be used anymore?

samuel40791765 (approved):

skmcgrail (approved):

skmcgrail (approved):

PR #820, opened 2023-02-18
HTTP header parsing when receiving an OCSP response

samuel40791765:
Relies on functionality from https://github.com/aws/aws-lc/pull/813, in draft for now.

### Issues:
Addresses `CryptoAlg-1536`

### Description of changes: 
Sends out a `OCSP_REQ_CTX` constructed from a `OCSP_REQUEST` via `OCSP_sendreq_bio`. `OCSP_sendreq_nbio`(called in `OCSP_sendreq_bio`) then expects a response in the `BIO` memory contents, and parses the HTTP response to get the `OCSP_response` within it.

Taken from: 

### Call-outs:
Code from OpenSSL was cleaned up to adhere to our code styling and library tools, but the actual logic stays the same.

### Testing:
Implemented tests focus on HTTP header parsing to get the OCSP response. Parsing of the actual OCSP response is covered by previously existing OCSP tests.

the ISC license.


crypto/ocsp/ocsp_http.c
```diff
@@ -22,11 +26,335 @@
...
+    while (*reason != '\0' && isspace(*reason)) {
+      reason++;
+    }
+    if (*reason != '\0') {
+        // Finally, zap any trailing white space in message (include CRLF).
+        // We know reason has a non-white space character so this is OK.
+        for (end = reason + strlen(reason) - 1; isspace(*end); end--) {
+          *end = '\0';
+        }
+    }
```
andrewhop, line 99: Is it possible to simplify all this using something like strtok to help split the line?

crypto/ocsp/ocsp_http.c
```diff
@@ -22,11 +26,335 @@
...
+
+        // Don't allow excessive lines.
+        if (n == rctx->iobuflen) {
+            rctx->state = OHS_ERROR;
+            return 0;
+        }
+
+        // First line.
+        if (rctx->state == OHS_FIRSTLINE) {
+            if (parse_http_line((char *)rctx->iobuf)) {
```
andrewhop, line 221: Do we know how long this line is at this point? Can this pass in iobuflen?

crypto/ocsp/ocsp_http.c
```diff
@@ -22,11 +26,335 @@
...
+
+    case OHS_ERROR:
+        return 0;
+
+    case OHS_FIRSTLINE:
+    case OHS_HEADERS:
+
+    // Attempt to read a line in.
+ next_line:
+        // Due to &%^*$" memory BIO behaviour with BIO_gets we have to check
```
andrewhop, line 192: Is `&%^*$"` a special BIO token? Or is this a censored expletive?

crypto/ocsp/ocsp_http.c
```diff
@@ -22,11 +26,335 @@
...
+        }
+        return 0;
+    }
+    return 1;
+}
+
+int OCSP_REQ_CTX_nbio(OCSP_REQ_CTX *rctx)
+{
+    int i, n;
+    const unsigned char *p;
```
andrewhop, line 116: What are these used for? Can they have a better name?

crypto/ocsp/ocsp_http.c
```diff
@@ -22,11 +26,335 @@
...
+        }
+
+        // Check it is an ASN1 SEQUENCE.
+        if (*p++ != (V_ASN1_SEQUENCE | V_ASN1_CONSTRUCTED)) {
+            rctx->state = OHS_ERROR;
+            return 0;
+        }
+
+        // Check out length field.
+        if ((*p & 0x80) != 0) {
```
andrewhop, line 259: What is 0x80 in this context? What is *p at this point?

crypto/ocsp/ocsp_http.c
```diff
@@ -22,11 +26,335 @@
...
+        }
+        else {
+          ERR_add_error_data(4, "Code=", code, ",Reason=", reason);
+        }
+        return 0;
+    }
+    return 1;
+}
+
+int OCSP_REQ_CTX_nbio(OCSP_REQ_CTX *rctx)
```
andrewhop, line 113: Can you include a high level overview of what this function is doing and why? I see in internal.h above it is "exchanging request and response via HTTP on (non-)blocking BIO in rctx", but this function is doing a lot of work that is not immediately obvious how it related back to the goal of sending and receiving a response.

crypto/ocsp/ocsp_http.c
```diff
@@ -22,11 +26,335 @@
...
 #define OHS_HTTP_HEADER         (9 | OHS_NOREAD)
 
+// Parse the HTTP response. This will look like this: "HTTP/1.0 200 OK". We
+// need to obtain the numeric code and (optional) informational message.
+static int parse_http_line(char *line)
+{
+    int http_code;
+    char *code, *reason, *end;
+    // Skip to first white space (passed protocol info)
+    for (code = line; *code != '\0' && !isspace(*code); code++) {
```
skmcgrail, line 55: Should we at least validate the lead in is HTTP 1.{0,1}?

crypto/ocsp/ocsp_http.c
skmcgrail, line 259: Worth indicating or using a constant here to make it clear that this expression is checking to see if the length is encoded as long-form (multiple bytes) versus being a length that can be encoded into 7 bits.

crypto/ocsp/ocsp_http.c
```diff
@@ -22,11 +26,335 @@
...
+    OCSP_RESPONSE *resp = NULL;
+    OCSP_REQ_CTX *ctx;
+    int rv;
+
+    ctx = OCSP_sendreq_new(b, path, req, -1);
+    if (ctx == NULL) {
+      return NULL;
+    }
+
+    // This waits indefinitely on a response.
```
skmcgrail, line 320: I assume this comment is 100% not accurate, I assume a socket timeout or some other error could cause a break-out this loop?

crypto/ocsp/ocsp_test.cc
```diff
@@ -766,3 +768,93 @@ TEST_P(OCSPRequestTest, OCSPRequestParse) {
...
+
+// This should parse. Only the status code is mandatory, subsequent lines are
+// optional.
+static const char no_type_http_response_hdr[] =
+        "HTTP/1.1 200 OK\r\n"
+        "Content-Length: ";
+
+// This should parse. Only the status code is mandatory, additional information
+// is optional.
+static const char no_info_http_response_hdr[] =
```
skmcgrail, line 796: Fuzzing might cover this, but should we add some tests for bad/malformed http responses?

crypto/ocsp/ocsp_http.c
samuel40791765: This waits indefinitely if `BIO_should_retry` is true, clarified the comment. Also added a documentation comment clarifying this API is only left for compatibility and people should avoid using it.

skmcgrail (dismissed):

include/openssl/ocsp.h
```diff
@@ -59,13 +59,30 @@ DECLARE_ASN1_FUNCTIONS(OCSP_REQUEST)
...
 DECLARE_ASN1_FUNCTIONS(OCSP_REQINFO)
 
+// Blocking OCSP request handler: now a special case of non-blocking I/O.
+// Combines OCSP_sendreq_new() with as many calls of OCSP_sendreq_nbio() as
+// needed and then OCSP_REQ_CTX_free(), with a response header maximum line
+// length 4k. It waits indefinitely on a response.
+//
+// WARNING: This does not support setting a timeout or adding your own HTTP
+// headers and is retained only for compatibility.
+// Use OCSP_sendreq_nbio and manually set your own timeout instead. You can also
```
andrewhop, line 69: How do you set a timeout for OCSP_sendreq_nbio?

samuel40791765 (commented):
Synced up with @andrewhop and we decided to revert the change to `strtok/strtok_r` parsing.
Called out in the "BUGS" section of [the manpage](https://man7.org/linux/man-pages/man3/strtok_r.3.html). These functions will modify the contents of their first argument, which will are the actual buffer contents of `OCSP_REQ_CTX`. This could result in undefined behavior when handling the rest of the buffer, so we've reverted back to the original implementation.

include/openssl/ocsp.h
```diff
@@ -59,13 +59,30 @@ DECLARE_ASN1_FUNCTIONS(OCSP_REQUEST)
 DECLARE_ASN1_FUNCTIONS(OCSP_SIGNATURE)
 DECLARE_ASN1_FUNCTIONS(OCSP_REQINFO)
 
+// Blocking OCSP request handler: now a special case of non-blocking I/O.
+// Combines OCSP_sendreq_new() with as many calls of OCSP_sendreq_nbio() as
+// needed and then OCSP_REQ_CTX_free(), with a response header maximum line
+// length 4k. It waits indefinitely on a response.
```
andrewhop, line 65: TIL about util/docs.go and this fails to parse (along with all the other public header file changes): 
```suggestion
// OCSP_sendreq_bio is a blocking OCSP request handler which is a special case of non-blocking I/O.
// |OCSP_sendreq_bio| combines |OCSP_sendreq_new| with as many calls of |OCSP_sendreq_nbio| as
// needed and then |OCSP_REQ_CTX_free|, with a response header maximum line
// length 4k. It waits indefinitely on a response.
```
>As a matter of style, comments should start with the name of the thing that they are commenting on

Also anything within |bars| is converted to a link if it is defined somewhere else, or printed as monospace.

https://github.com/aws/aws-lc/blob/main/util/doc.go#L442-L445

include/openssl/ocsp.h
samuel40791765: I actually meant setting a timer in the consumer’s own code, since these APIs don’t support it. I’ll have to clarify that:sweat_smile:

crypto/ocsp/ocsp_http.c
```diff
@@ -22,11 +26,335 @@
...
+// Request being flushed
+#define OHS_ASN1_FLUSH          (7 | OHS_NOREAD)
+// Completed
+#define OHS_DONE                (8 | OHS_NOREAD)
 // Headers set, no final \r\n included
 #define OHS_HTTP_HEADER         (9 | OHS_NOREAD)
 
+// Parse the HTTP response. This will look like this: "HTTP/1.0 200 OK". We
+// need to obtain the numeric code and (optional) informational message.
+static int parse_http_line(char *line)
```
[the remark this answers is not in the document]
andrewhop: Ahhh I see `BIO_gets(rctx->mem, (char *)rctx->iobuf, rctx->iobuflen)` reads up to rctx->iobuflen bytes, then we check `if (data_len >= rctx->iobuflen)` so we know we read the whole line.

crypto/ocsp/ocsp_http.c
andrewhop: As discussed it makes sense to keep this older parsing to avoid the dangers with strtok and strtok_r.

crypto/ocsp/ocsp_http.c
andrewhop: How does this work exactly? 0x80 is 0b10000000, won't this statement be true if the value is less than or greater than 0b10000000?
* `0b010000001 & 0b010000000 != 0`
* `0b110000000 & 0b010000000 != 0`

include/openssl/ocsp.h
andrewhop: Can we add a test that this works as expected?

crypto/ocsp/ocsp_http.c
samuel40791765: Yes the value has to exactly equal `0x80`, its part of the ASN.1 definition. If it does equal `0x80`, this means the length field (`N`) is encoded as followed after `0x80`. Anything else means that the length is encoded as long-form (multiple bytes).

andrewhop (approved):

skmcgrail (approved):

PR #835, opened 2023-02-24
Fuzz testing for OCSP response HTTP header parsing

samuel40791765:
### Issues:
Addresses `CryptoAlg-1536`

### Description of changes: 
Adds fuzz testing for the new header parsing done in `OCSP_sendreq_nbio`.

### Call-outs:
N/A

### Testing:
Followed runbook for fuzzing in https://github.com/aws/aws-lc/blob/main/FUZZING.md
Fuzz tests were ran for 1.5 hours to generate corpus.

the ISC license.


fuzz/ocsp_http.cc
```diff
@@ -0,0 +1,24 @@
...
+#include <openssl/mem.h>
+#include <openssl/ocsp.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *buf, size_t len) {
+  bssl::UniquePtr<BIO> bio(BIO_new_mem_buf(buf, len));
+  OCSP_REQ_CTX *req_ctx = OCSP_REQ_CTX_new(bio.get(), 0);
+  OCSP_RESPONSE *ocsp_response;
+
+  // OCSP_sendreq_nbio parses the HTTP response in the BIO.
+  if(OCSP_sendreq_nbio(&ocsp_response, req_ctx) > 0) {
```
andrewhop, line 14: Is this attempting to actually send the request?

samuel40791765: Yeah this hadn't been fuzzing what I initially wanted. I've updated it to use a dummy request, so we can properly get to parsing the OCSP http response. Also updated to fuzz `OCSP_REQ_CTX_nbio` instead, which is directly calling the `parse_http_line` static function.

skmcgrail (dismissed):

samuel40791765:
Apparently `BIO_new_mem_buf` creates a read-only BIO, so we weren't properly fuzzing at first. I've fixed the bug and we have a much larger corpus now.

andrewhop (approved):

skmcgrail (approved):

PR #839, opened 2023-02-26
Kyber verify fix

brian-jarvis-aws:
### Description of changes: 
When Kyber's common reference code was imported, changes were made to the `verify` function to avoid a compiler error that occurs on MSVC. However, the logic was inconsistent with the original reference code. This PR corrects the modification so that it will succeed on all supported compilers and be logically consistent with the original reference code.

### Testing:
A unit test has been added for the verify function to confirm it behaves as intended. This unit test is outside of the existing KEM tests in evp_extra_test.cc because it is Kyber-specific.

the ISC license.

### DIFF
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -709,7 +709,6 @@ if(BUILD_TESTING)
     hmac_extra/hmac_test.cc
     hrss/hrss_test.cc
     impl_dispatch_test.cc
-    kyber/kyber_ref_verify_test.cc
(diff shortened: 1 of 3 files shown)


brian-jarvis-aws:
Closed this PR to take a different approach. We're going to revert to the code as written in the upstream Kyber source and then suppress MSVC's complaints about it. Separately, we can pursue submitting a PR to the Kyber team with the change in logic and take that change ourselves if they accept it upstream.

PR #844, opened 2023-03-01
Allow RSA keys with d > n to be imported

dkostic:
### Issues:
CryptoAlg-1523/CryptoAlg-1626

### Description of changes: 
This PR reverts the change from #756 and enables the "relaxed" RSA key import by default.

### Call-outs:

### Testing:

the ISC license.

### DIFF
--- a/crypto/fipsmodule/rsa/rsa.c
+++ b/crypto/fipsmodule/rsa/rsa.c
@@ -742,24 +742,6 @@ static int check_mod_inverse(int *out_ok, const BIGNUM *a, const BIGNUM *ainv,
   return ret;
 }
 
-#if !defined(AWSLC_FIPS)
-static bool allow_rsa_keys_d_gt_n_flag = false;
-
-void allow_rsa_keys_d_gt_n(void) {
-  allow_rsa_keys_d_gt_n_flag = true;
-}
-
-static bool are_rsa_keys_with_d_gt_n_allowed(void) {
-  return allow_rsa_keys_d_gt_n_flag;
-}
-
-#else
-
-static bool are_rsa_keys_with_d_gt_n_allowed(void) {
-  return false;
-}
-#endif
-
 int RSA_validate_key(const RSA *key, rsa_asn1_key_encoding_t key_enc_type) {
   // TODO(davidben): RSA key initialization is spread across
   // |rsa_check_public_key|, |RSA_check_key|, |freeze_private_key|, and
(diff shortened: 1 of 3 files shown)


crypto/fipsmodule/rsa/rsa.c
```diff
@@ -777,18 +759,16 @@ int RSA_validate_key(const RSA *key, rsa_asn1_key_encoding_t key_enc_type) {
...
-  // translate to bounds on the running time of private key operations.
-  // See above this functions the explanation for the exception when keys
-  // with |d > n| are allowed.
-  if (key->d != NULL &&
-      are_rsa_keys_with_d_gt_n_allowed() == false &&
-      BN_cmp(key->d, key->n) >= 0) {
+  // Previously, we ensured that |key->d| is bounded by |key->n|.
+  // This ensures bounds on |RSA_bits| translate to bounds on
+  // the running time of private key operations.
+  // However, due to some users having to deal with private keys
```
andrewhop, line 765:
```suggestion
  // However, due to some users (V804729436) having to deal with private keys
```

crypto/rsa_extra/rsa_test.cc
```diff
@@ -967,30 +967,6 @@ TEST(RSATest, CheckKey) {
   ASSERT_TRUE(BN_hex2bn(&rsa->d, kDEuler));
   EXPECT_TRUE(RSA_check_key(rsa.get()));
 
-  // If d is completely out of range but otherwise valid, it is rejected.
```
andrewhop, line 970: Should we keep this test and validate we always accept keys when d > n?

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -742,24 +742,6 @@ static int check_mod_inverse(int *out_ok, const BIGNUM *a, const BIGNUM *ainv,
   return ret;
 }
 
-#if !defined(AWSLC_FIPS)
```
andrewhop, line 745: Is FIPS okay with the relaxed check and d > n?

crypto/rsa_extra/rsa_test.cc
dkostic: good point, added.

crypto/fipsmodule/rsa/rsa.c
dkostic: that's a very good question. I believe it's fine because FIPS specifies requirements for key generation and our generation method ensures that `d` is in the correct range regardless of the removed check.

andrewhop (approved):

samuel40791765 (approved):

PR #848, opened 2023-03-02
Spruce up OCSP code with our .clang-format

samuel40791765:
### Description of changes: 
Realigning indentation of OCSP code and adjusting documentation for docs.go.

### Call-outs:
Adjusted with CLion settings

### Testing:
N/A

the ISC license.


skmcgrail (approved):
Looks good CLion's diff makes it easier to condense the review since you can have it ignore the trivial whitespace changes.

include/openssl/ocsp.h
```diff
@@ -140,45 +144,52 @@ OPENSSL_EXPORT int OCSP_request_add1_cert(OCSP_REQUEST *req, X509 *cert);
...
+//          code will not be passed if |*status| is NULL.
+OPENSSL_EXPORT int OCSP_resp_find_status(OCSP_BASICRESP *bs,
+                                         OCSP_CERTID *id,
+                                         int *status,
+                                         int *reason,
+                                         ASN1_GENERALIZEDTIME **revtime,
+                                         ASN1_GENERALIZEDTIME **thisupd,
+                                         ASN1_GENERALIZEDTIME **nextupd);
+
+// OCSP_basic_verify verifies a basic response message. Returns 1 on success,
```
andrewhop, line 173: Does success == valid? Or just that the function ran to completion and you need to call something like OCSP_is_valid?
```suggestion
// OCSP_basic_verify verifies a basic response message. Returns 1 if the response is valid,
```

include/openssl/ocsp.h
```diff
@@ -140,45 +144,52 @@ OPENSSL_EXPORT int OCSP_request_add1_cert(OCSP_REQUEST *req, X509 *cert);
...
+// 0 on error, or -1 on fatal errors such as malloc failure.
```
andrewhop, line 174: Is this error the same as invalid signature?

include/openssl/ocsp.h
samuel40791765: Yeah, It test's that it's a valid response according to the certs passed in. Next step in a normal OCSP pipeline would to be check that the response hasn't expired.

include/openssl/ocsp.h
samuel40791765: Yes, clarified :)

samuel40791765 (commented):
> CLion's diff 
TIL. Just tried it, the interface is much cleaner

samuel40791765:
Turns out our .clang-format uses two space indents, so I've used that instead to align with the rest of the codebase

andrewhop (approved):

skmcgrail (approved):

PR #849, opened 2023-03-03
Additional OCSP Request functions required

samuel40791765:
### Issues:
Addresses `CryptoAlg-1483`

### Description of changes: 
This change focuses on adding `OCSP_parse_url` to support service teams who need it. More information about the API can be found in the API documentation.

### Call-outs:
Behavior should be the same as OpenSSL1.1.1. Only change made to the original parsing were clearer null checks, and addressing compiler warnings.

### Testing:
New valid and invalid test cases in `OCSPParseURL`. We can add additional fuzzing if needed

the ISC license.


crypto/ocsp/ocsp_http.c
```diff
@@ -520,3 +519,14 @@ BIO *OCSP_REQ_CTX_get0_mem_bio(OCSP_REQ_CTX *rctx)
 {
     return rctx->mem;
 }
+
+void OCSP_set_max_response_length(OCSP_REQ_CTX *rctx, unsigned long len)
```
andrewhop, line 523: What is max_resp_len used for? I expect that after changing this size some buffer needs to be re-allocated

crypto/ocsp/ocsp_test.cc
```diff
@@ -1134,3 +1162,85 @@ TEST_P(OCSPHTTPTest, OCSPRequestHTTP) {
...
+    {"http://[2001:db8::1]/ocsp", "2001:db8::1", "80", "/ocsp", 0,
+     OCSP_URL_PARSE_SUCCESS},
+    {"https://[2001:db8::1]/ocsp", "2001:db8::1", "443", "/ocsp", 1,
+     OCSP_URL_PARSE_SUCCESS},
+    // Parsing a URL with an invalid port, but OpenSSL still parses the string
+    // where the port is expected.
+    {"http://ocsp.example.com:invalid/", "ocsp.example.com", "invalid", "/", 0,
+     OCSP_URL_PARSE_SUCCESS},
+    // Empty host component.
+    {"http:///ocsp", "", "80", "/ocsp", 0, OCSP_URL_PARSE_SUCCESS},
```
andrewhop, line 1200: I have mixed feelings on this, at least it's documented and expected.

crypto/ocsp/ocsp_test.cc
```diff
@@ -1134,3 +1162,85 @@ TEST_P(OCSPHTTPTest, OCSPRequestHTTP) {
...
+
+    // === INVALID URLs ===
```
andrewhop, line 1202: Do you think more fuzz tests are useful for url parsing?

include/openssl/ocsp.h
```diff
@@ -107,6 +107,11 @@ OPENSSL_EXPORT OCSP_REQ_CTX *OCSP_REQ_CTX_new(BIO *io, int maxline);
 // Frees the memory allocated by |OCSP_REQ_CTX|.
 OPENSSL_EXPORT void OCSP_REQ_CTX_free(OCSP_REQ_CTX *rctx);
 
+// OCSP_set_max_response_length sets the maximum response length for an OCSP
+// request over HTTP.
```
andrewhop, line 111:
```suggestion
// request over HTTP to |len|.
```

include/openssl/ocsp.h
```diff
@@ -180,6 +185,30 @@ OPENSSL_EXPORT int OCSP_basic_verify(OCSP_BASICRESP *bs, STACK_OF(X509) *certs,
 OPENSSL_EXPORT OCSP_CERTID *OCSP_cert_to_id(const EVP_MD *dgst, const X509 *subject,
                                             const X509 *issuer);
 
+// OCSP_parse_url parses an OCSP responder URL and returns its component parts.
+// The url argument must be a null-terminated string containing the URL to be
```
andrewhop, line 189:
```suggestion
// |url| must be a null-terminated string containing the URL to be
```

include/openssl/ocsp.h
```diff
@@ -180,6 +185,30 @@ OPENSSL_EXPORT int OCSP_basic_verify(OCSP_BASICRESP *bs, STACK_OF(X509) *certs,
...
+// parsed. The other arguments are pointers to variables that will be set to the
+// parsed components of the URL:
```
andrewhop, line 191: This should cal out the pointers point to new memory that has a copy of the value and it's the caller's responsibility to free them.

crypto/ocsp/ocsp_http.c
samuel40791765: This is used when anticipating the maximum size of the HTTP OCSP response sent back from the OCSP responder in `OCSP_REQ_CTX_nbio`.
* https://github.com/aws/aws-lc/blob/ocsp-request/crypto/ocsp/ocsp_http.c#L314

This function should be set before the `OCSP_REQ_CTX` is sent out with `OCSP_REQ_CTX_nbio` (I'll add a comment for that). The variable is only used as a max response length check, so no memory needs to be reallocated here.

crypto/ocsp/ocsp_test.cc
samuel40791765: Yeah, I've thought about it, but this function doesn't do any validation of the URL and only parses the string based on a `http(s)://{host}:{port}/{path}` format. It wouldn't hurt to fuzz test this, I'm just not quite sure if there's any additional value. But if we're worried about any edge cases I hadn't thought of, let me know and I can look into doing so.

crypto/ocsp/ocsp_test.cc
```diff
@@ -822,6 +825,32 @@ TEST(OCSPRequestTest, AddCert) {
   certId = nullptr;
 }
 
+// Test that |OCSP_set_max_response_length| correctly sets the maximum response
+// length.
```
andrewhop, line 829: Do we have a test that tests the expected behavior of max_resp_len? If the max response is really short or long that OCSP_REQ_CTX_nbio does the right thing?

samuel40791765: Sure, I added a test to set `max_resp_len` to a really short length, and test if `OCSP_REQ_CTX_nbio` fails as expected. Our existing tests already test that `OCSP_REQ_CTX_nbio` does the right thing, when max response has a sufficient length.

andrewhop (approved):

skmcgrail (approved):
As an FYI the most recent upstream merge includes some new OPENSSL_isspace & OPENSSL_isidigit functions that we might want to look at using for some of the other code you've already merged in.

samuel40791765:
> As an FYI the most recent upstream merge includes some new OPENSSL_isspace & OPENSSL_isidigit functions that we might want to look at using for some of the other code you've already merged in.

Ah thanks for reminding me, the `OPENSSL_isspace` will definately come into play. I'll merge the feature branch with main and update the usages of `isspace` in the `parse_http_line` static function.

PR #853, opened 2023-03-06
Add AES-CCM/ECB/CTR/KW to 'test_fips' tool

billbo-yang:
### Issues:
Addresses CryptoAlg-1641

### Description of changes: 
AWS-LC currently only tests a few AES functions as a part of its tooling in `test_fips.c`. However, as a part of ATSEC's process, they wish to be able to step into functions for AES-CCM, ECB, CTR, AND KW with a debugger and thus have requested that we add them to our testing script.

In the interest of not having to repeat this patch for future validations, these functions were added to the `test_fips` tool the intention to bring these changes into the main, FIPS 1MU, and FIPS 2022 branches. Additionally, it would likely be prudent to bring the self_checks for the main and FIPS 2022 branches in line with these changes in another PR.

### Call-outs:
N/A

### Testing:
Changes tested by ensuring that all tests still pass and that the output of the `test_fips` tool is as expected.

the ISC license.

### DIFF
--- a/util/fipstools/test_fips.c
+++ b/util/fipstools/test_fips.c
@@ -146,20 +146,13 @@ int main(int argc, char **argv) {
   EVP_AEAD_CTX_cleanup(&aead_ctx);
 
   OPENSSL_memset(nonce, 0, sizeof(nonce));
-  if (!EVP_AEAD_CTX_init(&aead_ctx, EVP_aead_aes_128_ccm_bluetooth(), kAESKey, SIZEOF(kAESKey), 0, NULL))
+  if (!EVP_AEAD_CTX_init(&aead_ctx, EVP_aead_aes_128_ccm_bluetooth(), kAESKey, sizeof(kAESKey), 0, NULL))
   {
     fprintf(stderr, "EVP_AED_CTX_init for AES-128-CCM failed.\n");
     goto err;
   }
 
   /* AES-CCM Encryption */
-  memcpy(aes_iv, 0, sizeof(aes_iv));
-  if (AES_set_encrypt_key(kAESKey, 8 * sizeof(kAESKey), &aes_key) != 0)
-  {
-    fprintf(stderr, "AES_set_encrypt_key failed.\n");
-    goto err;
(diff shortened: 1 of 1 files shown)


util/fipstools/test_fips.c
```diff
@@ -140,8 +149,128 @@ int main(int argc, char **argv) {
   printf("  got ");
   hexdump(output, out_len);
 
+  EVP_AEAD_CTX_zero(&aead_ctx);
   EVP_AEAD_CTX_cleanup(&aead_ctx);
 
+  /* AES-CCM */
+  OPENSSL_memset(nonce, 0, sizeof(nonce));
```
skmcgrail, line 156: Probably should be consistent in the usage of either using `memset` or `OPENSSL_memset`.

billbo-yang: Will do, I'll switch `memset` to `OPENSSL_memset` since I believe that's what's being used already elsewhere in the file.

util/fipstools/test_fips.c
```diff
@@ -140,8 +149,128 @@ int main(int argc, char **argv) {
...
+    goto err;
+  }
+  printf("About to AES-KW Wrap ");
+  hexdump(output, out_len);
+  AES_wrap_key(&aes_key, NULL, output, kPlaintext, sizeof(kPlaintext));
+  printf("  got ");
+  hexdump(output, out_len);
+
+  OPENSSL_cleanse(&aes_key, sizeof(aes_key));
+
```
samuel40791765, line 273: Do we need to do an `AES_unwrap_key` here?

samuel40791765 (commented):
Just being a bit nit-picky here, but the `{` should generally be on the same line as the if/for statement

util/fipstools/test_fips.c
billbo-yang: ATSEC doesn't _need_ it per se but it can't hurt to put it in there for consistency lol

samuel40791765 (approved):

PR #854, opened 2023-03-06
Add OCSP functions needed to support nginx

samuel40791765:
### Issues:
Addresses `CryptoAlg-1562`

### Description of changes: 
Add additional OCSP functions to support nginx's OCSP use-case.
This includes:
*  `OCSP_response_status_str`
*  `OCSP_cert_status_str`
*  `OCSP_check_validity` 

### Call-outs:
`OCSP_check_validity` grabs the current system time via `time`, and checks the time field parameters for possible OCSP invalidity. Due to this behavior, we'll need to be careful of possible time bombs in our tests. Overriding the system's time in C without super-user permissions isn't possible, so I've opted to set the `nsec` parameter to the current system's time. This will force a "Status Not Yet Valid" and a "Status Expired" error in `OCSP_check_validity`.
All other tests regarding `OCSP_check_validity` have been carefully configured to avoid time bombs in our tests as well, feel free to let me know if there are any more hiding in the tests I've added.

### Testing:
CI

the ISC license.


crypto/ocsp/ocsp_test.cc
```diff
@@ -208,97 +210,103 @@ struct OCSPTestVectorExtended {
...
-     OCSP_RESPFINDSTATUS_SUCCESS, V_OCSP_CERTSTATUS_UNKNOWN},
+     OCSP_RESPONSE_STATUS_SUCCESSFUL, "successful", OCSP_VERIFYSTATUS_SUCCESS,
+     OCSP_RESPFINDSTATUS_SUCCESS, V_OCSP_CERTSTATUS_UNKNOWN, "unknown"},
     // Test OCSP response signed by the correct responder certificate, but not
-    // for the requested certificate. (So this would be a completely valid
-    // response to a different OCSP request for the other certificate.)
+    // for
+    // the requested certificate. (So this would be a completely valid response
+    // to a
+    // different OCSP request for the other certificate.)
```
andrewhop, line 251: Formatting is off

crypto/ocsp/ocsp_print.c
```diff
@@ -0,0 +1,46 @@
...
+}
+
+const char *OCSP_response_status_str(long s) {
+  static const OCSP_TBLSTR rstat_tbl[] = {
+      {OCSP_RESPONSE_STATUS_SUCCESSFUL, "successful"},
+      {OCSP_RESPONSE_STATUS_MALFORMEDREQUEST, "malformedrequest"},
+      {OCSP_RESPONSE_STATUS_INTERNALERROR, "internalerror"},
+      {OCSP_RESPONSE_STATUS_TRYLATER, "trylater"},
+      {OCSP_RESPONSE_STATUS_SIGREQUIRED, "sigrequired"},
+      {OCSP_RESPONSE_STATUS_UNAUTHORIZED, "unauthorized"}};
```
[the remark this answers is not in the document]
andrewhop: I thought more about this and the advantage of your original approach is it simplifies some of the error cases. If we used the status as the index then we'd need another test to verify the status values were reasonable (small positive values, a status of int_max would require creating an array with int_max length).

crypto/ocsp/ocsp_test.cc
```diff
@@ -208,97 +210,103 @@ struct OCSPTestVectorExtended {
...
   const EVP_MD *dgst;
   int expected_ocsp_status;
+  const char *expected_ocsp_status_string;
   int expected_ocsp_verify_status;
   int expected_ocsp_resp_find_status;
   int expected_ocsp_cert_status;
+  const char *expected_ocsp_cert_status_string;
 };
 
 static const OCSPTestVectorExtended nTestVectors[] = {
```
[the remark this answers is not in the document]
samuel40791765: Great idea. I've tried briefly applying the `Combine` generator, but the scope of the change seems to require a slight overhaul of this test suite. The diff for that would most likely muddle the changes we're focusing in this commit.

We've synced up and agreed to start working on this in a subsequent PR.

include/openssl/ocsp.h
```diff
@@ -164,9 +164,26 @@ OPENSSL_EXPORT int OCSP_resp_find_status(OCSP_BASICRESP *bs, OCSP_CERTID *id,
...
                                          ASN1_GENERALIZEDTIME **nextupd);
 
-// OCSP_basic_verify verifies a basic response message. Returns 1 if the
-// response is valid, 0 if the signature cannot be verified, or -1 on fatal
-// errors such as malloc failure.
+// OCSP_check_validity checks the validity of |thisUpdate| and |nextUpdate|
+// fields from an |OCSP_SINGLERESP|.
+//
+// Note: It is possible that the request will take a few  seconds to process
+// and/or the time won't be totally accurate. Therefore, to avoid rejecting
```
andrewhop, line 171:
```suggestion
// Note: It is possible that the request will take a few seconds to process
// and/or the local system time isn't exactly the same as the OCSP responder's time. Therefore, to avoid rejecting
```

include/openssl/ocsp.h
```diff
@@ -164,9 +164,26 @@ OPENSSL_EXPORT int OCSP_resp_find_status(OCSP_BASICRESP *bs, OCSP_CERTID *id,
...
+// |nextUpdate| field, an optional |max_age_seconds| parameter specifies the
+// maximum age the |thisUpdate| field can be. |max_age_seconds| should be the
+// number of seconds relative to |thisUpdate|. You can also set
+// |max_age_seconds| to "-1", if the maximum age should not be checked.
+OPENSSL_EXPORT int OCSP_check_validity(ASN1_GENERALIZEDTIME *thisUpdate,
+                                       ASN1_GENERALIZEDTIME *nextUpdate,
+                                       long drift_num_seconds,
+                                       long max_age_seconds);
+
+// OCSP_basic_verify verifies a basic response message. It checks that |bs| is
```
andrewhop, line 183: Awesome comment. What is the recommend way to fully verify an OCSP response? Is it call OCSP_basic_verify, if success check OCSP_check_validity, is there another step after that?

include/openssl/ocsp.h
```diff
@@ -211,6 +225,12 @@ OPENSSL_EXPORT OCSP_CERTID *OCSP_cert_to_id(const EVP_MD *dgst,
 OPENSSL_EXPORT int OCSP_parse_url(const char *url, char **phost, char **pport,
                                   char **ppath, int *pssl);
 
+// OCSP_response_status_str prints OCSP response status in a string.
+OPENSSL_EXPORT const char *OCSP_response_status_str(long s);
```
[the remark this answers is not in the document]
andrewhop: Missed this comment, or am I mistaken?

include/openssl/ocsp.h
samuel40791765: Yes, after `OCSP_basic_verify`, `OCSP_check_validity` should be used to verify the OCSP response's timestamps. Asides from that, the caller should also use `OCSP_resp_find_status` to verify if the OCSP response's Cert status is `V_OCSP_CERTSTATUS_GOOD`.
I'll add both as `Note 2.` below :) .

include/openssl/ocsp.h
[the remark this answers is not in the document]
samuel40791765: Haha I got confused for a bit too, but I think this thread is showing the past. 
* Updated version: https://github.com/aws/aws-lc/pull/854/files#diff-99acc9e006f0d0869a05a30d5b721624685365c53da01662dab1c17a317c0930R231-R233

include/openssl/ocsp.h
```diff
@@ -164,12 +164,34 @@ OPENSSL_EXPORT int OCSP_resp_find_status(OCSP_BASICRESP *bs, OCSP_CERTID *id,
                                          ASN1_GENERALIZEDTIME **thisupd,
                                          ASN1_GENERALIZEDTIME **nextupd);
 
-// OCSP_basic_verify verifies a basic response message. Returns 1 if the
-// response is valid, 0 if the signature cannot be verified, or -1 on fatal
-// errors such as malloc failure.
+// OCSP_check_validity checks the validity of |thisUpdate| and |nextUpdate|
+// fields from an |OCSP_SINGLERESP|.
+//
```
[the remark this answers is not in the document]
samuel40791765: >  is there a maximum future value we accept?
There shouldn't be a maximum value we accept. This time field is when newer information will be available about the status of the certificate, which is something the OCSP responder decides.

Added the additional documentation

andrewhop (approved):
I don't have a good suggestion to simplify the [error handling here](https://github.com/aws/aws-lc/pull/854#discussion_r1137848568) so this looks good to me. I'm still open to other ideas on how to improve it.

crypto/ocsp/ocsp_client.c
```diff
@@ -247,3 +248,61 @@ int OCSP_resp_find_status(OCSP_BASICRESP *bs, OCSP_CERTID *id, int *status,
...
+int OCSP_check_validity(ASN1_GENERALIZEDTIME *thisUpdate,
+                        ASN1_GENERALIZEDTIME *nextUpdate,
+                        long drift_num_seconds, long max_age_seconds) {
+  int ret = 1;
+  int64_t t_tmp;
+  int64_t t_now = time(NULL);
+
+  // Check |thisUpdate| is valid and not more than |drift_num_seconds| in the
+  // future.
+  if (!ASN1_GENERALIZEDTIME_check(thisUpdate)) {
```
skmcgrail, line 261: Is my understanding correct that the ASN.1 GeneralizedTime is expected to be UTC only, and not have any other timezone offset?

skmcgrail (approved):
Looks good, just one question related to timezones.

crypto/ocsp/ocsp_client.c
samuel40791765: Great question, I looked into a bit more and the RFC states that "GeneralizedTime values MUST be expressed in Greenwich Mean Time", but can optionally "include a representation of the time differential between local and Greenwich Mean Time".
* https://datatracker.ietf.org/doc/html/rfc6960#section-4.2.2.1
* https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.5.2

So the time is expected to UTC only (no local time allowed), but a differential offset can be included.

samuel40791765: This definition is just in the context of CRLs and OCSP

PR #855, opened 2023-03-07
KEM: fix key retrieval when public/secret key is missing

dkostic:
### Issues:
Resolves V845882454.

### Description of changes: 
Fixes key retrieval when public/secret key is missing. Specifically,
when `EVP_PKEY_get_raw_private_key` is called on a key that has
only the public key an error will be returned. The same when calling
`EVP_PKEY_get_raw_public_key` with a key that has only the
private key.

Also, got rid of the `has_secret_key` field in the `KEM_KEY`
structure because it is not necessary.

### Call-outs:

### Testing:

the ISC license.


crypto/evp_extra/p_kem.c
```diff
@@ -145,7 +143,7 @@ static int pkey_kem_decapsulate(EVP_PKEY_CTX *ctx,
   }
 
   KEM_KEY *key = ctx->pkey->pkey.kem_key;
-  if (!key->has_secret_key) {
+  if (key->secret_key == NULL) {
```
andrewhop, line 146: Should we add a similar safety check to encapsulate?

torben-hansen (commented):
lgtm, except https://github.com/aws/aws-lc/pull/855/files#r1128788636 makes sense for encaps.

Do we check `shared_secret` anywhere for encaps?
I tried chasing it down the call-stack. But the Kyber reference implementation doesn't exactly make this an easy exercise.

crypto/evp_extra/p_kem.c
dkostic: added

dkostic: btw, I'm working on a PR to add a bunch of test cases for failure modes of the API. Will add a test for this particular case as well.

andrewhop (approved):

torben-hansen (approved):

PR #861, opened 2023-03-10
Add OCSP_crl_reason_str to support libcurl

samuel40791765:
### Issues:
Resolves `CryptoAlg-1563`

### Description of changes: 
Once https://github.com/aws/aws-lc/pull/854 is merged in, the only missing gap to support curl's OCSP functionality is `OCSP_crl_reason_str`. There are some warnings to resolve once OCSP is turned on in Curl, but we can contribute the small changes to them once this is in (patch under SIM). 

### Call-outs:
N/A

### Testing:
Running against Curl's tests built with AWS-LC. https://github.com/jeking3/curl/actions/runs/3959544784/jobs/6782497658

the ISC license.


include/openssl/ocsp.h
```diff
@@ -250,6 +268,8 @@ OPENSSL_EXPORT const char *OCSP_response_status_str(long status_code);
 // a string.
 OPENSSL_EXPORT const char *OCSP_cert_status_str(long status_code);
 
+// OCSP_crl_reason_str prints the OCSP CRL reason into a string.
```
andrewhop, line 271:
```suggestion
// OCSP_crl_reason_str returns the OCSP CRL reason as a string.
```

include/openssl/ocsp.h
```diff
@@ -250,6 +268,8 @@ OPENSSL_EXPORT const char *OCSP_response_status_str(long status_code);
...
+OPENSSL_EXPORT const char *OCSP_crl_reason_str(long s);
```
andrewhop, line 272: How does this work? Where does a customer get one of these reason codes? This PR seems to be adding the 8 defines in a public header but we never use/return them.

andrewhop: Ahhhh so the OCSP responder would put these values in on their side. That explains why we're not using the defines when we create the object anywhere.

include/openssl/ocsp.h
```diff
@@ -250,6 +250,10 @@ OPENSSL_EXPORT const char *OCSP_response_status_str(long status_code);
 // a string.
 OPENSSL_EXPORT const char *OCSP_cert_status_str(long status_code);
 
+// OCSP_crl_reason_str returns the OCSP CRL reason of |status_code| a string.
```
andrewhop, line 253:
```suggestion
// OCSP_crl_reason_str returns the OCSP CRL reason of |status_code| as a string.
```

andrewhop (dismissed):

skmcgrail (approved):

andrewhop (approved):

skmcgrail (approved):

PR #862, opened 2023-03-10
Reference OPENSSL_ia32cap_P at x86 machine code level through a unique symbol instead of a common offset symbol avoiding add instruction

torben-hansen:
### Issues:

CryptoAlg-1577

### Description of changes: 

See [#856](https://github.com/aws/aws-lc/pull/856) for background and description of issue resolved in this PR.

This is the second PR out of two PRs that bridge the performance gap.

See ticket for performance comparisons.

The first PR took care of the C-level. But the machine-optimised algorithm implementation sometimes directly dereference `OPENSSL_ia32cap_P`. These also need to be fixed-up. As before, we can't just add a call instruction to `OPENSSL_ia32cap_get` because it would compromise soundness/correctness.

Recall, that the issue was the `add` instruction. Instead of injecting that instruction, we instead do the following for each occurrence of `OPENSSL_ia32cap_P` discovered in the textual assembly:
* Define a new unique symbol. Uniquness is ensured using both the original (not true in all cases, but see "call outs") register name and a unique global counter. Bit redundant using both, but makes it easier to read IMO.
* Under the new unique symbol, copy the address of `OPENSSL_ia32cap_P` into the original register. This will probably spawn a relocation, but this is fine, because we put this new unique symbol outside the FIPS module scope in `.text`. Putting it here also means that we know the address relative to RIP.
* To jump back to the original execution point, we inject a "return" symbol/label where `OPENSSL_ia32cap_P` was discovered. 

Examplewise, we have something looking like the following.
Original:
```
    mov OPENSSL_ia32cap_P(%RIP), %rax
```
Transformed into:
```
    jmp OPENSSL_ia32cap_P_rax0
OPENSSL_ia32cap_P_rax0_return:

"after fips scope"

OPENSSL_ia32cap_P_rax0:
    lea OPENSSL_ia32cap_P(%RIP), %rax
    jmp OPENSSL_ia32cap_P_rax0_return
```

Note, we need a new symbol for each occurrence of `OPENSSL_ia32cap_P` to ensure we can jump back to the correct place.

### Call-outs:

The original destination register is not always used. The reason is that the delocator script fix-up other things that switches out the original register. For example, for the `instrCombine` type instructions, [we switch register](https://github.com/aws/aws-lc/blob/main/util/fipstools/delocate/delocate.go#L1456-L1466), using [an ordered list of possible new registers](https://github.com/aws/aws-lc/blob/main/util/fipstools/delocate/delocate.go#L1220).

We use this ordered list in the tests to correctly predict the output.  For example, the **4th** occurrence of `OPENSSL_ia32cap_P`
```
orq OPENSSL_ia32cap_P@GOTPCREL(%rip), %rax
```
does **NOT** result in the following code output
```
	jmp	LOPENSSL_ia32cap_P_rax3
LOPENSSL_ia32cap_P_rax2_return:

"after fips scope"

.type LOPENSSL_ia32cap_P_rax3, @function
LOPENSSL_ia32cap_P_rax3:
	leaq OPENSSL_ia32cap_P(%rip), %rax
	jmp LOPENSSL_ia32cap_P_rax3_return
```
because of the `orq` instruction. **Instead**, the `rbx` register is used, producing
```
	jmp	LOPENSSL_ia32cap_P_rbx3
LOPENSSL_ia32cap_P_rbx3_return:

"after fips scope"
.type LOPENSSL_ia32cap_P_rbx3, @function
LOPENSSL_ia32cap_P_rbx3:
	leaq OPENSSL_ia32cap_P(%rip), %rbx
	jmp LOPENSSL_ia32cap_P_rbx3_return
```

### Testing:

Ran the delocator test script `delocate_script.go` from the delocate folder:
```
$ go test
PASS
ok      boringssl.googlesource.com/boringssl/util/fipstools/delocate    0.011s
```

Although, I had to "fix" a small regression like this:
```
diff --git a/util/fipstools/delocate/delocate.go b/util/fipstools/delocate/delocate.go
index df2ebbe3a..98499dbcd 100644
--- a/util/fipstools/delocate/delocate.go
+++ b/util/fipstools/delocate/delocate.go
@@ -1431,7 +1431,8 @@ Args:
 
                                classification := classifyInstruction(instructionName, argNodes)
                                if classification != instrThreeArg && classification != instrCompare && i != 0 {
-                                       return nil, fmt.Errorf("GOT access must be source operand, %w", classification)
+                                       //return nil, fmt.Errorf("GOT access must be source operand, %w", classification)
+                                       return nil, nil
                                }
```
the ISC license.

### DIFF
--- a/util/fipstools/delocate/delocate.go
+++ b/util/fipstools/delocate/delocate.go
@@ -1539,12 +1539,14 @@ Args:
 				}
 
 				if symbol == "OPENSSL_ia32cap_P" {
+
 					uniqueSymbol := newCpuCapUniqueSymbol(len(d.cpuCapUniqueSymbols), targetReg)
 					wrappers = append(wrappers, func(k func()) {
 						d.output.WriteString("\tjmp\t" + uniqueSymbol.getIntelSymbol() + "\n")
 						d.output.WriteString(uniqueSymbol.getIntelSymbolReturn() + ":\n")
 					})
 					d.cpuCapUniqueSymbols = append(d.cpuCapUniqueSymbols, uniqueSymbol)
+
 				} else if useGOT {
 					wrappers = append(wrappers, d.loadFromGOT(d.output, targetReg, symbol, section, redzoneCleared))
 				} else {
@@ -1965,6 +1967,8 @@ func transform(w stringWriter, inputs []inputFile) error {
 		w.WriteString("\tleaq OPENSSL_ia32cap_P(%rip), %rax\n")
 		w.WriteString("\tret\n")
 
+		// This is a fixed order iteration. such that we can soundly write tests
+		// for this in /testdata.
 		for _, uniqueSymbol := range d.cpuCapUniqueSymbols {
 			w.WriteString(".type " + uniqueSymbol.getIntelSymbol() + ", @function\n")
 			w.WriteString(uniqueSymbol.getIntelSymbol() + ":\n")
--- a/util/fipstools/delocate/delocate.go
+++ b/util/fipstools/delocate/delocate.go
@@ -1239,18 +1239,6 @@ func (d *delocation) loadFromGOT(w stringWriter, destination, symbol, section st
 	}
 }
 
-func saveFlags(w stringWriter, redzoneCleared bool) wrapperFunc {
-	return func(k func()) {
-		if !redzoneCleared {
-			w.WriteString("\tleaq -128(%rsp), %rsp\n") // Clear the red zone.
-			defer w.WriteString("\tleaq 128(%rsp), %rsp\n")
-		}
-		w.WriteString("\tpushfq\n")
-		k()
-		w.WriteString("\tpopfq\n")
-	}
-}
-
 func saveRegister(w stringWriter, avoidRegs []string) (wrapperFunc, string) {
 	candidates := []string{"%rax", "%rbx", "%rcx", "%rdx"}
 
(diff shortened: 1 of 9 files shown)


util/fipstools/delocate/delocate.go
```diff
@@ -59,6 +59,30 @@ const (
 	aarch64
 )
 
+// represents a unique symbol for an occurrence of OPENSSL_ia32cap_P.
+type cpuCapUniqueSymbol struct {
+	registerName string
+	suffixUniqueness string
+}
+
+const cpuCapIntelSymbolPrefix = "LOPENSSL_ia32cap_P_"
```
dkostic, line 68: can we call it `x86` because it's used both on Intel and AMD?

util/fipstools/delocate/delocate.go
```diff
@@ -59,6 +59,30 @@ const (
...
+	return cpuCapIntelSymbolPrefix + uniqueSymbol.registerName + uniqueSymbol.suffixUniqueness
+}
+
+func (uniqueSymbol cpuCapUniqueSymbol) getIntelSymbolReturn() string {
+	return uniqueSymbol.getIntelSymbol() + "_return"
+}
+
+// uniqueness must be a globally unique integer value.
+func newCpuCapUniqueSymbol(uniqueness int, registerName string) *cpuCapUniqueSymbol {
+    uniqueSymbol := new(cpuCapUniqueSymbol)
```
skmcgrail, line 80: Simplify to:
```go
return &uniqueSymbol{
   registerName: strings.Trim(registerName, "%"), // should work with both AT&T and Intel syntax.
   suffixUniqueness: strconv.Itoa(uniqueness),
}
```
Which is a bit more idiomatic

util/fipstools/delocate/delocate.go
torben-hansen: makes sense!

util/fipstools/delocate/delocate.go
torben-hansen: great!

skmcgrail (approved):

skmcgrail (approved):

PR #863, opened 2023-03-10
Avoid potential for buffer overflow in SHA3 ARMv8 assembly

brian-jarvis-aws:
### Description of changes: 
The `SHA3_Squeeze()` function is part of the internal API of the SHA3 and SHAKE implementations; it performs the "squeeze" operation of the sponge construction that is at the core of these hash functions. It takes as parameters a Keccak state (a 5×5 matrix of 64-bit words), an output buffer (of length `len` bytes), and the "rate" (i.e. the number of bytes that can be extracted from the Keccak state for each application of the Keccak-f permutation).

On ARMv8 architectures, a dedicated assembly implementation of Keccak is generated by the Perl script in _crypto/fipsmodule/sha/asm/keccak1600-armv8.pl_. In the assembly output, there is a potential for a buffer overflow if `len` is set to 0. This is currently no code path in the calling function (`SHA3_Final()`) which would produce a 0-valued `len`. Regardless, this patch addresses the _potential_ for the internal `SHA3_Squeeze()` method to be misused.

This patch also addresses the same issue with `SHA3_Squeeze_cext()`, which is a currently-unused specialized version of `SHA3_Squeeze()` that leverages the SHA3-specific opcodes present as an extension in some ARMv8 CPUs.

### Testing:
The existing testing suite was confirmed passing on a Graviton3 EC2 instance.

I also disassembled the `crypto_test` output executable and inspected the opcodes which were generated for `SHA3_Squeeze`:
```
0000000000917c60 <SHA3_Squeeze>:
  917c60:       a9bd7bfd        stp     x29, x30, [sp, #-48]!
  917c64:       910003fd        mov     x29, sp
  917c68:       f100005f        cmp     x2, #0x0                   //<---- New instruction
  917c6c:       54000600        b.eq    917d2c <SHA3_Squeeze+0xcc>  //<---- New instruction
<SNIP>
  917d2c:       a8c37bfd        ldp     x29, x30, [sp], #48 //<---- .Lsqueeze_abort
  917d30:       d65f03c0        ret
```
This shows that at address 0x917c68, the value of `len` (in register x2) is compared to 0. If it's equal, it jumps to address 0x917d2c, which is where `.Lsqueeze_abort` begins, effectively skipping the loop which would have resulted in a buffer overflow in this `len==0` condition.

This was verified in the generated opcodes for `SHA3_Squeeze_cext` as well: 
```
0000000000918140 <SHA3_Squeeze_cext>:
  918140:       a9bf7bfd        stp     x29, x30, [sp, #-16]!
  918144:       910003fd        mov     x29, sp
  918148:       f100005f        cmp     x2, #0x0                         //<---- New instruction
  91814c:       54000540        b.eq    9181f4 <SHA3_Squeeze_cext+0xb4>  //<---- New instruction
```


dkostic (dismissed):

nebeid (approved):

dkostic (approved):
Approved the fix, but I'm wondering if we can add a unit test for it (I don't mind it being in a separate PR)?

PR #874, opened 2023-03-15
Add OCSP integration tests with Amazon Trust Services

samuel40791765:
### Issues:
Resolves `CryptoAlg-1625`

### Description of changes: 
This PR adds an integration tests against Amazon's public CA, Amazon Trust Services. The certs used in this tests were taken from their examples: https://www.amazontrust.com/repository/. Amazon Trust Services has multiple OCSP responder endpoints, `ocsp.sca[0-4]a.amazontrust.com`, so we try to retrieve an OCSP responder against all of them.

### Call-outs:
This integration test dimension only tries to retrieve an OCSP response from the OCSP responder endpoint. All certs that are used in these tests will expire one day. We can look into updating these certs periodically every couple of years, but we should avoid relying on any time-related behavior for our unit tests to pass. We already test OCSP verification against hard-coded OCSP responses in the `OCSPTest` Test suites.

### Testing:
New test dimensions.

the ISC license.


crypto/ocsp/ocsp_integration_test.cc
```diff
@@ -0,0 +1,212 @@
...
+
+  // Try connecting to the OCSP responder endpoint with a timeout of 3 seconds.
+  // Sends out an OCSP request and expects an OCSP response.
+  OCSP_RESPONSE *resp = nullptr;
+  time_t start_time = time(nullptr);
+  double timeout = 3;
+  int rv;
+  do {
+    rv = OCSP_sendreq_nbio(&resp, ocsp_ctx.get());
+  } while ((rv == -1) && difftime(time(nullptr), start_time) < timeout);
```
andrewhop, line 62: Nice error/timeout handling.

crypto/ocsp/ocsp_integration_test.cc
```diff
@@ -0,0 +1,212 @@
...
+                                 STACK_OF(X509) *chain,
+                                 OCSP_CERTID *cert_id,
+                                 int expected_verify_status,
+                                 int expected_cert_status) {
+  ASSERT_EQ(OCSP_response_status(response),
+            authorized_responder ? OCSP_RESPONSE_STATUS_SUCCESSFUL
+                                 : OCSP_RESPONSE_STATUS_UNAUTHORIZED);
+  if (!authorized_responder) {
+    return;
+  }
```
andrewhop, line 79: Is it possible/interesting to move this down below after the OCSP_response_get1_basic and if it's not a authorized responder we expect there is garbage or nothing in the response and basic_response would be null? Basically a very tiny fuzz test that if someone connects to a bad endpoint our code fails and fails gracefully.

samuel40791765: Good point, I think that totally makes sense. I added an assert after the basic response parsing, it should always be false in these cases.

skmcgrail (approved):

andrewhop (approved):

PR #875, opened 2023-03-15
Add `KemKeyShare` to libssl

bbutch:
### Description of changes: 
This is the first of several PRs to add PQ/hybrid functionality to `libssl`. This PR adds:
* A new key share type: `KemKeyShare`. This handles the keygen/encapsulation/decapsulation flow for PQ KEMs in TLS 1.3. For the foreseeable future, `KemKeyShare` will not be available to negotiate on its own - it will be used only as part of a hybrid key share. In a future PR, an additional type, `HybridKeyShare`, will be added. `HybridKeyShare` will implement the generalized hybrid concept, allowing the hybrid combination of any other key share types. (In particular, we will be interested in combining an `ECKeyShare` with a `KemKeyShare`.)

### Call-outs:
[`NamedGroups`](https://github.com/aws/aws-lc/blob/6dcb1c327d4e9023855b92d6f7eb8e2c54e5616f/ssl/ssl_key_share.cc#L302-L315) defines which groups are made available for TLS negotiation. Since we are not adding anything to that, the `KemKeyShare` functionality is effectively feature-guarded from general use.

### Testing:
* Unit tests for `KemKeyShare` have been added. 
* Interoperability and integration tests will be added in future PRs.

the ISC license.

### DIFF
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -2366,6 +2366,11 @@ OPENSSL_EXPORT int SSL_set1_curves_list(SSL *ssl, const char *curves);
 #define SSL_CURVE_X25519 29
 #define SSL_CURVE_CECPQ2 16696
 
+// PQ and hybrid group IDs are not yet standardized. Current IDs are driven by
+// community consensus and are defined at
+// https://github.com/open-quantum-safe/openssl/blob/OQS-OpenSSL_1_1_1-stable/oqs-template/oqs-kem-info.md
+#define SSL_CURVE_KYBER512_R3 0x023A
+
 // SSL_get_curve_id returns the ID of the curve used by |ssl|'s most recently
 // completed handshake or 0 if not applicable.
 //
--- a/ssl/ssl_key_share.cc
+++ b/ssl/ssl_key_share.cc
@@ -24,6 +24,7 @@
 #include <openssl/curve25519.h>
 #include <openssl/ec.h>
 #include <openssl/err.h>
+#include <openssl/evp.h>
 #include <openssl/hrss.h>
 #include <openssl/mem.h>
 #include <openssl/nid.h>
@@ -325,11 +326,13 @@ class KEMKeyShare : public SSLKeyShare {
     }
 
     // Generate a KEM keypair and retrieve the length of the public key
+    size_t public_key_len = 0;
     EVP_PKEY *raw_key = nullptr;
     if (!EVP_PKEY_keygen_init(ctx_.get()) ||
         !EVP_PKEY_keygen(ctx_.get(), &raw_key) ||
         !raw_key ||
-        !EVP_PKEY_get_raw_public_key(raw_key, nullptr, &public_key_len_)) {
+        !EVP_PKEY_get_raw_public_key(raw_key, nullptr, &public_key_len)) {
+      EVP_PKEY_free(raw_key);
       OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
       return false;
     }
@@ -344,16 +347,16 @@ class KEMKeyShare : public SSLKeyShare {
 
     // Write the public key to |out|
     Array<uint8_t> public_key;
-    if (!public_key.Init(public_key_len_)) {
+    if (!public_key.Init(public_key_len)) {
       OPENSSL_PUT_ERROR(SSL, ERR_R_MALLOC_FAILURE);
       return false;
     }
 
-    size_t public_key_bytes_written = public_key_len_;
+    size_t public_key_bytes_written = public_key_len;
     if (!EVP_PKEY_get_raw_public_key(pkey.get(), public_key.data(),
                                      &public_key_bytes_written) ||
-                                     public_key_bytes_written != public_key_len_ ||
-                                     !CBB_add_bytes(out, public_key.data(), public_key_len_)) {
+                                     public_key_bytes_written != public_key_len ||
+                                     !CBB_add_bytes(out, public_key.data(), public_key_len)) {
       OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
       return false;
     }
@@ -361,6 +364,9 @@ class KEMKeyShare : public SSLKeyShare {
     return true;
   }
 
+  // Because this is a KEM key share, |out_public_key| will be the ciphertext
+  // resulting from the encapsulation of the shared secret under the
+  // received public key, |peek_key|.
   bool Accept(CBB *out_public_key, Array<uint8_t> *out_secret,
               uint8_t *out_alert, Span<const uint8_t> peer_key) override {
     // Basic nullptr checks
@@ -417,8 +423,10 @@ class KEMKeyShare : public SSLKeyShare {
     }
 
     // Retrieve the lengths of the ciphertext and shared secret
-    if (!EVP_PKEY_encapsulate(ctx_.get(), nullptr, &ciphertext_len_,
-                              nullptr, &secret_len_)) {
+    size_t ciphertext_len = 0;
+    size_t secret_len = 0;
+    if (!EVP_PKEY_encapsulate(ctx_.get(), nullptr, &ciphertext_len,
+                              nullptr, &secret_len)) {
       OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
       return false;
     }
@@ -428,19 +436,19 @@ class KEMKeyShare : public SSLKeyShare {
     // peer by writing it to |out_public_key|, then...
     Array<uint8_t> ciphertext;
     Array<uint8_t> shared_secret;
-    if (!ciphertext.Init(ciphertext_len_) || !shared_secret.Init(secret_len_)) {
+    if (!ciphertext.Init(ciphertext_len) || !shared_secret.Init(secret_len)) {
       OPENSSL_PUT_ERROR(SSL, ERR_R_MALLOC_FAILURE);
       return false;
     }
 
-    size_t ciphertext_bytes_written = ciphertext_len_;
-    size_t secret_bytes_written = secret_len_;
+    size_t ciphertext_bytes_written = ciphertext_len;
+    size_t secret_bytes_written = secret_len;
     if (!EVP_PKEY_encapsulate(ctx_.get(), ciphertext.data(),
                               &ciphertext_bytes_written, shared_secret.data(),
                               &secret_bytes_written) ||
-                              ciphertext_bytes_written != ciphertext_len_ ||
-                              secret_bytes_written != secret_len_ ||
-                              !CBB_add_bytes(out_public_key, ciphertext.data(), ciphertext_len_)) {
+                              ciphertext_bytes_written != ciphertext_len ||
+                              secret_bytes_written != secret_len ||
+                              !CBB_add_bytes(out_public_key, ciphertext.data(), ciphertext_len)) {
       OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
       return false;
     }
@@ -453,6 +461,10 @@ class KEMKeyShare : public SSLKeyShare {
     return true;
   }
 
+  // Because this is a KEM key share, |peer_key| is actually the ciphertext
+  // resulting from the peer encapsulating the shared secret under our public key.
+  // In Finish(), we use our previously generated private key to decrypt
+  // that ciphertext and obtain the shared secret.
   bool Finish(Array<uint8_t> *out_secret, uint8_t *out_alert,
               Span<const uint8_t> peer_key) override {
     // Basic nullptr checks
@@ -476,14 +488,16 @@ class KEMKeyShare : public SSLKeyShare {
     }
 
     // Retrieve the lengths of the ciphertext and shared secret
-    if (!EVP_PKEY_encapsulate(ctx_.get(), nullptr, &ciphertext_len_, nullptr,
-                              &secret_len_)) {
+    size_t ciphertext_len = 0;
+    size_t secret_len = 0;
+    if (!EVP_PKEY_encapsulate(ctx_.get(), nullptr, &ciphertext_len, nullptr,
+                              &secret_len)) {
       OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
       return false;
     }
 
     // Ensure that peer_key is valid
-    if (!peer_key.data() || peer_key.size() != ciphertext_len_) {
+    if (!peer_key.data() || peer_key.size() != ciphertext_len) {
       *out_alert = SSL_AD_DECODE_ERROR;
       OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
       return false;
@@ -494,16 +508,16 @@ class KEMKeyShare : public SSLKeyShare {
     // key to decapsulate it, and retain the shared secret by writing
     // it to |out_secret|.
     uint8_t *ciphertext = (uint8_t *)peer_key.begin();
-    if (!out_secret->Init(secret_len_)) {
+    if (!out_secret->Init(secret_len)) {
       OPENSSL_PUT_ERROR(SSL, ERR_R_MALLOC_FAILURE);
       return false;
     }
 
-    size_t secret_bytes_written = secret_len_;
+    size_t secret_bytes_written = secret_len;
     if (!EVP_PKEY_decapsulate(ctx_.get(), out_secret->data(),
                               &secret_bytes_written, ciphertext,
-                              ciphertext_len_) ||
-                              secret_bytes_written != secret_len_) {
+                              ciphertext_len) ||
+                              secret_bytes_written != secret_len) {
       OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
       return false;
     }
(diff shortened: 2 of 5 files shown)


ssl/ssl_key_share.cc
```diff
@@ -294,6 +295,233 @@ class CECPQ2KeyShare : public SSLKeyShare {
...
+      return false;
+    }
+
+    // Generate a KEM keypair and retrieve the length of the public key
+    EVP_PKEY *raw_key = nullptr;
+    if (!EVP_PKEY_keygen_init(ctx_.get()) ||
+        !EVP_PKEY_keygen(ctx_.get(), &raw_key) ||
+        !raw_key ||
+        !EVP_PKEY_get_raw_public_key(raw_key, nullptr, &public_key_len_)) {
+      OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
```
dkostic, line 333: `raw_key` is not freed in the failure path.

ssl/ssl_key_share.cc
```diff
@@ -294,6 +295,233 @@ class CECPQ2KeyShare : public SSLKeyShare {
...
+  bool Offer(CBB *out) override {
+    // Ensure |out| is valid, has no children, and has been initialized.
+    // We check that |out| has no children because otherwise CBB_data()
+    // will produce a fatal error by way of assert(cbb->child == NULL).
+    if (!out || out->child || !CBB_data(out)) {
+      OPENSSL_PUT_ERROR(SSL, ERR_R_PASSED_NULL_PARAMETER);
+      return false;
+    }
+
+    // ctx_ should not have been initialized (by e.g. a previous call
```
dkostic, line 313: why not? The EVP API is designed such that a single `EVP_PKEY_CTX` can be used for multiple key generations

dkostic, line 313: why not? The EVP API is designed such that a single `EVP_PKEY_CTX` can be used for multiple key generations

ssl/ssl_key_share.cc
```diff
@@ -294,6 +295,233 @@ class CECPQ2KeyShare : public SSLKeyShare {
...
+                                     public_key_bytes_written != public_key_len_ ||
+                                     !CBB_add_bytes(out, public_key.data(), public_key_len_)) {
+      OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
+      return false;
+    }
+
+    return true;
+  }
+
+  bool Accept(CBB *out_public_key, Array<uint8_t> *out_secret,
```
dkostic, line 364: `out_public_key`? shouldn't this be `out_ciphertext` or something like that?

ssl/ssl_key_share.cc
```diff
@@ -294,6 +295,233 @@ class CECPQ2KeyShare : public SSLKeyShare {
...
+
+    // Retrieve the lengths of the ciphertext and shared secret
+    if (!EVP_PKEY_encapsulate(ctx_.get(), nullptr, &ciphertext_len_, nullptr,
+                              &secret_len_)) {
+      OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
+      return false;
+    }
+
+    // Ensure that peer_key is valid
+    if (!peer_key.data() || peer_key.size() != ciphertext_len_) {
```
dkostic, line 486: the naming is very confusing, for example `peer_key.size() != ciphertext_len_`

ssl/ssl_key_share.cc
```diff
@@ -294,6 +295,233 @@ class CECPQ2KeyShare : public SSLKeyShare {
...
+    return true;
+  }
+
+ private:
+  int nid_;
+  uint16_t group_id_;
+  UniquePtr<EVP_PKEY_CTX> ctx_;
+  size_t public_key_len_;
+  size_t ciphertext_len_;
+  size_t secret_len_;
```
dkostic, line 522: do we need these three?

ssl/ssl_key_share.cc
```diff
@@ -303,12 +531,21 @@ CONSTEXPR_ARRAY NamedGroup kNamedGroups[] = {
     {NID_CECPQ2, SSL_CURVE_CECPQ2, "CECPQ2", "CECPQ2"},
 };
 
+CONSTEXPR_ARRAY uint16_t kPQGroups[] = {
+    SSL_CURVE_CECPQ2,
```
dkostic, line 535: the KEM API doesn't support `CECPQ2`.

include/openssl/ssl.h
```diff
@@ -2366,6 +2366,11 @@ OPENSSL_EXPORT int SSL_set1_curves_list(SSL *ssl, const char *curves);
 #define SSL_CURVE_X25519 29
 #define SSL_CURVE_CECPQ2 16696
 
+// PQ and hybrid group IDs are not yet standardized. Current IDs are driven by
+// community consensus and are defined at
+// https://github.com/open-quantum-safe/openssl/blob/OQS-OpenSSL_1_1_1-stable/oqs-template/oqs-kem-info.md
+#define SSL_CURVE_KYBER512_R3 0x023A
```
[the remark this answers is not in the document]
dkostic: why `CURVE`? Kyber is not based on elliptic curves.

ssl/ssl_key_share.cc
```diff
@@ -303,12 +531,21 @@ CONSTEXPR_ARRAY NamedGroup kNamedGroups[] = {
...
+    SSL_CURVE_KYBER512_R3,
```
dkostic, line 536: should we add Kyber to the named groups as well?

dkostic (commented):
I reviewed everything except `ssl_test.cc`. Will continue with the remaining file tomorrow.

ssl/ssl_key_share.cc
```diff
@@ -294,6 +295,233 @@ class CECPQ2KeyShare : public SSLKeyShare {
...
+
+    // ctx_ should not have been initialized (by e.g. a previous call
+    // to Offer()).
+    if (ctx_) {
+      OPENSSL_PUT_ERROR(SSL, ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
+      return false;
+    }
+
+    // Initialize the KEM ctx
+    ctx_.reset(EVP_PKEY_CTX_new_id(EVP_PKEY_KEM, nullptr));
```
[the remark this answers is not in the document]
bbutch: I was just following the [KEM README](https://github.com/aws/aws-lc/tree/dbb1ea4499804e046276a1d1ab61187e142d3ff4/crypto/kem):

> The context. How to create the context? Depending on the use case, this can be done in two ways:
> 
> 1. EVP_PKEY_CTX *ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_KEM, NULL);
>    EVP_PKEY_CTX_kem_set_params(ctx, NID_KYBER512_R3);

ssl/ssl_key_share.cc
bbutch: That is true for the EVP APIs, but I argue should not be the case for `SSLKeyShare` objects. This was my way of enforcing that a single `SSLKeyShare` object should be used only to generate/negotiate a single key share. Even if the case of an HRR or renegotiation, the handshake code should create a new `SSLKeyShare` object instead of reusing an existing one (this is the way the code is written at present). My reasoning for this is just based on code hygiene and preventing bugs from popping up in the future. The way things are written currently, it _should_ be safe to reuse a key share object, but since the key shares maintain some sort state, I could imagine situations where the code flow gets changed and bugs occur because the key share is trying to Accept() or Finish() based on a previously stored, incorrect private key.

bbutch: See above

ssl/ssl_key_share.cc
bbutch: Like above, I'm following the established convention set by CECPQ2:

https://github.com/aws/aws-lc/blob/dbb1ea4499804e046276a1d1ab61187e142d3ff4/ssl/ssl_key_share.cc#L229-L230

https://github.com/aws/aws-lc/blob/dbb1ea4499804e046276a1d1ab61187e142d3ff4/ssl/ssl_key_share.cc#L267-L268

 and the function definition:

https://github.com/aws/aws-lc/blob/dbb1ea4499804e046276a1d1ab61187e142d3ff4/ssl/internal.h#L1105-L1120

ssl/ssl_key_share.cc
```diff
@@ -294,6 +295,233 @@ class CECPQ2KeyShare : public SSLKeyShare {
...
+    }
+
+    // ctx_ should have been initialized by a previous call to Offer()
+    if (!ctx_) {
+      OPENSSL_PUT_ERROR(SSL, ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
+      return false;
+    }
+
+    // Retrieve the lengths of the ciphertext and shared secret
+    if (!EVP_PKEY_encapsulate(ctx_.get(), nullptr, &ciphertext_len_, nullptr,
```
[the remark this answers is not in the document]
bbutch: I used `encapsulate` to get both the ciphertext length and the shared secret length. (As far as I can tell, you cannot retrieve the ciphertext length via a similar call to `decapsulate`.) The ciphertext length is needed to validate that the received ciphertext is the correct size. I'm not 100% sure that this check is necessary as per TLS 1.3 spec, but it is definitely proper practice and is done by the other key shares:

https://github.com/aws/aws-lc/blob/dbb1ea4499804e046276a1d1ab61187e142d3ff4/ssl/ssl_key_share.cc#L168

https://github.com/aws/aws-lc/blob/dbb1ea4499804e046276a1d1ab61187e142d3ff4/ssl/ssl_key_share.cc#L168

https://github.com/aws/aws-lc/blob/dbb1ea4499804e046276a1d1ab61187e142d3ff4/ssl/ssl_key_share.cc#L276

ssl/ssl_key_share.cc
bbutch: Probably not. I don't actually see any place where I'm relying on their state between calls.

ssl/ssl_key_share.cc
```diff
@@ -294,6 +295,233 @@ class CECPQ2KeyShare : public SSLKeyShare {
...
+
+    // Success; clear the alert
+    *out_alert = 0;
+    return true;
+  }
+
+ private:
+  int nid_;
+  uint16_t group_id_;
+  UniquePtr<EVP_PKEY_CTX> ctx_;
```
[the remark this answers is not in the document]
bbutch: Yes - the client creates the context in and secret key `Offer()`, then needs the same context/secret key in `Finish()` to perform the decapsulation. (Unless there is a better approach for retaining the private key.)

ssl/ssl_key_share.cc
bbutch: No, but this is a collection of all PQ groups, regardless of whether the KEM APIs support them. This list is not used to determine if a KEM API should be called; its used for other handshake logic:

```
// is_post_quantum_group returns true if |id| is a group ID associated
// with a post-quantum group (either hybrid or standalone).
// Otherwise, it returns false.
static bool is_post_quantum_group(uint16_t id) {
  for (const uint16_t pq_group_id : PQGroups()) {
    if (id == pq_group_id) {
      return true;
    }
  }
  return false;
}
```

https://github.com/aws/aws-lc/blob/dbb1ea4499804e046276a1d1ab61187e142d3ff4/ssl/extensions.cc#L343-L346

https://github.com/aws/aws-lc/blob/dbb1ea4499804e046276a1d1ab61187e142d3ff4/ssl/extensions.cc#L408-L410

https://github.com/aws/aws-lc/blob/dbb1ea4499804e046276a1d1ab61187e142d3ff4/ssl/extensions.cc#L2303-L2311

https://github.com/aws/aws-lc/blob/dbb1ea4499804e046276a1d1ab61187e142d3ff4/ssl/extensions.cc#L2549-L2553

ssl/ssl_key_share.cc
bbutch: Adding it to the named groups will make it immediately publicly available for use in handshakes. I'm assuming we wouldn't want that because (a) we do not (yet) want kyber as a standalone group to be used, and (b) other testing (notably the Go integration tests) would need to be added.

ssl/ssl_key_share.cc
```diff
@@ -294,6 +295,233 @@ class CECPQ2KeyShare : public SSLKeyShare {
...
+    }
+
+    // Ensure that peer_key is valid
+    if (!peer_key.data() || peer_key.size() != ciphertext_len_) {
+      *out_alert = SSL_AD_DECODE_ERROR;
+      OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
+      return false;
+    }
+
+    // |peer_key| is the ciphertext, encapsulating the shared secret,
```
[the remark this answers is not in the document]
bbutch: Fixed

ssl/ssl_key_share.cc
bbutch: Added clarifying comment to the beginning of the function

ssl/ssl_key_share.cc
bbutch: Fixed

ssl/ssl_key_share.cc
bbutch: Removed the member variables from the class definition

ssl/ssl_key_share.cc
```diff
@@ -294,6 +296,243 @@ class CECPQ2KeyShare : public SSLKeyShare {
...
+    *out_secret = std::move(shared_secret);
+
+    // Success; clear the alert
+    *out_alert = 0;
+    return true;
+  }
+
+  // Because this is a KEM key share, |peer_key| is actually the ciphertext
+  // resulting from the peer encapsulating the shared secret under our public key.
+  // In Finish(), we use our previously generated private key to decrypt
```
dkostic, line 466:
```suggestion
  // In Finish(), we use our previously generated private key to decapsulate
```

ssl/ssl_key_share.cc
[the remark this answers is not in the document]
dkostic: we can retain only the actual private key `EVP_PKEY` instead of the entire context?

ssl/ssl_key_share.cc
dkostic: > this is the way the code is written at present

which code?

ssl/ssl_test.cc
```diff
@@ -44,6 +45,8 @@
 #include "../crypto/internal.h"
 #include "../crypto/test/test_util.h"
 #include "internal.h"
+#include "../crypto/kyber/kem_kyber.h"
```
dkostic, line 48: do we need this one?

ssl/ssl_test.cc
```diff
@@ -574,6 +588,23 @@ static const CurveTest kCurveTests[] = {
     },
 };
 
+// The key share sizes are taken from the corresponding specification.
+//
+// Kyber Round 3: https://pq-crystals.org/kyber/data/kyber-specification-round3-20210804.pdf
```
dkostic, line 593: please add a comment what these three correspond to from the spec.

ssl/ssl_key_share.cc
dkostic: right now nothing stops the user to call Finish multiple times with the same key share, but Offer and Accept can be called only once. Should we destroy the state at the end of Finish to ensure it can't be called more than once?

ssl/ssl_test.cc
```diff
@@ -9268,5 +9300,514 @@ TEST(SSLTest, InvalidSignatureAlgorithm) {
...
+        MakeConstSpan(server_out_public_key_data, CBB_len(&server_out_public_key));
+      EXPECT_TRUE(client_key_share->Finish(&client_secret, &client_alert,
+                                           server_public_key));
+
+      // The shared secrets are of the correct length...
+      EXPECT_EQ(client_secret.size(), t.shared_secret_size);
+      EXPECT_EQ(server_secret.size(), t.shared_secret_size);
+
+      // ... but they are not equal
+      EXPECT_NE(0, OPENSSL_memcmp(client_secret.data(), server_secret.data(),
```
dkostic, line 9679:
```suggestion
      EXPECT_NE(client_secret.data(), server_secret.data());
```
this doesn't work?

ssl/ssl_test.cc
```diff
@@ -9268,5 +9300,514 @@ TEST(SSLTest, InvalidSignatureAlgorithm) {
...
+                                 t.shared_secret_size));
+
+
+      OPENSSL_free(invalid_server_public_key_buf);
+    }
+
+    CBB_cleanup(&server_out_public_key);
+    CBB_cleanup(&client_out_public_key);
+  }
+}
```
dkostic, line 9810: nice, I appreciate all the failure mode tests!

ssl/ssl_key_share.cc
bbutch: You're right, the unit tests I'm adding do not get into this failure case. I also don't see a way to trigger it; this failure path only triggers if one of the EVP APIs fails. I'm not sure how it would be possible to force one of those functions to fail from the outside like this. I'll give it some thought and check the EVP/KEM unit tests; maybe it's done there somehow in a way we can replicate here.

ssl/ssl_key_share.cc
bbutch: > > this is the way the code is written at present
> 
> which code?

The [client key share/handshake code](https://github.com/aws/aws-lc/blob/d1c1bfcf015e88fd022f7cfcc205d9e9f2c3412b/ssl/extensions.cc#L2291-L2330). This code path is used for both the initial ClientHello, and a ClientHello sent as in response to an HRR. Point being that a new key share is always `Create()`ed.

PR #878, opened 2023-03-15
Fix OPENSSL_ia32cap Behavior on Windows 1MU FIPS build

billbo-yang:
### Issues:
Resolves CryptoAlg-1644

### Description of changes: 
We recently became aware of a bug on the 1MU branch where the FIPS integrity check would break if the `OPENSSL_ia32cap` environment was changed during runtime. This was caused by the fact that the file in which it was defined, `cpucap.c`, was included in `bcm.c`, which meant that by extension it was inside the FIPS boundary. As such, changing the variable from how it was when the library was built would break the FIPS integrity check. This PR fixes this problem by moving `cpucap.c` out of the FIPS boundary.

### Call-outs:
Note that these changes are built off of PR https://github.com/aws/aws-lc/pull/496, which was originally supposed to fix the static FIPS build on Linux in main. However, the changes were almost exactly what was needed to fix the bug we were encountering on Windows and as such the changes were appropriated. The static FIPS build for Linux was disabled again in this PR as it was still broken in this branch.

### Testing:
These changes were tested by ensuring that the FIPS integrity test no longer failed when `OPENSSL_ia32cap` is changed after building the library on Windows. They were also tested on Ubuntu to ensure that the static build was properly disabled.

the ISC license.


billbo-yang:
Closing as PR https://github.com/aws/aws-lc/pull/881 is a better way to address the problem and has already been merged into the 1MU branch.

PR #879, opened 2023-03-16
Add checks to ensure that OPENSSL_ia32cap is not in the fips module

billbo-yang:
### Issues:
N/A

### Description of changes: 

AWS-LC currently does not explicitly check that `OPENSSL_ia32cap`is not in the fips module during it's integrity check. Additionally, it only runs the urandom test - always returns success on x86 shared builds - with the environment variable set. This has proven to be a problem as outlined in the description for PR https://github.com/aws/aws-lc/pull/878.

As such, this PR adds checks to the integrity check and also tests `crypto_test` with the variable set in order to ensure that any change that accidentally adds OPENSSL_ia32cap back into the fips module is caught in the future.

### Call-outs:
N/A

### Testing:
Built locally on Windows x86 and MacOS arm to ensure that the tests run as expected.

the ISC license.

### DIFF
--- a/crypto/fipsmodule/bcm.c
+++ b/crypto/fipsmodule/bcm.c
@@ -70,6 +70,7 @@
 #include "cipher/e_aes.c"
 #include "cipher/e_aesccm.c"
 
+#include "cpucap/internal.h"
 #include "cpucap/cpu_aarch64_apple.c"
 #include "cpucap/cpu_aarch64_freebsd.c"
 #include "cpucap/cpu_aarch64_fuchsia.c"
@@ -155,12 +156,6 @@ extern const uint8_t BORINGSSL_bcm_text_hash[];
 #if defined(BORINGSSL_SHARED_LIBRARY)
 extern const uint8_t BORINGSSL_bcm_rodata_start[];
 extern const uint8_t BORINGSSL_bcm_rodata_end[];
-#endif
-
-// We need to get OPENSSL_ia32cap_P from cpucap.c without including the file
-// However, OPENSSL_ia32cap is only defined for x86 processors
-#if defined(OPENSSL_X86) || defined(OPENSSL_X86_64)
-extern uint32_t OPENSSL_ia32cap_P[4];
 #endif
 
 // assert_within is used to sanity check that certain symbols are within the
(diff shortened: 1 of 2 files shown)


util/all_tests.json
```diff
@@ -27,6 +27,12 @@
     "env": ["OPENSSL_armcap=0x3D"],
     "target_arch": "arm"
   },
+  {
+    "comment": "Test OPENSSL_ia32cap on crypto_test for x86, as urandom_test is disabled for shared builds on x86",
+    "cmd": ["crypto/crypto_test"],
```
andrewhop, line 32: This duplicates all the crypto_tests and takes a lot of time. Is there a single test we can execute that relies on OPENSSL_ia32cap that this can run with a `--gtest_filter`.

crypto/fipsmodule/bcm.c
```diff
@@ -278,6 +287,9 @@ int BORINGSSL_integrity_test(void) {
   assert_within(rodata_start, kPrimes, rodata_end);
   assert_within(rodata_start, kP256Params, rodata_end);
   assert_within(rodata_start, kPKCS1SigPrefixes, rodata_end);
+#if defined(OPENSSL_X86) || defined(OPENSSL_X86_64)
+  assert_not_within(rodata_start, OPENSSL_ia32cap_P, rodata_end);
```
andrewhop, line 291: Please add a check for OPENSSL_armcap_P in an if ARM block

andrewhop (approved):

crypto/fipsmodule/bcm.c
```diff
@@ -278,6 +284,11 @@ int BORINGSSL_integrity_test(void) {
...
```
torben-hansen, line 288: How can `OPENSSL_ia32cap_P` ever be in the `.rodate` segment? It's not a constant, hence can't be mapped as read-only.

billbo-yang: On Windows, the regular data segment and read-only data segments are the same: https://github.com/aws/aws-lc/blob/6dcb1c327d4e9023855b92d6f7eb8e2c54e5616f/crypto/fipsmodule/fips_shared_library_marker.c#L38-L39

This combined with the fact that in the 1MU branch, `cpucap.c` was included directly in `bcm.c`, mean that the variable was in the read-only data segment. While this has already been fixed in main (and a fix is coming for the 1MU branch), we want to ensure that if it somehow happens again we catch it early.

torben-hansen: I'm not sure I get this, on the other hand, my experience with Windows is minimal.

If the equivalent to `.data`/`.bss` in COFF land is lumped together with read-oonly data, how is the hash over it (always) deterministic (with respect to the initial hash computation at compile time) when mapped into memory? We just don't modify any data before the integrity check, apart from `OPENSSL_ia32cap_P`?

billbo-yang: You're right that this is a problem, but it's also not in the scope of this PR so I created another one that addresses your concerns here: https://github.com/aws/aws-lc/pull/880

dkostic (approved):

PR #893, opened 2023-03-21
Add SHA3_Squeeze output buffer overflow test

WillChilds-Klein:
### Issues:
Resolves CryptoAlg-1671

# Notes

# Add SHA3_Squeeze output buffer overflow test

The idea behind this new test draws from [stack canaries](https://en.wikipedia.org/wiki/Buffer_overflow_protection#Canaries). Basically, we append a segment of memory with well-known contents to our output buffer, failing the test if that canary value changes after the call to SHA3_Squeeze. This methodology isn't perfect, as it won't detect cases where the function overwrites the canary segment with that same canary value (`0xff` in this case). If we wanted to improve this, we might fill the canary with random data, however the static value is enough to catch the motivating overflow in testing.

#  Wrap SHA3 ARMv8 ASM optimizations in C function

Interestingly, we [mark all ARM assembly `.globl`s as `.hidden`](https://github.com/aws/aws-lc/commit/3ab3e3db6e5e060612d9f2172aa0efa3ff8f55a3), making it impossible to directly unit-test symbols defined only in ARM assembly on ARM platforms. To work around this, for each exported SHA3 function that also has an ASM implementation, we rename that implementation with a `_asm` suffix and wrap it with the exported C function.

# Testing

Manually testing on a c7g.xlarge (Graviton 3 ARMv8), I reverted the buffer overflow fix e1065a0 and confirmed that the test fails:

```
[ec2-user@ip-172-31-93-239 aws-lc]$ git log --oneline HEAD^..HEAD
18c024362 (HEAD -> main) Revert "Avoid potential for buffer overflow in SHA3 ARMv8 assembly (#863)"

[ec2-user@ip-172-31-93-239 aws-lc]$ cmake-build.sh && ./build/crypto/crypto_test --gtest_filter=KeccakInternalTest.SqueezeOutputBufferOverflow
...
[----------] 1 test from KeccakInternalTest
[ RUN      ] KeccakInternalTest.SqueezeOutputBufferOverflow
/home/ec2-user/aws-lc/crypto/fipsmodule/sha/sha3_test.cc:228: Failure
Value of: std::equal(out.end() - canary.size(), out.end(), canary.begin()) == true
  Actual: false
Expected: true
[  FAILED  ] KeccakInternalTest.SqueezeOutputBufferOverflow (0 ms)
[----------] 1 test from KeccakInternalTest (0 ms total)

[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] KeccakInternalTest.SqueezeOutputBufferOverflow

 1 FAILED TEST
```

---

the ISC license.

### DIFF
--- a/crypto/fipsmodule/sha/keccak1600.c
+++ b/crypto/fipsmodule/sha/keccak1600.c
@@ -393,4 +393,22 @@ void SHA3_Squeeze(uint64_t A[SHA3_ROWS][SHA3_ROWS], uint8_t *out, size_t len, si
         }
     }
 }
+
+#else
+
+size_t SHA3_Absorb_hw(uint64_t A[SHA3_ROWS][SHA3_ROWS], const uint8_t *inp, size_t len,
+                       size_t r);
+
+size_t SHA3_Absorb(uint64_t A[SHA3_ROWS][SHA3_ROWS], const uint8_t *inp, size_t len,
+                   size_t r) {
+    return SHA3_Absorb_hw(A, inp, len, r);
+}
+
+size_t SHA3_Squeeze_hw(uint64_t A[SHA3_ROWS][SHA3_ROWS], const uint8_t *out, size_t len,
+                        size_t r);
+
+void SHA3_Squeeze(uint64_t A[SHA3_ROWS][SHA3_ROWS], uint8_t *out, size_t len, size_t r) {
+    SHA3_Squeeze_hw(A, out, len, r);
+}
+
 #endif // !KECCAK1600_ASM
--- a/crypto/fipsmodule/sha/sha3_test.cc
+++ b/crypto/fipsmodule/sha/sha3_test.cc
@@ -211,6 +211,26 @@ TEST(SHA3Test, NISTTestVectors_SingleShot) {
   });
 }
 
+TEST(KeccakInternalTest, SqueezeOutputBufferOverflow) {
+    EVP_MD_unstable_sha3_enable(true);
+
+    KECCAK1600_CTX ctx;
+    std::vector<uint8_t> out;
+    std::vector<uint8_t> canary(8);
+    std::fill(canary.begin(), canary.end(), 0xff);
+
+    const size_t out_lens[] = { 0, 1, 2, 3, 4, 5, 6, 7, 8, (1<<5), (1<<16)+1 };
+    for (auto out_len : out_lens) {
+        EXPECT_TRUE(SHA3_Init(&ctx, SHA3_PAD_CHAR, SHA3_384_DIGEST_BITLENGTH));
+        out.resize(out_len + canary.size());
+        std::copy(canary.begin(), canary.end(), out.end() - canary.size());
+        SHA3_Squeeze(ctx.A, out.data(), out_len, ctx.block_size);
+        EXPECT_TRUE(std::equal(out.end() - canary.size(), out.end(), canary.begin()) == true);
+    }
+
+    EVP_MD_unstable_sha3_enable(false);
+}
+
 TEST(SHAKE128Test, NISTTestVectors) {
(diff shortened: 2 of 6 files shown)


nebeid:
We had assembly functions prefixed with `hw` (or `aes_hw`, in the case of AES) rather than suffixed with `asm`.

crypto/fipsmodule/sha/keccak1600.c
```diff
@@ -393,4 +393,24 @@ void SHA3_Squeeze(uint64_t A[SHA3_ROWS][SHA3_ROWS], uint8_t *out, size_t len, si
...
 }
+
+#else
+
+size_t SHA3_Absorb_asm(uint64_t A[SHA3_ROWS][SHA3_ROWS], const uint8_t *inp, size_t len,
+                       size_t r);
+
+size_t SHA3_Absorb(uint64_t A[SHA3_ROWS][SHA3_ROWS], const uint8_t *inp, size_t len,
+                   size_t r)
+{
```
torben-hansen, line 404: no newline

crypto/fipsmodule/sha/keccak1600.c
```diff
@@ -393,4 +393,24 @@ void SHA3_Squeeze(uint64_t A[SHA3_ROWS][SHA3_ROWS], uint8_t *out, size_t len, si
...
+    return SHA3_Absorb_asm(A, inp, len, r);
+}
+
+size_t SHA3_Squeeze_asm(uint64_t A[SHA3_ROWS][SHA3_ROWS], const uint8_t *out, size_t len,
+                        size_t r);
+
+void SHA3_Squeeze(uint64_t A[SHA3_ROWS][SHA3_ROWS], uint8_t *out, size_t len, size_t r)
+{
```
torben-hansen, line 412: no newline

crypto/fipsmodule/sha/sha3_test.cc
```diff
@@ -211,6 +211,26 @@ TEST(SHA3Test, NISTTestVectors_SingleShot) {
   });
 }
 
+TEST(KeccakInternalTest, SqueezeOutputBufferOverflow) {
+    EVP_MD_unstable_sha3_enable(true);
+
+    KECCAK1600_CTX ctx;
+    std::vector<uint8_t> out;
+    std::vector<uint8_t> canary(8);
+    std::fill(canary.begin(), canary.end(), 0xff);
```
torben-hansen, line 220: There is support for doing this "natively" with the address sanitizers. You can "poison" a memory region: https://learn.microsoft.com/en-us/cpp/sanitizers/asan-runtime?view=msvc-170#poisoning.

See https://github.com/aws/aws-lc/blob/main/crypto/mem.c for an example how it's used.

crypto/fipsmodule/sha/keccak1600.c
WillChilds-Klein: will remove.

crypto/fipsmodule/sha/sha3_test.cc
WillChilds-Klein: very nice. is this poisoning only available when run under ASAN? it looks like [the poisoning functions are stubbed out](https://github.com/aws/aws-lc/blob/main/crypto/mem.c#L82) when not run under ASAN, which would reduce our coverage across platforms here...

WillChilds-Klein:
> We had assembly functions prefixed with `hw` (or `aes_hw`, in the case of AES) rather than suffixed with `asm`.

got it, i'll rename the functions. thanks.

crypto/fipsmodule/sha/sha3_test.cc
torben-hansen: It's a fair point. Stick to your approach for something that is not cpu agnostic.

torben-hansen (approved):

PR #894, opened 2023-03-21
Improved support for OpenSSH

justsmth:
### Issues:
N/A

### Description of changes: 
* Added functions/symbols needed for integration with OpenSSH-portable.
* If pkcs11 support is disabled, OpenSSH-portable compiles and its tests pass on Linux when built against this commit of AWS-LC. (For all tests to pass, AWS-LC's `libcrypto.so` must be properly setup for use by the OS.)

**NOTE**: I reverted changes in this PR related to `BN_set_flags` to allow more time for discussion. `BN_set_flags` will need to be re-added in order to support openssh.

### Call-outs:
* Additional `CFLAGS` need to be specified in order for the OpenSSH-portable build to succeed. I submitted a [PR to openssh-portable](https://github.com/openssh/openssh-portable/pull/386) that would eliminate the need to set `CFLAGS` in our case.

### Testing:
* I setup a docker image. I built/installed this commit of AWS-LC. I setup `/etc/ld.so.conf` to find AWS-LC's `libcrypto.so`. I built/installed [OpenSSH portable](https://github.com/openssh/openssh-portable) against AWS-LC. All tests were ran and passed.
```
> cd openssh-portable
> export CFLAGS="-DHAVE_RSA_METH_FREE=1 -DHAVE_RSA_METH_DUP=1 -DHAVE_RSA_METH_SET1_NAME=1 -DHAVE_RSA_METH_SET_PRIV_ENC=1 -DHAVE_RSA_METH_SET_PRIV_DEC=1"
> ./configure --with-ssl-dir=`readlink -f ../AWS-LC-INSTALL` --prefix=`readlink -f ../OPENSSH-INSTALL` --disable-pkcs11
...
> make install
...
> make tests
...
unit tests passed
echo all tests passed
all tests passed
```

the ISC license.


include/openssl/bn.h
```diff
@@ -974,6 +974,12 @@ OPENSSL_EXPORT int BN_bn2binpad(const BIGNUM *in, uint8_t *out, int len);
 // BN_secure_new calls |BN_new|.
 OPENSSL_EXPORT BIGNUM *BN_secure_new(void);
 
+#define BN_FLG_MALLOCED 0x01
+#define BN_FLG_STATIC_DATA 0x02
+#define BN_FLG_CONSTTIME   0x04
```
torben-hansen, line 979: This can imply that code makes false assumptions about the run-time behaviour of AWS-LC. How is that mitigated?

include/openssl/ec.h
```diff
@@ -269,6 +269,23 @@ OPENSSL_EXPORT int EC_POINT_point2cbb(CBB *out, const EC_GROUP *group,
                                       point_conversion_form_t form,
                                       BN_CTX *ctx);
 
+// EC_POINT_point2buf allocates a buffer of suitable length and writes an
+// EC_POINT to it in octet format. The allocated buffer is written to *pbuf and
```
andrewhop, line 273: Is octet the same as X9.62? Is that more precise?

include/openssl/ec.h
```diff
@@ -269,6 +269,23 @@ OPENSSL_EXPORT int EC_POINT_point2cbb(CBB *out, const EC_GROUP *group,
...
+// its length is returned. The caller must free up the allocated buffer with a
+// call to OPENSSL_free(). Since the allocated buffer value is written to *pbuf
+// the pbuf parameter MUST NOT be NULL.
+OPENSSL_EXPORT size_t EC_POINT_point2buf(const EC_GROUP *group,
```
andrewhop, line 277: Outside the scope for this PR, but for this to be useful do we need a `EC_POINT_buf2point`? What would someone do with this serialized buffer?

andrewhop (commented):
Please add unit tests for the new APIs.

crypto/fipsmodule/cipher/cipher.c
```diff
@@ -590,6 +588,20 @@ unsigned EVP_CIPHER_CTX_key_length(const EVP_CIPHER_CTX *ctx) {
   return ctx->key_len;
 }
 
+int EVP_CIPHER_CTX_get_iv(const EVP_CIPHER_CTX *ctx, unsigned char *iv,
+                          size_t len) {
+  if (ctx == NULL || iv == NULL)
+    return 0;
```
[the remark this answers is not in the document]
justsmth: Fixed in the latest commit. I also added [InsertBraces](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces) to our `.clang-format` so this format is more consistent going forward.

crypto/fipsmodule/cipher/cipher.c
```diff
@@ -590,6 +588,20 @@ unsigned EVP_CIPHER_CTX_key_length(const EVP_CIPHER_CTX *ctx) {
...
+
+  if (len > EVP_MAX_IV_LENGTH || len != EVP_CIPHER_CTX_iv_length(ctx))
```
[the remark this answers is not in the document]
justsmth: It shouldn't be possible. I agree this might be a slightly excessive use of "defensive programming".  

The concern is that a client allocates a static buffer of size `EVP_MAX_IV_LENGTH` to use for the IV under the (hopefully correct) assumption that it can never be bigger -- the client might not even bother to verify that this "max" size is honored. If there is a bug elsewhere in our logic that might allow the IV length to be larger than this, we could overflow the client's allocation when we copy the value.

include/openssl/ec.h
justsmth: I agree, there's no need to expose this function. I'll change it to be a `static` function.

crypto/obj/objects.go
```diff
@@ -376,7 +376,7 @@ func writeNumbers(path string, objs *objects) error {
 
 func clangFormat(input string) (string, error) {
 	var b bytes.Buffer
-	cmd := exec.Command("clang-format")
+	cmd := exec.Command("clang-format", "--style=Google")
```
andrewhop, line 379: Do we need this? Our clang-format has `BasedOnStyle: Google`

include/openssl/ec.h
```diff
@@ -269,6 +269,13 @@ OPENSSL_EXPORT int EC_POINT_point2cbb(CBB *out, const EC_GROUP *group,
                                       point_conversion_form_t form,
                                       BN_CTX *ctx);
 
+// EC_POINT_point2bn converts an EC_POINT to BIGNUM. Returns the pointer to the
```
andrewhop, line 272: NP: I like links:
```suggestion
// EC_POINT_point2bn converts an |EC_POINT| to |BIGNUM.| Returns the pointer to the
```

crypto/ec_extra/ec_asn1.c
```diff
@@ -533,3 +533,45 @@ int i2o_ECPublicKey(const EC_KEY *key, uint8_t **outp) {
...
+
+  if (buf_len == 0) {
+    return NULL;
+  }
+
+  ret = BN_bin2bn(buf, buf_len, ret);
+
+  OPENSSL_free(buf);
+
+  return ret;
```
andrewhop, line 576: I realize we have to do it this way but it seems weird that this functions both returns a pointer to the bignum and ret to it.

crypto/obj/objects.go
justsmth: Without this option, the CI fails here due to it not recognizing the "InsertBraces" key I added to .clang-format. That option was only added in clang-format v15.

include/openssl/cipher.h
```diff
@@ -142,6 +142,11 @@ OPENSSL_EXPORT int EVP_CIPHER_CTX_copy(EVP_CIPHER_CTX *out,
 // |EVP_CIPHER_CTX_init| and returns one.
 OPENSSL_EXPORT int EVP_CIPHER_CTX_reset(EVP_CIPHER_CTX *ctx);
 
+// EVP_CIPHER_CTX_get_iv retrieves the IV from |*ctx|, copying it to |iv|. The
+// IV length, |len|, must equal the expected IV length for the context as
+// provided by |EVP_CIPHER_CTX_iv_length|.
+AWS_LC_DEPRECATED OPENSSL_EXPORT int EVP_CIPHER_CTX_get_iv(const EVP_CIPHER_CTX *ctx,
+                                         unsigned char *iv, size_t len);
```
andrewhop, line 149: This should go down in the deprecated function section on line 388.

include/openssl/ec.h
```diff
@@ -269,6 +269,13 @@ OPENSSL_EXPORT int EC_POINT_point2cbb(CBB *out, const EC_GROUP *group,
                                       point_conversion_form_t form,
                                       BN_CTX *ctx);
 
+// EC_POINT_point2bn converts an |EC_POINT| to |BIGNUM.| Returns the pointer to the
+// BIGNUM supplied or NULL on error.
+AWS_LC_DEPRECATED OPENSSL_EXPORT BIGNUM *EC_POINT_point2bn(const EC_GROUP *group,
+                                         const EC_POINT *point,
+                                         point_conversion_form_t form,
+                                         BIGNUM *ret, BN_CTX *ctx);
+
```
andrewhop, line 278: This should go down in the deprecated section around 313. We also still need a test for this function.

include/openssl/ec.h
```diff
@@ -269,6 +269,13 @@ OPENSSL_EXPORT int EC_POINT_point2cbb(CBB *out, const EC_GROUP *group,
                                       point_conversion_form_t form,
                                       BN_CTX *ctx);
 
+// EC_POINT_point2bn converts an |EC_POINT| to |BIGNUM.| Returns the pointer to the
+// BIGNUM supplied or NULL on error.
```
andrewhop, line 273: Does the user need to allocate ret/ctx?

crypto/fipsmodule/cipher/cipher.c
```diff
@@ -590,6 +588,27 @@ unsigned EVP_CIPHER_CTX_key_length(const EVP_CIPHER_CTX *ctx) {
...
+    return 0;
+  }
+
+  if (len > EVP_MAX_IV_LENGTH || len != EVP_CIPHER_CTX_iv_length(ctx)) {
+    return 0;
+  }
+
+  const EVP_CIPHER *cipher = EVP_CIPHER_CTX_cipher(ctx);
+  if (EVP_CIPHER_mode(cipher) == EVP_CIPH_GCM_MODE) {
+    return 0;
```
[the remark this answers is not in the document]
justsmth: This function doesn't exist in OpenSSL, it only exists [in LibreSSL](https://man.openbsd.org/EVP_EncryptInit.3#EVP_CIPHER_CTX_get_iv). OpenSSL 3 has a few [similar but different](https://www.openssl.org/docs/man3.0/man3/EVP_CIPHER_CTX_iv.html) functions for getting the IV. For compatibility, OpenSSH provides an implementation [similar to the one in this PR](https://github.com/openssh/openssh-portable/blob/master/openbsd-compat/libressl-api-compat.c#L30) when it's linking to OpenSSL.

include/openssl/ec.h
justsmth: I updated this description to clarify the use of ret/ctx in the latest commit.

andrewhop (dismissed):

crypto/CMakeLists.txt
```diff
@@ -747,6 +747,8 @@ if(BUILD_TESTING)
   target_link_libraries(${CRYPTO_TEST_EXEC} test_support_lib boringssl_gtest crypto)
   if(WIN32)
     target_link_libraries(${CRYPTO_TEST_EXEC} ws2_32)
+  else()
+    target_compile_options(${CRYPTO_TEST_EXEC} PUBLIC -Wno-deprecated-declarations)
```
skmcgrail, line 751: I assume there is no way to have a suppression file of sorts? Otherwise kind of annoying we need to turn off the warning completely.

crypto/ec_extra/ec_asn1.c
```diff
@@ -533,3 +533,45 @@ int i2o_ECPublicKey(const EC_KEY *key, uint8_t **outp) {
...
+                                 point_conversion_form_t form,
+                                 unsigned char **pbuf, BN_CTX *ctx) {
+  size_t len;
+  unsigned char *buf;
+
+  len = EC_POINT_point2oct(group, point, form, NULL, 0, NULL);
+  if (len == 0) {
+    return 0;
+  }
+  if ((buf = OPENSSL_malloc(len)) == NULL) {
```
skmcgrail, line 547: nit: Assignments and conditions in an if statement in C, while valid, has always felt a bit like a dirty practice in the language. Not sure what our practice has been in the rest of the codebase, but it might be clearer to separate this out.

crypto/err/evp.errordata
```diff
@@ -1,3 +1,4 @@
+EVP,138,BAD_DECRYPT
```
skmcgrail, line 1: Did you add this manually or did you run the `make_errors.go` utility? I'm surprised it doesn't sort them if you use the tool.

include/openssl/pem.h
```diff
@@ -475,5 +475,7 @@ OPENSSL_EXPORT int PEM_write_PKCS8PrivateKey(FILE *fp, const EVP_PKEY *x,
 #define PEM_R_SHORT_HEADER 112
 #define PEM_R_UNSUPPORTED_CIPHER 113
 #define PEM_R_UNSUPPORTED_ENCRYPTION 114
+// For OpenSSL compatibility. Not used internally.
+#define PEM_R_PROBLEMS_GETTING_PASSWORD 115
```
skmcgrail, line 479: Should run `make_errors.go` here too.

crypto/err/evp.errordata
justsmth: This is the output of that utility. (This caused me a little trouble -- the CI is setup to fail unless this file *exactly* matches that output.)

crypto/ec_extra/ec_asn1.c
```diff
@@ -533,3 +533,45 @@ int i2o_ECPublicKey(const EC_KEY *key, uint8_t **outp) {
   // Historically, this function used the wrong return value on error.
   return ret > 0 ? ret : 0;
 }
+
+static size_t EC_POINT_point2buf(const EC_GROUP *group, const EC_POINT *point,
+                                 point_conversion_form_t form,
+                                 unsigned char **pbuf, BN_CTX *ctx) {
```
torben-hansen, line 539: use `uint8_t` not `unsigned char`

crypto/ec_extra/ec_asn1.c
```diff
@@ -533,3 +533,45 @@ int i2o_ECPublicKey(const EC_KEY *key, uint8_t **outp) {
...
+  size_t len;
+  unsigned char *buf;
```
torben-hansen, line 541: `uint8_t`

crypto/ec_extra/ec_asn1.c
```diff
@@ -533,3 +533,45 @@ int i2o_ECPublicKey(const EC_KEY *key, uint8_t **outp) {
...
+  }
+  *pbuf = buf;
+  return len;
+}
+
+BIGNUM *EC_POINT_point2bn(const EC_GROUP *group, const EC_POINT *point,
+                          point_conversion_form_t form, BIGNUM *ret,
+                          BN_CTX *ctx) {
+  size_t buf_len = 0;
+  unsigned char *buf;
```
torben-hansen, line 564: `uint8_t`

skmcgrail (dismissed):

crypto/fipsmodule/ec/ec_test.cc
```diff
@@ -1036,6 +1037,15 @@ TEST(ECTest, ArbitraryCurve) {
   EXPECT_EQ(0, EC_GROUP_cmp(group.get(), group.get(), NULL));
   EXPECT_EQ(0, EC_GROUP_cmp(group2.get(), group.get(), NULL));
 
+  bssl::UniquePtr<BIGNUM> converted_generator1(EC_POINT_point2bn(
+      group.get(), generator.get(), POINT_CONVERSION_UNCOMPRESSED, NULL, NULL));
```
torben-hansen, line 1041: An elliptic curve point can be encoded as a compressed point as well. The enum is `POINT_CONVERSION_COMPRESSED`. Worth doing that path as well?

skmcgrail (approved):

PR #895, opened 2023-03-21
Revert back to using bssl

samuel40791765:
### Issues:
Resolves `P84022616`

### Description of changes: 
Internal imports are finding broken symlinks. CMake doesn't seem to support relative path symlinks, so directly removing it and using `awslc` from now should resolve our issues. 

### Call-outs:
N/A

### Testing:
N/A

the ISC license.


justsmth:
Several of our test scripts still use the name "bssl": https://github.com/aws/aws-lc/search?l=Shell&q=bssl

nebeid:
> Should we consider changing `awslc` back to `bssl`? Or would the better approach be to reconfigure our test scripts to use `awslc`?

I am leaning towards the first suggestion.

torben-hansen (approved):
lgtm - but remember to change the commit text before merging :)

andrewhop (approved):

PR #901, opened 2023-03-23
Add support for OCSP request nonce functions

samuel40791765:
### Issues:
Addresses `CryptoAlg-1564`

### Description of changes: 
Additional OCSP supporting functions for adding nonces to OCSP requests.

### Call-outs:
* Support for OCSP nonce parsing was taken out of `X509V3_EXT_METHOD`, so we needed to add support back in `crypto/x509v3/v3_ocsp.c`. OCSP nonces cannot be properly added without this code.
* Only two new external functions are `OCSP_request_add1_nonce` and `OCSP_check_nonce`
### Testing:
New unit tests for nonce comparison cases and adding new nonces.

the ISC license.


crypto/ocsp/ocsp_test.cc
```diff
@@ -1115,3 +1121,91 @@ TEST_P(OCSPURLTest, OCSPParseURL) {
...
+    // OCSP request with no nonce, but an OCSP response that does have one.
+    {"ocsp_request_no_nonce", "ocsp_response", false, nullptr, 0,
+     OCSP_NONCE_RESPONSE_ONLY},
+    // OCSP request with no nonce, but an OCSP response that does have one.
+    {"ocsp_request", "ocsp_response_no_nonce", false, nullptr, 0,
+     OCSP_NONCE_REQUEST_ONLY},
+    // OCSP request with no nonce, but an OCSP response that does have one.
+    {"ocsp_request_no_nonce", "ocsp_response_no_nonce", false, nullptr, 0,
+     OCSP_NONCE_BOTH_ABSENT},
+};
```
skmcgrail, line 1159: comment don't seem to match the test cases

samuel40791765: Got caught with the copy and paste, and forgot to update

crypto/ocsp/ocsp_test.cc
```diff
@@ -39,6 +39,12 @@ static const time_t invalid_after_ocsp_expire_time_sha256 = 1937505764;
 #define OCSP_URL_PARSE_SUCCESS 1
 #define OCSP_URL_PARSE_ERROR 0
 
+#define OCSP_NONCE_EQUAL 1
+#define OCSP_NONCE_BOTH_ABSENT 2
+#define OCSP_NONCE_RESPONSE_ONLY 3
+#define OCSP_NONCE_NOT_EQUAL 0
+#define OCSP_NONCE_REQUEST_ONLY -1
```
skmcgrail, line 46: not a blocker: But I'm wondering if the user experience would better if these were just public anyways? Then you can use them in the `OCSP_check_nonce` implementation as well.

skmcgrail (approved):

crypto/ocsp/ocsp_test.cc
samuel40791765: Good idea, I'd originally created these with only testing in mind, but these totally make sense in a broader context as well.

crypto/ocsp/ocsp_extension.c
```diff
@@ -0,0 +1,113 @@
...
+    goto err;
+  }
+  ret = 1;
+err:
+  OPENSSL_free(os.data);
+  return ret;
+}
+
+int OCSP_request_add1_nonce(OCSP_REQUEST *req, unsigned char *val, int len) {
+  return ocsp_add_nonce(&req->tbsRequest->requestExtensions, val, len);
```
andrewhop, line 78: From our recent retro this should check `req` is not null before accessing the internal parts of the struct.

crypto/ocsp/ocsp_extension.c
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  os.data = OPENSSL_malloc(os.length);
+  if (os.data == NULL) {
+    goto err;
+  }
+  tmpval = os.data;
+  ASN1_put_object(&tmpval, 0, len, V_ASN1_OCTET_STRING, V_ASN1_UNIVERSAL);
+  if (val != NULL) {
+    memcpy(tmpval, val, len);
```
andrewhop, line 63: This should use `OPENSSL_memcpy`. I also think this should be clarified in the documentation. What if someone passes in a 2 byte nonce and 0 for len, this would copy OCSP_DEFAULT_NONCE_LENGTH and read past the end of `val`. Is this a case where we should make our API a bit safer and require:
1. val == null and len can be 0 for default or a particular nonce size
2. val != null and len must be the length of val
Letting val !- null and picking a different len as default seems very dangerous.

crypto/ocsp/ocsp_extension.c
```diff
@@ -0,0 +1,113 @@
...
+err:
+  OPENSSL_free(os.data);
+  return ret;
+}
+
+int OCSP_request_add1_nonce(OCSP_REQUEST *req, unsigned char *val, int len) {
+  return ocsp_add_nonce(&req->tbsRequest->requestExtensions, val, len);
+}
+
+int OCSP_check_nonce(OCSP_REQUEST *req, OCSP_BASICRESP *bs) {
```
andrewhop, line 81: This is an interesting use case for the retro null check decision. Do we do the null checks here even though the values aren't used (only passed to OCSP_REQUEST_get_ext_by_NID) or: 
* OCSP_REQUEST_get_ext_by_NID
* X509v3_get_ext_by_NID
* X509v3_get_ext_by_OBJ
X509v3_get_ext_by_OBJ does already have a null check, so all use cases are "safe".

crypto/x509v3/v3_ocsp.c
```diff
@@ -68,6 +94,63 @@ static int i2r_ocsp_acutoff(const X509V3_EXT_METHOD *method, void *cutoff,
...
 
+// OCSP nonce. This is needs special treatment because it doesn't have an
+// ASN1 encoding at all: it just contains arbitrary data.
+
+static void *ocsp_nonce_new(void) { return ASN1_OCTET_STRING_new(); }
+
+static int i2d_ocsp_nonce(void *a, unsigned char **pp) {
+  ASN1_OCTET_STRING *os = a;
+  if (pp != NULL) {
+    memcpy(*pp, os->data, os->length);
```
andrewhop, line 105: `OPENSSL_memcpy`

include/openssl/ocsp.h
```diff
@@ -160,6 +160,26 @@ OPENSSL_EXPORT int OCSP_REQ_CTX_add1_header(OCSP_REQ_CTX *rctx,
...
 
+// OCSP_request_add1_nonce adds a nonce of value |val| and length |len| to
+// |req|. If |val| is NULL, a random nonce is generated and used. If |len| is
+// zero or negative, a default length of 16 bytes will be used.
+OPENSSL_EXPORT int OCSP_request_add1_nonce(OCSP_REQUEST *req,
+                                           unsigned char *val, int len);
+
+// OCSP_check_nonce checks nonce existence and equality in |req| and |bs|.
+//
+// Return value reflects result:
```
andrewhop, line 171: What about an error? Such as a null req or bs, or parsing issue?

andrewhop (approved):

skmcgrail (approved):

PR #903, opened 2023-03-24
Add support for OCSP_REQUEST_print and OCSP_RESPONSE_print

samuel40791765:
### Issues:
Resolves `CryptoAlg-1564`

### Description of changes: 
Once https://github.com/aws/aws-lc/pull/901 is merged in, these should be the only OCSP functions left we'll need to support for the time being.

### Call-outs:
N/A

### Testing:
N/A

the ISC license.

### DIFF
--- a/crypto/ocsp/ocsp_print.c
+++ b/crypto/ocsp/ocsp_print.c
@@ -8,8 +8,24 @@
  */
 
 #include <openssl/ocsp.h>
+#include <openssl/pem.h>
 
 #include "internal.h"
+
+static int ocsp_certid_print(BIO *bp, OCSP_CERTID *certid, int indent) {
+  BIO_printf(bp, "%*sCertificate ID:\n", indent, "");
+  indent += 2;
+  BIO_printf(bp, "%*sHash Algorithm: ", indent, "");
+  i2a_ASN1_OBJECT(bp, certid->hashAlgorithm->algorithm);
+  BIO_printf(bp, "\n%*sIssuer Name Hash: ", indent, "");
+  i2a_ASN1_STRING(bp, certid->issuerNameHash, 0);
+  BIO_printf(bp, "\n%*sIssuer Key Hash: ", indent, "");
+  i2a_ASN1_STRING(bp, certid->issuerKeyHash, 0);
+  BIO_printf(bp, "\n%*sSerial Number: ", indent, "");
+  i2a_ASN1_INTEGER(bp, certid->serialNumber);
+  BIO_printf(bp, "\n");
+  return 1;
+}
 
 typedef struct {
   long t;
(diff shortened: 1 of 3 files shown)


include/openssl/ocsp.h
```diff
@@ -250,6 +270,24 @@ OPENSSL_EXPORT const char *OCSP_response_status_str(long status_code);
...
 OPENSSL_EXPORT const char *OCSP_cert_status_str(long status_code);
 
+// OCSP_crl_reason_str returns the OCSP CRL reason of |status_code| as a string.
+// |OCSP_resp_find_status| can be used to retrieve the reason status code
+// if an OCSP response is revoked.
+OPENSSL_EXPORT const char *OCSP_crl_reason_str(long status_code);
+
+// OCSP_REQUEST_print prints the contents of an OCSP request to |bp|. |flags| is
+// used to configure printing of the |req|'s extensions (See
+// |X509V3_extensions_print| for more information).
```
andrewhop, line 360: 10/10 link

crypto/ocsp/ocsp_print.c
```diff
@@ -62,3 +78,190 @@ const char *OCSP_crl_reason_str(long s) {
...
 }
+
+int OCSP_REQUEST_print(BIO *bp, OCSP_REQUEST *req, unsigned long flags) {
+  long l;
+  OCSP_CERTID *cid = NULL;
+  OCSP_ONEREQ *one = NULL;
+  OCSP_REQINFO *inf = req->tbsRequest;
+  OCSP_SIGNATURE *sig = req->optionalSignature;
+
+  if (BIO_write(bp, "OCSP Request Data:\n", 19) <= 0) {
```
andrewhop, line 89: What is the difference between BIO_write, BIO_printf, and BIO_puts? Why do you alternate between them in this function?

crypto/ocsp/ocsp_print.c
```diff
@@ -62,3 +78,190 @@ const char *OCSP_crl_reason_str(long s) {
...
+  int ret = 0;
+  long l;
+  OCSP_CERTID *cid = NULL;
+  OCSP_BASICRESP *br = NULL;
+  OCSP_RESPID *rid = NULL;
+  OCSP_RESPDATA *rd = NULL;
+  OCSP_CERTSTATUS *cst = NULL;
+  OCSP_REVOKEDINFO *rev = NULL;
+  OCSP_SINGLERESP *single = NULL;
+  OCSP_RESPBYTES *rb = resp->responseBytes;
```
andrewhop, line 140: Missing null check on parameters, and this is a potential null de-reference.

crypto/ocsp/ocsp_print.c
```diff
@@ -62,3 +78,190 @@ const char *OCSP_crl_reason_str(long s) {
...
+
+  if (BIO_puts(bp, "OCSP Response Data:\n") <= 0) {
```
andrewhop, line 142: I assume yes, but does BIO_puts and the other bio functions handle a null bp?

crypto/ocsp/ocsp_print.c
```diff
@@ -62,3 +78,190 @@ const char *OCSP_crl_reason_str(long s) {
   size_t tbl_size = (sizeof(reason_tbl) / sizeof((reason_tbl)[0]));
   return do_table2string(s, reason_tbl, tbl_size);
 }
+
+int OCSP_REQUEST_print(BIO *bp, OCSP_REQUEST *req, unsigned long flags) {
+  long l;
+  OCSP_CERTID *cid = NULL;
+  OCSP_ONEREQ *one = NULL;
+  OCSP_REQINFO *inf = req->tbsRequest;
```
andrewhop, line 86: Missing null checks.

andrewhop (commented):
Please add a known answer style test so we can verify the final output is the expected format.

samuel40791765 (commented):
Tests added

crypto/ocsp/ocsp_print.c
```diff
@@ -62,3 +78,198 @@ const char *OCSP_crl_reason_str(long s) {
...
+      X509_print(bp, sk_X509_value(sig->certs, i));
+      PEM_write_bio_X509(bp, sk_X509_value(sig->certs, i));
+    }
+  }
+  return 1;
+}
+
+int OCSP_RESPONSE_print(BIO *bp, OCSP_RESPONSE *resp, unsigned long flags) {
+  if(bp == NULL|| resp ==NULL) {
+    OPENSSL_PUT_ERROR(OCSP, ERR_R_PASSED_NULL_PARAMETER);
```
andrewhop, line 135: This raises an interesting question: is the semantics of printing a null response printing the word "null response" to the bio (and returning success), or returning a null parameter error (and returning an error)?

crypto/ocsp/ocsp_test.cc
```diff
@@ -1380,3 +1380,66 @@ TEST(OCSPTest, CertIDDup) {
...
+      LoadOCSP_RESPONSE(ocsp_response_data);
+
+  bssl::UniquePtr<BIO> bio(BIO_new(BIO_s_mem()));
+  EXPECT_TRUE(OCSP_RESPONSE_print(bio.get(), ocspResponse.get(), 0));
+  char *out;
+  ASSERT_TRUE(BIO_get_mem_data(bio.get(), &out));
+
+  // Iterate through |kExpected| and verify that |OCSP_RESPONSE_print| has
+  // the expected format.
+  char *token = strtok(out, "\n");
```
andrewhop, line 1408: Since this is C++ you could avoid strtok and use some safer options like [std::istringstream](https://en.cppreference.com/w/cpp/io/basic_istringstream):
```
 std::istringstream out_stream(string(out));
 std::string line;
 while (std::getline(out_stream, line)) {
    // Check they're equal
}
```

crypto/ocsp/ocsp_test.cc
```diff
@@ -1380,3 +1380,66 @@ TEST(OCSPTest, CertIDDup) {
   // Check that the duplicated structure is not just a replicated pointer.
   EXPECT_NE(orig_id.get(), dup_id.get());
 }
+
+TEST(OCSPTest, OCSPResponsePrint) {
+  static const std::string kExpected[] = {
```
andrewhop, line 1385: This could be a C++ `std::array` which has  a nice `size()` function.

crypto/ocsp/ocsp_test.cc
```diff
@@ -1380,3 +1380,66 @@ TEST(OCSPTest, CertIDDup) {
...
+  ASSERT_TRUE(BIO_get_mem_data(bio.get(), &out));
+
+  // Iterate through |kExpected| and verify that |OCSP_RESPONSE_print| has
+  // the expected format.
+  char *token = strtok(out, "\n");
+  for (size_t i = 0; i < sizeof(kExpected) / sizeof(kExpected[0]); i++) {
+    EXPECT_EQ(OPENSSL_memcmp(token, kExpected[i].c_str(),
+                             strlen(kExpected[i].c_str())),
+              0);
+    token = strtok(nullptr, "\n");
```
andrewhop, line 1413: This is why I don't lile strtok, the weird hidden state is confusing. Especially because these tests are run in parallel, and can strtok handle that?

crypto/ocsp/ocsp_test.cc
```diff
@@ -1380,3 +1380,66 @@ TEST(OCSPTest, CertIDDup) {
...
+  bssl::UniquePtr<BIO> bio(BIO_new(BIO_s_mem()));
+  EXPECT_TRUE(OCSP_REQUEST_print(bio.get(), ocspRequest.get(), 0));
+  char *out;
+  ASSERT_TRUE(BIO_get_mem_data(bio.get(), &out));
+
+  // Iterate through |kExpected| and verify that |OCSP_REQUEST_print| has
+  // the expected format.
+  char *token = strtok(out, "\n");
+  for (size_t i = 0; i < sizeof(kExpected) / sizeof(kExpected[0]); i++) {
+    EXPECT_EQ(OPENSSL_memcmp(token, kExpected[i].c_str(),
```
andrewhop, line 1440: Does expect_eq work on c++ strings? Do you have to check memcmp is equal to 0?

crypto/ocsp/ocsp_test.cc
samuel40791765: Changed to use std::istringstream instead, thanks for the tip :)

crypto/ocsp/ocsp_print.c
samuel40791765: Good points, I'll leave it the way it is since it doesn't really make sense to pass nothing to print in the first place.
We're doing better than OpenSSL is here anyways, they would segfault if a NULL is passed in.

crypto/ocsp/ocsp_test.cc
samuel40791765: I assumed strtok's non-thread safeness meant that it couldn't handle `strtok` the same string across multiple threads? We're running against the same string in the same thread, so I didn't think this would be an issue.
I've changed the implementation to use `istringstream` though, like your suggestion.

crypto/ocsp/ocsp_test.cc
andrewhop: You're checking that Expect_eq(OPENSSL_memcmp(string1, string2), 0), I'm wondering can you Expect_eq(string1, strign2). This has the added benefit that if this test ever fails we get a better message "foo bar" does not match "bar foo" instead of got 1 expected 0.

andrewhop (approved):

skmcgrail (approved):

PR #904, opened 2023-03-24
Add missing APIs to better support Bind

andrewhop:
### Description of changes: 
* BN_GENCB_set_old sets old style callbacks on BN_GENCB
* DH_clear_flags which is a no op
* Define but don't support DH_FLAG_CACHE_MONT_P, AWS-LC always uses montgomery multiplication for DH

### Call-outs:
I considered using the existing [callback_wrapper](https://github.com/aws/aws-lc/blob/main/crypto/decrepit/dh/dh_decrepit.c#L57-L68) pattern used in DH_generate_parameters. However, that function can keep the `wrapped_callback` on the stack and not worry about memory management. For this use case users can call `BN_GENCB_set_old` directly. This would require allocating  `wrapped_callback` on the heap, but they don't expect to have to clean up any memory, and BN_GENCB_free wouldn't be able to tell what if anything it has to free. 

Therefore, using a union and `type` in bn_gencb_st seemed like the easiest path forward. All of this is internal implementation details we could change if we need to simplify/reduce the size of `bn_gencb_st`.

### Testing:
Added a new unit test that ensures the expected callback is called. This PR does not add all the required changes for Bind so I haven't added an integration test that attempts to build Bind with AWS-LC yet.

the ISC license.

### DIFF
--- a/crypto/rsa_extra/rsa_test.cc
+++ b/crypto/rsa_extra/rsa_test.cc
@@ -1259,6 +1259,50 @@ TEST(RSATest, KeygenInternalRetry) {
   EXPECT_TRUE(RSA_generate_key_ex(rsa.get(), 2048, e.get(), &cb));
 }
 
+TEST(RSATest, OldCallback) {
+  bssl::UniquePtr<RSA> rsa(RSA_new());
+  ASSERT_TRUE(rsa);
+  BN_GENCB cb;
+
+  int old_callback_call_count = 0;
+  void (*old_style_callback)(int, int, void *) = [](int event, int n,
+                                                   void *ptr) -> void {
+    BN_GENCB *cb_ptr = static_cast<BN_GENCB *>(ptr);
+    int *count_ptr = static_cast<int *>(cb_ptr->arg);
+    *count_ptr += 1;
+  };
+
+  int new_callback_call_count = 0;
+  int (*new_style_callback)(int, int, BN_GENCB *cb_ptr) = [](int event, int n,
+                                                    BN_GENCB *cb_ptr) -> int {
+    int *count_ptr = static_cast<int *>(cb_ptr->arg);
+    *count_ptr += 1;
+    return 1;
+  };
+
+  // Set the new style first, setting the old should overwrite the new callback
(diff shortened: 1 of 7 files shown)


include/openssl/bn.h
```diff
@@ -684,8 +688,10 @@ OPENSSL_EXPORT BN_GENCB *BN_GENCB_new(void);
 // BN_GENCB_free releases memory associated with |callback|.
 OPENSSL_EXPORT void BN_GENCB_free(BN_GENCB *callback);
 
-// BN_GENCB_set configures |callback| to call |f| and sets |callout->arg| to
-// |arg|.
+// BN_GENCB_set configures |callback| to call |f| and sets |callback->arg| to
```
dkostic, line 691: should we call-out that this one sets `new_style` callback?

crypto/rsa_extra/rsa_test.cc
```diff
@@ -1259,6 +1259,50 @@ TEST(RSATest, KeygenInternalRetry) {
...
+  EXPECT_EQ(new_callback_call_count, 0);
+  int previous_count = old_callback_call_count;
+
+  // Set the new style again and overwrite the old style
+  BN_GENCB_set(&cb, new_style_callback, &new_callback_call_count);
+  EXPECT_TRUE(RSA_generate_key_ex(rsa.get(), 2048, e.get(), &cb));
+  // The old callback should have been overwritten and not called
+  EXPECT_EQ(previous_count, old_callback_call_count);
+  // The new callback should have been called at least once
+  EXPECT_NE(new_callback_call_count, 0);
```
samuel40791765, line 1303: Is there a hard set number on what `new_callback_call_count/old_callback_call_count` is each time?

samuel40791765 (approved):

dkostic (approved):

crypto/rsa_extra/rsa_test.cc
andrewhop: No, it varies based on how lucky you get during RSA key generation. Generate a random big number, check if it's prime, call the callback. So depending on how many guess it takes the callback gets called a different number of times. In this test it's somewhere around 1,500-ish times.

PR #906, opened 2023-03-28
Disable 3DES by default

torben-hansen:
### Issues:

P84189141

### Description of changes: 

Disable `3DES` from default cipher suite list. The TLS version specific keyword rules still add `3DES` when appropriate (and that is TLS 1 and TLS 1.1).

Some keyword rules map to the default rule `ALL`. Instead of hard-coding these into the cipher suite mapping structure, we factor them out and just keep the default `ALL` rule in the structure. Future keywords that should map to default can then be added to a separate list.

### Call-outs:

Tests assumed 3DES for various tests categories... So, I went ahead and implemented a new post-processing detection. If a test config specifies that either 3DES or the NULL-cipher should be available or used, it would be highlighted in either the `CipherSuites` field or `renegotiateCiphers` field. If that is the case, populate the `cipher` options with the corresponding cipher suite. This option is given to `bssl_shim` when running tests that will then enable the cipher suites on the server when needed.

### Testing:

Disabled one tests since it requires a new object structure that is harder to generate. We don't use the handoff feature anyhow, so it's not a big deal to handle this a bit later on.

The change to the `SSLTest.ClientHello` test fixtures is because the `3DES` is not longer advertised, so it is removed from the cipher suite list in the client hello message.

the ISC license.


ssl/ssl_cipher.cc
```diff
@@ -1193,6 +1192,25 @@ static bool ssl_cipher_process_rulestr(const char *rule_str,
...
+static bool is_known_default_alias_keyword_filter_rule(const char *rule,
+  size_t *matched_rule_length) {
+
+  for (auto known_rule : kKnownKeywordFilterRulesMappingToDefault) {
+    if (strncmp(rule, known_rule, strlen(known_rule)) == 0) {
+      *matched_rule_length = (size_t) strlen(known_rule);
+      return true;
+    }
+  }
+  return false;
```
skmcgrail, line 1211: Should we just set `matched_rule_length` to zero here as well as just a safety? Will enforce that it is always initialized to some value after the call.

ssl/ssl_test.cc
```diff
@@ -5831,7 +5831,7 @@ TEST(SSLTest, SigAlgsList) {
   }
 }
 
-TEST(SSLTest, ApplyHandoffRemovesUnsupportedCiphers) {
+TEST(SSLTest, DISABLED_ApplyHandoffRemovesUnsupportedCiphers) {
```
skmcgrail, line 5834: You both modify and disable this test? Seems like it should be one or the other? Kind of confusing to change the test and disable it.

ssl/test/runner/runner.go
```diff
@@ -19571,6 +19567,52 @@ func checkTests() {
...
+func cipherSuitesContains(cipherSuites []uint16, id uint16) bool {
+	for _, cipherSuite := range cipherSuites {
+		if cipherSuite == id {
+			return true
+		}
+	}
+	return false
+}
+
+func flagsContains(flags []string, subStr string) bool {
```
skmcgrail, line 19579: nit: I'd call this `stringSliceContains`. Similar to the earlier comment.

ssl/test/runner/runner.go
```diff
@@ -19571,6 +19567,52 @@ func checkTests() {
 	}
 }
 
+func cipherSuitesContains(cipherSuites []uint16, id uint16) bool {
```
skmcgrail, line 19570: nit: I'd call this `uint16SliceContains`. Just so that someone doesn't implement another `[]uint16` search in the future.

ssl/test/runner/runner.go
```diff
@@ -19571,6 +19567,52 @@ func checkTests() {
...
+		}
+
+		if add_disabld_cipher_suite && flagsContains(testCases[index].flags, "-cipher") {
+			panic(fmt.Sprintf("Test %q duplicates -cipher in arguments", testCases[index].name))
+		}
+
+		if add_disabld_cipher_suite {
+			testCases[index].flags = append(testCases[index].flags, "-cipher", "DEFAULT" + disabled_ciphers)
+		}
+	}
```
skmcgrail, line 19613: I'd tweak it like this:

```go
for index, tc := range testCases {
	var add_disabld_cipher_suite bool
	var disabled_ciphers string
	if cipherSuitesContains(tc.config.CipherSuites, TLS_RSA_WITH_NULL_SHA) ||
		cipherSuitesContains(tc.renegotiateCiphers, TLS_RSA_WITH_NULL_SHA) {
		add_disabld_cipher_suite = true
		disabled_ciphers = ":NULL-SHA"
	}
	if cipherSuitesContains(tc.config.CipherSuites, TLS_RSA_WITH_3DES_EDE_CBC_SHA) ||
		cipherSuitesContains(tc.renegotiateCiphers, TLS_RSA_WITH_3DES_EDE_CBC_SHA) {
		add_disabld_cipher_suite = true
		disabled_ciphers = disabled_ciphers + ":3DES"
	}
	if add_disabld_cipher_suite && flagsContains(tc.flags, "-cipher") {
		panic(fmt.Sprintf("Test %q duplicates -cipher in arguments", testCases[index].name))
	}
	if add_disabld_cipher_suite {
		tc.flags = append(tc.flags, "-cipher", "DEFAULT" + disabled_ciphers)
		testCases[index] = tc
	}
}
```

ssl/test/runner/runner.go
```diff
@@ -19571,6 +19567,52 @@ func checkTests() {
...
+			disabled_ciphers = ":NULL-SHA"
+		}
+		if cipherSuitesContains(testCases[index].config.CipherSuites, TLS_RSA_WITH_3DES_EDE_CBC_SHA) ||
+			cipherSuitesContains(testCases[index].renegotiateCiphers, TLS_RSA_WITH_3DES_EDE_CBC_SHA) {
+			add_disabld_cipher_suite = true
+			disabled_ciphers = disabled_ciphers + ":3DES"
+		}
+
+		if add_disabld_cipher_suite && flagsContains(testCases[index].flags, "-cipher") {
+			panic(fmt.Sprintf("Test %q duplicates -cipher in arguments", testCases[index].name))
```
skmcgrail, line 19607: Style wise you should return an `error` from this function here, and then maybe panic in the main function if this returns a error from this function. Usually reserve panics for as high in the stack as possible.

ssl/ssl_test.cc
torben-hansen: Yeah I forgot that change.

andrewhop (commented):
Please add an explicit test that `ALL` includes the expected cipher suites and not 3DES.

ssl/ssl_cipher.cc
```diff
@@ -1267,12 +1285,14 @@ bool ssl_create_cipher_list(UniquePtr<SSLCipherPreferenceList> *out_cipher_list,
   // If the rule_string begins with DEFAULT, apply the default rule before
   // using the (possibly available) additional rules.
```
andrewhop, line 1286: Out of date comment.

skmcgrail (dismissed):

andrewhop (dismissed):

andrewhop (approved):

PR #913, opened 2023-03-29
Fix handling of EXFLAG_INVALID_POLICY on the leaf.

andrewhop:
X509_policy_check returns -1 if some certificate had an unparseable extension, in which case it sets EXFLAG_INVALID_POLICY on it. The calling code then iterates over the certificates to find the offending one, so the callback has a chance to undo it. But it skips i = 0, the leaf, and instead just silentely returns success.

We really should cut down on the callback's ability to mess things up here but, in the meantime, fix this. Also add a test covering this case.

While I'm here, I've updated make_invalid_extensions.go, which I pulled some code from, to rename fooOrPanic to mustFoo. That seems to be the convention in the Go standard library. (regexp.MustCompile, etc.)

Change-Id: Ib07c9f4175e66483bd7c0f7d49aea931bf36e53f Reviewed-on: https://boringssl-review.googlesource.com/c/boringssl/+/55748
Auto-Submit: David Benjamin <davidben@google.com>
Reviewed-by: Bob Beck <bbe@google.com>
Commit-Queue: David Benjamin <davidben@google.com>
(cherry picked from commit d1b20a9580aebb6fbb0b1b2408cf1221d83afb71)

### Issues:
Resolves P84593411

### Description of changes: 
This takes a recent change from AWS-LC main and ports it to the fips-2021-10-20 to address CVE-2023-0465.

### Call-outs:
There were some conflicts due to other changes since 2021, I resolved the conflicts while keeping the spirit of the original change.

### Testing:
This change adds new tests and passes locally.

the ISC license.


crypto/x509/x509_test.cc
```diff
@@ -4156,3 +4156,93 @@ TEST(X509Test, TrustedFirst) {
                X509_VERIFY_PARAM_clear_flags(param, X509_V_FLAG_TRUSTED_FIRST);
              }));
 }
+
+TEST(X509Test, Policy) {
```
torben-hansen, line 4160: Some of these are duplicating tests from https://github.com/aws/aws-lc/pull/905.

See [test fixture](https://github.com/aws/aws-lc/blob/fips-2021-10-20/crypto/x509/x509_test.cc#L1789) `X509Test.PolicyTreeBasic`.

PR #927, opened 2023-04-04
Let service indicator check for tls1.0/1.1 context

samuel40791765:
### Issues:
Resolves `CryptoAlg-1543`
Resolves `CryptoAlg-1744`

### Description of changes: 
OpenSSL's way of doing this is checking the label and throwing an error when the label "master secret"  is used. This is because "master secret" implies that the PRF is used in a TLS1.0/1.1 context. In TLS1.2, the label is expected to be "extended master secret". However, they only check if "master secret" is used instead of checking if "extended master secret" is used, because the PRF can be used for other purposes as well.
For our service indicator, we allow for "master secret" to be used in a TLS1.0/1.1 context, but disallow it when TLS1.2 is used. We use the label and the digests being used as an indicator of what TLS context the PRF is being used in.

* Commit: https://github.com/openssl/openssl/commit/50ea5cdcb735916591e35a04c1f5a659bf253ddc
* Internal Usage examples: https://github.com/aws/aws-lc/blob/fb42ac63b765d1be7bb8cb5537bbd31c6c906976/crypto/fipsmodule/self_check/self_check.c#L1130-L1158
* https://github.com/aws/aws-lc/blob/fb42ac63b765d1be7bb8cb5537bbd31c6c906976/util/fipstools/acvp/acvptool/subprocess/kdf-components.go#L198-L212

### Call-outs:
See under `CryptoAlg-1744`

### Testing:
New unit tests

the ISC license.


jvdsn:
> This is because "master secret" implies that the PRF is used in a TLS1.0/1.1 context. In TLS1.2, the label is expected to be "extended master secret".

Well, according to CMVP, "The TLS 1.0 and TLS 1.1 KDFs can also use either the master secret or the extended master secret as input into the HMAC-MD5/HMAC-SHA-1 PRF described in Section 4.2.1 of SP 800-135rev1." Looking at the extended master secret RFC, it seems like it only updates the TLS 1.2 RFC, but I wouldn't just disregard CMVP here.

Instead, what I would do is the following: `TLSKDF_verify_service_indicator` can always distinguish between 1.0/1.1 and 1.2 by looking at the hash function: if the hash function is MD5-SHA1, it must be 1.0/1.1. If the hash function is SHA-256, SHA-384, or SHA-512, it must be 1.2. If the hash function is anything else, it is non-approved.

Now if `TLSKDF_verify_service_indicator` knows the 1.2 KDF is used, it must check for the extended master secret: if this is present, the function is approved. Otherwise, the function is not approved.

skmcgrail (approved):

PR #721, opened 2022-12-08
AES-GCM optimised for AArch64 with EOR3 instruction and a wide AES/PMULL pipeline - 8x unrolling

nebeid:
### Issues:
Resolves CryptoAlg-1015

### Description of changes:
This is an integration of Armv8 optimised implementation of AES-GCM which employs stitching (interleaving) of the AES instructions with the GHASH calculation, which operates on 8 input blocks in parallel.
The code was first contributed to OpenSSL in https://github.com/openssl/openssl/commit/954f45ba4c50457
[PerformanceImprovement.md](https://github.com/aws/aws-lc/files/10900781/PerformanceImprovement.md)
0206ff5bed811e512cf92dc8e by xiaokang.qian@arm.com in Jan, 2022.
It was cherry-picked into 3.1.0 by the end of 2022 in https://github.com/openssl/openssl/commit/34ca334e5de6837f2c6bc0b0b0df28bdd237e4d7.

The code is targeting the Arm Neoverse V1 and Apple M1's architectures since it relies on 2 HW capabilities: SHA3 extension, particularly the 3-way XOR instruction, EOR3, and a wider ASIMD pipeline which processes the AES and PMULL instructions, the latter being the one used in the GHASH calculation.

The diff between the added file `aes-gcm-armv8-unroll8.pl` and its original version in OpenSSL is attached in the file `aes-gcm-armv8-unroll8.pl-openssl3.1.0-beta.diff`

### Call-outs:
Using ARMV8_NEOVERSE_V1 and ARMV8_APPLE_M1 as bits in OPENSSL_armcap_P is different from the original code:
* It tests for specific micro-architectures known to have a wide crypto, i.e. AES/PMULL pipeline.
* It replaces the bit ARMV8_AES_GCM_UNROLL8 which designated both a wide pipeline and the availability of SHA3 extension.

### Testing:
- CI
- AWS Graviton 3 EC2 instance
- Apple M1

### Size increase:
Measured in `tool/bssl` when using the same `speed.cc` file on both the main and PR branches:
27 kB

### Performance Improvement:
Numbers are in **MB/s**
#### EVP API
| Algorithm | Operation | Bytes | AWS-LC PR#721 | AWS-LC main | ratio to main |
| --- | --- | --- | --- | --- | --- |
| EVP-AES-128-GCM | Encrypt | 16    | 90.7   | 99.5   | 0.91 |
|                 |         | 128   | 648.6  | 692.4  | 0.94 |
[...]

PR #939, opened 2023-04-08
Wire up s2n bignum x25519 backend

torben-hansen:
### Issues:

CryptoAlg-1602

### Description of changes: 

PR chain:
* [DONE] PR1: factor out C-level implementation
* [DONE] PR2: refator backend implementation selection business logic to make it more sane
* [THIS PR] PR3: hook-up s2n-bignum x25519 backend to selection business logic
* PR4: retire existing x25519 arm implementation

https://github.com/aws/aws-lc/pull/932 refactored the logic to select the backend implementation. This PR wires up the s2n-bignum implementations.

To do that we need three things:
1. Detect when x25519 s2n-bignum is supported
2. Detect when each x25519 s2n-bignum implementation is supported
3. Decide which x25519 s2n-bignum implementation to use

#### For 1

This is implemented in `x25519_s2n_bignum_capable ()`. There should be at least one s2n-bignum implementation that works on pretty much all Armv8 and all x86_64 CPU. So, this function will return 1 if either we have `OPENSSL_AARCH64` or `OPENSSL_X86_64` defined.

This function is strictly needed, because of build options we support. For example, one can disable all assembly. If this is the case, we want to steer away from all the s2n-bignum implementations in the control flow.

#### For 2

There are two s2n-bignum implementations: s2n-bignum-no-alt and s2n-bignum-alt. They do different things depending on the architectures. I tried explaining this in a code comment above `x25519_s2n_bignum()`.

Run-time detection of whether a particular implementation is supported is handled by either `x25519_s2n_bignum_alt_capable()` or `x25519_s2n_bignum_no_alt_capable()`. These are a bit special, because they are also used to specifically match and catagorise certain CPUs. For example, on aarch64, the former is only returning 1 if the CPU has a wide multiplier. One could consider making this a separate check instead of overloading the semantics of the function with both a a) is capable of using implementation and b) is a certain type of capable CPU.

~~**Note:** To do the run-time detection on aarch64, I use the distinguisher implementation from https://github.com/aws/aws-lc/pull/721, that is not merged yet. That distinguisher specifically detects wide multipliers using some catagorisation. For example, it regards Apple M1 as a cpu with a wide multiplier. This PR will eventually be rebased on 721, or 721 will be rebased on this PR. But for now, I need that piece of code to make this PR work.~~

#### For 3

Lastly, we need to figure out which implementations to actually use. This is fairly simple using the functions from (2). The only thing we need to cater for is the order of which implementations to use. This is handled in `x25519_s2n_bignum()` and `x25519_s2n_bignum_public_from_private()`. These are also the functions that the public API will call if s2n-bignum capable at run-time.

### Call-outs:

We always prefer to use s2n-bignum if available. But in any case, always fall back to the native C-implementation if everything else fails.

I fixed a small "regression" from https://github.com/aws/aws-lc/pull/932. The neon implementation handles the internal demasking. So, we don't need to do this in `X25519_public_from_private()` before calling `x25519_NEON()`.

The delocator changes adds regex support to match offset formats: `#A +B`, `#(A*B)+C*D`, and `#A+B+C`.

Replace codebuild instance type with `BUILD_GENERAL1_LARGE` to avoid OOM.

### Testing:

We have specific tests in CI for x25519. But I'm also running tests on a small set of instances that have different cpus: c4, c5, c5a, c6g, c7g. As well as different Apple platforms.

the ISC license.


torben-hansen:
x25519 s2n-bignum source files have labels with the same names i.e. not all labels are unique to the source file where they appear. This doesn't work with the static FIPS build that join all assembly files. Need to fix this, upstream would be best.

```
[2/16] Generating bcm-delocated.S
FAILED: crypto/fipsmodule/bcm-delocated.S 
cd /home/ec2-user/humble/aws-lc/awslc-build/crypto/fipsmodule && ./delocate -a /home/ec2-user/humble/aws-lc/awslc-build/crypto/fipsmodule/libbcm_c_generated_asm.a -o bcm-delocated.S aesni-gcm-avx512.S aesni-gcm-x86_64.S aesni-x86_64.S ghash-ssse3-x86_64.S gh$
sh-x86_64.S md5-x86_64.S p256-x86_64-asm.S p256_beeu-x86_64-asm.S rdrand-x86_64.S rsaz-avx2.S sha1-x86_64.S sha256-x86_64.S sha512-x86_64.S vpaes-x86_64.S x86_64-mont5.S x86_64-mont.S p384/bignum_add_p384.S.S p384/bignum_sub_p384.S.S p384/bignum_neg_p384.S.S 
p384/bignum_tomont_p384.S.S p384/bignum_deamont_p384.S.S p384/bignum_montmul_p384.S.S p384/bignum_montmul_p384_alt.S.S p384/bignum_montsqr_p384.S.S p384/bignum_montsqr_p384_alt.S.S p384/bignum_nonzero_6.S.S p384/bignum_littleendian_6.S.S p521/bignum_add_p521$
S.S p521/bignum_sub_p521.S.S p521/bignum_neg_p521.S.S p521/bignum_mul_p521.S.S p521/bignum_mul_p521_alt.S.S p521/bignum_sqr_p521.S.S p521/bignum_sqr_p521_alt.S.S p521/bignum_tolebytes_p521.S.S p521/bignum_fromlebytes_p521.S.S curve25519/curve25519_x25519.S.S 
curve25519/curve25519_x25519_alt.S.S p384/bignum_tomont_p384_alt.S.S p384/bignum_deamont_p384_alt.S.S
panic: Duplicate symbol found: "copyloop" in "curve25519/curve25519_x25519_alt.S.S"

goroutine 1 [running]:
main.transform.func1(0xa5?)
        /home/ec2-user/humble/aws-lc/util/fipstools/delocate/delocate.go:1736 +0x12b
main.forEachPath(0xc0ce821260?, 0xc001193af0, {0xc00119372f?, 0x1, 0x90?})
        /home/ec2-user/humble/aws-lc/util/fipstools/delocate/delocate.go:2205 +0xa2
main.forEachPath(0xc0031b7fc0?, 0xc001193af0, {0xc00119372e?, 0x2, 0x1?})
        /home/ec2-user/humble/aws-lc/util/fipstools/delocate/delocate.go:2207 +0xb6
main.forEachPath(0x4e6e00?, 0xc001193af0, {0xc00119372d?, 0x3, 0x7efcd5a64800?})
        /home/ec2-user/humble/aws-lc/util/fipstools/delocate/delocate.go:2207 +0xb6
main.transform({0x5254b8?, 0xc009186278}, {0xc000117000, 0x29, 0x0?})
        /home/ec2-user/humble/aws-lc/util/fipstools/delocate/delocate.go:1733 +0x1fc8
main.main()
        /home/ec2-user/humble/aws-lc/util/fipstools/delocate/delocate.go:2180 +0x745
ninja: build stopped: subcommand failed.
```

torben-hansen:
Fix duplicate symbol names in s2n-bignum https://github.com/awslabs/s2n-bignum/pull/50

torben-hansen:
The prefix build still fails, but it's because there is a false positive here: https://github.com/aws/aws-lc/blob/main/tests/ci/common_posix_setup.sh#L90-L98

Need to filter out `x25519_x25519` symbols as well. Will make this change in next commit.

But this PR can still be reviewed now!

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,223 @@
...
 #if defined(OPENSSL_ARM) && !defined(OPENSSL_NO_ASM) && !defined(OPENSSL_APPLE)
-#define BORINGSSL_X25519_NEON
+#define BORINGSSL_X25519_NEON_CAPABLE
 #endif
 
+#if (defined(OPENSSL_X86_64) || defined(OPENSSL_AARCH64)) && \
+    (defined(OPENSSL_LINUX) || defined(OPENSSL_APPLE)) && \
+    !defined(OPENSSL_NO_ASM) && !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
+#include "../../third_party/s2n-bignum/include/s2n-bignum_aws-lc.h"
+#define CURVE25519_S2N_BIGNUM_CAPABLE
```
dkostic, line 42: nit: can we use `CURVE25519_USE_S2N_BIGNUM` to be more consistent with p384/p521

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,223 @@
...
+#endif
 
-void x25519_NEON(uint8_t out[32], const uint8_t scalar[32],
-                 const uint8_t point[32]);
 
-#if defined(BORINGSSL_X25519_NEON)
+OPENSSL_INLINE int x25519_s2n_bignum_capable(void) {
```
dkostic, line 46: why do we need a separate function for this check?

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,223 @@
...
+#if defined(CURVE25519_S2N_BIGNUM_CAPABLE)
+  return 1;
+#else
+  return 0;
+#endif
+}
 
-OPENSSL_INLINE int curve25519_asm_capable(void) {
+OPENSSL_INLINE int x25519_Armv7_neon_capable(void) {
```
dkostic, line 54: why `Armv7` in the function name?

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,223 @@
...
   abort();
 }
+#endif // !defined(BORINGSSL_X25519_NEON)
 
+// Run-time detection for each implementation
+
+OPENSSL_INLINE int x25519_s2n_bignum_alt_capable(void);
+OPENSSL_INLINE int x25519_s2n_bignum_no_alt_capable(void);
+
+// For aarch64, |x25519_s2n_bignum_alt_capable| returns 1 if we catagorise the
```
dkostic, line 105:
```suggestion
// For aarch64, |x25519_s2n_bignum_alt_capable| returns 1 if we categorize the
```

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,223 @@
...
+// CPU as a CPU having a wide multiplier (i.e. "higher" throughput). CPUs with
+// this feature is e.g.: AWS Graviton 3 and Apple M1. Return 0 otherwise, so we
```
dkostic, line 107:
```suggestion
// this feature are e.g.: AWS Graviton 3 and Apple M1. Return 0 otherwise, so we
```

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,223 @@
...
+#if !defined(BORINGSSL_X25519_NEON_CAPABLE)
 void x25519_NEON(uint8_t out[32], const uint8_t scalar[32],
                  const uint8_t point[32]) {
   abort();
 }
+#endif // !defined(BORINGSSL_X25519_NEON)
 
+// Run-time detection for each implementation
+
+OPENSSL_INLINE int x25519_s2n_bignum_alt_capable(void);
```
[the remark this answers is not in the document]
torben-hansen: But "capable" just means whether we can actually execute this implementation.

The logic that decides which is implemented in e.g. `x25519_s2n_bignum()`.

crypto/curve25519/curve25519.c
torben-hansen: I believe "capable" is correct for how I am using it here.

crypto/curve25519/curve25519.c
torben-hansen: yeah, I believe so. Since, this function is used to decide if we should use an s2n-bignum implementation or not. I would rather split this up: one function to say "yes we can use s2n-bignum" and other functions to say "this implementation can execute on this CPU we are on"

crypto/curve25519/curve25519.c
torben-hansen: armv7 is the 32-bit ISA, no?

crypto/curve25519/curve25519.c
torben-hansen: Really using correct spelling here! FWIW.

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,223 @@
...
+  const uint8_t point[32]);
+void curve25519_x25519_byte_alt(uint8_t res[32], const uint8_t scalar[32],
+  const uint8_t point[32]);
+void curve25519_x25519base_byte(uint8_t res[32], const uint8_t scalar[32]);
+void curve25519_x25519base_byte_alt(uint8_t res[32], const uint8_t scalar[32]);
+
+void curve25519_x25519_byte(uint8_t res[32], const uint8_t scalar[32],
+  const uint8_t point[32]) {
+  abort();
+}
```
andrewhop, line 76: Can we avoid these aborts? How did P384/521 manage this case? Can we gracefully fallback to the existing implementation and not call the s2n-bignum code if the runtime machine doesn't support it?

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,223 @@
...
+    return 1;
+  } else {
+    return 0;
+  }
+#else
+  return 0;
 #endif
+}
+
+// For aarch64, |x25519_s2n_bignum_alt_capable| always returns 1. If x25519
```
andrewhop, line 129:
```suggestion
// For aarch64, |x25519_s2n_bignum_no_alt_capable| always returns 1. If x25519
```

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,223 @@
...
+// of x25519 s2n-bignum we should use if x25519 s2n-bignum capable. Currently,
+// we support the following implementations.
+//
+// x86_64:
+//   - s2n-bignum-no-alt: hardware implementation using bmi2+adx instruction sets
+//   - s2n-bignum-alt: hardware implementation using standard instructions
+//
+// aarch64:
+//   - s2n-bignum-no-alt: hardware implementation for "small" multiplier throughput
+//   - s2n-bignum-alt: hardware implementation for "higher" multiplier throughput
```
andrewhop, line 162: NP small and higher aren't antonyms, is this small/big or low/high? Up above you used wide to describe it, so wide/narrow?

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,223 @@
...
+//
+// Through experiments we have found that:
```
andrewhop, line 164: Did you considered determining this at runtime? Could customers call a power on benchmark to do the local experiment and figure out for their CPU the best implementation? This is out of scope for this change, I'm just thinking about cases were our experiment wouldn't produce the best results for customers.

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,223 @@
...
+
+  if (x25519_s2n_bignum_no_alt_capable() == 1) {
+    curve25519_x25519_byte(out_shared_key, private_key_internal_demask,
+      peer_public_value);
+  } else if (x25519_s2n_bignum_alt_capable() == 1) {
+    curve25519_x25519_byte_alt(out_shared_key, private_key_internal_demask,
+      peer_public_value);
+  } else {
+    abort();
+  }
```
andrewhop, line 193: Is it possible to get here? Up above each of the functions have a guaranteed supported path between alt and non-alt. I prefer the p384 style of branching just on using alt or not.

crypto/curve25519/curve25519.c
torben-hansen: This code path is only taken if:
* architecture support is signaled at compile-time (i.e. x86_64 or aarch64).
* run-time checks pass

It's is equivalent to the p{384,521} implementation. Except, this implementation is more robust, since the p{384,521} cannot fall-back to the C-implementation at run-time: decision to use s2n-bignum is solely done at compile-time (AFAICT). But we do this in `X25519()`.

If both implementations somehow tries to use an s2n-bignum implementation when not supported (in this case of this PR this would be a bug in the logic), both this PR and p{384,521} will have the same failure mode: either abort of fail with an invalid instruction.

torben-hansen: The logic to fall-back to a different implementation is implemented in e.g. `X25519()`.

crypto/curve25519/curve25519.c
torben-hansen: That is true...

crypto/curve25519/curve25519.c
torben-hansen: I guess that is possible to do, at least in theory :)

crypto/curve25519/curve25519.c
torben-hansen: We could add a total fallback. That's possible.

Well, in theory this code here is more robust then p{384,521}. Not that I think that p{384,521} would ever reasonably fail.

torben-hansen: I did this. Not sure I like it(?)

torben-hansen: discussed offline: falling back to C-implementation would basically imply we are doing something very invalid e.g. running a 64-bit compiled program on a 32-bit architecture. This wouldn't work anyway.

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,226 @@
...
+#if defined(CURVE25519_S2N_BIGNUM_CAPABLE)
+  return 1;
+#else
+  return 0;
+#endif
+}
 
-OPENSSL_INLINE int curve25519_asm_capable(void) {
+OPENSSL_INLINE int x25519_Armv7_neon_capable(void) {
+#if defined(BORINGSSL_X25519_NEON)
```
dkostic, line 55: shouldn't this be `BORINGSSL_X25519_NEON_CAPABLE`?

torben-hansen: doh

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,226 @@
...
+  } else {
+    return 0;
+  }
+#else
+  return 0;
+#endif
+}
+
+// For aarch64, |x25519_s2n_bignum_no_alt_capable| always returns 1. If x25519
+// s2n-bignum capable, the Armv8 s2n-bignum-alt version should be supported on
```
nebeid, line 133: Did you mean
```suggestion
// s2n-bignum capable, the Armv8 s2n-bignum-no-alt version should be supported on
```
?

crypto/curve25519/curve25519.c
```diff
@@ -31,33 +31,226 @@
...
+// aarch64:
+//   - s2n-bignum-no-alt: hardware implementation for "low" multiplier throughput
+//   - s2n-bignum-alt: hardware implementation for "high" multiplier throughput
+//
+// Through experiments we have found that:
+//
+// For x86_64: bmi+adc will almost always give a performance boost. So, here we
+//   prefer s2n-bignum-no-alt over s2n-bignum-alt if the former is supported.
+// For aarch64: if a wide multiplier is supported, we prefer s2n-bignum-alt over
+//   s2n-bignum-no-alt if the former is supported.
```
nebeid, line 172: Do you need that extra part? The "if" should be what's at the beginning of the sentence, right?
```suggestion
//   s2n-bignum-no-alt.
```

crypto/curve25519/curve25519.c
```diff
@@ -189,17 +382,13 @@ int ED25519_verify(const uint8_t *message, size_t message_len,
 void X25519_public_from_private(uint8_t out_public_value[32],
                                 const uint8_t private_key[32]) {
 
-  uint8_t e[32];
-  OPENSSL_memcpy(e, private_key, 32);
-  e[0] &= 248;
-  e[31] &= 127;
-  e[31] |= 64;
```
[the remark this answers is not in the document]
torben-hansen: It's not getting a different input. Notice before it was using `private_key`, not `e`: https://github.com/aws/aws-lc/blob/main/crypto/curve25519/curve25519.c#L200

third_party/s2n-bignum/include/s2n-bignum_aws-lc.h
```diff
@@ -116,3 +116,17 @@ extern void bignum_fromlebytes_p521(uint64_t z[static 9], const uint8_t x[static
 
 // Convert 9-digit 528-bit bignum to little-endian bytes
 extern void bignum_tolebytes_p521(uint8_t z[static 66], const uint64_t x[static 9]);
+
+// curve25519_x25519_byte and curve25519_x25519_byte_alt computes the x25519
+// function specified in https://www.rfc-editor.org/rfc/rfc7748. |scalar| is the
+// scalar, |point| is the u-coordinate of the elliptic curve
+// point. The result, another u-coordinate, is saved in |res|.
+extern void curve25519_x25519_byte(uint8_t res[static 32], const uint8_t scalar[static 32], const uint8_t point[static 32]);
```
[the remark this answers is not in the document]
andrewhop: > the value of the corresponding actual argument shall provide access to the first element of an array with at least as many elements as specified by the size expression

Is that not true if you don't specify `static`? You just changed the capitlization, but should we add static to stuff like `void x25519_public_from_private_nohw(uint8_t out_public_value[32], const uint8_t private_key[32]);` so `out_public_value` actually has to be 32 bytes long?

andrewhop (approved):

dkostic (approved):

PR #941, opened 2023-04-11
Add support for users to define their own method to handle FIPS failures

andrewhop:
### Issues:
Addresses CryptoAlg-1737

### Description of changes: 
Add support for users to define their own method to handle FIPS failures by defining `AWS_LC_fips_failure_callback`. If that symbol is defined all failures will call that. This change does not affect the runtime tests from https://github.com/aws/aws-lc/pull/938, that will be handled in a separate PR. Those changes require making a lot of void functions return a int result.

### Call-outs:
I designed these changes with the idea of maintaining the default behavior for existing consumers of AWS-LC. If `AWS_LC_fips_failure_callback` is not defined AWS-LC will `abort` and `exit` on any FIPS failure.

### Testing:
I added a new test that defines a callback and asserts it gets called as expected. Right now the interaction between what exact FIPS test failed and the messaging to the consumer is not perfect. I plan to expand on the testing more next week.

the ISC license.


crypto/fipsmodule/self_check/self_check.c
```diff
@@ -1198,10 +1198,15 @@ int BORINGSSL_self_test(void) {
       // When requested to run self tests, also run the lazy tests.
       !boringssl_self_test_rsa() ||
       !boringssl_self_test_ecc() ||
-      !boringssl_self_test_ffdh()) {
+      !boringssl_self_test_ffdh() ||
+      !boringssl_self_test_sha256() ||
+      !boringssl_self_test_hmac_sha256()) {
```
torben-hansen, line 1203: why add these?

crypto/fipsmodule/self_check/self_check.c
```diff
@@ -1198,10 +1198,15 @@ int BORINGSSL_self_test(void) {
...
+#if defined(BORINGSSL_FIPS)
+    return AWS_LC_FIPS_error("BORINGSSL_self_test did not pass.", ERR_R_FIPS_TEST_FAILURE);
```
torben-hansen, line 1205: could this return 0 if not not build in fips mode? avoids the if-def things

util/fipstools/test-break-kat.sh
```diff
@@ -18,7 +18,7 @@
 set -x
```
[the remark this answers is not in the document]
andrewhop: Correct, I'm not sure why BoringSSL left it as not executable.

util/fipstools/break-kat.go
```diff
@@ -1,5 +1,3 @@
-//go:build
```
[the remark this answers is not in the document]
andrewhop: We're missing https://github.com/google/boringssl/commit/bab2f96e2637acb107fe9b099c58befbca918748 (haven't gotten to it yet), and the version of Go on my system doesn't like them.

crypto/fipsmodule/self_check/self_check.c
andrewhop: It made my testing simpler, so all the algorithms behave the same. Now if the constructor or if a customer calls BORINGSSL_self_test they get the same results.

crypto/fipsmodule/self_check/self_check.c
andrewhop: In non-FIPS mode this function return 0 or 1. In FIPS mode this function either:
* returns 1, or aborts if AWS_LC_fips_failure_callback is not defined
* returns 1, or calls AWS_LC_fips_failure_callback and returns 0

crypto/fipsmodule/self_check/self_check.c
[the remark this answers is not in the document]
andrewhop: They're not lazy, they're previously only called in the constructor. To simplify my testing, and in the spirit of BORINGSSL_self_test checking all the operational algorithms I added them. I like your spot better, I didn't see that originally.

crypto/fipsmodule/ec/ec_key.c
```diff
@@ -540,10 +540,10 @@ int EC_KEY_generate_key_fips(EC_KEY *eckey) {
   eckey->priv_key = NULL;
 
 #if defined(AWSLC_FIPS)
```
torben-hansen, line 542: Now that we have a function that can handle fips-mode or not, would prefer to push that stuff to the function.

But I think it would also need an additional input to signal whether it should abort or not when in non-fips mode. Which obviously complicates that a bit...

torben-hansen (approved):

samuel40791765 (approved):

crypto/internal.h
```diff
@@ -946,14 +946,12 @@ static inline uint64_t CRYPTO_rotr_u64(uint64_t value, int shift) {
...
-// process.
-#if defined(_MSC_VER)
-__declspec(noreturn) void BORINGSSL_FIPS_abort(void);
-#else
-void BORINGSSL_FIPS_abort(void) __attribute__((noreturn));
-#endif
+// AWS_LC_FIPS_error is called when a FIPS power-on or continuous test
+// fails. It calls |AWS_LC_fips_failure_callback| if the weak symbol is defined
+// in your application and then returns 0. If |AWS_LC_fips_failure_callback|
+// is not defined, it prevents any further cryptographic operations by the
```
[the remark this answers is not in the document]
nebeid: (forgot this comment was pending on my end, so you can leave it out if you want)

crypto/fipsmodule/cpucap/cpucap.c
```diff
@@ -79,7 +79,7 @@ HIDDEN uint32_t OPENSSL_armcap_P = 0;
 
 #if defined(BORINGSSL_DISPATCH_TEST)
 // This value must be explicitly initialized to zero. See similar comment above.
-HIDDEN uint8_t BORINGSSL_function_hit[7] = {0};
+HIDDEN uint8_t BORINGSSL_function_hit[8] = {0};
```
nebeid, line 82: Does this change need to go to main irrespective of the rest of the PR?

crypto/fipsmodule/self_check/self_check.c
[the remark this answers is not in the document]
nebeid: I agree, because I got confused and thought they sha256 and hmac_sha256 were lazy.
I'd say "run lazy tests and SHA-256 and HMAC-SHA256 which are part of the integrity test."

crypto/internal.h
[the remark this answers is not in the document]
andrewhop: Yes this comment is true.

crypto/fipsmodule/cpucap/cpucap.c
andrewhop: It already was made on main, this is cherry picking it onto this branch. https://github.com/aws/aws-lc/pull/934

PR #946, opened 2023-04-11
Add service indicator check for 0 <= slen <= hlen

samuel40791765:
### Issues:
Resolves `CryptoAlg-1753`

### Description of changes: 
 FIPS 186-4, Section 5.5, requires that if the RSA key size is 1024  bits (128 bytes) and the hash function output block is 512 bits (64 bytes), then the length of the salt shall be "0 <= sLen <= hLen-2". Otherwise it requires that the salt length used in PSS satisfies "0 <= sLen <= hLen".
* https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-5.pdf

### Call-outs:
Adding the check directly into the digest signing will create backwards incompatibility. We've gotten approval from our FIPS vendor to add this check in the service indicator instead. OpenSSL follows the same concept by marking a new default option in their "provider" APIs (their version of the service indicator).
* https://github.com/openssl/openssl/commit/6c73ca4a2f4ea71f4a880670624e7b2fdb6f32da

### Testing:
Adjust test that was failing to use a value larger than `hlen` instead.

the ISC license.

### DIFF
--- a/crypto/fipsmodule/service_indicator/service_indicator.c
+++ b/crypto/fipsmodule/service_indicator/service_indicator.c
@@ -225,22 +225,33 @@ static void evp_md_ctx_verify_service_indicator(const EVP_MD_CTX *ctx,
     if (!EVP_PKEY_CTX_get_rsa_padding(pctx, &padding)) {
       goto err;
     }
-    if (padding == RSA_PKCS1_PSS_PADDING) {
-      int salt_len;
-      const EVP_MD *mgf1_md;
-      if (!EVP_PKEY_CTX_get_rsa_pss_saltlen(pctx, &salt_len) ||
-          !EVP_PKEY_CTX_get_rsa_mgf1_md(pctx, &mgf1_md) ||
-          (salt_len != -1 && salt_len != (int)EVP_MD_size(pctx_md)) ||
-          EVP_MD_type(mgf1_md) != md_type) {
-        // Only PSS where saltLen == hashLen is tested with ACVP. Cases with
-        // non-standard padding functions are also excluded.
-        goto err;
-      }
-    }
-
     // The approved RSA key sizes for signing are 2048, 3072 and 4096 bits.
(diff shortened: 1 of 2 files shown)


crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -225,23 +225,34 @@ static void evp_md_ctx_verify_service_indicator(const EVP_MD_CTX *ctx,
...
           !EVP_PKEY_CTX_get_rsa_mgf1_md(pctx, &mgf1_md) ||
-          (salt_len != -1 && salt_len != (int)EVP_MD_size(pctx_md)) ||
           EVP_MD_type(mgf1_md) != md_type) {
-        // Only PSS where saltLen == hashLen is tested with ACVP. Cases with
-        // non-standard padding functions are also excluded.
+        goto err;
+      }
+      // RSA_PSS_SALTLEN_DIGEST (-1) has the special meaning of "sLen ==
+      // hLen", other values need additional checks.
+      // FIPS 186-4, Section 5.5, requires that if the RSA key size is 1024
```
torben-hansen, line 243: but 1024 is not an allowed length, so no need to consider it?

And the stronger check above `salt_len != (int)EVP_MD_size(pctx_md)` is already sufficient?

samuel40791765: > but 1024 is not an allowed length, so no need to consider it?

Sure, I think we can remove the check for that, since it's not doable.

> And the stronger check above salt_len != (int)EVP_MD_size(pctx_md) is already sufficient?

The original comment implied that the original check was implemented since ACVP test cases only had `saltLen == hashLen` for PSS. Outside of ACVP test cases, the FIPS guidance states that `0 <= salt_len <= hash_len` is usable for FIPS as well, so I thought we should be marking these cases as APPROVED.

samuel40791765:
Closing in favor of the stricter check.

PR #948, opened 2023-04-13
New FIPS break tests for ECDH and AES-GCM-256

samuel40791765:
### Issues:
Addresses `CryptoAlg-1739`

### Description of changes: 
Added new break tests for ECDH and AES-GCM-256. 

### Call-outs:
Since the ECDH functions call the EC self tests before running themselves, we need a special function called `ecdh_compute_shared_secret_no_self_test` to avoid an infinity loop. This follows the previous pattern for other EC self tests and shouldn't effect the original behavior of the existing ECDH APIs. The "no_self_test" function will be left internal like the other no_self_test functions.

### Testing:
Running the break test script with `./util/fipstools/test-break-kat.sh` passes the break testing.

the ISC license.


samuel40791765:
This seemed like a change that could go onto mainline, so I've opened it there first. We can cherrypick into onto `experimental-fips-refactoring` once it's approved.
I'll can reopen this if we feel like we should put this onto `experimental-fips-refactoring` first.

crypto/fipsmodule/self_check/self_check.c
```diff
@@ -509,6 +509,88 @@ static int boringssl_self_test_rsa(void) {
...
+      0xa3, 0x8d, 0x12, 0xb7, 0x8c, 0xb1, 0x21, 0x89, 0xec, 0x67, 0x9d, 0xdd,
+      0x2e, 0x9a, 0xa2, 0x1f, 0xa9, 0xa5, 0xa0, 0xb0,
+  };
+
+  EC_KEY *key = EC_KEY_new();
+  EC_GROUP *group = EC_GROUP_new_by_curve_name(NID_secp384r1);
+  EC_POINT *pub_key = EC_POINT_new(group);
+  EC_POINT *peer_pub_key = EC_POINT_new(group);
+
+  BIGNUM *qx = BN_bin2bn(kQx, sizeof(kQx), NULL);
```
torben-hansen, line 555: nit: you don't actually need all this intermediate stuff if you encode the coordinates as an EC point.

crypto/fipsmodule/ecdh/ecdh.c
```diff
@@ -131,6 +130,15 @@ int ECDH_compute_shared_secret(uint8_t *buf, size_t *buflen, const EC_POINT *pub
   return ret;
 }
 
+int ECDH_compute_shared_secret(uint8_t *buf, size_t *buflen,
+                               const EC_POINT *pub_key,
+                               const EC_KEY *priv_key) {
+  boringssl_ensure_ecc_self_test();
+
+  return ecdh_compute_shared_secret_no_self_test(buf, buflen, pub_key,
```
torben-hansen, line 138: OK, but we could prevent the same thread from entering `run_self_test_ecc` twice I guess. And we wouldn't need two functions?

torben-hansen (approved):

crypto/fipsmodule/ecdh/ecdh.c
samuel40791765: I'll keep this for now, so we're following our original pattern.
I'm curious though, is there a way to do that? Seems like `CRYPTO_once` only blocks concurrent threads from accessing the original executing thread, is there something that would help us block the original thread execution as well

torben-hansen: yeah, just a thread-local semaphore

torben-hansen (approved):

samuel40791765:
Closing, good news is that this isn't required anymore

PR #950, opened 2023-04-13
Update RAND_bytes to return return 0 on failure

andrewhop:
### Issues:
Addresses CryptoAlg-1738

### Description of changes: 
`RAND_bytes` has always returned an `int` but most callers assumed it would never fail. And there was no path that anything besides 1 could be returned, if any issue did happen it would `abort`. This PR adds a potential failure path that AWS-LC needs to handle, and updates all libcrypto callers. We could handle libssl in a separate PR. I believe this is okay for know because:
1. non-FIPS builds of AWS-LC will still abort, the only place that returns 0 is in the FIPS build
2. No one is currently using FIPS libcrypto with libssl

### Call-outs:
The most interesting logic/behavior changes are in:
* rand_nonzero
* ED25519_keypair/X25519_keypair
* HRSS_decap/HRSS_encap/HRSS_generate_key

All callers of RAND_bytes in libcrypto code now handle the return value. 

### Testing:
Existing unit tests and a new callback test pass.

the ISC license.

### DIFF
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -1872,7 +1872,12 @@ static void sc_muladd(uint8_t *s, const uint8_t *a, const uint8_t *b,
 
 void ED25519_keypair(uint8_t out_public_key[32], uint8_t out_private_key[64]) {
   uint8_t seed[ED25519_SEED_LEN];
-  RAND_bytes(seed, ED25519_SEED_LEN);
+  if (!RAND_bytes(seed, ED25519_SEED_LEN)) {
+    // This is a public void function and can't be updated
+    OPENSSL_cleanse(out_public_key, 32);
+    OPENSSL_cleanse(out_private_key, 64);
+    return;
+  }
   ED25519_keypair_from_seed(out_public_key, out_private_key, seed);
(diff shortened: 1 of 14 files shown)


crypto/curve25519/curve25519.c
```diff
@@ -2091,7 +2096,12 @@ static void x25519_scalar_mult(uint8_t out[32], const uint8_t scalar[32],
 }
 
 void X25519_keypair(uint8_t out_public_value[32], uint8_t out_private_key[32]) {
-  RAND_bytes(out_private_key, 32);
+  if (!RAND_bytes(out_private_key, 32)) {
+    // This is a public void function and can't be updated
+    OPENSSL_cleanse(out_public_value, 32);
+    OPENSSL_cleanse(out_private_key, 64);
```
samuel40791765, line 2102:
```suggestion
    OPENSSL_cleanse(out_private_key, 32);
```

andrewhop: This is a very good catch. This also shows we need a callback + asan/valgrind flavor.

samuel40791765 (approved):

torben-hansen (approved):

PR #954, opened 2023-04-13
Add RSA and EC FIPS pairwise failure tests with the new callback

andrewhop:
### Issues:
Addresses CryptoAlg-1737

### Description of changes: 
Add two new tests to make sure the RSA and EC pairwise tests can be broken and call the new FIPS callback as expected.

### Call-outs:
This just worked when `BORINGSSL_FIPS_abort` was updated to the new logic in `AWS_LC_FIPS_error` in all of libcrypto including these two pairwise tests.

### Testing:
This passes locally. 

the ISC license.

### DIFF
--- a/crypto/fips_callback_test.cc
+++ b/crypto/fips_callback_test.cc
@@ -7,8 +7,10 @@
 #if defined(__ELF__) && defined(__GNUC__)
 
 #include <gtest/gtest.h>
+#include <openssl/ec_key.h>
 #include <openssl/nid.h>
 #include <openssl/rand.h>
+#include <openssl/rsa.h>
 #include <algorithm>
 #include <list>
 #include <string>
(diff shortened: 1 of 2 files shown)


crypto/fips_callback_test.cc
```diff
@@ -106,15 +108,55 @@ TEST(FIPSCallback, DRBGRuntime) {
...
-    // TODO: make AWS_LC_FIPS_error update a new global state so FIPS_mode returns 0
-    ASSERT_EQ(1, FIPS_mode());
     ASSERT_EQ(1, failure_count);
   } else {
     // BORINGSSL_FIPS_BREAK_TEST has not been set and everything should work
     ASSERT_TRUE(RAND_bytes(buf, sizeof(buf)));
-    ASSERT_EQ(1, FIPS_mode());
     ASSERT_EQ(0, failure_count);
   }
+  ASSERT_EQ(1, FIPS_mode());
```
samuel40791765, line 117: Duplicate test to L107. I think keeping the earlier FIPS assertion test makes more sense though.

crypto/fips_callback_test.cc
```diff
@@ -106,15 +108,55 @@ TEST(FIPSCallback, DRBGRuntime) {
...
+                                         strcmp(broken_runtime_test, "CRNG" ) == 0)) {
+    ASSERT_FALSE(EC_KEY_generate_key_fips(key.get()));
+    // RSA key generation can call the DRBG multiple times before failing, we
+    // don't know how many times, but it should fail at least once.
+    ASSERT_NE(0, failure_count);
+  } else {
+    // BORINGSSL_FIPS_BREAK_TEST has not been set and everything should work
+    ASSERT_TRUE(EC_KEY_generate_key_fips(key.get()));
+  }
+  ASSERT_EQ(1, FIPS_mode());
```
samuel40791765, line 159: Same comment here for L127 and L148

samuel40791765 (dismissed):

crypto/fips_callback_test.cc
andrewhop: I wanted this here to verify the behavior: that `FIPS_mode` doesn't change.

samuel40791765: Should we also include an early FIPS assertion test for `TEST(FIPSCallback, PowerOnTests)` to verify this behavior for those tests as well?

samuel40791765 (approved):

justsmth (approved):

PR #955, opened 2023-04-14
add experimental break tests for RSA encryption

samuel40791765:
### Issues:
Addresses `CryptoAlg-1739`

### Description of changes: 
Added new experimental break tests for RSA OEAP SHA-256 encryption

### Call-outs:
This is ugly and hopes are we don't have to use this. RSA encryption uses a random value from `RAND_bytes`, which results in a non-deterministic output each time. This forces us to use a "special testing function" for the self tests KATs like [ECDSA is doing](https://github.com/aws/aws-lc/blob/fb42ac63b765d1be7bb8cb5537bbd31c6c906976/crypto/fipsmodule/ecdsa/internal.h#L25-L32). In the RSA encryption special KAT testing functions, we use a fixed value of `{0}` instead of pulling from `RAND_bytes`.

* https://github.com/aws/aws-lc/blob/fb42ac63b765d1be7bb8cb5537bbd31c6c906976/crypto/fipsmodule/ecdsa/internal.h#L25-L32

Did the work for this prematurely to avoid scrambling at the last minute for the hard deadline we have to meet. I'll clean up the PR and add more documentation once it's confirmed that this is needed.

### Testing:
Running the break test script with `./util/fipstools/test-break-kat.sh` passes the break testing.

the ISC license.


samuel40791765:
Closing, good news is that this isn't required anymore

PR #959, opened 2023-04-17
Add test for libssl off build option

samuel40791765:
### Description of changes: 
I noticed this when working on https://github.com/aws/aws-lc/pull/874. Our fv CI dimension would fail because they were consuming the build option, but this should have been caught in our regular CI dimension as well.

### Call-outs:
I removed the `AWSLC_FIPS` logic from the script since we use `./tests/ci/run_fips_tests.sh` for FIPS testing now.

### Testing:
CI

the ISC license.


andrewhop (approved):
I double checked none of the existing build dimensions specify AWSLC_FIPS.

justsmth (approved):

PR #964, opened 2023-04-18
Don't use a pre-calculated digest for signing and verification in the self tests

samuel40791765:
### Issues:
Addresses `CryptoAlg-1779`

### Description of changes: 
FIPS vendor feedback required us to do calculate digest along with the actual signing and verification for RSA and ECDSA.
`EVP_DigestSign/Verify` functions aren't usable for this, since they would result in an infinity loop.
We've gotten feedback from our FIPS vendor, that we can just calculate the digest and signature in the same function and continue using the `*_no_self_tests` function. They just don't want us using a pre-calculated digest in the self tests.

### Call-outs:
N/A

### Testing:
CI

the ISC license.


crypto/fipsmodule/rsa/internal.h
```diff
@@ -151,6 +151,14 @@ int rsa_sign_no_self_test(int hash_nid, const uint8_t *digest,
                           size_t digest_len, uint8_t *out, unsigned *out_len,
                           RSA *rsa);
 
+int rsa_digestsign_no_self_test(const EVP_MD *md, const uint8_t *input,
```
nebeid, line 154: I suggest adding comments here similar to those in ecdsa/internal.h.

samuel40791765: Sure, I initially didn't add them since the section on L133~150 encapsulated all of these, but I'll add more comments for additional clarification.

nebeid (dismissed):

andrewhop (approved):

andrewhop (approved):

nebeid (approved):

PR #979, opened 2023-04-25
Add missing symbols for postgres

samuel40791765:
### Issues:
Resolves `CryptoAlg-1746` and `CryptoAlg-1747`

### Description of changes: 
Postgres has some "missing symbols" that we need to add back to support. 
1. `cast` had originally been removed along with `decrepit`, so this is just putting them back. The test file had not been removed, so we only need to add back `cast.c`.
2. Postgres expects `SSL_CTX_set_min_proto_version` to be a macro, which was OpenSSL's way of wrapping around the `SSL_CTX_ctrl` functions. The simplest way around this is to redefine the `SSL_CTX_set_min_proto_version` as a macro at the end of `ssl.h`, like we've been doing for other instances of this.
* Postgres: https://github.com/postgres/postgres/blob/c8e1ba736b2b9e8c98d37a5b77c4ed31baf94147/src/include/common/openssl.h#L42-L45
* OpenSSL: https://github.com/openssl/openssl/blob/418c6c520764491262018c45481a20ef10cd3bca/include/openssl/ssl.h.in#L1490-L1491

### Call-outs:
N/A

### Testing:
Confirmed by building with Postgres.
```
./configure --with-openssl --enable-tap-tests --with-includes=$PWD/aws-lc/test_build_dir/include --with-libraries=$PWD/aws-lc/test_build_dir/lib --prefix=$PWD/build
make -j8 -C contrib all
```

the ISC license.


crypto/decrepit/cast/cast.c
```diff
@@ -0,0 +1,465 @@
...
+  for (i = 0; i < 16; i++) {
+    key->data[i * 2] = k[i];
+    key->data[i * 2 + 1] = ((k[i + 16]) + 16) & 0x1f;
+  }
+}
+
+// The input and output encrypted as though 64bit cfb mode is being used. The
+// extra state information to record how much of the 64bit block we have used
+// is contained in *num.
+void CAST_cfb64_encrypt(const uint8_t *in, uint8_t *out, size_t length,
```
torben-hansen, line 360: Is this actually neededI I don't see it mentioned in CryptoAlg-1747

include/openssl/cast.h
```diff
@@ -0,0 +1,97 @@
...
+#endif
+
+
+#define CAST_ENCRYPT 1
+#define CAST_DECRYPT 0
+
+#define CAST_BLOCK 8
+#define CAST_KEY_LENGTH 16
+
+typedef struct cast_key_st {
```
torben-hansen, line 73: Make opaque?

crypto/decrepit/cast/cast.c
```diff
@@ -0,0 +1,465 @@
...
+        iv = ivec;
+      }
+      cc = *(in++);
+      c = iv[n];
+      iv[n] = cc;
+      *(out++) = c ^ cc;
+      n = (n + 1) & 0x07;
+    }
+  }
+  v0 = v1 = ti[0] = ti[1] = t = c = cc = 0;
```
torben-hansen, line 412: brain freeze...

crypto/decrepit/cast/cast.c
```diff
@@ -0,0 +1,465 @@
...
+                         const uint8_t *iv, int enc) {
+  CAST_KEY *cast_key = ctx->cipher_data;
+  CAST_set_key(cast_key, ctx->key_len, key);
+  return 1;
+}
+
+static int cast_ecb_cipher(EVP_CIPHER_CTX *ctx, uint8_t *out, const uint8_t *in,
+                           size_t len) {
+  CAST_KEY *cast_key = ctx->cipher_data;
+
```
torben-hansen, line 426: Could check that `len` is divisible by `CAST_BLOCK`, otherwise this function will not really work correctly.

The assert below does that, but not really strongly.

include/openssl/cast.h
```diff
@@ -0,0 +1,97 @@
...
+  uint32_t data[32];
+  int short_key;  // Use reduced rounds for short key
+} CAST_KEY;
+
+OPENSSL_EXPORT void CAST_set_key(CAST_KEY *key, size_t len,
```
torben-hansen, line 78: Do we really need to export these or could they be static functions?

crypto/decrepit/cast/cast.c
samuel40791765: Added the assert :)

crypto/decrepit/cast/cast.c
samuel40791765: Good point, I removed this entire function and moved everything else internally. We're only exposing the EVP functions, which we've marked as deprecated.

include/openssl/cast.h
samuel40791765: I moved everything internally and removed the `cast.h` header file since it isn't needed anymore

samuel40791765 (commented):
I looked into some of our concerns for adding back Cast from decrepit and wanted to make the following points.
1. This had already existed in AWS-LC when we were working on removing decrepit. During the work for that, we had the consensus that we would remove everything that did not have usage internally and incrementally add back functions if customers needed the functionality.
2. We can consider influencing the Postgres upstream by submitting a PR to them to ifdef out the Cast ciphersuites. However, this change would only be added in new Postgres releases and won't be applicable to old releases. We can take a look at PostGres's versioning policy here: https://www.postgresql.org/support/versioning/
PostgreSQL Versions 11-15 are currently supported, with each major version of PostgreSQL having a support timeline of 5 years. Any upstream influence we make now, won't truly be felt until PostgreSQL15 is deprecated in 2027 (assuming an upstream contribution makes it to Version 16).

Considering these factors, we should look to add minimal support for the Cast ciphersuites through EVP now, so we can gain adoption in existing versions of PostGreSQL. We can still make an upstream contribution to [ifdef the CAST, blowfish, and des usage](https://github.com/postgres/postgres/blob/master/contrib/pgcrypto/openssl.c#L676-L742) out of PostGres when building with AWS-LC, so that new versions of PostGres don't continue rely on this behavior. 
We're making sure to only expose the minimal amount of symbols needed, which are the `EVP_CIPHER`s in this case. We're also adding the `AWS_LC_DEPRECATED` macro in front of the CAST ciphersuites to indicate that these function should not be relied on. This will intentionally cause current builds of PostGres to complain about deprecated symbols. We can make an upstream contribution to ifdef these deprecated symbols to silence these warnings.

torben-hansen (approved):

nebeid (approved):

PR #981, opened 2023-04-26
Define OPENSSL_NO_ASM macro if processor is unknown

francois-berder:
While trying to compile https://github.com/aws/aws-iot-device-sdk-cpp-v2 using yocto sdk for the Raspberry Pi 3, I got a number of undefined symbols at the linker stage.

This is due to `CMakeLists.txt` setting the `ARCH` variable (although it is already set to `arm` in my case) to `generic` because `CMAKE_SYSTEM_PROCESSOR` is set to `cortexa7t2hf-neon-vfpv4-hudl`.
If the processor is unknown then we cannot use ASM implementations. Thus, we must define the `OPENSSL_NO_ASM` macro to avoid errors at the linker stage.

### Issues:

No issues.

### Description of changes: 

Define `OPENSSL_NO_ASM` macro if processor is unknown.

### Call-outs:

### Testing:

No testing apart from building the project.

the ISC license.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -691,6 +691,7 @@ elseif (${CMAKE_SYSTEM_PROCESSOR} STREQUAL "riscv64")
   set(ARCH "riscv64")
 else()
   message(STATUS "Unknown processor found. Using generic implementations. Processor:" ${CMAKE_SYSTEM_PROCESSOR})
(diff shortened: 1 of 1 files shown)


torben-hansen:
I don't think this is the expected behaviour - please post build commands and all build output (including errors). Thanks.

PR #982, opened 2023-04-26
Add DH_check support to perf tool

torben-hansen:
### Description of changes: 

Makes it possible to measure performance of the function `DH_check()`.

### Call-outs:

Generating DH parameters for higher bit sizes gets pretty slow.

### Testing:

```
$ ./tool/bssl speed -filter dhcheck -timeout 10 -primes 2048,4096
DH check speed test is very slow. Speed test works best with higher timeouts.
Did 16 DH check(s) (2048 bits) operations in 10164348us (1.574 ops/sec)
Did 3 DH check(s) (4096 bits) operations in 15465898us (0.194 ops/sec)

$ ./tool/bssl speed -filter dhcheck -timeout 2 -primes 1024 -json
[
{"description": "DH check(s)", "numCalls": 20, "microseconds": 2045047, "primeSizePerCall": 1024}
]
```

**Checking new `-chuncks` and `-primes` parsing**

```
$ ./tool/bssl speed -filter EVP-AES-128-GCM -chunks 20,50,56
Did 5489000 EVP-AES-128-GCM encrypt init operations in 1000135us (5488259.1 ops/sec)
Did 10318000 EVP-AES-128-GCM encrypt (20 bytes) operations in 1000012us (10317876.2 ops/sec): 206.4 MB/s
Did 9833000 EVP-AES-128-GCM encrypt (50 bytes) operations in 1000011us (9832891.8 ops/sec): 491.6 MB/s
Did 9841000 EVP-AES-128-GCM encrypt (56 bytes) operations in 1000018us (9840822.9 ops/sec): 551.1 MB/s
Did 5481000 EVP-AES-128-GCM decrypt init operations in 1000071us (5480610.9 ops/sec)
Did 10456000 EVP-AES-128-GCM decrypt (20 bytes) operations in 1000013us (10455864.1 ops/sec): 209.1 MB/s
Did 10318500 EVP-AES-128-GCM decrypt (50 bytes) operations in 1000013us (10318365.9 ops/sec): 515.9 MB/s
Did 10318500 EVP-AES-128-GCM decrypt (56 bytes) operations in 1000007us (10318427.8 ops/sec): 577.8 MB/s


$ ./tool/bssl speed -filter EVP-AES-128-GCM -chunks 20,50,56
Did 5487000 EVP-AES-128-GCM encrypt init operations in 1000077us (5486577.5 ops/sec)
Did 10589000 EVP-AES-128-GCM encrypt (20 bytes) operations in 1000083us (10588121.2 ops/sec): 211.8 MB/s
Did 10230000 EVP-AES-128-GCM encrypt (50 bytes) operations in 1000024us (10229754.5 ops/sec): 511.5 MB/s
Did 10228250 EVP-AES-128-GCM encrypt (56 bytes) operations in 1000001us (10228239.8 ops/sec): 572.8 MB/s
Did 5497250 EVP-AES-128-GCM decrypt init operations in 1000019us (5497145.6 ops/sec)
Did 11285000 EVP-AES-128-GCM decrypt (20 bytes) operations in 1000073us (11284176.3 ops/sec): 225.7 MB/s
Did 11094000 EVP-AES-128-GCM decrypt (50 bytes) operations in 1000021us (11093767.0 ops/sec): 554.7 MB/s
Did 11095000 EVP-AES-128-GCM decrypt (56 bytes) operations in 1000021us (11094767.0 ops/sec): 621.3 MB/s
```

the ISC license.

### DIFF
--- a/tool/speed.cc
+++ b/tool/speed.cc
@@ -80,6 +80,13 @@ static std::string ChunkLenSuffix(size_t chunk_len) {
   return buf;
 }
 
+static std::string PrimeLenSuffix(size_t prime_length) {
+  char buf[32];
+  snprintf(buf, sizeof(buf), " (%zu bit%s)", prime_length,
+           prime_length != 1 ? "s" : "");
+  return buf;
+}
+
 // TimeResults represents the results of benchmarking a function.
 struct TimeResults {
   // num_calls is the number of function calls done in the time period.
@@ -113,6 +120,19 @@ struct TimeResults {
     }
   }
 
+  void PrintWithPrimes(const std::string &description,
+                      size_t prime_size) const {
+    if (g_print_json) {
+      PrintJSON(description, "primeSizePerCall", prime_size);
+    } else {
+      printf(
+          "Did %" PRIu64 " %s operations in %" PRIu64
+          "us (%.3f ops/sec)\n",
+          num_calls, (description + PrimeLenSuffix(prime_size)).c_str(), us,
+          (static_cast<double>(num_calls) / static_cast<double>(us)) * 1000000);
+    }
+  }
+
  private:
   void PrintJSON(const std::string &description,
                  size_t bytes_per_call = 0) const {
@@ -126,6 +146,25 @@ struct TimeResults {
 
     if (bytes_per_call > 0) {
       printf(", \"bytesPerCall\": %zu", bytes_per_call);
+    }
+
+    printf("}");
+    first_json_printed = true;
+  }
+
+  void PrintJSON(const std::string &description,
+                 const std::string &size_label,
+                 size_t size = 0) const {
+    if (first_json_printed) {
+      puts(",");
+    }
+
+    printf("{\"description\": \"%s\", \"numCalls\": %" PRIu64
+           ", \"microseconds\": %" PRIu64,
+           description.c_str(), num_calls, us);
+
+    if (size > 0) {
+      printf(", \"%s\": %zu", size_label.c_str(), size);
     }
 
     printf("}");
@@ -167,6 +206,7 @@ static uint64_t time_now() {
 
 static uint64_t g_timeout_seconds = 1;
 static std::vector<size_t> g_chunk_lengths = {16, 256, 1350, 8192, 16384};
+static std::vector<size_t> g_prime_bit_lengths = {2048, 3072};
 
 static bool TimeFunction(TimeResults *results, std::function<bool()> func) {
   // The first time |func| is called an expensive self check might run that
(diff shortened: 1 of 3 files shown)


tool/speed.cc
```diff
@@ -2072,6 +2164,32 @@ bool Speed(const std::vector<std::string> &args) {
     }
   }
 
+  if (args_map.count("-primes") != 0) {
```
andrewhop, line 2167: Did you consider making this a function to handle chunks and primes?

tool/speed.cc
```diff
@@ -2004,6 +2090,12 @@ static const argument_t kArguments[] = {
         "A comma-separated list of input sizes to run tests at (default is "
         "16,256,1350,8192,16384)",
     },
+    {
+        "-primes",
```
andrewhop, line 2094: Please update the other RSA tests like SpeedRSAKeyGen to use this as well.

tool/speed.cc
```diff
@@ -1892,6 +1932,52 @@ static bool SpeedJitter(std::string selected) {
...
+}
+
+static bool SpeedDHcheck(std::string selected) {
+  if (!selected.empty() && selected.find("dhcheck") == std::string::npos) {
+    return true;
+  }
+
+  if (!g_print_json) {
+    printf("DH check speed test is very slow. Speed test works best with higher timeouts.\n");
+  }
```
andrewhop, line 1970: Do you just want to defualt to 30 seconds? That's what the RSA keygen does.

tool/speed.cc
```diff
@@ -113,6 +120,19 @@ struct TimeResults {
     }
   }
 
+  void PrintWithPrimes(const std::string &description,
```
andrewhop, line 123: Would it be interesting to print out some of the other statistics like  min/mediian/max/std deviation?

tool/speed.cc
```diff
@@ -1892,6 +1932,52 @@ static bool SpeedJitter(std::string selected) {
...
+static bool SpeedDHcheck(size_t prime_bit_length) {
+
+  TimeResults results;
+  bssl::UniquePtr<DH> dh_params(DH_new());
+  if (dh_params == nullptr) {
+    return false;
+  }
+
+  // DH_generate_parameters_ex grows exponentially slower as prime length grows.
+  if (DH_generate_parameters_ex(dh_params.get(), prime_bit_length,
```
andrewhop, line 1944: Does performance vary with the parameters? Should this test use a fixed dh_params per size?

tool/speed.cc
torben-hansen: This seems reasonable. Have implemented this and now doesn't execute this speed dimension if not explicitly selected. Too slow.

`SpeedRSAKeyGen()` doesn't use the timing function tool. Actually that function is very different from everything else...

I have just set `g_timeout_seconds = 10` if no timeout was selected.

tool/speed.cc
andrewhop: I agree it can't, but would it make the data more useful/comparable?

tool/speed.cc
andrewhop: I was thinking of refactoring the shared logic to parse a comma delimited string into a list of integers which I think both primes and chunks are doing the same thing.

tool/speed.cc
torben-hansen: It would definitely make it faster. As prime size grows, this function does exponentially more work and is not even the thing we want to measure.

Can we do it at a later point? We need some more work to actually generate static parameters.

PR #1002, opened 2023-05-10
Add additional no-op symbols for support mySQL

samuel40791765:
### Issues:
Resolves `CryptoAlg-1726`

##### 1. CONF_modules_unload

mySQL consumes the `CONF_modules_unload` API when building with OpenSSL1.1.1/AWS-LC. `CONF_modules_unload` is used to finish and unload configuration modules in OpenSSL, but AWS-LC doesn't use the same configuration module logic. The `Conf_modules_*` family of functions in AWS-LC are all no-op functions, so we follow the same pattern here.

##### 2. SSL_CTX_get_security_level

mySQL consumes the `SSL_CTX_get_security_level` API. The definition of security levels is defined here in OpenSSL's documentation: https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_security_level.html
We should only be using `SSL_CTX_get_security_level` to retrieve a security level number. Actually using `SSL_CTX_set_security_level` to configure security levels in AWS-LC is not supported. I've set the returned number to 3, since we only support cipher suites that are at least 128 bits in security.

### Call-outs:
The additional header file is to get around an expectation that the `FIPS_mode` function can be called from `err.h`.
* https://github.com/mysql/mysql-server/blob/b9bd6fb7abf76e6a804095d2832c22e0d9f91fd6/mysys/my_openssl_fips.cc#L24

### Testing:

### DIFF
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -4886,6 +4886,11 @@ OPENSSL_EXPORT int SSL_CTX_set1_sigalgs_list(SSL_CTX *ctx, const char *str);
 // more convenient to codesearch for specific algorithm values.
 OPENSSL_EXPORT int SSL_set1_sigalgs_list(SSL *ssl, const char *str);
 
+// SSL_CTX_get_security_level returns 3. This is only to maintain compatibility
+// with OpenSSL and no security assumptions should be based on the number this
+// function returns.
+OPENSSL_EXPORT int SSL_CTX_get_security_level(const SSL_CTX *ctx);
+
 #define SSL_set_app_data(s, arg) (SSL_set_ex_data(s, 0, (char *)(arg)))
 #define SSL_get_app_data(s) (SSL_get_ex_data(s, 0))
 #define SSL_SESSION_set_app_data(s, a) \
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -4889,6 +4889,19 @@ OPENSSL_EXPORT int SSL_set1_sigalgs_list(SSL *ssl, const char *str);
 // SSL_CTX_get_security_level returns 3. This is only to maintain compatibility
 // with OpenSSL and no security assumptions should be based on the number this
 // function returns.
+//
+// Per OpenSSL's definition of Level 3:
+// Level 3:
+// Security level set to 128 bits of security. As a result RSA, DSA and DH keys
+// shorter than 3072 bits and ECC keys shorter than 256 bits are prohibited. In
+// addition to the level 2 exclusions cipher suites not offering forward secrecy
+// are prohibited. TLS versions below 1.1 are not permitted. Session tickets are
+// disabled.
+//
+// AWS-LC only supports atl least 128 bits of security, but we don't directly
+// prohibit session tickets and TLS 1.0 like Level 3 in OpenSSL states. This
+// behavior may change if we're asked to support actual Security level setting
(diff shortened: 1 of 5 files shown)


include/openssl/ssl.h
```diff
@@ -4886,6 +4886,11 @@ OPENSSL_EXPORT int SSL_CTX_set1_sigalgs_list(SSL_CTX *ctx, const char *str);
 // more convenient to codesearch for specific algorithm values.
 OPENSSL_EXPORT int SSL_set1_sigalgs_list(SSL *ssl, const char *str);
 
+// SSL_CTX_get_security_level returns 3. This is only to maintain compatibility
```
andrewhop, line 4889: Do all of our SSL configurations meet OpenSSL's definition of level 3 security?

samuel40791765: Yeah I was on the edge on documenting this, but I was worried that entire description might make things a bit bloated. Here's what OpenSSL's definition of level 2 and 3 is:
>Level 2
Security level set to 112 bits of security. As a result RSA, DSA and DH keys shorter than 2048 bits and ECC keys shorter than 224 bits are prohibited. In addition to the level 1 exclusions any cipher suite using RC4 is also prohibited. SSL version 3 is also not allowed. Compression is disabled.

 > Level 3
Security level set to 128 bits of security. As a result RSA, DSA and DH keys shorter than 3072 bits and ECC keys shorter than 256 bits are prohibited. In addition to the level 2 exclusions cipher suites not offering forward secrecy are prohibited. TLS versions below 1.1 are not permitted. Session tickets are disabled.

We only support 128 bits of security, but we don't seem to prohibit session tickets and TLS 1.0 like Level 3 states. But reporting Level 2 seems to be even more out of line since we don't support 112 bits.

andrewhop: I like the extra clarity. The question remains what if someone configs a connection with 256 bit cipher suites and expects this function to return 5 for them?

samuel40791765: I think this function only returns the "current security level" of the library. It seems to be based on how the library is configured, instead of per connection. Per the documentation:
> The default security level changes when OpenSSL is compiled by `DOPENSSL_TLS_SECURITY_LEVEL=level` ...

Or when setting the security level with "The functions SSL_CTX_set_security_level() and SSL_set_security_level() set the security level to level. If not set the library default security level is used."

We don't support setting security levels this way yet, so I think `3` is the most suitable number to return for our existing cipher suites now. But your question does address that we'll have to change this behavior once we get an ask to support actual "security_level" setting in AWS-LC.

andrewhop: Thanks, that clears this up, I agree 3 makes sense.

andrewhop (approved):

justsmth (approved):

PiotrSikora:
Doesn't OpenSSL's security level 3 prohibit using RSA 2048 (112 bits of security), which AWS-LC allows? Shouldn't this return security level 2 instead?

samuel40791765:
> Doesn't OpenSSL's security level 3 prohibit using RSA 2048 (112 bits of security), which AWS-LC allows? Shouldn't this return security level 2 instead?

Thanks for pointing this out! After some internal discussion, it does seem like this was a misunderstanding on my end. I'll put up a PR to lower the returned security level to 2 instead, since our default features do align with 2 better.

PR #1007, opened 2023-05-10
Add support for EVP_aes_256_wrap

samuel40791765:
### Issues:
Resolves `CryptoAlg-1724`

### Description of changes: 
mySQL consumes support for AES Key Wrap via the `EVP_aes_256_wrap` functions. AWS-LC already supports Key Wrapping, so we have existing functions we can try "wrapping" around :). No new crypto was implemented, but we should be staying consistent with the behavior of OpenSSL's EVP functions here.
1. The new `aes_wrap_init_key` static function was taken from OpenSSL's implementation.
2. The new `aes_wrap_cipher` wraps around our own implementation. The key wrapping should already be done when `EVP_(De/E)ncryptUpdate` is called, so there is no actual "final" operation for `EVP_aes_256_wrap`. OpenSSL addresses this by returning `0` during the final operation, so we do the same thing here.
* https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/evp/e_aes.c#L3814-L3815
I've also added service indicator locks here to keep the service indictor from incorrectly updating. Both of these behaviors are tested in the new service indicator tests.
3. I've chosen to mark `EVP_aes_256_wrap` as `NOT_APPROVED` for the service indicator, mainly because the complications of `EVP_CIPH_FLAG_CUSTOM_CIPHER` documented in `crypto/fipsmodule/cipher/cipher.c` still apply here. We can look into marking the consumption of this FIPS approved in another commit if needed. 
* https://github.com/aws/aws-lc/blob/d66b1bddf5942f2bf1232d1053f4073466f48954/crypto/fipsmodule/cipher/cipher.c#L537-L546

### Call-outs:
Above

### Testing:
The new AES Key wrap tests were extended upon our existing tests, but are ran with the EVP Ciphers instead.


crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -715,6 +715,63 @@ static int aes_xts_ctrl(EVP_CIPHER_CTX *c, int type, int arg, void *ptr) {
   return 1;
 }
 
+typedef struct {
+  union {
+      double align;
+      AES_KEY ks;
+  } ks;
+  // Indicates if IV has been set.
+  const uint8_t *iv;
```
andrewhop, line 724: Why do we need this? This just points to the iv in EVP_CIPHER_CTX.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -715,6 +715,63 @@ static int aes_xts_ctrl(EVP_CIPHER_CTX *c, int type, int arg, void *ptr) {
...
+  // function that indicates approval.
+  int ret;
+  FIPS_service_indicator_lock_state();
+  if (ctx->encrypt) {
+    ret = AES_wrap_key(&wctx->ks.ks, wctx->iv, out, in, inlen);
+  } else {
+    ret = AES_unwrap_key(&wctx->ks.ks, wctx->iv, out, in, inlen);
+  }
+  FIPS_service_indicator_unlock_state();
+  return ret;
```
andrewhop, line 772: Why do some EVP_CIPHER.cipher return success 0/1 (XTS and CBC) and others return a length (GCM and now key wrap)?

crypto/fipsmodule/aes/aes_test.cc
```diff
@@ -125,6 +125,73 @@ static void TestKeyWrap(FileTest *t) {
                                ciphertext.data(), ciphertext.size()));
 }
 
+static void TestEVPKeyWrap(FileTest *t) {
```
andrewhop, line 128: is it valuable to add a version of `WycheproofKeyWrap` that uses this new API? Also are there any other interesting error cases that should be tested by themselves?

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -902,6 +959,20 @@ DEFINE_METHOD_FUNCTION(EVP_CIPHER, EVP_aes_256_ofb) {
...
+DEFINE_METHOD_FUNCTION(EVP_CIPHER, EVP_aes_256_wrap) {
+  memset(out, 0, sizeof(EVP_CIPHER));
+
+  out->nid = NID_id_aes256_wrap;
+  out->block_size = 8;
+  out->key_len = 32;
+  out->iv_len = 8;
+  out->ctx_size = sizeof(EVP_AES_WRAP_CTX);
+  out->flags = EVP_CIPH_WRAP_MODE | EVP_CIPH_CUSTOM_IV |
+               EVP_CIPH_FLAG_CUSTOM_CIPHER | EVP_CIPH_ALWAYS_CALL_INIT;
```
andrewhop, line 971: Do you need to handle `EVP_CIPH_CUSTOM_IV` in `EVP_CipherInit_ex()` for `EVP_CIPH_WRAP_MODE`?

andrewhop (approved):

crypto/fipsmodule/aes/aes_test.cc
```diff
@@ -125,6 +125,77 @@ static void TestKeyWrap(FileTest *t) {
...
+  ctx.Reset();
+  ASSERT_TRUE(
+      EVP_DecryptInit_ex(ctx.get(), cipher, nullptr, key.data(), kDefaultIV));
+  ASSERT_TRUE(EVP_DecryptUpdate(ctx.get(), out.data(), &len, ciphertext.data(),
+                                ciphertext.size()));
+  out.resize(len);
+  ASSERT_TRUE(EVP_EncryptFinal(ctx.get(), out.data(), &len));
+  EXPECT_EQ(Bytes(plaintext), Bytes(out));
+
+  // Test corrupted ciphertext.
```
dkostic, line 190: why does `EVP_DecryptUpdate` fail with corrupted ciphertext?

samuel40791765: I think it's because the ciphertext can't be properly decrypted with same AES key here. In other words, this symmetric encryption scheme no longer recognizes the ciphertext with the same key. 
I took this from the test above in `TestKeyWrap`, which uses the AES Key `AES_wrap/unwrap_key` functions. Using `AES_unwrap_key` will also fail with a corrupted ciphertext.

dkostic: oh it's also authenticating... ok

dkostic (approved):

PR #1009, opened 2023-05-12
Unit test mode

torben-hansen:
### Description of changes: 

Adds a test mode that is entered when running unit tests. This will allow exported test-only functions to verify that it's only executed during testing.

In the future the test object `test_st` can be used to enable more granular tests e.g. test counting.

### Testing:

Verified behaviour of `CRYPTO_fork_detect_ignore_madv_wipeonfork_for_testing` through test fixture `ForkDetect.Test`.

### DIFF
--- a/crypto/crypto.c
+++ b/crypto/crypto.c
@@ -21,6 +21,26 @@
 #include "internal.h"
 
 
+struct test_st {
+  char test_mode;
+};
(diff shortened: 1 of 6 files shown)


samuel40791765 (approved):
If we add this, do we have to be careful that we're not running in "test mode" when we're testing regular behavior? Just worried that this could lead to differences in "what we test" vs "how we're consumed".

PR #1011, opened 2023-05-12
add support for SSL_get_key_update_type and SSL_KEY_UPDATE_NONE

samuel40791765:
### Issues:
Resolves `CryptoAlg-1851`

### Description of changes: 
mySQL consumes `SSL_get_key_update_type` along with `SSL_KEY_UPDATE_NONE` when building with OpenSSL1.1.1.  OpenSSL indicates that: 
> SSL_get_key_update_type() can be used to determine whether a key update operation has been scheduled but not yet performed. The type of the pending key update operation will be returned if there is one, or SSL_KEY_UPDATE_NONE otherwise.

AWS-LC does have existing related symbols (`SSL_key_update` and `SSL_KEY_UPDATE_(NOT_)REQUESTED`) and a `key_update_pending` boolean in the underlying SSL3 structure which may seem to be reusable in this case.
However, in OpenSSL, `SSL_get_key_update_type` has 3 possible values (`SSL_KEY_UPDATE_REQUESTED`, `SSL_KEY_UPDATE_NOT_REQUESTED`, `SSL_KEY_UPDATE_NONE`). We could return `SSL_KEY_UPDATE_NONE` when `key_update_pending` is false in AWS-LC, but we wouldn't be able to easily differentiate between `SSL_KEY_UPDATE_REQUESTED/NOT_REQUESTED` because both values are setting `key_pending` to true.
The way to work around this was to change `key_update_pending` into an `int`, so that it can maintain the 3 possible states that OpenSSL has.

* https://www.openssl.org/docs/man1.1.1/man3/SSL_get_key_update_type.html
* https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/include/openssl/ssl.h#L972-L977
* https://github.com/mysql/mysql-server/commit/98fa0818c8de42f8a26ba87b0f40f72ef75f6e57#diff-dec17245c4a9e498e6656bd12b5f8cb435bfba5a4dbc482913cbe03cc54ef324R195-R215

### Call-outs:
Above

### Testing:
New Unit Test

### DIFF
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -461,19 +461,29 @@ OPENSSL_EXPORT int SSL_write_ex(SSL *s, const void *buf, size_t num,
 #define SSL_KEY_UPDATE_REQUESTED 1
 
 // SSL_KEY_UPDATE_NOT_REQUESTED indicates that the peer should not reply with
-// it's own KeyUpdate message.
+// its own KeyUpdate message.
 #define SSL_KEY_UPDATE_NOT_REQUESTED 0
 
 // SSL_KEY_UPDATE_NONE should not be set by the user and is only used to
-// indicate that there is a pending key operation. OpenSSL indicates that -1 is
-// used, so that it will be an invalid value for the on-the-wire protocol.
+// indicate that there isn't a pending key operation. OpenSSL indicates that -1
+// is used, so that it will be an invalid value for the on-the-wire protocol
+// when calling |SSL_key_update|.
 #define SSL_KEY_UPDATE_NONE -1
 
 // SSL_key_update queues a TLS 1.3 KeyUpdate message to be sent on |ssl|
-// if one is not already queued. The |request_type| argument must be either be
+// if one is not already queued. The |request_type| argument must be either
 // |SSL_KEY_UPDATE_REQUESTED| or |SSL_KEY_UPDATE_NOT_REQUESTED|. This function
 // requires that |ssl| have completed a TLS >= 1.3 handshake. It returns one on
 // success or zero on error.
+//
+// If the updatetype parameter is set to |SSL_KEY_UPDATE_NOT_REQUESTED|, then
+// the sending keys for this connection will be updated and the peer will be
+// informed of the change.
+// If the updatetype parameter is set to |SSL_KEY_UPDATE_REQUESTED|, then the
+// sending keys for this connection will be updated and the peer will be
+// informed of the change along with a request for the peer to additionally
+// update its sending keys.
+// RFC: https://datatracker.ietf.org/doc/html/rfc8446#section-4.6.3
 //
 // Note that this function does not _send_ the message itself. The next call to
 // |SSL_write| will cause the message to be sent. |SSL_write| may be called with
@@ -481,11 +491,11 @@ OPENSSL_EXPORT int SSL_write_ex(SSL *s, const void *buf, size_t num,
 // pending.
 OPENSSL_EXPORT int SSL_key_update(SSL *ssl, int request_type);
 
-// SSL_get_key_update_type returns the internal |key_update_pending| status of
-// |ssl| and can be used to indicate whether a key update operation has been
-// scheduled but not yet performed. It will be of the values
+// SSL_get_key_update_type returns the state of the pending key operation in
+// |ssl|. The type of pending key operation will be either
 // |SSL_KEY_UPDATE_REQUESTED| or |SSL_KEY_UPDATE_NOT_REQUESTED| if there is one,
-// and |SSL_KEY_UPDATE_NONE| if otherwise.
+// and |SSL_KEY_UPDATE_NONE| if otherwise. This can be used to indicate whether
+// a key update operation has been scheduled but not yet performed.
 OPENSSL_EXPORT int SSL_get_key_update_type(const SSL *ssl);
 
 // SSL_shutdown shuts down |ssl|. It runs in two stages. First, it sends
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -476,13 +476,12 @@ OPENSSL_EXPORT int SSL_write_ex(SSL *s, const void *buf, size_t num,
 // requires that |ssl| have completed a TLS >= 1.3 handshake. It returns one on
 // success or zero on error.
 //
-// If the updatetype parameter is set to |SSL_KEY_UPDATE_NOT_REQUESTED|, then
-// the sending keys for this connection will be updated and the peer will be
-// informed of the change.
-// If the updatetype parameter is set to |SSL_KEY_UPDATE_REQUESTED|, then the
-// sending keys for this connection will be updated and the peer will be
-// informed of the change along with a request for the peer to additionally
-// update its sending keys.
+// If |request_type| is set to |SSL_KEY_UPDATE_NOT_REQUESTED|, then the sending
+// keys for this connection will be updated and the peer will be informed of the
+// change.
+// If |request_type| is set to |SSL_KEY_UPDATE_REQUESTED|, then the sending keys
+// for this connection will be updated and the peer will be informed of the change
+// along with a request for the peer to additionally update its sending keys.
 // RFC: https://datatracker.ietf.org/doc/html/rfc8446#section-4.6.3
 //
 // Note that this function does not _send_ the message itself. The next call to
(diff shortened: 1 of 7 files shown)


ssl/tls13_both.cc
```diff
@@ -632,7 +634,7 @@ bool tls13_add_key_update(SSL *ssl, int update_requested) {
   // Suppress KeyUpdate acknowledgments until this change is written to the
   // wire. This prevents us from accumulating write obligations when read and
   // write progress at different rates. See RFC 8446, section 4.6.3.
-  ssl->s3->key_update_pending = true;
```
dkostic, line 635: what's the equivalent function (`tls13_add_key_update`) in OpenSSL?

ssl/tls13_both.cc
```diff
@@ -655,7 +657,7 @@ static bool tls13_receive_key_update(SSL *ssl, const SSLMessage &msg) {
 
   // Acknowledge the KeyUpdate
   if (key_update_request == SSL_KEY_UPDATE_REQUESTED &&
-      !ssl->s3->key_update_pending &&
+      ssl->s3->key_update_pending == SSL_KEY_UPDATE_NONE &&
```
dkostic, line 660: what's the equivalent function (`tls13_receive_key_update `) in OpenSSL?

dkostic:
I'm not sure I completely understand the whole `key_update` story. Is the following correct:

In AWS-LC current implementation, `ssl->s3->key_update_pending` indicates if a key update operation was issued but not finished. The type of the operation can be either `SSL_KEY_UPDATE_REQUESTED` or `SSL_KEY_UPDATE_NOT_REQUESTED`. If so, why can't we do the change in a simpler and more clear way:
```
int SSL_get_key_update_type(const SSL *ssl) {
  if (ssl->s3->key_update_pending) return the_type_of_the_update_operation;
  else return SSL_KEY_UPDATE_NONE;
}
```

Maybe I'm misunderstanding something here, so could you please elaborate a bit what's supposed to happen with the connection when `SSL_key_update` is called with different params?

include/openssl/ssl.h
```diff
@@ -437,17 +437,30 @@ OPENSSL_EXPORT int SSL_write(SSL *ssl, const void *buf, int num);
 // it's own KeyUpdate message.
 #define SSL_KEY_UPDATE_NOT_REQUESTED 0
 
+// SSL_KEY_UPDATE_NONE should not be set by the user and is only used to
+// indicate that there is a pending key operation. OpenSSL indicates that -1 is
```
nebeid, line 441: I thought that 
```suggestion
// indicate that there isn't a pending key operation. OpenSSL indicates that -1 is
```

include/openssl/ssl.h
```diff
@@ -437,17 +437,30 @@ OPENSSL_EXPORT int SSL_write(SSL *ssl, const void *buf, int num);
...
+// used, so that it will be an invalid value for the on-the-wire protocol.
```
nebeid, line 442: Is it correct to say:
```suggestion
// used, so that it will be an invalid value for the on-the-wire protocol
// when calling SSL_key_update().
```
?

include/openssl/ssl.h
```diff
@@ -437,17 +437,30 @@ OPENSSL_EXPORT int SSL_write(SSL *ssl, const void *buf, int num);
 // it's own KeyUpdate message.
```
nebeid, line 437:
```suggestion
// its own KeyUpdate message.
```

ssl/ssl_test.cc
```diff
@@ -6173,6 +6173,40 @@ TEST(SSLTest, ZeroSizedWiteFlushesHandshakeMessages) {
...
+  bssl::UniquePtr<SSL> client, server;
+  ASSERT_TRUE(ConnectClientAndServer(&client, &server, client_ctx.get(),
+                                     server_ctx.get()));
+
+  // Test setting |SSL_key_update| with |SSL_KEY_UPDATE_REQUESTED|.
+  EXPECT_TRUE(SSL_key_update(client.get(), SSL_KEY_UPDATE_REQUESTED));
+  // |SSL_get_key_update_type| is used to determine whether a key update
+  // operation has been scheduled but not yet performed.
+  EXPECT_EQ(SSL_get_key_update_type(client.get()), SSL_KEY_UPDATE_REQUESTED);
+  EXPECT_EQ(0, SSL_write(client.get(), nullptr, 0));
```
nebeid, line 6197: Maybe it's obvious, but what does this step do? Does it advance the protocol?

samuel40791765 (commented):
> In AWS-LC current implementation, ssl->s3->key_update_pending indicates if a key update operation was issued but not finished. 

Yes, in AWS-LC `ssl->s3->key_update_pending` (boolean) is used to indicate if a key update operation was issued. However, `true` can mean either `SSL_KEY_UPDATE_REQUESTED` or `SSL_KEY_UPDATE_NOT_REQUESTED`. `false` means that `SSL_KEY_UPDATE_NONE`.
We need to return 3 possible states, but the current implementation only supports 2 states.

> If so, why can't we do the change in a simpler and more clear way:
int SSL_get_key_update_type(const SSL *ssl) {
  if (ssl->s3->key_update_pending) return the_type_of_the_update_operation;
  else return SSL_KEY_UPDATE_NONE;
}

That would be ideal, but the type of update operation is not saved in our `SSL` structure.  The difference between `SSL_KEY_UPDATE_REQUESTED` or `SSL_KEY_UPDATE_NOT_REQUESTED` is directly processed when calling `SSL_key_update`, and we're only left with `true` if we need to update the key.

> Maybe I'm misunderstanding something here, so could you please elaborate a bit what's supposed to happen with the connection when SSL_key_update is called with different params?

Yeah I was initially confused about the difference between `SSL_KEY_UPDATE_REQUESTED` or `SSL_KEY_UPDATE_NOT_REQUESTED` as well. I've added more documentation and an RFC link in `SSL_key_update` to help clarify.
* |SSL_KEY_UPDATE_NOT_REQUESTED| means that the sending keys for this connection will be updated and the peer will be informed of the change.
* |SSL_KEY_UPDATE_REQUESTED| means sending keys for this connection will be updated and the peer will be informed of the change along with a request for the peer to additionally update its sending keys.

RFC: https://datatracker.ietf.org/doc/html/rfc8446#section-4.6.3

include/openssl/ssl.h
```diff
@@ -434,20 +434,43 @@ OPENSSL_EXPORT int SSL_write(SSL *ssl, const void *buf, int num);
...
 // SSL_key_update queues a TLS 1.3 KeyUpdate message to be sent on |ssl|
-// if one is not already queued. The |request_type| argument must one of the
-// |SSL_KEY_UPDATE_*| values. This function requires that |ssl| have completed a
-// TLS >= 1.3 handshake. It returns one on success or zero on error.
+// if one is not already queued. The |request_type| argument must be either
+// |SSL_KEY_UPDATE_REQUESTED| or |SSL_KEY_UPDATE_NOT_REQUESTED|. This function
+// requires that |ssl| have completed a TLS >= 1.3 handshake. It returns one on
+// success or zero on error.
+//
+// If the updatetype parameter is set to |SSL_KEY_UPDATE_NOT_REQUESTED|, then
```
dkostic, line 452:
```suggestion
// If the |request_type| parameter is set to |SSL_KEY_UPDATE_NOT_REQUESTED|, then
```
?

include/openssl/ssl.h
```diff
@@ -434,20 +434,43 @@ OPENSSL_EXPORT int SSL_write(SSL *ssl, const void *buf, int num);
...
+// the sending keys for this connection will be updated and the peer will be
+// informed of the change.
+// If the updatetype parameter is set to |SSL_KEY_UPDATE_REQUESTED|, then the
```
dkostic, line 455:
```suggestion
// If the |request_type| parameter is set to |SSL_KEY_UPDATE_REQUESTED|, then the
```

include/openssl/ssl.h
```diff
@@ -434,20 +434,43 @@ OPENSSL_EXPORT int SSL_write(SSL *ssl, const void *buf, int num);
...
 // Note that this function does not _send_ the message itself. The next call to
 // |SSL_write| will cause the message to be sent. |SSL_write| may be called with
 // a zero length to flush a KeyUpdate message when no application data is
 // pending.
 OPENSSL_EXPORT int SSL_key_update(SSL *ssl, int request_type);
 
+// SSL_get_key_update_type returns the state of the pending key operation in
+// |ssl|. The type of pending key operation will be either
+// |SSL_KEY_UPDATE_REQUESTED| or |SSL_KEY_UPDATE_NOT_REQUESTED| if there is one,
+// and |SSL_KEY_UPDATE_NONE| if otherwise. This can be used to indicate whether
```
dkostic, line 470:
```suggestion
// and |SSL_KEY_UPDATE_NONE| otherwise. This can be used to indicate whether
```

include/openssl/ssl.h
samuel40791765: Yeah my bad `updatetype` is OpenSSL's...

nebeid (approved):

dkostic (approved):

PR #1018, opened 2023-05-17
Update speed.cc to support older versions of AWS-LC that have missing APIs.

andrewhop:
### Description of changes: 
If you want to compile the modern speed.cc with older versions of AWS-LC for a fair comparison we need to turn off some new algorithms.

Turns out the benchmark doesn't need ctrdrbg.h, the Jitter benchmark just needs jitterentropy.h.
### Testing:
Compiled locally, I don't think it's worth adding a dedicated CI test for this use case.


torben-hansen (approved):
I'm not sure this is the intended usage. But is it due to the fixed-up reuse of initialised contexts?

andrewhop:
> I'm not sure this is the intended usage. But is it due to the fixed-up reuse of initialised contexts?

Kind of, for a fair comparison of performance today vs an older releases IMO you need to use as close to the same benchmark code as possible. And in this case we changed what some of those encrypt benchmarks were measuring. This also helps if we add a new benchmark for an old algorithm and want to measure it against an old release of AWS-LC.

samuel40791765 (approved):

samuel40791765 (approved):

PR #1020, opened 2023-05-19
Fix RSA and ECDSA SigGen w/ SHA2-512/256

billbo-yang:
### Issues:
Resolves: CryptoAlg-1863

### Description of changes: 
When running the ACVP tool with vectors for RSA and ECDSA for the static build validation, the RSA SigGen w/ pkcs1v1.5 & SHA2-512/256 and ECDSA SigGen with SHA2-512/256 are fail to produce an output (no hard failure, the functions involved just return false). This is a problem as we're advertising that we support this functionality in our static build validation. This PR adds logic to handle SHA2-512/256 for the RSA and ECDSA SigGen use-cases.

### Call-outs:
N/A

### Testing:
Tested by running through the ACVP tool and verifying that an output is produced (and that said output is correct). Note that the ACVP tool currently isn't updated to handle RSA SigGen w/ SHA2-512/256 & pkcs1v1.5, so I had to modify it in order to test these changes. Those modifications are out of scope however and thus are coming in a separate PR.

`crypto_test` ran and passed locally as well


andrewhop (commented):
Please add an end to end test for this digest. Are there any known answer tests we can run?

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -2212,6 +2212,8 @@ static const struct ECDSATestVector kECDSATestVectors[] = {
      AWSLC_APPROVED},
     {NID_secp224r1, &EVP_sha512, AWSLC_APPROVED, AWSLC_APPROVED,
      AWSLC_APPROVED},
+    {NID_secp224r1, &EVP_sha512_256, AWSLC_APPROVED, AWSLC_APPROVED,
```
andrewhop, line 2215: Would these tests detect if the signature that is generated is invalid?

billbo-yang: They would not, added some new changes that do so 👍

crypto/evp_extra/evp_tests.txt
```diff
@@ -1655,6 +1655,16 @@ Digest = SHA256
...
 
+SignMessage = RSA-2048
+Digest = SHA512/256
+Input = "Hello world"
+Output = 8beb180f07b45b17af94c49badf8a7a630e89ffdaade1c3e2d0845ada27188a5d2d991926650773edfb5d36de5237abb82a1009e5dba5c7080b8fe19bd08227273420b73c060fb2efad53e0936b17cd341c7c02e588c7b0a3138c1205a7d46f8c5648267104aae1b285b390d90378aad9431499cf24feec2243ff668c98481940b08a10b63da8e7423f7e7e307485b6e7a6c73a5b68d70a36ab7cc5cb634773b6f1f8651932ea80c714501853e9fe1f2eff9b67fe03631e730345e02d14fa1da5fcbe52299c8b614fe5c4d17fdc9bd4f49b659154a95914fee33bce64c0cc4ba9aaf0809680b75cd891120025597a5b47e44cb3c669118591927133202768cd7
+
+VerifyMessage = RSA-2048
+Digest = SHA512/256
+Input = "Hello world"
+Output = 8beb180f07b45b17af94c49badf8a7a630e89ffdaade1c3e2d0845ada27188a5d2d991926650773edfb5d36de5237abb82a1009e5dba5c7080b8fe19bd08227273420b73c060fb2efad53e0936b17cd341c7c02e588c7b0a3138c1205a7d46f8c5648267104aae1b285b390d90378aad9431499cf24feec2243ff668c98481940b08a10b63da8e7423f7e7e307485b6e7a6c73a5b68d70a36ab7cc5cb634773b6f1f8651932ea80c714501853e9fe1f2eff9b67fe03631e730345e02d14fa1da5fcbe52299c8b614fe5c4d17fdc9bd4f49b659154a95914fee33bce64c0cc4ba9aaf0809680b75cd891120025597a5b47e44cb3c669118591927133202768cd7
```
andrewhop, line 1666: Is this duplicated on purpose?

billbo-yang: Discussed on slack but leaving this here for posterity:

Not a duplicate, the test above it is `SignMessage` while this one is `VerifyMessage`. They take the same parameter otherwise but just run `EVP_DigestSign()` vs. `EVP_DigestVerify()`.

andrewhop (approved):

torben-hansen:
Discussed offline. Reconsider adding this feature. pending.

torben-hansen (approved):
OK, but need to fix:
* service indicator which needs to be more granular - exclude this hash type for other padding types than PSS
* ACVP should be amended to be able to parse RSA pkcs 1v1.5 + sha2-512/256 from files.

Maybe this can be done in https://github.com/aws/aws-lc/pull/1022

PR #1022, opened 2023-05-19
Add Support for RSA SigGen w/ SHA2-512/256 & PKCS1v1.5 to the ACVP Tool

billbo-yang:
### Issues:
Resolves CryptoAlg-1863 in conjunction with PR https://github.com/aws/aws-lc/pull/1020

### Description of changes: 
We say that we support RSA SigGen with SHA2-512/256 and PKCS1v1.5, but the ACVP tool does not actually support vectors with that included. This PR adds support for handling it.

### Call-outs:
The length of the name generated by the go portion of the handler is larger than the current max name length supported by the modulewrapper. As such, I increased the max value to 35 from 30 in order to support it. The variable is only used in this case and so I felt it would be simpler to increase the max accepted characters in a name slightly as opposed to add dedicated logic each time a go module automatically generates a name larger than 30 characters.

Also note that until PR https://github.com/aws/aws-lc/pull/1020 is merged in, any test vector containing RSA SigGen with SHA2-512/256 & PKCS1v1.5 will fail, but now it won't be because of the ACVP tool.

### Testing:
Tested the ACVP tool locally and ensured everything ran correctly and that the output was as expected.


andrewhop (commented):
Do you have a test file we could include with this change?

andrewhop (approved):

PR #1032, opened 2023-05-25
Add few compatability functions needed for HAProxy

andrewhop:
### Issues:
Addresses CryptoAlg-1875

### Description of changes: 
* RAND_keep_random_devices_open which does nothing
* SSL_CTX_get_tlsext_status_cb returns the legacy OpenSSL OCSP callback
* X509_OBJECT_get0_X509_CRL returns the certificate revocation list (CRL) from an X509_OBJECT

### Testing:
I checked that these methods match OpenSSL but it would be good for the reviewer to check that as well. I can't add an integration test yet because there are other issues blocking the build. 


include/openssl/ssl.h
```diff
@@ -5286,6 +5286,13 @@ OPENSSL_EXPORT int SSL_CTX_set_tlsext_status_cb(SSL_CTX *ctx,
                                                 int (*callback)(SSL *ssl,
                                                                 void *arg));
 
+// SSL_CTX_get_tlsext_status_cb returns the legacy OpenSSL OCSP callback if set.
+OPENSSL_EXPORT int SSL_CTX_get_tlsext_status_cb(SSL_CTX *ctx, int (**callback)(SSL *, void *));
+// Provide a preprocessor macro for compatability with OpenSSL which uses a macro
+// to implement |SSL_CTX_get_tlsext_status_cb| and some callers expect to write
+// |#ifdef SSL_CTX_get_tlsext_status_cb|.
+#define SSL_CTX_get_tlsext_status_cb SSL_CTX_get_tlsext_status_cb
```
samuel40791765, line 5294: Should we define this at the bottom within `#if !defined(BORINGSSL_PREFIX)`?

crypto/rand_extra/rand_extra.c
```diff
@@ -71,4 +71,6 @@ const RAND_METHOD *RAND_get_rand_method(void) { return RAND_SSLeay(); }
 
 int RAND_set_rand_method(const RAND_METHOD *method) { return 1; }
 
+void RAND_keep_random_devices_open(int a) { }
```
darylmartin100, line 74: Can you add some documentation as to what this function was supposed to do and why it's safe to not have an implementation?

include/openssl/ssl.h
andrewhop: Yes, and the prefix build in the CI catches this

samuel40791765 (approved):

PR #1042, opened 2023-06-02
Fix OCSP integration tests to properly connect with endpoints

samuel40791765:
### Issues:
Fixes OCSP integration test issues we were having

### Description of changes: 
The new integration test was failing, once Amazon Trust Services reconfigured the behavior of their testing endpoints. The original way the certificate chain had been configured on the endpoint had been a bit off, and our tests had been depending on that irregular behavior.  The new behavior makes much more sense, but we've had to reconfigure our tests to align with them.
1. I've removed the expired cert tests, since these don't fail in OCSP anymore. They fail prematurely during the SSL connection due to the expired cert. This new endpoint behavior is more aligned with real-world behavior, but it doesn't belong within the realm of OCSP testing anymore.
2. The endpoints had been sending the expected root certificate along with the server's leaf/intermediate certificates over the wire. We had been dependent on this behavior, but this was wrong in the first place. We now store the intended root testing certificates in our local testing environment, although this means we'll have to update the certificates about 15 years down the line. I've added a warning print function so we can be aware of this.
3. Some of these endpoints don't immediately present the intended certificate chain to test. This is resolved by calling `SSL_set_tlsext_host_name`, which forces the connection to use the leaf certificate with the same host name as the endpoint.

### Call-outs:
N/A

### Testing:
CI, test adjustments.


crypto/ocsp/ocsp_integration_test.cc
```diff
@@ -18,6 +19,35 @@
...
+  // Get the current time and add 1 year to it.
+  int64_t warning_time = time(nullptr) + 31536000;
+
+  if (X509_cmp_time_posix(X509_get_notAfter(cert), warning_time) < 0) {
+    fprintf(stdout, "\n\n\n\n");
+    X509_NAME_print_ex_fp(stdout, X509_get_subject_name(cert), 0, 0);
+    fprintf(stdout, " WILL EXPIRE IN A YEAR, PLEASE REPLACE ME WITH THE NEW"
+            " EXPECTED ROOT CERTIFICATE FROM THE ENDPOINT.\n\n\n\n");
+  }
+}
```
andrewhop, line 34: 10/10. Did you test this locally by changing the warning time to 15 years and see it print out?

Hypothetically could we add a cloudwatch alarm on a regex looking at the codebuild logs to ticket us?

crypto/ocsp/ocsp_integration_test.cc
```diff
@@ -18,6 +19,35 @@
...
+
+static X509_STORE *SetupTrustStore() {
+  X509_STORE *trust_store = X509_STORE_new();
+  for (int i = 1; i <= 3; i++) {
+    bssl::UniquePtr<X509> ca_cert(CertFromPEM(
+        GetTestData(
+            std::string("crypto/ocsp/test/integration-tests/AmazonRootCA" +
+                        std::to_string(i) + ".pem")
+                .c_str())
+            .c_str()));
```
andrewhop, line 44: Can you remove the outermost round trip through the std::string?
```suggestion
            ("crypto/ocsp/test/integration-tests/AmazonRootCA" +
                        std::to_string(i) + ".pem") .c_str());
```

crypto/ocsp/ocsp_integration_test.cc
```diff
@@ -181,31 +205,48 @@ TEST_P(OCSPIntegrationTest, AmazonTrustServices) {
...
-  // the Cert should be the one right after it.
-  STACK_OF(X509) *chain = SSL_get_peer_cert_chain(ssl.get());
-  X509 *certificate = sk_X509_value(chain, 0);
-  ASSERT_TRUE(certificate);
-  X509 *issuer = sk_X509_value(chain, 1);
+  STACK_OF(X509) *chain = SSL_get_peer_full_cert_chain(ssl.get());
+  EXPECT_EQ(SSL_get_verify_result(ssl.get()), X509_V_OK);
+  // Leaf to be verified should be the first one in the chain.
+  X509 *subject = sk_X509_value(chain, 0);
+  ASSERT_TRUE(subject);
```
andrewhop, line 224: Is this getting the subject from the cert? Or is this still the entire x509 cert? 
```suggestion
  X509 *certificate = sk_X509_value(chain, 0);
  ASSERT_TRUE(certificate);
```

andrewhop (commented):
Please add a readme describing where the AmazonRootCAs are from, how to update them, and update the warning message to point at the readme.

crypto/ocsp/ocsp_integration_test.cc
```diff
@@ -181,31 +205,48 @@ TEST_P(OCSPIntegrationTest, AmazonTrustServices) {
...
   bssl::UniquePtr<SSL_CTX> ssl_ctx(SSL_CTX_new(TLS_method()));
   ASSERT_TRUE(ssl_ctx);
+
+  // Set up trust store.
+  bssl::UniquePtr<X509_STORE> trust_store(SetupTrustStore());
+  ASSERT_TRUE(SSL_CTX_set1_verify_cert_store(ssl_ctx.get(), trust_store.get()));
   bssl::UniquePtr<BIO> bio(BIO_new_socket(sock, BIO_CLOSE));
   bssl::UniquePtr<SSL> ssl(SSL_new(ssl_ctx.get()));
+  // Set the host name, so we don't retrieve the wrong intended certificate
+  // chain.
```
andrewhop, line 215: NP: this has more to do with letting the endpoint know what we're trying to connect to, and in this case SNI is used to host multiple different websites on one server.

crypto/ocsp/ocsp_integration_test.cc
samuel40791765: Both functions involved here take a `const *char`, but C++ doesn't allow directly appending of a value to a const char, thus the round trip :( 
I've used `ostringstream` to clean things up and make things prettier though

crypto/ocsp/ocsp_integration_test.cc
samuel40791765: Yeah I tested it by adding more time to it and got warnings.

The test output only seems to get dumped out when a test failure happens, but I think we can configure an alarm to warn us. Test failures always happen anyways :)

crypto/ocsp/ocsp_integration_test.cc
samuel40791765: Re-clarified the comment

andrewhop (approved):

skmcgrail (approved):

crypto/ocsp/ocsp_integration_test.cc
samuel40791765: I created `CryptoAlg-1894` to track making the alarm. I looked into it briefly, but I'm not well-versed enough in CloudWatch and ticketing configurations to make this a one-day task.

PR #1044, opened 2023-06-03
Test adjustments and add expired signing OCSP cert test

samuel40791765:
### Description of changes: 
Follow up from recent PR reviews on OCSP. 
* Clean up tests and refactor
* Add OCSP test where the OCSP response is signed by an expired certificate

### Call-outs:
N/A

### Testing:
CI


skmcgrail (approved):

crypto/ocsp/ocsp_test.cc
```diff
@@ -197,116 +193,97 @@ static void CheckOCSP_CERTSTATUS(
...
-     OCSP_RESPONSE_STATUS_SUCCESSFUL, "successful", OCSP_VERIFYSTATUS_SUCCESS,
-     OCSP_RESPFINDSTATUS_SUCCESS, V_OCSP_CERTSTATUS_UNKNOWN, "unknown"},
+     OCSP_VERIFYSTATUS_SUCCESS, OCSP_RESPFINDSTATUS_SUCCESS,
+     V_OCSP_CERTSTATUS_UNKNOWN, "unknown"},
     // for the requested certificate. (So this would be a completely valid
     // response to a different OCSP request for the other certificate.)
     {"ocsp_response", "ca_cert", "server_ecdsa_cert", EVP_sha1(),
-     OCSP_RESPONSE_STATUS_SUCCESSFUL, "successful", OCSP_VERIFYSTATUS_SUCCESS,
-     OCSP_RESPFINDSTATUS_ERROR, 0, nullptr},
+     OCSP_VERIFYSTATUS_SUCCESS, OCSP_RESPFINDSTATUS_ERROR, 0, nullptr},
```
andrewhop, line 237: Just curious why do some of these use a define like V_OCSP_CERTSTATUS_UNKNOWN, and here you use a 0? What happens if you get the status for something that fails? I'm also a little concerned since 0 is V_OCSP_CERTSTATUS_GOOD if I'm looking in the right spot.

samuel40791765: Yeah that's on me for using `0` which has the same value as `V_OCSP_CERTSTATUS_GOOD`. I should of used `-1` so it doesn't coincide with any of the other CERTSTATUS codes. I'll change it.
* OCSP_CERTSTATUS has three different possibilities: "good", "revoked", and "unknown". This value is extracted with `OCSP_resp_find_status`.
* `OCSP_resp_find_status` has to succeed in order to extract a status, so it's impossible to get a "cert status" for something that fails. It'll result in an error if it's ill-defined, which is why I was defining it with `0` since there was essentially no value to compare to.

andrewhop (dismissed):

skmcgrail (dismissed):

skmcgrail (approved):

andrewhop (approved):

PR #1045, opened 2023-06-05
Support changing OPENSSL_armcap with environment variable on Apple 64-bit ARM systems

billbo-yang:
### Issues:
Resolves CryptoAlg-1873

### Description of changes: 
AWS-LC currently uses a static ARM capabilities variable for 64 bit ARM Apple devices, as we know that certain hardware features are always present on them. This change gives Apple ARM users an option to define at runtime the CPU capabilities through an environment variable to determine whether to run the hardware vs. non-hardware implementation of an algorithm is used.

### Call-outs:
Refactored `handle_cpu_env()` function in `cpu_aarch64_linux.c` to a general `cpu_aarch64.c` file to avoid duplicating the code in `cpu_aarch64_apple.c`. Modified CMakeFIle accordingly.

### Testing:
Tests pass locally, confirmed that non-hardware APIs are used when `OPENSSL_armcap=0:0` and that the correct hardware APIs are used when it is left unset. Behavior for testing this already exists in `all_tests.json`, used by `all_tests.go`.

### DIFF
--- a/crypto/fipsmodule/cpucap/cpu_aarch64.c
+++ b/crypto/fipsmodule/cpucap/cpu_aarch64.c
@@ -26,7 +26,8 @@ void handle_cpu_env(uint32_t *out, const char *in) {
 
   // Detect if the user is trying to use the environment variable to set
   // a capability that is _not_ available on the CPU:
-  // If the runtime capability check (e.g via getauxval() on Linux) returned a non-zero hwcap in `armcap` (out)
+  // If the runtime capability check (e.g via getauxval() on Linux)
+  // returned a non-zero hwcap in `armcap` (out)
   // and a bit set in the requested `v` is not set in `armcap`,
   // abort instead of crashing later.
   // The case of invert cannot enable an unexisting capability;
--- a/crypto/fipsmodule/cpucap/cpu_aarch64.c
+++ b/crypto/fipsmodule/cpucap/cpu_aarch64.c
@@ -1,6 +1,8 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR ISC
 
+#if defined(OPENSSL_AARCH64) && !defined(OPENSSL_STATIC_ARMCAP)
+
 #include "cpu_aarch64.h"
 
 void handle_cpu_env(uint32_t *out, const char *in) {
@@ -45,3 +47,5 @@ void handle_cpu_env(uint32_t *out, const char *in) {
     out[0] = v;
   }
 }
+
+#endif // OPENSSL_AARCH64 && !OPENSSL_STATIC_ARMCAP
--- a/crypto/fipsmodule/cpucap/cpu_aarch64.c
+++ b/crypto/fipsmodule/cpucap/cpu_aarch64.c
@@ -1,0 +1,47 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include "cpu_aarch64.h"
+
+void handle_cpu_env(uint32_t *out, const char *in) {
+  const int invert = in[0] == '~';
+  const int or = in[0] == '|';
+  const int skip_first_byte = invert || or;
+  const int hex = in[skip_first_byte] == '0' && in[skip_first_byte+1] == 'x';
+  uint32_t armcap = out[0];
+
+  int sscanf_result;
+  uint32_t v;
+  if (hex) {
+    sscanf_result = sscanf(in + skip_first_byte + 2, "%" PRIx32, &v);
+  } else {
+    sscanf_result = sscanf(in + skip_first_byte, "%" PRIu32, &v);
+  }
+
+  if (!sscanf_result) {
+    return;
+  }
+
+  // Detect if the user is trying to use the environment variable to set
+  // a capability that is _not_ available on the CPU:
+  // If getauxval() returned a non-zero hwcap in `armcap` (out)
+  // and a bit set in the requested `v` is not set in `armcap`,
+  // abort instead of crashing later.
+  // The case of invert cannot enable an unexisting capability;
+  // it can only disable an existing one.
+  if (!invert && armcap && (~armcap & v))
+  {
+    fprintf(stderr,
+            "Fatal Error: HW capability found: 0x%02X, but HW capability requested: 0x%02X.\n",
+            armcap, v);
+    exit(1);
+  }
+
+  if (invert) {
+    out[0] &= ~v;
+  } else if (or) {
+    out[0] |= v;
+  } else {
+    out[0] = v;
+  }
+}
--- a/crypto/fipsmodule/cpucap/cpu_aarch64.c
+++ b/crypto/fipsmodule/cpucap/cpu_aarch64.c
@@ -26,7 +26,7 @@ void handle_cpu_env(uint32_t *out, const char *in) {
 
   // Detect if the user is trying to use the environment variable to set
   // a capability that is _not_ available on the CPU:
-  // If getauxval() returned a non-zero hwcap in `armcap` (out)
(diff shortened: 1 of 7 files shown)


crypto/fipsmodule/cpucap/cpu_aarch64.c
```diff
@@ -0,0 +1,51 @@
...
+    sscanf_result = sscanf(in + skip_first_byte, "%" PRIu32, &v);
+  }
+
+  if (!sscanf_result) {
+    return;
+  }
+
+  // Detect if the user is trying to use the environment variable to set
+  // a capability that is _not_ available on the CPU:
+  // If getauxval() returned a non-zero hwcap in `armcap` (out)
```
nebeid, line 29:
```suggestion
  // If the runtime capability check (e.g via getauxval() on Linux) returned a non-zero hwcap in `armcap` (out)
```
(The comment lines would need to be filled to original length)

crypto/fipsmodule/cpucap/cpu_aarch64_apple.c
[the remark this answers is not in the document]
billbo-yang: I don't believe this needs to be removed. I've tested the changes on both a m1 macbook as well as the iOS device we are concerned about by stepping through the code with a debugger and confirming that the non-hw implementation is used when OPENSSL_armcap is set to 0:0. I think this line is setting it for 32-bit Apple processors, which at this point is limited to very old (pre-2007 iirc) machines. I'll update the title and related tickets to reflect that we're only interested in aarch64

crypto/fipsmodule/cpucap/cpu_aarch64.c
```diff
@@ -26,7 +26,7 @@ void handle_cpu_env(uint32_t *out, const char *in) {
 
   // Detect if the user is trying to use the environment variable to set
   // a capability that is _not_ available on the CPU:
-  // If getauxval() returned a non-zero hwcap in `armcap` (out)
+  // If the runtime capability check (e.g via getauxval() on Linux) returned a non-zero hwcap in `armcap` (out)
```
nebeid, line 29: Line length

crypto/fipsmodule/cpucap/internal.h
```diff
@@ -167,7 +167,7 @@ extern uint32_t OPENSSL_armcap_P;
 // |CRYPTO_is_ARMv8_AES_capable| and |CRYPTO_is_ARMv8_PMULL_capable|
 // for checking the support for AES and PMULL instructions, respectively.
 OPENSSL_INLINE int CRYPTO_is_NEON_capable(void) {
-#if defined(OPENSSL_STATIC_ARMCAP_NEON) || defined(__ARM_NEON)
```
nebeid, line 170: Why were the second parts of the checks removed? Aren't these possible compiler flags that can set the capability statically?

dkostic:
Could we add the same `BORINGSLL_DISPATCH_TEST` as we do for x86?

crypto/fipsmodule/cpucap/internal.h
billbo-yang: According to [ACLE](https://developer.arm.com/documentation/ihi0053/d/?lang=en), `__ARM_NEON` is always set to 1 for AARCH64 platforms. As such, if we want `OPENSSL_armcap` to determine use of the C implementation, we can't check for this flag since it will always be set. We could use the `-march` flag when compiling to disable NEON, but since `OPENSSL_armcap` is supposed to determine processor capability at runtime I decided on this approach to reduce confusion.

`cpu_aarch64_apple.c` (as well as the counterpart files for other CPUs) still set `OPENSSL_armcap` to be the correct value for the processor in `OPENSSL_cpuid_setup()`, so removing the check doesn't disable NEON entirely, either.

This applies to `__ARM_FEATURE_AES` as well.

crypto/fipsmodule/cpucap/cpu_aarch64.c
billbo-yang: Yikes, I misunderstood the previous commit suggestion for this line! Will add 👍

billbo-yang:
> Could we add the same `BORINGSLL_DISPATCH_TEST` as we do for x86?

Discussed offline, will submit another PR with these changes as the scope is larger than the context of this one.

crypto/fipsmodule/cpucap/internal.h
nebeid: Couldn't these values be used on ARM processors (32-bit)? We have users who build AWS-LC on them.

billbo-yang: [cpucap.c](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/cpucap/cpucap.c#L53) detects if LC is built on 32-bit ARM and sets `OPENSSL_armcap` accordingly, so when we call a function like `CRYPTO_is_NEON_capable()`, even without a check for these values, it will correctly detect the CPU capabilities.

nebeid: I may be missing something, but I think if `OPENSSL_STATIC_ARMCAP` is defined but not `OPENSSL_STATIC_ARMCAP_NEON` due to an old compiler that defines `__ARM_NEON`, it will not reach the check for `OPENSSL_armcap` and will return 0.

dkostic (dismissed):

crypto/fipsmodule/cpucap/internal.h
billbo-yang: I changed the functions to no longer include the `ifdef` statements, similar to how the capability functions for x86 processors in the same file do it. Since `OPENSSL_armcap_P` is set correctly even when `OPENSSL_STATIC_ARMCAP` is defined, the functions should still return the right result.

Either way, a `BORINGSSL_DISPATCH_TEST` to ensure that this doesn't break anything will be introduced in a separate PR (which we should merge before this one)

nebeid: I think we'd also need to measure the performance in FIPS build with and without the ifdefs for static build (possibly for dynamic as well) when the STATIC_ARMCAPs are defined. Fetching the OPENSSL_armcap_P variable had some implications on the static FIPS build in x86 https://github.com/aws/aws-lc/pull/856.

billbo-yang: Performed a benchmark on both the dynamic and static builds comparing performance of AES on main to the PR branch:
* FIPS mode set
* `OPENSSL_STATIC_ARMCAP` and other static macros set
* Graviton2 on AL2

No performance differences between main & PR for either the dynamic nor static builds.

Full results can be seen here:
[aws-lc-dynamic-staticarmcap-aes_main_vs_nonhw_bm.csv](https://github.com/aws/aws-lc/files/12164501/aws-lc-dynamic-staticarmcap-aes_main_vs_nonhw_bm.csv)
[aws-lc-static-staticarmcap-aes_main_vs_nonhw_bm.csv](https://github.com/aws/aws-lc/files/12164502/aws-lc-static-staticarmcap-aes_main_vs_nonhw_bm.csv)

nebeid (dismissed):

crypto/fipsmodule/cpucap/internal.h
```diff
@@ -167,33 +167,15 @@ extern uint32_t OPENSSL_armcap_P;
...
-#if defined(OPENSSL_STATIC_ARMCAP_PMULL) || defined(__ARM_FEATURE_AES)
-  return 1;
-#elif defined(OPENSSL_STATIC_ARMCAP)
-  return 0;
-#else
   return (OPENSSL_armcap_P & ARMV8_PMULL) != 0;
-#endif
 }
 
 OPENSSL_INLINE int CRYPTO_is_ARMv8_GCM_8x_capable(void) {
```
dkostic, line 181: how about this and the next function? Should we remove the `#if defined(OPENSSL_STATIC_ARMCAP)` branch in those too?

nebeid (approved):

PR #1046, opened 2023-06-06
add integration test CI dimension for MariaDB

samuel40791765:
### Issues:
No Issue cut

### Description of changes: 
We've recently resolved missing symbols for MariaDB and are looking to add a CI dimension to ensure we don't break the build. Patches and flags necessary were taken from internal work on MariaDB.

### Call-outs:
N/A

### Testing:
New CI dimension


tests/ci/integration/run_mariadb_integration.sh
```diff
@@ -0,0 +1,93 @@
...
+# TODO: Remove this when we make an upstream contribution.
+function mariadb_patch() {
+  for patchfile in $(find -L "${MARIADB_PATCH_FOLDER}" -type f -name '*.patch'); do
+    echo "Apply patch $patchfile..."
+    patch -p1 --quiet -i "$patchfile"
+  done
+}
+
+# Get latest mariadb version, we can pin to a specific version if MariaDB's code changes break us too often.
+git clone https://github.com/MariaDB/server.git ${MARIADB_SRC_FOLDER} --depth 1
```
justsmth, line 83: Since we're also applying patches, testing with the latest commit seems particularly volatile. 

I'm curious about why we [test a specific tag from mysql](https://github.com/aws/aws-lc/pull/1051) (a project that doesn't require patches), but we test the latest commit from MariaDB? 

I'm ok with either decision, it just seems we should be consistent about how we integration test.

samuel40791765: We've noticed that MySQL's changes have been really closely tied with OpenSSL and are continuously introducing changes that introduce new dependencies on OpenSSL symbols. Asides from that, we recently had multiple test failures due to a bug in the latest version of mySQL 2-3 months ago. The bug was resolved a few weeks ago, but it shows that we can't quite rely on the latest mySQL version as a CI target.
MariaDB has less churn and there's a better chance of us sending these patches upstream to integrate with them. I could be proven wrong, but we can easily pin a version if that's the case. MySQL actually needs patches too, but we just haven't introduced them yet since we haven't turned on testing.
* https://github.com/aws/aws-lc/pull/1063

tests/ci/integration/mariadb_patch/0002-WIP-Stop-using-the-unchecked-version-of-DES_set_key.patch
```diff
@@ -0,0 +1,115 @@
...
+ 		       (uchar *) start, (int) (end-start),1,
+ 		       (uchar *) &keyblock,
+-		       ivec);
+-	DES_set_key_unchecked(&keyblock.key1,&(des_keyschedule[(int)offset].ks1));
+-	DES_set_key_unchecked(&keyblock.key2,&(des_keyschedule[(int)offset].ks2));
+-	DES_set_key_unchecked(&keyblock.key3,&(des_keyschedule[(int)offset].ks3));
++                       ivec.bytes);
++	DES_set_key(&keyblock.key1,&(des_keyschedule[(int)offset].ks1));
++	DES_set_key(&keyblock.key2,&(des_keyschedule[(int)offset].ks2));
++	DES_set_key(&keyblock.key3,&(des_keyschedule[(int)offset].ks3));
```
andrewhop, line 72: Did you consider adding a simple function that maps DES_set_key_unchecked -> DES_set_key? This isn't the only project to hit this issue.

tests/ci/integration/mariadb_patch/0004-WIP-Skip-test-that-is-disabled-when-using-OpenSSL-1..patch
```diff
@@ -0,0 +1,29 @@
+From 85cf96e689ef0df5ddefe12cfbcfb18a80e96210 Mon Sep 17 00:00:00 2001
+From: Robin Newhouse <robinnew@amazon.com>
+Date: Mon, 15 May 2023 22:37:36 +0000
+Subject: [PATCH 4/4] WIP: Skip test that is disabled when using OpenSSL 1.1.1
+
+OpenSSL 1.1.1 and AWS-LC use different cipher names than OpenSSL 1.0.2
```
andrewhop, line 6: Is this implying MariaDB only supports 1.0.2 right now?

tests/ci/integration/mariadb_patch/0001-WIP-Avoid-define-template-_template-so-AWS-LC-builds.patch
```diff
@@ -0,0 +1,56 @@
+From 10318f2e70b2cab11537b4fcf10266de254d8452 Mon Sep 17 00:00:00 2001
+From: Otto Kekalainen <ottoke@amazon.com>
+Date: Mon, 28 Nov 2022 20:27:35 -0800
+Subject: [PATCH 1/4] WIP: Avoid '#define template _template' so AWS-LC builds
+ can run
+
+Disable the WolfSSL specific bug fix as it breaks AWS-LC builds with
+errors like:
```
andrewhop, line 8: Why is the AWS-LC build enabling WolfSSL bug fixes?

tests/ci/integration/mariadb_patch/0003-WIP-Use-replace_regex-in-mysql-test-run-to-make-SSL-.patch
```diff
@@ -0,0 +1,28 @@
+From 557b4261a3120b037e2c705ce94453bc63d08823 Mon Sep 17 00:00:00 2001
+From: Otto Kekalainen <ottoke@amazon.com>
+Date: Mon, 28 Nov 2022 21:03:50 -0800
+Subject: [PATCH 3/4] WIP: Use `--replace_regex` in mysql-test-run to make SSL
+ tests pass
+
+Another potential area worthy of upstream contribution.
+
+The new error messages _seem_ to follow uppercase and underscore format.
```
andrewhop, line 9: Who is different here? Is this 1.0.2 vs AWS-LC **and** 1.1.1? Or is AWS-LC the only different one?

tests/ci/integration/run_mariadb_integration.sh
```diff
@@ -0,0 +1,93 @@
...
+}
+
+function mariadb_run_tests() {
+  pushd ${MARIADB_BUILD_FOLDER}
+  # More complicated integration tests. mtr expects to be launched in-place and with write access to it's own directories
+  #
+  # main.plugin_load passes, but is skipped over since it generates a warning when we run the script in Codebuild. Warnings will cause
+  # a failure in MariaDB's test runs, unless --nowarnings is turned on. The warning is not reproducable in Gitlab's CI or any local
+  # container runs. This test isn't relevant to AWS-LC integration so we skip over the Codebuild specific issue for now.
+  pushd mysql-test
```
andrewhop, line 64:
```suggestion
  pushd ${MARIADB_BUILD_FOLDER}/mysql-test
  # More complicated integration tests. mtr expects to be launched in-place and with write access to it's own directories
  #
  # main.plugin_load passes, but is skipped over since it generates a warning when we run the script in Codebuild. Warnings will cause
  # a failure in MariaDB's test runs, unless --nowarnings is turned on. The warning is not reproducable in Gitlab's CI or any local
  # container runs. This test isn't relevant to AWS-LC integration so we skip over the Codebuild specific issue for now.
```

tests/ci/integration/run_mariadb_integration.sh
```diff
@@ -0,0 +1,93 @@
...
+  echo "main.mysqldump : Field separator argument is not what is expected; check the manual when executing 'SELECT INTO OUTFILE'
+main.flush_logs_not_windows : query 'flush logs' succeeded - should have failed with error ER_CANT_CREATE_FILE (1004)
+main.mysql_upgrade_noengine : upgrade output order does not match the expected
+main.plugin_load : This test generates a warning in Codebuild. Skip over since this isn't relevant to AWS-LC. "> skiplist
+  ./mtr --suite=main --force --parallel=auto --skip-test-list=${MARIADB_BUILD_FOLDER}/mysql-test/skiplist --retry-failure=2
+  popd
+  popd
```
andrewhop, line 71:
```suggestion
  popd
```

tests/ci/integration/mariadb_patch/0002-WIP-Stop-using-the-unchecked-version-of-DES_set_key.patch
samuel40791765: I took this patch directly from the work ottoke@ had been doing. But I can add it in a subsequent PR and remove this patch once we do so

tests/ci/integration/mariadb_patch/0003-WIP-Use-replace_regex-in-mysql-test-run-to-make-SSL-.patch
samuel40791765: This is referring to our error messages which look like this `SSLV3_ALERT_CERTIFICATE_REVOKED`.
OpenSSL's error messages have lower cases and spaces instead: `sslv3 alert certificate revoked`.

tests/ci/integration/mariadb_patch/0004-WIP-Skip-test-that-is-disabled-when-using-OpenSSL-1..patch
samuel40791765: Nah, I think the original test was just more restrictive. It only allowed OpenSSL1.1.1 to skip the test. This adds both AWS-LC and OpenSSL1.1.1 to skip it.

tests/ci/integration/mariadb_patch/0001-WIP-Avoid-define-template-_template-so-AWS-LC-builds.patch
samuel40791765: Sorry, I don't have a good answer to what MariaDB/WolfSSL was doing under the hood here :( If it helps, I looked into the files and it seems like the WolfSSL bug fix was being performed, regardless of what backend crypto was being used.
* https://github.com/MariaDB/server/blob/11.1/include/violite.h#L150

justsmth (dismissed):

andrewhop (dismissed):

andrewhop (approved):

justsmth (approved):

PR #1050, opened 2023-06-07
add fuzz testing for OCSP_parse_url

samuel40791765:
### Issues:
Resolves `V924834538`

### Description of changes: 
Adds fuzz testing for the URL parsing done in `OCSP_parse_url `.

### Call-outs:
`fuzzer(ocsp_http)` was unintentionally removed when we updated the OCSP branch with main. We'll have to be more careful when updating feature branches.

### Testing:
Followed runbook for fuzzing in https://github.com/aws/aws-lc/blob/main/FUZZING.md
Fuzz tests were ran for 30 minutes to generate corpus.

the ISC license.


fuzz/ocsp_parse_url.cc
```diff
@@ -0,0 +1,32 @@
...
+  char *host = nullptr;
+  char *path = nullptr;
+  char *port = nullptr;
+  int is_ssl;
+
+  if (OCSP_parse_url(input.c_str(), &host, &port, &path, &is_ssl)) {
+    // Check that parameters have been assigned values.
+    if (!host && !path && !port && (is_ssl != 1 && is_ssl != 0)) {
+      return 1;
+    }
```
andrewhop, line 20: Is the parsing all or nothing? Could we end up parsing a part of the URL?

fuzz/CMakeLists.txt
```diff
@@ -21,6 +21,8 @@ fuzzer(der_roundtrip)
 fuzzer(dtls_client ssl)
 fuzzer(dtls_server ssl)
 fuzzer(ocsp)
+fuzzer(ocsp_http)
```
andrewhop, line 24: Where is this fuzzer defined? I don't get why what this is building and why it's not failing.

fuzz/ocsp_parse_url.cc
```diff
@@ -0,0 +1,32 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include <openssl/mem.h>
+#include <openssl/ocsp.h>
+#include <string>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *buf, size_t len) {
+  std::string input(reinterpret_cast<const char *>(buf),
+                    std::next(reinterpret_cast<const char *>(buf), len));
```
andrewhop, line 10: What is the std::next part doing? Does this add a null terminator to the end of the string? Is it safe to pass this to OCSP_parse_url?

fuzz/CMakeLists.txt
skmcgrail: Looks like it is already there in the tree, just wasn't enabled.

fuzz/ocsp_parse_url.cc
samuel40791765: Yeah it's all or nothing. We have some test cases covering certain empty components.
* https://github.com/aws/aws-lc/blob/ocsp-request/crypto/ocsp/ocsp_test.cc#L1116-L1117

skmcgrail (dismissed):

andrewhop (approved):

skmcgrail (approved):

PR #1051, opened 2023-06-08
add integration build CI dimension for mySQL

samuel40791765:
### Issues:
Resolves `CryptoAlg-1859`

### Description of changes: 
We've recently resolved missing symbols for mySQL and are looking to add a CI dimension to ensure we don't break the build. This doesn't run the tests yet since there are test failures still pending investigation, so this just runs the build.

### Call-outs:
N/A

### Testing:
New CI dimension

### DIFF
--- a/tests/ci/codebuild/integration/mysql_integration.yml
+++ b/tests/ci/codebuild/integration/mysql_integration.yml
@@ -1,0 +1,13 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+version: 0.2
+
+env:
+  variables:
+    GOPROXY: https://proxy.golang.org,direct
+
+phases:
+  build:
+    commands:
+      - ./tests/ci/integration/run_mysql_integration.sh
--- a/tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
+++ b/tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
@@ -45,6 +45,7 @@ RUN set -ex && \
     flex \
     bison \
     curl \
+    jq \
     unzip && \
     # Based on https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-linux.html
     # The awscli is used to publish data to CloudWatch Metrics in some jobs. This requires additional IAM permission
@@ -59,7 +60,7 @@ RUN set -ex && \
     # Download a copy of LLVM's libcxx which is required for building and running with Memory Sanitizer
     git clone https://github.com/llvm/llvm-project.git --branch llvmorg-11.1.0  --depth 1 && \
     cd llvm-project && rm -rf $(ls -A | grep -Ev "(libcxx|libcxxabi)") && \
-    apt-get --purge remove -y curl unzip && \
+    apt-get --purge remove -y unzip && \
     apt-get autoremove --purge -y && \
     apt-get clean && \
     apt-get autoclean && \
--- a/tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
+++ b/tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
@@ -7,6 +7,7 @@ SHELL ["/bin/bash", "-c"]
 
 ENV DEBIAN_FRONTEND=noninteractive
 
+ENV BOOST_PACKAGE_NAME=boost_1_77_0
 ENV DEPENDENCIES_DIR=/home/dependencies
 ENV LLVM_PROJECT_HOME=${DEPENDENCIES_DIR}/llvm-project
 ENV ASAN_SYMBOLIZER_PATH=/usr/bin/llvm-symbolizer
@@ -50,9 +51,12 @@ RUN set -ex && \
     unzip awscliv2.zip && \
     ./aws/install --bin-dir /usr/bin && \
     rm -rf awscliv2.zip aws/ && \
-    # Download a copy of LLVM's libcxx which is required for building and running with Memory Sanitizer
     mkdir -p ${DEPENDENCIES_DIR} && \
     cd ${DEPENDENCIES_DIR} && \
+    # Extract and install Boost, dependency needed for mySQL.
+    wget https://boostorg.jfrog.io/artifactory/main/release/1.77.0/source/${BOOST_PACKAGE_NAME}.tar.bz2 && \
+    tar xfj ${BOOST_PACKAGE_NAME}.tar.bz2 && mv ./${BOOST_PACKAGE_NAME} ./boost && \
+    # Download a copy of LLVM's libcxx which is required for building and running with Memory Sanitizer
     git clone https://github.com/llvm/llvm-project.git --branch llvmorg-11.1.0  --depth 1 && \
     cd llvm-project && rm -rf $(ls -A | grep -Ev "(libcxx|libcxxabi)") && \
     apt-get --purge remove -y curl unzip && \
--- a/tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
+++ b/tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
@@ -8,6 +8,8 @@ SHELL ["/bin/bash", "-c"]
 ENV DEBIAN_FRONTEND=noninteractive
 
 ENV BOOST_PACKAGE_NAME=boost_1_77_0
+ENV BOOST_TARBALL="${BOOST_PACKAGE_NAME}.tar.bz2"
+ENV BOOST_SRC_URL="https://boostorg.jfrog.io/artifactory/main/release/1.77.0/source/${BOOST_TARBALL}"
 ENV DEPENDENCIES_DIR=/home/dependencies
 ENV LLVM_PROJECT_HOME=${DEPENDENCIES_DIR}/llvm-project
 ENV ASAN_SYMBOLIZER_PATH=/usr/bin/llvm-symbolizer
@@ -23,7 +25,6 @@ RUN set -ex && \
     apt-get -y --no-install-recommends install \
     software-properties-common \
     cmake \
-    cpp \
     make \
     ninja-build \
     perl \
@@ -53,9 +54,8 @@ RUN set -ex && \
     rm -rf awscliv2.zip aws/ && \
     mkdir -p ${DEPENDENCIES_DIR} && \
     cd ${DEPENDENCIES_DIR} && \
-    # Extract and install Boost, dependency needed for mySQL.
-    wget https://boostorg.jfrog.io/artifactory/main/release/1.77.0/source/${BOOST_PACKAGE_NAME}.tar.bz2 && \
-    tar xfj ${BOOST_PACKAGE_NAME}.tar.bz2 && mv ./${BOOST_PACKAGE_NAME} ./boost && \
+    # Extract and install Boost 1.77.0. mySQL 8.33 depends on this specific version.
+    wget ${BOOST_SRC_URL} && tar xfj ${BOOST_TARBALL} && mv ./${BOOST_PACKAGE_NAME} ./boost && \
     # Download a copy of LLVM's libcxx which is required for building and running with Memory Sanitizer
     git clone https://github.com/llvm/llvm-project.git --branch llvmorg-11.1.0  --depth 1 && \
     cd llvm-project && rm -rf $(ls -A | grep -Ev "(libcxx|libcxxabi)") && \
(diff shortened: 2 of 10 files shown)


tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
```diff
@@ -22,6 +23,7 @@ RUN set -ex && \
     apt-get -y --no-install-recommends install \
     software-properties-common \
     cmake \
+    cpp \
```
justsmth, line 26: I find it odd that you specifically need to install the C preprocessor (cpp)? Installing "build-essential" would eliminate a few of these lines.

```
❯ apt-cache depends build-essential                    
build-essential
 |Depends: libc6-dev
  Depends: <libc-dev>
    libc6-dev
  Depends: gcc
  Depends: g++
  Depends: make
    make-guile
  Depends: dpkg-dev
```

tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
```diff
@@ -47,9 +51,12 @@ RUN set -ex && \
     unzip awscliv2.zip && \
     ./aws/install --bin-dir /usr/bin && \
     rm -rf awscliv2.zip aws/ && \
-    # Download a copy of LLVM's libcxx which is required for building and running with Memory Sanitizer
     mkdir -p ${DEPENDENCIES_DIR} && \
     cd ${DEPENDENCIES_DIR} && \
+    # Extract and install Boost, dependency needed for mySQL.
+    wget https://boostorg.jfrog.io/artifactory/main/release/1.77.0/source/${BOOST_PACKAGE_NAME}.tar.bz2 && \
+    tar xfj ${BOOST_PACKAGE_NAME}.tar.bz2 && mv ./${BOOST_PACKAGE_NAME} ./boost && \
```
justsmth, line 58: * Since both the URL and variable `BOOST_PACKAGE_NAME` imply version number "1.77.0", I'm not sure that the variable provides much value. For visibility, perhaps the variable should contain the whole URL.
* Do you specifically need version 1.77.0? My Ubuntu (22.04) host has a package for Boost "1.74.0":
```
❯ apt-cache show libboost-all-dev | head -n 3
Package: libboost-all-dev
Architecture: amd64
Version: 1.74.0.3ubuntu7
```

tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
samuel40791765: Yeah great point. I had considered installing `build-essential` to install `cpp`, but I noticed that it installs gcc/g++ along with it (which is gcc-11). The extended `gcc-12` docker image specifically installs gcc-12, so I chose to directly install cpp so we wouldn't muddle two versions of gcc in the same container.
I don't think it's a big blocker though since we do specify `CC=gcc-12` in the environment. Let me know if we're OK with using `build-essential` instead.
* https://github.com/aws/aws-lc/blob/main/tests/ci/docker_images/linux-x86/ubuntu-22.04_gcc-12x/Dockerfile

tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
samuel40791765: * Sure, I took this from mySQL, but I think we can clean this up a bit.
* Yes, the mySQL build requires at least 1.77.0 which is why I had to install it from source. I should have clarified that in the comments, I'll add that.
```
CMake Warning at cmake/boost.cmake:305 (MESSAGE):
  Boost minor version found is 74 we need 77
```

justsmth (dismissed):

tests/ci/integration/run_mysql_integration.sh
```diff
@@ -0,0 +1,71 @@
...
+}
+
+function mysql_run_tests() {
+  pushd ${MYSQL_BUILD_FOLDER}
+  ninja test
+  popd
+}
+
+# Get latest MySQL version. MySQL often updates with large changes depending on OpenSSL all at once, so we pin to a specific version.
+git clone https://github.com/mysql/mysql-server.git ${MYSQL_SRC_FOLDER} -b ${MYSQL_VERSION_TAG} --depth 1
```
justsmth, line 62: We don't want our CI to fail unpredictably when there's a new MySql releases, but we also don't want to forget to update the tag periodically.

It would be nice if we could generate a metric/alarm when the tag we're using doesn't match the latest release.
```
LATEST_MYSQL_VERSION_TAG=mysql-`curl https://api.github.com/repos/mysql/mysql-server/tags | jq '.[].name' |grep '\-8.0' |sed -e 's/"mysql-cluster-\(.*\)"/\1/' |sort | tail -n 1`
```

justsmth (approved):

tests/ci/integration/run_mysql_integration.sh
samuel40791765: Great point, I'll see if there's something we can do to cut a ticket to us and remind us to do so. Thanks for the Tag code :).

samuel40791765: I updated the script to print us a warning. I briefly looked into it again, but Codebuild doesn't seem to have a "log tracking" mechanism in place. We had some similar discussion about creating an alarm here: https://github.com/aws/aws-lc/pull/1042#discussion_r1214951789. I'll update the SIM to make it more general.
I think we could consider creating a Github Action to poke us instead

tests/ci/docker_images/linux-x86/ubuntu-22.04_gcc-12x/Dockerfile
```diff
@@ -16,6 +16,9 @@ RUN set -ex && \
     rm -rf /var/lib/apt/lists/* && \
     rm -rf /tmp/*
 
+# Create alias to use cpp-12 preprocessor as default. MySQL relies on cpp being available.
+RUN update-alternatives --install /usr/bin/cpp cpp-bin /usr/bin/cpp-12 100
```
justsmth, line 20: Although build scripts should always use the `CC` and `CXX` environment variables, we should use `update-alternatives` to also setup `gcc-12` and `g++-12` for consistency with `cpp`. This is especially true since we're using this image to build third-party projects.

tests/ci/integration/run_mysql_integration.sh
andrewhop: In this it's actually a bit easier, we can just call cloudwatch put metric in this bash script, what's a little harder from 1042 is it's a string output from our test.

tests/ci/docker_images/linux-x86/ubuntu-22.04_gcc-12x/Dockerfile
samuel40791765: Updated 👍

tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
```diff
@@ -36,23 +39,28 @@ RUN set -ex && \
...
     # The awscli is used to publish data to CloudWatch Metrics in some jobs. This requires additional IAM permission
     curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
     unzip awscliv2.zip && \
     ./aws/install --bin-dir /usr/bin && \
     rm -rf awscliv2.zip aws/ && \
-    # Download a copy of LLVM's libcxx which is required for building and running with Memory Sanitizer
     mkdir -p ${DEPENDENCIES_DIR} && \
     cd ${DEPENDENCIES_DIR} && \
+    # Extract and install Boost 1.77.0. mySQL 8.33 depends on this specific version.
+    wget ${BOOST_SRC_URL} && tar xfj ${BOOST_TARBALL} && mv ./${BOOST_PACKAGE_NAME} ./boost && \
```
andrewhop, line 59:
```suggestion
    wget ${BOOST_SRC_URL} && tar xfj ${BOOST_TARBALL} && mv ./${BOOST_PACKAGE_NAME} ./boost && rm ${BOOST_TARBALL}  && \
```

tests/ci/integration/run_mysql_integration.sh
```diff
@@ -0,0 +1,79 @@
...
+#    |
+#    - mysql
+#    - AWS_LC_BUILD_FOLDER
+#    - AWS_LC_INSTALL_FOLDER
+#    - MYSQL_BUILD_FOLDER
+
+# Assumes script is executed from the root of aws-lc directory
+AWS_LC_DIR=$(pwd)
+cd ../
+ROOT=$(pwd)
```
andrewhop, line 30: These are already defined in common_posix.sh

tests/ci/integration/run_mysql_integration.sh
```diff
@@ -0,0 +1,79 @@
...
+  ninja -C "${AWS_LC_BUILD_FOLDER}" install
+  ls -R "${AWS_LC_INSTALL_FOLDER}"
+  rm -rf "${AWS_LC_BUILD_FOLDER:?}"/*
+}
+
+function mysql_patch_reminder() {
+  LATEST_MYSQL_VERSION_TAG=mysql-`curl https://api.github.com/repos/mysql/mysql-server/tags | jq '.[].name' |grep '\-8.0' |sed -e 's/"mysql-cluster-\(.*\)"/\1/' |sort | tail -n 1`
+  if [[ "${LATEST_MYSQL_VERSION_TAG}" != "${MYSQL_VERSION_TAG}" ]]; then
+    echo -e '\n\nMYSQL HAS RELEASED A NEW VERSION. REMEMBER TO UPDATE MYSQL_VERSION_TAG SOON.\n\n'
+  fi
```
andrewhop, line 53: I don't think we'll ever see this message.
```suggestion
  if [[ "${LATEST_MYSQL_VERSION_TAG}" != "${MYSQL_VERSION_TAG}" ]]; then
    aws cloudwatch put-metric-data --namespace AWS-LC --metric-name MySQLVersionMismatch --value 1
  else
      aws cloudwatch put-metric-data --namespace AWS-LC --metric-name MySQLVersionMismatch --value 0
  fi
```

tests/ci/codebuild/integration/mysql_integration.yml
```diff
@@ -0,0 +1,13 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+version: 0.2
+
+env:
+  variables:
+    GOPROXY: https://proxy.golang.org,direct
```
andrewhop, line 8: Is this the recommended way Sean fixed all the other GOPROXY issues?

tests/ci/integration/run_mysql_integration.sh
samuel40791765: Had to attach an IAM policy to get it working, so I made the change.

tests/ci/integration/run_mysql_integration.sh
samuel40791765: Only something similar to `AWS_LC_DIR` was defined, but I went ahead and added a new variable for `ROOT` too.

tests/ci/integration/run_mysql_integration.sh
```diff
@@ -0,0 +1,77 @@
...
+  rm -rf "${AWS_LC_BUILD_FOLDER:?}"/*
+}
+
+function mysql_patch_reminder() {
+  LATEST_MYSQL_VERSION_TAG=mysql-`curl https://api.github.com/repos/mysql/mysql-server/tags | jq '.[].name' |grep '\-8.0' |sed -e 's/"mysql-cluster-\(.*\)"/\1/' |sort | tail -n 1`
+  if [[ "${LATEST_MYSQL_VERSION_TAG}" != "${MYSQL_VERSION_TAG}" ]]; then
+    aws cloudwatch put-metric-data --namespace AWS-LC --metric-name MySQLVersionMismatch --value 1
+  else
+      aws cloudwatch put-metric-data --namespace AWS-LC --metric-name MySQLVersionMismatch --value 0
+  fi
```
justsmth, line 51: nit: indentation is off

tests/ci/integration/run_openssh_integration.sh
```diff
@@ -6,9 +6,9 @@ source tests/ci/common_posix_setup.sh
 
 # Set up environment.
 
-# ROOT
+# SYS_ROOT
 #  |
-#  - AWS_LC_DIR
+#  - SRC_ROOT
 #    |
 #    - aws-lc
```
justsmth, line 13: The documentation makes it look like "aws-lc" is a directory under `$SRC_ROOT`, but "aws-lc" is the `$SRC_ROOT`.

samuel40791765: Fixed :)

justsmth (approved):

andrewhop (approved):

PR #1055, opened 2023-06-10
Add SSL_get0_verified_chain 

andrewhop:
### Issues:
Resolves CryptoAlg-1876

### Description of changes: 
Add SSL_get0_verified_chain which returns the verified chain from an SSL. The built in X509 verifier is responsible for saving the reference to the verified chain if verification is successful.

### Call-outs:
Like other `get0` methods this does not increase the reference counter and the caller is not expected to free the chain when they are done with it. 

However, this change does use `X509_STORE_CTX_get1_chain` to get and own a reference to the chain so it persists after the verification is done. 

### Testing:
Add a new basic test that ensures only the chain needed for verification is returned. Still working on a full integration test.

### DIFF
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -1653,10 +1653,19 @@ OPENSSL_EXPORT STACK_OF(X509) *SSL_get_peer_cert_chain(const SSL *ssl);
 //
 // This is the same as |SSL_get_peer_cert_chain| except that this function
 // always returns the full chain, i.e. the first element of the return value
-// (if any) will be the leaf certificate. In constrast,
+// (if any) will be the leaf certificate. In contrast,
 // |SSL_get_peer_cert_chain| returns only the intermediate certificates if the
 // |ssl| is a server.
 OPENSSL_EXPORT STACK_OF(X509) *SSL_get_peer_full_cert_chain(const SSL *ssl);
+
+// SSL_get0_verified_chain returns the verified certificate chain of the peer
+// including the peer's end entity certificate. It must be called after a
+// session has been successfully established. If peer verification was not
+// successful (as indicated by |SSL_get_verify_result| not returning |X509_V_OK|)
+// the chain may be incomplete or invalid. If a verification callback was set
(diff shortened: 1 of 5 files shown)


samuel40791765 (dismissed):

samuel40791765 (approved):

include/openssl/ssl.h
```diff
@@ -1653,11 +1653,20 @@ OPENSSL_EXPORT STACK_OF(X509) *SSL_get_peer_cert_chain(const SSL *ssl);
...
-// (if any) will be the leaf certificate. In constrast,
+// (if any) will be the leaf certificate. In contrast,
 // |SSL_get_peer_cert_chain| returns only the intermediate certificates if the
 // |ssl| is a server.
 OPENSSL_EXPORT STACK_OF(X509) *SSL_get_peer_full_cert_chain(const SSL *ssl);
 
+// SSL_get0_verified_chain returns the verified certificate chain of the peer
+// including the peer's end entity certificate. It must be called after a
+// session has been successfully established. If peer verification was not
+// successful (as indicated by |SSL_get_verify_result| not returning |X509_V_OK|)
```
torben-hansen, line 1664: If this is essentially UB atm, why not just check `SSL_get_verify_result` before returning the chain, strengthen the function contract?

include/openssl/ssl.h
```diff
@@ -1653,11 +1653,20 @@ OPENSSL_EXPORT STACK_OF(X509) *SSL_get_peer_cert_chain(const SSL *ssl);
...
+// the chain may be incomplete or invalid. If a verification callback was set
+// with |SSL_CTX_set_cert_verify_callback| or |SSL_set_custom_verify|
+// |SSL_get0_verified_chain| may return null even if verification was successful.
```
torben-hansen, line 1667: I think you can just say that the result is undefined for this one?

include/openssl/ssl.h
andrewhop: Decided to make it explicit, and added a test for it.

samuel40791765 (approved):

torben-hansen (approved):

PR #1058, opened 2023-06-13
Change the Jitter entropy recovery mechanism

dkostic:
Previously, we were calling the |jent_entropy_read_safe| Jitter function to get entropy. This function has a built-in recover mechanism where if the source fails to produce entropy then the function changes some parameters and tries again. Due to FIPS rules, this is not possible any more. We change the call to |jent_entropy_read| instead and handle the potential failure in the AWS-LC part of the module by retrying three times (note that in this case the same entropy source parameters do not change).

### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 

### Call-outs:

### Testing:

### DIFF
--- a/crypto/fipsmodule/rand/rand.c
+++ b/crypto/fipsmodule/rand/rand.c
@@ -238,6 +238,13 @@ static void CRYPTO_get_fips_seed(uint8_t *out_entropy, size_t out_entropy_len,
                           out_entropy_len) == (ssize_t) out_entropy_len) {
         break;
     }
+    // If Jitter entropy failed to produce entropy we need to reset it.
+    jent_entropy_collector_free(state->jitter_ec);
+    state->jitter_ec = NULL;
+    jent_entropy_collector_alloc(0, JENT_FORCE_FIPS);
+    if (state->jitter_ec == NULL) {
+      abort();
+    }
   }
 
   if (num_tries == max_num_tries) {
--- a/crypto/fipsmodule/rand/rand.c
+++ b/crypto/fipsmodule/rand/rand.c
@@ -240,7 +240,7 @@ static void CRYPTO_get_fips_seed(uint8_t *out_entropy, size_t out_entropy_len,
     // If Jitter entropy failed to produce entropy we need to reset it.
     jent_entropy_collector_free(state->jitter_ec);
     state->jitter_ec = NULL;
-    jent_entropy_collector_alloc(0, JENT_FORCE_FIPS);
+    state->jitter_ec = jent_entropy_collector_alloc(0, JENT_FORCE_FIPS);
     if (state->jitter_ec == NULL) {
       abort();
     }
--- a/crypto/fipsmodule/rand/rand.c
+++ b/crypto/fipsmodule/rand/rand.c
@@ -217,7 +217,7 @@ static int rdrand(uint8_t *buf, size_t len) {
 #if defined(BORINGSSL_FIPS)
 
 static void CRYPTO_get_fips_seed(uint8_t *out_entropy, size_t out_entropy_len,
-                             int *out_want_additional_input) {
+                                 int *out_want_additional_input) {
   *out_want_additional_input = 0;
   // Every thread has its own Jitter instance so we fetch the one assigned
   // to the current thread.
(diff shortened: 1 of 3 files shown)


crypto/fipsmodule/rand/rand.c
```diff
@@ -227,9 +227,20 @@ static void CRYPTO_get_fips_seed(uint8_t *out_entropy, size_t out_entropy_len,
...
   }
 
-  // Generate the required number of bytes with Jitter.
-  if (jent_read_entropy_safe(&state->jitter_ec, (char *) out_entropy,
-                             out_entropy_len) != (ssize_t) out_entropy_len) {
+  // |jent_read_entropy| has a false positive health test failure rate of 2^-22.
+  // To avoid aborting so frequently, we retry 3 times.
+  size_t num_tries;
+  const size_t max_num_tries = 3;
+  for (num_tries = 0; num_tries < max_num_tries; num_tries++) {
```
torben-hansen, line 234: Can you make `CALL_RDRAND_WITH_RETRY` more general and use that?

crypto/fipsmodule/rand/rand.c
```diff
@@ -227,9 +227,20 @@ static void CRYPTO_get_fips_seed(uint8_t *out_entropy, size_t out_entropy_len,
     abort();
   }
 
-  // Generate the required number of bytes with Jitter.
-  if (jent_read_entropy_safe(&state->jitter_ec, (char *) out_entropy,
-                             out_entropy_len) != (ssize_t) out_entropy_len) {
+  // |jent_read_entropy| has a false positive health test failure rate of 2^-22.
+  // To avoid aborting so frequently, we retry 3 times.
+  size_t num_tries;
+  const size_t max_num_tries = 3;
```
torben-hansen, line 233: Should be global constant. Useful for testing as well (of which there is no coverage of this retries, is there(?))

crypto/fipsmodule/rand/rand.c
```diff
@@ -227,9 +227,20 @@ static void CRYPTO_get_fips_seed(uint8_t *out_entropy, size_t out_entropy_len,
...
+  for (num_tries = 0; num_tries < max_num_tries; num_tries++) {
+    // Try to generate the required number of bytes with Jitter.
+    // If successful break out from the loop, otherwise try again.
+    if (jent_read_entropy(state->jitter_ec, (char *) out_entropy,
+                          out_entropy_len) == (ssize_t) out_entropy_len) {
+        break;
+    }
+  }
+
+  if (num_tries == max_num_tries) {
```
torben-hansen, line 243: Regardless of what you decide to do with `CALL_RDRAND_WITH_RETRY`. This should be

```suggestion
  if (num_tries >= max_num_tries) {
```

crypto/fipsmodule/rand/rand.c
dkostic: I could have, but I was informed that we need to reset the Jitter object upon failure (see the latest update). With that, I don't see how `CALL_RDRAND_WITH_RETRY` can be used cleanly.

crypto/fipsmodule/rand/rand.c
dkostic: > Should be global constant

why is that?

> there is no coverage of this retries, is there(?)

nope, I'll add a test case.

crypto/fipsmodule/rand/rand.c
dkostic: I'll change it, but I don't see what's the difference :)

crypto/fipsmodule/rand/rand.c
dkostic: done.

samuel40791765 (dismissed):

crypto/fipsmodule/rand/rand.c
```diff
@@ -227,9 +227,26 @@ static void CRYPTO_get_fips_seed(uint8_t *out_entropy, size_t out_entropy_len,
...
+    // Try to generate the required number of bytes with Jitter.
+    // If successful break out from the loop, otherwise try again.
+    if (jent_read_entropy(state->jitter_ec, (char *) out_entropy,
+                          out_entropy_len) == (ssize_t) out_entropy_len) {
+        break;
+    }
+    // If Jitter entropy failed to produce entropy we need to reset it.
+    jent_entropy_collector_free(state->jitter_ec);
+    state->jitter_ec = NULL;
+    jent_entropy_collector_alloc(0, JENT_FORCE_FIPS);
```
torben-hansen, line 243: You'll want to assign the return value  to `state->jitter_ec`.

torben-hansen: (need some error path tests or try to test it manually?)

torben-hansen (approved):

samuel40791765 (approved):

PR #1059, opened 2023-06-14
Change the Jitter entropy recovery mechanism (#1058)

samuel40791765:
Previously, we were calling |jent_entropy_read_safe| Jitter function to get entropy. This function has a built-in recover mechanism where if the source fails to produce entropy then the function changes some parameters and tries again. Due to FIPS rules, this is not possible any more. We change the call to |jent_entropy_read| instead and handle the potential failure in the AWS-LC part of the module by retrying three times (note that in this case the same entropy source parameters do not change).

(cherry picked from commit 64d8a6b4b1f518173d07edd25a22a09f5d7e6c3d)


darylmartin100 (approved):

darylmartin100 (approved):

nebeid (approved):

skmcgrail (commented):
Looks good to me

PR #1060, opened 2023-06-14
Change the Jitter entropy recovery mechanism (#1058)

samuel40791765:
Previously, we were calling |jent_entropy_read_safe| Jitter function to get entropy. This function has a built-in recover mechanism where if the source fails to produce entropy then the function changes some parameters and tries again. Due to FIPS rules, this is not possible any more. We change the call to |jent_entropy_read| instead and handle the potential failure in the AWS-LC part of the module by retrying three times (note that in this case the same entropy source parameters do not change).

(cherry picked from commit 64d8a6b4b1f518173d07edd25a22a09f5d7e6c3d)

### DIFF
--- a/crypto/fipsmodule/rand/cpu_jitter_test.cc
+++ b/crypto/fipsmodule/rand/cpu_jitter_test.cc
@@ -57,6 +57,12 @@ TEST(CPUJitterEntropyTest, Basic) {
   EXPECT_NE(jitter_ec.instance, nullptr);
   EXPECT_EQ(jitter_ec.instance->osr, osr);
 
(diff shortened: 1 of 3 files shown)


billbo-yang (approved):

samuel40791765:
This is expected to fail the FV CI, since we don't have https://github.com/aws/aws-lc/commit/2240ea707a1861c26f9a098a4f76299be9417f44 pulled in to the FIPS branch. I'll confirm that it passes all the other tests before overriding the check and merging it in.

nebeid (approved):

PR #1062, opened 2023-06-15
align X509_VERIFY_PARAM host and email behavior with OpenSSL

samuel40791765:
### Issues:
Resolves `CryptoAlg-1865`

### Description of changes: 
A dig into some of the mySQL/AWS-LC test failures revealed that this was the root cause.
* https://github.com/mysql/mysql-server/blob/8.0/router/src/harness/src/tls_client_context.cc#L100
* https://github.com/mysql/mysql-server/blob/8.0/vio/viosslfactories.cc#L709

BoringSSL briefly made this backwards compatible, but enforced the stricter check a week later since Chrome was not effected by this.
* https://github.com/aws/aws-lc/commit/e759a9cd84198613199259dbed401f4951747cff
* https://github.com/aws/aws-lc/commit/1902d818ac4fef9497dfe5d0ce6f2c99f585bdff

This doesn't apply to us however, so we're reverting the strict check in an attempt to be more compatible with OpenSSL.

### Call-outs:
I've reverted the check for `X509_VERIFY_PARAM_set1_email` while I was at it as well. `X509_VERIFY_PARAM_set1_ip` was using the same underlying logic as `X509_VERIFY_PARAM_set1_email` which made things a bit complicated.
`X509_VERIFY_PARAM_set1_ip` takes an `unsigned char *`, but changes to make it more compatible with OpenSSL required `strlen` to detect the length of the string. I've separated the logic for  `X509_VERIFY_PARAM_set1_email/ip` into two functions to get around this.
`X509_VERIFY_PARAM_set1_ip` now maintains the original BoringSSL behavior, while `X509_VERIFY_PARAM_set1_email` is compatible with OpenSSL now. We can look into making `X509_VERIFY_PARAM_set1_ip` more compatible with OpenSSL later on if needed.

### Testing:
CI

### DIFF
--- a/crypto/x509/x509_vpm.c
+++ b/crypto/x509/x509_vpm.c
@@ -80,9 +80,11 @@ static int int_x509_param_set_hosts(X509_VERIFY_PARAM *param, int mode,
                                     const char *name, size_t namelen) {
   char *copy;
 
-  if (name == NULL || namelen == 0) {
-    // Unlike OpenSSL, we reject trying to set or add an empty name.
-    return 0;
+  // Setting 0 to automatically detect the length of |name| is an OpenSSL quirk
+  // that AWS-LC isn't keen on supporting. However, consumers often assume
+  // OpenSSL semantics from AWS-LC, so it's supported in this case.
+  if (name != NULL && namelen == 0) {
+    namelen = strlen(name);
   }
 
   // Refuse names with embedded NUL bytes.
@@ -94,6 +96,12 @@ static int int_x509_param_set_hosts(X509_VERIFY_PARAM *param, int mode,
   if (mode == SET_HOST && param->hosts) {
     string_stack_free(param->hosts);
     param->hosts = NULL;
+  }
+  // OpenSSL returns 1 when trying to set or add an empty name. This is also a
+  // quirk that AWS-LC isn't keen on supporting, but we maintain for backwards
+  // compatibility.
+  if (name == NULL || namelen == 0) {
+    return 1;
   }
 
   copy = OPENSSL_strndup(name, namelen);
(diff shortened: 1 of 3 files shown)


crypto/x509/x509_vpm.c
```diff
@@ -80,9 +80,11 @@ static int int_x509_param_set_hosts(X509_VERIFY_PARAM *param, int mode,
                                     const char *name, size_t namelen) {
   char *copy;
 
-  if (name == NULL || namelen == 0) {
-    // Unlike OpenSSL, we reject trying to set or add an empty name.
-    return 0;
+  // Setting 0 to automatically detect the length of |name| is an OpenSSL quirk
+  // that AWS-LC isn't keen on supporting. However, consumers often assume
+  // OpenSSL semantics from AWS-LC, so it's supported in this case.
+  if (name != NULL && namelen == 0) {
```
andrewhop, line 86: Are we missing a return 0 if name == null?

crypto/x509/x509_vpm.c
```diff
@@ -313,8 +321,50 @@ int X509_VERIFY_PARAM_set1(X509_VERIFY_PARAM *to,
...
+  return 1;
+}
+
+// IP addresses work slightly differently, so we use another function to
+// differentiate from emails. |X509_VERIFY_PARAM_set1_ip| takes a const
+// unsigned char*, instead of a const char*, so the same strlen logic that was
+// being used is not quite suitable here.
+// We keep the original behavior that BoringSSL left, but only for IP addresses.
+// We can align the behavior with |int_x509_param_set1_email| like OpenSSL has
+// been doing if needed.
```
andrewhop, line 365: How hard is this to do now? This will lead to a confusing experience with different functions treating 0 in different ways.

crypto/x509/x509_vpm.c
```diff
@@ -313,8 +321,50 @@ int X509_VERIFY_PARAM_set1(X509_VERIFY_PARAM *to,
...
+                                     const char *src, size_t srclen) {
+  void *tmp;
+  if (src != NULL) {
+    // Setting |srclen| to 0 to automatically detect the length of |src| is an
+    // OpenSSL quirk that AWS-LC isn't keen on supporting. However, consumers
+    // often assume OpenSSL semantics from AWS-LC, so it's supported in this
+    // case.
+    if (srclen == 0) {
+      srclen = strlen(src);
+    }
```
andrewhop, line 334: Should this have an else that ensures there are no internal \0 like int_x509_param_set_hosts? I see the original code in int_x509_param_set1 didn't enforce this for ip address or emails, any idea why they are different?

crypto/x509/x509_vpm.c
samuel40791765: `X509_VERIFY_PARAM_set1_ip` takes a `const unsigned char*` with the `ip` parameter, instead of a `const char*`. OpenSSL then attempts to transform the `unsigned char` pointer into an `const char` and passes it into `int_x509_param_set1`. `int_x509_param_set1` calls `strlen` on the pointer, but this results in a fatal failure if the  `const unsigned char*` pointer is non-NULL terminated.

* https://github.com/openssl/openssl/blob/7197abddb891933f52ec84dafb41b685d4a1d122/crypto/x509/x509_vpm.c#L466-L467
* https://github.com/openssl/openssl/blob/7197abddb891933f52ec84dafb41b685d4a1d122/crypto/x509/x509_vpm.c#L243

I chose to keep the original behavior since it's effectively getting around the weird behavior OpenSSL was doing. Consumers of `X509_VERIFY_PARAM_set1_ip` were likely defining the length since using 0 would often result in fatal failures, unless they were manually appending `\0` at the end of the IP string.
We could do this now, but it would easily expose the obvious fatal failures when misusing `X509_VERIFY_PARAM_set1_ip`.

crypto/x509/x509_vpm.c
samuel40791765: IP addresses are allowed to have NULL bytes due to their nature: https://github.com/aws/aws-lc/blob/8402a6eecfff195793a57b5bf1cf371f1218da3c/crypto/x509/x509_test.cc#L1790-L1791

We could look into enforcing this for emails if we wanted to? `int_x509_param_set_hosts` was a different function since it deals with a Stack instead of a single string like the others.

samuel40791765: Emails already have this enforced in `X509_VERIFY_PARAM_set1_email`. We also have a test for it.
* https://github.com/aws/aws-lc/blob/15bd84a7a63b44c7a94965a13a9df74b1db25483/crypto/x509/x509_vpm.c#L457
* https://github.com/aws/aws-lc/blob/main/crypto/x509/x509_test.cc#L1749-L1755

skmcgrail (approved):

andrewhop (approved):

PR #1067, opened 2023-06-22
Fix potential NULL pointer dereference

dkostic:
### Issues:
Resolves V935691760

### Description of changes: 
Fix potential NULL pointer dereference in two places.

### Call-outs:

### Testing:

### DIFF
--- a/crypto/x509/name_print.c
+++ b/crypto/x509/name_print.c
@@ -178,6 +178,10 @@ static int do_name_ex(BIO *out, const X509_NAME *n, int indent,
     } else {
       objbuf = OBJ_nid2sn(fn_nid);
     }
+    if (objbuf == NULL) {
+      return -1;
+    }
+
     int objlen = strlen(objbuf);
     if (!maybe_write(out, objbuf, objlen) ||
         !maybe_write(out, sep_eq, sep_eq_len)) {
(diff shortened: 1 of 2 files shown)


torben-hansen (approved):

crypto/x509/x509_req.c
```diff
@@ -185,6 +185,9 @@ int X509_REQ_get_attr_by_OBJ(const X509_REQ *req, const ASN1_OBJECT *obj,
   for (; lastpos < n; lastpos++) {
     const X509_ATTRIBUTE *attr =
         sk_X509_ATTRIBUTE_value(req->req_info->attributes, lastpos);
+    if (attr == NULL) {
```
nebeid, line 188: I'm not sure if this can happen, since n was set the number of attributes, if I understood correctly. There is a similar code here https://github.com/aws/aws-lc/blob/8402a6eecfff195793a57b5bf1cf371f1218da3c/crypto/x509/t_req.c#L153-L155

crypto/x509/name_print.c
```diff
@@ -178,6 +178,10 @@ static int do_name_ex(BIO *out, const X509_NAME *n, int indent,
     } else {
```
nebeid, line 178: From the documentation of `OBJ_obj2nid()` called on l. 174, https://github.com/aws/aws-lc/blob/cfce39dccb49bc61252784c5270c928a51ecace6/include/openssl/obj.h#L110-L113, the `else` branch should be for a known nid. If we prefer to be safe still, maybe the check for NULL would be inside the `else` branch.

crypto/x509/x509_req.c
dkostic: yeah, I had the same thought but I'm not 100% certain it can't happen. The additional check is very cheap anyway, so no harm in having it.

crypto/x509/name_print.c
dkostic: I think it's safer to have the `NULL` check right before calling `strlen` unconditionally?

nebeid (approved):

PR #1077, opened 2023-06-27
Align OCSP behavior with OpenSSL for nginx

samuel40791765:
### Issues:
Resolves `CryptoAlg-1849`

### Description of changes: 
Asides from the behavior fixed in https://github.com/aws/aws-lc/pull/1074, there was some slight OCSP errors when building with nginx's OCSP stapling tests. There are still a couple of tests failing due to us missing multiple certs support, but this should resolve all other issues with us failing the test.
* https://github.com/nginx/nginx-tests/blob/master/ssl_stapling.t

1. The NULL check in `ocsp_find_signer_sk` static function was a bit too pedantic, so I relaxed it. It's possible for the OCSP Basic response to be missing a cert stack, so we shouldn't be throwing an error on the stack if that happens.
2. Implemented functionality for the two OCSP flags, `OCSP_NOVERIFY` and `OCSP_TRUSTOTHER`

### Call-outs:
N/A

### Testing:
N/A

### DIFF
--- a/crypto/ocsp/ocsp_verify.c
+++ b/crypto/ocsp/ocsp_verify.c
@@ -9,7 +9,6 @@
 // https://datatracker.ietf.org/doc/html/rfc6960#section-4.2.2.3
 static X509 *ocsp_find_signer_sk(STACK_OF(X509) *certs, OCSP_RESPID *id) {
   if (certs == NULL || id == NULL) {
-    OPENSSL_PUT_ERROR(OCSP, ERR_R_PASSED_NULL_PARAMETER);
     return NULL;
   }
 
@@ -54,7 +53,7 @@ static int ocsp_find_signer(X509 **psigner, OCSP_BASICRESP *bs,
   signer = ocsp_find_signer_sk(certs, rid);
   if (signer != NULL) {
     *psigner = signer;
-    return 1;
+    return 2;
   }
 
   // look in certs stack the responder may have included in |OCSP_BASICRESP|,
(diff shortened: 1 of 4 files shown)


crypto/ocsp/ocsp_verify.c
```diff
@@ -54,7 +53,7 @@ static int ocsp_find_signer(X509 **psigner, OCSP_BASICRESP *bs,
   signer = ocsp_find_signer_sk(certs, rid);
   if (signer != NULL) {
     *psigner = signer;
-    return 1;
+    return 2;
```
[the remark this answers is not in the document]
samuel40791765: Good point. I took this directly from OpenSSL, so I'll refine it.

skmcgrail (dismissed):

include/openssl/ocsp.h
```diff
@@ -27,26 +27,25 @@ extern "C" {
...
 // OCSP_NOEXPLICIT.
 #define OCSP_NOEXPLICIT 0x20
-// OCSP_TRUSTOTHER is for |OCSP_basic_verify|. This is a no-op flag in AWS-LC.
-// When setting this flag in OpenSSL, if the reponse signer's cert is one of
-// those in the |certs| stack then it is implicitly trusted.
-#define OCSP_TRUSTOTHER 0
+// OCSP_TRUSTOTHER is for |OCSP_basic_verify|. When setting this flag in OpenSSL,
+// if the reponse signer's cert is one of those in the |certs| stack then it is
+// implicitly trusted.
+#define OCSP_TRUSTOTHER 0x200
```
andrewhop, line 48:
```suggestion
// OCSP_TRUSTOTHER is for |OCSP_basic_verify|. When this flag is set
// if the reponse signer's cert is one of those in the |certs| stack then it is
// implicitly trusted.
#define OCSP_TRUSTOTHER 0x200
```

This isn't just for OpenSSL right? We have this behavior now

include/openssl/ocsp.h
```diff
@@ -27,26 +27,25 @@ extern "C" {
...
 #define OCSP_NOCHAIN 0x8
-// OCSP_NOVERIFY is for |OCSP_basic_verify|. This is a no-op flag in AWS-LC.
-// When setting this flag in OpenSSL, the |OCSP_BASICRESP|'s signature will
-// still be verified, but setting this flag skips verifying the signer's
-// certificate.
-#define OCSP_NOVERIFY 0
+// OCSP_NOVERIFY is for |OCSP_basic_verify|. When setting this flag, the
+// |OCSP_BASICRESP|'s signature will still be verified, but setting this flag
+// skips additionally verifying the signer's certificate.
+#define OCSP_NOVERIFY 0x10
```
andrewhop, line 40:
```suggestion
// OCSP_NOVERIFY is for |OCSP_basic_verify|. When setting this flag, the
// |OCSP_BASICRESP|'s signature will still be verified, but
// skips additionally verifying the signer's certificate.
#define OCSP_NOVERIFY 0x10
```

include/openssl/ocsp.h
```diff
@@ -27,26 +27,25 @@ extern "C" {
 #define OCSP_NOCERTS 0x1
 // OCSP_NOINTERN is for |OCSP_basic_verify|. Searches for certificates the
 // responder may have included in |bs| will be done, unless the flags contain
-// OCSP_NOINTERN.
+// |OCSP_NOINTERN|.
```
andrewhop, line 30: NP: can you clean up this comment to match the ones below.

andrewhop (commented):
I know this behavior is covered by Nginx's tests but can you add some targeted tests in our crypto_test.

andrewhop (approved):

skmcgrail (approved):

PR #1086, opened 2023-07-07
add back CERT_PKEY structure and refactor pointers

samuel40791765:
### Issues:
Addresses `CryptoAlg-1844`

### Description of changes: 
This doesn't actually implement multiple cert functionality, but sets up the structures to do so. All certs currently use the first index (`SSL_PKEY_RSA`) during the SSL connection. 
A subsequent PR will be introduced to implement usage of the new certificate slots.

### Call-outs:
Overall functionality of the code shouldn't be effected right now, this is just letting the code point to new pointers.

### Testing:
Original tests

### DIFF
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -1361,7 +1361,8 @@ struct ssl_private_key_method_st {
 };
 
 // SSL_set_private_key_method configures a custom private key on |ssl|.
-// |key_method| must remain valid for the lifetime of |ssl|.
+// |key_method| must remain valid for the lifetime of |ssl|. Using custom
+// keys with the multiple certificate slots feature is not supported.
 OPENSSL_EXPORT void SSL_set_private_key_method(
     SSL *ssl, const SSL_PRIVATE_KEY_METHOD *key_method);
 
--- a/ssl/internal.h
+++ b/ssl/internal.h
@@ -2425,12 +2425,14 @@ bool tls12_check_peer_sigalg(const SSL_HANDSHAKE *hs, uint8_t *out_alert,
 // From RFC 4492, used in encoding the curve type in ECParameters
 #define NAMED_CURVE_TYPE 3
 
-struct CERT {
-  static constexpr bool kAllowUniquePtr = true;
+// SSL_PKEY_* denote certificate types. These represent an index within
+// |CERT->cert_privatekeys|.
+#define SSL_PKEY_RSA 0
+#define SSL_PKEY_ECC 1
+#define SSL_PKEY_ED25519 2
+#define SSL_PKEY_NUM 3
 
-  explicit CERT(const SSL_X509_METHOD *x509_method);
-  ~CERT();
-
+struct CERT_PKEY {
   UniquePtr<EVP_PKEY> privatekey;
 
   // chain contains the certificate chain, with the leaf at the beginning. The
@@ -2450,6 +2452,22 @@ struct CERT {
   // is only used as a cache in order to implement “get0” functions that return
   // a non-owning pointer to the certificate chain.
   X509 *x509_leaf = nullptr;
+};
+
+struct CERT {
+  static constexpr bool kAllowUniquePtr = true;
+
+  explicit CERT(const SSL_X509_METHOD *x509_method);
+  ~CERT();
+
+  // cert_privatekey_idx ALWAYS points to an element of the |cert_pkeys|
+  // array. OpenSSL implements this as a pointer, but an index is more
+  // efficient.
+  int cert_privatekey_idx = SSL_PKEY_RSA;
+
+  Array<CERT_PKEY> cert_privatekeys;
+
+  /// We'lll see what we want to do about the |x509_stash| below later.
 
   // x509_stash contains the last |X509| object append to the chain. This is a
   // workaround for some third-party code that continue to use an |X509| object
--- a/ssl/internal.h
+++ b/ssl/internal.h
@@ -2465,9 +2465,9 @@ struct CERT {
   // cert_privatekey_idx ALWAYS points to an element of the |cert_pkeys|
   // array. OpenSSL implements this as a pointer, but an index is more
   // efficient.
-  int cert_privatekey_idx = -1;
+  int cert_private_key_idx = -1;
 
-  Array<CERT_PKEY> cert_privatekeys;
+  Array<CERT_PKEY> cert_private_keys;
 
   /// We'lll see what we want to do about the |x509_stash| below later.
 
@@ -3236,6 +3236,11 @@ bool ssl_is_key_type_supported(int key_type);
 bool ssl_compare_public_and_private_key(const EVP_PKEY *pubkey,
                                         const EVP_PKEY *privkey);
 bool ssl_cert_check_private_key(const CERT *cert, const EVP_PKEY *privkey);
+
+// ssl_cert_check_cert_private_keys_usage returns true if |cert_private_keys|
+// in |cert| has a valid index and a sufficient amount of slots.
+bool ssl_cert_check_cert_private_keys_usage(const CERT *cert);
+
 bool ssl_get_new_session(SSL_HANDSHAKE *hs);
 bool ssl_encrypt_ticket(SSL_HANDSHAKE *hs, CBB *out,
                         const SSL_SESSION *session);
--- a/ssl/internal.h
+++ b/ssl/internal.h
@@ -2430,7 +2430,7 @@ bool tls12_check_peer_sigalg(const SSL_HANDSHAKE *hs, uint8_t *out_alert,
 #define SSL_PKEY_RSA 0
 #define SSL_PKEY_ECC 1
 #define SSL_PKEY_ED25519 2
-#define SSL_PKEY_NUM 3
+#define SSL_PKEY_SIZE 3
 
 struct CERT_PKEY {
   UniquePtr<EVP_PKEY> privatekey;
@@ -2448,9 +2448,11 @@ struct CERT_PKEY {
   // pointer to the certificate chain.
   STACK_OF(X509) *x509_chain = nullptr;
 
-  // x509_leaf may contain a parsed copy of the first element of |chain|. This
-  // is only used as a cache in order to implement “get0” functions that return
-  // a non-owning pointer to the certificate chain.
+  // x509_leaf retains the |X509| structure of the first element of |chain|.
+  // However, if certs are set with |SSL_CTX_use_certificate_ASN1| or
+  // |SSL_use_certificate_ASN1|, this is only used as a cache in order to
+  // implement “get0” functions that return a non-owning pointer to the
+  // certificate chain.
   X509 *x509_leaf = nullptr;
 };
 
@@ -2463,7 +2465,7 @@ struct CERT {
   // cert_privatekey_idx ALWAYS points to an element of the |cert_pkeys|
   // array. OpenSSL implements this as a pointer, but an index is more
   // efficient.
-  int cert_privatekey_idx = SSL_PKEY_RSA;
+  int cert_privatekey_idx = -1;
 
   Array<CERT_PKEY> cert_privatekeys;
 
--- a/ssl/ssl_cert.cc
+++ b/ssl/ssl_cert.cc
@@ -160,7 +160,7 @@ UniquePtr<CERT> ssl_cert_dup(CERT *cert) {
   }
 
   ret->cert_private_key_idx = cert->cert_private_key_idx;
-  if (!ssl_cert_check_cert_private_keys_usage(cert) &&
+  if (!ssl_cert_check_cert_private_keys_usage(cert) ||
       !ssl_cert_check_cert_private_keys_usage(ret.get())) {
     return nullptr;
   }
--- a/ssl/ssl_cert.cc
+++ b/ssl/ssl_cert.cc
@@ -136,8 +136,8 @@ BSSL_NAMESPACE_BEGIN
 
 CERT::CERT(const SSL_X509_METHOD *x509_method_arg)
     : x509_method(x509_method_arg) {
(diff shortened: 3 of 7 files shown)


include/openssl/ssl.h
```diff
@@ -1361,7 +1361,8 @@ struct ssl_private_key_method_st {
 };
 
 // SSL_set_private_key_method configures a custom private key on |ssl|.
-// |key_method| must remain valid for the lifetime of |ssl|.
+// |key_method| must remain valid for the lifetime of |ssl|. Using custom
+// keys with the multiple certificate slots feature is not supported.
```
samuel40791765, line 1365: This comment was also backported from the original commit: https://github.com/google/boringssl/commit/d1d8078025cc654d60ee873c88740f8a987a77b7
It's a BoringSSL specific function that doesn't exist in OpenSSL, so adding multiple cert slots support for this function shouldn't be a priority for now.

ssl/internal.h
```diff
@@ -2450,6 +2452,22 @@ struct CERT {
...
+struct CERT {
+  static constexpr bool kAllowUniquePtr = true;
+
+  explicit CERT(const SSL_X509_METHOD *x509_method);
+  ~CERT();
+
+  // cert_privatekey_idx ALWAYS points to an element of the |cert_pkeys|
+  // array. OpenSSL implements this as a pointer, but an index is more
+  // efficient.
+  int cert_privatekey_idx = SSL_PKEY_RSA;
```
skmcgrail, line 2466: Does OpenSSL do this for purposes of tracking whether this got initialized to the correct value? E.g. they can check that if this is `nullptr` to see if it wasn't correctly initialized, but here we are default to always pointing at the first index.

ssl/ssl_cert.cc
```diff
@@ -300,49 +310,55 @@ static int cert_set_chain_and_key(
     }
   }
 
-  cert->privatekey = UpRef(privkey);
+  int idx = cert->cert_privatekey_idx;
```
skmcgrail, line 313: What sets this index value to the correct position?

ssl/internal.h
samuel40791765: Don't think there's a particular purpose, I applied the idea since their documentation called it out as a potential improvement. It would be worth doing this right from the start, now that we're reimplementing it.
*  https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/ssl/ssl_local.h#L1879-L1885.

In the original, `cert->key` value is assigned the first index during initialization. There's an expectation that the pointer always exists, so it isn't checked (based on the logic within the commit).
* https://github.com/google/boringssl/commit/d1d8078025cc654d60ee873c88740f8a987a77b7#diff-ea838e3fccacc2f14349aefa3d01ba807d2a2cb218878a93cd0fa227215558a4L151-L161

ssl/ssl_cert.cc
samuel40791765: Great question, that's part of the broader functionalities being introduced in the next PR (when we actually allow different cert types to use different slots). As of now, `cert_privatekey_idx` can only be `SSL_PKEY_RSA` and is never changed.
Also please let me know if we should come up with a better variable name other than `cert_privatekey_idx` :)

skmcgrail (approved):

ssl/internal.h
```diff
@@ -2425,12 +2425,14 @@ bool tls12_check_peer_sigalg(const SSL_HANDSHAKE *hs, uint8_t *out_alert,
...
-
-  explicit CERT(const SSL_X509_METHOD *x509_method);
-  ~CERT();
-
+// SSL_PKEY_* denote certificate types. These represent an index within
+// |CERT->cert_privatekeys|.
+#define SSL_PKEY_RSA 0
+#define SSL_PKEY_ECC 1
+#define SSL_PKEY_ED25519 2
+#define SSL_PKEY_NUM 3
```
justsmth, line 2433: Should this be called `SSL_PKEY_COUNT` or `SSL_PKEY_SIZE`? It's only used to indicate the size of the array.

ssl/ssl_cert.cc
```diff
@@ -135,7 +135,9 @@
 BSSL_NAMESPACE_BEGIN
 
 CERT::CERT(const SSL_X509_METHOD *x509_method_arg)
-    : x509_method(x509_method_arg) {}
+    : x509_method(x509_method_arg) {
+  this->cert_privatekeys.Init(SSL_PKEY_NUM);
+}
```
justsmth, line 140: There's no check on the return value here. The call to `Init` could potentially fail, invalidating logic that assumes the size of always `SSL_PKEY_NUM`.

ssl/ssl_cert.cc
```diff
@@ -153,20 +155,26 @@ UniquePtr<CERT> ssl_cert_dup(CERT *cert) {
...
   }
 
-  if (cert->chain) {
-    ret->chain.reset(sk_CRYPTO_BUFFER_deep_copy(
-        cert->chain.get(), buffer_up_ref, CRYPTO_BUFFER_free));
-    if (!ret->chain) {
-      return nullptr;
+  for (int i = 0; i < SSL_PKEY_NUM; i++) {
+    CERT_PKEY &cert_pkey = cert->cert_privatekeys[i];
+    CERT_PKEY &ret_pkey = ret->cert_privatekeys[i];
```
justsmth, line 160: There should be a check on the `size()` prior to iterating -- either `cert`  or `ret` might not have properly allocated arrays.

ssl/internal.h
justsmth, line 2466: Should this instead be initialized to `-1` (analogous to a `NULL` pointer) to indicate that there's not yet a confirmed "cert_privatekey" available for reference?

ssl/ssl_cert.cc
```diff
@@ -153,20 +155,26 @@ UniquePtr<CERT> ssl_cert_dup(CERT *cert) {
...
     }
-  }
 
-  if (cert->x509_leaf != nullptr) {
-    X509_up_ref(cert->x509_leaf);
-    ret->x509_leaf = cert->x509_leaf;
+    if (cert_pkey.x509_leaf != nullptr) {
+      X509_up_ref(cert_pkey.x509_leaf);
+      ret_pkey.x509_leaf = cert_pkey.x509_leaf;
+    }
```
justsmth, line 173: Do you also need to copy the `x509_chain`?

ssl/ssl_cert.cc
```diff
@@ -205,8 +213,10 @@ void ssl_cert_clear_certs(CERT *cert) {
 
   cert->x509_method->cert_clear(cert);
 
-  cert->chain.reset();
-  cert->privatekey.reset();
+  for (int i = 0; i < SSL_PKEY_NUM; i++) {
+    cert->cert_privatekeys[i].chain.reset();
+    cert->cert_privatekeys[i].privatekey.reset();
+  }
```
justsmth, line 219: Do `x509_leaf` or `x509_chain` need to be cleared, or "down ref'd"?

justsmth (changes requested):

ssl/ssl_cert.cc
samuel40791765: Great question, `cert->x509_method->cert_clear(cert);` above actually does that for us. 
`cert_clear` is a pointer function to `ssl_crypto_x509_cert_clear`, which clears the X509 related structures within it.
https://github.com/aws/aws-lc/blob/ad53db8f1c1753799a13c39d53a9410fe7a49b2b/ssl/ssl_x509.cc#L262-L268

ssl/ssl_cert.cc
```diff
@@ -300,49 +319,55 @@ static int cert_set_chain_and_key(
...
 
 bool ssl_has_certificate(const SSL_HANDSHAKE *hs) {
-  return hs->config->cert->chain != nullptr &&
-         sk_CRYPTO_BUFFER_value(hs->config->cert->chain.get(), 0) != nullptr &&
+  CERT_PKEY &cert_pkey =
+      hs->config->cert->cert_privatekeys[hs->config->cert->cert_privatekey_idx];
+  return cert_pkey.chain != nullptr &&
+         sk_CRYPTO_BUFFER_value(cert_pkey.chain.get(), 0) != nullptr &&
          ssl_has_private_key(hs);
 }
```
justsmth, line 372: My impression is that almost everywhere that `cert_privatekeys` is accessed by an index needs to be guarded by a  `if(cert->cert_privatekeys.size() < 1) { /* error handling and return */ }`
```
bool ssl_has_certificate(const SSL_HANDSHAKE *hs) {
  CERT* cert = hs->config->cert.get();
  if(cert->cert_privatekeys.size() < 1) {
      return false;
  }
  CERT_PKEY &cert_pkey =
      cert->cert_privatekeys[cert->cert_privatekey_idx];
  return cert_pkey.chain != nullptr &&
         sk_CRYPTO_BUFFER_value(cert_pkey.chain.get(), 0) != nullptr &&
         ssl_has_private_key(hs);
}
```

ssl/ssl_cert.cc
```diff
@@ -153,20 +159,31 @@ UniquePtr<CERT> ssl_cert_dup(CERT *cert) {
...
   }
 
-  if (cert->chain) {
-    ret->chain.reset(sk_CRYPTO_BUFFER_deep_copy(
-        cert->chain.get(), buffer_up_ref, CRYPTO_BUFFER_free));
-    if (!ret->chain) {
-      return nullptr;
-    }
+  ret->cert_private_key_idx = cert->cert_private_key_idx;
+  if (!ssl_cert_check_cert_private_keys_usage(cert) &&
```
justsmth, line 163: I think this should be an "or".

samuel40791765: My bad, good catch. I double checked all usages of `ssl_cert_check_cert_private_keys_usage` again to verify.

justsmth (approved):

justsmth (dismissed):
A few more "nits", none of which are blocking.

ssl/internal.h
```diff
@@ -2446,10 +2448,28 @@ struct CERT {
   // pointer to the certificate chain.
   STACK_OF(X509) *x509_chain = nullptr;
```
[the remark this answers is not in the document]
samuel40791765: I'm not so clear on why, probably just never got to refactoring it since `chain` is the main certificate chain used for verification. The same question also applies to `x509_leaf`.
We could consider doing a larger refactor on this later.

skmcgrail (dismissed):

ssl/ssl_cert.cc
```diff
@@ -300,49 +323,67 @@ static int cert_set_chain_and_key(
     }
   }
 
-  cert->privatekey = UpRef(privkey);
+  if (!ssl_cert_check_cert_private_keys_usage(cert)) {
+    return 0;
+  }
```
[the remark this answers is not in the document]
samuel40791765: I considered that, but there are multiple error cases above that could be thrown onto the stack before this is actually used. I didn't want to break this behavior and the current placement still makes the most sense to me.

ssl/ssl_cert.cc
```diff
@@ -522,15 +563,16 @@ bool ssl_cert_check_private_key(const CERT *cert, const EVP_PKEY *privkey) {
     return false;
   }
 
-  if (cert->chain == nullptr ||
-      sk_CRYPTO_BUFFER_value(cert->chain.get(), 0) == nullptr) {
+  STACK_OF(CRYPTO_BUFFER) *chain =
+      cert->cert_private_keys[cert->cert_private_key_idx].chain.get();
+
+  if (chain == nullptr || sk_CRYPTO_BUFFER_value(chain, 0) == nullptr) {
```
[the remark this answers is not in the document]
samuel40791765: It's not. `ssl_set_pkey` calls `ssl_cert_check_private_key` with a custom private key.
https://github.com/aws/aws-lc/blob/769995e9afbb0d576c247ac02d66ef60a5ece81a/ssl/ssl_privkey.cc#L79-L88

ssl/ssl_lib.cc
```diff
@@ -1757,16 +1757,25 @@ int SSL_has_pending(const SSL *ssl) {
...
+  }
+  return ssl_cert_check_private_key(
+      ctx->cert.get(),
+      ctx->cert->cert_private_keys[ctx->cert->cert_private_key_idx]
+          .privatekey.get());
 }
 
 int SSL_check_private_key(const SSL *ssl) {
-  if (!ssl->config) {
+  if (!ssl->config ||
```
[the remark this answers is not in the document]
samuel40791765: Yeah I agree. I've noticed it in multiple places, but I didn't want to touch code irrelevant to this PR. I've been trying to make my implementation checks consistent with the former for now.

ssl/ssl_x509.cc
```diff
@@ -966,40 +1003,47 @@ int SSL_clear_chain_certs(SSL *ssl) {
...
-  UniquePtr<STACK_OF(X509)> chain(sk_X509_new_null());
-  if (!chain) {
+  UniquePtr<STACK_OF(X509)> new_chain(sk_X509_new_null());
+  if (!new_chain) {
     return 0;
   }
 
-  for (size_t i = 1; i < sk_CRYPTO_BUFFER_num(cert->chain.get()); i++) {
-    CRYPTO_BUFFER *buffer = sk_CRYPTO_BUFFER_value(cert->chain.get(), i);
+  for (size_t i = 1; i < sk_CRYPTO_BUFFER_num(chain.get()); i++) {
```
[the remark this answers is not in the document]
samuel40791765: I don't think that's the reason, it's the expected behavior of the fields.
It's already documented here:
https://github.com/aws/aws-lc/blob/769995e9afbb0d576c247ac02d66ef60a5ece81a/ssl/internal.h#L2436-L2447

justsmth (approved):

skmcgrail (approved):

PR #1093, opened 2023-07-12
Add BoringSSL Dispatch Test for aarch64

billbo-yang:
### Issues:
Addresses CryptoAlg-1873 by unblocking https://github.com/aws/aws-lc/pull/1045

### Description of changes: 
Currently AWS-LC only checks whether the correct hardware code is being reached on x86 and x86_64 and not aarch64. This PR fixes that so that we can ensure that the correct code path is taken whenever change OPENSSL_armcap on aarch64 systems.

### Call-outs:
N/A

### Testing:
Tests pass locally

### DIFF
--- a/crypto/fipsmodule/aes/asm/aesv8-armx.pl
+++ b/crypto/fipsmodule/aes/asm/aesv8-armx.pl
@@ -102,7 +102,7 @@ ${prefix}_set_encrypt_key:
     adrp	x6,:pg_hi21:BORINGSSL_function_hit
     add x6, x6, :lo12:BORINGSSL_function_hit
 	mov w7, #1
-	strb w7, [x6,#2]
+	strb w7, [x6,#3]
 #endif
 ___
 $code.=<<___	if ($flavour =~ /64/);
@@ -355,7 +355,7 @@ ${prefix}_${dir}crypt:
     adrp	x6,:pg_hi21:BORINGSSL_function_hit
     add x6, x6, :lo12:BORINGSSL_function_hit
 	mov w7, #1
-	strb w7, [x6]
+	strb w7, [x6,#1]
 #endif
 	AARCH64_VALID_CALL_TARGET
 	ldr	$rounds,[$key,#240]
@@ -734,6 +734,13 @@ $code.=<<___;
 .type	${prefix}_ctr32_encrypt_blocks,%function
 .align	5
 ${prefix}_ctr32_encrypt_blocks:
+#ifdef BORINGSSL_DISPATCH_TEST
+.extern        BORINGSSL_function_hit
+    adrp	x6,:pg_hi21:BORINGSSL_function_hit
+    add x6, x6, :lo12:BORINGSSL_function_hit
+	mov w7, #1
+	strb w7, [x6]
+#endif
 ___
 $code.=<<___	if ($flavour =~ /64/);
 	// Armv8.3-A PAuth: even though x30 is pushed to stack it is not popped later.
--- a/crypto/fipsmodule/aes/asm/aesv8-armx.pl
+++ b/crypto/fipsmodule/aes/asm/aesv8-armx.pl
@@ -97,15 +97,15 @@ $code.=<<___;
 .align	5
 ${prefix}_set_encrypt_key:
 .Lenc_key:
+___
+$code.=<<___	if ($flavour =~ /64/);
 #ifdef BORINGSSL_DISPATCH_TEST
 .extern        BORINGSSL_function_hit
     adrp	x6,:pg_hi21:BORINGSSL_function_hit
     add x6, x6, :lo12:BORINGSSL_function_hit
-	mov w7, #1
-	strb w7, [x6,#3]
+    mov w7, #1
+    strb w7, [x6,#3]
 #endif
-___
-$code.=<<___	if ($flavour =~ /64/);
 	// Armv8.3-A PAuth: even though x30 is pushed to stack it is not popped later.
 	AARCH64_VALID_CALL_TARGET
 	stp	x29,x30,[sp,#-16]!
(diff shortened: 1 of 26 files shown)


crypto/fipsmodule/aes/asm/aesv8-armx.pl
```diff
@@ -99,6 +99,13 @@
 .Lenc_key:
 ___
 $code.=<<___	if ($flavour =~ /64/);
+#ifdef BORINGSSL_DISPATCH_TEST
+.extern        BORINGSSL_function_hit
+	adrp	x6,:pg_hi21:BORINGSSL_function_hit
+	add x6, x6, :lo12:BORINGSSL_function_hit
+	mov w7, #1
+	strb w7, [x6,#3]
```
dkostic, line 107: could you please add a comment stating the name of the flag that's being set (throughout the PR)?

crypto/fipsmodule/cpucap/cpucap.c
```diff
@@ -85,8 +85,12 @@ HIDDEN uint32_t OPENSSL_armcap_P = 0;
 
 #if defined(BORINGSSL_DISPATCH_TEST)
 // This value must be explicitly initialized to zero. See similar comment above.
+#if defined(OPENSSL_AARCH64)
+HIDDEN uint8_t BORINGSSL_function_hit[7] = {0};
```
dkostic, line 89: can we keep only the larger definition here? we have enough ifdef's already :)

crypto/impl_dispatch_test.cc
```diff
@@ -31,14 +31,19 @@
 #include "fipsmodule/cpucap/internal.h"
 #include "fipsmodule/modes/internal.h"
 
-
+  // bool vaes_vpclmulqdq_ = false;
```
dkostic, line 34: what's this comment supposed to tell us?

crypto/impl_dispatch_test.cc
```diff
@@ -31,14 +31,19 @@
...
   void SetUp() override {
 #if defined(OPENSSL_X86) || defined(OPENSSL_X86_64)
-    aesni_ = CRYPTO_is_AESNI_capable();
+    aes_hw = CRYPTO_is_AESNI_capable();
     avx_movbe_ = CRYPTO_is_AVX_capable() && CRYPTO_is_MOVBE_capable();
-    ssse3_ = CRYPTO_is_SSSE3_capable();
+    aes_vpaes = CRYPTO_is_SSSE3_capable();
     sha_ext_ = CRYPTO_is_SHAEXT_capable();
     vaes_vpclmulqdq_ =
         (OPENSSL_ia32cap_P[2] & 0xC0030000) &&         // AVX512{F+DQ+BW+VL}
```
dkostic, line 49: while you are hear, can you please change this to use the appropriate `CRYPTO_is_*` functions.

crypto/impl_dispatch_test.cc
```diff
@@ -61,7 +66,12 @@ class ImplDispatchTest : public ::testing::Test {
 #else
         false;
 #endif // MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX
-#endif  // X86 || X86_64
+#elif defined(OPENSSL_AARCH64)
+    aes_hw = CRYPTO_is_ARMv8_AES_capable();
+    aes_vpaes = CRYPTO_is_NEON_capable();
+    armv8_gcm_pmull_ = CRYPTO_is_ARMv8_PMULL_capable();
```
dkostic, line 72: nit: can we stick to `aes_*` naming for these two (without `armv8`) for consistency?

crypto/impl_dispatch_test.cc
```diff
@@ -88,50 +98,72 @@ class ImplDispatchTest : public ::testing::Test {
...
+          {kFlag_aesv8_gcm_8x_enc_128, aes_hw &&
+           armv8_gcm_pmull_ && armv8_gcm_8x_}
+#endif
       },
       [] {
         const uint8_t kZeros[16] = {0};
+#if defined(OPENSSL_X86) || defined(OPENSSL_X86_64)
         const uint8_t kPlaintext[40] = {1, 2, 3, 4, 0};
+#else
+        const uint8_t kPlaintext[256] = {1, 2, 3, 4, 0};
```
dkostic, line 165: can we keep only this one and avoid the ifdef?

crypto/impl_dispatch_test.cc
billbo-yang: I thought had removed these before marking the PR as ready, will remove.

crypto/internal.h
```diff
@@ -1021,7 +1022,19 @@ OPENSSL_INLINE int boringssl_fips_break_test(const char *test) {
...
 //   7: aes_gcm_encrypt_avx512
+// On AARCH64:
+//   0: aes_hw_ctr32_encrypt_blocks
+//   1: aes_hw_encrypt
+//   2: aes_gcm_enc_kernel
+//   3: aes_hw_set_encrypt_key
+//   4: vpaes_encrypt
+//   5: vpaes_set_encrypt_key
+//   6: aesv8_gcm_8x_enc_128
+#if defined(OPENSSL_AARCH64)
```
nebeid, line 1033: Does this need to match the change made in cpucap.c to keep the larger array only?

billbo-yang: I must have missed this when removing the larger array, although it seems like it isn't causing a problem with the code working I'll remove it anyway so that the code will be cleaner / have fewer ifdefs :)

nebeid (commented):
The bits for SHA-256 and SHA-512 are not tested here. Do we want to include them or leave them to another PR?

billbo-yang:
> The bits for SHA-256 and SHA-512 are not tested here. Do we want to include them or leave them to another PR?

These tests are looking to see if we're correctly hitting any special hardware instructions present on the CPU. We support the Intel SHA Extensions on supported x86 processors, hence why we test for it on x86 (as opposed to testing whether we hit the regular assembly instructions) in the dispatch test. However, looking at [cpucap/internal.h](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/cpucap/internal.h#L169), I don't see any support for any special SHA hardware extension on aarch64.

Given that on x86 we don't test to see if we hit the regular SHA assembly instructions, I don't think we need tests to hit the regular SHA assembly instruction on aarch64 either.

nebeid (approved):

dkostic (approved):

PR #1100, opened 2023-07-19
Allow `SSL_CTX` to use the new multiple certificate slots internally

samuel40791765:
### Issues:
Addresses `CryptoAlg-1844`

### Description of changes: 
This builds upon the new framework in https://github.com/aws/aws-lc/pull/1086 to use the new multiple certificate slots available. The SSL connection should use the corresponding index of the certificate and pkey type assigned. As of now, the connection will directly use the last certificate that was set for it's connection. This corresponds with our original behavior before we reintroduced the multiple certificate slots framework, where we only maintained one `x509_leaf` and one `pkey per `CERT`.`
A subsequent PR will be introduced to allow the user to manually select which certificate slots they wish to use.

### Call-outs:
These functions change the certificate slot index:
* `SSL_CTX_use_certificate_ASN1` / `SSL_use_certificate_ASN1`/ `SSL_CTX_use_certificate` / `SSL_use_certificate`  (functions that call `ssl_set_cert` internally)
* `SSL_use_*_PrivateKey_*` (functions that call `ssl_set_pkey` internally)

The naming for the list of functions are pretty straightforward. Users setting these functions would have expected `SSL_CTX` to use the new set certificate/pkey, so it's natural that we switch over to that slot. Despite it seeming straightforward, definately let me know if I should add more documentation regarding this anyways

`SSL_CTX_select_current_cert` is a bit more interesting since it selects an existing cert in `SSL_CTX` to use. This will be implemented in one of the next PRs.
* https://github.com/aws/aws-lc/commit/44628096232d22160840b2fa2ea8438343879f6f

### Testing:
I've added a new test that sets up a connection with different certificate types and checks if the correct slot is being used.
`ED25519` is incompatible with older versions of TLS, so those are skipped.

### DIFF
--- a/ssl/ssl_cert.cc
+++ b/ssl/ssl_cert.cc
@@ -250,6 +250,34 @@ enum leaf_cert_and_privkey_result_t {
   leaf_cert_and_privkey_mismatch,
 };
 
+// do_leaf_cert_and_privkey_checks does the necessary checks against |pubkey|
+// and |privkey|. It's  expected that the |pubkey| has been parsed from
+// |cert_cbs|.
+static enum leaf_cert_and_privkey_result_t do_leaf_cert_and_privkey_checks(
+    const CBS *cert_cbs, EVP_PKEY *pubkey, EVP_PKEY *privkey) {
+  if (!ssl_is_key_type_supported(EVP_PKEY_id(pubkey))) {
+    OPENSSL_PUT_ERROR(SSL, SSL_R_UNKNOWN_CERTIFICATE_TYPE);
+    return leaf_cert_and_privkey_error;
+  }
+
+  // An ECC certificate may be usable for ECDH or ECDSA. We only support ECDSA
+  // certificates, so sanity-check the key usage extension.
+  if (EVP_PKEY_id(pubkey) == EVP_PKEY_EC &&
+      !ssl_cert_check_key_usage(cert_cbs, key_usage_digital_signature)) {
+    OPENSSL_PUT_ERROR(SSL, SSL_R_UNKNOWN_CERTIFICATE_TYPE);
+    return leaf_cert_and_privkey_error;
+  }
+
+  if (privkey != nullptr &&
+      // Sanity-check that the private key and the certificate match.
+      !ssl_compare_public_and_private_key(pubkey, privkey)) {
+    ERR_clear_error();
+    return leaf_cert_and_privkey_mismatch;
+  }
+
+  return leaf_cert_and_privkey_ok;
+}
+
 // check_leaf_cert_and_privkey checks whether the certificate in |leaf_buffer|
 // and the private key in |privkey| are suitable and coherent. It returns
 // |leaf_cert_and_privkey_error| and pushes to the error queue if a problem is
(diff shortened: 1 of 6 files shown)


ssl/ssl_lib.cc
```diff
@@ -2476,16 +2485,21 @@ EVP_PKEY *SSL_get_privatekey(const SSL *ssl) {
     assert(ssl->config);
     return NULL;
   }
-  if (ssl->config->cert != NULL) {
-    return ssl->config->cert->privatekey.get();
+  if (ssl->config->cert != NULL &&
+      ssl_cert_check_cert_private_keys_usage(ssl->config->cert.get())) {
```
justsmth, line 2489: If the null check is added to `ssl_cert_check_cert_private_keys_usage`, then it's not needed here.

ssl/ssl_privkey.cc
```diff
@@ -665,7 +691,7 @@ static bool parse_sigalg_pairs(Array<uint16_t> *out, const int *values,
 
   for (size_t i = 0; i < num_values; i += 2) {
     const int hash_nid = values[i];
-    const int pkey_type = values[i+1];
+    const int pkey_type = values[i + 1];
```
justsmth, line 694: Nit: For safety, the loop should terminate if  `i+1 >= num_values`
```
for (size_t i = 0; i < num_values - 1; i += 2) {
```

ssl/ssl_lib.cc
```diff
@@ -1757,16 +1757,25 @@ int SSL_has_pending(const SSL *ssl) {
...
+    return 0;
+  }
+  return ssl_cert_check_private_key(
+      ctx->cert.get(),
+      ctx->cert->cert_private_keys[ctx->cert->cert_private_key_idx]
+          .privatekey.get());
 }
 
 int SSL_check_private_key(const SSL *ssl) {
-  if (!ssl->config) {
```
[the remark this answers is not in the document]
samuel40791765: * These were probably left over from OpenSSL that hadn't been refactored. OpenSSL code has a really bad habit of not doing proper null checks. It might be worth cleaning these up all at once in another SIM/PR.
* The behavior is well documented in `SSL_set_shutdown` so I believe its intentional. `SSL_CTX_set_quiet_shutdown` is encouraged instead.

skmcgrail (dismissed):

ssl/ssl_cert.cc
```diff
@@ -334,12 +334,29 @@ static int cert_set_chain_and_key(
...
 
-  CERT_PKEY &cert_pkey = cert->cert_private_keys[cert->cert_private_key_idx];
+  int slot_index = ssl_get_certificate_slot_index_from_buffer(buffer.get());
+  if (slot_index < 0) {
+    return false;
+  }
+  CERT_PKEY &cert_pkey = cert->cert_private_keys[slot_index];
 
   switch (
       check_leaf_cert_and_privkey(buffer.get(), cert_pkey.privatekey.get())) {
```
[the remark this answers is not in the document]
samuel40791765: Great point, I hadn't looked quite into that since it require some additional logic upon the stuff that's already here, but I'll work on it :).

samuel40791765:
While working on Justin's PR comments, I noticed that `SSL_CTX_set_chain_and_key` (custom BoringSSL only function) also sets up the certificate and private key for use. I fixed it and added a test for it while I was at it.

ssl/ssl_cert.cc
```diff
@@ -250,45 +250,52 @@ enum leaf_cert_and_privkey_result_t {
...
     OPENSSL_PUT_ERROR(SSL, SSL_R_UNKNOWN_CERTIFICATE_TYPE);
     return leaf_cert_and_privkey_error;
   }
 
-  if (privkey != NULL &&
+  if (privkey != nullptr &&
       // Sanity-check that the private key and the certificate match.
-      !ssl_compare_public_and_private_key(pubkey.get(), privkey)) {
+      !ssl_compare_public_and_private_key(pubkey, privkey)) {
     ERR_clear_error();
```
justsmth, line 274: Is there a reason we "clear" errors when the keys mismatch?

justsmth (approved):

skmcgrail (approved):

PR #1109, opened 2023-07-20
Fix DH_check() excessive time with oversized modulus

skmcgrail:
### Description of changes: 
Addresses CVE-2023-3446 which was low severity issue reported to OpenSSL on July 13th 2023. (See [bulletin](https://www.openssl.org/news/secadv/20230719.txt).

Excessively long DH keys can be slow to check which may lead to denial of service.

### DIFF
--- a/crypto/fipsmodule/dh/check.c
+++ b/crypto/fipsmodule/dh/check.c
@@ -60,6 +60,7 @@
 
 #include "internal.h"
 
+#define OPENSSL_DH_CHECK_MAX_MODULUS_BITS  32768
 
 int DH_check_pub_key(const DH *dh, const BIGNUM *pub_key, int *out_flags) {
   *out_flags = 0;
@@ -125,6 +126,13 @@ int DH_check(const DH *dh, int *out_flags) {
   BIGNUM *t1 = NULL, *t2 = NULL;
 
   *out_flags = 0;
+
+  /* Don't do any checks at all with an excessively large modulus */
+  if (BN_num_bits(dh->p) > OPENSSL_DH_CHECK_MAX_MODULUS_BITS) {
+    OPENSSL_PUT_ERROR(DH, DH_R_MODULUS_TOO_LARGE);
+    return 0;
+  }
+
   ctx = BN_CTX_new();
   if (ctx == NULL) {
     goto err;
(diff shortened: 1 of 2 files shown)


crypto/fipsmodule/dh/check.c
```diff
@@ -60,6 +60,7 @@
 
 #include "internal.h"
 
+#define OPENSSL_DH_CHECK_MAX_MODULUS_BITS  32768
```
justsmth, line 63: Is there a reason that `32768` was chosen?  Why not smaller? Why not larger?

The bulletin you linked only mentions 10,000:
```
Trying to use
a very large modulus is slow and OpenSSL will not normally use a modulus which
is over 10,000 bits in length.
```

skmcgrail: The 10,000 bits is the size OpenSSL limits for generation currently, but they chose to use a larger number for checking, as keys not generated by OpenSSL may have been potentially larger.

skmcgrail: >The DH_check() function checks numerous aspects of the key or parameters
that have been supplied. Some of those checks use the supplied modulus
value even if it is excessively large.
>
>There is already a maximum DH modulus size (10,000 bits) over which
OpenSSL will not generate or derive keys. DH_check() will however still
perform various tests for validity on such a large modulus. We introduce a
new maximum (32,768) over which DH_check() will just fail.
>
>An application that calls DH_check() and supplies a key or parameters
obtained from an untrusted source could be vulnerable to a Denial of
Service attack.
>
>The function DH_check() is itself called by a number of other OpenSSL
functions. An application calling any of those other functions may
similarly be affected. The other functions affected by this are
DH_check_ex() and EVP_PKEY_param_check().

The original openssl commit message details explains better.

crypto/fipsmodule/dh/check.c
```diff
@@ -125,6 +126,13 @@ int DH_check(const DH *dh, int *out_flags) {
   BIGNUM *t1 = NULL, *t2 = NULL;
 
   *out_flags = 0;
+
+  /* Don't do any checks at all with an excessively large modulus */
```
samuel40791765, line 130: Tiny tiny nit for our coding style
```suggestion
  // Don't do any checks at all with an excessively large modulus
```

samuel40791765 (approved):

torben-hansen (approved):
Also note we use 64 rounds of MR, not 128.

justsmth (approved):

samuel40791765 (approved):

justsmth (approved):

samuel40791765 (approved):

PR #1113, opened 2023-07-24
Abstract fips entropy functions

torben-hansen:
### Description of changes: 

Draft for now.

Abstract getting data from the configured entropy source of FIPS. Atm, the configured source is hard-coded in multiple workflows. This allows adding another source more easily.

In addition, add some machinery to configure another source of build-time using the new define `FIPS_ENTROPY_SOURCE_JITTER_CPU`.

Depends on: https://github.com/aws/aws-lc/pull/1112.

### Testing

This doesn't mutate any behaviour.

### DIFF
--- a/crypto/fipsmodule/rand/rand.c
+++ b/crypto/fipsmodule/rand/rand.c
@@ -65,6 +65,8 @@
 
 #if defined(BORINGSSL_FIPS)
 
+// In the future this will be demoted to a build-time option.
+// Another entropy source will be promoted to the default source.
 #define FIPS_ENTROPY_SOURCE_JITTER_CPU 1
 
 #if !defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
@@ -208,7 +210,8 @@ static void rand_thread_state_clear_all(void) {
   // The locks are deliberately left locked so that any threads that are still
   // running will hang if they try to call |RAND_bytes|.
 }
-#endif
+
+#endif // defined(BORINGSSL_FIPS)
 
 // rand_thread_state_free frees a |rand_thread_state|. This is called when a
 // thread exits.
--- a/crypto/fipsmodule/rand/rand.c
+++ b/crypto/fipsmodule/rand/rand.c
@@ -78,7 +78,7 @@ static const unsigned kReseedInterval = 16777216;
 #else // defined(BORINGSSL_FIPS)
 
 #if defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
-#error "Jitter CPU must not be configured as the entropy source in FIPS mode"
+#error "Jitter CPU must not be configured as the entropy source in non-FIPS mode"
 #endif
 
 static const unsigned kReseedInterval = 4096;
--- a/crypto/fipsmodule/rand/rand.c
+++ b/crypto/fipsmodule/rand/rand.c
@@ -64,10 +64,24 @@
 // reseeding.
 
 #if defined(BORINGSSL_FIPS)
-static const unsigned kReseedInterval = 16777216;
-#else
-static const unsigned kReseedInterval = 4096;
+
+#define FIPS_ENTROPY_SOURCE_JITTER_CPU 1
+
+#if !defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
+#error "Currently, Jitter CPU must be configured as the entropy source in FIPS mode"
 #endif
+
+static const unsigned kReseedInterval = 16777216;
+
+#else // defined(BORINGSSL_FIPS)
+
+#if defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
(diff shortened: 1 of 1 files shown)


crypto/fipsmodule/rand/rand.c
```diff
@@ -64,14 +64,24 @@
 // reseeding.
 
 #if defined(BORINGSSL_FIPS)
+
+#define FIPS_ENTROPY_SOURCE_JITTER_CPU 1
+
+#if !defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
```
skmcgrail, line 70: Won't this always be true due to line 68?

crypto/fipsmodule/rand/rand.c
```diff
@@ -64,14 +64,24 @@
...
+#endif
+
 static const unsigned kReseedInterval = 16777216;
-#else
-static const unsigned kReseedInterval = 4096;
+
+#else // defined(BORINGSSL_FIPS)
+
+#if defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
+#error "Jitter CPU must not be configured as the entropy source in FIPS mode"
```
skmcgrail, line 79: nit: "non-FIPS mode"

crypto/fipsmodule/rand/rand.c
```diff
@@ -107,6 +114,78 @@ DEFINE_BSS_GET(struct rand_thread_state *, thread_states_list)
...
+
+static void rand_state_fips_init(struct rand_thread_state *state) {
+
+#if defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
+  // Initialize the thread-local Jitter instance.
+  state->jitter_ec = NULL;
+  // The first parameter passed to |jent_entropy_collector_alloc| function is
+  // the desired oversampling rate. Passing a 0 tells Jitter module to use
+  // the default rate (which is 3 in Jitter v3.1.0).
+  state->jitter_ec = jent_entropy_collector_alloc(0, JENT_FORCE_FIPS);
```
skmcgrail, line 133: Will this eventually move outside the FIPS boundary when we implemented passive entropy, or will jitter entropy within the FIPS boundary always be an option?

dkostic (dismissed):

crypto/fipsmodule/rand/rand.c
torben-hansen: Yes, for now it will! But later this will be demoted to a choice. Just paranoid.

crypto/fipsmodule/rand/rand.c
torben-hansen: It will be a build-time option. When demoting jitter entropy, I will implement a cmake option that can be used.

crypto/fipsmodule/rand/rand.c
```diff
@@ -119,16 +198,12 @@ __declspec(allocate(".CRT$XCU")) void(*fips_library_destructor)(void) =
...
 static void rand_thread_state_clear_all(void) {
   CRYPTO_STATIC_MUTEX_lock_write(thread_states_list_lock_bss_get());
   CRYPTO_STATIC_MUTEX_lock_write(state_clear_all_lock_bss_get());
   for (struct rand_thread_state *cur = *thread_states_list_bss_get();
        cur != NULL; cur = cur->next) {
-    CTR_DRBG_clear(&cur->drbg);
-    OPENSSL_cleanse(cur->last_block, sizeof(cur->last_block));
-
-    jent_entropy_collector_free(cur->jitter_ec);
+    rand_state_fips_clear(cur);
```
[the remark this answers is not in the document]
torben-hansen: Sorta - `rand_thread_state_clear_all` is strictly a fips-only use. So, it made sense to move it there IMO.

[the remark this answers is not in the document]
torben-hansen: In non-fips, the thread-local destructor `rand_thread_state_free` fires and will release the thread-local rand state (also note that the drbg state buffers are flat memory in the rand state)

dkostic (approved):

skmcgrail (approved):

crypto/fipsmodule/rand/rand.c
```diff
@@ -98,6 +116,78 @@ DEFINE_BSS_GET(struct rand_thread_state *, thread_states_list)
...
+#endif
+}
+
+static void rand_state_fips_maybe_want_additional_input(
+  uint8_t additional_input[CTR_DRBG_ENTROPY_LEN],
+  size_t *additional_input_len) {
+
+  *additional_input_len = 0;
+
+#if defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
```
nebeid, line 148: Do we want to surround this with `if defined(OPENSSL_URANDOM)` as it was before?

torben-hansen: `CRYPTO_sys` should always be defined.

PR #1116, opened 2023-07-25
ABI Diff GitHub Action

skmcgrail:
### Description of changes: 
Uses [libabigail](https://sourceware.org/libabigail/), which is available in Amazon Linux repositories for doing simple ABI Diffing between a pull request and the target branch to be merged into.

#### Notes
* Currently just diffs the "standard" non-FIPS build, on x86-64 just to give some base coverage. We can expand this further at a later date.
* The comparison is between the PR branch and the target branch. In a future revision I intend to also have it diff the PR branch to the last tagged release.

### DIFF
--- a/.github/docker_images/abidiff/build.sh
+++ b/.github/docker_images/abidiff/build.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+set -ex
+cmake -S . -B build -GNinja -DBUILD_SHARED_LIBS=1 -DCMAKE_BUILD_TYPE=RelWithDebInfo
+cmake --build build
--- a/.github/docker_images/abidiff/diff.sh
+++ b/.github/docker_images/abidiff/diff.sh
@@ -19,8 +19,22 @@ else
 fi
 
 set +e
+
+# https://sourceware.org/libabigail/manual/abidiff.html
 abidiff --hd1 "/previous/include" --hd2 "/next/include" "/previous/${ARTIFACT_PATH}" "/next/${ARTIFACT_PATH}"
-if [[ $? -ge 8 ]]; then
+
+# From the manual page:
+#
+# The exit code of the abidiff command is either 0 if the ABI of the binaries being compared are equal,
+# or non-zero if they differ or if the tool encountered an error.
+# ...
+# The third bit, of value 4, named ABIDIFF_ABI_CHANGE means the ABI of the binaries being compared are different.
+# The fourth bit, of value 8, named ABIDIFF_ABI_INCOMPATIBLE_CHANGE means the ABI of the binaries compared are different
+# in an incompatible way. If this bit is set, then the ABIDIFF_ABI_CHANGE bit must be set as well. If the
+# ABIDIFF_ABI_CHANGE is set and the ABIDIFF_INCOMPATIBLE_CHANGE is NOT set, then it means that the ABIs being compared
+# might or might not be compatible. In that case, a human being needs to review the ABI changes to decide if they are
+# compatible or not.
+if [[ $? -ge 4 ]]; then
   exit 1
 else
   exit 0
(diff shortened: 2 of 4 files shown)


.github/docker_images/abidiff/diff.sh
```diff
@@ -0,0 +1,27 @@
...
+  ARTIFACT_PATH="${ARTIFACT_PATH}/crypto/libcrypto.so"
+elif [[ "${CHECK_LIB}" == "ssl" ]]; then
+  ARTIFACT_PATH="${ARTIFACT_PATH}/ssl/libssl.so"
+else
+  exit 1
+fi
+
+set +e
+abidiff --hd1 "/previous/include" --hd2 "/next/include" "/previous/${ARTIFACT_PATH}" "/next/${ARTIFACT_PATH}"
+if [[ $? -ge 8 ]]; then
```
andrewhop, line 23: What is special about 8? Can you add a link to the documentation.

.github/workflows/abidiff.yml
```diff
@@ -0,0 +1,50 @@
...
+          ref: ${{ github.base.ref }}
+          path: ${{ github.workspace }}/previous
+      - name: Build Docker Image
+        working-directory: ${{ github.workspace }}/next/.github/docker_images/abidiff
+        run: |
+          docker build -t abidiff .
+      - name: Perform ABI Diff
+        run: |
+          docker run -v ${{ github.workspace }}/previous:/previous -v ${{ github.workspace }}/next:/next -- abidiff crypto
+  ssl:
```
andrewhop, line 30: This is very minor, just curious: does this result in build AWS-LC 4 times? next/previous each for libcrypto and libssl?

justsmth: 💰💰💰

Would we need to double this to 8 builds to get 4 CI tasks:
* Crypto API Change
* Crypto Incompatible API Change
* SSL API Change
* SSL Incompatible API Change

.github/docker_images/abidiff/diff.sh
```diff
@@ -0,0 +1,27 @@
...
+if [[ "${CHECK_LIB}" == "crypto" ]]; then
+  ARTIFACT_PATH="${ARTIFACT_PATH}/crypto/libcrypto.so"
+elif [[ "${CHECK_LIB}" == "ssl" ]]; then
+  ARTIFACT_PATH="${ARTIFACT_PATH}/ssl/libssl.so"
+else
+  exit 1
+fi
+
+set +e
+abidiff --hd1 "/previous/include" --hd2 "/next/include" "/previous/${ARTIFACT_PATH}" "/next/${ARTIFACT_PATH}"
```
justsmth, line 22: Does the fact that "crypto" and "ssl" headers are mixed potentially affect the results when the test should only be considering one of them?

.github/workflows/abidiff.yml
justsmth: 💰💰💰

Would we need to double this to 8 builds to get 4 CI tasks:
* Crypto API Change
* Crypto Incompatible API Change
* SSL API Change
* SSL Incompatible API Change

.github/docker_images/abidiff/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+set -ex
+cmake -S . -B build -GNinja -DBUILD_SHARED_LIBS=1 -DCMAKE_BUILD_TYPE=RelWithDebInfo
```
WillChilds-Klein, line 7: does including debug info affect the ABI?

.github/workflows/abidiff.yml
skmcgrail: I think I found a way to combine them such that the overall action will correctly fail, but it will still diff libssl if the libcrypto step fails, thus allowing the sharing of the built artifacts.

.github/docker_images/abidiff/diff.sh
skmcgrail: No, its just a means to do an initial pass at filtering the symbols in the library, and only focusing on the ones that match with those defined in the headers. So think something along the lines of `nm <library so> | grep <symbols from headers>`

torben-hansen (approved):

andrewhop (approved):

PR #1117, opened 2023-07-27
tweak back error string functions with exclamation marks

samuel40791765:
### Issues:
Resolves `CryptoAlg-1995`

### Description of changes: 
Tweak back string functions with exclamation marks.
Reverted from: https://github.com/aws/aws-lc/commit/c2ae53db6dff3b8a773c43e16e3efaf20ff4c7da

### Call-outs:
N/A

### Testing:
N/A

### DIFF
--- a/ssl/ssl_stat.cc
+++ b/ssl/ssl_stat.cc
@@ -98,7 +98,7 @@ const char *SSL_state_string_long(const SSL *ssl) {
 }
 
 const char *SSL_state_string(const SSL *ssl) {
-  return "!!!!!!";
+  return SSL_state_string_long(ssl);
 }
 
 const char *SSL_alert_type_string_long(int value) {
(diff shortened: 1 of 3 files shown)


skmcgrail (dismissed):

ssl/ssl_stat.cc
```diff
@@ -113,11 +113,111 @@ const char *SSL_alert_type_string_long(int value) {
...
 const char *SSL_alert_type_string(int value) {
-  return "!";
+  value >>= 8;
+  if (value == SSL3_AL_WARNING) {
+    return "W";
+  } else if (value == SSL3_AL_FATAL) {
+    return "F";
+  }
+
+  return "U";
```
justsmth, line 123: Should we to mask out the bits we don't care about?
```
value &= 0x0f00;
value >>= 8;
```

justsmth (approved):

ssl/ssl_stat.cc
samuel40791765: Sure, might as well make the improvement :)

justsmth (dismissed):

ssl/ssl_test.cc
```diff
@@ -9683,5 +9683,19 @@ TEST(SSLTest, InvalidSignatureAlgorithm) {
       ctx.get(), kDuplicatePrefs, OPENSSL_ARRAY_SIZE(kDuplicatePrefs)));
 }
 
+TEST(SSLTest, ErrorStrings) {
+  int warning_value = SSL3_AD_CLOSE_NOTIFY | (SSL3_AL_WARNING << 8);
+  int fatal_value = SSL3_AD_UNEXPECTED_MESSAGE | (SSL3_AL_FATAL << 8);
+  int unknown_value = 99999;
+
+  EXPECT_EQ(strncmp(SSL_alert_desc_string(warning_value), "CN", 2), 0);
```
andrewhop, line 9691: NP: for next time you can use `Bytes(SSL_alert_desc_string(warning_value))` so if the assert is wrong you can a nice output like "expected foo, got bar" instead of "expected 0, got 1".

andrewhop (approved):

andrewhop (approved):

nebeid (approved):

PR #1120, opened 2023-07-27
Retrieve multiple certificate slot for TLS1.2/1.3 based on negotiated sigalgs

samuel40791765:
### Issues:
Addresses `CryptoAlg-1844`

### Description of changes: 
There are two issues with our SSL handshake blocking us from supporting automated certificate slot selection. 

1. We set the server certificate too early in the configuration.
2. We computes the signature algorithms too late in the handshake. 

OpenSSL’s multiple certificate functionality depends on switching to a valid configured certificate based on what signature algorithms the client indicates support for. They negotiate the available cipher suites first. They then negotiate the signature algorithm right after and decide which certificate to use while doing so.

* OpenSSL all TLS signature algorithm negotiation and certificate handling: https://github.com/openssl/openssl/blob/master/ssl/t1_lib.c#L3701-L3851
* OpenSSL all TLS signature algorithm negotiation location: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/ssl/statem/statem_srvr.c#L2278-L2291

This will require a small restructure of our TLS handshakes. We would basically have to move our TLS certificate selection configuration during signature algorithm negotiation like what OpenSSL is doing. This gets convoluted since we have some behavior that relies on the public key extracted from the certificate we’ve decided to use earlier, instead of the private key like OpenSSL is relying on.

AWS-LC sets the certificate for all TLS versions early in [do_select_certificate](https://github.com/aws/aws-lc/blob/7bdae649ee6a723a5197e27ec3770c37f6a64e85/ssl/handshake_server.cc#L757-L775). We branch out to TLS1.3 later if supported, or move further on within the TLS1.0-1.2 handshake. `ssl_on_certificate_selected` should be called once we know which certificate we wish to use.

### Changes

1. TLS 1.0-1.2 has to determine the key exchange and authentication cipher suite masks compatible with the server configuration during cipher suite negotiation. This is done in the [ssl_get_compatible_server_ciphers](https://github.com/aws/aws-lc/blob/ca8d8d5fb0efdd37f67ae1d9c8b657509351c247/ssl/handshake_server.cc#L295-L326) function, where we return a mask based on the extracted public key. To let our server configuration know that we support more authentication cipher suites, we can just additionally loop over the available private keys and add masks accordingly. This isn’t needed for TLS1.3 since all cipher suites are compatible with available authentication schemes.
2. BoringSSL/AWS-LC signature algorithm negotiation is done on demand right before the signature is calculated. We can replicate what OpenSSL is doing by moving the signature algorithm negotiation right after cipher suite negotiation. Our current signature algorithm negotiation uses the handshake’s certificate to check for support. We will have to reconfigure this to check against all available certificates in the connection instead, and switch to one that is supported when a signature algorithm is negotiated. We have [support for signature algorithm priority](https://github.com/aws/aws-lc/blob/ca8d8d5fb0efdd37f67ae1d9c8b657509351c247/ssl/extensions.cc#L4112C34-L4122) in the handshake, so stronger signature schemes are already prioritized. We will need to move signature algorithm negotiation after cipher suite negotiation for both TLS1.0-1.2 and TLS1.3. Since certificate selection is now done later, we will have to move `ssl_on_certificate_selected` till after signature algorithm negotiation is done. 
3. OCSP stapling data should be configured after the certificate is finalized. We will have to move the OCSP stapling callback function to after `ssl_on_certificate_selected` is called.

### Call-outs:
More details in internal doc

### Testing:
New tests with multiple certificate slots configured


skmcgrail:
Is there a way to verify these changes / feature support with the `ssl/test/runner` framework?

samuel40791765:
> Is there a way to verify these changes / feature support with the ssl/test/runner framework?

Great point, I'll look into it

ssl/ssl_test.cc
```diff
@@ -5160,9 +5166,10 @@ class MultipleCertificateSlotTest
     return std::get<1>(GetParam());
   }
 
-  void StandardCertificateSlotIndexTests(SSL_CTX *client_ctx,
-                                         SSL_CTX *server_ctx,
-                                         std::vector<uint16_t> sigalgs) {
+  const void StandardCertificateSlotIndexTests(SSL_CTX *client_ctx,
```
skmcgrail, line 5169: Are any additional tests required to specifically account for the changes in when the OCSP callbacks are handled? What about validating functionality in AWS-LC that requires the public key to be stored (ensuring the right key gets cached etc)?

samuel40791765: * The OCSP callbacks are already tested in our SSL runner tests. If the callback isn't configured or called in the handshake, the tests will fail correspondingly.
https://github.com/aws/aws-lc/blob/a818ddf61ed8bbf41516fbd0b8dabd64e166a9b1/ssl/test/test_config.cc#L1577-L1584

* Great point, I'll have to add tests for that. Also thinking if there's a better way to help us enforce that we reload the public key whenever the certificate is updated.

samuel40791765:
* I added some basic tests with the `ssl/test/runner framework` as requested, but held off on adding more since I didn't want the diff to grow too complicated to review. My current plan is to add more tests to the go framework in a subsequent PR once we add support for TLS1.0/1.1. 
* I tried adding a test to ensure the right public key gets cached, but the state of the used public key gets lost after the handshake ends. I've added an additional sanity check in `ssl_on_certificate_selected` instead, to check if the cached key is the same type as our chosen slot index.

ssl/handshake_server.cc
```diff
@@ -303,10 +303,23 @@ static void ssl_get_compatible_server_ciphers(SSL_HANDSHAKE *hs,
   uint32_t mask_a = 0;
 
   if (ssl_has_certificate(hs)) {
```
darylmartin100, line 305: Should we add a null check for hs here?  There's a lot of functions that just assume hs is safe to use,  tls1_call_ocsp_stapling_callback included.

ssl/ssl_privkey.cc
```diff
@@ -368,28 +368,69 @@ enum ssl_private_key_result_t ssl_private_key_decrypt(SSL_HANDSHAKE *hs,
...
   // emLen be at least hLen + sLen + 2. Both hLen and sLen are the size of the
   // hash in TLS. Reasonable RSA key sizes are large enough for the largest
   // defined RSASSA-PSS algorithm, but 1024-bit RSA is slightly too small for
   // SHA-512. 1024-bit RSA is sometimes used for test credentials, so check the
   // size so that we can fall back to another algorithm in that case.
   const SSL_SIGNATURE_ALGORITHM *alg = get_signature_algorithm(sigalg);
-  if (alg->is_rsa_pss && (size_t)EVP_PKEY_size(hs->local_pubkey.get()) <
-                             2 * EVP_MD_size(alg->digest_func()) + 2) {
+  if (alg->is_rsa_pss &&
+      (size_t)EVP_PKEY_size(pubkey) < 2 * EVP_MD_size(alg->digest_func()) + 2) {
```
darylmartin100, line 380: Should this be less than or equal to?

ssl/handshake_server.cc
samuel40791765: `SSL_HANDSHAKE` is essentially the structure that contains the state of the entire handshake. It's the center piece of the all the handshake state functions.
I assume that these aren't checked since `hs` already has to be non-null in order to progress to this stage. The internal `ssl_run_handshake` function is what runs the handshake code. Before the only place where it's called (`SSL_do_handshake`), we perform a NULL check on `hs` in `SSL_in_init`.
* https://github.com/aws/aws-lc/blob/b99e3d60b9be4e9ee07e0cefdc4c87ca8fbd1e9b/ssl/ssl_lib.cc#L846-L862
* https://github.com/aws/aws-lc/blob/b99e3d60b9be4e9ee07e0cefdc4c87ca8fbd1e9b/ssl/internal.h#L2181-L2184

ssl/ssl_privkey.cc
samuel40791765: From the comment above, "at least hLen + sLen + 2" makes it seems like `hLen + sLen + 2` is a sufficient size to be used.
Just looked into OpenSSL's code and the same check is also enforced :) 
https://github.com/openssl/openssl/blob/master/ssl/t1_lib.c#L1628-L1636

skmcgrail (approved):

justsmth (approved):

PR #1125, opened 2023-08-01
Implementation passive entropy

torben-hansen:
### Issues

CryptoAlg-1953

### Description of changes: 

Implement passive entropy FIPS source.

### Call-outs:

Currently, to compile with this source use `-DUSE_FIPS_ENTROPY_SOURCE_PASSIVE`. In the future this will be swapped over to be the default and the Jitter CPU source will be an option that must be enabled at build-time.

This PR adds back the state parameter to `rand_get_seed` - it makes more sense since the state might be needed and there is no need to query the thread list twice...

### Testing:

Adds an additional target to CI that uses the special build-option `-DUSE_FIPS_ENTROPY_SOURCE_PASSIVE` and pass over one `run_tests` target.


CMakeLists.txt
```diff
@@ -25,6 +25,15 @@ include(cmake/go.cmake)
 
 enable_language(C)
 
+# Configure entropy source in case of FIPS build
+if(FIPS)
+  if(USE_FIPS_ENTROPY_SOURCE_PASSIVE)
```
andrewhop, line 30: Would it be better to inverse this if? 
```suggestion
  if(USE_JITTER)
```

Then you could also enforce you can't set USE_JITTER if FIPS isn't set as well. Otherwise we need to decide what does USE_FIPS_ENTROPY_SOURCE_PASSIVE if FIPS isn't set.

crypto/fipsmodule/rand/internal.h
```diff
@@ -129,10 +129,30 @@ OPENSSL_INLINE int have_fast_rdrand(void) {
 #define MAX_BACKOFF_RETRIES 9
 OPENSSL_EXPORT void HAZMAT_set_urandom_test_mode_for_testing(void);
 
+// Final factor TBD
+#define PASSIVE_ENTROPY_WHITEN_FACTOR 10
```
andrewhop, line 133: Is this a typo? What do you mean by whiten factor?

crypto/fipsmodule/rand/internal.h
```diff
@@ -129,10 +129,30 @@ OPENSSL_INLINE int have_fast_rdrand(void) {
...
 #define JITTER_MAX_NUM_TRIES (3)
+
+#elif defined(FIPS_ENTROPY_SOURCE_PASSIVE)
+
+void RAND_module_entropy_depleted(uint8_t out_entropy[CTR_DRBG_ENTROPY_LEN],
+                                  int *out_want_additional_input);
+void CRYPTO_get_seed_entropy(uint8_t entropy[PASSIVE_ENTROPY_LOAD_LENGTH],
+                             int *out_want_additional_input);
+void RAND_load_entropy(uint8_t out_entropy[CTR_DRBG_ENTROPY_LEN],
+                       uint8_t entropy[PASSIVE_ENTROPY_LOAD_LENGTH]);
```
andrewhop, line 150: From CBMC doc is it worth adding an explicit length parameter to the array?

crypto/fipsmodule/rand/rand.c
```diff
@@ -65,26 +65,28 @@
...
 #if defined(BORINGSSL_FIPS)
 
-// In the future this will be demoted to a build-time option.
-// Another entropy source will be promoted to the default source.
-#define FIPS_ENTROPY_SOURCE_JITTER_CPU 1
-
-#if !defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
-#error "Currently, Jitter CPU must be configured as the entropy source in FIPS mode"
+#if defined(FIPS_ENTROPY_SOURCE_JITTER_CPU) && defined(FIPS_ENTROPY_SOURCE_PASSIVE)
+#error "Only one FIPS entropy source must not be defined"
```
andrewhop, line 69:
```suggestion
#error "Only one FIPS entropy source can be enabled at a time"
```

crypto/fipsmodule/rand/rand.c
```diff
@@ -291,33 +312,52 @@ static int rdrand(uint8_t *buf, size_t len) {
...
+                       uint8_t entropy[PASSIVE_ENTROPY_LOAD_LENGTH]) {
+
+  OPENSSL_memcpy(out_entropy, entropy, CTR_DRBG_ENTROPY_LEN);
+
+  // Whiten the resulting entropy
+  for (size_t i = 1; i < PASSIVE_ENTROPY_WHITEN_FACTOR; i++) {
+    for (size_t j = 0; j < CTR_DRBG_ENTROPY_LEN; j++) {
+      out_entropy[j] ^= entropy[CTR_DRBG_ENTROPY_LEN * i + j];
+    }
   }
```
andrewhop, line 326: What does this do, do you have a link? Is there a reason to do it here and not in the Jitter/non-FIPS case?

CMakeLists.txt
torben-hansen: Not sure I understand your point. I'm gonna flip this later on to passive being the default - as you suggest.

Is it easier to just make this flip now?

crypto/fipsmodule/rand/internal.h
torben-hansen: Whitening is a technique/method to reduce bias in the distribution.

crypto/fipsmodule/rand/internal.h
torben-hansen: I don't see why(?)

crypto/fipsmodule/rand/rand.c
torben-hansen: not sure what my original sentence is actually trying to say. Will corrrect.

CMakeLists.txt
andrewhop: My concern is I don't want to switch the behavior because that's another "breaking" change. But since this is the FIPS release and customers have to follow the specific build instructions and use a specific version of code it's probably fine.

crypto/fipsmodule/rand/internal.h
andrewhop: Having an implicit length allegedly makes it a bit harder to verify.

torben-hansen: Wanna stick with current implementation.

skmcgrail (dismissed):

crypto/fipsmodule/rand/rand.c
```diff
@@ -65,26 +65,28 @@
...
-#if !defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
-#error "Currently, Jitter CPU must be configured as the entropy source in FIPS mode"
+#if defined(FIPS_ENTROPY_SOURCE_JITTER_CPU) && defined(FIPS_ENTROPY_SOURCE_PASSIVE)
+#error "Only one FIPS entropy source can be enabled at a time"
 #endif
 
+#if defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
 static const unsigned kReseedInterval = 16777216;
+#elif defined(FIPS_ENTROPY_SOURCE_PASSIVE)
+static const unsigned kReseedInterval = 4096;
```
andrewhop, line 75: :partying_face:

tests/ci/run_posix_tests.sh
```diff
@@ -26,6 +26,9 @@ build_and_test -DOPENSSL_NO_ASM=1 -DCMAKE_BUILD_TYPE=Release
 echo "Testing building shared lib."
 build_and_test -DBUILD_SHARED_LIBS=1 -DCMAKE_BUILD_TYPE=Release
 
+echo "Testing FIPS entropy source: Passive."
+build_and_test -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DUSE_FIPS_ENTROPY_SOURCE_PASSIVE=1
```
andrewhop, line 30: Is this testing with passive entropy? From the CMake list I think you also need to turn on FIPS:
```suggestion
build_and_test -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DUSE_FIPS_ENTROPY_SOURCE_PASSIVE=1 -DFIPS=1
```

crypto/fipsmodule/rand/rand.c
```diff
@@ -159,6 +174,7 @@ static void CRYPTO_fips_get_from_entropy_source(struct rand_thread_state *state,
   uint8_t *out_entropy, size_t out_entropy_len, int *out_want_additional_input) {
 
 #if defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
+
```
andrewhop, line 177: NP: up on line 172 the comment is out of date for passive entropy which doesn't use the state

torben-hansen: I think it's better just to have a general requirement instead of bi-modal ones.

tests/ci/run_posix_tests.sh
torben-hansen: Nice find

torben-hansen: Using the `fips_build_and_test` function instead.

andrewhop (approved):

skmcgrail (approved):

andrewhop (approved):

andrewhop (approved):

skmcgrail (approved):

PR #1133, opened 2023-08-02
Start of Big Endian PPC support

andrewhop:
### Issues:
Addresses CryptoAlg-1954

### Description of changes: 
* Update CMake to allow the build to continue on Big Endian platforms
* Fix the common crypto load/store functions to work on big endian, this fixes all digests except Blake
* Fix big number loading and storing which also fixes RSA operation on big endian platforms

### Call-outs:
All tests are not passing yet. Known failing things include AES, ECDH, ECDSA, Blake.

Known passing things include: Digests (MD*, SHA*), DES, ChaCha Poly, HMAC, RSA, FFDH.

### Testing:
Tested locally with qemu on 32 bit big endian Power PC, and tested on an actual 64 bit big endian Power PC device. 

### DIFF
--- a/crypto/fipsmodule/bcm.c
+++ b/crypto/fipsmodule/bcm.c
@@ -228,7 +228,7 @@ static void BORINGSSL_bcm_power_on_self_test(void) __attribute__ ((constructor))
 #endif
 
 static void BORINGSSL_bcm_power_on_self_test(void) {
-#if !defined(OPENSSL_NO_ASM)
+#if !defined(OPENSSL_NO_ASM) && !defined(OPENSSL_PPC32BE)
   OPENSSL_cpuid_setup();
 #endif
 
--- a/crypto/fipsmodule/bn/bytes.c
+++ b/crypto/fipsmodule/bn/bytes.c
@@ -136,9 +136,22 @@ BIGNUM *BN_le2bn(const uint8_t *in, size_t len, BIGNUM *ret) {
   // Make sure the top bytes will be zeroed.
   ret->d[num_words - 1] = 0;
 
-  // We only support little-endian platforms, so we can simply memcpy the
-  // internal representation.
+#ifdef OPENSSL_BIG_ENDIAN
+  BN_ULONG word = 0;
+  unsigned m;
+
+  m = (len - 1) % BN_BYTES;
+  for (size_t i = len -1; i < len; i--) {
+    word = (word << 8) | in[i];
+    if (m-- == 0) {
+      ret->d[--num_words] = word;
+      word = 0;
+      m = BN_BYTES - 1;
+    }
+  }
+#else
   OPENSSL_memcpy(ret->d, in, len);
+#endif
   return ret;
 }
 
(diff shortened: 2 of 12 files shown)


crypto/fipsmodule/bcm.c
```diff
@@ -228,7 +228,7 @@ static void BORINGSSL_bcm_power_on_self_test(void) __attribute__ ((constructor))
 #endif
 
 static void BORINGSSL_bcm_power_on_self_test(void) {
-#if !defined(OPENSSL_NO_ASM)
+#if !defined(OPENSSL_NO_ASM) && !defined(OPENSSL_PPC32BE)
```
skmcgrail, line 231: Would it be better to set `OPENSSL_NO_ASM` define in the CMakeLists.txt when you detect big-endian / if the processor is ppc64 in the CMakeLists.txt?

crypto/fipsmodule/bn/bytes.c
```diff
@@ -136,9 +136,22 @@ BIGNUM *BN_le2bn(const uint8_t *in, size_t len, BIGNUM *ret) {
...
   ret->d[num_words - 1] = 0;
 
-  // We only support little-endian platforms, so we can simply memcpy the
-  // internal representation.
+#ifdef OPENSSL_BIG_ENDIAN
+  BN_ULONG word = 0;
+  unsigned m;
+
+  m = (len - 1) % BN_BYTES;
+  for (size_t i = len -1; i < len; i--) {
```
skmcgrail, line 144:
```suggestion
  for (size_t i = len - 1; i < len; i--) {
```

crypto/fipsmodule/bcm.c
PiotrSikora: IMHO, this should define `OPENSSL_ASM_INCOMPATIBLE` (which is later evaluated to `OPENSSL_NO_ASM`) in `include/openssl/base.h` when defining platform-specific variables, since there is a precedent for that there.

Doing this in `CMakeLists.txt` splits logic between build system and source code, which is neither ideal nor necessary in this case.

justsmth (approved):

crypto/fipsmodule/bn/bytes.c
justsmth, line 144: I think the logic would much more clear with the iterators starting are 0:
```
  BN_ULONG output_word = 0;
  size_t byte_index = 0;
  size_t output_word_index = 0;

  for (size_t i = 0; i < len; i++) {
    BN_ULONG input_byte = in[i];
    output_word |= (input_byte << (byte_index * 8));

    if (++byte_index == BN_BYTES) {
      ret->d[output_word_index++] = output_word;
      output_word = 0;
      byte_index = 0;
    }
  }

  if (byte_index != 0) {
    ret->d[output_word_index] = output_word;
  }
```

andrewhop:
Closing in favor of https://github.com/aws/aws-lc/pull/1205 which is rebased on recent changes and going to main.

PR #1134, opened 2023-08-03
add curl integration test CI dimension

samuel40791765:
### Issues:
Resolves `CryptoAlg-1886`

### Description of changes: 
We've recently done integration efforts with libcurl. This is to maintain our compatibility with their tip of main.

### Call-outs:
N/A

### Testing:
New CI dimension

### DIFF
--- a/tests/ci/cdk/cdk/codebuild/github_ci_integration_omnibus.yaml
+++ b/tests/ci/cdk/cdk/codebuild/github_ci_integration_omnibus.yaml
@@ -66,3 +66,13 @@ batch:
         image: 620771051181.dkr.ecr.us-west-2.amazonaws.com/aws-lc-docker-images-linux-x86:ubuntu-22.04_gcc-12x_latest
         variables:
           AWS_LC_CI_TARGET: "tests/ci/integration/run_haproxy_integration.sh"
+
+    - identifier: curl_integration
+      buildspec: tests/ci/codebuild/common/run_simple_target.yml
+      env:
+        type: LINUX_CONTAINER
+        privileged-mode: false
+        compute-type: BUILD_GENERAL1_MEDIUM
+        image: 620771051181.dkr.ecr.us-west-2.amazonaws.com/aws-lc-docker-images-linux-x86:ubuntu-22.04_gcc-12x_latest
+        variables:
+          AWS_LC_CI_TARGET: "tests/ci/integration/run_curl_integration.sh"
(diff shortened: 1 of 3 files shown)


tests/ci/integration/run_curl_integration.sh
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function curl_run_tests() {
+  make -j ${NUM_CPU_THREADS} tests
+  make test-ci
+  cd ${SCRATCH_FOLDER}
+}
+
+# Get latest curl version.
+git clone https://github.com/curl/curl.git ${CURL_SRC_FOLDER}
```
billbo-yang, line 45: Does curl make a guarantee that their main branch will always be buildable/release quality? If not, is that something we should be concerned about in regards to our CI?

tests/ci/cdk/cdk/codebuild/github_ci_integration_omnibus.yaml
```diff
@@ -66,3 +66,13 @@ batch:
...
+
+    - identifier: curl_integration
+      buildspec: tests/ci/codebuild/common/run_simple_target.yml
+      env:
+        type: LINUX_CONTAINER
+        privileged-mode: false
+        compute-type: BUILD_GENERAL1_MEDIUM
+        image: 620771051181.dkr.ecr.us-west-2.amazonaws.com/aws-lc-docker-images-linux-x86:ubuntu-22.04_gcc-12x_latest
+        variables:
+          AWS_LC_CI_TARGET: "tests/ci/integration/run_curl_integration.sh"
```
billbo-yang, line 78: nit: new line is missing 😅

samuel40791765: Yeah I had debated whether to add one, but this pattern is consistent through our other yaml files. There wasn't a new line before I made this change either.

tests/ci/integration/run_curl_integration.sh
samuel40791765: Not much of a guarantee since they do periodic releases, but they run a similar CI dimension against a version of AWS-LC in their CI as well. This helps us have some faith that we'll always be buildable with the tip of main.
* https://github.com/curl/curl/blob/master/.github/workflows/awslc.yml

We've been following a pattern of building with main for most of these integrations. If it's deemed too unstable, we can always pin to a specific version instead :).

andrewhop (approved):

billbo-yang (approved):

PR #1114, opened 2023-07-25
Pull in the latest changes from s2n-bignum (2023-07-25) 

aqjune-aws:
### Issues:
N/A

### Description of changes: 
(2022-07-25) Pull in the latest changes from s2n-bignum.

The commands in the internal Quip document was used, with PATHS_TO_KEEP defined as follows:
```
PATHS_TO_KEEP="./arm/p384 ./x86_att/p384 ./arm/p521 ./x86_att/p521 ./x86_att/curve25519/curve25519_x25519.S ./x86_att/curve25519/curve25519_x25519base.S ./x86_att/curve25519/curve25519_x25519_alt.S ./x86_att/curve25519/curve25519_x25519base_alt.S  ./arm/curve25519/curve25519_x25519.S ./arm/curve25519/curve25519_x25519base.S ./arm/curve25519/curve25519_x25519_alt.S ./arm/curve25519/curve25519_x25519base_alt.S ./arm/curve25519/curve25519_x25519_byte.S ./arm/curve25519/curve25519_x25519base_byte.S ./arm/curve25519/curve25519_x25519_byte_alt.S ./arm/curve25519/curve25519_x25519base_byte_alt.S ./arm/fastmul/bignum_emontredc_8n.S ./arm/fastmul/bignum_kmul_16_32.S ./arm/fastmul/bignum_kmul_32_64.S ./arm/fastmul/bignum_ksqr_16_32.S ./arm/fastmul/bignum_ksqr_32_64.S ./arm/generic/bignum_ge.S ./arm/generic/bignum_mul.S ./arm/generic/bignum_optsub.S ./arm/generic/bignum_sqr.S ./include/_internal_s2n_bignum.h"
```

### Call-outs:
Point out areas that need special attention or support during the review process. Discuss architecture or design changes.

### Testing:
How is this change tested (unit tests, fuzz tests, etc.)? Are there any testing steps to be verified by the reviewer?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

PR #1135, opened 2023-08-03
Update montgomery multiplication to use s2n-bignum's verified scalar bignum functions

aqjune-aws:
### Description of changes: 

This patch updates aws-lc's montgomery multiplication to use s2n-bignum's verified bignum functions.
This is a follow-up of #1114, and is splitted from #1108.
To selectively apply s2n-bignum to Graviton 2, this patch invokes CRYPTO_is_ARMv8_wide_multiplier_capable() and runs the s2n-bignum functions only when wide multipliers are not capable.

The performance numbers of RSA signing are as follows. Graviton 2 is used, and `tool/bssl speed -filter RSA` has been used. (Unit: ops/sec).

Bits | Operation | baseline AWS-LC | s2n-bignum | speedup vs baseline
-- | -- | -- | -- | --
2048 | RSA sign | 299.3 | 399 | 33.31%
  | verify (fresh key) | 10736.3 | 15491 | 44.29%
3072 | RSA sign | 95.4 | 113.2 | 18.66%
  | verify (fresh key) | 4917.7 | 6001.7 | 22.04%
4096 | RSA sign | 41.7 | 63.2 | 51.56%
  | verify (fresh key) | 2781.6 | 3451 | 24.07%

This is only adopted for AArch64 that has narrow multiplication instruction bandwidths.

### Testing:
Tested via bssl speed -filter RSA


dkostic:
nice improvements, can we get perf numbers for GV3 and possibly Apple M1/M2?

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -119,8 +119,47 @@
...
+  return 0;
+
+#endif
+}
+
+OPENSSL_INLINE int montgomery_use_s2n_bignum(unsigned int num) {
+#if defined(BN_MONTGOMERY_USE_S2N_BIGNUM)
+
+  // Use s2n-bignum's functions only if (1) the ARM architecture has slow
+  // multipliers, and (2) temporary buffer's size does not exceed
```
dkostic, line 152: I see two more conditions in the code, can we document them as well? And also explain the `(2 * (uint64_t)num + 96) <= BN_MONTGOMERY_MAX_WORDS` part -- why `2*` and why `96`?

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -418,6 +457,77 @@ static int bn_mod_mul_montgomery_fallback(BIGNUM *r, const BIGNUM *a,
...
+                                           const BN_ULONG *bp,
+                                           const BN_ULONG *np,
+                                           const BN_ULONG *n0, size_t num) {
+
+#if defined(BN_MONTGOMERY_USE_S2N_BIGNUM)
+
+  // t is a temporary buffer used by big-int multiplication.
+  // bignum_kmul_32_64 requires 96 words at maximum.
+  uint64_t t[96];
+  // l is the output buffer of big-int multiplication.
```
dkostic, line 473: what's `l`?

aqjune-aws:
@dkostic For microarchitectures supporting fast enough multiplications (such as Graviton 2 and M1), the current version of s2n-bignum didn't bring effective performance improvement. To selectively apply s2n-bignum to Graviton 2, this patch invokes `CRYPTO_is_ARMv8_wide_multiplier_capable()` and runs the s2n-bignum functions only when wide multipliers are not capable.
This is the relevant part in arm_arch.h:
```c
// The Neoverse V1 and Apple M1 micro-architectures are detected to enable
// high unrolling factor of AES-GCM and other algorithms that leverage a
// wide crypto pipeline and fast multiplier.
#define ARMV8_NEOVERSE_V1 (1 << 12)
#define ARMV8_APPLE_M1 (1 << 13)
```

crypto/fipsmodule/bn/montgomery.c
aqjune-aws: The buffer size (2*num+96) is the arrays' size allocated in `montgomery_s2n_bignum_mul_mont()`, and `num % 8 == 0` is the precondition that s2n-bignum's extended montgomery reduction should work.
I will address these comments in the following commits.

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -119,8 +119,51 @@
...
+    defined(OPENSSL_AARCH64) && !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX) &&   \
+    defined(OPENSSL_BN_ASM_MONT)
+
+#include "../../../third_party/s2n-bignum/include/s2n-bignum_aws-lc.h"
+
+#define BN_MONTGOMERY_USE_S2N_BIGNUM 1
+
+#endif
+
+OPENSSL_INLINE int montgomery_s2n_bignum_capable(void) {
```
dkostic, line 136: not sure why we need this function, can't we combine this one and `montgomery_use_s2n_bignum` into one?

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -119,8 +119,51 @@
...
+
+  // Use s2n-bignum's functions only if
+  // (1) The ARM architecture has slow multipliers, and
+  // (2) num (which is the number of words) is multiplie of 8, because
+  //     s2n-bignum's bignum_emontredc_8n requires it
+  // (3) The word size is 64 bits, and
+  // (4) Temporary buffer's size used in montgomery_s2n_bignum_mul_mont
+  //     does not exceed BN_MONTGOMERY_MAX_WORDS.
+  return !CRYPTO_is_ARMv8_wide_multiplier_capable() && (num % 8 == 0) &&
+         BN_BITS2 == 64 && (2 * (uint64_t)num + 96) <= BN_MONTGOMERY_MAX_WORDS;
```
dkostic, line 159: the temporary buffer size in `montgomery_s2n_bignum_mul_mont` is 96, no (line 476)? This is a bit confusing to me, do you think there is a way to simplify/clarify?

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -418,6 +461,78 @@ static int bn_mod_mul_montgomery_fallback(BIGNUM *r, const BIGNUM *a,
...
+  // mulres is the output buffer of big-int multiplication.
+  // If BN_MONTGOMERY_MAX_WORDS - 96 is larger than num*2, its low num*2
+  // elements are used.
+  // It is montgomery_use_s2n_bignum() that checks whether num*2 fits in the
+  // size of mulres array.
+  uint64_t mulres[BN_MONTGOMERY_MAX_WORDS - 96];
+
+  // BN_ULONG is uint64_t since BN_BITS2 is 64.
+  // m is the prime number, and m * w = -1 mod 2^64.
+  uint64_t *m = (uint64_t *)np;
```
dkostic, line 486: if we are in the `BN_MONTGOMERY_USE_S2N_BIGNUM` case, isn't it guaranteed that `BN_ULONG` is `uint64_t`?

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -418,6 +461,78 @@ static int bn_mod_mul_montgomery_fallback(BIGNUM *r, const BIGNUM *a,
   return ret;
 }
 
+
+#if defined(OPENSSL_BN_ASM_MONT)
+
+static void montgomery_s2n_bignum_mul_mont(BN_ULONG *rp, const BN_ULONG *ap,
```
dkostic, line 467: Can we explicitly state in a comment which parameter sizes are supported by this function?

crypto/fipsmodule/bn/montgomery.c
aqjune-aws: We have two temporary buffers: `mulres` and `t`. Will clarify this further.

crypto/fipsmodule/bn/montgomery.c
aqjune-aws: In general, all parameter sizes whose number of words is a multiple of 8 and fits in buffer. For specific sizes where num = 32 or 64 (if 64-bit, 1024bit or 2048bit), this function uses faster implementations (bignum_k{sqr,mul}_16_32, bignum_k{sqr,mul}_32_64). Will be commented in the next commit

crypto/fipsmodule/bn/montgomery.c
aqjune-aws: This is true. I think this code was silently stripping the const quantifier away. Should be fixed in the next commit

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -418,6 +450,82 @@ static int bn_mod_mul_montgomery_fallback(BIGNUM *r, const BIGNUM *a,
...
+// Additionally, montgomery_s2n_bignum_mul_mont allocates arrays at a stack, and
+// large num leads to out of bounds accesses of the arrays.
+// montgomery_use_s2n_bignum(num) must be called in advance to check these
+// conditions.
+static void montgomery_s2n_bignum_mul_mont(BN_ULONG *rp, const BN_ULONG *ap,
+                                           const BN_ULONG *bp,
+                                           const BN_ULONG *np,
+                                           const BN_ULONG *n0, size_t num) {
+
+#if defined(BN_MONTGOMERY_USE_S2N_BIGNUM)
```
nebeid, line 469: This `#if` can replace the one above, 'OPENSSL_BN_ASM_MONT', so there is only one `#if..#else` pair, since it's more constrained.

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -418,6 +450,82 @@ static int bn_mod_mul_montgomery_fallback(BIGNUM *r, const BIGNUM *a,
   return ret;
 }
 
+
+#if defined(OPENSSL_BN_ASM_MONT)
```
nebeid: Sorry, it turns out you need this `#if` for the NO_ASM builds on x86.

aqjune-aws:
From the CI failure, it seems in ARM the s2n-bignum assembly files are not linked to executables if `MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX` is set.

Relevant CMakeLists.txt lines: https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/CMakeLists.txt#L170-L171

Should we update the CMakeLists.txt lines in this PR?

nebeid (dismissed):

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -418,6 +447,82 @@ static int bn_mod_mul_montgomery_fallback(BIGNUM *r, const BIGNUM *a,
...
+  //    The comparison is true if either A or B holds:
+  //    A. The result of step 1 >= 2^(64*num), meaning that bignum_emontredc_8n
+  //       returned 1. Since m is less than 2^(64*num), (result of step 1) >= m holds.
+  //    B. The result of step 1 fits in 2^(64*num), and the result >= m.
+  uint64_t c;
+  c = bignum_emontredc_8n(num, mulres, np, w); // c: case A
+  c |= bignum_ge(num, mulres + num, num, np);  // c: case B
+  // Optionally subtract and store the result at rp
+  bignum_optsub(num, rp, mulres + num, c, np);
+
```
dkostic, line 514: is there a reason why the full montgomery reduction is not offered by s2n-bignum as a single function?

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -418,6 +447,82 @@ static int bn_mod_mul_montgomery_fallback(BIGNUM *r, const BIGNUM *a,
...
+static void montgomery_s2n_bignum_mul_mont(BN_ULONG *rp, const BN_ULONG *ap,
+                                           const BN_ULONG *bp,
+                                           const BN_ULONG *np,
+                                           const BN_ULONG *n0, size_t num) {
+
+#if defined(BN_MONTGOMERY_USE_S2N_BIGNUM)
+
+  // t is a temporary buffer used by big-int multiplication.
+  // bignum_kmul_32_64 requires 96 words.
+  uint64_t t[96];
```
dkostic, line 470: if `96` is a constant related to the internals of s2n-bignum than I think it should be defined as a macro in an s2n-bignum header file. Otherwise, s2n-bignum internals could change and we'll forget to change the appropriate values in aws-lc.

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -119,8 +119,37 @@
...
+
+#define BN_MONTGOMERY_USE_S2N_BIGNUM 1
+
+OPENSSL_INLINE int montgomery_use_s2n_bignum(unsigned int num) {
+  // Use s2n-bignum's functions only if
+  // (1) The ARM architecture has slow multipliers, and
+  // (2) num (which is the number of words) is multiplie of 8, because
+  //     s2n-bignum's bignum_emontredc_8n requires it
+  // (3) The word size is 64 bits, and
+  // (4) Temporary buffer's size (t and mulres) used in
```
dkostic, line 139: what I still don't understand is if this is a limitation of s2n-bignum or aws-lc?

crypto/fipsmodule/bn/montgomery.c
jargh: There is the generic function bignum_montredc that does full Montgomery reduction (and indeed bignum_montmul doing full Montgomery multiplication), but that is not intended to be used for the performance-critical cases like this.

The argument for breaking this primitive out is that emontredc can also be a building-block for doing "almost-Montgomery" where you would do the subtraction only if you exceed the 2^64N size (that comparison is already the return value from the emontredc function). I think the whole modexp operation would be slightly faster maintaining this weaker normalization through the tower of multiplications (and just reducing strictly one time at the very end), but because we are trying to be plug-compatible at the level of BN_mul_montgomery, we do need to do the full reduction here every time, hence the additional bignum_ge.

crypto/fipsmodule/bn/montgomery.c
aqjune-aws: Addressed it in the latest commit. For each Karatsuba multiplication function (16x16 and 32x32 square/mults), a macro designating the temporary buffer word size is defined and used. To avoid usage of `max()` (which actually doesn't seem to exist in `stdlib.h` or `math.h`..? C is a weird language) an assertion that checks `S2NBIGNUM_KMUL_32_64_TEMP_NWORDS` is the largest one is added to this checker function

crypto/fipsmodule/bn/montgomery.c
aqjune-aws: I think the necessity of buffer comes from how the montgomery multiplication is being done in this performance improvement.
First we calculate the NxN->2N big int multiplication using Karatsuba multiplication, and then reduce 2N to N using montgomery reduction.
In order to do this, we need 2N-word temporary buffer.
Also, Karatsuba multiplication itself requires storing its intermediate result in a buffer, adding more usage.

crypto/fipsmodule/bn/montgomery.c
aqjune-aws: In addition to @jargh, I somewhat thought that the function invocations were straightforward since it was following the steps of montgomery reduction.. It was my thought though.

aqjune-aws: Yes. `bignum_emontredc_8n` relies on the assumption that the input big integer is a multiple of 8 words for faster calculation. Specifically, it splits the input big integer into large chunks (4-words - which is 256 bits integers - that is a half of 8-words) and does fast multiplication on 256x256->512 bits, which internally uses Karatsuba multiplication. For Graviton 2, which has slow multiplier, using Karatsuba at this level brings significant performance benefit. :)

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -119,8 +119,42 @@
...
+  //     montgomery_s2n_bignum_mul_mont does not exceed
+  //     BN_MONTGOMERY_MAX_WORDS.
+  assert(S2NBIGNUM_KSQR_16_32_TEMP_NWORDS <= S2NBIGNUM_KMUL_32_64_TEMP_NWORDS &&
+         S2NBIGNUM_KSQR_32_64_TEMP_NWORDS <= S2NBIGNUM_KMUL_32_64_TEMP_NWORDS &&
+         S2NBIGNUM_KMUL_16_32_TEMP_NWORDS <= S2NBIGNUM_KMUL_32_64_TEMP_NWORDS);
+  const uint64_t temp_buffer_nwords =
+      S2NBIGNUM_KMUL_32_64_TEMP_NWORDS + 2 * (uint64_t)num;
+  return !CRYPTO_is_ARMv8_wide_multiplier_capable() && (num % 8 == 0) &&
+         BN_BITS2 == 64 && temp_buffer_nwords <= BN_MONTGOMERY_MAX_WORDS;
```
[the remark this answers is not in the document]
aqjune-aws: Makes sense! From `include/openssl.base.h` (which defines `OPENSSL_AARCH64`) I can see that `OPENSSL_64_BIT` and `OPENSSL_AARCH64` are defined together, and if `OPENSSL_64_BIT` is defined, `BN_BITS2` must be 64. I will leave an assertion checking `BN_BITS2==64` here.

crypto/fipsmodule/bn/montgomery.c
dkostic: > s2n-bignum may use slightly more buffer
but why can't we just allocate large enough `t` and `mulres`? Are `s2n-bignum` functions that are used here writing to any other buffer?

crypto/fipsmodule/bn/montgomery.c
```diff
@@ -119,8 +119,43 @@
...
+#include "../cpucap/internal.h"
 #include "../../internal.h"
 
+#if !defined(OPENSSL_NO_ASM) &&                                                \
+    (defined(OPENSSL_LINUX) || defined(OPENSSL_APPLE)) &&                      \
+    defined(OPENSSL_AARCH64) && defined(OPENSSL_BN_ASM_MONT)
+
+#include "../../../third_party/s2n-bignum/include/s2n-bignum_aws-lc.h"
+
+#define BN_MONTGOMERY_USE_S2N_BIGNUM 1
```
torben-hansen, line 131: It's a nit. But this is a capability definition. It doesn't mean it's actually used a run-time. Hence the naming is misleading.

crypto/fipsmodule/bn/montgomery.c
aqjune-aws: There is a minor issue about the direction (even though I am aware that it is desirable to make things 'more correct') - it is that s2n-bignum does not (yet) support formal verification of linking of code. In this case linking is necessary because the three functions are defined in all different assembly files.
The reason why linking is unsupported is that (IIUC), linking requires the 'relocation' step which does not exist in s2n-bignum's assembly decoder yet. Adding this might be a fairly complicated process.

One possible workaround is to make a giant assembly file that copy-and-pastes the three assembly functions used in here. But this might require a really significant engineering effort. Proofs for these three functions are pretty large, and simply concatenating these proofs won't work in general due to... how it is written. Maybe this is why writing top-level, verified functions for elliptic curves is a big goal for this year as well (to avoid misinterpretation, I wanted to emphasize that verifying combined functions is non-trivial).

Of course we can put an unverified code that has the three function calls in s2n-bignum, but since s2n-bignum's policy is to always have a pair of code and proof, not sure whether this is a good direction too....

crypto/fipsmodule/bn/montgomery.c
aqjune-aws: Would `BN_MONTGOMERY_S2N_BIGNUM_CAPABLE` be ok? Then I will apply it in the next commit.

crypto/fipsmodule/bn/montgomery.c
dkostic: I see. I’m ok with the compromise to get this code in as is with the expectation that s2n-bignum in the future will provide a single function call to do montgomery reduction (or even the full mul+redc).

Discussed with June offline, we agreed this is ok.

crypto/curve25519/curve25519.c
```diff
@@ -31,9 +31,13 @@
 #include "../internal.h"
 #include "../fipsmodule/cpucap/internal.h"
 
-#if (defined(OPENSSL_X86_64) || defined(OPENSSL_AARCH64)) && \
-    (defined(OPENSSL_LINUX) || defined(OPENSSL_APPLE)) && \
-    !defined(OPENSSL_NO_ASM) && !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
+// If (1) x86 or aarch64, (2) linux or apple, and (3) OPENSSL_NO_ASM is not
```
dkostic, line 34:
```suggestion
// If (1) x86_64 or aarch64, (2) linux or apple, and (3) OPENSSL_NO_ASM is not
```

crypto/fipsmodule/ec/p384.c
```diff
@@ -32,10 +32,12 @@
 //   #define p384_felem_add(out, in0, in1) bignum_add_p384(out, in0, in1)
 // when s2n-bignum is used.
 //
-#if !defined(OPENSSL_NO_ASM) && \
-    (defined(OPENSSL_LINUX) || defined(OPENSSL_APPLE)) && \
-    (defined(OPENSSL_X86_64) || defined(OPENSSL_AARCH64)) && \
-    !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
+// If (1) x86 or aarch64, (2) linux or apple, and (3) OPENSSL_NO_ASM is not
```
dkostic, line 35:
```suggestion
// If (1) x86_64 or aarch64, (2) linux or apple, and (3) OPENSSL_NO_ASM is not
```

crypto/fipsmodule/ec/p521.c
```diff
@@ -33,10 +33,12 @@
 // when Fiat-crypto is used, or as:
 //   #define p521_felem_add(out, in0, in1) bignum_add_p521(out, in0, in1)
 // when s2n-bignum is used.
-#if !defined(OPENSSL_NO_ASM) && \
-    (defined(OPENSSL_LINUX) ||  defined(OPENSSL_APPLE)) && \
-    (defined(OPENSSL_X86_64) || defined(OPENSSL_AARCH64)) && \
-    !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
+// If (1) x86 or aarch64, (2) linux or apple, and (3) OPENSSL_NO_ASM is not
```
dkostic, line 36:
```suggestion
// If (1) x86_64 or aarch64, (2) linux or apple, and (3) OPENSSL_NO_ASM is not
```

dkostic (approved):

nebeid (approved):

PR #1140, opened 2023-08-08
Avoid UB in test shim

torben-hansen:
### Issues:

V980180360

### Description of changes: 

`NULL` arguments to standard library functions is generally undefined behaviour. This is test code, but prevent it anyway.

One such occurrence is if `SSL_get_signature_algorithm_name( SSL_get_peer_signature_algorithm(ssl), 0)` ends up returning `NULL`:
```
CheckListContains("sigalg", SSL_get_all_signature_algorithm_names,  SSL_get_signature_algorithm_name( SSL_get_peer_signature_algorithm(ssl), 0))
```

### DIFF
--- a/ssl/test/bssl_shim.cc
+++ b/ssl/test/bssl_shim.cc
@@ -305,6 +305,9 @@ static bool CheckListContains(const char *type,
   std::vector<const char *> list(list_func(nullptr, 0));
   list_func(list.data(), list.size());
   for (const char *expected : list) {
+    if (str == nullptr) {
+      break;
+    }
     if (strcmp(expected, str) == 0) {
       return true;
     }


ssl/test/bssl_shim.cc
```diff
@@ -305,6 +305,9 @@ static bool CheckListContains(const char *type,
   std::vector<const char *> list(list_func(nullptr, 0));
   list_func(list.data(), list.size());
   for (const char *expected : list) {
+    if (str == nullptr) {
```
dkostic, line 308: should we also check `expected`?

torben-hansen: The list can't contain `NULL` because `list_func` callback just returns some static list of names. basically.

If you insist, I could.

dkostic (approved):

andrewhop (approved):

PR #1141, opened 2023-08-09
Add more ssl runner tests for multiple certificate support

samuel40791765:
### Issues:
Resolves `CryptoAlg-1844`

### Description of changes: 
Add additional ssl runner tests to test our support for multiple certificate slots.
* Tests for multiple certificates slots in certificate callback
* Failure tests when the correct certificate isn't configured
* Failure tests for the client side

### Call-outs:
N/A

### Testing:
New SSL runner tests


ssl/test/runner/runner.go
```diff
@@ -19692,8 +19704,41 @@ func addMultipleCertSlotTests() {
 				strictAlgTest.config.CipherSuites = []uint16{alg.cipher}
 			}
 
+			// Extend the original test, but force it to use the early certificate callback instead.
+			// Expected behavior should be the same.
+			strictAlgCallbackTest := strictAlgTest
+			strictAlgCallbackTest.name = prefix + "Multiple-Cert-Strict-Alg-Callback" + suffix
+			strictAlgCallbackTest.flags = append(strictAlgCallbackTest.flags, "-use-early-callback")
```
andrewhop, line 19711: What is this testing exactly? What happens when the callback and multiple certs are configured?

ssl/test/runner/runner.go
```diff
@@ -19704,27 +19749,94 @@ func addMultipleCertSlotTests() {
...
 				peerSignatureAlgorithm: signatureEd25519,
 			},
 		}
 
+		// Below tests verify that multiple certificates only works because we have a valid cert.
+
+		// Connection shouldn't be accepted without RSA cert.
+		noRSACertTest := testCase{
+			testType: serverTest,
+			name:     "Server-Multiple-Cert-No-RSA" + suffix,
```
andrewhop, line 19771: Why doesn't this show up in the transfer test case name?

samuel40791765: I'm speculating that the transfer test case names only include test cases that succeed. These tests are testing for specific failures. We also only include tests on the server side.
https://github.com/aws/aws-lc/blob/c516faf2bf3f788ff842317e779fe13cf901acfd/ssl/test/runner/runner.go#L2052-L2059

ssl/test/runner/runner.go
samuel40791765: This is testing when the certificates are set via the early certificate callback, rather than setting up everything in `SSL_CTX/SSL` before the handshake is initiated. The expected behavior should be the exact same, it's just setting the certs beforehand VS during the callback.

ssl/test/runner/runner.go
```diff
@@ -19618,17 +19618,31 @@ var testMultipleCertSlotsAlgorithms = []struct {
...
+		multipleCertsFlag, path.Join(*resourceDir, getShimCertificate(testCertECDSAP256)) + "," + path.Join(*resourceDir, getShimKey(testCertECDSAP256)),
+	}
+	ed25519CertSlot := []string{
+		multipleCertsFlag, path.Join(*resourceDir, getShimCertificate(testCertEd25519)) + "," + path.Join(*resourceDir, getShimKey(testCertEd25519)),
+	}
+	certificateSlotFlags := []string{
+		rsaCertSlot[0], rsaCertSlot[1],
+		ecdsaCertSlot[0], ecdsaCertSlot[1],
+		ed25519CertSlot[0], ed25519CertSlot[1],
+	}
```
skmcgrail, line 19642:
```suggestion
	certificateSlotFlags := append([]string{}, rsaCertSlot...)
	certificateSlotFlags = append(certificateSlotFlags, ecdsaCertSlot...)
	certificateSlotFlags = append(certificateSlotFlags, ed25519CertSlot...)
```

skmcgrail (approved):

andrewhop (approved):

PR #1143, opened 2023-08-10
add integration CI dimension for nginx

samuel40791765:
### Issues:
Resolves `CryptoAlg-1961`

### Description of changes: 
We've recently ironed out some of the gaps for nginx support. We're not 100% in line with what OpenSSL supports, but we're satisfied with what we have now. 
The remaining gaps are:
* Stateful Session Resumption (Session Caches)
* SSL_Conf commands

This adds a CI dimension to verify our newfound support.

### Call-outs:
We'll be looking to contribute this patch to nginx some time soon.

### Testing:
New CI dimension

### DIFF
--- a/tests/ci/cdk/cdk/codebuild/github_ci_integration_omnibus.yaml
+++ b/tests/ci/cdk/cdk/codebuild/github_ci_integration_omnibus.yaml
@@ -78,12 +78,22 @@ batch:
         variables:
           AWS_LC_CI_TARGET: "tests/ci/integration/run_curl_integration.sh"
 
-    - identifier: nginx_integration
+    - identifier: nginx_integration_x86
       buildspec: tests/ci/codebuild/common/run_nonroot_target.yml
       env:
         type: LINUX_CONTAINER
         privileged-mode: false
         compute-type: BUILD_GENERAL1_MEDIUM
         image: 620771051181.dkr.ecr.us-west-2.amazonaws.com/aws-lc-docker-images-linux-x86:ubuntu-22.04_gcc-12x_latest
+        variables:
+          AWS_LC_CI_TARGET: "tests/ci/integration/run_nginx_integration.sh"
+
+    - identifier: nginx_integration_aarch
+      buildspec: tests/ci/codebuild/common/run_nonroot_target.yml
+      env:
+        type: ARM_CONTAINER
+        privileged-mode: false
+        compute-type: BUILD_GENERAL1_LARGE
+        image: 620771051181.dkr.ecr.us-west-2.amazonaws.com/aws-lc-docker-images-linux-aarch:ubuntu-22.04_gcc-12x_latest
         variables:
           AWS_LC_CI_TARGET: "tests/ci/integration/run_nginx_integration.sh"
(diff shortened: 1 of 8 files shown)


tests/ci/integration/run_nginx_integration.sh
```diff
@@ -0,0 +1,93 @@
...
+    --with-stream \
+    --with-stream_realip_module \
+    --with-stream_ssl_module \
+    --with-stream_ssl_preread_module \
+    --with-mail \
+    --with-mail_ssl_module \
+    --with-cc-opt="-I${AWS_LC_INSTALL_FOLDER}/include" \
+    --with-ld-opt="-L${AWS_LC_INSTALL_FOLDER}/lib"
+  make -j
+  make -j install
```
andrewhop, line 46: I assume the install target depends on the nginx executable target so `make install` will build everything then install it. Much like you can just run `ninja run_tests` for aws-lc: ninja sees if all of the `run_tests` dependencies are up to date and then does the right thing.
```suggestion
  make -j install
```

tests/ci/cdk/cdk/codebuild/github_ci_integration_omnibus.yaml
```diff
@@ -74,3 +77,13 @@ batch:
         image: 620771051181.dkr.ecr.us-west-2.amazonaws.com/aws-lc-docker-images-linux-x86:ubuntu-22.04_gcc-12x_latest
         variables:
           AWS_LC_CI_TARGET: "tests/ci/integration/run_curl_integration.sh"
+
+    - identifier: nginx_integration
+      buildspec: tests/ci/codebuild/common/run_nonroot_target.yml
+      env:
+        type: LINUX_CONTAINER
+        privileged-mode: false
+        compute-type: BUILD_GENERAL1_MEDIUM
```
andrewhop, line 86: What about an ARM build? How long does the overall build and test take?

tests/ci/codebuild/common/run_nonroot_target.yml
```diff
@@ -11,11 +11,11 @@ phases:
   install:
     run-as: root
     commands:
-      # Let postgres user in docker image take ownership of codebuild artifacts.
+      # Let non-root user in docker image take ownership of codebuild artifacts.
       - chown -R postgres:postgres /codebuild/output
```
andrewhop, line 15: What's up with this change? It's still giving ownership to the postgres user.

samuel40791765: I'm just trying to clarify that postgres is the non-root user here. I tried using changing the user name to a better name, but that'll result in issues where existing scripts on main still expect `postgres`.

tests/ci/integration/run_nginx_integration.sh
samuel40791765: Sure, I just threw the install on there and forgot to prune the commands needed. `make install` was needed so that `nginx-tests` can find the right binary to test against.

tests/ci/cdk/cdk/codebuild/github_ci_integration_omnibus.yaml
samuel40791765: I can give it a shot. The overall build+test takes 10 minutes.

samuel40791765: Added :)

andrewhop (approved):

skmcgrail (approved):

PR #1152, opened 2023-08-14
Upstream merge 2023 08 09

justsmth:
### Description of changes: 
Merge upstream changes from https://github.com/google/boringssl/commit/706846d7a84c3ce8d089b7089c298bd7449227b3 to https://github.com/google/boringssl/commit/8c7e925b5dab1f826f08f38e4b9e1543b8413476

### Call-outs:
See internal document.

### DIFF
--- a/crypto/fipsmodule/rand/rand.c
+++ b/crypto/fipsmodule/rand/rand.c
@@ -52,14 +52,14 @@
 // This might be a bit of a leap of faith, esp on Windows, but there's nothing
 // that we can do about it.)
 
-// When in FIPS mode we use the CPU Jitter entropy source to seed our DRBG.  
+// When in FIPS mode we use the CPU Jitter entropy source to seed our DRBG.
 // This entropy source is very slow and can incur a cost anywhere between 10-60ms
-// depending on configuration and CPU.  Increasing to 2^24 will amortize the 
-// penalty over more requests.  This is the same value used in OpenSSL 3.0  
+// depending on configuration and CPU.  Increasing to 2^24 will amortize the
+// penalty over more requests.  This is the same value used in OpenSSL 3.0
 // and meets the requirements defined in SP 800-90B for a max reseed of interval (2^48)
 //
 // CPU Jitter:  https://www.chronox.de/jent/doc/CPU-Jitter-NPTRNG.html
-// 
+//
 // kReseedInterval is the number of generate calls made to CTR-DRBG before
 // reseeding.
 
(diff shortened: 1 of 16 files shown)


crypto/fipsmodule/rand/rand.c
```diff
@@ -94,6 +94,12 @@ struct rand_thread_state {
   // calls is the number of generate calls made on |drbg| since it was last
   // (re)seeded. This is bound by |kReseedInterval|.
   unsigned calls;
+  // last_block_valid is non-zero iff |last_block| contains data from
+  // |get_seed_entropy|.
+  int last_block_valid;
```
dkostic, line 99: this field was removed in https://github.com/aws/aws-lc/pull/1112/

torben-hansen, line 99: unused

util/fipstools/acvp/acvptool/subprocess/subprocess.go
```diff
@@ -107,16 +133,18 @@ func NewWithIO(cmd *exec.Cmd, in io.WriteCloser, out io.ReadCloser) *Subprocess
 		"hmacDRBG":          &drbg{"hmacDRBG", map[string]bool{"SHA-1": true, "SHA2-224": true, "SHA2-256": true, "SHA2-384": true, "SHA2-512": true}},
 		"KDF":               &kdfPrimitive{},
 		"KDA":               &hkdf{},
+		"TLS-v1.2":          &tlsKDF{},
+		"TLS-v1.3":          &tls13{},
 		"CMAC-AES":          &keyedMACPrimitive{"CMAC-AES"},
 		"RSA":               &rsa{},
 		"kdf-components":    &kdfComp{"kdf-components"},
-		"TLS-v1.2":          &kdfComp{"TLS-v1.2"},
```
billbo-yang, line 113: Looks like the acvp commit somehow stealth added support for TLS 1.2 and 1.3 for Google. The TLS 1.3 stuff is great, but we already support TLS 1.2 through some code in `kdf-components.go`.

I have a commit where I renamed their `tlskdf.go` to `kdf-components` because we wanted to support SSH KDFs (and for some reason NIST uses the same algorithm name in their ACVP specs for SSH and TLS KDFs). We also support TLS 1.0/1.1 in addition to TLS 1.2, whereas it looks like Google's implementation still only supports TLS 1.2.

All that just to say that we should probably continue using our implementation via `&kdfComp{"TLS-v1.2"}` instead of what Google has here.

crypto/fipsmodule/rand/rand.c
justsmth: Removed in latest update.

justsmth: Removed in latest update.

util/fipstools/acvp/acvptool/subprocess/subprocess.go
justsmth: Agreed. This should be resolved with latest update.

billbo-yang (approved):

torben-hansen (approved):

dkostic (approved):

PR #1153, opened 2023-08-14
Add support for BIO_FP_TEXT in file BIOs

WillChilds-Klein:
# Issues
Addresses CryptoAlg-2036

# Notes

OpenSSL defines a "close flag" `BIO_FP_TEXT` which is used by callers to indicate that file-backed BIOs should be opened in "text" mode as opposed to the default of raw "binary" mode. Of platforms currently supported by AWS-LC, this is only meaningful on windows. From linux's `fopen` [man page](https://man7.org/linux/man-pages/man3/fopen.3.html):

```
        The mode string can also include the letter 'b' either as a last
        character or as a character between the characters in any of the
        two-character strings described above.  This is strictly for
        compatibility with ISO C and has no effect; the 'b' is ignored on
        all POSIX conforming systems, including Linux.  (Other systems
        may treat text files and binary files differently, and adding the
        'b' may be a good idea if you do I/O to a binary file and expect
        that your program may be ported to non-UNIX environments.)
```

And from [Windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170#generic-text-routine-mappings):

```
t     Open in text (translated) mode. Carriage return-line feed (CR-LF)
      combinations are translated into single line feeds (LF) on input and LF
      characters are translated to CR-LF combinations on output. Also, CTRL+Z is
      interpreted as an end-of-file character on input.
b     Open in binary (untranslated) mode; translations involving
      carriage-return and line feed characters are suppressed.
```

The terminology here is a bit confusing. In text mode, "Input" here refers to reading files and "output" refers to writing files to disk. So, if an application is using conventional unixy LF line endings (`\n`) and then writes a line of text to a file, the Windows CRT will instead write a CRLF line ending (`\r\n`). When reading from a file, the CRT will translate CRLF back to LF transparently to the application. In binary mode, what you see is what you get and no translation occurs.

OpenSSL supports setting text mode for a variety of BIO operations (including [when opening files with `fopen`](https://github.com/openssl/openssl/blob/3d254b31344e82b8f10fda8bab196757a377eb63/crypto/bio/bss_file.c#L267-L290)), but only sets this flag internally [when calling `BIO_new_fp`](https://github.com/search?q=repo%3Aopenssl/openssl%20BIO_FP_TEXT&type=code). Note that `BIO_new_fp` operates on open `FILE *`s, and [does not itself open the file](https://github.com/openssl/openssl/blob/3d254b31344e82b8f10fda8bab196757a377eb63/crypto/bio/bss_file.c#L207-L266). Instead, it uses the Windows CRT-specific function `_setmode` to set binary/text mode on the already opened file.

To implement this functionality in AWS-LC, we only set translation mode in functions [using the `BIO_C_SET_FILE_PTR` control directive](https://github.com/aws/aws-lc/blob/main/crypto/bio/file.c#L191C10-L191C28): `BIO_set_fp` and `BIO_new_fp`. AWS-LC's BIO functions that call `fopen` along the way [do not provide the caller a parameter for specifying flags](https://github.com/aws/aws-lc/blob/main/crypto/bio/file.c#L287-L306), so we're limited here by the interface.

Finally, it's a little odd to refer to `BIO_FP_TEXT` as a "close flag", as the behavior it determines has nothing to do with how or if the file is closed. However, this is how OpenSSL's code refers to it, so it's the terminology we use here.

## Links

Documentation links for the Windows-specific functions used in this change can be found here:

- [`_setmode`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setmode?view=msvc-170): used to set translation mode on an open file
- [`_fileno`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fileno?view=msvc-170): used to convert a `FILE *` to a file descriptor

# Testing
- CI runs with new platform-aware test

---

### DIFF
--- a/crypto/bio/bio_test.cc
+++ b/crypto/bio/bio_test.cc
@@ -169,34 +169,26 @@ TEST(BIOTest, CloseFlags) {
 
   const char *test_str = "test\ntest\ntest\n";
 
-  printf("FOOBAR 1\n");
-
   // Assert that CRLF line endings get inserted on write and translated back out on
   // read for text mode.
   TempFILE text_bio_file(tmpfile(), fclose);
   ASSERT_TRUE(text_bio_file);
-  printf("FOOBAR 2\n");
   bssl::UniquePtr<BIO> text_bio(BIO_new_fp(text_bio_file.get(), BIO_NOCLOSE | BIO_FP_TEXT));
   int bytes_written = BIO_write(text_bio.get(), test_str, strlen(test_str));
   EXPECT_GE(bytes_written, 0);
-  printf("FOOBAR 3\n");
   ASSERT_TRUE(BIO_flush(text_bio.get()));
   ASSERT_EQ(0, BIO_seek(text_bio.get(), 0));    // 0 indicates success here
   char contents[256];
-  printf("FOOBAR 4\n");
   OPENSSL_memset(contents, 0, sizeof(contents));
   int bytes_read = BIO_read(text_bio.get(), contents, sizeof(contents));
   EXPECT_GE(bytes_read, bytes_written);
   EXPECT_EQ(test_str, std::string(contents));
 
-  printf("FOOBAR 5\n");
   // Windows should have translated '\n' to '\r\n' on write, so validate that
   // by opening the file in raw binary mode (i.e. no BIO_FP_TEXT).
   bssl::UniquePtr<BIO> text_bio_raw(BIO_new_fp(text_bio_file.get(), BIO_NOCLOSE));
   ASSERT_EQ(0, BIO_seek(text_bio.get(), 0));    // 0 indicates success here
-  printf("FOOBAR 6\n");
   OPENSSL_memset(contents, 0, sizeof(contents));
-  printf("FOOBAR 7\n");
   bytes_read = BIO_read(text_bio_raw.get(), contents, sizeof(contents));
   EXPECT_GT(bytes_read, 0);
 #if defined(OPENSSL_WINDOWS)
@@ -205,18 +197,15 @@ TEST(BIOTest, CloseFlags) {
   EXPECT_EQ(test_str, std::string(contents));
 #endif
 
-  printf("FOOBAR 8\n");
   // Assert that CRLF line endings don't get inserted on write for
   // (default) binary mode.
   TempFILE binary_bio_file(tmpfile(), fclose);
   ASSERT_TRUE(binary_bio_file);
-  printf("FOOBAR 9\n");
   bssl::UniquePtr<BIO> binary_bio(BIO_new_fp(binary_bio_file.get(), BIO_NOCLOSE));
   bytes_written = BIO_write(binary_bio.get(), test_str, strlen(test_str));
   EXPECT_EQ((int) strlen(test_str), bytes_written);
   ASSERT_TRUE(BIO_flush(binary_bio.get()));
   ASSERT_EQ(0, BIO_seek(binary_bio.get(), 0));    // 0 indicates success here
-  printf("FOOBAR 10\n");
   OPENSSL_memset(contents, 0, sizeof(contents));
   bytes_read = BIO_read(binary_bio.get(), contents, sizeof(contents));
   EXPECT_GE(bytes_read, bytes_written);
@@ -228,47 +217,13 @@ TEST(BIOTest, CloseFlags) {
   long pos;
   char b1[256], b2[256];
   binary_bio.reset(BIO_new_fp(binary_bio_file.get(), BIO_NOCLOSE));
-  printf("FOOBAR 11\n");
   ASSERT_EQ(0, BIO_seek(binary_bio.get(), 0));    // 0 indicates success here
   BIO_gets(binary_bio.get(), b1, sizeof(b1));
   pos = BIO_tell(binary_bio.get());
   BIO_gets(binary_bio.get(), b1, sizeof(b1));
-  printf("FOOBAR 12\n");
   BIO_seek(binary_bio.get(), pos);
   BIO_gets(binary_bio.get(), b2, sizeof(b2));
   EXPECT_EQ(std::string(b1), std::string(b2));
-
-  // Assert that BIO_CLOSE causes the underlying file to be closed on BIO free
-  // (ftell will return < 0)
-  FILE *tmp = tmpfile();
-  ASSERT_TRUE(tmp);
-  printf("FOOBAR 13\n");
-  BIO *bio = BIO_new_fp(tmp, BIO_CLOSE);
-  EXPECT_EQ(0, BIO_tell(bio));
-  printf("FOOBAR 14\n");
-  // save off fd to avoid referencing |tmp| after free and angering valgrind
-  int tmp_fd = fileno(tmp);
-  EXPECT_LT(0, tmp_fd);
-  EXPECT_TRUE(BIO_free(bio));
-  printf("FOOBAR 15\n");
-  EXPECT_EQ(-1, lseek(tmp_fd, 0, SEEK_CUR));
-  EXPECT_EQ(errno, EBADF);  // EBADF indicates taht |BIO_free| closed the file
-
-  printf("FOOBAR 16\n");
-  // Assert that BIO_NOCLOSE does not closethe underlying file on BIO free
-  tmp = tmpfile();
-  ASSERT_TRUE(tmp);
-  printf("FOOBAR 17\n");
-  bio = BIO_new_fp(tmp, BIO_NOCLOSE);
-  EXPECT_EQ(0, BIO_tell(bio));
-  EXPECT_TRUE(BIO_free(bio));
-  printf("FOOBAR 18\n");
-  EXPECT_TRUE(tmp);
-  EXPECT_EQ(0, ftell(tmp));     // 0 indicates file is still open
-  printf("FOOBAR 19\n");
-  EXPECT_LT(0, tmp_fd);
-  EXPECT_EQ(0, fclose(tmp));    // 0 indicates success for fclose
-  printf("FOOBAR 20\n");
 }
 
 TEST(BIOTest, ReadASN1) {
--- a/crypto/bio/bio_test.cc
+++ b/crypto/bio/bio_test.cc
@@ -234,10 +234,13 @@ TEST(BIOTest, CloseFlags) {
   int tmp_fd = fileno(tmp);
   EXPECT_LT(0, tmp_fd);
   EXPECT_TRUE(BIO_free(bio));
-  // use fcntl here because Windows CRT fails an assertion when lseek or _ftell
-  // is called on a closed file descriptor
-  EXPECT_EQ(-1, fcntl(tmp_fd, F_GETFD));
+  // Windows CRT hits an assertion error and stack overflow (exception code
+  // 0xc0000409) when calling ftell or lseek on an already-closed file
+  // descriptor, so only consider the non-Windows case here.
+#if !defined(OPENSSL_WINDOWS)
+  EXPECT_EQ(-1, lseek(tmp_fd, 0, SEEK_CUR));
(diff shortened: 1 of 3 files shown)


include/openssl/bio.h
```diff
@@ -396,10 +396,15 @@ OPENSSL_EXPORT int BIO_read_asn1(BIO *bio, uint8_t **out, size_t *out_len,
...
-// flag" is passed to a BIO function.
+// |BIO_NOCLOSE|, |BIO_CLOSE|, and |BIO_FP_TEXT|  can be used as symbolic
+// arguments when a "close flag" is passed to a BIO function. |BIO_CLOSE| (on
+// by default) will free the backing buffer on BIO close, whereas |BIO_NOCLOSE|
+// will not. Setting |BIO_FP_TEXT| will cause the file to be opened as a text
+// file (this is only relevant on Windows due to CRLF endings) instead of the
+// default behavior of opening the file in binary mode.
 #define BIO_NOCLOSE 0
 #define BIO_CLOSE 1
+#define BIO_FP_TEXT 0x10
```
andrewhop, line 407: If you wanted to you could create a new section (2 blank lines, title and description followed by 1 blank lines) and then document each one:
```suggestion

// BIO close flags
// These  can be used as symbolic arguments when a "close flag" is passed to a BIO function. 

// BIO_NOCLOSE does foo
#define BIO_NOCLOSE 0
// BIO_CLOSE does bar
#define BIO_CLOSE 1
// BIO_FP_TEXT does bazz
#define BIO_FP_TEXT 0x10
```
See some of the bn.h examples https://github.com/aws/aws-lc/blob/main/include/openssl/bn.h#L866-L869

crypto/bio/bio_test.cc
```diff
@@ -152,6 +152,40 @@ TEST(BIOTest, Printf) {
...
+  FILE *text_bio_file = fmemopen(nullptr, 10, "r+");
+  ASSERT_TRUE(text_bio_file);
+  bssl::UniquePtr<BIO> text_bio(BIO_new_fp(text_bio_file, BIO_CLOSE | BIO_FP_TEXT));
+  int bytes_written = BIO_write(text_bio.get(), test_str, strlen(test_str));
+  ASSERT_GE(bytes_written, 0);
+  ASSERT_TRUE(BIO_flush(text_bio.get()));
+  ASSERT_EQ(0, BIO_seek(text_bio.get(), 0));    // 0 indicates success here
+  char contents[10];
+  int bytes_read = BIO_read(text_bio.get(), contents, sizeof(contents));
+  printf("FOOBAR :: %d :: %s\n", bytes_written, contents);
```
andrewhop, line 168: Debug message?

include/openssl/bio.h
WillChilds-Klein: good call, will do.

crypto/bio/bio_test.cc
WillChilds-Klein: yeah, i'm still debugging some windows stuff (no local test env...). I'll clean this up before publishing the PR.

crypto/bio/file.c
```diff
@@ -193,6 +198,13 @@ static long file_ctrl(BIO *b, int cmd, long num, void *ptr) {
       b->shutdown = (int)num & BIO_CLOSE;
       b->ptr = ptr;
       b->init = 1;
+#if defined(OPENSSL_WINDOWS)
+      // Windows differentiates between "text" and "binary" file modes, so set
+      // the file to text mode if caller specifies BIO_FP_TEXT flag.
+      //
+      // https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setmode?view=msvc-170#remarks
+      _setmode(_fileno(b->ptr), num & BIO_FP_TEXT ? _O_TEXT : _O_BINARY);
```
justsmth, line 206: OpenSSL also [makes a call to `setvbuf` on Windows](https://github.com/openssl/openssl/blob/84364b9dc693a30fa55c22e684b45978a5bcc77b/crypto/bio/bss_file.c#L238C1-L246C55) when `BIO_FP_TEXT` is set.  Should we do the same?
```
            /*
             * Reports show that ftell() isn't trustable in text mode.
             * This has been confirmed as a bug in the Universal C RTL, see
             * https://developercommunity.visualstudio.com/content/problem/425878/fseek-ftell-fail-in-text-mode-for-unix-style-text.html
             * The suggested work-around from Microsoft engineering is to
             * turn off buffering until the bug is resolved.
             */
            if ((num & BIO_FP_TEXT) != 0)
                setvbuf((FILE *)ptr, NULL, _IONBF, 0);
```

crypto/bio/bio_test.cc
```diff
@@ -154,6 +154,64 @@ TEST(BIOTest, Printf) {
...
+  // On Android, when running from an APK, |tmpfile| does not work. See
+  // b/36991167#comment8.
+  GTEST_SKIP();
+#endif
+
+
+  // unique_ptr will automatically call fclose on the file descriptior when the
+  // variable goes out of scope, so we need to specify BIO_NOCLOSE close flags
+  // to avoid a double-free condition.
+  using TempFILE = std::unique_ptr<FILE, decltype(&fclose)>;
```
justsmth, line 168: I don't see any tests here that utilize `BIO_CLOSE` to verify that it does close the file when specified.

WillChilds-Klein: Huh, [searching around for `BIO_CLOSE`](https://github.com/search?q=repo%3Aaws%2Faws-lc+BIO_CLOSE&type=code) it looks like we don't have any direct test coverage on that flag being set explicitly, even though it's [documented as the default](https://github.com/aws/aws-lc/blob/aa90e509f2e940916fbe9fdd469a4c90c51824f6/include/openssl/bio.h#L392). I'll add a test case for this.

include/openssl/bio.h
```diff
@@ -452,6 +447,23 @@ OPENSSL_EXPORT int BIO_set_mem_buf(BIO *bio, BUF_MEM *b, int take_ownership);
...
+// BIO_NOCLOSE will not free the backing buffer on BIO close
+#define BIO_NOCLOSE 0
+
+// BIO_CLOSE (default) will free the backing buffer on BIO close, whereas
+#define BIO_CLOSE 1
+
+// BIO_FP_TEXT will cause the file to be opened as a text file (this is only
+// relevant on Windows due to CRLF endings) instead of the default behavior of
+// opening the file in binary mode.
+#define BIO_FP_TEXT 0x10
```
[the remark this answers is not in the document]
WillChilds-Klein: This is a good question. This shouldn't matter if all we're doing is calling `_setmode` after open, as windows `fopen` [considers](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170#generic-text-routine-mappings) "text" mode to be the default and will honor `b` when passed to `fopen`. however, if we're doing more than `_setmode`, I think we need to perform this check-and-set-`BIO_FP_TEXT`.

If the fseek/ftell tests I write indicate that  we need to call `setvbuf`, then I'll implement this. Else, I don't think we need it.

crypto/bio/file.c
WillChilds-Klein: Another good question. I'll write some test cases to determine whether it's needed.

crypto/bio/bio_test.cc
justsmth: Calling `BIO_CLOSE` the "default" could be misleading. If you invoke `BIO_new_fp(text_bio_file.get(), BIO_FP_TEXT);`, there's no way for [`BIO_new_fp`](https://github.com/aws/aws-lc/blob/aa90e509f2e940916fbe9fdd469a4c90c51824f6/crypto/bio/file.c#L118) to know that you might expect the "default" to provide the `BIO_CLOSE` flag behavior.

include/openssl/bio.h
```diff
@@ -452,6 +447,23 @@ OPENSSL_EXPORT int BIO_set_mem_buf(BIO *bio, BUF_MEM *b, int take_ownership);
...
+// BIO close flags.
+//
+// These can be used as symbolic arguments when a "close flag" is passed to a
+// BIO function.
+
+// BIO_NOCLOSE will not free the backing buffer on BIO close
+#define BIO_NOCLOSE 0
+
+// BIO_CLOSE (default) will free the backing buffer on BIO close, whereas
+#define BIO_CLOSE 1
```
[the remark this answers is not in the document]
WillChilds-Klein: will fix.

include/openssl/bio.h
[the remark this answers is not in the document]
WillChilds-Klein: The test I added isn't able to reproduce the issue, so I don't think we need the `setvmode` and I'd like to avoid porting over OpenSSL's legacy cruft for dealing with fixed bugs.

To your initial comment, I guess the question is whether we want to pass through to the default behavior in this case (i.e. windows treating `mode` str w/o `'b'` set as "text mode")? I believe that this default behavior will need to remain stable through time as by now, surely many windows applications have come to rely on it. That said, it probably wouldn't _hurt_ to be explicit and set `BIO_FP_TEXT` internally in this case. What do you think?

crypto/bio/bio_test.cc
WillChilds-Klein: Hm, fair point. I'll see if I can reword this.

crypto/bio/file.c
WillChilds-Klein: Based on the reproduction test I've written (specifically, that it passes without the call to `setvbuf` on our windows CI), I don't think we need this call. I'd like to avoid porting over OpenSSL's legacy cruft for dealing with fixed bugs.

justsmth (dismissed):
Not part of this PR, but I noticed that `BIO_get_fp` also doesn't have much test coverage.

include/openssl/bio.h
```diff
@@ -452,6 +447,23 @@ OPENSSL_EXPORT int BIO_set_mem_buf(BIO *bio, BUF_MEM *b, int take_ownership);
...
+// BIO_NOCLOSE will not close the underlying file on BIO free
+#define BIO_NOCLOSE 0
+
+// BIO_CLOSE will close the underlying file on BIO free
+#define BIO_CLOSE 1
+
+// BIO_FP_TEXT will cause the file to be opened as a text file (this is only
+// relevant on Windows due to CRLF endings) instead of the default behavior of
+// opening the file in binary mode.
+#define BIO_FP_TEXT 0x10
```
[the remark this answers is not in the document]
WillChilds-Klein: > Is there any situation where this flag is in effect when the file is "opened"? (The documentation here only addresses its use when opening a file.) I see it can be passed as part of close_flag for BIO_set_fp and BIO_new_fp, but in those cases the file has already been "opened", we just potentially change its mode.

I think my language is imprecise here. No file is ever actually `open`ed with this flag; it's set after open (although some functions immediately set it after opening before returning to the caller). The functions available to callers that support this flag are:

- `BIO_new_file`
- `BIO_new_fp`
- `BIO_set_fp`

I'll update the doc comments of those functions to reflect this and modify this doc comment to be more precise.

> Should this flag also be called out in the documentation for BIO_set_fp and BIO_new_fp?

Yes.

justsmth (approved):

justsmth (dismissed):

WillChilds-Klein:
@andrewhop had a helpful suggestion to try [`_get_osfhandle`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/get-osfhandle?view=msvc-170), but unfortunately that still appears to trigger the "stack overflow" (reason for scare quotes [here](https://devblogs.microsoft.com/oldnewthing/20190108-00/?p=100655)):

```
[----------] 1 test from BIOTest
[ RUN      ] BIOTest.CloseFlags
crypto/crypto_test failed to complete: exit status 0xc0000409
```

andrewhop (approved):

justsmth (approved):

PR #1155, opened 2023-08-15
add missing symbols for SSLProxy support

samuel40791765:
### Issues:
Resolves `CryptoAlg-1957`

### Description of changes: 
This pulls in the missing symbols needed to integrate with SSLProxy.
1. I've followed the same pattern with `X509V3_EXT_conf_nid` and added the `X509V3_EXT_conf` version here as well. The goal is to only maintain it for basic compatibility and push users to use `X509V3_EXT_nconf*` instead.
2. Added `SSL_CTX_set_security_level` and `CONF_modules_finish` as no-ops. We don't support security levels and adding any configurations to do so should be it's own story.
3. Added tests and functionality for `SSL_SESSION_print`. Taken from: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/ssl/ssl_txt.c#L32-L162

### Call-outs:
N/A

### Testing:
New unit test

### DIFF
--- a/include/openssl/x509v3.h
+++ b/include/openssl/x509v3.h
@@ -632,13 +632,23 @@ OPENSSL_EXPORT X509_EXTENSION *X509V3_EXT_nconf_nid(const CONF *conf,
 
 // X509V3_EXT_conf_nid calls |X509V3_EXT_nconf_nid|. |conf| must be NULL.
 //
-// TODO(davidben): This is the only exposed instance of an LHASH in our public
-// headers. cryptography.io wraps this function so we cannot, yet, replace the
-// type with a dummy struct.
+// TODO(davidben): This and |X509V3_EXT_conf| are the only exposed instance of
+// |LHASH| in our public headers. cryptography.io wraps this function so we
+// cannot replace the type with a dummy struct.
 OPENSSL_EXPORT X509_EXTENSION *X509V3_EXT_conf_nid(LHASH_OF(CONF_VALUE) *conf,
                                                    const X509V3_CTX *ctx,
                                                    int ext_nid,
                                                    const char *value);
+
+// X509V3_EXT_conf calls |X509V3_EXT_nconf|. |conf| must be NULL.
+//
+// NOTE: This is only provided for compatibility. See |X509V3_EXT_nconf|
+// instead.
+OPENSSL_EXPORT X509_EXTENSION *X509V3_EXT_conf(LHASH_OF(CONF_VALUE) *conf,
+                                               X509V3_CTX *ctx,
+                                               const char *name,
+                                               const char *value);
+
 
 // X509V3_EXT_add_nconf_sk looks up the section named |section| in |conf|. For
 // each |CONF_VALUE| in the section, it constructs an extension as in
--- a/ssl/ssl_text.cc
+++ b/ssl/ssl_text.cc
@@ -1,0 +1,139 @@
+/*
+ * Copyright 1995-2022 The OpenSSL Project Authors. All Rights Reserved.
+ * Copyright 2005 Nokia. All rights reserved.
+ */
+
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+// Modifications Copyright Amazon.com, Inc. or its affiliates.
+
+
+#include <openssl/ssl.h>
+
+#include "internal.h"
+
+
+int SSL_SESSION_print(BIO *bp, const SSL_SESSION *sess) {
+  if (sess == nullptr) {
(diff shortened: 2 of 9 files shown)


ssl/ssl_text.cc
```diff
@@ -0,0 +1,139 @@
...
+  if (BIO_puts(bp, "SSL-Session:\n") <= 0) {
+    return 0;
+  }
+  if (BIO_printf(bp, "    Protocol  : %s\n", SSL_SESSION_get_version(sess)) <=
+      0) {
+    return 0;
+  }
+
+  if (sess->cipher == nullptr) {
+    if (((sess->cipher->id) & 0xff000000) == 0x02000000) {
```
justsmth, line 30: * You verify that `sess->cipher == nullptr` then immediately dereference `sess->cipher->id`?
* Why `0x02000000`? What does that imply about the value?  (The [comment here](https://github.com/aws/aws-lc/blob/4f852cd6f2ef9ee638013636dfb54fa1482d8f62/ssl/ssl_cipher.cc#L1452) says "All OpenSSL cipher IDs are prefaced with 0x03"?)

ssl/ssl_text.cc
```diff
@@ -0,0 +1,139 @@
...
+
+
+int SSL_SESSION_print(BIO *bp, const SSL_SESSION *sess) {
+  if (sess == nullptr) {
+    return 0;
+  }
+
+  bool tls13 = (sess->ssl_version == TLS1_3_VERSION);
+  if (BIO_puts(bp, "SSL-Session:\n") <= 0) {
+    return 0;
```
justsmth, line 22: Should any of these `return 0` lines also have a `OPENSSL_PUT_ERROR`?

justsmth (changes requested):

ssl/ssl_text.cc
samuel40791765: I don't think this necessarily needs an error since it's a print function for debugging purposes. The original OpenSSL function doesn't push any errors onto the stack either.
* https://github.com/openssl/openssl/blob/3d254b31344e82b8f10fda8bab196757a377eb63/ssl/ssl_txt.c#L32

ssl/ssl_text.cc
samuel40791765: Yeah my bad, I think we can take this out. I took from the original OpenSSL code and hadn't noticed that `sess->cipher->id` was under `sess->cipher`.

All of our available ciphers have names, so the `cipher->id` doesn't really make sense in AWS-LC: https://github.com/aws/aws-lc/blob/4f852cd6f2ef9ee638013636dfb54fa1482d8f62/ssl/ssl_cipher.cc#L158

justsmth (approved):

nebeid:
> 3\. Added tests and functionality for `SSL_SESSION_print`.

I suggest you add a reference to the origin of this code in the PR description.

include/openssl/x509v3.h
```diff
@@ -632,14 +632,24 @@ OPENSSL_EXPORT X509_EXTENSION *X509V3_EXT_nconf_nid(const CONF *conf,
 
 // X509V3_EXT_conf_nid calls |X509V3_EXT_nconf_nid|. |conf| must be NULL.
 //
-// TODO(davidben): This is the only exposed instance of an LHASH in our public
-// headers. cryptography.io wraps this function so we cannot, yet, replace the
-// type with a dummy struct.
+// TODO(davidben): This and |X509V3_EXT_conf| are the only exposed instance of
```
nebeid, line 635: Is this valid if we assert that `conf` is NULL in `X509V3_EXT_nconf`?

samuel40791765: ahh this was a typo on my side, it should be `X509V3_EXT_conf`

samuel40791765 (commented):
> I suggest you add a reference to the origin of this code in the PR description.

Added!

include/openssl/x509v3.h
samuel40791765: I think what this comment's trying to say is that `X509V3_EXT_conf_nid` is the only function that exposes the `LHASH_OF(*)` type in it's argument. They wanted to deprecate this structure, but ran into libraries that were still consuming it. 

I updated this comment to include `X509V3_EXT_conf` since there are now two functions exposing `LHASH_OF(*)`. Both `X509V3_EXT_conf_nid` and `X509V3_EXT_conf` do the `NULL` assertion for `conf`.

nebeid (approved):

PR #1156, opened 2023-08-15
Fix build on gcc 4.8.3

samuel40791765:
### Description of changes: 
We recently had a team reach out to us about our broken build on gcc 4.8.3. This is due to a bug that we've previously encountered in this commit: https://github.com/aws/aws-lc/commit/65661035692572f9e399a162a4e908ad5b4bd509

### Call-outs:
N/A

### Testing:
We may have to add new CI to check for this, but it seems like this specific version is hard to find. We already have CI for gcc4.8.5.

### DIFF
--- a/tool/speed.cc
+++ b/tool/speed.cc
@@ -21,7 +21,12 @@
 
 #include <assert.h>
 #include <errno.h>
+
+#ifndef __STDC_FORMAT_MACROS
+#define __STDC_FORMAT_MACROS
+#endif
 #include <inttypes.h>
+
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>


justsmth:
I found a couple of other place we use `intypes.h` without this define.  Should we add it to these as well?
* https://github.com/aws/aws-lc/blob/main/crypto/asn1/a_strex.c#L61
* https://github.com/aws/aws-lc/blob/main/crypto/asn1/posix_time.c#L19

samuel40791765:
> I found a couple of other place we use intypes.h without this define. Should we add it to these as well?

I noticed that too and was on the fence on that. I'm speculating that the bug is C++ specific. The team seems to have gotten the build working by just configuring this C++ file, and [this fix](https://stackoverflow.com/questions/14535556/why-doesnt-priu64-work-in-this-code) also called out C++.

I'm still waiting for the team to test if this fix is sufficient enough for them. If not, we can add them to the C files as well.

justsmth (approved):

skmcgrail (approved):

PR #1157, opened 2023-08-15
Lower returned number for SSL_CTX_get_security_level

samuel40791765:
### Description of changes: 
A contributor pointed out that our default SSL settings don't quite align with OpenSSL's definition for different security levels.
* https://github.com/aws/aws-lc/pull/1002#issuecomment-1678675162

I've lowered the return number for `SSL_CTX_get_security_level`, but this function is still only for compatibility purposes.

### Call-outs:
N/A

### Testing:
N/A

### DIFF
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -5109,6 +5109,7 @@ OPENSSL_EXPORT int SSL_set1_sigalgs_list(SSL *ssl, const char *str);
 // function returns.
 //
 // Per OpenSSL's definition of Level 0, 1, and 2:
+//
 // Level 0:
 // Everything is permitted. This retains compatibility with previous versions of
 // OpenSSL.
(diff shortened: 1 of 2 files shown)


PiotrSikora (changes requested):
It looks that you've changed only the documentation, but not the actual return value.

andrewhop (dismissed):

include/openssl/ssl.h
```diff
@@ -5115,22 +5115,34 @@ OPENSSL_EXPORT int SSL_CTX_set1_sigalgs_list(SSL_CTX *ctx, const char *str);
...
+// than 80 bits of security. SSL version 2 is prohibited. Any cipher suite using
+// MD5 for the MAC is also prohibited.
+//
+// Level 2
+// Security level set to 112 bits of security. As a result RSA, DSA and DH keys
+// shorter than 2048 bits and ECC keys shorter than 224 bits are prohibited. In
+// addition to the level 1 exclusions any cipher suite using RC4 is also
+// prohibited. SSL version 3 is also not allowed. Compression is disabled.
+//
+// AWS-LC's libssl doesn't support SSL version 2 or 3, and we have no support
```
torben-hansen, line 5141: "AWS-LC's libssl doesn't support SSL version 2 or 3"

Unclear what this means.

samuel40791765: I took this from directly the OpenSSL documentation above, which is referring to the ancient SSL protocol. Would changing it to SSLv2 and SSLv3 be better?

torben-hansen: A'ha, yes that's better.

justsmth (approved):

PR #1159, opened 2023-08-16
Implement SSL_get_client_ciphers

WillChilds-Klein:
# Issues
Addresses CryptoAlg-2036

# Notes

This commit implements `SSL_get_client_ciphers`, which returns a "stack" (really just a list) of `SSL_CIPHER` structs [presented by the client](https://github.com/openssl/openssl/blob/master/ssl/statem/statem_srvr.c#L2002) during the `|ssl|`'s most recent handshake. [Like OpenSSL](https://github.com/openssl/openssl/blob/e2972982c64f3f1ac10b3ebe1086d99ec67631bd/ssl/ssl_lib.c#L1433), we free this list when the connection is freed. OpenSSL [obtains this list](https://github.com/openssl/openssl/blob/7a5f58b2cf0d7b2fa0451603a88c3976c657dae9/ssl/ssl_lib.c#L3205-L3212) from from an `SSL_CONNECTION` object, something that we or BoringSSL appear to have removed.

So, we're left with the choice of where to store client ciphers. Options include internal/opaque structs `SSL_CTX`, `SSL_HANDSHAKE`, `SSL_SESSION`, `SSL_CONFIG`, and `SSL3_STATE`. Of these, `SSL_HANDSHAKE` or `SSL_SESSION` seemed most appropriate, but `SSL_HANDSHAKE` is [reset](https://github.com/aws/aws-lc/blob/main/ssl/ssl_lib.cc#L871) immediately after the handshake phase completes and `SSL_SESSION` incurs complications around caching. `SSL_CTX` appears to hold more configuration than mutable state, so we went with storing the client ciphers on `SSL3_STATE`. One important drawback to note here is that `SSL3_STATE` has "experimental" support for ASN.1 serialization. To avoid introducing irrevocable changes to that format, we do not include the client ciphers in that serialized object. When deserialized, client ciphers will be populated as "null". We can add support for this as an optional field in the future if required.

# Testing
- CI runs with new test

---

### DIFF
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -1664,7 +1664,6 @@ OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_ciphers(const SSL *ssl);
 // available, or |ssl| is not operating in server mode, NULL is returned.
 OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_client_ciphers(const SSL *ssl);
 
-
 // Connection information.
 
 // SSL_is_init_finished returns one if |ssl| has completed its initial handshake
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -1659,11 +1659,6 @@ OPENSSL_EXPORT int SSL_CTX_cipher_in_group(const SSL_CTX *ctx, size_t i);
 // SSL_get_ciphers returns the cipher list for |ssl|, in order of preference.
 OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_ciphers(const SSL *ssl);
 
-// SSL_get_client_ciphers returns the stack of available SSL_CIPHERs matching
-// the list received from the client on |ssl|. If |ssl| is NULL, no ciphers are
-// available, or |ssl| is not operating in server mode, NULL is returned.
-OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_client_ciphers(const SSL *ssl);
-
 // Connection information.
 
 // SSL_is_init_finished returns one if |ssl| has completed its initial handshake
@@ -1774,6 +1769,11 @@ OPENSSL_EXPORT int SSL_get_extms_support(const SSL *ssl);
 // SSL_get_current_cipher returns cipher suite used by |ssl|, or NULL if it has
 // not been negotiated yet.
 OPENSSL_EXPORT const SSL_CIPHER *SSL_get_current_cipher(const SSL *ssl);
+
+// SSL_get_client_ciphers returns the stack of available SSL_CIPHERs matching
+// the list received from the client on |ssl|. If |ssl| is NULL, no ciphers are
+// available, or |ssl| is not operating in server mode, NULL is returned.
+OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_client_ciphers(const SSL *ssl);
 
 // SSL_session_reused returns one if |ssl| performed an abbreviated handshake
 // and zero otherwise.
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -1659,6 +1659,11 @@ OPENSSL_EXPORT int SSL_CTX_cipher_in_group(const SSL_CTX *ctx, size_t i);
 // SSL_get_ciphers returns the cipher list for |ssl|, in order of preference.
 OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_ciphers(const SSL *ssl);
 
+// SSL_get_client_ciphers returns the stack of available SSL_CIPHERs matching
+// the list received from the client on |ssl|. If |ssl| is NULL, no ciphers are
+// available, or |ssl| is not operating in server mode, NULL is returned.
+OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_client_ciphers(const SSL *ssl);
+
 
 // Connection information.
 
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -1659,6 +1659,7 @@ OPENSSL_EXPORT int SSL_CTX_cipher_in_group(const SSL_CTX *ctx, size_t i);
 // SSL_get_ciphers returns the cipher list for |ssl|, in order of preference.
 OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_ciphers(const SSL *ssl);
 
+
 // Connection information.
 
 // SSL_is_init_finished returns one if |ssl| has completed its initial handshake
@@ -2601,7 +2602,9 @@ OPENSSL_EXPORT const char *SSL_get_group_name(uint16_t group_id);
 // finished.
 // WARNING: Currently only supports |SSL| as server.
 // WARNING: CRYPTO_EX_DATA |ssl->ex_data| is not encoded. Remember set |ex_data|
-// back after decode. WARNING: BIO |ssl->rbio| and |ssl->wbio| are not encoded.
+// back after decode.
+// WARNING: BIO |ssl->rbio| and |ssl->wbio| are not encoded.
+// WARNING: STACK_OF(SSL_CIPHER) |ssl->client_cipher_suites| is not encoded.
 //
 // Initial implementation of this API is made by Evgeny Potemkin.
 OPENSSL_EXPORT int SSL_to_bytes(const SSL *in, uint8_t **out_data,
@@ -2617,6 +2620,8 @@ OPENSSL_EXPORT int SSL_to_bytes(const SSL *in, uint8_t **out_data,
 //          Otherwise, the connections use the same key material.
 // WARNING: Remember set |ssl->rbio| and |ssl->wbio| before using |ssl|.
 // WARNING: Remember set callback functions and |ex_data| back if needed.
+// WARNING: STACK_OF(SSL_CIPHER) |ssl->client_cipher_suites| is not encoded and
+//          will be repopulated on next handshake.
 // WARNING: To ensure behavior unchange, |ctx| setting should be the same.
 //
 // Initial implementation of this API is made by Evgeny Potemkin.
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -1770,9 +1770,11 @@ OPENSSL_EXPORT int SSL_get_extms_support(const SSL *ssl);
 // not been negotiated yet.
 OPENSSL_EXPORT const SSL_CIPHER *SSL_get_current_cipher(const SSL *ssl);
 
-// SSL_get_client_ciphers returns the stack of available SSL_CIPHERs matching
-// the list received from the client on |ssl|. If |ssl| is NULL, no ciphers are
-// available, or |ssl| is not operating in server mode, NULL is returned.
+// SSL_get_client_ciphers returns stack of ciphers offered by the client during
+// a handshake. If the |ssl| is a client or the handshake hasn't occurred yet,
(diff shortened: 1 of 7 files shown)


ssl/handshake_server.cc
```diff
@@ -264,32 +264,41 @@ static bool negotiate_version(SSL_HANDSHAKE *hs, uint8_t *out_alert,
...
+    const SSL_CLIENT_HELLO *client_hello, UniquePtr<STACK_OF(SSL_CIPHER)> *ciphers_out) {
+  ciphers_out->reset();
+
   CBS cipher_suites;
   CBS_init(&cipher_suites, client_hello->cipher_suites,
            client_hello->cipher_suites_len);
 
+  // Cipher suites are encoded as 2-byte unsigned integers
+  if (CBS_len(&cipher_suites) % 2 != 0) {
+    return false;
```
skmcgrail, line 277: Should this push an error before returning? Previous error would have been a `SSL_R_ERROR_IN_RECEIVED_CIPHER_LIST` while trying to read a u16 from the CBS. Seems like this would mask malformed data.

skmcgrail: I guess this check was down in the TLS 1.3 chose cipher logic. I'm not sure what the protocol behavior should be here, and whether its the same between 1.2 & 1.3.

ssl/internal.h
```diff
@@ -2970,6 +2976,10 @@ struct SSL3_STATE {
   // immutable.
   UniquePtr<SSL_SESSION> established_session;
 
+  // client_ciphers holds the peer's declared cipher preferences, in order. This
```
torben-hansen, line 2979:
```suggestion
  // client_cipher_suites contains cipher suites offered by the client during the handshake, with preference order maintained. This
```

ssl/handshake_server.cc
WillChilds-Klein: I believe the prior behavior for odd length CBS would have been to [not set an error](https://github.com/aws/aws-lc/blob/main/ssl/s3_both.cc#L691) until two calls up the stack where [we'd set a `SSL_R_NO_SHARED_CIPHER` error and send a `SSL_AD_HANDSHAKE_FAILURE`](https://github.com/aws/aws-lc/blob/main/ssl/tls13_server.cc#L243-L245). This change currently sends a `SSL_AD_INTERNAL_ERROR` alert on parsing failure, which doesn't seem [per RFC 5246](https://datatracker.ietf.org/doc/html/rfc5246#section-7.2.2).

I'll change the alert sent to the client, and set the error to `SSL_R_ERROR_IN_RECEIVED_CIPHER_LIST` on parse failure as that seems most appropriate. I don't think setting the error actually impacts what goes on the wire, just how the error is reported locally.

ssl/internal.h
```diff
@@ -2970,6 +2976,10 @@ struct SSL3_STATE {
...
+  // field is NOT serialized and is only populated if used in a server context.
+  bssl::UniquePtr<STACK_OF(SSL_CIPHER)> client_ciphers;
```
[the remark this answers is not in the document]
WillChilds-Klein: sure, updated.

ssl/internal.h
WillChilds-Klein: sure, updated.

ssl/handshake_server.cc
skmcgrail: I think those changes proposed would be appropriate.

andrewhop:
Would any of this be simpler if you just put the state in our `SSL`? Our `SSL/ssl_st` is the same as OpenSSL's `SSL_CONNECTION/ssl_connection_st`. We're missing `SSL_CONNECTION` because OpenSSL appears to be doing some [casting to fit both Quic and TLS](https://github.com/openssl/openssl/blob/master/ssl/ssl_local.h#L1820-L1826) into the same `SSL`. Our implementation is different and tracks two different function pointer look ups in `method` and `quic_method`.

WillChilds-Klein:
> Would any of this be simpler if you just put the state in our `SSL`? Our `SSL/ssl_st` is the same as OpenSSL's `SSL_CONNECTION/ssl_connection_st`.

Thanks for the suggestion @andrewhop, I hadn't realized that this path was used by QUIC handshakes as well but reading back through the code, I don't see anywhere we'd branch out of the handshake state machine or parse client Hello differently due to QUIC-ness of the connection. I've added a commit to store the client ciphers stack on the `SSL` struct so it's available on both QUIC and TLS connections (as long as they're not rehydrated from ASN.1!).

andrewhop:
Is this function supposed to work for DTLS/QUIC? Can you add a test that it works/doesn't work as expected?

ssl/handshake_server.cc
```diff
@@ -810,12 +812,18 @@ static enum ssl_hs_wait_t do_select_certificate(SSL_HANDSHAKE *hs) {
     return ssl_hs_error;
   }
 
+  if (!ssl_parse_client_cipher_list(&client_hello, &ssl->client_cipher_suites)) {
+    OPENSSL_PUT_ERROR(SSL, SSL_R_ERROR_IN_RECEIVED_CIPHER_LIST);
+    ssl_send_alert(ssl, SSL3_AL_FATAL, SSL_AD_HANDSHAKE_FAILURE);
+    return ssl_hs_error;
+  }
```
andrewhop, line 819: Is this changing the error we return? Previously if something failed in choose_cipher or ssl_parse_client_cipher_list (which was called by choose_cipher) the below error block would be hit which sends SSL_R_NO_SHARED_CIPHER. Was this intentional, does OpenSSL return this error for this case?

include/openssl/ssl.h
```diff
@@ -1771,6 +1770,13 @@ OPENSSL_EXPORT int SSL_get_extms_support(const SSL *ssl);
 // not been negotiated yet.
 OPENSSL_EXPORT const SSL_CIPHER *SSL_get_current_cipher(const SSL *ssl);
 
+// SSL_get_client_ciphers returns stack of ciphers offered by the client during
+// a handshake. If the |ssl| is a client or the handshake hasn't occurred yet,
+// NULL is returned. The stack of ciphers IS NOT de/serialized, so NULL will
```
andrewhop, line 1775: Great call out, can you also add this to SSL_to/from_bytes to try and help future users find this caveat as well.

include/openssl/ssl.h
```diff
@@ -1659,7 +1659,6 @@ OPENSSL_EXPORT int SSL_CTX_cipher_in_group(const SSL_CTX *ctx, size_t i);
 // SSL_get_ciphers returns the cipher list for |ssl|, in order of preference.
 OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_ciphers(const SSL *ssl);
 
-
```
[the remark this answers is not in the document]
WillChilds-Klein: ah, thanks for catching this. will fix.

include/openssl/ssl.h
WillChilds-Klein: will do.

ssl/handshake_server.cc
WillChilds-Klein: It was intentional, as I'd thought this error was more precise for describing a parse error. OpenSSL seems to [prefer](https://github.com/search?q=repo%3Aopenssl%2Fopenssl+SSL_R_ERROR_IN_RECEIVED_CIPHER_LIST&type=code) `SSL_R_ERROR_IN_RECEIVED_CIPHER_LIST` for parsing issues and [prefer](https://github.com/search?q=repo%3Aopenssl%2Fopenssl+SSL_R_NO_SHARED_CIPHER&type=code) `SSL_R_NO_SHARED_CIPHER` for not being able to find common cipher value.

Reconsidering this, though, I see the benefit of preserving backwards compatibility with local errors we return. I'll expand the places we push `SSL_R_ERROR_IN_RECEIVED_CIPHER_LIST` onto the error queue in `ssl_parse_client_cipher_list` and also push `SSL_R_NO_SHARED_CIPHER` in this failure case.

WillChilds-Klein:
> Is this function supposed to work for DTLS/QUIC? Can you add a test that it works/doesn't work as expected?

@andrewhop -- I've added checks to `QUICMethodTest.ZeroRTTAccept`. Does that seem like the right place?

ssl/s3_both.cc
```diff
@@ -686,36 +686,28 @@ class CipherScorer {
...
-      return nullptr;
-    }
-
-    SSL_CIPHER *candidate = nullptr;
+  for (size_t i = 0; i < sk_SSL_CIPHER_num(client_cipher_suites); i++) {
+    const SSL_CIPHER *client_cipher = sk_SSL_CIPHER_value(client_cipher_suites, i);
+    const SSL_CIPHER *candidate = nullptr;
     if (tls13_ciphers != nullptr) {
       // Limit to customized TLS 1.3 ciphers if configured.
-      uint32_t cipher_suite_id = 0x03000000L | cipher_suite;
```
andrewhop, line 708: What does `0x03000000L | cipher_suite` do? Why don't we need it anymore?

WillChilds-Klein: Ah, great question. I should have called this out on the PR. The magical value `0x03000000L` is `|`'d onto each wire-protocol ciphersuite ID [by OpenSSL](https://github.com/aws/aws-lc/blob/6a1e61a81bb5a1f77bcbf94a1b451b89d939d59b/ssl/ssl_cipher.cc#L1452-L1454). Previously, this section code was doing direct ID equality comparisons on L711 between wire protocol `cipher_suite_id` and OpenSSL ciphersuite IDs, so they had to do the `| 0x03...L` operation here for that to work.

After this refactor, we obtain the `SSL_CIPHER` struct by wire protocol value using [a function which prepends](https://github.com/aws/aws-lc/blob/6a1e61a81bb5a1f77bcbf94a1b451b89d939d59b/ssl/ssl_cipher.cc#L1443) the magic value for us. We then do comparisons using only `SSL_CIPHER.id`, so we [don't need to worry about the conversion](https://github.com/aws/aws-lc/blob/6a1e61a81bb5a1f77bcbf94a1b451b89d939d59b/ssl/internal.h#L544).

To prove this out, I modified this loop with the following diff:

```
$ git diff
diff --git a/ssl/s3_both.cc b/ssl/s3_both.cc
index d09954daa..4b80bc161 100644
--- a/ssl/s3_both.cc
+++ b/ssl/s3_both.cc
@@ -695,11 +695,17 @@ const SSL_CIPHER *ssl_choose_tls13_cipher(

   for (size_t i = 0; i < sk_SSL_CIPHER_num(client_cipher_suites); i++) {
     const SSL_CIPHER *client_cipher = sk_SSL_CIPHER_value(client_cipher_suites, i);
+    if ((client_cipher->id & 0xff000000) != 0x03000000) {
+        abort();
+    }
     const SSL_CIPHER *candidate = nullptr;
     if (tls13_ciphers != nullptr) {
       // Limit to customized TLS 1.3 ciphers if configured.
       for (size_t j = 0; j < sk_SSL_CIPHER_num(tls13_ciphers); j++) {
         const SSL_CIPHER *cipher = sk_SSL_CIPHER_value(tls13_ciphers, j);
+        if ((client_cipher->id & 0xff000000) != 0x03000000) {
+            abort();
+        }
         if (cipher != nullptr && cipher->id == client_cipher->id) {
           candidate = (SSL_CIPHER *)cipher;
           break;
```

and successfully ran the new test after rebuilding:

```
$ cmake-build.sh && ./build/ssl/ssl_test --gtest_filter='*SSLVersionTest.SSLClientCiphers*'; echo $?
...
[==========] 8 tests from 1 test suite ran. (8 ms total)
[  PASSED  ] 8 tests.
0
```

andrewhop (approved):

skmcgrail (approved):

PR #1164, opened 2023-08-17
Conditionally enable Neon version of s2n-bignum implementations for montgomery multiplications

aqjune-aws:
### Description of changes: 
This patch is the last patch splitted from #1108 for improving RSA signature speed for Graviton 2.
This conditionally enables the the Neon versions of verified primitives of s2n-bignum when `__ARM_NEON` is defined.
I checked this `__ARM_NEON` macro to be consistent with what BoringSSL does, according to [BUILDING.md](https://github.com/aws/aws-lc/blob/main/BUILDING.md).
The scalar versions of s2n-bignum primitives are already merged in b706d7e.

The performance improvements of RSA signatures on Graviton 2 are as follows (Unit: ops/sec).

|Bits|Operation|AWS-LC before b706d7e|AWS-LC after b706d7e|This patch|speedup vs. before b706d7e|
|:----|:----|:----|:----|:----|:----|
|2048|RSA sign|299.3|399|495.8|65.65%|
| |verify (fresh key)|10736.3|15491|18836.6|75.45%|
|3072|RSA sign|95.4|113.2|126.4|32.49%|
| |verify (fresh key)|4917.7|6001.7|6579.1|33.78%|
|4096|RSA sign|41.7|63.2|78.3|87.77%|
| |verify (fresh key)|2781.6|3451|3800.3|36.62%|

### Testing:
Tested via `bssl speed -filter RSA`

### DIFF
--- a/crypto/fipsmodule/bn/montgomery.c
+++ b/crypto/fipsmodule/bn/montgomery.c
@@ -477,15 +477,29 @@ static void montgomery_s2n_bignum_mul_mont(BN_ULONG *rp, const BN_ULONG *ap,
   uint64_t w = n0[0];
 
   if (num == 32) {
+#if defined(__ARM_NEON)
+    if (ap == bp)
+      bignum_ksqr_32_64_neon(mulres, ap, t);
+    else
+      bignum_kmul_32_64_neon(mulres, ap, bp, t);
+#else
     if (ap == bp)
       bignum_ksqr_32_64(mulres, ap, t);
     else
       bignum_kmul_32_64(mulres, ap, bp, t);
+#endif
   } else if (num == 16) {
+#if defined(__ARM_NEON)
+    if (ap == bp)
+      bignum_ksqr_16_32_neon(mulres, ap, t);
+    else
+      bignum_kmul_16_32_neon(mulres, ap, bp, t);
+#else
     if (ap == bp)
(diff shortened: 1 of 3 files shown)


crypto/fipsmodule/bn/montgomery.c
```diff
@@ -477,15 +477,29 @@ static void montgomery_s2n_bignum_mul_mont(BN_ULONG *rp, const BN_ULONG *ap,
   uint64_t w = n0[0];
 
   if (num == 32) {
+#if defined(__ARM_NEON)
```
nebeid, line 480: It's preferable to replace the occurrences of `#if defined` with `if CRYPTO_is_NEON_capable()`. That's possible since this code is surrounded by `BN_MONTGOMERY_S2N_BIGNUM_CAPABLE`, which is defined only for AArch64.
We could have assumed that neon is always available on AArch64, but in FIPS tests, we disable all capabilities as one of the configurations we test. So we need this check and this choice between vectorised and non-vectorised implementations.

nebeid (approved):

dkostic (approved):

aqjune-aws:
Thank you for reviews! It seems I don't have a permission to merge this pull request. Could anyone merge this for me please?

PR #1166, opened 2023-08-23
Calibrate BN_window_bits_for_ctime_exponent_size to use 5

aqjune-aws:
### Description of changes: 

This patch sets the value of `BN_window_bits_for_ctime_exponent_size` to 5 instead of 6.

The expression goes back to 2005, this commit: https://github.com/openssl/openssl/commit/46a643763de6d8e39ecf6f76fa79b4d04885aa59#diff-a55637d1fa727ef3bb6c79efd1713b54aeb21c76d45ec0d205200b015233d635R187-R191
The experimental results show that setting the maximum value to 5 brings performance improvements for Graviton 2 and 3:
[w-experiments2.xlsx](https://github.com/aws/aws-lc/files/12423434/w-experiments2.xlsx)
For Intel machines, they are already set to 5 because `OPENSSL_BN_ASM_MONT5` is defined.
For Graviton 2, this improves the 2048 bit RSA performance from ~495 ops/sec to 515 ops/sec.
Since small-bit integer modular exponentiation won't happen in realistic operations, set it to constant 5.

Hanno Becker suggested this idea and provided initial experimentation result.

### Call-outs:

### Testing:


aqjune-aws:
If the very small bits are ignored, `BN_window_bits_for_ctime_exponent_size` can be further simplified to 5.

crypto/fipsmodule/bn/exponentiation.c
```diff
@@ -872,8 +872,8 @@ static int copy_from_prebuf(BIGNUM *b, int top, const BN_ULONG *table, int idx,
 // lines with a cache-line-dependent constant-time mitigation. They can probably
```
torben-hansen, line 872: This comment is no longer relevant?

torben-hansen (commented):
You should be able to simplify this now?

https://github.com/aws/aws-lc/blob/5f4c79cce21943807ab9b750845ce8beca5d7230/crypto/fipsmodule/bn/exponentiation.c#L968C19-L974

dkostic:
>For Graviton 2, this improves the 2048 bit RSA performance from ~495 ops/sec to 515 ops/sec.

which RSA operation is this?

aqjune-aws:
> > For Graviton 2, this improves the 2048 bit RSA performance from ~495 ops/sec to 515 ops/sec.
> 
> which RSA operation is this?

It is the signature operation.


Regarding the small bits (<= 306) cases, well I couldn't find their use cases at least from `bssl speed` and also `SpeedDHcheck`. (DH check only uses 2048 and 3072 bit primes).
Simply remove the smaller cases, or should we preserve the heritage?

dkostic (approved):

torben-hansen (approved):
There is another `window` branch further down the method. But I suppose the compiler will just elide this for high-optimisation builds. So, OK.

```
    // If the window size is greater than 1, then calculate
    // val[i=2..2^winsize-1]. Powers are computed as a*a^(i-1)
    // (even powers could instead be computed as (a^(i/2))^2
    // to use the slight performance advantage of sqr over mul).
    if (window > 1) {
```

aqjune-aws:
Thanks! I merged the latest main branch to here. Waiting for CI...

PR #1170, opened 2023-08-28
Add EVP support for SHA512-224

WillChilds-Klein:
# Issues

CryptoAlg-2034

# Notes

We also add support for using this truncated hash function for HMAC + RSA signature padding and add the new algorithm to the FIPS module and its tooling.

# Testing
- CI runs with added unit tests + KATs


dkostic:
the last 4 bytes are not correct because `sha512_final_impl` is not implemented with 224 variant in mind. All the other possible digest lengths (512, 384, 256) are multiples of 64, while 224 is not: https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/sha/sha512.c#L291. You need to adapt this part so it can correctly handle the 224 variant as well: https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/sha/sha512.c#L292C17-L296

WillChilds-Klein:
> the last 4 bytes are not correct because `sha512_final_impl` is not implemented with 224 variant in mind. All the other possible digest lengths (512, 384, 256) are multiples of 64, while 224 is not: https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/sha/sha512.c#L291. You need to adapt this part so it can correctly handle the 224 variant as well: https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/sha/sha512.c#L292C17-L296

Thanks @dkostic -- I wish I'd seen this a few days ago when you posted it! It took me a few hours to figure this out independently through reading and re-reading `sha512.c` -- the last 4 bytes were consistently incorrect in my initial draft. I think that particular issue is [fixed now](https://github.com/aws/aws-lc/pull/1170/commits/2e954c6b01b072e6f53ba8f42cb876eca3680e08).

crypto/fipsmodule/sha/sha512.c
```diff
@@ -288,13 +333,21 @@ static int sha512_final_impl(uint8_t *out, size_t md_len, SHA512_CTX *sha) {
...
+  const size_t remainder = md_len % 8;
+  assert(remainder == 0 || (remainder == 4 && out_words * 8 + remainder == md_len));
   for (size_t i = 0; i < out_words; i++) {
     CRYPTO_store_u64_be(out, sha->h[i]);
     out += 8;
   }
 
+  // TODO [childw] explain this
+  if (remainder == 4) {
+    const uint64_t trailer = CRYPTO_bswap8(sha->h[out_words]);
```
andrewhop, line 347: Do you always need to swap here, or should it use `CRYPTO_store_u64_be` which swaps on little endian platform and does not swap on big endian platforms (pending https://github.com/aws/aws-lc/pull/1133).

WillChilds-Klein: Hm, good question. In the current small-endian world, we want to take the rightmost 4 bytes of the last element of the state array, swap the order of those bytes, and store them in the last 4 bytes of the output buffer. That byte order swap would not be required on big-endian systems. However, we can't use `CRYPTO_store_u64_be` here to write directly to the output buffer, as that would potentially overwrite the end of the output buffer by 4 bytes (SHA512-224 digest size is 28 bytes, which is not 8-byte-word-aligned).

To avoid `#ifdef`'ing here, I'll use `CRYPTO_store_u64_be` to store the full 8-byte word into a temporary variable and then only write the first 4 bytes of that to the output buffer.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -197,8 +198,9 @@ static int is_md_fips_approved_for_verifying(int md_type, int pkey_type) {
     case NID_sha384:
     case NID_sha512:
       return 1;
+    case NID_sha512_224:
     case NID_sha512_256:
-      // SHA512/256 is only approved for verifying with RSA PSS
+      // Truncated SHA512 is only approved for verifying with RSA PSS
       if (pkey_type == EVP_PKEY_RSA_PSS) {
         return 1;
       }
```
WillChilds-Klein, line 206: I'm pattern-matching to SHA512-256 here. Does anyone know of a NIST document (or similar) that stipulates which usages of these truncated hash algorithms are FIPS-approved?

util/convert_wycheproof/convert_wycheproof.go
```diff
@@ -224,6 +224,8 @@ var defaultInputs = []string{
 	"hmac_sha256_test.json",
 	"hmac_sha384_test.json",
 	"hmac_sha512_test.json",
+	"hmac_sha512_224_test.json",
+	"hmac_sha512_256_test.json",
```
WillChilds-Klein, line 228: HMAC w/ SHA512-256 wasn't previously tested here, so I added the wycheproof KAT files from [their github repository](https://github.com/google/wycheproof).

crypto/fipsmodule/sha/sha512.c
```diff
@@ -288,13 +333,23 @@ static int sha512_final_impl(uint8_t *out, size_t md_len, SHA512_CTX *sha) {
     return 0;
   }
 
-  assert(md_len % 8 == 0);
   const size_t out_words = md_len / 8;
+  const size_t remainder = md_len % 8;
```
dkostic, line 337: I'd prefer to be more specific here. Since we support only 512, 256, and 224 digest lengths, how about
```
assert(md_len == SHA512_LEN || md_len == SHA512_256_LEN || md_len == SHA512_224_LEN)
```
and then further down
```
if (md_len == SHA512_224_LEN) {
...
}
```

util/convert_wycheproof/convert_wycheproof.go
```diff
@@ -26,11 +26,11 @@ import (
...
 type wycheproofTest struct {
-	Algorithm        string            `json:"algorithm"`
-	GeneratorVersion string            `json:"generatorVersion"`
-	NumberOfTests    int               `json:"numberOfTests"`
-	Notes            map[string]string `json:"notes"`
-	Header           []string          `json:"header"`
+	Algorithm        string         `json:"algorithm"`
+	GeneratorVersion string         `json:"generatorVersion"`
+	NumberOfTests    int            `json:"numberOfTests"`
+	Notes            map[string]any `json:"notes"`
```
dkostic, line 32: just curious, why the change from `string` to `any`?

crypto/fipsmodule/sha/sha512.c
WillChilds-Klein: Tightening this up makes sense. I'll update.

third_party/wycheproof_testvectors/hmac_sha512_224_test.json
```diff
@@ -0,0 +1,1978 @@
+{
+  "algorithm" : "HMACSHA512/224",
+  "schema" : "mac_test_schema.json",
+  "generatorVersion" : "0.9",
+  "numberOfTests" : 173,
+  "header" : [
+    "Test vectors of type MacTest are intended for testing the",
+    "generation and verification of MACs."
+  ],
+  "notes" : {
```
WillChilds-Klein, line 10: @dkostic

util/convert_wycheproof/convert_wycheproof.go
WillChilds-Klein: I'm not sure why, but the new test vectors have a `notes` section that's a struct rather than a string.

crypto/fipsmodule/sha/sha512.c
WillChilds-Klein: on second pass, it looks like we also [use this function impl for SHA384](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/sha/sha512.c#L188). rather than case out all of the functions that call this, I'm going to keep the assertion as it was before this PR, but extend it to also accept the `md_len == SHA512_224_LEN` case.

dkostic (approved):

andrewhop (approved):

crypto/fipsmodule/service_indicator/service_indicator.c
WillChilds-Klein: page 22 of [this NIST document](https://pages.nist.gov/ACVP/draft-celi-acvp-rsa.txt) seems to indicate that SHA512-224 is viable for RSA padding.

PR #1175, opened 2023-09-06
Align with OpenSSL on TLS 1.3 cipher suite constants.

andrewhop:
Addresses CryptoAlg-1310 and CryptoAlg-1314, these were initially skipped.


samuel40791765 (approved):

samuel40791765:
We also need this commit to turn on the [--with-http_v3_module](https://nginx.org/en/docs/configure.html#http_v3_module) configuration and enable QUIC in NGINX.

justsmth (approved):

PR #1176, opened 2023-09-08
Fix and add CI for postgres build on arm

samuel40791765:
### Description of changes: 
Postgres doesn't allow functions that call `exit` according to the documentation here:
* https://github.com/postgres/postgres/blob/master/src/interfaces/libpq/Makefile#L119

With the `exit` function called in `crypto/fipsmodule/cpucap/cpu_aarch64.c`, our build breaks prematurely when built with PostGres on ARM.

### Call-outs:
N/A

### Testing:
New CI dimension.

### DIFF
--- a/tests/ci/docker_images/linux-aarch/ubuntu-22.04_gcc-12x/Dockerfile
+++ b/tests/ci/docker_images/linux-aarch/ubuntu-22.04_gcc-12x/Dockerfile
@@ -16,5 +16,11 @@ RUN set -ex && \
     rm -rf /var/lib/apt/lists/* && \
     rm -rf /tmp/*
 
+# Postgres's integration tests cannot be ran as root, so we have to define
+# a non-root user here to use in Codebuild.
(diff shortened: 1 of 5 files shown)


tests/ci/docker_images/linux-aarch/ubuntu-22.04_gcc-12x/Dockerfile
```diff
@@ -16,5 +16,11 @@ RUN set -ex && \
     rm -rf /var/lib/apt/lists/* && \
     rm -rf /tmp/*
 
+# Postgres's integration tests cannot be ran as root, so we have to define
```
nebeid, line 19: Nit
```suggestion
# Postgres's integration tests cannot be run as root, so we have to define
```

nebeid (approved):

billbo-yang (approved):

nebeid (approved):

PR #1177, opened 2023-09-08
Upstream Merge - Remove unions in EC_SCALAR and EC_FELEM.

justsmth:
### Issues:
Relates to: CryptoAlg-1206 and CryptoAlg-2021

### Description of changes:

This a merge of an [upstream commit](https://github.com/google/boringssl/commit/227ff6e6425283b83594a91a1aa81cc78f1a88df) that transitions `EC_SCALAR` and `EC_FELEM` from unions to structs. Most of that commit was accepted.

Original commit message:
---------
---------
When introducing EC_SCALAR and EC_FELEM, I used unions as convenience for converting to and from the byte representation. However, type-punning with unions is not allowed in C++ and hard to use correctly in C. As I understand the rules, they are:

- The abstract machine knows what member of union was last written to.

- In C, reading from an inactive member is defined to type-pun. In C++, it is UB though some compilers promise the C behavior anyway.

- However, if you read or write from a *pointer* to a union member, the strict aliasing rule applies. (A function passed two pointers of different types otherwise needs to pessimally assume they came from the same union.)

That last rule means the type-punning allowance doesn't apply if you take a pointer to an inactive member, and it's common to abstract otherwise direct accesses of members via pointers.

https://github.com/openssl/openssl/issues/18225 is an example where similar union tricks have caused problems for OpenSSL. While we don't have that code, EC_SCALAR and EC_FELEM play similar tricks.

We do get a second lifeline because our alternate view is a uint8_t, which we require to be unsigned char. Strict aliasing always allows the pointer type to be a character type, so pointer-indirected accesses of EC_SCALAR.bytes aren't necessarily UB. But if we ever write to EC_SCALAR.bytes directly (and we do), we'll switch the active arm and then pointers to EC_SCALAR.words become strict aliasing violations!

This is all far too complicated to deal with. Ideally everyone would build with -fno-strict-aliasing because no real C code actually follows these rules. But we don't always control our downstream consumers' CFLAGS, so let's just avoid the union. This also avoids a pitfall if we ever move libcrypto to C++.

For p224-64.c, I just converted the representations directly, which avoids worrying about the top 32 bits in p224_felem_to_generic. Most of the rest was words vs. bytes conversions and boils down to a cast (we're still dealing with a character type, at the end of the day). But I took the opportunity to extract some more "words"-based helper functions out of BIGNUM, so the casts would only be in one place. That too saves us from the top bits problem in the bytes-to-words direction.

Bug: 301
Change-Id: I3285a86441daaf824a4f6862e825d463a669efdb
Reviewed-on: https://boringssl-review.googlesource.com/c/boringssl/+/52505
Commit-Queue: Bob Beck <bbe@google.com>
Reviewed-by: Bob Beck <bbe@google.com>

(cherry picked from commit 227ff6e6425283b83594a91a1aa81cc78f1a88df)
---------
---------

### Call-outs:
I removed remaining references to the `bytes` fields of `EC_SCALAR` and `EC_FELEM`.

### Testing:
* crypt_test and ssl_test pass for me locally.
* The corresponding update for aws-lc-verification is here: https://github.com/awslabs/aws-lc-verification/pull/118

### DIFF
--- a/crypto/fipsmodule/ec/p224-64.c
+++ b/crypto/fipsmodule/ec/p224-64.c
@@ -52,11 +52,6 @@ typedef uint128_t p224_widelimb;
 typedef p224_limb p224_felem[4];
 typedef p224_widelimb p224_widefelem[7];
 
-// Field element represented as a byte arrary. 28*8 = 224 bits is also the
-// group order size for the elliptic curve, and we also use this type for
-// scalars for point multiplication.
-typedef uint8_t p224_felem_bytearray[28];
-
 // Precomputed multiples of the standard generator
 // Points are given in coordinates (X, Y, Z) where Z normally is 1
 // (0 for the point at infinity).
@@ -180,31 +175,20 @@ static const p224_felem g_p224_pre_comp[2][16][3] = {
       {0x32477c61b6e8c6, 0xb46a97570f018b, 0x91176d0a7e95d1, 0x3df90fbc4c7d0e},
       {1, 0, 0, 0}}}};
 
-static uint64_t p224_load_u64(const uint8_t in[8]) {
-  uint64_t ret;
-  OPENSSL_memcpy(&ret, in, sizeof(ret));
-  return ret;
-}
 
 // Helper functions to convert field elements to/from internal representation
-static void p224_bin28_to_felem(p224_felem out, const uint8_t in[28]) {
-  out[0] = p224_load_u64(in) & 0x00ffffffffffffff;
-  out[1] = p224_load_u64(in + 7) & 0x00ffffffffffffff;
-  out[2] = p224_load_u64(in + 14) & 0x00ffffffffffffff;
-  out[3] = p224_load_u64(in + 20) >> 8;
-}
 
-static void p224_felem_to_bin28(uint8_t out[28], const p224_felem in) {
-  for (size_t i = 0; i < 7; ++i) {
-    out[i] = in[0] >> (8 * i);
-    out[i + 7] = in[1] >> (8 * i);
-    out[i + 14] = in[2] >> (8 * i);
-    out[i + 21] = in[3] >> (8 * i);
-  }
-}
-
+#if defined(AWSLC_FIPS)
+// Related to delocate issue with gcc-11
+__attribute__((noinline))
+#endif
 static void p224_generic_to_felem(p224_felem out, const EC_FELEM *in) {
-  p224_bin28_to_felem(out, in->bytes);
+  // |p224_felem|'s minimal representation uses four 56-bit words. |EC_FELEM|
+  // uses four 64-bit words. (The top-most word only has 32 bits.)
+  out[0] = in->words[0] & 0x00ffffffffffffff;
+  out[1] = ((in->words[0] >> 56) | (in->words[1] << 8)) & 0x00ffffffffffffff;
+  out[2] = ((in->words[1] >> 48) | (in->words[2] << 16)) & 0x00ffffffffffffff;
+  out[3] = ((in->words[2] >> 40) | (in->words[3] << 24)) & 0x00ffffffffffffff;
 }
 
 // Requires 0 <= in < 2*p (always call p224_felem_reduce first)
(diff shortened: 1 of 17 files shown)


dkostic (approved):

samuel40791765 (dismissed):

crypto/fipsmodule/ec/p224-64.c
```diff
@@ -204,7 +199,7 @@ static void p224_felem_to_bin28(uint8_t out[28], const p224_felem in) {
 }
 
 static void p224_generic_to_felem(p224_felem out, const EC_FELEM *in) {
-  p224_bin28_to_felem(out, in->bytes);
```
nebeid, line 207: There were some changes (deletions) before this one in the original commit, why we didn't take them?

justsmth: The deletion in the original commit were removing `p224_bin28_to_felem` and `p224_load_u64`. But, because I didn't take the change to `p224_generic_to_felem`, those functions are still in use.

nebeid: Why we didn't take the change to `p224_generic_to_felem`?

crypto/fipsmodule/ec/p384.c
```diff
@@ -162,7 +162,7 @@ static void p384_felem_cmovznz(p384_limb_t out[P384_NLIMBS],
 
 // NOTE: the input and output are in little-endian representation.
 static void p384_from_generic(p384_felem out, const EC_FELEM *in) {
-  p384_felem_from_bytes(out, in->bytes);
+  p384_felem_from_bytes(out, (const uint8_t *)in->words);
```
dkostic, line 165: can you please add a TODO and maybe open a SIM for adding `p384_felem_from/to_words` as Boring did in P-256  so we can remove all the casts in P-384/521 files?

crypto/fipsmodule/ec/p384.c
```diff
@@ -685,10 +689,10 @@ OPENSSL_STATIC_ASSERT(P384_MUL_WSIZE == 5,
 // It forces an odd scalar and outputs digits in
 // {\pm 1, \pm 3, \pm 5, \pm 7, \pm 9, ...}
 // i.e. signed odd digits with _no zeroes_ -- that makes it "regular".
-static void p384_felem_mul_scalar_rwnaf(int16_t *out, const unsigned char *in) {
+static void p384_felem_mul_scalar_rwnaf(int16_t *out, const EC_SCALAR *in) {
   int16_t window, d;
 
-  window = (in[0] & P384_MUL_WSIZE_MASK) | 1;
+  window = (in->words[0] & P384_MUL_WSIZE_MASK) | 1;
```
dkostic, line 695: I don't like that we are mixing `int16_t` with `uint64_t` here but I don't have a better suggestion.

crypto/fipsmodule/ec/p224-64.c
justsmth: Fixed in latest update.

crypto/fipsmodule/ec/p224-64.c
```diff
@@ -180,31 +175,20 @@ static const p224_felem g_p224_pre_comp[2][16][3] = {
...
-    out[i] = in[0] >> (8 * i);
-    out[i + 7] = in[1] >> (8 * i);
-    out[i + 14] = in[2] >> (8 * i);
-    out[i + 21] = in[3] >> (8 * i);
-  }
-}
 
+#if defined(AWSLC_FIPS)
+// Related to delocate issue with gcc-11
+__attribute__((noinline))
```
torben-hansen, line 183: Is this attribute portable enough? It's a GNU thing. I don't recall if we need to guard against other compilers.

justsmth: This file is not compiled on Windows:
* [BORINGSSL_HAS_UINT128](https://github.com/aws/aws-lc/blob/6a1e61a81bb5a1f77bcbf94a1b451b89d939d59b/crypto/internal.h#L160) is not defined for Windows:
```
#if (!defined(_MSC_VER) || defined(__clang__)) && defined(OPENSSL_64_BIT)
#define BORINGSSL_HAS_UINT128
```
* [p224-64.c](https://github.com/aws/aws-lc/blob/6a1e61a81bb5a1f77bcbf94a1b451b89d939d59b/crypto/fipsmodule/ec/p224-64.c#L34) requires `BORINGSSL_HAS_UINT128` to be set.
```
#if defined(BORINGSSL_HAS_UINT128) && !defined(OPENSSL_SMALL)
```

dkostic (approved):

samuel40791765 (approved):

nebeid:
^
https://github.com/aws/aws-lc/pull/1240/files#r1355669244
[crypto/fipsmodule/ec/internal.h L.112](https://github.com/aws/aws-lc/pull/1240/files#diff-9053bbc88f62c34ad85263cee90396124b4a5200a2cc390e8525e067aaeb4d60L112)
```
- typedef union {
+ typedef struct {
```

PR #1178, opened 2023-09-08
Simplify Prefix Build Support

skmcgrail:
### Description of changes: 
* `boringssl_prefix_symbols.h`, `boringssl_prefix_symbols_asm.h`, and `boringssl_prefix_symbols_nasm.inc` have been relocated to `openssl/boringssl_prefix_symbols_nasm.h`, `openssl/boringssl_prefix_symbols_asm.h`, and `openssl/boringssl_prefix_symbols_nasm.inc` respectively.
* Empty (stub) files for the prefix headers are now in `include/openssl` directory of the repository, these will be installed by `cmake --install <build folder>` if a prefix build was not performed.
* If the user performs a prefix build the prefix headers will be generated and will be installed in the final installation directory instead of the empty stubs.
* The `BORINGSSL_PREFIX` define is now embedded into the generated prefix headers, so the user is no longer required to provide it when linking to an aws-lc installation that was built with a prefix.
* All conditional `#include` directives for the prefix headers have been removed, as the files will always be present as either the empty stubs or the generated content.


CMakeLists.txt
```diff
@@ -150,26 +153,34 @@ if(BORINGSSL_PREFIX AND BORINGSSL_PREFIX_SYMBOLS AND GO_EXECUTABLE)
...
+           symbol_prefix_include/openssl/boringssl_prefix_symbols_asm.h
+           symbol_prefix_include/openssl/boringssl_prefix_symbols_nasm.inc
+    COMMAND ${GO_EXECUTABLE} run ${CMAKE_CURRENT_SOURCE_DIR}/util/make_prefix_headers.go -out ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/openssl -prefix ${BORINGSSL_PREFIX} ${BORINGSSL_PREFIX_SYMBOLS_PATH}
+    COMMAND sed -i.bak '/ bignum_/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/openssl/boringssl_prefix_symbols.h
+    COMMAND sed -i.bak '/ bignum_/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/openssl/boringssl_prefix_symbols_asm.h
+    COMMAND sed -i.bak '/ bignum_/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/openssl/boringssl_prefix_symbols_nasm.inc
+    COMMAND sed -i.bak '/ curve25519_x25519/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/openssl/boringssl_prefix_symbols.h
+    COMMAND sed -i.bak '/ curve25519_x25519/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/openssl/boringssl_prefix_symbols_asm.h
+    COMMAND sed -i.bak '/ curve25519_x25519/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/openssl/boringssl_prefix_symbols_nasm.inc
```
justsmth, line 165: There's no need to create the (*.bak) backups since they're immediately deleted. If you exclude the extension (`-i`) sed should just do an in-place file modification.

CMakeLists.txt
```diff
@@ -150,26 +153,34 @@ if(BORINGSSL_PREFIX AND BORINGSSL_PREFIX_SYMBOLS AND GO_EXECUTABLE)
...
-           symbol_prefix_include/boringssl_prefix_symbols_nasm.inc
-    COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include
-    COMMAND ${GO_EXECUTABLE} run ${CMAKE_CURRENT_SOURCE_DIR}/util/make_prefix_headers.go -out ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include ${BORINGSSL_PREFIX_SYMBOLS_PATH}
-    COMMAND sed -i.bak '/ bignum_/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/boringssl_prefix_symbols.h
-    COMMAND sed -i.bak '/ bignum_/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/boringssl_prefix_symbols_asm.h
-    COMMAND sed -i.bak '/ bignum_/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/boringssl_prefix_symbols_nasm.inc
-    COMMAND sed -i.bak '/ curve25519_x25519/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/boringssl_prefix_symbols.h
-    COMMAND sed -i.bak '/ curve25519_x25519/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/boringssl_prefix_symbols_asm.h
-    COMMAND sed -i.bak '/ curve25519_x25519/d' ${CMAKE_CURRENT_BINARY_DIR}/symbol_prefix_include/boringssl_prefix_symbols_nasm.inc
+    OUTPUT symbol_prefix_include/openssl/boringssl_prefix_symbols.h
```
justsmth, line 156: Since we're refactoring this process, should we rename these files to `awslc_prefix_symbols.h`?

include/openssl/boringssl_prefix_symbols.h
```diff
@@ -0,0 +1,5 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+// This file is intentionally empty, when building with the prefix enabled the
+// generated file will overwrite this one.
```
justsmth, line 5: An "overwrite" would only occur in the file placed during install after  a prefix build/installation had completed, right?  Would this statement ever be true for a consumer reading it? Perhaps the comment should be hypothetical -- "Were AWS-LC installed with symbol prefixes, this file would contain..."

CMakeLists.txt
skmcgrail: Yeah I actually looked into removing this cause I didn't think we really needed it either. I opted to keep it due to this message in the man page:
> It is not recommended to give a zero-length extension when in-place editing files, as you risk corruption or
             partial content in situations where disk space is exhausted, etc.

CMakeLists.txt
skmcgrail: I want to avoid that for now, as this at least matches the `BORINGSSL_PREFIX` define as well. I'd say that would be part of a more broader project to remove any reference to `BORINGSSL_*` defines that are all over the codebase.

justsmth (dismissed):

samuel40791765 (dismissed):

util/make_prefix_headers.go
```diff
@@ -209,6 +246,10 @@ func main() {
 		os.Exit(1)
 	}
 
+	if len(*prefix) == 0 {
+		fmt.Fprint(os.Stderr, "-prefix must be non-zero length")
```
samuel40791765, line 250: Should this do an `os.Exit(1)` like the rest below?

justsmth (approved):

samuel40791765 (approved):

PR #1179, opened 2023-09-11
Add BN_CTX_secure_new alias for BN_CTX_new

geedo0:
### Issues:
Helps with AWS-LC compatibility with strongSwan. [Discussion link](https://github.com/strongswan/strongswan/discussions/1878)

### Description of changes: 
OpenSSL defines BN_CTX_secure_new which allocates BNs using
BN_secure_new. In LC, all BNs are allocated "securely" (i.e. they 0 out
memory during alloc/free).

We already define an alias for BN_secure_new like this, so we are just
completing the compatibility layer with this change.

### Call-outs:
- Double check that BN_CTX_secure_new doesn't do anything else noteworthy.

### Testing:

CI

### DIFF
--- a/include/openssl/bn.h
+++ b/include/openssl/bn.h
@@ -1005,6 +1005,8 @@ OPENSSL_EXPORT int BN_bn2binpad(const BIGNUM *in, uint8_t *out, int len);
 // BN_secure_new calls |BN_new|.
 OPENSSL_EXPORT BIGNUM *BN_secure_new(void);
 
+// BN_CTX_secure_new calls |BN_CTX_new|.
+OPENSSL_EXPORT BN_CTX *BN_CTX_secure_new(void);
 
 // Private functions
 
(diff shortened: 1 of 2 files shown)


skmcgrail (approved):

include/openssl/bn.h
```diff
@@ -1005,6 +1005,8 @@ OPENSSL_EXPORT int BN_bn2binpad(const BIGNUM *in, uint8_t *out, int len);
 // BN_secure_new calls |BN_new|.
 OPENSSL_EXPORT BIGNUM *BN_secure_new(void);
 
+// BN_CTX_secure_new calls |BN_CTX_new|.
+OPENSSL_EXPORT BN_CTX *BN_CTX_secure_new(void);
```
justsmth, line 1009: NP: This would be better placed up around line 370 where `BN_CTX_new` is declared.

justsmth (approved):

include/openssl/bn.h
justsmth: We did recently add a [`AWS_LC_DEPRECATED`](https://github.com/aws/aws-lc/blob/166687838b9a1e063cef02cd7119c64261a851ae/include/openssl/base.h#L488) macro that makes the fact that a function is deprecated more visible.

PR #1194, opened 2023-09-14
Reset IV in AES-GCM ABI Test

nwetzler:
### Description of changes: 

This modifies the ABI test for AES-GCM to reset the `ivec` (initialization vector) in between encrypt and decrypt calls for ARM assembly.  The `ivec` buffer currently is overwritten during the `aes_gcm_enc_kernel` routine which, when passed to the `aes_gcm_dec_kernel` will produce a decrypted plaintext that does not match the original plaintext.  This change ensures that a decrypt of an encrypt produces the same buffer during testing.

### Call-outs:

Although this test is directed at the assembly interface, another suggestion would be to add an EXPECT_EQ of the input buffer and result after encrypt and decrypt to catch problems like this in the future.

### Testing:

This change was tested on an ARM machine with passing regression.

### DIFF
--- a/crypto/fipsmodule/modes/gcm_test.cc
+++ b/crypto/fipsmodule/modes/gcm_test.cc
@@ -240,13 +240,15 @@ TEST(GCMTest, ABI) {
 #if defined(OPENSSL_AARCH64) && defined(HW_GCM)
   if (hwaes_capable() && gcm_pmull_capable()) {
     static const uint8_t kKey[16] = {0};
-    uint8_t iv[16] = {0};
+    uint8_t iv[16];
 
     for (size_t key_bits = 128; key_bits <= 256; key_bits += 64) {
       AES_KEY aes_key;
       aes_hw_set_encrypt_key(kKey, key_bits, &aes_key);
+      OPENSSL_memset(iv, 0, 16);
       CHECK_ABI(aes_gcm_enc_kernel, buf, sizeof(buf) * 8, buf, X, iv, &aes_key,
                 Htable);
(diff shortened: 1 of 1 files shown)


justsmth:
Hello -- Thanks for your PR!

I'm curious what prompted this change. Was this test failing for you?  If so, we'd love to know more about the test environment that resulted in its failure so we can reproduce it.

justsmth (approved):

nwetzler:
> I'm curious what prompted this change. Was this test failing for you? If so, we'd love to know more about the test environment that resulted in its failure so we can reproduce it.

The test was not failing, but rather we wanted to see what the resulting ciphertext would be for a call of Arm assembly function `aes_gcm_enc_kernel`.  When looking at the AES-GCM ABI test where the `aes_gcm_enc_kernel` function is exercised in GDB, I noticed that the decrypt of the encrypt of a buffer containing a specific repeated value was not producing the original buffer.  I worked with a contact in the AWS-LC team to root cause the issue to the initialization vector not being properly reset.

It is my impression that the ABI tests are only intended to test the interface and not to test correctness of the function.  But the code change is rather minor to additionally have a key property of the algorithm work.

nebeid (approved):

nebeid:
We discussed internally and decided to close the PR:
- As nwetzler mentioned, the goal of the ABI test is not correctness nor self-consistency testing.
- There are other test vectors in aead_test.cc and cipher_test.cc
- The changes are particular to a certain architecture and are not replicated in other architectures.
- The changes are not blocking the requester's development.

PR #1196, opened 2023-09-15
No prefix for OPENSSL_armcap_P on FIPS

justsmth:
### Issues:
CryptoAlg-1845

### Description of changes: 
* In FIPS builds `OPENSSL_armcap_P` also should be excluded from prefixing.

### Call-outs:
N/A

### Testing:
I performed both FIPS and non-FIPS builds on a Linux/aarch64 host and checked the symbols in the resulting binary:
```
...
cpucap.c.o:
3912   │ 0000000000000000 B OPENSSL_armcap_P
3913   │ 0000000000000004 B TESTING_4__OPENSSL_cpucap_initialized
...
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,18 +145,20 @@ if(BORINGSSL_PREFIX AND BORINGSSL_PREFIX_SYMBOLS AND GO_EXECUTABLE)
     set(BORINGSSL_PREFIX_SYMBOLS_PATH ${PROJECT_BINARY_DIR}/${BORINGSSL_PREFIX_SYMBOLS})
   endif()
 
+  set(DELETE_PATTERN " bignum_| curve25519_x25519")
+  if(FIPS)
+    set(DELETE_PATTERN "${DELETE_PATTERN}| OPENSSL_armcap_P")
+  endif()
+
   add_custom_command(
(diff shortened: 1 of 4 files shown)


nebeid:
`OPENSSL_ia32cap_P` and `OPENSSL_ppc64le_hwcap2` don't need to be excluded as well?

justsmth:
> `OPENSSL_ia32cap_P` and `OPENSSL_ppc64le_hwcap2` don't need to be excluded as well?

Apparently not. We also needed to filter out this symbol when performing the [prefix build for aws-lc-rs](https://github.com/aws/aws-lc-rs/blob/main/scripts/generate/_collect_symbols_build.sh#L26).

W/o this change the build fails with a message like the following:
```
...
[ 71%] Linking C executable test_fips
inserted_by_delocate.c:2: error: undefined reference to 'OPENSSL_armcap_P'
inserted_by_delocate.c:2: error: undefined reference to 'OPENSSL_armcap_P' 
...
```

nebeid (dismissed):

justsmth:
I believe [this](https://github.com/aws/aws-lc/blob/main/util/fipstools/delocate/delocate.go#L427) needs to be updated for prefixing:
```
	var helperFunc string
	if symbol == "OPENSSL_armcap_P" {
		helperFunc = ".LOPENSSL_armcap_P_addr"
	} else {
```

PR #1199, opened 2023-09-19
Expose SHAKE through the EVP API

WillChilds-Klein:
# Issues
Resolves CryptoAlg-2034

# Notes

This commit exposes our extant SHAKE implementation through the EVP digest API and incorporates that implementation into the FIPS service indicator and ACVP tooling. SHAKE differs from traditional digest functions in that it does not have a fixed output digest size. When using the incremental EVP API, the output size is specified on finalization through `EVP_DigestFinalXOF`. This additional parameter percolates throughout the EVP and test code, as the function signature of `EVP_DigestFinalXOF` is different from that of `EVP_DigestFinal` or `EVP_DigestFinal_ex`.

We also add SHA3 to the self tests. As noted [below](https://github.com/aws/aws-lc/pull/1199#issuecomment-1743260032), while the self-tests don't include actual tests for every approved digest algorithm we support, they do cover the core implementation of every approved digest. SHA224 is covered by the SHA256 self-test; SHA384, SHA512-224, and SHA512-256 are covered by the SHA512 self-test; SHAKE-128, SHAKE-256, and all the SHA3 variants are covered by the SHA3-512 self-test.

---


crypto/fipsmodule/digest/internal.h
```diff
@@ -85,6 +85,10 @@ struct env_md_st {
   // final completes the hash and writes |md_size| bytes of digest to |out|.
   void (*final)(EVP_MD_CTX *ctx, uint8_t *out);
 
+  // finalXOF completes the hash and writes |len| bytes of digest extended output
+  // to |out|.
+  void (*finalXOF)(EVP_MD_CTX *ctx, uint8_t *out, size_t len);
```
WillChilds-Klein, line 90: Looks like our formal verification isn't happy about this struct field being added:
Screenshot 2023-09-19 at 12 19 07 PM

WillChilds-Klein: Moving this field to the end of the struct appears to have resolved the backwards incompatibility.

samuel40791765:
Only reviewed https://github.com/aws/aws-lc/pull/1199/commits/48be6d7a908f0693c7ded53ec30e95fd7e561e92, but hte service indicator changes/tests lgtm 👍

crypto/fipsmodule/digest/digest.c
```diff
@@ -269,8 +281,16 @@ int EVP_Digest(const void *data, size_t count, uint8_t *out_md,
 
   EVP_MD_CTX_init(&ctx);
   ret = EVP_DigestInit_ex(&ctx, type, impl) &&
-        EVP_DigestUpdate(&ctx, data, count) &&
-        EVP_DigestFinal_ex(&ctx, out_md, out_size);
+        EVP_DigestUpdate(&ctx, data, count);
+  if (EVP_MD_flags(type) & EVP_MD_FLAG_XOF) {
+    if (out_size == NULL) {
+      OPENSSL_PUT_ERROR(DIGEST, ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
```
andrewhop, line 287: Is this consistent with OpenSSL? This feels more like ERR_R_PASSED_NULL_PARAMETER than ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED

util/fipstools/acvp/acvptool/test/expected/SHAKE-128.bz2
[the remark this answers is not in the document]
WillChilds-Klein: I have not. I'll do that.

crypto/fipsmodule/digest/digest.c
```diff
@@ -269,8 +281,16 @@ int EVP_Digest(const void *data, size_t count, uint8_t *out_md,
 
   EVP_MD_CTX_init(&ctx);
   ret = EVP_DigestInit_ex(&ctx, type, impl) &&
-        EVP_DigestUpdate(&ctx, data, count) &&
-        EVP_DigestFinal_ex(&ctx, out_md, out_size);
+        EVP_DigestUpdate(&ctx, data, count);
+  if (EVP_MD_flags(type) & EVP_MD_FLAG_XOF) {
```
dkostic, line 285: can we extract `EVP_MD_flags(type) & EVP_MD_FLAG_XOF` to a function `evp_md_is_xof` and call the function instead of having the expression?

crypto/fipsmodule/sha/internal.h
```diff
@@ -126,6 +126,14 @@ OPENSSL_EXPORT uint8_t *SHAKE128(const uint8_t *data, const size_t in_len,
 OPENSSL_EXPORT uint8_t *SHAKE256(const uint8_t *data, const size_t in_len, 
                                  uint8_t *out, size_t out_len);
 
+// SHAKE_Init initializes |ctx| with specified |block_size|, returns 1 on
+// success and 0 on failure. Calls SHA3_Init under the hood.
+OPENSSL_EXPORT int SHAKE_Init(KECCAK1600_CTX *ctx, size_t block_size);
+
+// SHAKE_Init writes |len| bytes of finalized digest to |md|, returns 1 on
```
dkostic, line 133:
```suggestion
// SHAKE_Final writes |len| bytes of finalized digest to |md|, returns 1 on
```

darylmartin100 (commented):
Since you are marking this as an approved algorithm, do we need to add a corresponding self test in self_check.c?

crypto/fipsmodule/digest/internal.h
```diff
@@ -90,6 +90,10 @@ struct env_md_st {
 
   // ctx_size contains the size, in bytes, of the state of the hash function.
   unsigned ctx_size;
+
+  // finalXOF completes the hash and writes |len| bytes of digest extended output
+  // to |out|.
+  void (*finalXOF)(EVP_MD_CTX *ctx, uint8_t *out, size_t len);
```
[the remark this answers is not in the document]
WillChilds-Klein: that makes sense. additionally, it's just a function pointer, so shouldn't take up more than 4-8 bytes per MD function we support, right?

crypto/fipsmodule/digest/digest.c
WillChilds-Klein: Hm, my view is that this pattern is common, safe, and simple enough that it doesn't really justify its own function. I don't feel very strongly about this, though. What do you think?

WillChilds-Klein:
> Since you are marking this as an approved algorithm, do we need to add a corresponding self test in self_check.c?

@darylmartin100 --That's a good question. Right now, we're missing coverage in `self_check.c` for the following approved hash algorithms:

- SHA2: SHA384, SHA512-224, SHA512-256
- SHA3: SHA3-224, SHA3-256, SHA3-384, SHA3-512
- SHAKE: SHAKE-128, SHAKE-256

I'm not sure what the decision criterion is for adding an algorithm to `self_check.c`, but it seems like if we add SHAKE, we should add the other missing ones as well.

dkostic:
one general question: can the `finalXOF` function be called more than once on the same context object and what's the expected behavior in that case?

WillChilds-Klein:
> one general question: can the `finalXOF` function be called more than once on the same context object and what's the expected behavior in that case?

@dkostic -- I don't think we should allow that (neither [does OpenSSL](https://github.com/openssl/openssl/blob/ad4af6dfca8344516bb658b1745a530635af9433/crypto/evp/digest.c#L515C23-L526)). For regular digests, we technically [allow](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/digest/digest.c#L245-L250) multiple calls to `EVP_DigestFinal_ex`, but [advise against it](https://github.com/aws/aws-lc/blob/main/include/openssl/digest.h#L179-L181) in our documentation. `EVP_DigestFinal` [cleans up the `ctx`](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/digest/digest.c#L261), so it [can't](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/digest/digest.c#L245) be called again without re-initialization.

I've added a test case to `digest_test.cc` to assert this behavior.

crypto/fipsmodule/digest/digest.c
```diff
@@ -131,8 +131,17 @@ void EVP_MD_CTX_free(EVP_MD_CTX *ctx) {
...
+  if (ctx->digest == NULL) {
+    return 0;
+  }
+  if ((EVP_MD_flags(ctx->digest) & EVP_MD_FLAG_XOF) == 0) {
+    OPENSSL_PUT_ERROR(DIGEST, ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
+    return 0;
+  }
+  ctx->digest->finalXOF(ctx, out, len);
+  OPENSSL_cleanse(ctx->md_data, ctx->digest->ctx_size);
+  EVP_MD_CTX_cleanup(ctx);
```
dkostic, line 143:
```suggestion
  EVP_MD_CTX_cleanse(ctx);
```

crypto/fipsmodule/digest/digest.c
```diff
@@ -269,8 +281,16 @@ int EVP_Digest(const void *data, size_t count, uint8_t *out_md,
...
-        EVP_DigestFinal_ex(&ctx, out_md, out_size);
+        EVP_DigestUpdate(&ctx, data, count);
+  if (EVP_MD_flags(type) & EVP_MD_FLAG_XOF) {
+    if (out_size == NULL) {
+      OPENSSL_PUT_ERROR(DIGEST, ERR_R_PASSED_NULL_PARAMETER);
+      return 0;
+    }
+    ret &= EVP_DigestFinalXOF(&ctx, out_md, *out_size);
+  } else {
+    ret &= EVP_DigestFinal_ex(&ctx, out_md, out_size);
```
dkostic, line 292: `EVP_DigestFinalXOF` cleans up the context and `EVP_DigestFinal_ex` doesn't but the context is cleaned in line 294. Why not use `EVP_DigestFinal` instead and remove line 294?

WillChilds-Klein: I was preserving prior code, but I agree that your suggested changes are functionally equivalent and cleaner. will update.

crypto/fipsmodule/digest/digest.c
```diff
@@ -269,9 +280,16 @@ int EVP_Digest(const void *data, size_t count, uint8_t *out_md,
...
   ret = EVP_DigestInit_ex(&ctx, type, impl) &&
-        EVP_DigestUpdate(&ctx, data, count) &&
-        EVP_DigestFinal_ex(&ctx, out_md, out_size);
-  EVP_MD_CTX_cleanup(&ctx);
+        EVP_DigestUpdate(&ctx, data, count);
+  if (EVP_MD_flags(type) & EVP_MD_FLAG_XOF) {
+    if (out_size == NULL) {
+      OPENSSL_PUT_ERROR(DIGEST, ERR_R_PASSED_NULL_PARAMETER);
+      return 0;
+    }
```
andrewhop, line 288: Can this be moved up to line 280?

WillChilds-Klein: Sure thing. I've refactored this and added a test.

andrewhop (approved):

dkostic (approved):

PR #1201, opened 2023-09-20
Add HybridKeyShare support for SecP256r1Kyber768Draft00 and X25519Kyber768Draft00

alexw91:
### Issues:
Resolves `CryptoAlg-2073` and `CryptoAlg-2074`

Note: This PR is heavily based on [this previously abandoned PR](https://github.com/aws/aws-lc/pull/875)

### Description of changes: 
This PR makes the following changes:
 1. Adds new `KEMKeyShare` and `HybridKeyShare` classes to `ssl_key_share.cc`.
 2. Adds definitions for new Hybrid Post-Quantum SupportedGroups `SecP256r1Kyber768Draft00` and `X25519Kyber768Draft00`
     -  [`SecP256r1Kyber768Draft00` Draft RFC](https://datatracker.ietf.org/doc/html/draft-kwiatkowski-tls-ecdhe-kyber)
     - [`X25519Kyber768Draft00` Draft RFC](https://datatracker.ietf.org/doc/html/draft-tls-westerbaan-xyber768d00)
 3. Adds various self-talk unit tests that negotiates these new SupportedGroups

These hybrid PQ groups are still opt-in, and have [NOT been added to the groups that will be offered or accepted by default.](https://github.com/aws/aws-lc/blob/dabfbbcb7a577b2fd9ba0ead5b14782133ba2724/ssl/extensions.cc#L304)

### Call-outs:
Integration tests and automated interoperability tests with other TLS libraries are [in progress in this s2n-tls PR](https://github.com/aws/s2n-tls/pull/4267). Manual interoperability tests confirm that these new Hybrid groups work with both BoringSSL and s2n-tls. 

### Testing:
1. AWS-LC self-talk unit tests.
2. Confirmed interoperability with BoringSSL for `X25519Kyber768Draft00`
3. Confirmed interoperability with s2n-tls for both `SecP256r1Kyber768Draft00` and `X25519Kyber768Draft00`


ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,550 @@ class X25519KeyShare : public SSLKeyShare {
...
+    // ctx_ should have been initialized by a previous call to Offer()
+    if (!ctx_) {
+      OPENSSL_PUT_ERROR(SSL, ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
+      return false;
+    }
+
+    // Retrieve the lengths of the ciphertext and shared secret
+    size_t ciphertext_len = 0;
+    size_t secret_len = 0;
+    if (!EVP_PKEY_encapsulate(ctx_.get(), nullptr, &ciphertext_len, nullptr,
```
andrewhop, line 375: NP: you can do the same thing with decapsulate and it looks just a little strange to call encapsulate before decapsulate.

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,550 @@ class X25519KeyShare : public SSLKeyShare {
...
+      OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
+      return false;
+    }
+
+    // |peer_key| is the ciphertext, encapsulating the shared secret,
+    // that the peer generated using our public key. We use our private
+    // key to decapsulate it, and retain the shared secret by writing
+    // it to |out_secret|.
+    uint8_t *ciphertext = (uint8_t *)peer_key.begin();
+    if (!out_secret->Init(secret_len)) {
```
andrewhop, line 393: Any reason you initialized out_secret here and wrote to it directly instead of moving the final result like other places above?

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,550 @@ class X25519KeyShare : public SSLKeyShare {
...
+  uint16_t group_id_;
+  UniquePtr<EVP_PKEY_CTX> ctx_;
+};
+
+// A HybridKeyShare consists of key shares from two or more component groups,
+// all of which are used to generate a hybrid shared secret.
+// See https://datatracker.ietf.org/doc/html/draft-ietf-tls-hybrid-design.
+  class HybridKeyShare : public SSLKeyShare {
+  public:
+    HybridKeyShare(uint16_t group_id) : group_id_(group_id), hybrid_group_(nullptr) {
```
andrewhop, line 423: How do you initialize a hybrid keyshare if you always set the hybrid_group to null?

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,550 @@ class X25519KeyShare : public SSLKeyShare {
...
+      for (UniquePtr<SSLKeyShare> &key_share : key_shares_) {
+        key_share = nullptr;
+      }
```
andrewhop, line 426: Any reason we have to clear this state but not in other key share constructors?

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,550 @@ class X25519KeyShare : public SSLKeyShare {
...
+
+      // Success; clear the alert
+      *out_alert = 0;
+      return true;
+    }
+
+  private:
+   bool init_component_key_shares() {
+      // The component key shares should only be initialized once
```
andrewhop, line 625: Should this be done in the constructor?

ssl/internal.h
```diff
@@ -1197,12 +1198,40 @@ class SSLKeyShare {
...
+  const char name[32], alias[32];
 };
 
 // NamedGroups returns all supported groups.
 Span<const NamedGroup> NamedGroups();
 
+// Hybrid groups adhere to the draft specification:
+// https://datatracker.ietf.org/doc/html/draft-ietf-tls-hybrid-design
+#define NUM_HYBRID_COMPONENTS 2
+  struct HybridGroup {
```
andrewhop, line 1210: Indentation looks a little off here.

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,550 @@ class X25519KeyShare : public SSLKeyShare {
...
+    SSL_GROUP_X25519_KYBER768_DRAFT00
+};
+
+// The key share sizes are taken from the corresponding specification.
+//
+// Kyber Round 3: https://pq-crystals.org/kyber/data/kyber-specification-round3-20210804.pdf
+// SECP256R1:     https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8.2
+// X25519:        https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8.2
+static const size_t p256_share_size = ((32 * 2) + 1);
+static const size_t x25519_share_size = 32;
```
andrewhop, line 671: Should the hybrid key share care about the size of it's sub-key shares? Or just let them use read/write as much as they need and let the sub-keyshare throw an error if the size is wrong?

andrewhop (commented):
I have not looked at ssl_test.cc yet.

include/openssl/ssl.h
```diff
@@ -2561,6 +2561,15 @@ OPENSSL_EXPORT int SSL_set1_groups_list(SSL *ssl, const char *groups);
 #define SSL_GROUP_SECP384R1 24
 #define SSL_GROUP_SECP521R1 25
 #define SSL_GROUP_X25519 29
+#define SSL_GROUP_SECP256R1_KYBER768_DRAFT00 0x639A
```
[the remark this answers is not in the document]
alexw91: Yes, the draft RFC's state that this is for the 3rd round of Kyber.
 - https://datatracker.ietf.org/doc/html/draft-kwiatkowski-tls-ecdhe-kyber
  - https://datatracker.ietf.org/doc/html/draft-tls-westerbaan-xyber768d00

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,550 @@ class X25519KeyShare : public SSLKeyShare {
...
+      return false;
+    }
+
+    // Generate a KEM keypair and retrieve the length of the public key
+    size_t public_key_len = 0;
+    EVP_PKEY *raw_key = nullptr;
+    if (!EVP_PKEY_keygen_init(ctx_.get()) ||
+        !EVP_PKEY_keygen(ctx_.get(), &raw_key) ||
+        !raw_key ||
+        !EVP_PKEY_get_raw_public_key(raw_key, nullptr, &public_key_len)) {
```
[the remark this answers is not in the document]
alexw91: Calling `EVP_PKEY_get_raw_public_key` with the 2nd argument set to null results in `public_key_len` being populated with the length of the public key. 

`If the buffer pub is NULL then *len is populated with the number of bytes required to hold the key. ` - https://www.openssl.org/docs/man3.1/man3/EVP_PKEY_get_raw_public_key.html

include/openssl/ssl.h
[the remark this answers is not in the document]
andrewhop: I meant should the name include r3 in the name like `SSL_GROUP_SECP256R1_KYBER768r3_DRAFT00`

ssl/ssl_key_share.cc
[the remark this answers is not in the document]
andrewhop: Ahhh and you need the size down below.

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,550 @@ class X25519KeyShare : public SSLKeyShare {
   uint8_t private_key_[32];
 };
 
+class KEMKeyShare : public SSLKeyShare {
```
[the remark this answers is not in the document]
alexw91: I might be wrong, but I don't think there's a reason to implement these API's. The only locations that I can find that use the `SerializePrivateKey`/`DeserializePrivateKey` API's are related to experimental API's: [handshake hints](https://github.com/google/boringssl/commit/b571e77773ef875f7ba9f1d2bba81f305152703e#diff-b909baec2e9ef4ed52c083aa5c5dede6956b01327a5e71b3331cd06d540d351bR3801-R3809) and [split handshaking](https://github.com/aws/aws-lc/commit/3fe8fa74acf0a7bcd8ab5ad08698e8a99a667703#diff-b909baec2e9ef4ed52c083aa5c5dede6956b01327a5e71b3331cd06d540d351bR4462-R4504) (that has been removed from upstream BorinSSL but is still in AWS-LC. I'm not aware of users of either of these experimental API's in AWS-LC.

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,550 @@ class X25519KeyShare : public SSLKeyShare {
...
+ public:
+  KEMKeyShare(int nid, uint16_t group_id) : nid_(nid), group_id_(group_id) {}
+
+  uint16_t GroupID() const override { return group_id_; }
+
+  bool Offer(CBB *out) override {
```
dkostic, line 187: Boring changed the function names to `Encap, Decap, Generate` which makes much more sense in the KEM context (https://github.com/google/boringssl/blob/master/ssl/internal.h#L1107-L1121). Can we do something similar?

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,550 @@ class X25519KeyShare : public SSLKeyShare {
...
+  bool Accept(CBB *out_public_key, Array<uint8_t> *out_secret,
+              uint8_t *out_alert, Span<const uint8_t> peer_key) override {
+    // Basic nullptr checks
+    if (!out_alert) {
+      OPENSSL_PUT_ERROR(SSL, ERR_R_PASSED_NULL_PARAMETER);
+      return false;
+    }
+
+    // Set alert to internal error by default
+    *out_alert = SSL_AD_INTERNAL_ERROR;
```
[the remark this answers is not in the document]
alexw91: [A TLS alert is only sent if the `Accept()` call fails and returns false](https://github.com/aws/aws-lc/blob/main/ssl/tls13_server.cc#L80-L82). This pattern of initializing alert to `SSL_AD_INTERNAL_ERROR` at the start of the `Accept()` call is [part of the base SSLKeyShare implementation.](https://github.com/aws/aws-lc/blob/b405bb968846ec3b7ff2fc5b0307c812f41b1260/ssl/ssl_key_share.cc#L210-L215)

ssl/ssl_key_share.cc
alexw91: For encapsulate, `ciphertext_len` is an output of the function. For decapsulate, `ciphertext_len` is an input to the function.

andrewhop: Not sure I get it, aren't you calling encapsulate with null to get the size? You can do that with [EVP_PKEY_decapsulate](https://github.com/aws/aws-lc/blob/main/include/openssl/evp.h#L750-L752) too.

ssl/ssl_key_share.cc
alexw91: I believe the other places that perform the `std::move()` pattern to ensure that in the event of any of the function calls between `shared_secret.Init()` and the Encapsulate/Decapsulate failing (such as CBB memory allocations), that `out_secret` isn't modified. Here, there aren't multiple function calls between those two points, so the move pattern isn't used.

include/openssl/ssl.h
```diff
@@ -2561,6 +2561,15 @@ OPENSSL_EXPORT int SSL_set1_groups_list(SSL *ssl, const char *groups);
...
+#define SSL_GROUP_X25519_KYBER768_DRAFT00 0x6399
+
+// PQ and hybrid group IDs are not yet standardized. Current IDs are driven by
+// community consensus and are defined at
+// https://github.com/open-quantum-safe/openssl/blob/OQS-OpenSSL_1_1_1-stable/oqs-template/oqs-kem-info.md
```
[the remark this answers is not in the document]
alexw91: Fixed

ssl/internal.h
```diff
@@ -1197,12 +1198,40 @@ class SSLKeyShare {
...
+    // identifies the group as a whole unit...
+    uint16_t group_id;
+
+    // ...and each component of the hybrid group has its own group ID
+    uint16_t component_group_ids[NUM_HYBRID_COMPONENTS];
+
+    // The sizes, in bytes, of the key shares that are output by each
+    // component's SSLKeyShare::Offer() and SSLKeyShare::Accept() fucntions
+    size_t offer_share_sizes[NUM_HYBRID_COMPONENTS];
+    size_t accept_share_sizes[NUM_HYBRID_COMPONENTS];
```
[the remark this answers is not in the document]
alexw91: I was able to remove these constants into a private lookup function that's called at runtime. `get_component_offer_key_share_size()` and `get_component_accept_key_share_size()` have been added to the `HybridKeyShare` class

ssl/ssl_key_share.cc
alexw91: This was initialized by the `init_component_key_shares()` function. This function has been moved to the class constructor

ssl/ssl_key_share.cc
alexw91: I'm happy to make this change, but I'd prefer to do so in a separate Pull Request

ssl/ssl_key_share.cc
alexw91: The ECKeyShare will fail with an error if the KeyShare data passed to it is too large, so the HybridKeyShare class needs to know where to split the two component KeyShares in it to avoid those errors.

ssl/ssl_key_share.cc
alexw91: This was just initializing these values to null in the constructor to avoid them containing uninitialized data. This initialization has been moved to the constructor now in the latest update

include/openssl/ssl.h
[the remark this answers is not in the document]
alexw91: My preference is to not include "r3" in the name. This keeps the variable name closer to the draft standard name, and it also matches the naming convention for [`SSL_GROUP_X25519_KYBER768_DRAFT00`](https://github.com/google/boringssl/blob/master/include/openssl/ssl.h#L2365) that Boringssl is already using which will make it easier for future upstream imports.

ssl/ssl_key_share.cc
alexw91: +1, moved this function into the constructor

ssl/internal.h
alexw91: Fixed

ssl/ssl_key_share.cc
alexw91: Yes, [the call to `EVP_PKEY_encapsulate()` with NULL values is setting the sizes of both **`ciphertext_len`** and **`secret_len`**](https://github.com/aws/aws-lc/blob/main/crypto/evp_extra/p_kem.c#L76-L77). 

However the type signatures for `EVP_PKEY_encapsulate()` and `EVP_PKEY_decapsulate()` are different. [Calling `EVP_PKEY_decapsulate()` with NULL values will only set **`shared_secret_len`**.](https://github.com/aws/aws-lc/blob/main/crypto/evp_extra/p_kem.c#L139)

`EVP_PKEY_encapsulate()` has a `size_t *ciphertext_len` so it can modify the callers value of `ciphertext_len` that the pointer is pointing at, whearas `EVP_PKEY_decapsulate()` only has a `size_t ciphertext_len` so any modifications can't be passed up to the caller.

```
int EVP_PKEY_encapsulate(EVP_PKEY_CTX *ctx,
                         uint8_t *ciphertext, size_t *ciphertext_len,
                         uint8_t *shared_secret, size_t *shared_secret_len);
```
and 

```
int EVP_PKEY_decapsulate(EVP_PKEY_CTX *ctx,
                         uint8_t *shared_secret, size_t *shared_secret_len,
                         uint8_t *ciphertext, size_t ciphertext_len);
```

ssl/ssl_test.cc
```diff
@@ -10071,6 +10525,150 @@ TEST(SSLTest, NameLists) {
   }
 }
 
+// Test a successful round-trip for KemKeyShare
```
andrewhop, line 10528: Are these tests operating at the right abstraction level? All the other tests seem to be using `CreateClientAndServer/ConnectClientAndServer/CompleteHandshakes`. I see this is why you changed the visibility in internal.h. I think these are good tests to have, but are they testing something different than what actually happens?

ssl/ssl_test.cc
```diff
@@ -10071,6 +10525,150 @@ TEST(SSLTest, NameLists) {
...
+TEST(SSLTest, KemKeyShare) {
+  for (const GroupTest &t : kKemGroupTests) {
```
andrewhop, line 10530: If you want nicer test output and filtering you could make this a [parameterized test](https://github.com/google/googletest/blob/main/docs/advanced.md#value-parameterized-tests).

ssl/ssl_test.cc
```diff
@@ -10118,5 +10716,1082 @@ TEST(SSLTest, SessionPrint) {
...
+      EXPECT_TRUE(client_key_share->Finish(&client_secret, &client_alert,
+                                           server_public_key));
+
+      // The shared secrets are of the correct length...
+      EXPECT_EQ(client_secret.size(), t.shared_secret_size);
+      EXPECT_EQ(server_secret.size(), t.shared_secret_size);
+
+      // ... but they are not equal
+      EXPECT_NE(0, OPENSSL_memcmp(client_secret.data(), server_secret.data(),
+                                  t.shared_secret_size));
```
andrewhop, line 10952: Reading these secrets probably isn't valuable but for the future we have a `Bytes` test class which pretty prints any failures and implements comparable. 
```suggestion
      // ... but they are not equal
      EXPECT_NE(Bytes(client_secret), Bytes(server_secret));
```

ssl/ssl_key_share.cc
dkostic: but that's because for `encapsulate` both the ciphertext and the shared secret are output arguments. For `decapsulate`, the shared secret is an output argument and the ciphertext is an input argument.

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,569 @@ class X25519KeyShare : public SSLKeyShare {
...
+
+      // Accept() each component key share. Each component's Accept() function
+      // will generate a shared secret and a public key to be sent back to
+      // the peer. The hybrid public key is the concatenation of all component
+      // public keys; the hybrid shared secret is the concatenation of all
+      // component shared secrets.
+      size_t peer_key_index = 0;
+      for (size_t i = 0; i < NUM_HYBRID_COMPONENTS; i++) {
+        size_t component_key_size = 0;
+        if (!get_component_offer_key_share_size(&component_key_size, hybrid_group_->component_group_ids[i])) {
```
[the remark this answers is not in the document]
alexw91: Boringssl only supports a single combination of x25519 + Kyber768, and they just hardcoded each value. 
 - https://github.com/google/boringssl/blob/master/ssl/ssl_key_share.cc#L203-L210

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,569 @@ class X25519KeyShare : public SSLKeyShare {
...
+  UniquePtr<EVP_PKEY_CTX> ctx_;
+};
+
+// The key share sizes are taken from the corresponding specification.
+//
+// Kyber Round 3: https://pq-crystals.org/kyber/data/kyber-specification-round3-20210804.pdf
+// SECP256R1:     https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8.2
+// X25519:        https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8.2
+static const size_t p256_uncompressed_share_size = ((32 * 2) + 1);
+static const size_t x25519_share_size = 32;
```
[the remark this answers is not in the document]
alexw91: Not as far as I can tell, but I can create a new `#define` for it. 

The EC code defines the [size of a single EC point coordinate length for P256 here](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/ec/ec.c#L283), but the [TLS KeyShare spec requires both an X and Y coordinate and a 1 byte header](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8.2). The x25519 code has `32` hardcoded everywhere.

ssl/ssl_key_share.cc
```diff
@@ -231,7 +790,7 @@ bool ssl_name_to_group_id(uint16_t *out_group_id, const char *name, size_t len)
       *out_group_id = group.group_id;
       return true;
     }
-    if (len == strlen(group.alias) &&
+    if ((strlen(group.alias) > 0) && len == strlen(group.alias) &&
```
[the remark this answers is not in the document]
alexw91: There are multiple naming conventions for the NIST curves, and [the `NamedGroups` allows for a primary name as well as a secondary alias name.](https://github.com/aws/aws-lc/blob/main/ssl/internal.h#L1200). There aren't any secondary alias names for the Hybrid curves, so an empty string was used as the alias name for the Hybrid groups, but `ssl_name_to_group_id()` doesn't correctly handle zero-length aliases, resulting in a zero-length alias matching always matching every group name.

[BoringSSL found this same bug in `ssl_name_to_group_id()` when they were adding Hybrid PQ groups.](https://github.com/google/boringssl/commit/1b724a6b8cefd6d3facf7141222bca57f9e090c7#diff-b4d1fee360b068e022d5ca8c8712cee12257f1666de6c9b91718cf8823c8c3dbR338)

ssl/ssl_test.cc
```diff
@@ -580,8 +707,335 @@ static const char *kBadCurvesLists[] = {
...
+    TLS1_3_VERSION,
+    "X25519Kyber768Draft00:secp384r1:x25519:SecP256r1Kyber768Draft00",
+    TLS1_3_VERSION,
+    SSL_GROUP_SECP256R1_KYBER768_DRAFT00,
+    false,
+  },
+
+  // The client lists PQ/hybrid groups as both first and second preferences.
+  // The key share logic is implemented such that the client will always
+  // attempt to send one hybrid key share and one classical key share.
```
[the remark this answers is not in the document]
alexw91: https://github.com/aws/aws-lc/blob/31a322e7256867f5f12e4b25014188878a5867b5/ssl/extensions.cc#L2240-L2260

ssl/ssl_test.cc
```diff
@@ -580,8 +707,335 @@ static const char *kBadCurvesLists[] = {
...
+    "prime256v1:x25519",
+    TLS1_2_VERSION,
+    SSL_GROUP_SECP256R1,
+    false,
+  },
+
+  // Same as above, but server also has SecP256r1Kyber768Draft00 in it's
+  // supported list, but can't use it since TLS 1.3 is the minimum version that
+  // supports PQ.
+  {
```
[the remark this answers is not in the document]
alexw91: This is in line with existing behavior. Configuring Curves/Groups that are impossible to negotiate due to the TLS Ciphersuites only allowing RSA key exchange still results in a successful TLS connection that does not use those curves/groups. 

```
# Configure client to only use RSA Key Exchange, but also provide curve preference list that will not be used
$ bssl s_client -connect google.com:443 -max-version tls1.2 -cipher AES128-GCM-SHA256 -curves prime256v1:secp384r1
Connecting to 142.251.33.110:443
Connected.
  Version: TLSv1.2
  Resumed session: no
  Cipher: TLS_RSA_WITH_AES_128_GCM_SHA256
  Secure renegotiation: yes
  Extended master secret: yes
  Next protocol negotiated: 
  ALPN protocol: 
  OCSP staple: no
  SCT list: no
  Early data: no
  Encrypted ClientHello: no
  Cert subject: CN = *.google.com
  Cert issuer: C = US, O = Google Trust Services LLC, CN = GTS CA 1C3
```

```
# Configure client to only use RSA Key Exchange, but also provide curve preference list that will not be used
$ openssl s_client -connect google.com:443 -tls1_2 -cipher AES128-GCM-SHA256 -curves prime256v1:secp384r1 -brief
CONNECTION ESTABLISHED
Protocol version: TLSv1.2
Ciphersuite: AES128-GCM-SHA256
Peer certificate: CN = *.google.com
Verification: OK
```

ssl/ssl_test.cc
```diff
@@ -580,8 +707,335 @@ static const char *kBadCurvesLists[] = {
...
+    TLS1_2_VERSION,
+    SSL_GROUP_SECP256R1,
+    false,
+  },
+
+  // If the client configures the curve list to include a hybrid
+  // curve, then initiates a 1.2 handshake, it will not advertise
+  // hybrid groups because hybrid is not supported for 1.2. So
+  // a 1.2 EC handshake will be negotiated (even if the server
+  // supports 1.3 with corresponding hybrid group).
```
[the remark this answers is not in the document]
alexw91: Same as above. People today can configure curves/groups with ciphersuites that only allow RSA key exchange. If they do, then the groups will be silently ignored.

ssl/ssl_test.cc
```diff
@@ -580,8 +707,335 @@ static const char *kBadCurvesLists[] = {
...
+  // preferences, in order, as [SecP256r1Kyber768Draft00, X25519Kyber768Draft00,
+  // x25519, prime256v1]. From those groups, it will send key shares
+  // [SecP256r1Kyber768Draft00, x25519]. The server supports, and receives a
+  // key share for, x25519. However, when selecting a mutually supported group
+  // to negotiate, the server recognizes that the client prefers
+  // X25519Kyber768Draft00 over x25519. Since the server also supports
+  // X25519Kyber768Draft00, but did not receive a key share for it, it will
+  // select it and send an HRR. This ensures that the client's highest
+  // preference group will be negotiated, even at the expense of an additional
+  // round-trip.
```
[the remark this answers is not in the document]
alexw91: 1. AWS-LC currently prefers to use the client preference ordering when selecting an algorithm. These tests don't set `SSL_OP_CIPHER_SERVER_PREFERENCE`, but enabling that option just effectively swaps which preference list the selection code assumes is the client's and server's.
2. For TLS 1.2, the ciphersuite name includes the key exchange type (RSA/ECDHE). For TLS 1.3 the ciphersuite only selects the bulk encryption algorithm (AES128-GCM, AES256-GCM, and ChaChaPoly) and does not restrict which KeyExchange group to use, and either ciphersuites or groups can be selected first. ([AWS-LC currently selects SupportedGroups, and then Ciphersuites for TLS 1.3](https://github.com/aws/aws-lc/blob/e673f6e508cc5887191f89ad04df308f46a6c809/ssl/tls13_server.cc#L228-L239), and [even though the selected group is passed into the TLS 1.3 selection function, it's not actually used.](https://github.com/aws/aws-lc/blob/e673f6e508cc5887191f89ad04df308f46a6c809/ssl/s3_both.cc#L688-L728))
3. There is a 3-way tradeoff between security, performance, and respecting the client's wishes. AWS-LC can choose between selecting the SupportedGroup with the highest security level, the group that requires the fewest round trips, or the group that is highest on the client's preference list. For TLS ciphersuites and signature algorithms, AWS-LC selects the algorithm that is the highest mutually supported algorithm in the client's preference list, and the same decision to respect the client's preference order is made here for SupportedGroups. In that sense, I think this tradeoff to respect the client's preference is worth it even if it results in an extra round trip.

ssl/ssl_test.cc
```diff
@@ -580,8 +707,335 @@ static const char *kBadCurvesLists[] = {
...
+    "SecP256r1Kyber768Draft00",
+    TLS1_3_VERSION,
+    "X25519Kyber768Draft00",
+    TLS1_3_VERSION,
+    0,
+    false,
+  },
+
+  // If the client supports hybrid TLS 1.3, but the server
+  // only supports TLS 1.2, then TLS 1.2 EC should be negotiated.
```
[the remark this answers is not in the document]
alexw91: Locally updated this test to assert that the actual TLS version negotiated is always equal to `std::min(t.client_version, t.server_version)`. Will include that change in an update to this PR soon.

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,569 @@ class X25519KeyShare : public SSLKeyShare {
...
+      OPENSSL_PUT_ERROR(SSL, ERR_R_MALLOC_FAILURE);
+      return false;
+    }
+
+    size_t secret_bytes_written = secret_len;
+    if (!EVP_PKEY_decapsulate(ctx_.get(), shared_secret.data(),
+                              &secret_bytes_written, ciphertext,
+                              ciphertext_len) ||
+                              secret_bytes_written != secret_len) {
+      OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
```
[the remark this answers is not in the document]
alexw91: [Updated code to return `SSL_R_BAD_KEM_CIPHERTEXT`](https://github.com/aws/aws-lc/pull/1201/commits/29515a994c202712cafa3a4682515e44ccd951e2#diff-b4d1fee360b068e022d5ca8c8712cee12257f1666de6c9b91718cf8823c8c3dbR401)

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,569 @@ class X25519KeyShare : public SSLKeyShare {
...
+    if (!EVP_PKEY_encapsulate(ctx_.get(), nullptr, &ciphertext_len, nullptr,
+                              &secret_len)) {
+      OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
+      return false;
+    }
+
+    // Ensure that peer_key is valid
+    if (!peer_key.data() || peer_key.size() != ciphertext_len) {
+      *out_alert = SSL_AD_DECODE_ERROR;
+      OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
```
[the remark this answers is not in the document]
alexw91: Fixed

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,569 @@ class X25519KeyShare : public SSLKeyShare {
...
+        if (!get_component_offer_key_share_size(&component_key_size, hybrid_group_->component_group_ids[i])) {
+          OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
+          return false;
+        }
+
+        // Verify that |peer_key| contains enough data
+        if (peer_key_index + component_key_size > peer_key.size()) {
+          CBB_cleanup(&hybrid_shared_secret);
+          *out_alert = SSL_AD_DECODE_ERROR;
+          OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
```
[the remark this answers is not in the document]
alexw91: Fixed

ssl/ssl_key_share.cc
```diff
@@ -176,20 +178,569 @@ class X25519KeyShare : public SSLKeyShare {
...
+//
+// Kyber Round 3: https://pq-crystals.org/kyber/data/kyber-specification-round3-20210804.pdf
+// SECP256R1:     https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8.2
+// X25519:        https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8.2
+static const size_t p256_uncompressed_share_size = ((32 * 2) + 1);
+static const size_t x25519_share_size = 32;
+
+// Kyber Round 3: https://pq-crystals.org/kyber/data/kyber-specification-round3-20210804.pdf
+static const size_t kyber768_r3_public_key_size = 1184;
+static const size_t kyber768_r3_ciphertext_size = 1088;
```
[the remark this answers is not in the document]
alexw91: [Defined new Kyber constants here, and am now using them in both `ssl_key_share.cc` and `ssl_test.cc`](https://github.com/aws/aws-lc/pull/1201/commits/29515a994c202712cafa3a4682515e44ccd951e2#diff-65e9a26b94926cae6214ce303f9e075bbee8e548e8ebe30cb39009b6edd3011eR12-R24)

ssl/ssl_key_share.cc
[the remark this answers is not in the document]
alexw91: [Added new Elliptic curve field element size constants here, and am using these constants in both the KeyShare and testing code.](https://github.com/aws/aws-lc/pull/1201/commits/29515a994c202712cafa3a4682515e44ccd951e2#diff-9053bbc88f62c34ad85263cee90396124b4a5200a2cc390e8525e067aaeb4d60R103-R107)

ssl/ssl_test.cc
```diff
@@ -569,6 +604,98 @@ static const CurveTest kCurveTests[] = {
...
+
+static const size_t x25519_key_share_size = 32;
+static const size_t x25519_secret_size = 32;
+
+// The key share sizes are taken from the corresponding specification.
+//
+// Kyber Round 3: https://pq-crystals.org/kyber/data/kyber-specification-round3-20210804.pdf
+static const size_t kyber768_r3_offer_key_share_size = 1184;
+static const size_t kyber768_r3_accept_key_share_size = 1088;
+static const size_t kyber768_r3_secret_size = 32;
```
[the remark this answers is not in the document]
alexw91: Fixed

ssl/ssl_test.cc
alexw91: Updated all new tests to be parameterized

ssl/ssl_test.cc
alexw91: Followed this pattern in all new test code that I have added.

ssl/ssl_test.cc
alexw91: If it is valuable to test the failure paths of how we handle unexpected input of the new HybridKeyShare and KemKeyShare classes that we are adding, then I'd argue that these tests are at the correct abstraction level. We still have coverage of the high level "happy path" using the API's you mentioned in [PerformHybridHandshakeTest](https://github.com/aws/aws-lc/pull/1201/files#diff-39cb80e1b51b8f538bf9667decca8d5f25849138ec5623ddece489113dc3c601R11767), but we also have coverage that these SSLKeyShares will fail on unexpected input how we expect them to. 

If the AWS-LC team would prefer an alternate approach that can test these function without exporting them, or not to test these error paths at all, I can also make those changes.

ssl/ssl_key_share.cc
andrewhop: In this case I don't think you need the ciphertext length, the KEM decapsulate checks this already https://github.com/aws/aws-lc/blob/main/crypto/evp_extra/p_kem.c#L143-L148

crypto/kem/kem.c
```diff
@@ -25,39 +26,39 @@ static const uint8_t kOIDKyber1024r3[] = {0xff, 0xff, 0xff, 0xff};
...
-    1632,                    // kem.secret_key_len
-    768,                     // kem.ciphertext_len
-    32,                      // kem.shared_secret_len
-    &kem_kyber512r3_method,  // kem.method
+    NID_KYBER512_R3,                // kem.nid
+    kOIDKyber512r3,                 // kem.oid
+    sizeof(kOIDKyber512r3),         // kem.oid_len
+    "Kyber512 Round-3",             // kem.comment
+    KYBER512_R3_PUBLIC_KEY_BYTES,   // kem.public_key_len
+    KYBER512_R3_PRIVATE_KEY_BYTES,  // kem.secret_key_len
```
dkostic, line 34: Nit: we are calling it `secret_key` not `private_key`.

ssl/ssl_key_share.cc
```diff
@@ -176,20 +180,557 @@ class X25519KeyShare : public SSLKeyShare {
...
+    // Retrieve the lengths of the ciphertext and shared secret
+    size_t ciphertext_len = 0;
+    size_t secret_len = 0;
+    if (!EVP_PKEY_encapsulate(ctx_.get(), nullptr, &ciphertext_len,
+                              nullptr, &secret_len)) {
+      OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
+      return false;
+    }
+
+    // EVP_PKEY_encapsulate() will generate a shared secret, then encrypt it
```
dkostic, line 312:
```suggestion
    // EVP_PKEY_encapsulate() will generate a shared secret, then encapsulate it
```

ssl/ssl_key_share.cc
```diff
@@ -176,20 +180,557 @@ class X25519KeyShare : public SSLKeyShare {
...
+        OPENSSL_PUT_ERROR(SSL, ERR_R_PASSED_NULL_PARAMETER);
+        return false;
+      }
+
+      if (!hybrid_group_ || this->exchange_performed) {
+        OPENSSL_PUT_ERROR(SSL, ERR_R_INTERNAL_ERROR);
+        return false;
+      }
+
+      // A hybrid shared secret with two components should be 64 bytes.
```
dkostic, line 494: where does 64 come from? is it defined in some spec?

alexw91: It is the initial capacity of the Crypto Byte Buffer that will be storing the hybrid secret in. It will grow as needed as data is added.

ssl/ssl_key_share.cc
```diff
@@ -176,20 +180,557 @@ class X25519KeyShare : public SSLKeyShare {
...
+          *out_alert = SSL_AD_DECODE_ERROR;
+          OPENSSL_PUT_ERROR(SSL, SSL_R_BAD_HYBRID_KEYSHARE);
+          return false;
+        }
+
+        Span<const uint8_t> component_key =
+          peer_key.subspan(peer_key_index, component_key_size);
+
+        Array<uint8_t> component_secret;
+        if (!key_shares_[i] ||
```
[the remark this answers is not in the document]
alexw91: Calling `this->` is optional when accessing a class member variable when there are no conflicting variables with the same name. Either of the following works: `this->key_shares_` or `key_shares_`.

ssl/ssl_key_share.cc
```diff
@@ -176,20 +180,557 @@ class X25519KeyShare : public SSLKeyShare {
...
+
+      // Accept() each component key share. Each component's Accept() function
+      // will generate a shared secret and a public key to be sent back to
+      // the peer. The hybrid public key is the concatenation of all component
+      // public keys; the hybrid shared secret is the concatenation of all
+      // component shared secrets.
+      size_t peer_key_index = 0;
+      for (size_t i = 0; i < NUM_HYBRID_COMPONENTS; i++) {
+        size_t component_key_size = 0;
+        if (!get_component_offer_key_share_size(&component_key_size, hybrid_group_->component_group_ids[i])) {
```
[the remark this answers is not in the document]
alexw91: What logic should be moved to the constructor? Do you mean that we should use an array lookup instead of a getter function, and add and initialize two new private class variables `uint16_t offer_share_size[2]` and `uint16_t accept_share_size[2]` to the HybridKeyShare class?

ssl/ssl_key_share.cc
alexw91: Fixed

crypto/kem/kem.c
alexw91: Fixed

ssl/ssl_key_share.cc
alexw91: Made this fix. However, since the size checks now happen inside the decapsulate function instead of before at the SSL layer, the client will now receive `SSL_AD_INTERNAL_ERROR` TLS Alert instead of `SSL_AD_DECODE_ERROR` if they send an improperly sized KEM ciphertext.

ssl/ssl_key_share.cc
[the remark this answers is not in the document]
dkostic: ha, didn't know that, cool.

ssl/ssl_key_share.cc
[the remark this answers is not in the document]
dkostic: yes, I meant something like that. Or maybe even have the base `SSLKeyShare` class have fields for the share sizes? Seems to me like a normal information to have in this class, but I'm not very familiar with this code so I might be wrong :)

ssl/ssl_key_share.cc
```diff
@@ -176,20 +180,548 @@ class X25519KeyShare : public SSLKeyShare {
...
+      return false;
+    }
+
+    // ctx_ should have been initialized by a previous call to Offer()
+    if (!ctx_) {
+      OPENSSL_PUT_ERROR(SSL, ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
+      return false;
+    }
+
+    // Retrieve the lengths of the ciphertext and shared secret
```
dkostic, line 369:
```suggestion
    // Retrieve the lengths of the shared secret
```

alexw91: Fixed

dkostic (approved):

andrewhop (approved):

PR #1204, opened 2023-09-21
Handle x30 target register in delocator for aarch64 to avoid clobbering the register

torben-hansen:
### Description of changes: 

When taking the commit: https://github.com/justsmth/aws-lc/commit/3800f0fcb53c3fc2bb2ffa5bba57de0a67774a5f gcc 11.4 emitted the following in the routine `ec_GFp_nistp224_felem_sqr()`:
```
adrp    x30, :got:__stack_chk_guard
```
This uncovered an unsupported case in the delocater. 

When seeing the line above, the delocater resolved the relocation like this:
```
// WAS adrp    x30, :got:__stack_chk_guard
    sub sp, sp, 128
    stp x0, lr, [sp, #-16]!
    bl .Lboringssl_loadgot___stack_chk_guard
    mov x30, x0
    ldp x0, lr, [sp], #16
    add sp, sp, 128
```
But `x30` is an alias for the aarch64 link register `lr`.  Hence, `mov x30, x0` would clobber `x30` that is used as a general-purpose register in the un-modified routine; the routine pushes the state onto the stack and pops before returning making x30 available for scratch purposes. This caused a false-positive "stack smashing" event in the stack guard detector. The clobbered register `x30` contained the stack guard value that is verified before returning from the routine. When the register got clobbered, the stack guard value got corrupted causing the validation to fail with very high probability.

Fix this by handling `x30` as a special case in the delocater.  If `x30` is a target register, we assume that the routine use `x30` as a (volatile) general-purpose register and will recover the link register value, using an appropriate method, before returning. This is really similar to how the `x0` case is handled since x0 holds the return value from the subroutine and is not to be overwritten by the value saved on the stack..

### Testing:

Added a test case in the delocater tests:
```
$ pwd
/[...]/aws-lc/util/fipstools/delocate

$ go test
PASS
ok      boringssl.googlesource.com/boringssl/util/fipstools/delocate    0.010s
```

### DIFF
--- a/crypto/fipsmodule/ec/p224-64.c
+++ b/crypto/fipsmodule/ec/p224-64.c
@@ -178,10 +178,6 @@ static const p224_felem g_p224_pre_comp[2][16][3] = {
 
 // Helper functions to convert field elements to/from internal representation
 
-#if defined(AWSLC_FIPS)
-// Related to delocate issue with gcc-11
-__attribute__((noinline))
-#endif
 static void p224_generic_to_felem(p224_felem out, const EC_FELEM *in) {
   // |p224_felem|'s minimal representation uses four 56-bit words. |EC_FELEM|
   // uses four 64-bit words. (The top-most word only has 32 bits.)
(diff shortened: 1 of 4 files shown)


torben-hansen:
`aws-lc-rs sanity tests` dimension is expected to fail atm.

justsmth (approved):

nebeid:
The PR description is great. I suggest you add "This is really similar to how the x0 case is handled since x0 holds the return value from the subroutine and is not to be overwritten by the value saved on the stack."

nebeid (approved):

PR #1205, opened 2023-09-21
Add BIGNUM support for big-endian architectures

justsmth:
### Issues:
N/A

### Description of changes: 
* Updates work done on previous PR: https://github.com/aws/aws-lc/pull/1133
* Adds `EndianTest` to test our support for little- and big-endian architectures.
* Adds macros for `OPENSSL_PPC64BE` and `OPENSSL_PPC32BE`.

### Call-outs:
* The FIPS build for ppc64be is failing - delocator.go cannot parse the assembly code.
* There's a lot more work to fully support big-endian

### Testing:
For ppc64be:
* All `EndianTest.*` tests pass except `EndianTest.AES`.
```
...
[----------] 16 tests from EndianTest (18 ms total)

[==========] 16 tests from 1 test suite ran. (23 ms total)
[  PASSED  ] 15 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] EndianTest.AES

 1 FAILED TEST
```
* All 'BNTest.*` tests pass.
```
...
[----------] 36 tests from BNTest (935247 ms total)

[==========] 36 tests from 1 test suite ran. (935252 ms total)
[  PASSED  ] 36 tests

```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -701,7 +701,7 @@ endif()
 
 TEST_BIG_ENDIAN(BIG_ENDIAN)
 if(BIG_ENDIAN)
-  message(FATAL_ERROR "Big Endian is not supported.")
+  message(STATUS "Continuing with experimental support on big endian platform")
 endif()
 
 if(OPENSSL_NO_SSE2_FOR_TESTING)
@@ -741,7 +741,7 @@ elseif (${CMAKE_SYSTEM_PROCESSOR} STREQUAL "riscv64")
 elseif (${CMAKE_SYSTEM_PROCESSOR} STREQUAL "loongarch64")
   set(ARCH "loongarch64")
 else()
-  message(STATUS "Unknown processor found. Using generic implementations. Processor:" ${CMAKE_SYSTEM_PROCESSOR})
+  message(STATUS "Unknown processor found. Using generic implementations. Processor: " ${CMAKE_SYSTEM_PROCESSOR})
   set(ARCH "generic")
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -701,9 +701,11 @@ endif()
 
 TEST_BIG_ENDIAN(BIG_ENDIAN)
 if(BIG_ENDIAN)
-  message(FATAL_ERROR "Big Endian is not supported.")
-# TODO: Enable BIG_ENDIAN once we have better support
-#  message(STATUS "Continuing with experimental support on big endian platform")
+  if(ENABLE_EXPERIMENTAL_BIG_ENDIAN_SUPPORT)
+    message(STATUS "Continuing with experimental support on big endian platform")
+  else()
+    message(FATAL_ERROR "Big Endian is not supported.")
+  endif()
 endif()
 
 if(OPENSSL_NO_SSE2_FOR_TESTING)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -701,7 +701,9 @@ endif()
 
 TEST_BIG_ENDIAN(BIG_ENDIAN)
 if(BIG_ENDIAN)
-  message(STATUS "Continuing with experimental support on big endian platform")
+  message(FATAL_ERROR "Big Endian is not supported.")
+# TODO: Enable BIG_ENDIAN once we have better support
+#  message(STATUS "Continuing with experimental support on big endian platform")
 endif()
 
 if(OPENSSL_NO_SSE2_FOR_TESTING)
--- a/crypto/bytestring/cbs.c
+++ b/crypto/bytestring/cbs.c
@@ -521,8 +521,18 @@ int CBS_get_asn1_int64(CBS *cbs, int64_t *out) {
   uint8_t sign_extend[sizeof(int64_t)];
   memset(sign_extend, is_negative ? 0xff : 0, sizeof(sign_extend));
   for (size_t i = 0; i < len; i++) {
+#ifdef OPENSSL_BIG_ENDIAN
+    // sign_extend[sizeof(int64_t) - 0 - 1] = data[0];
+    // sign_extend[8 - 0 - 1] = data[0];
+    // sign_extend[7] = data[0];
+    sign_extend[sizeof(sign_extend) - i - 1] = data[len - i - 1];
+#else
+    // sign_extend[0] = data[1 - 0 - 1];
+    // sign_extend[0] = data[0];
     sign_extend[i] = data[len - i - 1];
+#endif
   }
+
   memcpy(out, sign_extend, sizeof(sign_extend));
   return 1;
 }
--- a/crypto/bytestring/cbs.c
+++ b/crypto/bytestring/cbs.c
@@ -521,14 +521,13 @@ int CBS_get_asn1_int64(CBS *cbs, int64_t *out) {
   uint8_t sign_extend[sizeof(int64_t)];
   memset(sign_extend, is_negative ? 0xff : 0, sizeof(sign_extend));
   for (size_t i = 0; i < len; i++) {
+// `data` is big-endian.
+// Values are always shifted toward the "little" end.
 #ifdef OPENSSL_BIG_ENDIAN
-    // sign_extend[sizeof(int64_t) - 0 - 1] = data[0];
-    // sign_extend[8 - 0 - 1] = data[0];
-    // sign_extend[7] = data[0];
+    // Bytes are written starting at the highest index.
     sign_extend[sizeof(sign_extend) - i - 1] = data[len - i - 1];
 #else
-    // sign_extend[0] = data[1 - 0 - 1];
-    // sign_extend[0] = data[0];
+    // Bytes are written starting at the lowest index.
     sign_extend[i] = data[len - i - 1];
 #endif
   }
--- a/crypto/endian_test.cc
+++ b/crypto/endian_test.cc
@@ -53,8 +53,6 @@ uint8_t expected_le[4] = {0x78, 0x56, 0x34, 0x12};
 
 TEST(EndianTest, TestRotate32) {
   uint32_t value = 0b00000010000000000000000000000;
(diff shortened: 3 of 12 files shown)


crypto/fipsmodule/bn/bytes.c
```diff
@@ -143,41 +143,72 @@ BIGNUM *BN_le2bn(const uint8_t *in, size_t len, BIGNUM *ret) {
...
   ret->d[num_words - 1] = 0;
 
-  // We only support little-endian platforms, so we can simply memcpy the
-  // internal representation.
+#ifdef OPENSSL_BIG_ENDIAN
+  BN_ULONG word = 0;
+  unsigned m;
+
+  m = (len - 1) % BN_BYTES;
+  for (size_t i = len - 1; i < len; i--) {
```
WillChilds-Klein, line 151: if we're initializing `i = len-1`, and the loop condition is `i < len` as we decrement, then the loop will only terminate when `i` underflows, right? would making the loop condition `i >= 0` be more clear?

crypto/endian_test.cc
```diff
@@ -0,0 +1,250 @@
...
+  bssl::UniquePtr<BIGNUM> x(BN_new());
+  uint8_t input[256];
+  OPENSSL_memset(input, 0, sizeof(input));
+  input[0] = 0xaa;
+  input[1] = 0x01;
+  input[254] = 0x01;
+  input[255] = 0x02;
+  ASSERT_NE(nullptr, BN_bin2bn(input, sizeof(input), x.get()));
+  EXPECT_FALSE(BN_is_zero(x.get()));
+  for (int i = 1; i < (256*8/BN_BITS2) - 1; i++) {
```
WillChilds-Klein, line 136: minor nit: `sizeof(input)` instead of `256` here and L141

crypto/endian_test.cc
```diff
@@ -0,0 +1,250 @@
...
+  OPENSSL_memset(out, 0, sizeof(out));
+  EXPECT_EQ(1, BN_bn2bin_padded(out, sizeof(out), x.get()));
+  EXPECT_EQ(Bytes(input), Bytes(out));
+}
+
+TEST(EndianTest, AES) {
+  // Initialize the key and message buffers with zeros
+  uint8_t key[AES_BLOCK_SIZE] = {0};
+  key[0] = 0xaa;
+  key[0] = 0xbb;
```
WillChilds-Klein, line 214: should this be `key[1]`, or is the desired array `[0xbb, 0x00, 0x00, 0x00, ...]`?

crypto/fipsmodule/bn/bytes.c
andrewhop: size_t is unsigned so i >= 0 is always true.

WillChilds-Klein: Right, that makes sense. And we need to cover `i = 0`, so i guess the underflow is necessary. Perhaps we can add a small comment stating the intention?

andrewhop (commented):
Why didn't you have to update bn_big_endian_to_words?

justsmth:
> Why didn't you have to update bn_big_endian_to_words?

The call to [`CRYPTO_load_word_be`](https://github.com/aws/aws-lc/blob/7ce723e94f36092d0f4a30e4a4e6504a10a6c27b/crypto/internal.h#L907) on [this line](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/bn/bytes.c#L81C14-L81C33) handles differences due to endian-ness when loading whole "words". The rest of the logic is correct for either big- or little-endian.

crypto/bytestring/cbs.c
```diff
@@ -521,8 +521,18 @@ int CBS_get_asn1_int64(CBS *cbs, int64_t *out) {
   uint8_t sign_extend[sizeof(int64_t)];
   memset(sign_extend, is_negative ? 0xff : 0, sizeof(sign_extend));
   for (size_t i = 0; i < len; i++) {
+#ifdef OPENSSL_BIG_ENDIAN
+    // sign_extend[sizeof(int64_t) - 0 - 1] = data[0];
+    // sign_extend[8 - 0 - 1] = data[0];
+    // sign_extend[7] = data[0];
```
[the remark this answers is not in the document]
justsmth: I'll remove these notes and add some comments for clarity:
```C
// `data` is big-endian.
// Values are always shifted toward the "little" end.
#ifdef OPENSSL_BIG_ENDIAN
    // Bytes written starting at the highest index.
    sign_extend[sizeof(sign_extend) - i - 1] = data[len - i - 1];
#else
    // Bytes written starting at the lowest index.
    sign_extend[i] = data[len - i - 1];
#endif
```

crypto/fipsmodule/bn/bytes.c
justsmth: Adding comment:
```c
  // size_t is unsigned so i >= 0 is always true 
  for (size_t i = len - 1; i < len; i--) {
```

crypto/internal.h
```diff
@@ -855,68 +864,114 @@ static inline void *OPENSSL_memset(void *dst, int c, size_t n) {
 static inline uint32_t CRYPTO_load_u32_le(const void *in) {
   uint32_t v;
   OPENSSL_memcpy(&v, in, sizeof(v));
+#if defined(OPENSSL_BIG_ENDIAN)
```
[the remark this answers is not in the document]
justsmth: I'm OK with having it this way -- each function definition only appears once (helping to minimize redundant lines). There's also no chance this way of being confused by reading the "wrong" function definition (i.e., imagine that the BE block is very large and not realizing you're looking at the big-endian-specific definition of a function and thinking it applies to little-endian).

CMakeLists.txt
```diff
@@ -702,6 +702,8 @@ endif()
 TEST_BIG_ENDIAN(BIG_ENDIAN)
 if(BIG_ENDIAN)
   message(FATAL_ERROR "Big Endian is not supported.")
+# TODO: Enable BIG_ENDIAN once we have better support
+#  message(STATUS "Continuing with experimental support on big endian platform")
```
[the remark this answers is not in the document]
justsmth: Good idea, I'll add a flag:
```cmake
if(BIG_ENDIAN)
  if(ENABLE_EXPERIMENTAL_BIG_ENDIAN_SUPPORT)
    message(STATUS "Continuing with experimental support on big endian platform")
  else()
    message(FATAL_ERROR "Big Endian is not supported.")
  endif()
endif()
```

crypto/endian_test.cc
justsmth: I agree.
```c
for (size_t i = 1; i < (sizeof(input)*8/BN_BITS2) - 1; i++) {
```

crypto/endian_test.cc
justsmth: Fixed:
```c
  key[0] = 0xaa;
  key[1] = 0xbb;
...
  const uint8_t known_value_bytes[AES_BLOCK_SIZE] = {
      0xf6, 0x8a, 0x88, 0x76, 0x05, 0x23, 0xd4, 0x14, 0x7e, 0x5a, 0xb8, 0x52, 0x8f, 0x56, 0x62, 0x3e
  };
```

crypto/fipsmodule/bcm.c
```diff
@@ -228,7 +228,8 @@ static void BORINGSSL_bcm_power_on_self_test(void) __attribute__ ((constructor))
 #endif
 
 static void BORINGSSL_bcm_power_on_self_test(void) {
-#if !defined(OPENSSL_NO_ASM)
+#if !defined(OPENSSL_NO_ASM) && !defined(OPENSSL_PPC32BE) && !defined(OPENSSL_PPC64BE)
+// TODO: Do we need to implement this for PPC BE architectures?
```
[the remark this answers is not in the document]
nebeid: In my AES experiments, I left it as
```
#if !defined(OPENSSL_NO_ASM) && !defined(OPENSSL_PPC32BE)
```
Because testing on an emulated POWER8 (ISA 2.07) Big-endian 64-bit architecture that has VCRYPTO capability, it could use the assembly of [crypto/fipsmodule/aes/asm/aesp8-ppc.pl](https://github.com/aws/aws-lc/blob/9db98588b97078acdcda13d9f4b5b7be86e5b602/crypto/fipsmodule/aes/asm/aesp8-ppc.pl) and [crypto/fipsmodule/modes/asm/ghashp8-ppc.pl](https://github.com/aws/aws-lc/blob/9db98588b97078acdcda13d9f4b5b7be86e5b602/crypto/fipsmodule/modes/asm/ghashp8-ppc.pl)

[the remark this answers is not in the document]
justsmth: Per conversation with Nevine, I'll add a comment above this line:
```
// TODO: remove !defined(OPENSSL_PPC64BE) from the check below when starting to support
```

crypto/fipsmodule/bcm.c
```diff
@@ -228,8 +228,8 @@ static void BORINGSSL_bcm_power_on_self_test(void) __attribute__ ((constructor))
 #endif
 
 static void BORINGSSL_bcm_power_on_self_test(void) {
+// TODO: remove !defined(OPENSSL_PPC64BE) from the check below when starting to support
```
nebeid, line 231:
```
// TODO: remove !defined(OPENSSL_PPC64BE) from the check below when starting to support
// PPC64BE that has VCRYPTO capability. In that case, add `|| defined(OPENSSL_PPC64BE)`
// to `#if defined(OPENSSL_PPC64LE)` wherever it occurs.
```

WillChilds-Klein (approved):

crypto/internal.h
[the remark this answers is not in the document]
nebeid: I agree with justsmth because when I look at a single function, I can confirm my understanding of what it does when I see both branches.

andrewhop (approved):

PR #1213, opened 2023-09-28
Fix AES on PPCBE (32-bit and 64-bit).

nebeid:
### Issues:
Addresses CryptoAlg-1954

### Description of changes: 
Fix AES C code on PPC 64 BE (that doesn't support the HW capability AT_HWCAP2 VCRYPTO; otherwise the `p8` assembly implementations would have been used):
Only compact_block and uncompact_block functions handled the input and output and were the only ones needing change.

### Call-outs:
In `aes_nohw_compact_block`, I opted out to reduce the work done in the case of the little-endian build by
using the `load_u64` only in the case of big-endian build.

### Testing:
By cross-compiling for ppc64 and ppc using qemu-user tools, and running crypto_test filtering on AES, XTS, GCM, ECB, CBC, CTR, CCM, AEAD.


crypto/fipsmodule/aes/aes_nohw.c
```diff
@@ -381,8 +385,13 @@ static inline void aes_nohw_uncompact_block(
       aes_nohw_uncompact_word((a0 & UINT64_C(0x00000000ffffffff)) | (a1 << 32));
   uint64_t b1 =
       aes_nohw_uncompact_word((a1 & UINT64_C(0xffffffff00000000)) | (a0 >> 32));
+#if defined(OPENSSL_BIG_ENDIAN)
+  CRYPTO_store_u64_le(&out[0], b0);
+  CRYPTO_store_u64_le(&out[8], b1);
+#else
   memcpy(out, &b0, 8);
   memcpy(out + 8, &b1, 8);
+#endif // !OPENSSL_BIG_ENDIAN
```
andrewhop, line 394: [CRYPTO_store_u64_le](https://github.com/aws/aws-lc/blob/main/crypto/internal.h#L911-L917) now handles both BE and LE correctly. 
```suggestion
  CRYPTO_store_u64_le(&out[0], b0);
  CRYPTO_store_u64_le(&out[8], b1);
```

justsmth:
This works for ppc64, but it's not yet working for ppc32.

nebeid:
Removed commented out code was there in order to create the correct order of 32-bit key rounds in AES_KEY, but it was deemed unnecessary. The order of each 2 subsequent 32-bit round keys is flipped, but because they get read again into 64-bit words where they were ordered correctly, the calculations proceeded successfully.

justsmth (approved):

crypto/fipsmodule/modes/xts.c
```diff
@@ -62,7 +62,6 @@ size_t CRYPTO_xts128_encrypt(const XTS128_CONTEXT *ctx,
                              uint8_t *out, size_t len, int enc) {
   union {
     uint64_t u[2];
-    uint32_t d[4];
     uint8_t c[16];
```
andrewhop, line 65: Any idea why there isn't an issue mixing the uint64 and uint8? I get the issue with  `d` that you got rid of below by using `u` and a different shift.

nebeid: There is an issue with mixing them that is handled by loading the 64-bit values into separate variables, `tweak_u0` and `tweak_u1`, which was not needed in the little-endian case and the shifts were done directly on the union values.

andrewhop (approved):

justsmth (approved):

Issue #1185, opened 2023-09-12
Uninitialised variable in `ec_GFp_mont_batch_get_window`


giordano:
### Problem:

When building v1.15.0 targeting 32-bit ARM architectures like armv6l or armv7l [I get](https://buildkite.com/julialang/yggdrasil/builds/5076#018a86c1-81e2-4c38-9d64-17db3b001a83/656-1217):
```
[00:20:47] In file included from /workspace/srcdir/aws-lc/crypto/fipsmodule/bcm.c:96:0:
[00:20:47] /workspace/srcdir/aws-lc/crypto/fipsmodule/ec/simple_mul.c: In function ‘ec_GFp_mont_batch_get_window’:
[00:20:47] /workspace/srcdir/aws-lc/crypto/fipsmodule/ec/felem.c:65:19: error: ‘neg_Y.words[0]’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
[00:20:47]      out->words[i] &= mask;
[00:20:47]                    ^~
[00:20:47] cc1: all warnings being treated as errors
```

#### Relevant details
AWS-LC commit: (aa90e509f2e940916fbe9fdd469a4c90c51824f6)

System information: for linux, below info can be collected by running `uname -srvmp`
 * target CPU architecture: armv6l or armv7l
 * CPU name: N/A
 * OS: Linux

[Build log](https://buildkite.com/julialang/yggdrasil/builds/5076#018a86c1-81e2-4c38-9d64-17db3b001a83/656-808): 

```text
[00:20:24]  ---> cmake -DCMAKE_INSTALL_PREFIX=${prefix} -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TARGET_TOOLCHAIN} -DBUILD_TESTING=OFF -DCMAKE_BUILD_TYPE=Release -DDISABLE_GO=ON -DBUILD_SHARED_LIBS=OFF -GNinja ..
[00:20:24] -- The C compiler identification is GNU 7.1.0
[00:20:24] -- Detecting C compiler ABI info
[00:20:25] -- Detecting C compiler ABI info - done
[00:20:25] -- Check for working C compiler: /opt/bin/armv6l-linux-gnueabihf-libgfortran4-cxx11/arm-linux-gnueabihf-gcc - skipped
[00:20:25] -- Detecting C compile features
[00:20:25] -- Detecting C compile features - done
[00:20:25] -- The CXX compiler identification is GNU 7.1.0
[...]

PR #1215, opened 2023-09-29
Flag so compiler warnings do not fail build

justsmth:
### Issues:
Addresses #1185

### Description of changes: 
* Accept flag to allow warnings to not fail the build.

### Call-outs:
N/A

### Testing:
* I set a commonly used function as deprecated, then verified that build would succeed with flag.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -308,7 +308,9 @@ if (GCC)
   endif ()
 endif ()
 
+# Private build configuration. Subject to deprecation at any time.
 if(AWS_LC_WARNINGS_ARE_NOT_ERRORS)
+  message(WARNING "Compiler warning-diagnostics will not fail the build")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -308,6 +308,12 @@ if (GCC)
   endif ()
 endif ()
 
+if(AWS_LC_WARNINGS_ARE_NOT_ERRORS)
```
torben-hansen, line 311:
```suggestion
# Private build configuration. Subject to deprecation at any time.
if(AWS_LC_WARNINGS_ARE_NOT_ERRORS)
  message(WARNING "Compiler warning-diagnostics will not fail the build")
```
Suggestion to allow deprecation.

dkostic:
We have discussed this change internally as a team and decided that we won't add the flag at this time.

PR #1219, opened 2023-10-01
Replace s2n-bignum's Montgomery reduction with its latest, faster version

aqjune-aws:
### Description of changes: 

This is going to be the last patch for improving RSA signatures on Graviton 2.
This patch replaces s2n-bignum's bignum_emontredc_8n_neon with its latest, faster version in s2n-bignum.
(At the moment, the code is not verified yet; once verification is done, I will deliver the verified code to s2n-bignum first, and switch this PR from draft to ready.)
The faster version uses two techniques: (1) Cache differences used in Arbitrary Degree Karatsuba and store it in a temporary buffer to reuse it in its innermost loop (2) Schedule Arm instructions in the main loop using [SLOTHY](https://github.com/slothy-optimizer/slothy) (a superoptimization tool for Arm).

```
| RSA signing bits | baseline (b405bb9) | new  | speedup |
|------------------|--------------------|------|---------|
| 2048             | 540                | 583  | 8.0%    |
| 3072             | 128.6              | 139  | 8.1%    |
| 4096             | 81                 | 91.4 | 12.8%   |
| 8192             | 7.6                | 8.4  | 10.5%   |
```

To store the differences, a temporary buffer of `uint64_t[12 * (BN_MONTGOMERY_MAX_WORDS / 4 - 1)]` is created at stack.
Since `BN_MONTGOMERY_MAX_WORDS` is set to `8192 / 8 = 1024`, this will use 255 * 12 * `size_of(uint64_t)` ~= 24 KB.
A question is whether allocating this size on stack is acceptable. If this is too big, I can make this PR use this newer montgomery reduction only when the montgomery multiplication's bitwidth is sufficiently small (which will still cover the major use cases).

### Testing:
`bssl speed -filter RSA`

### DIFF
--- a/crypto/fipsmodule/bn/montgomery.c
+++ b/crypto/fipsmodule/bn/montgomery.c
@@ -518,8 +518,13 @@ static void montgomery_s2n_bignum_mul_mont(BN_ULONG *rp, const BN_ULONG *ap,
   //    A. The result of step 1 >= 2^(64*num), meaning that bignum_emontredc_8n
   //       returned 1. Since m is less than 2^(64*num), (result of step 1) >= m holds.
   //    B. The result of step 1 fits in 2^(64*num), and the result >= m.
+
+  // temp's size is 12 * (num / 4 - 1) where num's upperbound is
+  // BN_MONTGOMERY_MAX_WORDS (caller's assertion), and num is divisible by 8
+  // by montgomery_use_s2n_bignum(num).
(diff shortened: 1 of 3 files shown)


aqjune-aws:
Oops, the undering assembly code is still under verification.
To avoid confusion, I will close this pull request and open a new one when everything is ready. :)

PR #1233, opened 2023-10-09
self destruct ec2 instances after certain amount of time

samuel40791765:
### Description of changes: 
The new ec2 test framework doesn't mesh well with our job pruner. Codebuild just "stops" the instance without giving a signal to the executing script, so `trap cleanup` is not executed. This lets the ec2 self-terminate once a certain amount of time has passed.

### DIFF
--- a/tests/ci/cdk/cdk/ssm/general_test_run_ssm_document.yaml
+++ b/tests/ci/cdk/cdk/ssm/general_test_run_ssm_document.yaml
@@ -10,6 +10,9 @@ mainSteps:
     inputs:
       timeoutSeconds: '7200'
       runCommand:
+        # Fallback plan to shut down the ec2 instance in 60 minutes in case it's not terminated.
+        # Codebuild just "stops" the instance calling the script, so "trap cleanup" is not executed.
+        - shutdown -P +60
         - sudo -i
         - export DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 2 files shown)


tests/ci/cdk/cdk/ssm/general_test_run_ssm_document.yaml
```diff
@@ -10,6 +10,9 @@ mainSteps:
     inputs:
       timeoutSeconds: '7200'
       runCommand:
+        # Fallback plan to shut down the ec2 instance in 60 minutes in case it's not terminated.
+        # Codebuild just "stops" the instance calling the script, so "trap cleanup" is not executed.
+        - shutdown -P +60
```
billbo-yang, line 15: Is there any chance that this will cause a shutdown while we're still running tests? Also, unless I'm misunderstanding something - if Codebuild stops the instance calling the script when pruning, then isn't the added line in `run_ec2_test_framework.sh` enough?

samuel40791765: 1. We try to let all our CI jobs finish within 60 minutes, so this should be enough for the tests to finish running. 
2. Thought the same, but apparently Codebuild just kills the instance when STOP is called and the script doesn't get the chance to execute `cleanup`.

billbo-yang (approved):

skmcgrail (approved):

PR #1234, opened 2023-10-09
Add support for RSA KeyGen AFT tests for FIPS186-5 to ACVP tool

billbo-yang:
### Issues:
Addresses CryptoAlg-1838

### Description of changes: 
[FIPS 186-5](https://csrc.nist.gov/pubs/fips/186-5/final) updates some things related to RSA. For ACVP, it introduces a new capabilities registration. Luckily, the format of the vectors is largely the same, but we need to enable support for "AFT" tests in order to correctly parse FIPS 186-5 compatible test vectors.

### Call-outs:
N/A

### Testing:
Tool passes when running against ACVP demo server.

### DIFF
--- a/util/fipstools/acvp/acvptool/subprocess/rsa.go
+++ b/util/fipstools/acvp/acvptool/subprocess/rsa.go
@@ -126,10 +126,9 @@ func processKeyGen(vectorSet []byte, m Transactable) (any, error) {
 	var ret []rsaKeyGenTestGroupResponse
 
 	for _, group := range parsed.Groups {
-		// GDT means "Generated data test", i.e. "please generate an RSA key".
-		const expectedType = "GDT"
-		if group.Type != expectedType {
-			return nil, fmt.Errorf("RSA KeyGen test group has type %q, but only generation tests (%q) are supported", group.Type, expectedType)
+		// We support both GDT and AFT tests, which are formatted the same and expect the same output.
+		if !(group.Type == "GDT" || group.Type == "AFT") {
+			return nil, fmt.Errorf("RSA KeyGen test group has type %q, but only GDT and AFT tests are supported", group.Type)
 		}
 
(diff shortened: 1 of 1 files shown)


util/fipstools/acvp/acvptool/subprocess/rsa.go
```diff
@@ -126,10 +126,9 @@ func processKeyGen(vectorSet []byte, m Transactable) (any, error) {
 	var ret []rsaKeyGenTestGroupResponse
 
 	for _, group := range parsed.Groups {
-		// GDT means "Generated data test", i.e. "please generate an RSA key".
-		const expectedType = "GDT"
-		if group.Type != expectedType {
-			return nil, fmt.Errorf("RSA KeyGen test group has type %q, but only generation tests (%q) are supported", group.Type, expectedType)
+		// We support both GDT and AFT tests, which are formatted the same and expect the same output.
```
samuel40791765, line 129: Curious what the exact difference between GDT and AFT is?

billbo-yang: In this case, the GDT and AFT tests are exactly the same. For whatever reason, NIST renamed the test when they issued FIPS 186-5.

Most of the changes are related to how we register our capabilities with the ACVP server now. If we choose to register with the new way (under FIPS 186-5), we get vectors that are labeled as "AFT" tests. If we choose to register with the old way (under FIPS 186-4), we get vectors that are labeled as "GDT".

Other than the labels though, it appears that both the vectors that we get from NIST for RSA as well as the expected format/contents of the corresponding response vectors are the exact same as before...

You can see more details on the ACVP specifications page for RSA: https://pages.nist.gov/ACVP/draft-celi-acvp-rsa.html

samuel40791765 (approved):

andrewhop (approved):

PR #1240, opened 2023-10-10
Fix 32-bit big-endian p521 bug

justsmth:
### Issues:
N/A

### Description of changes: 
* Fixes 32-bit big-endian bug in P521.
  * `P521_NLIMBS` is 19, but the `EC_FELEM` type only contains `EC_MAX_WORDS` (= 17) words.

### Call-outs:
N/A

### Testing:
```
❮ ppc-qemu.sh ./crypto/crypto_test --gtest_filter="EC*.*:*/EC*.*"                         
Note: Google Test filter = EC*.*:*/EC*.*
[==========] Running 156 tests from 6 test suites.
[----------] 7 tests from ECDHTest
...
[----------] 80 tests from All/ECCurveTest (7543 ms total)

[==========] 156 tests from 6 test suites ran. (407363 ms total)
[  PASSED  ] 156 tests.
```

### DIFF
--- a/crypto/fipsmodule/ec/internal.h
+++ b/crypto/fipsmodule/ec/internal.h
@@ -109,7 +109,7 @@ OPENSSL_STATIC_ASSERT(EC_MAX_WORDS <= BN_SMALL_MAX_WORDS,
 // An EC_SCALAR is an integer fully reduced modulo the order. Only the first
 // |order->width| words are used. An |EC_SCALAR| is specific to an |EC_GROUP|
 // and must not be mixed between groups.
-typedef union {
+typedef struct {
   // words is the representation of the scalar in little-endian order.
   BN_ULONG words[EC_MAX_WORDS];
 } EC_SCALAR;
--- a/crypto/fipsmodule/ec/internal.h
+++ b/crypto/fipsmodule/ec/internal.h
@@ -196,6 +196,12 @@ void ec_scalar_select(const EC_GROUP *group, EC_SCALAR *out, BN_ULONG mask,
 
 // Field elements.
 
+#define P384_EC_FELEM_BYTES (48)
+#define P384_EC_FELEM_WORDS ((P384_EC_FELEM_BYTES + BN_BYTES - 1) / BN_BYTES)
+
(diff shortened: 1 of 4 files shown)


crypto/fipsmodule/ec/internal.h
```diff
@@ -109,7 +109,7 @@ OPENSSL_STATIC_ASSERT(EC_MAX_WORDS <= BN_SMALL_MAX_WORDS,
 // An EC_SCALAR is an integer fully reduced modulo the order. Only the first
 // |order->width| words are used. An |EC_SCALAR| is specific to an |EC_GROUP|
 // and must not be mixed between groups.
-typedef union {
+typedef struct {
```
justsmth, line 112: This should have been done as part of the upstream merge of #1177.

crypto/fipsmodule/ec/p384.c
```diff
@@ -144,6 +144,7 @@ static p384_limb_t p384_felem_nz(const p384_limb_t in1[P384_NLIMBS]) {
 #endif // P384_USE_S2N_BIGNUM_FIELD_ARITH
 
 #define P384_FELEM_BYTES (48)
+#define P384_FELEM_WORDS ((P384_FELEM_BYTES + BN_BYTES - 1) / BN_BYTES)
```
dkostic, line 147: in the low level EC implementations (for all curves) we are making an assumption on the real size of `EC_FELEM` for a specific curve. For example, for P-384 we assume that `EC_FELEM` associated with an `EC_GROUP` for P-384 is 48 bytes long (line 146). I have two suggestions:
- rename `P384_FELEM_BYTES/WORDS` to `P384_EC_FELEM_BYTES/WORDS` (same for other curves).
- remove the asserts you added in the latest change and add a unit test instead that checks `P384_EC_FELEM_BYTES/WORDS` are equal to the real size of a `EC_FELEM` object instantiated with the corresponding `EC_GROUP`.
what do you think?

dkostic (approved):

skmcgrail (approved):

PR #1245, opened 2023-10-12
Slight build fix and migrate to GHA for MacOS ARM CI

samuel40791765:
Testing


crypto/fipsmodule/CMakeLists.txt
```diff
@@ -458,17 +458,29 @@ elseif(FIPS_SHARED)
...
+        OUTPUT fips_apple_start.o
+        COMMAND ${CMAKE_C_COMPILER} -arch ${CMAKE_SYSTEM_PROCESSOR} -isysroot ${CMAKE_OSX_SYSROOT} -c ${CMAKE_CURRENT_SOURCE_DIR}/fips_shared_library_marker.c -DAWSLC_FIPS_SHARED_START -o fips_apple_start.o
+        DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/fips_shared_library_marker.c
+      )
+      add_custom_command(
+        OUTPUT fips_apple_end.o
+        COMMAND ${CMAKE_C_COMPILER} -arch ${CMAKE_SYSTEM_PROCESSOR} -isysroot ${CMAKE_OSX_SYSROOT} -c ${CMAKE_CURRENT_SOURCE_DIR}/fips_shared_library_marker.c -DAWSLC_FIPS_SHARED_END -o fips_apple_end.o
+        DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/fips_shared_library_marker.c
+      )
+    endif()
```
skmcgrail, line 483: Instead of duplicating the entire, can we just have this append to a variable that we set that is empty by default? And then just template that variable into the command? It would be just an empty-string when it expands or it would have the "-mmacosx-version-min=value"

skmcgrail (approved):

justsmth (approved):

PR #1251, opened 2023-10-16
 Fix SipHash for big-endian

justsmth:
### Issues:
N/A

### Description of changes: 
Fix SipHash implementation for big-endian architectures.

### Call-outs:
N/A

### Testing:
`SipHash.*` test pass for both ppc32 and ppc64.

ppc32:
```
❯ ppc-qemu.sh ./crypto/crypto_test --gtest_filter="SipHash.*"
Note: Google Test filter = SipHash.*
[==========] Running 2 tests from 1 test suite.
[----------] 2 tests from SipHash
[ RUN      ] SipHash.Basic
[       OK ] SipHash.Basic (3 ms)
[ RUN      ] SipHash.Vectors
[       OK ] SipHash.Vectors (115 ms)
[----------] 2 tests from SipHash (120 ms total)

[==========] 2 tests from 1 test suite ran. (129 ms total)
[  PASSED  ] 2 tests.
```

ppc64:
```
❯ ppc64-qemu.sh ./crypto/crypto_test --gtest_filter="SipHash.*"   
Note: Google Test filter = SipHash.*
[==========] Running 2 tests from 1 test suite.
[----------] 2 tests from SipHash
[ RUN      ] SipHash.Basic
[       OK ] SipHash.Basic (2 ms)
[ RUN      ] SipHash.Vectors
[       OK ] SipHash.Vectors (122 ms)
[----------] 2 tests from SipHash (126 ms total)

[==========] 2 tests from 1 test suite ran. (133 ms total)
[  PASSED  ] 2 tests.
```

### DIFF
--- a/crypto/siphash/siphash.c
+++ b/crypto/siphash/siphash.c
@@ -77,6 +77,5 @@ uint64_t SIPHASH_24(const uint64_t key[2], const uint8_t *input,
   siphash_round(v);
   siphash_round(v);
 
-  uint64_t b = v[0] ^ v[1] ^ v[2] ^ v[3];
-  return CRYPTO_load_u64_le(&b);
+  return v[0] ^ v[1] ^ v[2] ^ v[3];
 }
--- a/crypto/siphash/siphash.c
+++ b/crypto/siphash/siphash.c
@@ -42,8 +42,13 @@ uint64_t SIPHASH_24(const uint64_t key[2], const uint8_t *input,
   const size_t orig_input_len = input_len;
 
   uint64_t v[4], k0, k1;
-  CRYPTO_store_u64_le(&k0, key[0]);
-  CRYPTO_store_u64_le(&k1, key[1]);
+#ifdef OPENSSL_BIG_ENDIAN
+  k0 = CRYPTO_bswap8(key[0]);
+  k1 = CRYPTO_bswap8(key[1]);
+#else
+  k0 = key[0];
+  k1 = key[1];
+#endif
   v[0] = k0 ^ UINT64_C(0x736f6d6570736575);
   v[1] = k1 ^ UINT64_C(0x646f72616e646f6d);
   v[2] = k0 ^ UINT64_C(0x6c7967656e657261);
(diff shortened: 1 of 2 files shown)


crypto/siphash/siphash.c
```diff
@@ -75,5 +77,6 @@ uint64_t SIPHASH_24(const uint64_t key[2], const uint8_t *input,
   siphash_round(v);
   siphash_round(v);
 
-  return v[0] ^ v[1] ^ v[2] ^ v[3];
+  uint64_t b = v[0] ^ v[1] ^ v[2] ^ v[3];
+  return CRYPTO_load_u64_le(&b);
```
andrewhop, line 81: This introduces a little more work for the existing LE path. How do you see the tradeoff of simplifying our code (same for both LE and BE) and performance? In other functions we've handled the two orders differently, why not here?

crypto/siphash/siphash_test.cc
```diff
@@ -34,8 +34,8 @@ TEST(SipHash, Basic) {
   for (unsigned i = 0; i < sizeof(input); i++) {
     input[i] = i;
   }
-
-  EXPECT_EQ(UINT64_C(0xa129ca6149be45e5),
+  uint64_t expected = UINT64_C(0xa129ca6149be45e5);
+  EXPECT_EQ(CRYPTO_load_u64_le(&expected),
```
andrewhop, line 38: Do you know why you need this? I thought creating a variable with a compile time constant value is the same on BE and LE.

crypto/siphash/siphash.c
justsmth: My bad. This wasn't needed.

crypto/siphash/siphash_test.cc
justsmth: I was confused. The [SipHash paper](https://cr.yp.to/siphash/siphash-20120918.pdf) explicitly says "k0 and k1 are the little-endian 64-bit words encoding the key k", and I assumed that the resulting hash bytes would also need to be flipped on big-endian; they do not.

The SipHash result **is** the *64-bit-integer-value* (represented in whatever the native byte ordering is) -- the SipHash result is not "bytes with a particular ordering".

I'm updating the logic and the tests for the next commit.

andrewhop (approved):

andrewhop (approved):

skmcgrail (approved):

PR #1252, opened 2023-10-16
Add a build option to the assembler to retain local symbols.

nebeid:
### Issues:
Resolves CryptoAlg-2142

### Description of changes: 
Formal verification needed some local symbols to be visible in the symbol table. This build option helps achieve that.

### Testing:
Building with `cmake -DKEEP_LOCAL_SYMBOLS=1 -DCMAKE_BUILD_TYPE=Release -GNinja .. `
- On Graviton3, searching for `K512` in the output of `objdump --syms crypto_test | less` yielded a result.
- On Intel x86_64 (with avx capability but not vaes and vpclmulqdq, which are needed for AES AVX512 implementation), searching for the desired symbols in the symbol table of `crypto_test` yielded results as well.


nebeid:
Is it worth adding a test to `tests/ci/run_minimal_tests.sh`?

skmcgrail (dismissed):

skmcgrail:
> Is it worth adding a test to `tests/ci/run_minimal_tests.sh`?

If this will be used by the formal verification tooling then I assume that would provide coverage? Regardless I'm not sure how useful this is to test if its just for visibility to aid in debugging / writing FV tooling.

Out of curiosity odes this cause issues if you do this with a FIPS build? (`-DFIPS=1`)

CMakeLists.txt
```diff
@@ -633,6 +633,10 @@ if(GCOV)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fprofile-arcs -ftest-coverage")
 endif()
 
+if(KEEP_LOCAL_SYMBOLS)
```
justsmth, line 636: Since this only affects the assembler should the flag be `KEEP_LOCAL_ASM_SYMBOLS`?

nebeid: I changed it to `KEEP_ASM_LOCAL_SYMBOLS`.

nebeid:
> 
> Out of curiosity odes this cause issues if you do this with a FIPS build? (`-DFIPS=1`) (Edit: I guess no since the CI passed).

I added a test to the FIPS tests and used a side effect of `break-hash.go` to check that the module hash didn't change with this option in a FIPS release build.

nebeid:
> I added a test to the FIPS tests and used a side effect of `break-hash.go` to check that the module hash didn't change with this option in a FIPS release build. I can revert it if deemed to be an overkill.

The hash with and without the option is the same in Release and Debug FIPS builds. But Release build with gcc7 doesn't finish the break-kat tests. That's why the test uses a Debug build.

skmcgrail (approved):

justsmth (approved):

PR #1254, opened 2023-10-16
Add SDE+ASAN CI dimension

samuel40791765:
### Issues:
Resolves `CryptoAlg-1854`

### Description of changes: 
Rehashing the old PR of https://github.com/aws/aws-lc/pull/847. Hopefully the CI magically passes without hanging after a SDE version update. If this still doesn't work, I'll try against valgrind instead.

### Call-outs:
N/A

### Testing:
New CI


crypto/dynamic_loading_test.c
```diff
@@ -105,13 +105,15 @@ int main(int argc, char *argv[]) {
     exit(1);
   }
 
-  printf("PASS\n");
+  fprintf(stdout, "PASS\n");
+  fflush(stdout);
```
andrewhop, line 109: What was happening if you didn't have this? Shouldn't it get flushed when main exits? How did you figure this out?

tests/ci/cdk/cdk/codebuild/github_ci_linux_x86_omnibus.yaml
```diff
@@ -341,14 +361,6 @@ batch:
         variables:
           AWS_LC_CI_TARGET: "tests/ci/run_prefix_tests.sh"
 
-    - identifier: amazonlinux2_gcc7x_intel_sde_x86_64
```
andrewhop, line 344: It makes sense that we needed the new version to test ASAN + SDE, what are your thoughts about keeping AL2 + SDE? Most internal teams will be running on AL2 and I like testing all CPUs with it. Granted if we have Ubuntu + SDE, I'm not sure we need AL2 + SDE. At that point are we testing the OS/system calls on various CPUs more than AWS-LC?

tests/ci/common_posix_setup.sh
```diff
@@ -221,6 +221,16 @@ function print_executable_information {
   fi
 }
 
+function sde_getenforce_check {
+  # Based on Intel SDE README, SELinux should be turned off to allow pin to work.
```
andrewhop, line 225: Where do we turn this off?

skmcgrail: It's dependent on the host Linux kernel, so that will be left up to the host running this job. So we are going to be at the mercy of CodeBuild. It appears (at least for AL2023 SELinux is enabled in permissive mode by default), not sure what CodeBuild is using.

skmcgrail (dismissed):

tests/ci/common_posix_setup.sh
samuel40791765: Yeah it depends on the host. It's already disabled in our case, we just have an additional check here to make sure.

tests/ci/cdk/cdk/codebuild/github_ci_linux_x86_omnibus.yaml
samuel40791765: Sure we could keep the AL2 dimension and use Ubuntu for SDE+ASAN instead. I like the idea of testing both platforms. I was initially trying to upgrade this to AL2023, but chose ubuntu instead since AL2023 was [phasing out 32-bit runtime support](https://docs.aws.amazon.com/linux/al2023/ug/compare-with-al2.html#i686) (which `sde` needs).

crypto/dynamic_loading_test.c
samuel40791765: The issue was `main` never exits if the tests hang :(. Without `fflush`, `PASS\n` would be buffered and never printed. I configured the go test runner to dump the output if the tests were hanging, so this had been pretty obvious when I was debugging this.
* `fflush` documentation under `EXAMPLES`: https://pubs.opengroup.org/onlinepubs/000095399/functions/fflush.html

andrewhop: Oh interesting, so the tests ran and did all the work but hangs on exit.

skmcgrail (approved):

andrewhop (approved):

PR #1256, opened 2023-10-18
Document Ed25519 verification operation

torben-hansen:
### Description of changes: 

Document Ed25519 workflow with respect to https://datatracker.ietf.org/doc/html/rfc8032.

### Call-outs:

Swapped the order of a few (data independent) operations. I see no cache impact.

Renamed few variables to align with naming in rfc8032.

### DIFF
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -301,29 +301,23 @@ int ED25519_sign(uint8_t out_sig[64], const uint8_t *message,
 
 int ED25519_verify(const uint8_t *message, size_t message_len,
                    const uint8_t signature[64], const uint8_t public_key[32]) {
-  ge_p3 A;
-  if ((signature[63] & 224) != 0 ||
-      !x25519_ge_frombytes_vartime(&A, public_key)) {
+
+  // Ed25519 verify: rfc8032 5.1.7
+
+  // Step: rfc90832 5.1.7.1
+  // Decode signature as:
+  //  - signature[0:31]: encoded point R, aliased to R_expected.
+  //  - signature[32:61]: integer S.
+  if ((signature[63] & 224) != 0) {
     return 0;
   }
+  uint8_t R_expected[32];
+  OPENSSL_memcpy(R_expected, signature, 32);
+  uint8_t S[32];
+  OPENSSL_memcpy(S, signature + 32, 32);
 
-  fe_loose t;
-  fe_neg(&t, &A.X);
-  fe_carry(&A.X, &t);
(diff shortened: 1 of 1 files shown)


crypto/curve25519/curve25519.c
```diff
@@ -301,37 +301,31 @@ int ED25519_sign(uint8_t out_sig[64], const uint8_t *message,
...
-  OPENSSL_memcpy(scopy, signature + 32, 32);
+  // Ed25519 verify: rfc8032 5.1.7
 
-  // https://tools.ietf.org/html/rfc8032#section-5.1.7 requires that s be in
-  // the range [0, order) in order to prevent signature malleability.
+  // Step: rfc90832 5.1.7.1
+  // Decode signature as:
+  //  - signature[0:31]: encoded point R, aliased to R_expected.
+  //  - signature[32:61]: integer S.
+  if ((signature[63] & 224) != 0) {
```
skmcgrail, line 311: What's the purpose of the magic value `224` out of curiosity? Didn't find it in the RFC at a quick glance.

crypto/curve25519/curve25519.c
```diff
@@ -341,23 +335,51 @@ int ED25519_verify(const uint8_t *message, size_t message_len,
...
 
-  ge_p2 R;
-  ge_double_scalarmult_vartime(&R, h, &A, scopy);
+  // Compute R_have <- [S]B - [k]A'.
+  ge_p2 R_have;
+  ge_double_scalarmult_vartime(&R_have, k, &A, S);
 
-  uint8_t rcheck[32];
-  x25519_ge_tobytes(rcheck, &R);
+  uint8_t R_have_enc[32];
```
nebeid, line 378: Maybe add a comment here or rename it to `R_have_encoded` since "enc" can be read as "encrypted".

crypto/curve25519/curve25519.c
nebeid: I understand it's coming from this guarantee when signing in 5.1.6.6:
> Form the signature of the concatenation of R (32 octets) and the little-endian encoding of S (32 octets; the three most significant bits of the final octet are always zero).

But I didn't dive into the exact details of encoding.
It's worth adding a comment.

torben-hansen: 5.1.6.6 correct. We have 224 = 0xE0

crypto/curve25519/curve25519.c
torben-hansen: makes sense!

skmcgrail (approved):

nebeid (approved):

PR #1269, opened 2023-10-27
Create dirs in cmake build dir with default permissions

sfod:
### Description of changes: 

The downstream project, aws-iot-device-sdk-cpp-v2, fails to perform cmake configuration when sources are read-only. One of the issues is that aws-lc fails to perform cmake configuration for read-only sources.

By default, the `file(COPY)` cmake command preserves input file permissions. When source directory is read-only, this causes destination directory to be also read-only, which causes cmake configuration to fail when it tries to create a new file in the destination directory.
NO_SOURCE_PERMISSIONS option resolves the issue.

### Testing:

The issue appear in Linux systems:
```
git clone --recursive https://github.com/aws/aws-lc.git
chmod -R -w aws-lc
cmake -S aws-lc -B build/
```


torben-hansen:
Thanks. However, `NO_SOURCE_PERMISSIONS` was not [introduced before 3.19](https://cmake.org/cmake/help/latest/command/configure_file.html#:~:text=they%20are%20created.-,NO_SOURCE_PERMISSIONS,New%20in%20version%203.19.,-Do%20not%20transfer). Which might turn out to be problematic.

torben-hansen:
oops, wrong link. This is for `file` command: https://cmake.org/cmake/help/latest/command/file.html#:~:text=TARGET_PROPERTY%3Aprop%3E).-,NO_SOURCE_PERMISSIONS,New%20in%20version%203.20.,-The%20generated%20file

Which states 3.20, slightly worse.

sfod:
"new in version 3.20" is about `file(GENERATE)`. If you scroll up, you'll see that.
I was also confused first time I looked at the docs.

torben-hansen (approved):

sfod:
I didn't find in documentation which permissions will be used for a destination directory when NO_SOURCE_PERMISSIONS option is used. Fortunately, the source code is pretty straightforward:
https://github.com/Kitware/CMake/blob/v3.27.7/Source/cmFileCopier.cxx#L154

skmcgrail (approved):

PR #1271, opened 2023-10-27
Refactor ED25519_keypair into hw and nohw backend

torben-hansen:
### Description of changes: 

Refactors `ED25519_keypair` into a no-hardware and a hardware path. Existing backend is the former. Preparation for s2n-bignum ED25519.

s2n-bignum is lacking SHA512. So unfortunately, I can't punt this computation to s2n-bignum. 

PR chain:
1. (this PR) refactor `ED25519_keypair`
2. refactor `ED25519_sign`
3. refactor `ED25519_verify`
4. various small changes to improve code quality
5. import s2n-bignum Ed25519 routines
6. connect s2n-bignum Ed25519 backend to frontend

### DIFF
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -31,6 +31,25 @@
 #include "../internal.h"
 #include "../fipsmodule/cpucap/internal.h"
 
+// X25519 [1] and Ed25519 [2] is an ECDHE protocol and signature scheme,
+// respectively. This file contains an implementation of both using two
+// different backends:
+// 1) One backed is a pure C backend that should work on any platform.
+// 2) The other backend is machine-optimized using s2n-bignum [3] as backend.
+//
+// [1]: https://datatracker.ietf.org/doc/html/rfc7748
+// [2]: https://datatracker.ietf.org/doc/html/rfc8032
+// [3]: https://github.com/awslabs/s2n-bignum
+//
+// "Clamping":
+// Both X25519 and Ed25519 contain "clamping" steps; bit-twiddling, masking or
+// setting specific bits. Generally, the bit-twiddling is to avoid common
+// implementation errors and weak instances. Details can be found through the
+// following two references:
+// * https://mailarchive.ietf.org/arch/msg/cfrg/pt2bt3fGQbNF8qdEcorp-rJSJrc/
+// * https://neilmadden.blog/2020/05/28/whats-the-curve25519-clamping-all-about
+
+
 // If (1) x86_64 or aarch64, (2) linux or apple, and (3) OPENSSL_NO_ASM is not
 // set, s2n-bignum path is capable.
 #if ((defined(OPENSSL_X86_64) &&                                               \
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -240,35 +240,54 @@ static void x25519_s2n_bignum_public_from_private(
 }
 
 // Stub function until ED25519 lands in s2n-bignum
-static void ed25519_keypair_from_seed_s2n_bignum(uint8_t out_public_key[32],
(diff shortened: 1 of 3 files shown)


crypto/curve25519/curve25519.c
```diff
@@ -240,35 +240,54 @@ static void x25519_s2n_bignum_public_from_private(
...
+  // Compute SHA512(seed).
   uint8_t az[SHA512_DIGEST_LENGTH];
   SHA512(seed, ED25519_SEED_LEN, az);
 
+  // Step: rfc8032 5.1.5.2
+  // 248 = 11111000_2 clears lowest 3 bits.
+  // For last octet: clear highest highest bit and set second highest bit,
+  // respectively.
+  // 127 = 01111111_2
+  // 64 = 01000000_2
```
rod-chapman: Could you explain _why_ these bit manipultions are important, rather than just what they do? I had to look... it's all explained here:
 https://neilmadden.blog/2020/05/28/whats-the-curve25519-clamping-all-about/
Essentially, it ensures the the number is "large" (setting the next-to-top-bit), and also an exact multiple of 8 which is the co-factor of the group...

torben-hansen: This is a general step in all subroutines. I added some general information.

rod-chapman: Looks good

crypto/curve25519/curve25519.c
```diff
@@ -31,6 +31,25 @@
 #include "../internal.h"
 #include "../fipsmodule/cpucap/internal.h"
 
+// X25519 [1] and Ed25519 [2] is an ECDHE protocol and signature scheme,
+// respectively. This file contains an implementation of both using two
+// different backends:
+// 1) One backed is a pure C backend that should work on any platform.
```
dkostic, line 37:
```suggestion
// 1) One backend is a pure C backend that should work on any platform.
```

dkostic (approved):

crypto/curve25519/curve25519.c
```diff
@@ -234,28 +258,53 @@ static void x25519_s2n_bignum_public_from_private(
...
+void ED25519_keypair(uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
+  uint8_t out_private_key[ED25519_PRIVATE_KEY_LEN]) {
+
+  // Ed25519 key generation: rfc8032 5.1.5
+  // Private key is 32 octets of random data.
   uint8_t seed[ED25519_SEED_LEN];
   RAND_bytes(seed, ED25519_SEED_LEN);
+
+  // Public key generation is handled in a separate function. See function
+  // description why this is useful.
```
nebeid, line 307: I don't know where in the function it's explained that it's useful to separate it?

torben-hansen: I believe the standard is to put the description at the function declaration: https://github.com/aws/aws-lc/blob/main/include/openssl/curve25519.h#L96-L100

nebeid, line 307: Nit: The first comment in this section refers to "Private key", but it's the seed that's generated. What would be a good description for it? The seed of the private key?
The second comment refers to the public key generation, is it not a keypair generation, i.e. both private and public keys?

nebeid: Ah, yes, thanks.

torben-hansen: Naming used refers to the standard, such that one can follow the code and spec at the same time.

The private key is the seed. But in the spec it's initially called seed. The public key is generated from that seed.

torben-hansen: In our implementation, the private key is the unhashed seed. Not sure why it was implemented this way.

But one could have chosen to use the hashed seed as well. This would break the API by now though.

nebeid (approved):

dkostic (approved):

PR #1272, opened 2023-10-28
Rearrange X509 symbols for Monit support

samuel40791765:
### Issues:
Addresses `CryptoAlg-2167`

### Description of changes: 
* Monit depends on [these symbols](https://bitbucket.org/tildeslash/monit/src/a869c3aad38f34a3b24e9b510b9ee48e3d7b3d9f/src/ssl/Ssl.c?at=master#lines-436:448).
* `X509_OBJECT_free_contents` was removed from OpenSSL 1.0.2 -> 1.1.1 in https://github.com/openssl/openssl/commit/6ddbb4cd92ad8c9ebf874e5d9920c122696b7477. We already expose our version string as 1.1.1, so we can remove this.
* OpenSSL 1.1.1 renamed `X509_STORE_get_by_subject` to `X509_STORE_CTX_get_by_subject` and made an alias to the original: https://github.com/openssl/openssl/commit/6ddbb4cd92ad8c9ebf874e5d9920c122696b7477

### Call-outs:
N/A

### Testing:
Minor new test

### DIFF
--- a/include/openssl/x509.h
+++ b/include/openssl/x509.h
@@ -2906,9 +2906,6 @@ OPENSSL_EXPORT X509_LOOKUP_METHOD *X509_LOOKUP_file(void);
 OPENSSL_EXPORT int X509_STORE_add_cert(X509_STORE *ctx, X509 *x);
 OPENSSL_EXPORT int X509_STORE_add_crl(X509_STORE *ctx, X509_CRL *x);
 
-OPENSSL_EXPORT int X509_STORE_get_by_subject(X509_STORE_CTX *vs, int type,
-                                             X509_NAME *name, X509_OBJECT *ret);
-
 OPENSSL_EXPORT int X509_LOOKUP_ctrl(X509_LOOKUP *ctx, int cmd, const char *argc,
                                     long argl, char **ret);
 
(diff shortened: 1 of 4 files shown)


include/openssl/x509.h
```diff
@@ -2997,6 +3002,19 @@ OPENSSL_EXPORT void X509_STORE_CTX_set0_param(X509_STORE_CTX *ctx,
 OPENSSL_EXPORT int X509_STORE_CTX_set_default(X509_STORE_CTX *ctx,
                                               const char *name);
 
+// X509_STORE_get_by_subject is an alias to |X509_STORE_CTX_get_by_subject| in
+// OpenSSL 1.1.1.
+#define X509_STORE_get_by_subject X509_STORE_CTX_get_by_subject
```
andrewhop, line 3007: Does this cause an issue with `X509_STORE_get_by_subject` being exported up on line 2909 as a symbol but it's just a macro?

samuel40791765: Oh man, great catch I just forgot about removing the original symbol when I discovered it was changed to an alias.
For the record, I think subsequent symbols defined at the bottom of the file take priority over the ones at the top, so this would cause an overwrite.

andrewhop (approved):

skmcgrail (approved):

PR #1282, opened 2023-11-02
Don't use expired certificates if possible.

samuel40791765:
### Issues:
Resolves `CryptoAlg-2052`

### Description of changes: 
Some internal teams have ran into unsuspecting issues when migrating over to AWS-LC. If an expired cert was placed before a possible valid cert within the stack, AWS-LC would break and fail certificate verification. This would not happen if the valid certificate was placed first. The behavioral difference was pinned down to an older commit from OpenSSL 1.0.2 -> OpenSSL 1.1.1: https://github.com/openssl/openssl/commit/0930251df814f3993bf2c598761e0c7c6d0d62a2, that we didn't have support for.

### Call-outs:
Logic was applied from the original commit. 

### Testing:
New certificate test

### DIFF
--- a/crypto/x509/x509_lu.c
+++ b/crypto/x509/x509_lu.c
@@ -643,7 +643,7 @@ int X509_STORE_CTX_get1_issuer(X509 **issuer, X509_STORE_CTX *ctx, X509 *x) {
   }
   CRYPTO_MUTEX_unlock_write(&ctx->ctx->objs_lock);
   if(*issuer) {
-    CRYPTO_refcount_inc(&(*issuer)->references);
+    X509_up_ref(*issuer);
   }
   return ret;
 }
--- a/crypto/x509/x509_lu.c
+++ b/crypto/x509/x509_lu.c
@@ -604,8 +604,10 @@ int X509_STORE_CTX_get1_issuer(X509 **issuer, X509_STORE_CTX *ctx, X509 *x) {
   }
   // If certificate matches all OK
   if (ctx->check_issued(ctx, x, obj.data.x509)) {
-    *issuer = obj.data.x509;
-    return 1;
+    if (x509_check_cert_time(ctx, obj.data.x509, /*suppress_error*/1)) {
+      *issuer = obj.data.x509;
+      return 1;
+    }
   }
   X509_OBJECT_free_contents(&obj);
 
@@ -627,13 +629,21 @@ int X509_STORE_CTX_get1_issuer(X509 **issuer, X509_STORE_CTX *ctx, X509 *x) {
       }
       if (ctx->check_issued(ctx, x, pobj->data.x509)) {
         *issuer = pobj->data.x509;
-        X509_OBJECT_up_ref_count(pobj);
         ret = 1;
-        break;
+        // Break the loop with a match if the time check is valid, otherwise
+        // we continue searching. We leave the last tested issuer certificate in
+        // |issuer| on purpose. This returns the closest match if none of the
+        // candidate issuer certificates' timestamps were valid.
+        if (x509_check_cert_time(ctx, *issuer, /*suppress_error*/1)) {
+          break;
+        }
       }
     }
   }
   CRYPTO_MUTEX_unlock_write(&ctx->ctx->objs_lock);
+  if(*issuer) {
+    CRYPTO_refcount_inc(&(*issuer)->references);
+  }
   return ret;
 }
 
--- a/crypto/x509/x509_lu.c
+++ b/crypto/x509/x509_lu.c
@@ -598,6 +598,7 @@ int X509_STORE_CTX_get1_issuer(X509 **issuer, X509_STORE_CTX *ctx, X509 *x) {
   X509_OBJECT obj, *pobj;
   int idx, ret;
   size_t i;
+  *issuer = NULL;
   xn = X509_get_issuer_name(x);
   if (!X509_STORE_CTX_get_by_subject(ctx, X509_LU_X509, xn, &obj)) {
     return 0;
--- a/crypto/x509/x509_vfy.c
+++ b/crypto/x509/x509_vfy.c
@@ -491,7 +491,10 @@ end:
 }
 
 // Given a STACK_OF(X509) find the issuer of cert (if any)
-
+//
+// |find_issuer| will directly return the pointer of the corresponding index
+// within |sk|. Callers of |find_issuer| should remember to bump the reference
+// count of the returned |X509| if the call is successful.
 static X509 *find_issuer(X509_STORE_CTX *ctx, STACK_OF(X509) *sk, X509 *x) {
   size_t i;
   X509 *issuer, *candidate = NULL;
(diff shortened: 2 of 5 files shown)


crypto/x509/x509_lu.c
```diff
@@ -613,13 +615,21 @@ int X509_STORE_CTX_get1_issuer(X509 **issuer, X509_STORE_CTX *ctx, X509 *x) {
       }
       if (ctx->check_issued(ctx, x, pobj->data.x509)) {
         *issuer = pobj->data.x509;
-        X509_OBJECT_up_ref_count(pobj);
         ret = 1;
```
WillChilds-Klein, line 618: we consider the lookup successful even if all of the issuer certs are expired?

crypto/x509/x509_lu.c
```diff
@@ -613,13 +615,21 @@ int X509_STORE_CTX_get1_issuer(X509 **issuer, X509_STORE_CTX *ctx, X509 *x) {
...
-        break;
+        // Break the loop with a match if the time check is valid, otherwise
+        // we continue searching. We leave the last tested issuer certificate in
+        // |issuer| on purpose. This returns the closest match if none of the
+        // candidate issuer certificates' timestamps were valid.
+        if (x509_check_cert_time(ctx, *issuer, /*suppress_error*/1)) {
+          break;
```
WillChilds-Klein, line 624: is the comment on L579 still true?

```
// However it will fill the cache with all matching certificates, so we can
// examine the cache for all matches.
```

crypto/x509/x509_lu.c
samuel40791765: Yeah the reason being that `X509_STORE_CTX_get1_issuer` is mainly used to find a matching issuer. The cert may be expired, but it's still considered a match.
This aligns with what OpenSSL1.1.1 has: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/x509/x509_lu.c#L724-L735

crypto/x509/x509_lu.c
samuel40791765: I took a closer look and I think it still holds. The "cache" being mentioned here is `ctx->ctx->objs`, which is being looped over and holds all matching certs.

crypto/x509/x509_lu.c
samuel40791765: Your comments led me to take a closer look at `X509_STORE_CTX_get1_issuer` and I realized that this hadn't been properly tested with my initial set up. I added some tests to specifically test the new function behavior!

crypto/x509/x509_lu.c
```diff
@@ -598,14 +598,17 @@ int X509_STORE_CTX_get1_issuer(X509 **issuer, X509_STORE_CTX *ctx, X509 *x) {
...
   xn = X509_get_issuer_name(x);
   if (!X509_STORE_CTX_get_by_subject(ctx, X509_LU_X509, xn, &obj)) {
     return 0;
   }
   // If certificate matches all OK
   if (ctx->check_issued(ctx, x, obj.data.x509)) {
-    *issuer = obj.data.x509;
-    return 1;
+    if (x509_check_cert_time(ctx, obj.data.x509, /*suppress_error*/1)) {
+      *issuer = obj.data.x509;
```
skmcgrail, line 609: Not obvious at quick glance, is there a reason why this reference doesn't need to be bumped?

justsmth: A ref count gets bumped in [`X509_STORE_CTX_get_by_subject`](https://github.com/aws/aws-lc/blob/044d529f6c84c30849a2beff60b1660932f31c93/crypto/x509/x509_lu.c#L311).  I think we would only need to "free" if we return w/o setting the `*issuer` to point to this value, which it does below.

samuel40791765: @justsmth had the same question. 

The code is pretty confusing. 
Step 1: We define `obj` at the top on L598 and `X509_STORE_CTX_get_by_subject` does a reference bump on the retrieved `obj` if the function succeeds. 
Step 2: If successful, the data within `obj` is then assigned to `issuer` and returned. Technically the reference bump has been done `X509_STORE_CTX_get_by_subject`, so we don't need to do an additional bump here.
Step 3: If failed, we free `obj` which contains the bumped reference on L613.

skmcgrail: Ah I should have expanded the resolved comments. Thanks!

skmcgrail (dismissed):

crypto/x509/x509_vfy.c
```diff
@@ -494,14 +494,17 @@ int X509_verify_cert(X509_STORE_CTX *ctx) {
...
-  X509 *issuer;
+  X509 *issuer, *candidate = NULL;
   for (i = 0; i < sk_X509_num(sk); i++) {
     issuer = sk_X509_value(sk, i);
     if (ctx->check_issued(ctx, x, issuer)) {
-      return issuer;
+      candidate = issuer;
+      if (x509_check_cert_time(ctx, candidate, /*suppress_error*/1)) {
+        break;
+      }
```
justsmth, line 504: With expired certs, we might iterate over multiple candidates. When we change candidates, don't we need a call to "free" the previous?

crypto/x509/x509_lu.c
```diff
@@ -627,13 +630,21 @@ int X509_STORE_CTX_get1_issuer(X509 **issuer, X509_STORE_CTX *ctx, X509 *x) {
...
+        // candidate issuer certificates' timestamps were valid.
+        if (x509_check_cert_time(ctx, *issuer, /*suppress_error*/1)) {
+          break;
+        }
       }
     }
   }
   CRYPTO_MUTEX_unlock_write(&ctx->ctx->objs_lock);
+  if(*issuer) {
+    CRYPTO_refcount_inc(&(*issuer)->references);
```
justsmth, line 646: NP: For consistency, I think it would be better to use `X509_up_ref(*issuer)` here.

crypto/x509/x509_vfy.c
samuel40791765: I think `candidate` is assigned the pointer within the `sk` X509 stack. These return pointers within the stack without additional reference bumps.
https://github.com/aws/aws-lc/blob/fbe7a0646d707016d661c2471ad82908404c0d19/crypto/stack/stack.c#L128

If we try freeing the assigned pointer, the original `X509` index pointer within `sk` will be unintentionally freed as well. Because of this behavior, users of the `find_issuer` function have to bump the reference each time the call is successful.
* https://github.com/aws/aws-lc/blob/fbe7a0646d707016d661c2471ad82908404c0d19/crypto/x509/x509_vfy.c#L260-L267
* https://github.com/aws/aws-lc/blob/fbe7a0646d707016d661c2471ad82908404c0d19/crypto/x509/x509_vfy.c#L529-L531

This is pretty confusing as well, I'll add a comment above `find_issuer` to clarify this usage behavior.

WillChilds-Klein (dismissed):

justsmth (approved):

skmcgrail (dismissed):

crypto/x509/x509_lu.c
samuel40791765: No problem, I like that too.

justsmth (approved):

WillChilds-Klein (approved):

skmcgrail (approved):

PR #1299, opened 2023-11-10
Revert "Give up on qsort for sk_FOO_sort"

samuel40791765:
This reverts commit 1b87d59d7fdbcefc096965efb4af22068c491a08.

### Issues:
Resolves `P104663041`

### Description of changes: 
There's the [existing issue](https://github.com/openssl/openssl/issues/18708) in AWS-LC/OpenSSL where certificate verification will fail depending on an invalid certificate in the trust store coming first. The new sorting introduced a rearrangement of identical certificates within the trust store. This could technically increase the chance of teams running into the existing AWS-LC/OpenSSL issue mentioned earlier, if their invalid identical certificate is rearranged to the top of the stack.

We will have to introduce this change sooner or later to get around the UB sanitizer issue this commit was trying to resolve. But we should do additional verification to make sure we aren't changing the original certificate ordering. To be fair, sorting behavior has always been considered "unstable" and certain sorting expectations shouldn't be depended on. We can clarify this within the "revert" commit.

### Testing:
Apparently one of the new tests I had written was being effected as well.

### DIFF
--- a/crypto/stack/stack.c
+++ b/crypto/stack/stack.c
@@ -393,52 +393,20 @@ err:
   return NULL;
 }
 
-static size_t parent_idx(size_t idx) {
-  assert(idx > 0);
-  return (idx - 1) / 2;
+#if defined(_MSC_VER)
+struct sort_compare_ctx {
+  OPENSSL_sk_call_cmp_func call_cmp_func;
+  OPENSSL_sk_cmp_func cmp_func;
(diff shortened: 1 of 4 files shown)


andrewhop (approved):
Please don't squash these two commits.

skmcgrail (approved):

PR #1309, opened 2023-11-15
Wire up s2n-bignum Ed25519 backend

torben-hansen:
### Issues:

CryptoAlg-1603

### Description of changes: 

Wires up the Ed25519 s2n-bignum backend to the frontend. From now on, all Ed25519 operations will use s2n-bignum on supported hardware.

Most code-changes are due to the delocater.

### Call-outs:

Quite a few call-outs on this one.

Firstly, the new s2n-bignum code-path performs the same operations as the nohw code-path, just using the s2n-bignum functions. Since the interface currently exposed by s2n-bignum Ed25519 functions is not byte-level (i.e. `uint8_t`), we have to do a fair bit of copying. I considered doing this using a union with automatic type-punning. I believe we could have stayed inside the restrictive conditions of non-UB type-punning, but I just didn't do the detailed analysis. Instead, s2n-bignum should expose `uint8_t` level API, as the rest of AWS-LC does. Performance-wise, the copying is just tiny 32-bytes or so. It doesn't matter much in the grand scheme of things. I did try to implement a PoC with unions, and there were no true difference.

Secondly, I had to expand the delocater slightly. Immediates can have different offset computations. This is quite annoying to handle atm. But I made it work without breaking anything else. Expanded test slightly to not break this in the future. But the delocater test coverage is not exactly great atm.

Finally, there is an issue with the FIPS build. This is partly due to the delocater. We concatenate everything under the same compilation unit and resolve relocations. On trick is to regard known labels as a simple address grab. That is, we can just `adrp known_label`. But `adrp` spawns a relocation, so the delocater replaces this with `adr`. Unfortunately, the reach of `adr` is a mere 1MiB and it seems we reach this limit. I believe this is solvable. But to unblock for now, just don't build for fips... See CryptoAlg-2198 for details.

### Testing:

Existing testing will path through all this. But also added additional to the delocater.


torben-hansen:
Import missing functions and fix-up duplicate labels: https://github.com/aws/aws-lc/pull/1314

crypto/curve25519/curve25519.c
```diff
@@ -261,27 +263,192 @@ static void x25519_s2n_bignum_public_from_private(
...
+    edwards25519_scalarmulbase(uint64_point, uint64_hashed_seed);
+  } else if (curve25519_s2n_bignum_alt_capable() == 1) {
+    edwards25519_scalarmulbase_alt(uint64_point, uint64_hashed_seed);
+  } else {
+    abort();
+  }
+
+#elif defined(OPENSSL_AARCH64)
+
+  if (curve25519_s2n_bignum_alt_capable() == 1) {
```
skmcgrail, line 286: Does the order of these condition checks matter? For example for x86-64 we check `curve25519_s2n_bignum_no_alt_capable() == 1` first, but here for aarch64 we check `curve25519_s2n_bignum_alt_capable() == 1` first. Same question for similar code further below.

third_party/s2n-bignum/include/s2n-bignum_aws-lc.h
```diff
@@ -246,3 +246,81 @@ extern void bignum_copy_row_from_table_16_neon (uint64_t *z, const uint64_t *tab
 // Input table[height*32]; output z[32]
 extern void bignum_copy_row_from_table_32_neon (uint64_t *z, const uint64_t *table,
         uint64_t height, uint64_t idx);
+
+#if !defined(BORINGSSL_FIPS)
+// Reduction is modulo the order of the curve25519/edwards25519 basepoint,
+// which is n_25519 = 2^252 + 27742317777372353535851937790883648493.
+// Reduce modulo basepoint order, z := x mod n_25519
+// Input x[k]; output z[4]
+extern void bignum_mod_n25519(uint64_t z[static 4], uint64_t k, uint64_t *x);
```
skmcgrail, line 255: This is a data point for not making this a submodule, and leaving it as a subtree. Its much more complicated if we are patching the sources rather then using them as-is.

crypto/curve25519/curve25519.c
torben-hansen: Yes, definitely.

It's explained here: https://github.com/aws/aws-lc/blob/main/crypto/curve25519/curve25519.c#L164-L184. It uses the same reasoning so it wasn't changed for this PR.

third_party/s2n-bignum/include/s2n-bignum_aws-lc.h
torben-hansen: Yes - something that would need to be fixed before actually switching over...

crypto/curve25519/curve25519.c
```diff
@@ -73,40 +72,43 @@ OPENSSL_INLINE int curve25519_s2n_bignum_capable(void) {
...
+  return 1;
+#else
   return 0;
+#endif
 }
 
 // Stub functions if implementations are not compiled.
 // These functions have to abort, otherwise we risk applications assuming they
 // did work without actually doing anything.
+#if !defined(CURVE25519_S2N_BIGNUM_CAPABLE) || defined(BORINGSSL_FIPS)
```
dkostic, line 87: why do we even define the functions when `!defined(CURVE25519_S2N_BIGNUM_CAPABLE)`?

crypto/curve25519/curve25519.c
```diff
@@ -261,27 +263,192 @@ static void x25519_s2n_bignum_public_from_private(
...
+    scalarmulbase = edwards25519_scalarmulbase;
+    madd = bignum_madd_n25519;
+  } else {
+    abort();
+  }
+
+#else
+
+  scalarmulbase = edwards25519_scalarmulbase;
+  madd = bignum_madd_n25519;
```
dkostic, line 341: delete since we abort in this case?

crypto/curve25519/curve25519.c
torben-hansen: to avoid having id/def's everywhere below. So, just handle it here.

crypto/curve25519/curve25519.c
torben-hansen: I do it to avoid unused warnings (that turns into errors)

crypto/curve25519/curve25519.c
```diff
@@ -261,27 +263,192 @@ static void x25519_s2n_bignum_public_from_private(
...
+
+  if (curve25519_s2n_bignum_alt_capable() == 1) {
+    edwards25519_scalarmulbase_alt(uint64_point, uint64_hashed_seed);
+  } else if (curve25519_s2n_bignum_no_alt_capable() == 1) {
+    edwards25519_scalarmulbase(uint64_point, uint64_hashed_seed);
+  } else {
+    abort();
+  }
+
+#else
```
dkostic, line 294: I think something like this would be simpler:
```
static bool curve25519_s2n_bignum_use_alt() {
#if defined(X86_64)
  return X;
#else
  return Y;
#endif
}
```
and then you could do:

```suggestion
#if defined(CURVE25519_S2N_BIGNUM_CAPABLE)
  if (curve25519_s2n_bignum_use_alt() == 1) {
    edwards25519_scalarmulbase_alt(uint64_point, uint64_hashed_seed);
  } else {
    edwards25519_scalarmulbase(uint64_point, uint64_hashed_seed);
  }

#else
```

in all the other places as well.

torben-hansen: this has to fail at run-time, otherwise we return without doing any work.

dkostic: agreed offline that we'll revisit the design later.

dkostic (approved):

crypto/curve25519/curve25519.c
nebeid: maybe you can use
```
(void) scalarmulbase;
(void) madd;
```
for readability.

crypto/curve25519/curve25519.c
```diff
@@ -261,27 +263,192 @@ static void x25519_s2n_bignum_public_from_private(
 #endif
 }
 
-// Stub function until ED25519 lands in s2n-bignum
 static void ed25519_public_key_from_hashed_seed_s2n_bignum(
   uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
   uint8_t az[SHA512_DIGEST_LENGTH]) {
+
```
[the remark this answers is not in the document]
torben-hansen: I attempted to factor out as much as possible. The remaining "duplication" is due to using different functions working on different internal representations. For example, this can illustrate it:
nohw path:
```
  // Reduce r modulo the order of the base-point B.
  x25519_sc_reduce(r);
  ge_p3 R;
  // Compute [r]B.
  x25519_ge_scalarmult_base(&R, r);
```
while s2n-bignum patch:
```
  uint64_t R[8] = {0};  
  // Reduce r modulo the order of the base-point B.
  bignum_mod_n25519(uint64_r, 8, uint64_r);

  // Compute [r]B.
  scalarmulbase(R, uint64_r);
```
so, in former case R is some internal representation of a group element. In the latter case, R is an uin64_t  buffer.

I think in this case, having a bit of duplication is actually more read-friendly than not having it. Hopefully, in the future we can just have a high-level s2n-bignum function that would remove this duplication.

nebeid (approved):

PR #1311, opened 2023-11-16
Implement support for EVP_chacha20_poly1305

geedo0:
### Issues:
Resolves CryptoAlg-2084

### Description of changes: 
ChaCha20-Poly1305 is only exposed via the BoringSSL EVP_AEAD interface. While it provides a cleaner API to our customers it lacks compatibility with the corpus of applications that are developed against upstream OpenSSL's EVP_CIPHER interface. This PR implements that interface and aims for compatibility with existing consumers of that interface.

From a behavioral perspective, the fundamental difference between the EVP_AEAD and EVP_CIPHER interface is that they operate as one-shot and streaming respectively. By providing the EVP_CIPHER interface, we also introduce ChaCha20-Poly1305 streaming (i.e. the ability to run the the cipher in arbitray increments down to a single byte) support as a new capability.

We implement this by leveraging a fallback ChaCha20-Poly1305 implementation that already exists in AWS-LC which traces back to upstream OpenSSL and supports streaming. This implementation is still assembly optimized, but presumably not as performant as the preferred implementation. We then draw inspiration from upstream OpenSSL to implement the EVP_CIPHER APIs.

### Call-outs:
I know we're doing work in compiling AWS-LC for big endian systems, but it's
unclear to me the full scope/requirements are around that effort. There are
certainly areas in this code that assume little endian on a 32-bit word size and
that's because endianess was not considered. If that was a mistake, please flag
it now.

### Testing:
- Added unit tests which run through all the same test vectors that the EVP_AEAD
interface used, but with the EVP_CIPHER API.
- Implemented unit test for the chacha20-poly1305 WychefProof vectors.
- If lucky, CI will pass.


geedo0:
Failing CI for builds that define `OPENSSL_NO_ASM`. Need to add a fallback to a C-based ChaCha20 implementation.

geedo0:
All CI failures fixed. This had the added benefit of improving re-use between the EVP_CIPHER implementation and the fallback EVP_AEAD implementation. This also fixed a lot of endianess issues because the re-used code block makes a more conscious effort at that.

crypto/cipher_extra/cipher_test.cc
```diff
@@ -538,6 +540,161 @@ TEST(CipherTest, CAVP_TDES_ECB) {
...
+  // Finish decryption
+  out_len = 0;
+  res = EVP_DecryptFinal(dctx, NULL, &out_len);
+
+  // Tag validation, a valid result should end with a valid tag
+  ASSERT_EQ(res, result.IsValid() ? 1 : 0);
+  ASSERT_EQ(out_len, 0);
+
+  // Valid results should match the KATs
+  if (result.IsValid()) {
```
andrewhop, line 676: Did you look at `RunWycheproofTestCase` in aead_test.cc? That flow is easier to follow, there is no need to do a lot of this test if the test is supposed to fail.

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -78,11 +114,11 @@ static void poly1305_update_length(poly1305_state *poly1305, size_t data_len) {
 
 // calc_tag fills |tag| with the authentication tag for the given inputs.
 static void calc_tag(uint8_t tag[POLY1305_TAG_LEN], const uint8_t *key,
-                     const uint8_t nonce[12], const uint8_t *ad, size_t ad_len,
-                     const uint8_t *ciphertext, size_t ciphertext_len,
-                     const uint8_t *ciphertext_extra,
+                     const uint8_t nonce[CHACHA_IV_LEN], const uint8_t *ad,
+                     size_t ad_len, const uint8_t *ciphertext,
+                     size_t ciphertext_len, const uint8_t *ciphertext_extra,
```
andrewhop, line 119: For future PRs please separate simple refactors like this into a separate PR, it makes it much easier to review.

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,317 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+                                 const uint8_t user_key[CHACHA_KEY_LEN],
+                                 const uint8_t counter_nonce[CHACHA_CTR_IV_LEN])
+{
+  CIPHER_CHACHA_KEY *key = CC_KEY(ctx);
+  uint32_t i;
+  if (user_key) {
+    for (i = 0; i < (CHACHA_KEY_LEN / 4); i++) {
+      key->key[i] = *(((const uint32_t *) user_key) + i);
+    }
+  }
```
andrewhop, line 396: This will most likely hit a endian issue. Can this use something like `CRYPTO_load_*`?

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,317 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+  key->partial_len = 0;
+  return 1;
+}
+
+static int cipher_chacha20_poly1305_init(EVP_CIPHER_CTX *ctx,
+                                         const uint8_t *key,
+                                         const uint8_t *iv, int32_t enc) {
+  CIPHER_CHACHA_POLY_CTX *cipher_ctx = CCP_CTX(ctx);
+  if (!key && !iv) {
+    return 1;
```
andrewhop, line 411: This looks weird but is consistent with other EVP_CIPHER implementations. Is this documented?

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,317 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+  cipher_ctx->len.aad = 0;
+  cipher_ctx->len.text = 0;
+  cipher_ctx->pad_aad = 0;
+  cipher_ctx->poly_initialized = 0;
+  // Init can be called multiple times before starting the cipher to
+  // independently initialize any combination of Key/IV/NULL.
+  if (iv != NULL) {
+    // Start the counter at 0 and copy over the nonce(iv)
+    uint8_t counter_nonce[CHACHA_CTR_IV_LEN] = {0};
+    memcpy(counter_nonce + CHACHA_CTR_IV_LEN - CHACHA_IV_LEN, iv,
```
andrewhop, line 422:
```suggestion
    OPENSSL_memcpy(counter_nonce + CHACHA_CTR_IV_LEN - CHACHA_IV_LEN, iv,
```

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,317 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+    cipher_ctx->iv[1] = cipher_ctx->key.counter_nonce[2];
+    cipher_ctx->iv[2] = cipher_ctx->key.counter_nonce[3];
+  } else {
+    cipher_chacha20_poly1305_init_key(cipher_ctx, key, NULL);
+  }
+  return 1;
+}
+
+static int cipher_chacha20_do_cipher(EVP_CIPHER_CTX *ctx, uint8_t *out,
+                                     const uint8_t *inp, size_t len)
```
andrewhop, line 436:
```suggestion
                                     const uint8_t *in, size_t len)
```

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,317 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+
+    // If we consumed all the input, we're done
+    if (len == 0) {
+      return 1;
+    }
+
+    // If we completed a block, increment the counter
+    if (n == CHACHA_BLOCK_LEN) {
+      key->partial_len = 0;
+      key->counter_nonce[0]++;
```
andrewhop, line 460: Is this sufficient? Isn't the counter_nonce 4 u32 long? What happens if 0 overflows?

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,317 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+
+static int cipher_chacha20_do_cipher(EVP_CIPHER_CTX *ctx, uint8_t *out,
+                                     const uint8_t *inp, size_t len)
+{
+  CIPHER_CHACHA_POLY_CTX  *cipher_ctx = CCP_CTX(ctx);
+  CIPHER_CHACHA_KEY *key = CC_KEY(cipher_ctx);
+  uint32_t n, rem, ctr32;
+
+  // Complete any partial block
+  if ((n = key->partial_len)) {
```
andrewhop, line 443: Personally I prefer separating assignment out of any conditional blocks to make it easier to read/reason about.

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,317 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+    // Compute the cipher using our partially used key and any new input up
+    // to the next block
+    while (len && n < CHACHA_BLOCK_LEN) {
+      *out++ = *inp++ ^ key->buf[n++];
```
andrewhop, line 447: Can this be split up, there is a lot going on in this one line.

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,317 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+    // 1<<28 is just a not-so-small yet not-so-large number... Below
+    // condition is practically never met, but it has to be checked for code
+    // correctness.
+    if (sizeof(size_t) > sizeof(uint32_t) && blocks > (1U<<28)) {
+      blocks = (1U << 28);
+    }
+
+    // As ChaCha20_ctr32 operates on 32-bit counter, caller has to handle
+    // overflow. 'if' below detects the overflow, which is then handled by
+    // limiting the amount of blocks to the exact overflow point...
```
andrewhop, line 479: What does the caller need to do? Does this need to change the documentation for any of the EVP_update/encrypt functions? Only got to this point.

crypto/cipher_extra/e_chacha20poly1305.c
andrewhop: I'll send the WIP docs, https://github.com/aws/aws-lc/pull/1318 will add CI for BE and help you test for it. The short answer is most algorithms (e.g. [ChaCha20](https://datatracker.ietf.org/doc/html/rfc7539#section-2.3) takes everything as little) specify the input/output is big/little, and historically all of our code assumed little, but swapping to the correct CRYPTO_load_u32_le will do the correct ([either swap or not](https://github.com/aws/aws-lc/blob/main/crypto/internal.h#L894-L898)) thing on big/little platforms. Without this you're copying bytes into a uint32 in the wrong order.

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,320 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+static int cipher_chacha20_poly1305_do_cipher(
+    EVP_CIPHER_CTX *ctx, unsigned char *out, const unsigned char *inp,
+    size_t inl) {
+  CIPHER_CHACHA_POLY_CTX *cipher_ctx = CCP_CTX(ctx);
+  poly1305_state *poly_ctx = POLY_CTX(cipher_ctx);
+  size_t remainder;
+
+  if (!cipher_ctx->poly_initialized) {
+    // Obtain the poly1305 key by computing the 0th chacha20 key
+    const uint8_t *chacha_key = (const uint8_t *) cipher_ctx->key.key;
```
andrewhop, line 522: This will also hit an endian issue.

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,320 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+    }
+  }
+  return (int32_t) inl;
+}
+
+static void cipher_chacha20_poly1305_cleanup(EVP_CIPHER_CTX *ctx) {
+  CIPHER_CHACHA_POLY_CTX *cipher_ctx = CCP_CTX(ctx);
+  if (cipher_ctx) {
+    OPENSSL_cleanse(ctx->cipher_data, sizeof(CIPHER_CHACHA_POLY_CTX));
+  }
```
andrewhop, line 616: This looks a little weird, you're checking cipher_ctx but freeing ctx->cipher_data.

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,320 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+      cipher_ctx->len.text = 0;
+      cipher_ctx->pad_aad = 0;
+      cipher_ctx->poly_initialized = 0;
+      cipher_ctx->tag_len = 0;
+      return 1;
+    case EVP_CTRL_COPY:
+      if (cipher_ctx && cipher_ctx->poly_initialized) {
+        // The poly1305 context needs to be aligned on a 64-byte boundary.
+        // The destination context doesn't necessarily have the same
+        // alignment so we have to fix that here.
```
andrewhop, line 646: Will this be an issue if the user copies/moves the context? @justsmth what was the issue you hit, was it passing a XTS context by value instead which changed the alignment?

geedo0: This logic exists precisely to support copying the context. I set a flag in the cipher to force EVP to call this CTRL operation during all copy operations. This routine figures out the base pointer of the original poly-context and copies it to an aligned position in the destination context. It does costs us some redundant memory I/O though so I'd be interested in other ways to do this.

crypto/cipher_extra/e_chacha20poly1305.c
geedo0: Good catch. I can fix this, but it's not gonna be pretty.

crypto/cipher_extra/e_chacha20poly1305.c
geedo0: ChaCha20-Poly1305 is defined in RFC 8439 and the limits on the algorithm are defined in [section 2.8]
(https://datatracker.ietf.org/doc/html/rfc8439#section-2.8). They merely state the limits but do not define what should happen if we exceed them. It's up to us to define that. OpenSSL and BoringSSL have chosen to wrap the counter so I'll follow that approach as there's no conventional way of erroring out here.

>1.  The amount of encrypted data possible in a single invocation is
> 2^32-1 blocks of 64 bytes each, because of the size of the block
> counter field in the ChaCha20 block function.  This gives a total
> of 274,877,906,880 bytes, or nearly 256 GB.  This should be
> enough for traffic protocols such as IPsec and TLS, but may be
> too small for file and/or disk encryption.  For such uses, we can
> return to the original design, reduce the nonce to 64 bits, and
> use the integer at position 13 as the top 32 bits of a 64-bit
> block counter, increasing the total message size to over a
> million petabytes (1,180,591,620,717,411,303,360 bytes to be
> exact).
>
>
> 2.  Despite the previous item, the ciphertext length field in the
> construction of the buffer on which Poly1305 runs limits the
> ciphertext (and hence, the plaintext) size to 2^64 bytes, or
> sixteen thousand petabytes (18,446,744,073,709,551,616 bytes to
> be exact).

crypto/cipher_extra/e_chacha20poly1305.c
geedo0: The caller has to update their code to not do >256GB of data in a single ChaCha20-Poly1305 cipher. I can drop a note in the header file.

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,353 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+
+    // Apply padding for the text
+    remainder = cipher_ctx->len.text % POLY1305_TAG_LEN;
+    if (remainder != 0) {
+      CRYPTO_poly1305_update(poly_ctx, padding, sizeof(padding) - remainder);
+    }
+
+    // Apply len values
+    CRYPTO_poly1305_update(poly_ctx, (uint8_t *) &cipher_ctx->len,
+                           POLY1305_TAG_LEN);
```
geedo0, line 629: This looks like the culprit behind our remaining endian issues.

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,361 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+}
+
+static void cipher_chacha20_poly1305_cleanup(EVP_CIPHER_CTX *ctx) {
+  if (ctx->cipher_data) {
+    OPENSSL_cleanse(ctx->cipher_data, sizeof(CIPHER_CHACHA_POLY_CTX));
+  }
+}
+
+// Returns a pointer to the first 64-byte alignment within |poly_ctx|.
+static void *cipher_chacha20_poly1305_aligned_poly(EVP_CIPHER_CTX *ctx) {
```
nebeid, line 661: Should we make this function the same as https://github.com/aws/aws-lc/blob/7e6aef83ecf7bec3a0c6d38e38c64ac079647c78/crypto/poly1305/poly1305.c#L45?

WillChilds-Klein (commented):
The tests added only appear to function in a sort of one-shot mode -- add the AAD, do a single update, finalize. Can we add some tests that exercise multiple updates with ~block-misaligned~ variable sizes?

crypto/cipher_extra/cipher_test.cc
```diff
@@ -538,6 +540,147 @@ TEST(CipherTest, CAVP_TDES_ECB) {
...
+  // Initialize the decrypt context
+  std::unique_ptr<EVP_CIPHER_CTX, decltype(&EVP_CIPHER_CTX_free)> udctx(
+          EVP_CIPHER_CTX_new(), EVP_CIPHER_CTX_free);
+  EVP_CIPHER_CTX *dctx = udctx.get();
+  ASSERT_TRUE(dctx);
+  ASSERT_TRUE(EVP_DecryptInit_ex(dctx, cipher, NULL, key.data(), iv.data()));
+
+  // Insert AAD
+  out_len = 0;
+  ASSERT_TRUE(EVP_DecryptUpdate(dctx, NULL, &out_len, aad.data(), aad.size()));
```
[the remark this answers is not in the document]
WillChilds-Klein: Ahhh, I [RTFM'd](https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption):

> AAD data is passed through in zero or more calls to EVP_EncryptUpdate, with the output buffer set to NULL

include/openssl/cipher.h
```diff
@@ -111,6 +111,13 @@ OPENSSL_EXPORT const EVP_CIPHER *EVP_rc2_cbc(void);
 // deliberately not exported.
 const EVP_CIPHER *EVP_rc2_40_cbc(void);
 
+// EVP_chacha20_poly1305 returns a cipher that implements chacha20-poly1305 as
+// described in RFC 8439. This cipher implementation is added for
```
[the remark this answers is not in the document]
WillChilds-Klein: Nevermind, not needed.

crypto/cipher_extra/e_chacha20poly1305.c
```diff
@@ -346,3 +382,361 @@ const EVP_AEAD *EVP_aead_chacha20_poly1305(void) {
...
+static int cipher_chacha20_poly1305_init(EVP_CIPHER_CTX *ctx,
+                                         const uint8_t *key,
+                                         const uint8_t *iv, int32_t enc) {
+  CIPHER_CHACHA_POLY_CTX *cipher_ctx = CCP_CTX(ctx);
+  cipher_ctx->len.aad = 0;
+  cipher_ctx->len.text = 0;
+  cipher_ctx->pad_aad = 0;
+  cipher_ctx->poly_initialized = 0;
+  if (!key && !iv) {
+    return 1;
```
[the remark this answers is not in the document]
WillChilds-Klein: hm, alright. sold on compatibility with OpenSSL.

crypto/cipher_extra/cipher_test.cc
```diff
@@ -538,6 +540,147 @@ TEST(CipherTest, CAVP_TDES_ECB) {
...
+      ASSERT_TRUE(t->GetBytes(&ciphertext, "CT"));
+      if (EVP_CIPHER_iv_length(cipher) > 0) {
+        ASSERT_TRUE(t->GetBytes(&iv, "NONCE"));
+      }
+      ASSERT_TRUE(EVP_CIPHER_flags(cipher) & EVP_CIPH_FLAG_AEAD_CIPHER);
+      ASSERT_TRUE(t->GetBytes(&aad, "AD"));
+      ASSERT_TRUE(t->GetBytes(&tag, "TAG"));
+
+      Operation op = Operation::kBoth;
+      TestCipher(cipher, op, /*padding=*/false, key, iv, plaintext, ciphertext,
```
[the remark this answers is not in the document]
WillChilds-Klein: confirmed. from [the RFC](https://datatracker.ietf.org/doc/html/rfc7539#section-2.4):

```
                                    There is no requirement for the
   plaintext to be an integral multiple of 512 bits.  If there is extra
   keystream from the last block, it is discarded.  Specific protocols
   MAY require that the plaintext and ciphertext have certain length.
   Such protocols need to specify how the plaintext is padded and how
   much padding it receives.

```

nebeid (approved):

crypto/cipher_extra/e_chacha20poly1305.c
WillChilds-Klein: I read through some of the code to see what our AES CTR impl. does when exposed through the EVP_CIPHER interface. It looks like [the cipher update function](https://github.com/aws/aws-lc/blob/f3e49b01fb9a58bddd5bac218d9369e36d5bfbd0/crypto/fipsmodule/cipher/e_aes.c#L276) has _no error path_ and thus erroring on counter overflow would be impossible for that function (however, other AES impls. [do appear to guard against overflow](https://github.com/aws/aws-lc/blob/f3e49b01fb9a58bddd5bac218d9369e36d5bfbd0/crypto/fipsmodule/cipher/e_aes.c#L1499)).

anyway, since other relevant EVP_CIPHERs don't perform this check and it's not mandated by the RFC, I don't think this is a blocking concern. it might be a useful security feature to be added to all counter-based EVP_CIPHERs in the future.

WillChilds-Klein (approved):

PR #1316, opened 2023-11-20
Allow 0 byte read/write in SSL_{read,write}_ex

WillChilds-Klein:
### Issues:
Addresses CryptoAlg-2136

### Description of changes: 

The primary motivation of this change is to match OpenSSL's behavior for these functions. Relevant links:

- [SSL_read_ex docs][1]
- [SSL_write_ex docs][2]
- [CPython update accounting for this behavior][3]

[1]: https://www.openssl.org/docs/man1.1.1/man3/SSL_read_ex.html
[2]: https://www.openssl.org/docs/man1.1.1/man3/SSL_write_ex.html
[3]: https://github.com/python/cpython/issues/75892#issuecomment-1093761352

### Call-outs:
- n/a

### Testing:
- CI checks
---

### DIFF
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -375,7 +375,7 @@ OPENSSL_EXPORT int SSL_read(SSL *ssl, void *buf, int num);
 // SSL_read_ex reads up to |num| bytes from |ssl| into |buf|. It is similar to
 // |SSL_read|, but instead of returning the number of bytes read, it returns
 // 1 on success or 0 for failure. The number of bytes actually read is stored in
-// |read_bytes|. SSL_read_ex() can be called with num=0, but will not read
+// |read_bytes|. |SSL_read_ex| can be called with |num| set as 0, but will not read
 // application data from the peer.
 
 //
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -375,7 +375,9 @@ OPENSSL_EXPORT int SSL_read(SSL *ssl, void *buf, int num);
 // SSL_read_ex reads up to |num| bytes from |ssl| into |buf|. It is similar to
 // |SSL_read|, but instead of returning the number of bytes read, it returns
 // 1 on success or 0 for failure. The number of bytes actually read is stored in
-// |read_bytes|.
+// |read_bytes|. SSL_read_ex() can be called with num=0, but will not read
+// application data from the peer.
+
 //
 // This is only maintained for OpenSSL compatibility. Use |SSL_read| instead.
 OPENSSL_EXPORT int SSL_read_ex(SSL *ssl, void *buf, size_t num,
@@ -449,7 +451,9 @@ OPENSSL_EXPORT int SSL_write(SSL *ssl, const void *buf, int num);
 // SSL_write_ex writes up to |num| bytes from |buf| into |ssl|. It is similar to
 // |SSL_write|, but instead of returning the number of bytes written, it returns
 // 1 on success or 0 for failure. The number bytes actually written is stored in
-// |written|.
+// |written|. SSL_write_ex() can be called with num=0, but will not send
+// application data to the peer.
+
 //
 // This is only maintained for OpenSSL compatibility. Use |SSL_write| instead.
 OPENSSL_EXPORT int SSL_write_ex(SSL *s, const void *buf, size_t num,
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -451,7 +451,7 @@ OPENSSL_EXPORT int SSL_write(SSL *ssl, const void *buf, int num);
 // SSL_write_ex writes up to |num| bytes from |buf| into |ssl|. It is similar to
 // |SSL_write|, but instead of returning the number of bytes written, it returns
 // 1 on success or 0 for failure. The number bytes actually written is stored in
-// |written|. SSL_write_ex() can be called with num=0, but will not send
+// |written|. |SSL_write_ex| can be called with |num| set as 0, but will not send
 // application data to the peer.
 
 //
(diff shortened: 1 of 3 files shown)


include/openssl/ssl.h
```diff
@@ -449,7 +451,9 @@ OPENSSL_EXPORT int SSL_write(SSL *ssl, const void *buf, int num);
 // SSL_write_ex writes up to |num| bytes from |buf| into |ssl|. It is similar to
 // |SSL_write|, but instead of returning the number of bytes written, it returns
 // 1 on success or 0 for failure. The number bytes actually written is stored in
-// |written|.
+// |written|. SSL_write_ex() can be called with num=0, but will not send
```
darylmartin100, line 454: Should we add similar behavior to SSL_write() while we are doing this? It's not needed but might be good to add in case another application tries to use this method to write zero bytes for testing...

ssl/ssl_lib.cc
```diff
@@ -1026,6 +1026,10 @@ static int ssl_read_impl(SSL *ssl) {
 }
 
 int SSL_read_ex(SSL *ssl, void *buf, size_t num, size_t *read_bytes) {
+  if (num == 0 && read_bytes != nullptr) {
```
darylmartin100, line 1029: is the null pointer needed here?  Should we always set read_bytes to 0 and return 1 when num == 0?

include/openssl/ssl.h
WillChilds-Klein: This runs counter to OpenSSL's [documented behavior](https://www.openssl.org/docs/man1.1.1/man3/SSL_write_ex.html), so if the goal is behavioral compatibility I don't think we should diverge unless we need to.

> You should not call SSL_write() with num=0, it will return an error. SSL_write_ex() can be called with num=0, but will not send application data to the peer.

ssl/ssl_lib.cc
WillChilds-Klein: My goal here was to balance safety with compatibility. from my reading, OpenSSL won't return 1 in that case; it'll segfault. Admittedly, because of the amount of indirection via function pointers in OpenSSL's libssl, I'm not 100% confident in this reading of that code (see below).

So, if indeed my reading is correct and they segfault, I guess the expected behavior of passing null pointer here is "undefined" for OpenSSL. I thought it would be good for us to return an error in that case (as opposed to returning success) because the API's contract is for `*read_bytes` to be set to the number of bytes read on success. Since we can't do that with a null pointer, per the contract, returning success doesn't make sense in my opinion.

---

My reading of OpenSSL's relevant code:

- `SSL_read_ex` calls `ssl_read_internal`, which in turn [calls](https://github.com/openssl/openssl/blob/master/ssl/ssl_lib.c#L2328) the `ssl_read` implementation on `s->method`.
- I believe the implementation for our purposes here (i.e. not DTLS) [is](https://github.com/openssl/openssl/blob/master/ssl/ssl_local.h#L2294) `ssl3_write`, which in turn [calls](https://github.com/openssl/openssl/blob/master/ssl/s3_lib.c#L4510) an `ssl_write_bytes` implementation attached to `s->method`.
- Presumably, that implementation is `ssl3_write_bytes`, which never checks the `written` pointer before [dereferencing](https://github.com/openssl/openssl/blob/master/ssl/record/rec_layer_s3.c#L317) in the case when 0 bytes have been requested to be written (and 0 bytes have been written, i.e. `len == tot == 0` in the code).

Side  note -- somebody summed up the whole situation pretty well in [an internal comment](https://github.com/openssl/openssl/blob/master/ssl/ssl_local.h#L2118):

```
/*
 * This is for the SSLv3/TLSv1.0 differences in crypto/hash stuff It is a bit
 * of a mess of functions, but hell, think of it as an opaque structure :-)
 */
```

include/openssl/ssl.h
darylmartin100: Sounds good!

ssl/ssl_test.cc
```diff
@@ -6510,6 +6510,18 @@ TEST_P(SSLVersionTest, SSLPendingEx) {
   ASSERT_EQ(buf_len, (size_t)2);
   EXPECT_EQ(3, SSL_pending(client_.get()));
   EXPECT_EQ(1, SSL_has_pending(client_.get()));
+
+  // 0-sized IO with valied inputs should succeed but not read/write nor effect
+  // buffer state. However, NULL outlen pointer should fail.
```
samuel40791765, line 6515: Nit:
```suggestion
  // 0-sized IO with valid inputs should succeed but not read/write nor effect
  // buffer state. However, NULL outlen pointer should fail.
```
Also I'm assuming `outlen` refers to `read_bytes/written` here?

include/openssl/ssl.h
```diff
@@ -375,7 +375,9 @@ OPENSSL_EXPORT int SSL_read(SSL *ssl, void *buf, int num);
 // SSL_read_ex reads up to |num| bytes from |ssl| into |buf|. It is similar to
 // |SSL_read|, but instead of returning the number of bytes read, it returns
 // 1 on success or 0 for failure. The number of bytes actually read is stored in
-// |read_bytes|.
+// |read_bytes|. SSL_read_ex() can be called with num=0, but will not read
```
samuel40791765, line 378:
```suggestion
// |read_bytes|. |SSL_read_ex| can be called with |num| set as 0, but will not read
```

include/openssl/ssl.h
samuel40791765, line 454:
```suggestion
// |written|. |SSL_write_ex| can be called with |num| set as 0, but will not send
```

ssl/ssl_test.cc
WillChilds-Klein: yes, i can and should be more precise.

samuel40791765 (approved):

PR #1318, opened 2023-11-21
CI for Build/Testing on PPC64BE

justsmth:
### Issues:
N/A

### Description of changes: 
* Add CI for build and testing of ppc64 (big-endian).

### Call-outs:
N/A

### Testing:
N/A

### DIFF
--- a/crypto/fipsmodule/bn/bytes.c
+++ b/crypto/fipsmodule/bn/bytes.c
@@ -212,6 +212,9 @@ void bn_assert_fits_in_bytes(const BIGNUM *bn, size_t num) {
   size_t tot_bytes = bn->width * sizeof(BN_ULONG);
   if (tot_bytes > num) {
     CONSTTIME_DECLASSIFY(bytes + num, tot_bytes - num);
+// Avoids compiler error: unused variable 'byte' or 'word'
+// The assert statements below are only effective in DEBUG builds
+#ifndef NDEBUG
 #ifdef OPENSSL_BIG_ENDIAN
     for (int i = num / BN_BYTES; i < bn->width; i++) {
       BN_ULONG word = bn->d[i];
@@ -229,6 +232,7 @@ void bn_assert_fits_in_bytes(const BIGNUM *bn, size_t num) {
     for (size_t i = num; i < tot_bytes; i++) {
       assert(bytes[i] == 0);
     }
+#endif
 #endif
     (void)bytes;
   }
--- a/tests/ci/common_posix_setup.sh
+++ b/tests/ci/common_posix_setup.sh
@@ -1,15 +1,19 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0 OR ISC
 
+SRC_ROOT="$(pwd)"
 if [ -v CODEBUILD_SRC_DIR ]; then
   SRC_ROOT="$CODEBUILD_SRC_DIR"
-else
-  SRC_ROOT=$(pwd)
+elif [ "$(basename "${SRC_ROOT}")" != 'aws-lc' ]; then
+  SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+  SRC_ROOT="${SCRIPT_DIR}/../.."
 fi
+SRC_ROOT="$(readlink -f "${SRC_ROOT}")"
 echo "$SRC_ROOT"
 
-cd ../
-SYS_ROOT=$(pwd)
+SYS_ROOT="$(readlink -f "${SRC_ROOT}/..")"
+echo "$SYS_ROOT"
+
 cd $SRC_ROOT
 
 BUILD_ROOT="${SRC_ROOT}/test_build_dir"
--- a/tests/ci/common_posix_setup.sh
+++ b/tests/ci/common_posix_setup.sh
@@ -9,15 +9,15 @@ elif [ "$(basename "${SRC_ROOT}")" != 'aws-lc' ]; then
   SRC_ROOT="${SCRIPT_DIR}/../.."
 fi
 SRC_ROOT="$(readlink -f "${SRC_ROOT}")"
-echo "$SRC_ROOT"
+echo SRC_ROOT="$SRC_ROOT"
 
 SYS_ROOT="$(readlink -f "${SRC_ROOT}/..")"
-echo "$SYS_ROOT"
+echo SYS_ROOT="$SYS_ROOT"
 
 cd $SRC_ROOT
 
 BUILD_ROOT="${SRC_ROOT}/test_build_dir"
-echo "$BUILD_ROOT"
+echo BUILD_ROOT="$BUILD_ROOT"
 
 PLATFORM=$(uname -m)
 
--- a/tests/ci/gtest_util.sh
+++ b/tests/ci/gtest_util.sh
@@ -10,17 +10,22 @@ function shard_gtest() {
         GTEST_TOTAL_SHARDS=4
     fi
 
+    echo shard_gtest-Command: ${1}
     PIDS=()
     COUNTER=0
     while [ $COUNTER -lt $GTEST_TOTAL_SHARDS ]; do
         export GTEST_SHARD_INDEX=$COUNTER
-        echo shard_gtest-Command: ${1}
         ${1} &
         PIDS[${COUNTER}]=$!
         COUNTER=$(( COUNTER+1 ))
     done
 
+    RESULT=0
     for PID in ${PIDS[*]}; do
         wait -f $PID
+        if "${?}" -ne "0"; then
+          RESULT=${?}
+        fi
     done
+    return $RESULT
 }
--- a/tests/ci/gtest_util.sh
+++ b/tests/ci/gtest_util.sh
@@ -1,0 +1,26 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+function shard_gtest() {
+    export GTEST_TOTAL_SHARDS=$(nproc --all)
+    if [ -n "${2}" ]; then
+        GTEST_TOTAL_SHARDS="${2}"
+    fi
+    if [ -z ${GTEST_TOTAL_SHARDS} -o ${GTEST_TOTAL_SHARDS} -lt 1 ]; then
+        GTEST_TOTAL_SHARDS=4
+    fi
+
+    PIDS=()
+    COUNTER=0
+    while [ $COUNTER -lt $GTEST_TOTAL_SHARDS ]; do
+        export GTEST_SHARD_INDEX=$COUNTER
+        echo shard_gtest-Command: ${1}
+        ${1} &
+        PIDS[${COUNTER}]=$!
+        COUNTER=$(( COUNTER+1 ))
+    done
+
+    for PID in ${PIDS[*]}; do
+        wait -f $PID
+    done
+}
--- a/tests/ci/run_cross_ppc64_tests.sh
+++ b/tests/ci/run_cross_ppc64_tests.sh
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+set -ex
+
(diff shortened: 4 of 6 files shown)


tests/ci/run_cross_ppc64_tests.sh
```diff
@@ -0,0 +1,55 @@
...
+echo "Testing AWS-LC shared library for PPC64 big-endian."
+
+run_build -DCMAKE_TOOLCHAIN_FILE="${SCRATCH_FOLDER}/ppc64.cmake"
+
+export QEMU_LD_PREFIX="${SCRATCH_FOLDER}/powerpc64-unknown-linux-gnu/powerpc64-unknown-linux-gnu/sysroot"
+export LD_LIBRARY_PATH="${SCRATCH_FOLDER}/powerpc64-unknown-linux-gnu/powerpc64-unknown-linux-gnu/sysroot/lib"
+
+shard_gtest ${BUILD_ROOT}/crypto/crypto_test
+shard_gtest ${BUILD_ROOT}/ssl/ssl_test
+
```
andrewhop, line 54: We might need a better long term solution (parsing all_tests.json), but for now:
```suggestion
shard_gtest ${BUILD_ROOT}/ssl/urandom_test
shard_gtest ${BUILD_ROOT}/ssl/integration_test
shard_gtest ${BUILD_ROOT}/crypto/mem_test
shard_gtest ${BUILD_ROOT}/crypto/mem_set_test
shard_gtest ${BUILD_ROOT}/crypto/dynamic_loading_test

```

tests/ci/run_cross_ppc64_tests.sh
```diff
@@ -0,0 +1,55 @@
...
+EOF
+
+echo "Testing AWS-LC shared library for PPC64 big-endian."
+
+run_build -DCMAKE_TOOLCHAIN_FILE="${SCRATCH_FOLDER}/ppc64.cmake"
+
+export QEMU_LD_PREFIX="${SCRATCH_FOLDER}/powerpc64-unknown-linux-gnu/powerpc64-unknown-linux-gnu/sysroot"
+export LD_LIBRARY_PATH="${SCRATCH_FOLDER}/powerpc64-unknown-linux-gnu/powerpc64-unknown-linux-gnu/sysroot/lib"
+
+shard_gtest ${BUILD_ROOT}/crypto/crypto_test
```
andrewhop, line 52: We have a few slow tests disabled with valgrind, I'm hoping they're not too slow in qemu:
```suggestion
shard_gtest ${BUILD_ROOT}/crypto/crypto_test --gtest_also_run_disabled_tests
```

tests/ci/run_cross_ppc64_tests.sh
```diff
@@ -0,0 +1,55 @@
...
+set(CMAKE_SYSROOT ${SCRATCH_FOLDER}/powerpc64-unknown-linux-gnu/powerpc64-unknown-linux-gnu/sysroot)
+set(CMAKE_SYSTEM_INCLUDE_PATH ${SCRATCH_FOLDER}/powerpc64-unknown-linux-gnu/powerpc64-unknown-linux-gnu/sysroot/usr/include)
+
+set(ENABLE_EXPERIMENTAL_BIG_ENDIAN_SUPPORT true)
+set(CMAKE_GENERATOR Ninja)
+EOF
+
+echo "Testing AWS-LC shared library for PPC64 big-endian."
+
+run_build -DCMAKE_TOOLCHAIN_FILE="${SCRATCH_FOLDER}/ppc64.cmake"
```
andrewhop, line 47:
```suggestion
run_build -DCMAKE_TOOLCHAIN_FILE="${SCRATCH_FOLDER}/ppc64.cmake" -DCMAKE_BUILD_TYPE=Release
```

tests/ci/run_cross_ppc64_tests.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Specify the sysroot for the target system
+set(CMAKE_SYSROOT ${SCRATCH_FOLDER}/powerpc64-unknown-linux-gnu/powerpc64-unknown-linux-gnu/sysroot)
+set(CMAKE_SYSTEM_INCLUDE_PATH ${SCRATCH_FOLDER}/powerpc64-unknown-linux-gnu/powerpc64-unknown-linux-gnu/sysroot/usr/include)
+
+set(ENABLE_EXPERIMENTAL_BIG_ENDIAN_SUPPORT true)
+set(CMAKE_GENERATOR Ninja)
+EOF
+
+echo "Testing AWS-LC shared library for PPC64 big-endian."
```
andrewhop, line 45: Can you put the below logic in a loop and run the non-FIPS release and FIPS release build in a reasonable amount of time? I'm hoping the release option speeds things up enough that it's not too slow.

tests/ci/run_cross_ppc64_tests.sh
```diff
@@ -0,0 +1,55 @@
...
+
+run_build -DCMAKE_TOOLCHAIN_FILE="${SCRATCH_FOLDER}/ppc64.cmake"
+
+export QEMU_LD_PREFIX="${SCRATCH_FOLDER}/powerpc64-unknown-linux-gnu/powerpc64-unknown-linux-gnu/sysroot"
+export LD_LIBRARY_PATH="${SCRATCH_FOLDER}/powerpc64-unknown-linux-gnu/powerpc64-unknown-linux-gnu/sysroot/lib"
```
andrewhop, line 50: This sets the QEMU paths, but where do you actually run the tests in qemu?

tests/ci/run_cross_ppc64_tests.sh
```diff
@@ -0,0 +1,55 @@
...
+    # Some directories in the archive lack write permission, preventing deletion of files
+    chmod +w -R "${SCRATCH_FOLDER}"
+    rm -rf "${SCRATCH_FOLDER}"
+fi
+mkdir -p "${SCRATCH_FOLDER}"
+
+pushd "${SCRATCH_FOLDER}"
+
+wget https://aws-libcrypto.s3.us-west-2.amazonaws.com/cross-compile-toolchains/host-x86_64-pc-linux-gnu/ppc64-x-tools.tar.xz
+tar Jxvf ppc64-x-tools.tar.xz --no-same-owner --no-same-permissions
```
andrewhop, line 26: Small preference for less scrolling, any errors should still get printed:
```suggestion
tar Jxf ppc64-x-tools.tar.xz --no-same-owner --no-same-permissions
```

tests/ci/run_cross_ppc64_tests.sh
```diff
@@ -0,0 +1,55 @@
...
+if [ -e "${SCRATCH_FOLDER}" ]; then
+    # Some directories in the archive lack write permission, preventing deletion of files
+    chmod +w -R "${SCRATCH_FOLDER}"
+    rm -rf "${SCRATCH_FOLDER}"
+fi
+mkdir -p "${SCRATCH_FOLDER}"
+
+pushd "${SCRATCH_FOLDER}"
+
+wget https://aws-libcrypto.s3.us-west-2.amazonaws.com/cross-compile-toolchains/host-x86_64-pc-linux-gnu/ppc64-x-tools.tar.xz
```
andrewhop, line 25: That's a lot of scrolling
```suggestion
wget -q https://aws-libcrypto.s3.us-west-2.amazonaws.com/cross-compile-toolchains/host-x86_64-pc-linux-gnu/ppc64-x-tools.tar.xz
```

tests/ci/common_posix_setup.sh
```diff
@@ -1,15 +1,19 @@
...
+  SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+  SRC_ROOT="${SCRIPT_DIR}/../.."
 fi
+SRC_ROOT="$(readlink -f "${SRC_ROOT}")"
 echo "$SRC_ROOT"
 
-cd ../
-SYS_ROOT=$(pwd)
+SYS_ROOT="$(readlink -f "${SRC_ROOT}/..")"
+echo "$SYS_ROOT"
```
WillChilds-Klein, line 15: nit: here and L12, should we echo the variable name as well? otherwise it's just some paths outputted to stdout...


```
echo "SYS_ROOT $SYS_ROOT"
```

tests/ci/gtest_util.sh
```diff
@@ -0,0 +1,26 @@
...
+    fi
+    if [ -z ${GTEST_TOTAL_SHARDS} -o ${GTEST_TOTAL_SHARDS} -lt 1 ]; then
+        GTEST_TOTAL_SHARDS=4
+    fi
+
+    PIDS=()
+    COUNTER=0
+    while [ $COUNTER -lt $GTEST_TOTAL_SHARDS ]; do
+        export GTEST_SHARD_INDEX=$COUNTER
+        echo shard_gtest-Command: ${1}
```
WillChilds-Klein, line 17: nit: this is invariant over each shard; should we just output it once before the loop?

tests/ci/gtest_util.sh
```diff
@@ -0,0 +1,26 @@
...
+        ${1} &
+        PIDS[${COUNTER}]=$!
+        COUNTER=$(( COUNTER+1 ))
+    done
+
+    for PID in ${PIDS[*]}; do
+        wait -f $PID
```
WillChilds-Klein, line 24: will this correctly propagate each process's exit code? we wouldn't want to let an individual shard silently fail in our CI...

tests/ci/run_cross_ppc64_tests.sh
```diff
@@ -0,0 +1,55 @@
...
+
+echo "Testing AWS-LC shared library for PPC64 big-endian."
+
+run_build -DCMAKE_TOOLCHAIN_FILE="${SCRATCH_FOLDER}/ppc64.cmake"
+
+export QEMU_LD_PREFIX="${SCRATCH_FOLDER}/powerpc64-unknown-linux-gnu/powerpc64-unknown-linux-gnu/sysroot"
+export LD_LIBRARY_PATH="${SCRATCH_FOLDER}/powerpc64-unknown-linux-gnu/powerpc64-unknown-linux-gnu/sysroot/lib"
+
+shard_gtest ${BUILD_ROOT}/crypto/crypto_test
+shard_gtest ${BUILD_ROOT}/ssl/ssl_test
```
WillChilds-Klein, line 53: What's the value in running the same test suites each `$(nproc)` times in parallel with identical parameters? Isn't that redundant?

tests/ci/run_cross_ppc64_tests.sh
justsmth: Yeah, the tests are actually being run with Qemu. This setup requires [`qemu-user-binfmt`](https://packages.ubuntu.com/jammy/qemu-user-binfmt) to be installed, which registers the Qemu user static binaries as handlers for various [cpu binary formats](https://www.man7.org/linux/man-pages/man5/binfmt.d.5.html) including ppc64be.

tests/ci/gtest_util.sh
justsmth: I was assuming that our scripts all use `set -e`, but that's not a safe assumption.  I'll modify this to explicitly check the exit status of the child and return an appropriate value.

tests/ci/run_cross_ppc64_tests.sh
justsmth: By exporting [`GTEST_TOTAL_SHARDS` and `GTEST_SHARD_INDEX`](https://google.github.io/googletest/advanced.html#distributing-test-functions-to-multiple-machines) from the environment, it informs the GoogleTest to shard so that each test only runs on one of the shards. Each process/shard has a distinct `GTEST_SHARD_INDEX` value so the tests it runs are unique.

crypto/fipsmodule/bn/bytes.c
```diff
@@ -212,6 +212,9 @@ void bn_assert_fits_in_bytes(const BIGNUM *bn, size_t num) {
   size_t tot_bytes = bn->width * sizeof(BN_ULONG);
   if (tot_bytes > num) {
     CONSTTIME_DECLASSIFY(bytes + num, tot_bytes - num);
+// Avoids compiler error: unused variable 'byte' or 'word'
+// The assert statements below are only effective in DEBUG builds
```
andrewhop, line 216: Huh, why doesn't this hit the little endian path for release mode?

andrewhop (approved):

crypto/fipsmodule/bn/bytes.c
justsmth: Only in release mode (when the assert statements are no-ops) do the values go unused. I didn't catch this due to only running debug builds locally.

WillChilds-Klein (approved):

PR #1320, opened 2023-11-22
Increment session hit counter for ticket resumptions

WillChilds-Klein:
### Issues:
Addresses CryptoAlg-2136

### Description of changes: 
OpenSSL increments session hits in two places:

1. Handshake finish [on the client side](https://github.com/openssl/openssl/blob/1750689767cc922bdbe73358f7256475f0838c67/ssl/statem/statem_lib.c#L1494)
2. Fetching potentially pre-existing session [on the server side](https://github.com/openssl/openssl/blob/1750689767cc922bdbe73358f7256475f0838c67/ssl/ssl_sess.c#L665)

Prior to this change, we failed to increment the session "cache" hit counter for TLS 1.3 servers. I put "cache" in scare-quotes because whether that behavior is correct depends on how you define a cache, as TLS 1.3 ticket resumption is stateless and doesn't require a true session cache. Semantics aside, the change in this PR conforms to OpenSSL's behavior. OpenSSL [documents](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_sess_hits.html) session hit counts as:

> SSL_CTX_sess_hits() returns the number of successfully reused sessions. In client mode a session set with [SSL_set_session(3)](https://www.openssl.org/docs/man1.1.1/man3/SSL_set_session.html) successfully reused is counted as a hit. In server mode a session successfully retrieved from internal or external cache is counted as a hit.

### Call-outs:
- n/a

### Testing:
- CI tests
- newly added test assertions failed before source change, succeed after
- built and tested [my python 3.10 fork](https://github.com/WillChilds-Klein/cpython/tree/3.10-debugging) against this change

---

### DIFF
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -5012,7 +5012,8 @@ OPENSSL_EXPORT int SSL_CTX_sess_accept_renegotiate(const SSL_CTX *ctx);
 // SSL/TLS sessions in server mode.
 OPENSSL_EXPORT int SSL_CTX_sess_accept_good(const SSL_CTX *ctx);
 
-// SSL_CTX_sess_hits returns the number of successfully reused sessions.
+// SSL_CTX_sess_hits returns the number of successfully reused sessions, from
+// both session cache and session tickets.
 OPENSSL_EXPORT int SSL_CTX_sess_hits(const SSL_CTX *ctx);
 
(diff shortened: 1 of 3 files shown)


andrewhop (commented):
Can you update the comment in ssl.h to note this behavior.

samuel40791765 (approved):

skmcgrail (approved):

PR #1322, opened 2023-11-27
Return 0 if using default proto min/max

WillChilds-Klein:
### Issues:
Addresses CryptoAlg-2136

### Description of changes: 

While both AWS-LC/BoringSSL and OpenSSL allow `0` to be used in `SSL_CTX_set_min_proto_version` and `SSL_CTX_set_max_proto_version` to configure an `SSL_CTX` to use default TLS versions for its TLS "method", the two libraries differ in the behavior of their corresponding getters. When configured to use default versions, OpenSSL returns `0`:

- [OpenSSL 1.1.1 docs](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_min_proto_version.html)
- [OpenSSL 3 docs](https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_min_proto_version.html)
- relevant code pointers [here](https://github.com/openssl/openssl/blob/b0e9d0370262ade64c55f2385fbb09ec6aa81e76/ssl/ssl_lib.c#L3894) and [here](https://github.com/openssl/openssl/blob/master/ssl/statem/statem_lib.c#L2055-L2058)

AWS-LC/BoringSSL, on the other hand, will return [the actual default value being used](https://github.com/aws/aws-lc/blob/main/include/openssl/ssl.h#L736-L740) instead of the `0` sentinel value. This PR modifies AWS-LC to conform with OpenSSL's behavior in this respect. I don't think OpenSSL's choice is generally preferable (how can applications tell which version will actually be used by default?), but this change will allow AWS-LC greater compatibility with applications that rely on AWS-LC's behavior.

### Call-outs:
- n/a

### Testing:
- CI checks
---

### DIFF
--- a/ssl/internal.h
+++ b/ssl/internal.h
@@ -3660,6 +3660,20 @@ struct ssl_ctx_st {
   // and is further constrainted by |SSL_OP_NO_*|.
   uint16_t conf_min_version = 0;
 
+  // conf_max_version_use_default indicates whether the |SSL_CTX| is configured
+  // to use the default maximum protocol version for the relevant protocol
+  // method. By default, |SSL_CTX_new| will set this to 0 and use that default.
+  // callers can change this by calling |SSL_CTX_set_max_proto_version| with a
+  // non-zero value.
+  bool conf_max_version_use_default;
+
+  // conf_max_version_use_default indicates whether the |SSL_CTX| is configured
+  // to use the default maximum protocol version for the relevant protocol
+  // method. By default, |SSL_CTX_new| will set this to 0 and use that default.
+  // callers can change this by calling |SSL_CTX_set_max_proto_version| with a
+  // non-zero value.
+  bool conf_min_version_use_default;
+
   // num_tickets is the number of tickets to send immediately after the TLS 1.3
   // handshake. TLS 1.3 recommends single-use tickets so, by default, issue two
   /// in case the client makes several connections before getting a renewal.
--- a/ssl/internal.h
+++ b/ssl/internal.h
@@ -3303,6 +3303,21 @@ struct SSL_CONFIG {
   // of support for AES hw. The value is only considered if |aes_hw_override| is
   // true.
   bool aes_hw_override_value : 1;
+
+
+  // conf_max_version_use_default indicates whether the |SSL_CTX| is configured
+  // to use the default maximum protocol version for the relevant protocol
+  // method. By default, |SSL_CTX_new| will set this to 0 and use that default.
+  // callers can change this by calling |SSL_CTX_set_max_proto_version| with a
+  // non-zero value.
+  bool conf_max_version_use_default;
+
+  // conf_max_version_use_default indicates whether the |SSL_CTX| is configured
+  // to use the default maximum protocol version for the relevant protocol
+  // method. By default, |SSL_CTX_new| will set this to 0 and use that default.
+  // callers can change this by calling |SSL_CTX_set_max_proto_version| with a
+  // non-zero value.
+  bool conf_min_version_use_default;
 };
 
 // From RFC 8446, used in determining PSK modes.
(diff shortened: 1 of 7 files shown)


torben-hansen (commented):
Please check whether this brings `SSL_set_{max,min}_proto_version` out-of-sync with the corresponding context versions of those functions. The SSL objets inherit config data from the SSL context object.

Skimming, something looks a bit unaligned now.

Maybe just leave as-is if nothing is broken from this?

WillChilds-Klein:
I've ported this behavior to `SSL_get_{min,max}_proto_version` and have updated the relevant ASN.1 parsing code and tests. The ASN.1 structure for `SSL_CONFIG` is documented as "experimental"; curious to hear others' opinions on what that means for backwards (in)compatibility.

ssl/ssl_versions.cc
```diff
@@ -335,44 +335,54 @@ BSSL_NAMESPACE_END
 using namespace bssl;
 
 int SSL_CTX_set_min_proto_version(SSL_CTX *ctx, uint16_t version) {
+  ctx->conf_min_version_use_default = version == 0;
```
justsmth, line 338: NP: This (and the others like it) would be easier to visually parse as:
```
ctx->conf_min_version_use_default = (version == 0);
```

justsmth (approved):

ssl/ssl_versions.cc
WillChilds-Klein: sure, i'll do that.

ssl/ssl_transfer_asn1.cc
```diff
@@ -988,12 +1007,20 @@ static int SSL_CONFIG_from_bytes(SSL_CONFIG *out, CBS *cbs) {
...
+      !CBS_get_optional_asn1_bool(&config, &conf_min_version_use_default,
+                                  kSSLConfigConfMinVersionUseDefault,
+                                  1 /* default to true */) ||
       CBS_len(&config) != 0) {
     OPENSSL_PUT_ERROR(SSL, SSL_R_SERIALIZATION_INVALID_SSL_CONFIG);
     return 0;
   }
   out->conf_max_version = conf_max_version;
   out->conf_min_version = conf_min_version;
+  out->conf_max_version_use_default = !!conf_max_version_use_default;
```
torben-hansen, line 1022: a'ha I had forgotten about the `!!` trick

ssl/ssl_transfer_asn1.cc
```diff
@@ -930,6 +930,10 @@ static const unsigned kSSLConfigOcspStaplingEnabledTag =
 static const unsigned kSSLConfigJdk11WorkaroundTag =
     CBS_ASN1_CONSTRUCTED | CBS_ASN1_CONTEXT_SPECIFIC | 1;
 
+static const unsigned kSSLConfigConfMaxVersionUseDefault = 2;
```
torben-hansen, line 933: Default tagging is `EXPLICIT` per https://github.com/aws/aws-lc/blob/main/ssl/tls_transfer.asn#L1. Hence this needs `CBS_ASN1_CONSTRUCTED` and `CBS_ASN1_CONTEXT_SPECIFIC`.

ssl/ssl_transfer_asn1.cc
```diff
@@ -944,7 +948,9 @@ static const unsigned kSSLConfigJdk11WorkaroundTag =
 //    version                           INTEGER (1),  -- SSL_CONFIG structure
```
torben-hansen, line 948: Version needs a bump

ssl/internal.h
```diff
@@ -3303,6 +3303,20 @@ struct SSL_CONFIG {
   // of support for AES hw. The value is only considered if |aes_hw_override| is
   // true.
   bool aes_hw_override_value : 1;
+
+  // conf_max_version_use_default indicates whether the |SSL_CONFIG| is configured
+  // to use the default maximum protocol version for the relevant protocol
+  // method. By default, |SSL_new| will set this to 0 and use that default.
```
torben-hansen, line 3309: Won't it inherit from `SSL_CTX` where e.g. max version might not be 0? See https://github.com/aws/aws-lc/blob/main/ssl/ssl_lib.cc#L651-L652

Same for `conf_min_version_use_default`.

justsmth (approved):

torben-hansen (commented):
For the updated transfer format, update documentation: https://github.com/aws/aws-lc/blob/main/ssl/tls_transfer.asn#L122-L130

ssl/internal.h
WillChilds-Klein: Hm, this comment is poorly worded. Let me see if I can clarify.

torben-hansen (approved):

justsmth (approved):

PR #1323, opened 2023-11-27
Improve decision logic for s2n-bignum implementation

torben-hansen:
### Description of changes: 

Follow-up to https://github.com/aws/aws-lc/pull/1309. Simplifies the decision logic for which s2n-bignum implementation that should be used at run-time. Basically, before the overall ISA type decision logic was repeated twice.

Suggested by @dkostic.

### Call-outs:

The aborts before calling s2n-bignum is actually not important, so removed it. The reason is that we already assume that the fallback s2n-bignum implementation works on all available cpu (for the specific type).

### Testing:

Manually inspected the same implementation is chosen as before. Also ran a couple of speed experiments to see that the same speed was reported.

### DIFF
--- a/crypto/curve25519/curve25519.c
+++ b/crypto/curve25519/curve25519.c
@@ -64,23 +64,6 @@
 #define CURVE25519_S2N_BIGNUM_CAPABLE
 #endif
 
-OPENSSL_INLINE int curve25519_s2n_bignum_capable(void) {
-#if defined(CURVE25519_S2N_BIGNUM_CAPABLE)
-  return 1;
-#else
-  return 0;
-#endif
-}
-
-// Temporarily use separate function for Ed25519. See CryptoAlg-2198.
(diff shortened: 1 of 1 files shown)


crypto/curve25519/curve25519.c
```diff
@@ -184,6 +134,30 @@ OPENSSL_INLINE int curve25519_s2n_bignum_no_alt_capable(void) {
...
+  }
+#endif
+  // Have to return some default value.
+  return 0;
+}
+
+
+// s2n-bignum wrappers
 
 static void x25519_s2n_bignum(uint8_t out_shared_key[32],
```
nebeid, line 162: Not related to this review, can this function's name be changed to describe what it does?

torben-hansen: will change in a follow-up I have planned (moving s2n-bignum wrappers to separate file.

nebeid (approved):

PR #1325, opened 2023-11-27
Use mkstemp instead of tmpname when available

samuel40791765:
### Description of changes: 
Despite silencing tmpname deprecation warnings, the warning still appears on Linux platforms when compiling tests.

### DIFF
--- a/ssl/ssl_test.cc
+++ b/ssl/ssl_test.cc
@@ -11850,14 +11850,10 @@ TEST(SSLTest, SSLFileTests) {
   using ScopedFILE = std::unique_ptr<FILE, fclose_deleter>;
 
 #if defined(OPENSSL_WINDOWS)
-  OPENSSL_BEGIN_ALLOW_DEPRECATED
-  // |tmpnam| is deprecated due to security concerns, but we only use this to
-  // run tests.
   char rsa_pem_filename[L_tmpnam];
   char ecdsa_pem_filename[L_tmpnam];
-  ASSERT_TRUE(tmpnam(rsa_pem_filename));
-  ASSERT_TRUE(tmpnam(ecdsa_pem_filename));
-  OPENSSL_END_ALLOW_DEPRECATED
+  ASSERT_EQ(tmpnam_s(rsa_pem_filename, sizeof(rsa_pem_filename)), 0);
+  ASSERT_EQ(tmpnam_s(ecdsa_pem_filename, sizeof(ecdsa_pem_filename)), 0);
 #else
   char rsa_pem_filename[] = "/tmp/fileXXXXXX";
   char ecdsa_pem_filename[] = "/tmp/fileXXXXXX";
--- a/ssl/ssl_test.cc
+++ b/ssl/ssl_test.cc
@@ -11836,10 +11836,6 @@ TEST_P(PerformHybridHandshakeTest, PerformHybridHandshake) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


ssl/ssl_test.cc
```diff
@@ -11852,10 +11848,22 @@ TEST(SSLTest, SSLFileTests) {
...
+  OPENSSL_END_ALLOW_DEPRECATED
+#else
+  char rsa_pem_filename[] = "/tmp/fileXXXXXX";
+  char ecdsa_pem_filename[] = "/tmp/fileXXXXXX";
+  ASSERT_TRUE(mkstemp(rsa_pem_filename));
+  ASSERT_TRUE(mkstemp(ecdsa_pem_filename));
+#endif
 
   ScopedFILE rsa_pem(fopen(rsa_pem_filename, "w"));
   ScopedFILE ecdsa_pem(fopen(ecdsa_pem_filename, "w"));
```
WillChilds-Klein, line 11869: If we don't care about the file name, can't you just use  `tmpfile()` [as in bio_test.cc](https://github.com/aws/aws-lc/blob/main/crypto/bio/bio_test.cc#L176)? It should be [portable enough](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/tmpfile?view=msvc-170). These lines would just become:

```
ScopedFile rsa_pem(tmpfile());
ScopedFile ecdsa_pem(tmpfile());
```

justsmth:
Another option is `tempnam` (a [POSIX.1-2008 alternative](https://pubs.opengroup.org/onlinepubs/9699919799/functions/tempnam.html) that's thread-safe?) that appears to be supported on [Windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/tempnam?view=msvc-160), [Linux](https://www.man7.org/linux/man-pages/man3/tempnam.3.html) and [Mac](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/tempnam.3.html)

justsmth (approved):

ssl/ssl_test.cc
samuel40791765: That's the conundrum I've been trying to solve with these tests :( 
`SSL_CTX_use_certificate_chain_file` and `SSL_use_certificate_chain_file` directly take the file name as the second argument and parse the certificates for the user. `tmpfile` was the initial plan, but it doesn't provide an interface to extract the created file name.

samuel40791765 (commented):
> Another option is tempnam (a [POSIX.1-2008 alternative](https://pubs.opengroup.org/onlinepubs/9699919799/functions/tempnam.html) that's thread-safe?) that appears to be supported on [Windows](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/tempnam?view=msvc-160), [Linux](https://www.man7.org/linux/man-pages/man3/tempnam.3.html) and [Mac](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/tempnam.3.html)

Sure, `tempname` has the same security and deprecation warning as `tmpnam`, but we can opt for the additional tread-safety on Windows.

justsmth (approved):

ssl/ssl_test.cc
WillChilds-Klein: Ah, I see. That makes sense.

WillChilds-Klein (approved):

PR #1326, opened 2023-11-28
Add null-check for input args in SSL_{read|write}_ex

dkostic:
### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 
Add null-check for input args in SSL_{read|write}_ex to avoid potentially dereferencing nullptr.

### Call-outs:

### Testing:

### DIFF
--- a/ssl/ssl_lib.cc
+++ b/ssl/ssl_lib.cc
@@ -1026,7 +1026,10 @@ static int ssl_read_impl(SSL *ssl) {
 }
 
 int SSL_read_ex(SSL *ssl, void *buf, size_t num, size_t *read_bytes) {
-  if (num == 0 && read_bytes != nullptr) {
+  if (read_bytes == nullptr) {
+    return 0;
+  }
+  if (num == 0) {
     *read_bytes = 0;
     return 1;
   }
@@ -1123,7 +1126,10 @@ int SSL_write(SSL *ssl, const void *buf, int num) {
 }
 
 int SSL_write_ex(SSL *ssl, const void *buf, size_t num, size_t *written) {
-  if (num == 0 && written != nullptr) {
+  if (written == nullptr) {
+    return 0;
+  }
+  if (num == 0) {
     *written = 0;
     return 1;
   }
--- a/ssl/ssl_lib.cc
+++ b/ssl/ssl_lib.cc
@@ -1026,10 +1026,7 @@ static int ssl_read_impl(SSL *ssl) {
 }
 
 int SSL_read_ex(SSL *ssl, void *buf, size_t num, size_t *read_bytes) {
-  if (read_bytes == nullptr) {
-    return 0;
-  }
-  if (num == 0) {
+  if (num == 0 && read_bytes != nullptr) {
     *read_bytes = 0;
     return 1;
   }
@@ -1037,7 +1034,9 @@ int SSL_read_ex(SSL *ssl, void *buf, size_t num, size_t *read_bytes) {
   if (ret <= 0) {
     return 0;
   }
-  *read_bytes = ret;
+  if (read_bytes != nullptr) {
+    *read_bytes = ret;
+  }
   return 1;
 }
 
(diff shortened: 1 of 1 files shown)


samuel40791765 (approved):

ssl/ssl_lib.cc
```diff
@@ -1026,7 +1026,10 @@ static int ssl_read_impl(SSL *ssl) {
 }
 
 int SSL_read_ex(SSL *ssl, void *buf, size_t num, size_t *read_bytes) {
-  if (num == 0 && read_bytes != nullptr) {
+  if (read_bytes == nullptr) {
+    return 0;
+  }
+  if (num == 0) {
```
torben-hansen, line 1032: is this actually what we want? Aren't there some behaviour and state stuff that happens when you did `num = 0` before that would now never happen now that it short-circuits

WillChilds-Klein: I outlined my reasoning for the prior change [here](https://github.com/aws/aws-lc/pull/1316#discussion_r1400823717), but I suppose I missed a case (which, admittedly, would be pretty silly of the caller to do) where they specify `num > 1` and `read_bytes = NULL`. They would've gotten a segfault for those inputs before my change, as they would still before my change. But not after this change.

WillChilds-Klein (commented):
Thanks dusan. Could we also add two test cases [here](https://github.com/aws/aws-lc/blob/main/ssl/ssl_test.cc#L6525)? Perhaps something like

```
  ASSERT_EQ(0, SSL_read_ex(client_.get(), (void *)"A", 1, nullptr));
  ASSERT_EQ(0, SSL_write_ex(client_.get(), (void *)"A", 1, nullptr));
```

ssl/ssl_lib.cc
torben-hansen: Not sure I see why we needed to change behaviour in AWS-LC and short-circuit, from the linked comment.

WillChilds-Klein: Ah, sorry, yeah that was a bit confusing. A few things to tease apart:

1. our libssl's behavior [prior](https://github.com/aws/aws-lc/blob/77aa3bd9ae71f5c6d483396631b2defd59a9d69a/ssl/ssl_lib.cc#L1033) to my change merged earlier this week
2. our libssl's behavior after [my change](https://github.com/aws/aws-lc/commit/89e1fd0d93360fee0b0ed54cd534ca44f6965367) but before dusan's change in this PR
3. OpenSSL's behavior as described in [that comment](https://github.com/aws/aws-lc/pull/1316/files#r1400823717)

Per my reading, 1. would segfault whenever `read_bytes=nullptr` and the internal call to `SSL_read` is successful, regardless of the value of `num`. Both 1. and 2. would segfault when `read_bytes=nullptr`,  the internal call to `SSL_read` is successful, and `num > 1`. In both of those cases, there's nothing being short-circuited, as `read_bytes` isn't passed anywhere -- the internal call to `SSL_read` doesn't use it. After dusan's change, no chance of segfault due to `read_bytes=nullptr`.

For 3., I'm a little bit more shaky  on my reading of OpenSSL's code because of all the indirection, but I don't think we're doing anything here that would be "incompatible" with OpenSSL (unless you count preventing a segfault, which I'd argue is a universally good thing).

WillChilds-Klein: @torben-hansen and I did a closer reading of the code offline. The "short-circuit" he was referring to was irrelevant to `read_bytes` and the potential segfaults addressed above. His concern was that by returning early in `SSL_read_ex` when `num=0`, we're skipping some underlying logic that would otherwise be executed. We confirmed this. Specifically, `SSL_read_ex` would call down into `SSL_read` which calls `SSL_peek` which finally calls `ssl_read_impl`. Down in `ssl_read_impl`, some [handshake completion](https://github.com/aws/aws-lc/blob/main/ssl/ssl_lib.cc#L978) and even [reading/decrypting](https://github.com/aws/aws-lc/blob/main/ssl/ssl_lib.cc#L1013-L1014) might be performed (the results of the latter being thrown away). I argue this is acceptable, perhaps even preferable for performance reasons, and that any handshake completion or other setup logic would be performed on subsequent IO requests for `num > 0`.

Torben also dug up [this artifact of history](https://bugs.chromium.org/p/boringssl/issues/detail?id=24), which may be relevant.

dkostic:
@torben-hansen @WillChilds-Klein 
I decided to only fix the potential null pointer dereference in this commit without changing the behavior of the functions. If you want to actually change the behavior we can do it separately.

torben-hansen (approved):

WillChilds-Klein (approved):

PR #1333, opened 2023-12-01
Implement SSL_MODE_AUTO_RETRY

WillChilds-Klein:
### Issues:

Addresses CryptoAlg-2136

### Description of changes: 

# Notes

This change implements `SSL_MODE_AUTO_RETRY`, which was previously ignored. When this option (off by default) is set, `SSL_get_error`s behavior hews a little more closely to OpenSSL when processing a return code of 0 (i.e. empty I/O). With this option set, we allow error processing for empty reads if we're in a retryable state. With this option unset, we treat (most) 0-valued return codes as `SSL_ERROR_SYSCALL`, using it as a sort-of alias for EOF in the underlying transport. Internal calls to `SSL_get_error` (such as [those in `bio_ssl.cc`](https://github.com/aws/aws-lc/blob/main/ssl/bio_ssl.cc#L29)) then observe the  `SSL_ERROR_SYSCALL` and fail to process potential retries.

OpenSSL implements this mode differently, only checking its value in its [main read loop](https://github.com/openssl/openssl/blob/398011848468c7e8e481b295f7904afc30934217/ssl/record/rec_layer_s3.c#L992). Neither OpenSSL [1.1.1][6] nor [OpenSSL 3.x][10] return `SSL_ERROR_SYSCALL` early for empty reads, instead allowing them to be processed for retryable state. BoringSSL [appears to have diverged][7] from this behavior in 2015, but that commit was just a revert of a regression committed ~1 month prior. Their [previous implementation][18] closely resembles [that of OpenSSL 1.0.2][13], guarding retryable error processing with `i < 0` checks. `SSL_get_error` in [OpenSSL 1.1.1][12] and [OpenSSL 3.0][11], however, have no such guards. This change resembles the latter two. 

Why is it important that we process 0-valued return codes more fully? The recommended SSL IO functions (`SSL_read`, `SSL_write`, etc.) are, by contract, allowed to return negative numbers to indicate an error or non-success condition. However, their `size_t`-clean counterparts (`SSL_read_ex`, `SSL_write_ex`, etc.) must, by contract, only return 0 or 1. This means that passing the return code from the various `_ex` functions to `SSL_get_error` loses error reporting information, resulting in inconsistent handling of otherwise identical error cases between the two function families.

Another solution to the `_ex` functions' loss of error reporting information would be to change the contract of those functions to allow negative return values, directly returning the return code from internal calls to `SSL_read` or `SSL_write`. I decided not to take this approach as it would break callers who do not account for negative return values in their error handling.

# Relevant Links

- [OpenSSL 1.0.2 documentation for `SSL_get_error`][3]
- [OpenSSL 1.1.1 documentation for `SSL_get_error`][4]
- [OpenSSL 3.0 documentation for `SSL_get_error`][5]
- [OpenSSL 1.0.2 implementation for `SSL_get_error`][13]
- [OpenSSL 1.1.1 implementation for `SSL_get_error`][12]
- [OpenSSL 3.0 implementation for `SSL_get_error`][11]
- [OpenSSL master (3.2) implementation for `SSL_get_error`][10]
- [OpenSSL 1.1.1 commit dropping `ret < 0` guards for hint processing][6]
- [Unresolved OpenSSL issue for distinguishing EOF from failure in `BIO_read_ex`][1]
- [BoringSSL commit introducing `SSL_ERROR_SYSCALL` for `ret == 0` case][7]
- [BoringSSL Issue 503: Stop preserving arbitrary BIO error return values][2]
- [BoringSSL Issue 24: Sanitize SSL and BIO return codes.][15]
- [OpenSSL 3.0 commit returning `SSL_ERROR_SSL`, add reason to err stack on EOF][14]
- [OpenSSL 3.0 commit introducing `SSL_OP_IGNORE_UNEXPECTED_EOF`  option][8]
- [CPython commit from @davidben removing CPython's use of `SSL_OP_IGNORE_UNEXPECTED_EOF`][9]
- [CPython commit switching from `SSL_read` and `SSL_write` to `SSL_read_ex` and `SSL_write_ex`][16]
- [OpenSSL 1.1.1 documentation for `SSL_MODE_AUTO_RETRY`][17]

[1]: https://github.com/openssl/openssl/issues/8208
[2]: https://bugs.chromium.org/p/boringssl/issues/detail?id=503
[3]: https://www.openssl.org/docs/man1.0.2/man3/SSL_get_error.html
[4]: https://www.openssl.org/docs/man1.1.1/man3/SSL_get_error.html
[5]: https://www.openssl.org/docs/man3.0/man3/SSL_get_error.html
[6]: https://github.com/openssl/openssl/commit/8051ab2b6f8e1fb9e957771afcc3555560f9694f
[7]: https://github.com/google/boringssl/commit/9a38e924aafa96b4f1d7495ce140ba06b956854f
[8]: https://github.com/openssl/openssl/commit/09b90e0ed7915809fcd4ee1e250d881b77d06d45
[9]: https://github.com/python/cpython/pull/95495
[10]: https://github.com/openssl/openssl/blob/master/ssl/ssl_lib.c#L4601
[11]: https://github.com/openssl/openssl/blob/openssl-3.0/ssl/ssl_lib.c#L3839
[12]: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/ssl/ssl_lib.c#L3617
[13]: https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/ssl/ssl_lib.c#L2709
[14]: https://github.com/openssl/openssl/commit/d924dbf4ae127c68463bcbece04b6e06abc58928
[15]: https://bugs.chromium.org/p/boringssl/issues/detail?id=24
[16]: https://github.com/python/cpython/commit/89d1550d14ba689af12eeb726e4ff8ce73cee7e1
[17]: https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_mode.html
[18]: https://github.com/google/boringssl/commit/fcf25833bc4d057377da1ea755f57b2f5126e555#diff-38e1eecb6013c2fe8bf194e5020ce1afd6e31c371d7849e041464ad822fe745bL2271

### Call-outs:
- n/a

### Testing:
- CI tests
- PR #1359 runs relevant CPython module tests against this change


samuel40791765:
This seemed relevant to the test failures I'm running into with Bind9, so I tried applying the change to see if it would resolve my issues. It results in the tests from Bind9 hanging forever though :(. 
I'm pasting this code block from bind9 here, it seems they were dependent on detecting the original behaviour:
* https://gitlab.isc.org/isc-projects/bind9/-/blob/4330014fce5ab2ac1a6e58cd330762318a34f8d9/lib/isc/netmgr/tlsstream.c#L731-750

ssl/ssl_test.cc
```diff
@@ -10458,6 +10458,57 @@ TEST(SSLTest, ErrorSyscallAfterCloseNotify) {
...
+  ASSERT_TRUE(ConnectClientAndServer(&client, &server, client_ctx.get(),
+                                     server_ctx.get()));
+
+  // Create a fake read BIO that returns 0 on read to simulate empty read
+  bssl::UniquePtr<BIO_METHOD> method(BIO_meth_new(0, nullptr));
+  ASSERT_TRUE(method);
+  BIO_meth_set_create(method.get(), [](BIO *b) -> int {
+    BIO_set_init(b, 1);
+    return 1;
+  });
```
andrewhop, line 10478:
```suggestion
  ASSERT_TRUE(BIO_meth_set_create(method.get(), [](BIO *b) -> int {
    BIO_set_init(b, 1);
    return 1;
  }));
```

ssl/ssl_test.cc
```diff
@@ -10458,6 +10458,57 @@ TEST(SSLTest, ErrorSyscallAfterCloseNotify) {
...
+  BIO_meth_set_read(method.get(), [](BIO *, char *, int) -> int {
+    return 0;
+  });
```
andrewhop, line 10481:
```suggestion
  ASSERT_TRUE(BIO_meth_set_read(method.get(), [](BIO *, char *, int) -> int {
    return 0;
  }));
```

ssl/ssl_test.cc
```diff
@@ -10458,6 +10458,57 @@ TEST(SSLTest, ErrorSyscallAfterCloseNotify) {
...
+  ret = SSL_read(client.get(), buf, sizeof(buf));
+  EXPECT_EQ(ret, 0);
+  EXPECT_EQ(SSL_get_error(client.get(), ret), SSL_ERROR_WANT_READ);
+
+  // Reset client rbio, read should succeed
+  SSL_set0_rbio(client.get(), client_rbio.release());
+  ret = SSL_read(client.get(), buf, sizeof(buf));
+  EXPECT_EQ(ret, (int) sizeof(buf));
+  EXPECT_EQ(SSL_get_error(client.get(), ret), SSL_ERROR_NONE);
+}
```
andrewhop, line 10510: If you tried to read another byte here what do you get?

WillChilds-Klein: we get `SSL_ERROR_WANT_READ` from `SSL_get_error`, but this time with a ret val of -1 (as opposed to 0 retval on ssl_test.cc L10502). `SSL_ERROR_WANT_READ` is retryable, that doesn't seem right....

moving PR back to draft pending further investigation.

WillChilds-Klein: well, I confirmed that even before this change, trying to read additional bytes after exhausting the `rbio` resulted in `SSL_ERROR_WANT_READ`. so, it looks like this change doesn't effect that specific behavior.

justsmth (approved):

justsmth (approved):

WillChilds-Klein:
> This seemed relevant to the test failures I'm running into with Bind9, so I tried applying the change to see if it would resolve my issues. It results in the tests from Bind9 hanging forever though (the tests used to directly fail) :(. I'm pasting this code block from bind9 here, it seems they were dependent on detecting the original behaviour:
> 
>     * https://gitlab.isc.org/isc-projects/bind9/-/blob/4330014fce5ab2ac1a6e58cd330762318a34f8d9/lib/isc/netmgr/tlsstream.c#L731-750

@samuel40791765 -- from what I can tell, bind9 [does not set SSL_MODE_AUTO_RETRY](https://gitlab.isc.org/search?search=SSL_MODE_AUTO_RETRY&nav_source=navbar&project_id=1&group_id=2&search_code=true&repository_ref=main), so this change should not affect that integration.

andrewhop (approved):

PR #1337, opened 2023-12-04
Improve backwards compatability for the TLS transfer parser

torben-hansen:
### Description of changes: 

Allows the parser to deserialise TLS transfer state as long as the state version is as old or older than the parser.

https://github.com/aws/aws-lc/pull/1322 added a few more (optional) fields to the TLS transfer format. We want this to impose a version bump. But the parser currently doesn't allow inputting state serialised through an older parser. Improve backwards compatibility to allow future version bumps.

This makes deploying to a large fleet with a long delta of software versions being out-of-sync much easier. 

### DIFF
--- a/ssl/ssl_transfer_asn1.cc
+++ b/ssl/ssl_transfer_asn1.cc
@@ -922,7 +922,7 @@ static int SSL3_STATE_from_bytes(SSL *ssl, CBS *cbs, const SSL_CTX *ctx) {
 
 // SSL_CONFIG serialization.
 
-static const unsigned kSSLConfigVersion = 1;
+static const unsigned kSSLConfigVersion = 2;
 
 static const unsigned kSSLConfigOcspStaplingEnabledTag =
     CBS_ASN1_CONSTRUCTED | CBS_ASN1_CONTEXT_SPECIFIC | 0;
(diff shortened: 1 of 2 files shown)


skmcgrail:
Still need to bump `SSL_SERIAL_VERSION` to version `2` right? And explicitly check that prior to de-serializing the new fields?

torben-hansen:
> Still need to bump `SSL_SERIAL_VERSION` to version `2` right? Maybe I didn't understand our discussion outcome earlier, but I thought that was the intent of the follow-up was to bump and handle the backwards compatibility with the older version?

I could bump it or just leave it as-is. There is no functional difference. I left it out in this PR with no real arguments for why.

torben-hansen:
> Still need to bump `SSL_SERIAL_VERSION` to version `2` right? Maybe I didn't understand our discussion outcome earlier, but I thought that was the intent of the follow-up was to bump and handle the backwards compatibility with the older version?

bumped config parser version: https://github.com/aws/aws-lc/pull/1337/commits/ed630eeb5cf013d48e1f9687ddcc17fe18d41264

skmcgrail (approved):

justsmth (approved):

PR #1324, opened 2023-11-27
Add back support for EVP_PKEY_HMAC

samuel40791765:
### Issues:
Resolves `CryptoAlg-1704`

### Description of changes: 
Some legacy libraries still depend on `EVP_PKEY_HMAC` unfortunately, so we're adding back support. OpenSSL 3.0 hasn't deprecated support for this either, so we still need to support this for the SHIM layer long term. Much of logic was reapplied from the [original commit removal](https://github.com/google/boringssl/commit/65ee9b7ce94872b8265f7a5e4a09a5929ed61d61), but I've cleaned some unneeded branching logic to makes things more straight forward.

The original `EVP_PKEY_HMAC` code from OpenSSL 1.1.1 exists in these two files:
* https://github.com/openssl/openssl/blame/OpenSSL_1_1_1-stable/crypto/hmac/hm_pmeth.c
* https://github.com/openssl/openssl/blame/OpenSSL_1_1_1-stable/crypto/hmac/hm_ameth.c

The code has remained relatively untouched since the removal in BoringSSL.

We've also made modifications to remove the need for additional `EVP_PKEY` function pointers and `ctrl` logic. Since HMAC needs to be specially handled in many of the `EVP_DigestSign*` APIs, the additional function pointers make much less sense. We can just directly make calls to `HMAC_CTX` without the `EVP_PKEY_METHOD` redirections.

### Call-outs:
- [X] Reapply the [original removal]((https://github.com/google/boringssl/commit/65ee9b7ce94872b8265f7a5e4a09a5929ed61d61),)
- [X] Reapply tests from the original removal.
- [X] Clean up code 
- [x] Add tests against HMAC [wycheproof test vectors
](https://github.com/aws/aws-lc/tree/main/third_party/wycheproof_testvectors)
- [x] Update documentation that mentions the deprecation of `EVP_PKEY_HMAC` and add follow up comments for service indicator adjustments
- [x] Migrate ASN1 usage to use CBB/CBS
- [x] Add test using `EVP_PKEY_CTX_dup` with `EVP_PKEY_HMAC`

### Testing:
New tests

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

PR #1338, opened 2023-12-05
allow HMAC via EVP_PKEY raw privkey functions

samuel40791765:
### Issues:
Resolves `CryptoAlg-1704`

### Description of changes: 
Follow up from https://github.com/aws/aws-lc/pull/1324#issuecomment-1832848738, Bind consumes `EVP_PKEY_HMAC` though this codepath, so we'll have to add support for this within the EVP_PKEY function pointers.
OpenSSL also encourages using `EVP_PKEY_new_raw_private_key` instead of `EVP_PKEY_new_mac_key`: https://www.openssl.org/docs/manmaster/man3/EVP_PKEY_new_mac_key.html

### Call-outs:
N/A

### Testing:
New test for `EVP_PKEY_{new,get}_raw_private_key`

### DIFF
--- a/crypto/evp_extra/p_hmac_asn1.c
+++ b/crypto/evp_extra/p_hmac_asn1.c
@@ -87,6 +87,7 @@ static int hmac_set_key(EVP_PKEY *pkey, const uint8_t *priv, size_t len,
 
   key->key = OPENSSL_memdup(priv, len);
   if (key->key == NULL && len > 0) {
+    OPENSSL_free(key);
     return 0;
   }
   key->key_len = len;
--- a/crypto/evp_extra/p_hmac_asn1.c
+++ b/crypto/evp_extra/p_hmac_asn1.c
@@ -57,6 +57,7 @@
 #include <openssl/evp.h>
 #include <openssl/mem.h>
 
+#include "../internal.h"
 #include "internal.h"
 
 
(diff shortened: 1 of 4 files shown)


crypto/evp_extra/p_hmac_asn1.c
```diff
@@ -66,9 +67,52 @@ static int hmac_size(OPENSSL_UNUSED const EVP_PKEY *pkey) {
...
+
+static int hmac_get_key(const EVP_PKEY *pkey, uint8_t *priv, size_t *len) {
+  HMAC_KEY *key = pkey->pkey.ptr;
+  if (key == NULL) {
+    return 0;
+  }
+
+  // The semantics of the EVP APIs are to return the length, if |priv| is NULL.
+  if (priv == NULL) {
+    *len = key->key_len;
```
WillChilds-Klein, line 102: should we do a null check on `len` here, or can we always assume that to be a valid pointer?

crypto/fipsmodule/evp/evp.c
```diff
@@ -290,7 +290,7 @@ EVP_PKEY *EVP_PKEY_new_mac_key(int type, ENGINE *engine, const uint8_t *mac_key,
   if(key == NULL) {
     goto err;
   }
-  key->key = mac_key;
+  key->key = OPENSSL_memdup(mac_key, mac_key_len);
```
WillChilds-Klein, line 293: we're allocating here; do we need to handle failure?

himsen: yes

crypto/fipsmodule/evp/evp.c
```diff
@@ -290,8 +290,12 @@ EVP_PKEY *EVP_PKEY_new_mac_key(int type, ENGINE *engine, const uint8_t *mac_key,
...
+  if (key->key == NULL && mac_key_len > 0) {
+    return 0;
```
WillChilds-Klein, line 295: will this leak memory after the allocation of `ret`? other error cases `goto err;`

samuel40791765: Ahhh nice catch

crypto/evp_extra/p_hmac_asn1.c
```diff
@@ -66,9 +67,55 @@ static int hmac_size(OPENSSL_UNUSED const EVP_PKEY *pkey) {
...
+  }
+
+  HMAC_KEY *key = HMAC_KEY_new();
+  if (key == NULL) {
+    return 0;
+  }
+
+  key->key = OPENSSL_memdup(priv, len);
+  if (key->key == NULL && len > 0) {
+    return 0;
```
WillChilds-Klein, line 90: free `key`?

WillChilds-Klein (approved):

torben-hansen (approved):

PR #1342, opened 2023-12-05
Fix delocator and FIPS static build prefixing behaviors

skmcgrail:
### Description of changes: 

This pull request addresses two primary "behaviors" that were occurring with the FIPS static build when prefixing symbols.

1. On aarch64 architectures, if you attempt to prefix the `OPENSSL_armcap_P` symbol, you would get a build failure:
```
[build] /usr/bin/ld: crypto/libcrypto.a(bcm.o): in function `wait_for_entropy_once_bss_get':
[build] bcm.c:(.text+0xde644): undefined reference to `OPENSSL_armcap_P'
[build] /usr/bin/ld: bcm.c:(.text+0xde648): undefined reference to `OPENSSL_armcap_P'
[build] collect2: error: ld returned 1 exit status
```

2. The x86-64 and aarch64 architectures prefix builds each had unexpected behaviors when the delocator operated on the assembly for each architecture. First on x86-64 platforms the assembly output from the compilation of `bcm.c` would have symbols prefixed, but the Perl/generated-src assembly would not be prefixed. This is problematic as the delocator is not prefix build aware. All its symbol parsing logic expects the unprefixed names. So special handling of `OPENSSL_ia32cap_P` and `OPENSSL_armcap_P` would not get handled correctly in all instances of their usage. For example, in the case of `OPENSSL_ia32cap_P` if the symbol was prefixed a redirector would be generated, for the unprefixed assembly a proper `OPENSSL_ia32cap_get` local symbol would have been generated. The later behavior is what is expected regardless of prefixed or standard build. Similarly aarch64 architectures had the same problem. But in this instance it was the Perl/generated-src's were also getting prefixed since that assembly was ran through the C preprocessor, it then subsequently would have incorrect handling of special symbols similar to the x86-64 presentation.

The solution is to not add the prefix headers include path when compiling the `bcm.c` (`bcm_c_generated_asm` target) to assembly, and to not include it when executing the C preprocessing the other Perl/generated-src files. Rather we will emit C preproeccor include directives in the resulting `bcm-Delocated.S` file. This will ensure that when the final compilation target is ran (`bcm_hashunset` target) we can include the prefix header include path, and the C preprocessor include directives will esnure that the symbol prefixing occurs in the final target.

As an additional simplification I've updated the CMake build to just allow the delocator to perform the preprocessing step of the assembly (now regardless of platform). This functionality already existed, but we weren't leveraging it. This also provides us the opportunity to know what headers we want to inject for include paths into the resulting bcm-Delocated.S file.


crypto/fipsmodule/CMakeLists.txt
```diff
@@ -305,7 +305,7 @@ function(cpreprocess dest src)
   string(REGEX REPLACE "[ ]+" ";" CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS}")
   add_custom_command(
     OUTPUT ${dest}
-    COMMAND ${CMAKE_ASM_COMPILER} ${TARGET} ${CMAKE_ASM_FLAGS} -E ${src} ${PREFIX_INCLUDE} -I${PROJECT_SOURCE_DIR}/include > ${dest}
+    COMMAND ${CMAKE_ASM_COMPILER} ${TARGET} ${CMAKE_ASM_FLAGS} -E -dI ${src} ${PREFIX_INCLUDE} -I${PROJECT_SOURCE_DIR}/include > ${dest}
```
nebeid, line 308: My understanding is that the solution was to keep the #include directives using this preprocessor compilation flag, correct?
https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html
> -dI
    Output ‘#include’ directives in addition to the result of preprocessing.

skmcgrail: This comment isn't necessarily relevant to the actual fix, but is related to the solution. The real solution was to have the delocator inject in `#include` C preprocesser directives so that we can do prefixing after the delocator has executed.

nebeid (approved):

torben-hansen (approved):
Do you need to handle `BORINGSSL_bcm_text_[start,end]` on aarch64 archs?

I think I implemented them as globals

skmcgrail:
> Do you need to handle `BORINGSSL_bcm_text_[start,end]` on aarch64 archs?
> 
> I think I implemented them as globals

We do as those are not prefixed at all today, but the pose some challenges that I didn't want to try to solve in this PR. Ideally I'd like to get those converted to local symbols if at all possible, but that will a bit more work. That will be in a future follow-up, this at least gets us into a more proper state of delocator behavior while fixing some of the prefixing issues for aarch64.

PR #1346, opened 2023-12-06
More caution with 'rm -rf'

justsmth:
### Issues:
N/A

### Description of changes: 
Prevent accidental deletions (via `rm -rf`) when a variable is not defined.

### Call-outs:
* See Bash documentation for [parameter expansion](https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion).

> ${parameter:?word}
>
> If parameter is null or unset, the expansion of word (or a message to that effect if 
> word is not present) is written to the standard error and the shell, if it is not 
> interactive, exits. Otherwise, the value of parameter is substituted. 

### Testing:
N/A

### DIFF
--- a/tests/ci/integration/run_s2n_integration.sh
+++ b/tests/ci/integration/run_s2n_integration.sh
@@ -25,7 +25,7 @@ S2N_TLS_BUILD_FOLDER="${SCRATCH_FOLDER}/s2n-tls-build"
 
 # Make script execution idempotent.
 mkdir -p ${SCRATCH_FOLDER}
-rm -rf ${SCRATCH_FOLDER}/*
+rm -rf "${SCRATCH_FOLDER:?}"/*
 cd ${SCRATCH_FOLDER}
 
 # Test helper functions.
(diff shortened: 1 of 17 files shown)


andrewhop (approved):

tests/ci/integration/run_s2n_integration.sh
```diff
@@ -25,7 +25,7 @@ S2N_TLS_BUILD_FOLDER="${SCRATCH_FOLDER}/s2n-tls-build"
 
 # Make script execution idempotent.
 mkdir -p ${SCRATCH_FOLDER}
-rm -rf ${SCRATCH_FOLDER}/*
+rm -rf "${SCRATCH_FOLDER:?}"/*
```
WillChilds-Klein, line 28: is it worth pulling this "logic" up into an alias or function in `common_posix_setup.sh` and adding a descriptive error message (i.e. `word` in the bash documentation)?

WillChilds-Klein (approved):

PR #1352, opened 2023-12-08
Mark bcm_redirector functions as local

skmcgrail:
### Description of changes: 
Today the `bcm_redirector_*` functions are symbols that are present in the bcm.o file, and are subject to potential collisions if they are not correctly "prefixed". Today since we have the odd behavior of prefixing symbols prior to running delocate this hasn't cause an issue as we would get redirect symbols like `bcm_redirector_myprefix_OPENSSL_malloc` etc. While working on https://github.com/aws/aws-lc/pull/1342 I realized a side-effect to prefixing after running the delocator would cause the `bcm_redirector_` functions to more then likely collide since they would no longer contain the prefix e.g. `bcm_redirector_OPENSSL_malloc`. Since these symbols are only referenced within the bcm.o file this PR marks these symbols as local so that they don't appear in the object file.

### Testing:
With the FIPS static build enabled I have validated that `go run ./util/all_tests.go` passes.

### DIFF
--- a/util/fipstools/delocate/delocate.go
+++ b/util/fipstools/delocate/delocate.go
@@ -2294,7 +2294,7 @@ func isFipsScopeMarkers(symbol string) bool {
 }
 
 func redirectorName(symbol string) string {
-	return "bcm_redirector_" + symbol
+	return ".Lbcm_redirector_" + symbol
 }
 
 // sectionType returns the type of a section. I.e. a section called “.text.foo”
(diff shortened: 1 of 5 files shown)


torben-hansen (approved):
lgtm. doesn't work for COFF or macho. But non of our stuff works there anyway...

dkostic (approved):

PR #1355, opened 2023-12-12
Add helper functions needed for NTP

andrewhop:
### Issues:
Addresses CryptoAlg-2187

### Description of changes: 
Add:
* EVP_MD_get_pkey_type
* EVP_MD_get0_name
* RAND_write_file
* DES_key_sched

### Call-outs:
AWS-LC doesn't support reading in entropy from a file with `RAND_load_file` so it seemed appropriate to not implement `RAND_write_file`. However, customers could try and use that with another library so we might want to implement it. For now it just returns an error and does nothing. At least for NTP they only use RAND load/write in the case that `RAND_status` returns 0, and for AWS-LC it always returns 1. 

EVP_MD_get_pkey_type is really weird, based on the [OpenSSL documentation](https://www.openssl.org/docs/manmaster/man3/EVP_MD_get_pkey_type.html) and reading [their code all current digests](https://github.com/openssl/openssl/blob/986c48c4eb26861f25bc68ea252d8f2aad592735/crypto/evp/legacy_sha.c#L124) are "associated" with RSA and they expect to return that PKEY type even though you could use ECDSA. 

### Testing:
Tested with NTP and locally with new tests.

### DIFF
--- a/crypto/fipsmodule/evp/evp.c
+++ b/crypto/fipsmodule/evp/evp.c
@@ -221,7 +221,7 @@ int EVP_MD_get_pkey_type(const EVP_MD *md) {
 }
 
 const char *EVP_MD_get0_name(const EVP_MD *md) {
-  if (md) {
+  if (md != NULL) {
     return OBJ_nid2sn(EVP_MD_nid(md));
   }
   return NULL;
--- a/crypto/fipsmodule/evp/evp.c
+++ b/crypto/fipsmodule/evp/evp.c
@@ -210,6 +210,24 @@ int EVP_PKEY_id(const EVP_PKEY *pkey) {
   return pkey->type;
 }
 
+int EVP_MD_get_pkey_type(const EVP_MD *md) {
+  if (md) {
+    int sig_nid, result = 0;
+    result = OBJ_find_sigid_by_algs(&sig_nid, md->type, NID_rsaEncryption);
+    if (result) {
+      return sig_nid;
+    }
+  }
+  return 0;
+}
+
+const char *EVP_MD_get0_name(const EVP_MD *md) {
+  if (md) {
+    return OBJ_nid2sn(EVP_MD_nid(md));
+  }
+  return NULL;
+}
+
 extern const EVP_PKEY_ASN1_METHOD *const *AWSLC_non_fips_pkey_evp_asn1_methods(void);
 
 // evp_pkey_asn1_find returns the ASN.1 method table for the given |nid|, which
--- a/crypto/fipsmodule/evp/evp.c
+++ b/crypto/fipsmodule/evp/evp.c
@@ -220,11 +220,19 @@ int EVP_MD_get_pkey_type(const EVP_MD *md) {
   return 0;
 }
 
+int EVP_MD_pkey_type(const EVP_MD *md){
+  return EVP_MD_get_pkey_type(md);
+}
+
 const char *EVP_MD_get0_name(const EVP_MD *md) {
   if (md != NULL) {
     return OBJ_nid2sn(EVP_MD_nid(md));
   }
   return NULL;
+}
+
+const char *EVP_MD_name(const EVP_MD *md) {
+  return EVP_MD_get0_name(md);
 }
 
 extern const EVP_PKEY_ASN1_METHOD *const *AWSLC_non_fips_pkey_evp_asn1_methods(void);
--- a/crypto/fipsmodule/evp/evp.c
+++ b/crypto/fipsmodule/evp/evp.c
@@ -212,9 +212,8 @@ int EVP_PKEY_id(const EVP_PKEY *pkey) {
 
 int EVP_MD_get_pkey_type(const EVP_MD *md) {
   if (md) {
-    int sig_nid, result = 0;
-    result = OBJ_find_sigid_by_algs(&sig_nid, md->type, NID_rsaEncryption);
-    if (result) {
+    int sig_nid = 0;
+    if (OBJ_find_sigid_by_algs(&sig_nid, md->type, NID_rsaEncryption)) {
       return sig_nid;
     }
   }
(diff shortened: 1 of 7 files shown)


include/openssl/rand.h
```diff
@@ -82,6 +82,9 @@ OPENSSL_EXPORT void RAND_seed(const void *buf, int num);
 // RAND_load_file returns a nonnegative number.
 OPENSSL_EXPORT int RAND_load_file(const char *path, long num);
 
+// RAND_write_file does nothing and returns negative 1
```
torben-hansen, line 85:
```suggestion
// RAND_write_file does nothing and returns negative 1.
```

Inconsistent use of `.`s :)

include/openssl/evp.h
```diff
@@ -141,6 +140,16 @@ OPENSSL_EXPORT int EVP_PKEY_id(const EVP_PKEY *pkey);
 // otherwise.
 OPENSSL_EXPORT int EVP_PKEY_type(int nid);
 
+// EVP_MD_get_pkey_type returns the NID of the public key signing algorithm
+// associated with |md| and RSA.
+OPENSSL_EXPORT int EVP_MD_get_pkey_type(const EVP_MD *md);
+#define EVP_MD_pkey_type EVP_MD_get_pkey_type
```
torben-hansen, line 146: why these defines?

include/openssl/evp.h
```diff
@@ -141,6 +140,16 @@ OPENSSL_EXPORT int EVP_PKEY_id(const EVP_PKEY *pkey);
 // otherwise.
 OPENSSL_EXPORT int EVP_PKEY_type(int nid);
 
+// EVP_MD_get_pkey_type returns the NID of the public key signing algorithm
+// associated with |md| and RSA.
+OPENSSL_EXPORT int EVP_MD_get_pkey_type(const EVP_MD *md);
```
torben-hansen, line 145: say deprecated?

crypto/fipsmodule/evp/evp.c
```diff
@@ -210,6 +210,24 @@ int EVP_PKEY_id(const EVP_PKEY *pkey) {
   return pkey->type;
 }
 
+int EVP_MD_get_pkey_type(const EVP_MD *md) {
+  if (md) {
+    int sig_nid, result = 0;
+    result = OBJ_find_sigid_by_algs(&sig_nid, md->type, NID_rsaEncryption);
+    if (result) {
```
torben-hansen, line 217:
```suggestion
    if (OBJ_find_sigid_by_algs(&sig_nid, md->type, NID_rsaEncryption)) {
```

torben-hansen (dismissed):

crypto/fipsmodule/evp/evp.c
```diff
@@ -210,6 +210,23 @@ int EVP_PKEY_id(const EVP_PKEY *pkey) {
...
+    int sig_nid = 0;
+    if (OBJ_find_sigid_by_algs(&sig_nid, md->type, NID_rsaEncryption)) {
+      return sig_nid;
+    }
+  }
+  return 0;
+}
+
+const char *EVP_MD_get0_name(const EVP_MD *md) {
+  if (md) {
```
samuel40791765, line 224:
```suggestion
  if (md != NULL) {
```
And above on L214

include/openssl/evp.h
```diff
@@ -141,6 +140,10 @@ OPENSSL_EXPORT int EVP_PKEY_id(const EVP_PKEY *pkey);
 // otherwise.
 OPENSSL_EXPORT int EVP_PKEY_type(int nid);
 
+// EVP_MD_get0_name returns the short name of |md|
+OPENSSL_EXPORT const char *EVP_MD_get0_name(const EVP_MD *md);
+#define EVP_MD_name EVP_MD_get0_name
```
samuel40791765, line 145: Is there a particular way that NTP is consuming this as a macro?  We could define `EVP_MD_name`  and `EVP_MD_pkey_type` as functions anyways and add them to this section below?
https://github.com/aws/aws-lc/blob/95cc2ce94a00304eaa887a80b851b89e5f0c65d8/include/openssl/evp.h#L1158-L1169

andrewhop: Not sure what you mean, NTP is just calling `name = EVP_MD_name(m);` in ntpq.c. They are using it in a way that this could be a function. But in theory other consumers could try to use it as a macro in `if defined(EVP_MD_name)`.

samuel40791765: The section I linked to redefines the functions as macros, so consumers can still detect it with `#if defined(EVP_MD_name)`. The comment within the section also specifies that this was created to avoid breaking `#idefs` in consumers: https://github.com/aws/aws-lc/blob/95cc2ce94a00304eaa887a80b851b89e5f0c65d8/include/openssl/evp.h#L1160-L1162

I think we could still define `EVP_MD_name` and `EVP_MD_pkey_type` as functions, and add them to the `Preprocessor compatibility section`. It's a consistent trend we've been doing, `ssl.h` has more examples here:
https://github.com/aws/aws-lc/blob/95cc2ce94a00304eaa887a80b851b89e5f0c65d8/include/openssl/ssl.h#L5657-L5803

include/openssl/evp.h
```diff
@@ -141,6 +140,12 @@ OPENSSL_EXPORT int EVP_PKEY_id(const EVP_PKEY *pkey);
 // otherwise.
 OPENSSL_EXPORT int EVP_PKEY_type(int nid);
 
+// EVP_MD_get0_name returns the short name of |md|
+OPENSSL_EXPORT const char *EVP_MD_get0_name(const EVP_MD *md);
+
+// EVP_MD_name calls |EVP_MD_get0_name|
+OPENSSL_EXPORT const char *EVP_MD_name(const EVP_MD *md);
+#define EVP_MD_name EVP_MD_name
```
samuel40791765, line 148: We should move these down to the section below or the prefix build could fail.
https://github.com/aws/aws-lc/blob/95cc2ce94a00304eaa887a80b851b89e5f0c65d8/include/openssl/evp.h#L1164-L1169

samuel40791765 (approved):

torben-hansen (approved):

PR #1359, opened 2023-12-15
 Add cpython integration test

WillChilds-Klein:
### Issues:
Addresses i/CryptoAlg-2136

### Description of changes: 

This change adds an integration test for CPython versions 3.10 through tip-of-main. Running _all_ 3.10 python module tests on my AL2 dev machine takes ~5m:

```
$ time ./tests/ci/integration/run_python_integration.sh
...
== Tests result: FAILURE then SUCCESS ==

404 tests OK.

21 tests skipped:
    test_bz2 test_ctypes test_devpoll test_idle test_ioctl test_kqueue
    test_msilib test_ossaudiodev test_readline test_sqlite
    test_startfile test_tcl test_tix test_tk test_ttk_guionly
    test_ttk_textonly test_turtle test_winconsoleio test_winreg
    test_winsound test_zipfile64

1 re-run test:
    test___all__
```

However, in [CI testing](https://us-west-2.console.aws.amazon.com/codesuite/codebuild/620771051181/projects/aws-lc-ci-integration/build/aws-lc-ci-integration%3A35a4d7f1-c75f-4db5-b39a-d328a4884efa?region=us-west-2), I found that some tests (unrelated to our `ssl` module integration) failed on our ubuntu 22.02 docker image:

```
== Tests result: FAILURE then FAILURE ==

395 tests OK.

9 tests failed:
    test_email test_generators test_multiprocessing_fork
    test_multiprocessing_forkserver test_multiprocessing_spawn
    test_pdb test_regrtest test_signal test_threading

21 tests skipped:
    test_bz2 test_dbm_gnu test_dbm_ndbm test_devpoll test_gdb
    test_idle test_ioctl test_kqueue test_msilib test_ossaudiodev
    test_startfile test_tcl test_tix test_tk test_ttk_guionly
    test_ttk_textonly test_turtle test_winconsoleio test_winreg
    test_winsound test_zipfile64

10 re-run tests:
    test___all__ test_email test_generators test_multiprocessing_fork
    test_multiprocessing_forkserver test_multiprocessing_spawn
    test_pdb test_regrtest test_signal test_threading
```

As a result, I've narrowed down our python test suite to only the modules that actually use our libssl (this list was discovered by calling `abort()` in `SSL_CTX_new` and seeing which module tests failed) as well as `hashlib`, which is the only other module other than `ssl` to include libcrypto. From the CPython 3.10 source to discover which modules use libcrypto:

```
$ ag -l 'include.+crypto' Modules/
Modules/_hashopenssl.c
Modules/_ssl.c
```

### Call-outs:

As we add support for further versions, we may want to parallelize the build and test processes.

### Testing:
- see description section above
 - CI integration tests

---

### DIFF
--- a/tests/ci/integration/python_patch/3.11/aws-lc-cpython.patch
+++ b/tests/ci/integration/python_patch/3.11/aws-lc-cpython.patch
@@ -65,18 +65,6 @@ index bd0fc9c2da..3ab70f5c1f 100644
              with self.reaped_server(SimpleIMAPHandler) as server:
                  client = self.imap_class(*server.server_address,
                                           ssl_context=ssl_context)
-diff --git a/Lib/test/test_site.py b/Lib/test/test_site.py
-index 8d336649ec..fa96b5964f 100644
---- a/Lib/test/test_site.py
-+++ b/Lib/test/test_site.py
-@@ -473,6 +473,7 @@ def test_sitecustomize_executed(self):
-     def test_license_exists_at_url(self):
-         # This test is a bit fragile since it depends on the format of the
-         # string displayed by license in the absence of a LICENSE file.
-+        #import ssl
-         url = license._Printer__data.split()[1]
-         req = urllib.request.Request(url, method='HEAD')
-         # Reset global urllib.request._opener
 diff --git a/Lib/test/test_ssl.py b/Lib/test/test_ssl.py
 index 1f881038c3..36fa1e32e8 100644
 --- a/Lib/test/test_ssl.py
--- a/tests/ci/integration/python_patch/3.11/aws-lc-cpython.patch
+++ b/tests/ci/integration/python_patch/3.11/aws-lc-cpython.patch
@@ -1,8 +1,8 @@
 diff --git a/Lib/test/test_asyncio/test_events.py b/Lib/test/test_asyncio/test_events.py
-index 5728d25254..3f86c14127 100644
+index d7871d3e53..fb65ee3b8b 100644
 --- a/Lib/test/test_asyncio/test_events.py
 +++ b/Lib/test/test_asyncio/test_events.py
-@@ -1099,12 +1099,12 @@ def test_create_server_ssl_match_failed(self):
+@@ -1103,12 +1103,12 @@ def test_create_server_ssl_match_failed(self):
          # incorrect server_hostname
          f_c = self.loop.create_connection(MyProto, host, port,
                                            ssl=sslcontext_client)
@@ -20,10 +20,10 @@ index 5728d25254..3f86c14127 100644
  
          # close connection
 diff --git a/Lib/test/test_httplib.py b/Lib/test/test_httplib.py
-index acae0127a1..3d9f0c031d 100644
+index 015a3d1e87..ec565ad181 100644
 --- a/Lib/test/test_httplib.py
 +++ b/Lib/test/test_httplib.py
-@@ -2009,7 +2009,7 @@ def test_host_port(self):
+@@ -2040,7 +2040,7 @@ def test_host_port(self):
  
      def test_tls13_pha(self):
          import ssl
@@ -33,10 +33,10 @@ index acae0127a1..3d9f0c031d 100644
          # just check status of PHA flag
          h = client.HTTPSConnection('localhost', 443)
 diff --git a/Lib/test/test_imaplib.py b/Lib/test/test_imaplib.py
-index ed26fa16d6..b0f0b0b2ab 100644
+index bd0fc9c2da..3ab70f5c1f 100644
 --- a/Lib/test/test_imaplib.py
 +++ b/Lib/test/test_imaplib.py
-@@ -558,9 +558,10 @@ def test_ssl_raises(self):
+@@ -561,9 +561,10 @@ def test_ssl_raises(self):
          self.assertEqual(ssl_context.check_hostname, True)
          ssl_context.load_verify_locations(CAFILE)
  
@@ -50,7 +50,7 @@ index ed26fa16d6..b0f0b0b2ab 100644
              _, server = self._setup(SimpleIMAPHandler)
              client = self.imap_class(*server.server_address,
                                       ssl_context=ssl_context)
-@@ -963,10 +964,10 @@ def test_ssl_verified(self):
+@@ -967,10 +968,10 @@ def test_ssl_verified(self):
          ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
          ssl_context.load_verify_locations(CAFILE)
  
@@ -66,7 +66,7 @@ index ed26fa16d6..b0f0b0b2ab 100644
                  client = self.imap_class(*server.server_address,
                                           ssl_context=ssl_context)
 diff --git a/Lib/test/test_site.py b/Lib/test/test_site.py
-index b5dc381a2f..65c78264e6 100644
+index 8d336649ec..fa96b5964f 100644
 --- a/Lib/test/test_site.py
 +++ b/Lib/test/test_site.py
 @@ -473,6 +473,7 @@ def test_sitecustomize_executed(self):
@@ -78,10 +78,10 @@ index b5dc381a2f..65c78264e6 100644
          req = urllib.request.Request(url, method='HEAD')
          # Reset global urllib.request._opener
 diff --git a/Lib/test/test_ssl.py b/Lib/test/test_ssl.py
-index 85a561f6b7..ba8a81c2fb 100644
+index 1f881038c3..36fa1e32e8 100644
 --- a/Lib/test/test_ssl.py
 +++ b/Lib/test/test_ssl.py
-@@ -40,6 +40,7 @@
+@@ -44,6 +44,7 @@
  
  Py_DEBUG = hasattr(sys, 'gettotalrefcount')
  Py_DEBUG_WIN32 = Py_DEBUG and sys.platform == 'win32'
@@ -89,7 +89,7 @@ index 85a561f6b7..ba8a81c2fb 100644
  
  PROTOCOLS = sorted(ssl._PROTOCOL_NAMES)
  HOST = socket_helper.HOST
-@@ -166,7 +167,7 @@ def is_ubuntu():
+@@ -170,7 +171,7 @@ def is_ubuntu():
      except FileNotFoundError:
          return False
  
@@ -98,7 +98,7 @@ index 85a561f6b7..ba8a81c2fb 100644
      def seclevel_workaround(*ctxs):
          """"Lower security level to '1' and allow all ciphers for TLS 1.0/1"""
          for ctx in ctxs:
-@@ -544,7 +545,7 @@ def test_openssl_version(self):
+@@ -548,7 +549,7 @@ def test_openssl_version(self):
          else:
              openssl_ver = f"OpenSSL {major:d}.{minor:d}.{fix:d}"
          self.assertTrue(
@@ -107,7 +107,7 @@ index 85a561f6b7..ba8a81c2fb 100644
              (s, t, hex(n))
          )
  
-@@ -1316,7 +1317,6 @@ def test_min_max_version(self):
+@@ -1320,7 +1321,6 @@ def test_min_max_version(self):
  
          if has_tls_protocol(ssl.PROTOCOL_TLSv1_1):
              ctx = ssl.SSLContext(ssl.PROTOCOL_TLSv1_1)
@@ -115,7 +115,7 @@ index 85a561f6b7..ba8a81c2fb 100644
              self.assertIn(
                  ctx.minimum_version, minimum_range
              )
-@@ -1376,24 +1376,25 @@ def test_load_cert_chain(self):
+@@ -1380,24 +1380,25 @@ def test_load_cert_chain(self):
          with self.assertRaises(OSError) as cm:
              ctx.load_cert_chain(NONEXISTINGCERT)
          self.assertEqual(cm.exception.errno, errno.ENOENT)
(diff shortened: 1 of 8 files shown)


tests/ci/integration/python_patch/main/aws-lc-cpython.patch
```diff
@@ -0,0 +1,555 @@
...
+-# _ssl _ssl.c $(OPENSSL_INCLUDES) $(OPENSSL_LDFLAGS) \
+-#     -l:libssl.a -Wl,--exclude-libs,libssl.a \
+-#     -l:libcrypto.a -Wl,--exclude-libs,libcrypto.a
+-# _hashlib _hashopenssl.c $(OPENSSL_INCLUDES) $(OPENSSL_LDFLAGS) \
+-#     -l:libcrypto.a -Wl,--exclude-libs,libcrypto.a
++_ssl _ssl.c $(OPENSSL_INCLUDES) $(OPENSSL_LDFLAGS) \
++    -l:libssl.a -Wl,--exclude-libs,libssl.a \
++    -l:libcrypto.a -Wl,--exclude-libs,libcrypto.a
++_hashlib _hashopenssl.c $(OPENSSL_INCLUDES) $(OPENSSL_LDFLAGS) \
++    -l:libcrypto.a -Wl,--exclude-libs,libcrypto.a
```
samuel40791765, line 390: Is this diff needed or was this auto-generated

tests/ci/integration/run_python_integration.sh
```diff
@@ -0,0 +1,116 @@
...
+
+mkdir -p ${PYTHON_SRC_FOLDER}
+pushd ${PYTHON_SRC_FOLDER}
+
+# Some environments disable IPv6 by default
+which sysctl && ( sysctl -w net.ipv6.conf.all.disable_ipv6=0 || /bin/true )
+echo 0 >/proc/sys/net/ipv6/conf/all/disable_ipv6 || /bin/true
+
+# NOTE: cpython keeps a unique branch per version, add version branches here
+# TODO: As we add more versions to support, we may want to parallelize here
```
samuel40791765, line 109: Could we also call out the list of known things we had to patch and tests we had to skip (e.g. PHA, FFDH, etc.)? The patch files show some great insight, but it would be nice to have a single source of truth here.

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
WillChilds-Klein: it's needed. the diff for each branch is slightly different, each was constructed manually, tested, and then `git diff`ed here.

tests/ci/integration/python_patch/3.11/aws-lc-cpython.patch
```diff
@@ -0,0 +1,554 @@
...
+-
+             self.assertIn(
+                 ctx.minimum_version, minimum_range
+             )
+@@ -1380,24 +1380,25 @@ def test_load_cert_chain(self):
+         with self.assertRaises(OSError) as cm:
+             ctx.load_cert_chain(NONEXISTINGCERT)
+         self.assertEqual(cm.exception.errno, errno.ENOENT)
+-        with self.assertRaisesRegex(ssl.SSLError, "PEM lib"):
++        with self.assertRaisesRegex(ssl.SSLError, "PEM (lib|routines)"):
```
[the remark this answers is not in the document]
WillChilds-Klein: [yes](https://github.com/aws/aws-lc/blob/eaa19c77d000175da90da7b4ef22ffe24dfc2a2f/crypto/err/err.c#L446).

tests/ci/integration/run_python_integration.sh
WillChilds-Klein: yeah, that makes sense. I'll update the comment here.

samuel40791765 (dismissed):

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
```diff
@@ -0,0 +1,555 @@
+diff --git a/Lib/test/test_asyncio/test_events.py b/Lib/test/test_asyncio/test_events.py
+index b25c097573..8f07868b82 100644
+--- a/Lib/test/test_asyncio/test_events.py
++++ b/Lib/test/test_asyncio/test_events.py
+@@ -1125,12 +1125,12 @@ def test_create_server_ssl_match_failed(self):
+         # incorrect server_hostname
+         f_c = self.loop.create_connection(MyProto, host, port,
+                                           ssl=sslcontext_client)
++        regex = "IP address mismatch, certificate is not valid for '127.0.0.1'"
```
andrewhop, line 9: Is this what AWS-LC returns?

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
```diff
@@ -0,0 +1,555 @@
...
+diff --git a/Lib/test/test_httplib.py b/Lib/test/test_httplib.py
+index 089bf5be40..1713a5fbae 100644
+--- a/Lib/test/test_httplib.py
++++ b/Lib/test/test_httplib.py
+@@ -2066,7 +2066,7 @@ def test_host_port(self):
+ 
+     def test_tls13_pha(self):
+         import ssl
+-        if not ssl.HAS_TLSv1_3:
++        if not ssl.HAS_TLSv1_3 or "AWS-LC" in ssl.OPENSSL_VERSION:
```
andrewhop, line 31: Instead of checking for AWS-LC is it reasonable to add a new flag like ssl.HAS_PHA?

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
```diff
@@ -0,0 +1,555 @@
...
+                                          ssl_context=ssl_context)
+diff --git a/Lib/test/test_site.py b/Lib/test/test_site.py
+index 0502181854..e23531f147 100644
+--- a/Lib/test/test_site.py
++++ b/Lib/test/test_site.py
+@@ -564,6 +564,7 @@ def test_customization_modules_on_startup(self):
+     def test_license_exists_at_url(self):
+         # This test is a bit fragile since it depends on the format of the
+         # string displayed by license in the absence of a LICENSE file.
++        #import ssl
```
andrewhop, line 76: Accidental comment?

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
```diff
@@ -0,0 +1,555 @@
...
+-
+             self.assertIn(
+                 ctx.minimum_version, minimum_range
+             )
+@@ -1162,24 +1162,25 @@ def test_load_cert_chain(self):
+         with self.assertRaises(OSError) as cm:
+             ctx.load_cert_chain(NONEXISTINGCERT)
+         self.assertEqual(cm.exception.errno, errno.ENOENT)
+-        with self.assertRaisesRegex(ssl.SSLError, "PEM lib"):
++        with self.assertRaisesRegex(ssl.SSLError, "PEM (lib|routines)"):
```
andrewhop, line 123: Should we standardize our error strings?

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
```diff
@@ -0,0 +1,555 @@
...
+     if (PySSL_ChainExceptions(self) < 0)
+@@ -2060,7 +2068,12 @@ _ssl__SSLSocket_shared_ciphers_impl(PySSLSocket *self)
+     len = 0;
+     for (i = 0; i < sk_SSL_CIPHER_num(server_ciphers); i++) {
+         cipher = sk_SSL_CIPHER_value(server_ciphers, i);
++#if defined(OPENSSL_IS_AWSLC)
++        size_t unused_idx;
++        if (sk_SSL_CIPHER_find(client_ciphers, &unused_idx, cipher) < 0)
++#else
+         if (sk_SSL_CIPHER_find(client_ciphers, cipher) < 0)
```
andrewhop, line 472: Why not add a find that doesn't take an out index that wraps the existing sk_find for better compatibility? Also can our version of find ever return a negative number?

tests/ci/integration/run_python_integration.sh
```diff
@@ -0,0 +1,139 @@
...
+#   - Modify the hashlib module's backing C code to only register BLAKE
+#     functions if the expected NID is available in linked libcrypto
+#   - Modify the ssl module's backing C code to separate notions of supporting
+#     TLSv1.3 from supporting post-handshake authentication as some libraries
+#     (namely AWS-LC) support TLSv1.3, but not the post-handshake
+#     authentication portion of that protocol.
+#   - Modify the ssl module's backing C code to account for AWS-LC's divergent
+#     function signature for |sk_SSL_CIPHER_find|
+#   - Modify the ssl module's backing C code to set |SSL_MODE_AUTO_RETRY| in
+#     all calls to |SSL{_CTX}_set_mode|
```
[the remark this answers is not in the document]
WillChilds-Klein: As of 1.1.1, [yes](https://github.com/openssl/openssl/commit/693cf80c6ff54ae276a44d305d4ad07168ec6895).

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
WillChilds-Klein: no, this is what OpenSSL returns. we return something like `CERTIFICATE_VERIFY_FAILED`.

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
WillChilds-Klein: We could propose that upstream, but I'm not sure whether it's worth adding a new public property to the `ssl` module API. Regarding this change, it would just streamline a few tests.

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
WillChilds-Klein: yep, thanks.

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
WillChilds-Klein: > Why not add a find that doesn't take an out index that wraps the existing sk_find for better compatibility?

I don't think we can do this, as (to my knowledge) C doesn't support function overloading. Or are you proposing that we create a new funciton/symbol altogether that wraps our existing `sk_SSL_CIPHER_find` implementation and use that here?

I suppose we could also just [change the signature to match OpenSSL's](https://github.com/aws/aws-lc/blob/main/include/openssl/stack.h#L202-L204), but that might risk breaking any consumers that already use that function.

> Also can our version of find ever return a negative number?

Not [according to our documentation](https://github.com/aws/aws-lc/blob/main/include/openssl/stack.h#L199-L200), no. Good catch! I'll update this.

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
WillChilds-Klein: That's certainly an option. If we don't consider error string contents part of our backwards compatibility "contract", then we're free to change them to hew more closely to some version(s) of OpenSSL. The scope of such an effort would probably be pretty sizeable, and I don't know if it would provide any material value other than satisfying brittle tests that key on error string contents.

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
andrewhop: I was thinking we could have `sk_SSL_CIPHER_find_internal` which is our current implementation and `sk_SSL_CIPHER_find` which matches OpenSSL. I don't have a sense for how common OpenSSL stack usage outside of libcrypto/libssl is.

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
andrewhop: Ahhh is this an external ssl module? I was thinking this was a part of some test config.

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
WillChilds-Klein: > I don't have a sense for how common OpenSSL stack usage outside of libcrypto/libssl is.

It looks like it's [non-zero](https://github.com/search?q=sk_SSL_CIPHER_find&type=code). What percentage of those projects depend on OSSL's signature vs. BSSL's signature? I don't know, but I would guess it would skew towards OSSL given its ubiquity. If we change the signature of `sk_SSL_CIPHER_find`, though, it would necessarily become a point of incompatibility with BSSL.

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
WillChilds-Klein: Sorry, the term "module" is a bit [overloaded](https://docs.python.org/3/tutorial/modules.html). This `ssl` thing is CPython's standard library implementation of TLS, which is a thin wrapper around OpenSSL. So, whatever we properties we put on the `ssl` object become public APIs requiring stability, maintenance, etc.

tests/ci/integration/run_python_integration.sh
```diff
@@ -0,0 +1,139 @@
...
+    popd
+}
+
+# The per-branch patch files do a few things:
+#
+#   - Modify various unit tests to account for error string differences between
+#     OpenSSL and AWS-LC.
+#   - In |test_bio|handshake|, check whether protocol is TLSv1.3 before testing
+#     tls-unique channel binding behavior, as channel bindings are not defined
+#     on that protocol
```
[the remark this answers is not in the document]
WillChilds-Klein: I believe it's a difference in channel binding "support" between AWS-LC and OpenSSL. I re-built my python fork against OSSL 3.0, reverted changes to this test, and dumped the return value of `get_channel_binding()`. On a TLSv1.3 connection OSSL will actually return a byte array, which is surprising given that channel bindings [aren't defined for TLSv1.3](https://datatracker.ietf.org/doc/html/rfc8446#appendix-C.5). The CPython team [has been made aware](https://github.com/python/cpython/issues/95341) of this oddity.

samuel40791765 (dismissed):

samuel40791765 (approved):

andrewhop (approved):

tests/ci/integration/run_python_integration.sh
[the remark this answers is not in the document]
Neustradamus: Dear @andrewhop, @WillChilds-Klein, I am happy to see your comments.

It is possible to talk on the CPython specified tickets, the problems with missing "tls-exporter" RFC9266 support, and "tls-server-end-point" support too?

I have done several tickets here without security solutions:
- https://github.com/python/cpython/issues?q=is%3Aissue+author%3A%40me+

-> 2 years, no security solutions...

Several projects are impacted, I do not know all, example Slixmpp project has done a recent annnouncement:
- https://blog.mathieui.net/en/slixmpp-1.8.5.html
- https://codeberg.org/poezio/slixmpp/issues/3498

Thanks a lot in advance.

PR #1367, opened 2023-12-19
Add CMake GitHub Actions CI

skmcgrail:

### DIFF
--- a/.github/docker_images/cmake_build_versions/Dockerfile
+++ b/.github/docker_images/cmake_build_versions/Dockerfile
@@ -1,0 +1,33 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+FROM amazonlinux:2023
+
+ARG CMAKE_VERSION
+ARG CMAKE_DOWNLOAD_URL
+ARG CMAKE_SHA256
+
+VOLUME ["/awslc"]
+
+RUN dnf install -y tar ninja-build golang gcc gcc-c++ libcurl-devel libarchive-devel zlib-devel xz-devel
+
+RUN mkdir /cmake
+
+COPY cmake_build.sh /
+COPY awslc_build.sh /
+COPY entry.sh /
+
+WORKDIR /cmake
+
+RUN curl -L -o source.tar.gz "${CMAKE_DOWNLOAD_URL}"  && \
+    echo "${CMAKE_SHA256} source.tar.gz" | sha256sum -c - && \
+    mkdir source && \
+    tar -x -f source.tar.gz -v --strip-components=1 -C source
+
+WORKDIR /cmake/source
+
+RUN /cmake_build.sh
+
+WORKDIR /
+
+ENTRYPOINT ["/entry.sh"]
--- a/.github/docker_images/cmake_build_versions/awslc_build.sh
+++ b/.github/docker_images/cmake_build_versions/awslc_build.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+set -ex -o pipefail
+
+echo "Building with CMake Version: $(cmake --version)"
+
+BUILD_DIR=$(mktemp -d)
+SRC_DIR="${SRC_DIR:-/awslc}"
+CMAKE_CONFIGURE_ARGS=("${argv[@]}")
+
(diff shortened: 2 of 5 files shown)


.github/docker_images/cmake_build_versions/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+FROM amazonlinux:2023
+
+ARG CMAKE_VERSION
+ARG CMAKE_DOWNLOAD_URL
+ARG CMAKE_SHA256
+
+VOLUME ["/awslc"]
+
+RUN dnf install -y tar ninja-build golang gcc gcc-c++ libcurl-devel libarchive-devel zlib-devel xz-devel
```
WillChilds-Klein, line 12: I'm not familiar with `dnf`. Is there some analog to `apt update` or `yum update` that we should run before installing?

.github/docker_images/cmake_build_versions/awslc_build.sh
```diff
@@ -0,0 +1,25 @@
...
+CMAKE_CONFIGURE_ARGS=("${argv[@]}")
+
+pushd "${BUILD_DIR}"
+
+cmake "${SRC_DIR}" "${CMAKE_CONFIGURE_ARGS[@]}"
+cmake --build .
+
+"${BUILD_DIR}/crypto/crypto_test"
+
+if [[ -a "${BUILD_DIR}/ssl/ssl_test"  ]]; then
```
WillChilds-Klein, line 21: when would we not build libssl?

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,40 @@
...
+          - 0
+          - 1
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build Docker Image
+        working-directory: .github/docker_images/cmake_build_versions
+        run: |
+          docker build -t "cmake-${{ matrix.cmake.version }}" --build-arg CMAKE_VERSION=${{ matrix.cmake.version }} --build-arg CMAKE_DOWNLOAD_URL=${{ matrix.cmake.url }} --build-arg CMAKE_SHA256=${{ matrix.cmake.hash }} .
+      - name: ${{ matrix.generator }} (Static)
```
WillChilds-Klein, line 35: why not parameterize `FIPS` and `BUILD_SHARED_LIBS` in the same way?

.github/docker_images/cmake_build_versions/cmake_build.sh
```diff
@@ -0,0 +1,14 @@
...
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+set -ex -o pipefail
+
+echo "Building CMake Version: ${CMAKE_VERSION:-unknown}"
+
+# At the moment this works fine for all versions, in the future build logic can be modified to
+# look at it ${CMAKE_VERSION}.
+./configure --prefix=/opt/cmake --system-curl --system-libarchive
```
WillChilds-Klein, line 12: huh, does CMake depend on libcurl?

WillChilds-Klein (commented):
nice solution of building new containers in each task to get around version-namespacing CMake builds!

.github/docker_images/cmake_build_versions/Dockerfile
skmcgrail: `dnf` is technically the replacement for `yum` that occurred in Amazon Linux 2023. Not that you can't still use the `yum` commands. https://docs.aws.amazon.com/linux/al2023/ug/compare-with-al2.html

.github/docker_images/cmake_build_versions/awslc_build.sh
skmcgrail: By default we do, but since I have the script taking in the CMake configuration arguments in a generic manner it could be disabled and not compiled in the build if we wanted to. But at the moment in this PR, it will always be built and thus always tested.

.github/workflows/cmake.yml
skmcgrail: Was Just balancing how many times we would need to fetch and compile a given CMake version. I'd like to explore if we could just build these once, and put them in an S3 Bucket with a public cloudfront distribution so we can just pull the rebuilt artifacts. But I want to do that at the moment.

.github/docker_images/cmake_build_versions/cmake_build.sh
skmcgrail: Indeed it does, thus it depends on OpenSSL!

andrewhop (approved):

WillChilds-Klein (approved):

PR #1368, opened 2023-12-20
Add option to run speed for specified milliseconds

andrewhop:
### Description of changes: 
Currently speed only supports running for a set number of integer seconds, which means the shortest time is 1 second. That's a good default but to generate a lot of data for various sizes (everything from 1 byte up to 1000) of symmetric ciphers this can end up taking hours.

### Call-outs:
I thought about changing the existing timeout option to take milliseconds but that might break other existing tools so I decided to add a new option. 

### Testing:
Tested that both can't be set:
```
./tool/bssl speed -timeout_ms 10 -timeout 10
'-timeout' and '-timeout_ms' are mutually exclusive
-filter A comma-separated list of filters on the speed tests to run. Each filter is applied in independent runs.
-timeout        The number of seconds to run each test for (default is 1)
-timeout_ms     The number of milliseconds to run each test for (default is 1000)
...
```

### DIFF
--- a/tool/speed.cc
+++ b/tool/speed.cc
@@ -224,8 +224,8 @@ static uint64_t time_now() {
 }
 #endif
 
-#define TIMEOUT_SECONDS_DEFAULT 1
-static uint64_t g_timeout_seconds = TIMEOUT_SECONDS_DEFAULT;
+#define TIMEOUT_MS_DEFAULT 1000
+static uint64_t g_timeout_ms = TIMEOUT_MS_DEFAULT;
 static std::vector<size_t> g_chunk_lengths = {16, 256, 1350, 8192, 16384};
 static std::vector<size_t> g_prime_bit_lengths = {2048, 3072};
 static std::vector<std::string> g_filters = {""};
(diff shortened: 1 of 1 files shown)


tool/speed.cc
```diff
@@ -2546,8 +2551,18 @@ bool Speed(const std::vector<std::string> &args) {
...
 
+  if (args_map.count("-timeout") != 0 && args_map.count("-timeout_ms") != 0) {
+    puts("'-timeout' and '-timeout_ms' are mutually exclusive");
+    PrintUsage(kArguments);
+    return false;
+  }
+
   if (args_map.count("-timeout") != 0) {
-    g_timeout_seconds = atoi(args_map["-timeout"].c_str());
+    g_timeout_ms = atoi(args_map["-timeout"].c_str()) * 1000;
```
WillChilds-Klein, line 2561: not specific to this change, what happens if the user specifies 0 or a negative number? are those valid?

WillChilds-Klein (approved):

skmcgrail (approved):

justsmth (approved):

PR #1354, opened 2023-12-12
Add print helper functions for RSA and DSA

andrewhop:
### Issues:

Addresses CryptoAlg-2187

### Description of changes: 
Add the following 3 print functions.
* DSA_print
* DSA_print_fp
* RSA_print_fp

### Call-outs:
We already had an implementation of RSA_print. I tried to think of ways to share more code between DSA and RSA but I don't think it's worth the complexity to make the function work for both types.

OpenSSL has OPENSSL_NO_STDIO macro that AWS-LC has in some spots, but we already include stdio.h in other spots not guarded by this macro so I decided to always include it. 

### Testing:
Included new tests.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

PR #1369, opened 2023-12-20
Add an integration test for NTP

andrewhop:
### Issues:
Resolves CryptoAlg-2187

### Description of changes: 
With https://github.com/aws/aws-lc/pull/1354 and https://github.com/aws/aws-lc/pull/1355 AWS-LC is compatible with the latest version of NTP. This change adds a test to make sure we continue to work. 

### Call-outs:
AWS-LC already defines `OPENSSL_NO_MDC2` in our public header files, however NTP currently doesn't check for that in their digest tests and expects to use MDC2. For now I added a patch to update the test and check `OPENSSL_NO_MDC2` is not defined. 

NTP does not have an up to date GitHub mirror, it seems like the easiest way to get the code is from https://www.eecis.udel.edu/~ntp/ntp_spool/ntp4/ntp-4.2/. Right now the integration script is hard coded to download 4.2.8p17 which is the latest right now.

### Testing:
Built and ran locally. 


tests/ci/integration/run_ntp_integration.sh
```diff
@@ -0,0 +1,66 @@
...
+}
+
+function ntp_build() {
+  ./configure --with-openssl-incdir="${AWS_LC_INSTALL_FOLDER}/include" --with-openssl-libdir="${AWS_LC_INSTALL_FOLDER}/lib/"
+  make -j "${NUM_CPU_THREADS}"
+}
+
+function ntp_run_tests() {
+  export LD_LIBRARY_PATH="${AWS_LC_INSTALL_FOLDER}/lib"
+  make check
```
WillChilds-Klein, line 41: nit: `make -j ${NUM_CPU_THREADS} check`?

tests/ci/integration/run_ntp_integration.sh
```diff
@@ -0,0 +1,66 @@
...
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+source tests/ci/common_posix_setup.sh
+
+# Set up environment.
+
+# SYS_ROOT
+#  - SRC_ROOT(aws-lc)
+#    - SCRATCH_FOLDER
+#      - ntp
```
WillChilds-Klein, line 12: this is `NTP_SRC_FOLDER`?

tests/ci/integration/run_ntp_integration.sh
```diff
@@ -0,0 +1,66 @@
...
+#      - AWS_LC_BUILD_FOLDER
+#      - AWS_LC_INSTALL_FOLDER
+#      - NTP_BUILD_FOLDER
+
+# Assumes script is executed from the root of aws-lc directory
+SCRATCH_FOLDER="${SRC_ROOT}/NTP_BUILD_ROOT"
+NTP_VERSION="ntp-4.2.8p17"
```
WillChilds-Klein, line 19: perhaps this is too brittle, but you could try to auto-detect the latest version and fall back to this one if there's no match

```
$ set -o pipefail
$ curl -s https://www.ntp.org/downloads/ | grep -oP 'The current release of NTP is <a href=".+">([0-9a-z\.]+)' | cut -d'>' -f2
4.2.8p17
```

WillChilds-Klein (approved):

justsmth (approved):

tests/ci/integration/run_ntp_integration.sh
andrewhop: Let's see how stable it is 👍  I went with a slightly different approach to get both the folder and specific version.

justsmth (approved):

WillChilds-Klein (approved):

PR #1373, opened 2023-12-22
Implement AES-CCM

geedo0:
### Issues:
Resolves: CryptoAlg-2085

### Description of changes: 

This exposes the AES-CCM block cipher algorithm via the EVP_CIPHER API. Currently we only support a limited set of AES-CCM modes via the EVP_AEAD interface for the Bluetooth, Bluetooth-8, and Matter wireless protocols. This PR introduces support for all AES key sizes (128, 192, and 256) as well as the ability to configure arbitrary AES-CCM modes via the M and L parameters.

### Call-outs:
- AES-CCM in OpenSSL has a number of idiosyncrasies. I chose to prefer the behavior of upstream OpenSSL rather than imposing "more consistent" behavior relative to our other AEAD algorithms like AES-GCM and ChaCha20-Poly1305. This approach will improve our OpenSSL compatibility rather than introduce issues in potential consumers in the future.
  + AES-CCM only works in one-shot mode. i.e. incremental updates are not allowed. This is documented in the [wiki](https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption#Authenticated_Decryption_using_CCM_mode) and inferred from the [source](https://github.com/openssl/openssl/blob/1567a821a4616f59748fa8982724f88e542867d6/crypto/evp/e_aes.c#L3633)
  + Tag validation happens in EVP_Decrypt_Update rather than EVP_Decrypt_Final. This is in-line with the [OpenSSL wiki](https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption#Authenticated_Decryption_using_CCM_mode), [OpenSSL source](https://github.com/openssl/openssl/blob/1567a821a4616f59748fa8982724f88e542867d6/crypto/evp/e_aes.c#L3641), and [strongSwan consuming source](https://github.com/strongswan/strongswan/blob/c8ef91c786863f58bb07433b3967a2c4b41cb64b/src/libstrongswan/plugins/openssl/openssl_aead.c#L129).
  + The NULL-ness of the `in` and `out` parameters going into EVP_Cipher_Update is very important. The unit tests have been updated to ensure that non-NULL pointers are passed in regardless of whether or not the length is 0.
- We are introducing new to LC CCM operating modes and key sizes for AES-CCM that were not exposed via the EVP_AEAD mode. Please review that the cryptography for these are sound.

### Testing:

- All the NIST CAVP test vectors have been imported from upstream [OpenSSL](https://github.com/openssl/openssl/blob/master/test/recipes/30-test_evp_data/evpciph_aes_ccm_cavs.txt) and run into the EVP_CIPHER test suite.
- All the existing EVP_AEAD test vectors for Bluetooth, Bluetooth-8, and Matter have been run into the EVP_AEAD test vector runner.
- All the AES-CCM Wycheproof test vectors were generated using `convert_wycheproof.go` and run through the test runner developed for ChaCha20-Poly1305.
- All CI is passing aside from `aws-lc-ci-ec2-test-framework` which appears universal to all active PRs.


crypto/cipher_extra/cipher_test.cc
```diff
@@ -256,15 +273,20 @@ static void TestCipherAPI(const EVP_CIPHER *cipher, Operation op, bool padding,
       EVP_CIPHER_CTX_encrypting(ctx.get())) {
     max_out += block_size - (max_out % block_size);
   }
-  std::vector<uint8_t> result(max_out);
+  std::vector<uint8_t> result;
+  max_out ? result.resize(max_out) : result.reserve(1);
```
WillChilds-Klein, line 277: why `reserve(1)` here if `max_out == 0`?

geedo0: This forces an allocation of the vector's underlying array so when you do `result.data()` you get a non-NULL pointer even if `result.size() == 0`. This is important because the cipher interface is really sensitive to the NULL-ness of in/out to activate different steps of the algorithm.

crypto/fipsmodule/cipher/e_aesccm.c
```diff
@@ -96,8 +133,8 @@ static int ccm128_init_state(const struct ccm128_context *ctx,
                              const uint8_t *aad, size_t aad_len,
                              size_t plaintext_len) {
   const block128_f block = ctx->block;
-  const unsigned M = ctx->M;
-  const unsigned L = ctx->L;
+  const uint32_t M = ctx->M;
+  const uint32_t L = ctx->L;
```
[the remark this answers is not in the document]
geedo0: Just converting the ambiguous types to the corresponding explicit integers. Reclaiming a few bytes of stack space was not a goal, unless you feel strongly about it.

crypto/fipsmodule/cipher/e_aesccm.c
```diff
@@ -456,3 +491,239 @@ DEFINE_METHOD_FUNCTION(EVP_AEAD, EVP_aead_aes_128_ccm_matter) {
...
+    ctr128_f ctr = aes_ctr_set_key(&cipher_ctx->ks.ks, NULL, &block, key,
+                                   ctx->key_len);
+    CRYPTO_ccm128_init(&cipher_ctx->ccm, block, ctr, cipher_ctx->M,
+                       cipher_ctx->L);
+    cipher_ctx->key_set = 1;
+  }
+  if (iv) {
+    CRYPTO_ccm128_init(&cipher_ctx->ccm, NULL, NULL, cipher_ctx->M,
+                       cipher_ctx->L);
+    OPENSSL_memcpy(cipher_ctx->nonce, iv, 15 - cipher_ctx->L);
```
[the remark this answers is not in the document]
geedo0: The nonce length is `15 - L` as defined by the [RFC](https://datatracker.ietf.org/doc/html/rfc3610#section-2.1). Can add a comment/macro to make this clearer.

crypto/fipsmodule/cipher/e_aesccm.c
```diff
@@ -456,3 +491,239 @@ DEFINE_METHOD_FUNCTION(EVP_AEAD, EVP_aead_aes_128_ccm_matter) {
...
+        return 0;
+      }
+      if ((size_t) arg != cipher_ctx->M) {
+        return 0;
+      }
+      OPENSSL_memcpy(ptr, cipher_ctx->tag, cipher_ctx->M);
+      cipher_ctx->tag_set = 0;
+      cipher_ctx->iv_set = 0;
+      cipher_ctx->len_set = 0;
+      cipher_ctx->ccm_set = 0;
```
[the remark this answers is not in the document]
geedo0: The simple answer, it's what [OpenSSL does](https://github.com/openssl/openssl/blob/master/crypto/evp/e_aes.c#L3461) and I was preserving its oddities. The why of it, I'm not entirely sure. They seem to want the tag to only be read once and the cipher to be "reset" once that happens.

include/openssl/cipher.h
```diff
@@ -582,9 +586,11 @@ OPENSSL_EXPORT void EVP_CIPHER_CTX_set_flags(const EVP_CIPHER_CTX *ctx,
 #define EVP_CTRL_AEAD_SET_TAG 0x11
 #define EVP_CTRL_AEAD_SET_IV_FIXED 0x12
 #define EVP_CTRL_GCM_IV_GEN 0x13
+#define EVP_CTRL_CCM_SET_L 0x14
 #define EVP_CTRL_AEAD_SET_MAC_KEY 0x17
 // EVP_CTRL_GCM_SET_IV_INV sets the GCM invocation field, decrypt only
 #define EVP_CTRL_GCM_SET_IV_INV 0x18
+#define EVP_CTRL_GET_IVLEN 0x25
```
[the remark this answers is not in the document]
geedo0: > do we need to expose EVP_CTRL_CCM_SET_L to callers?

Yes, this is a CTRL value exposed by upstream OpenSSL and is [in use in the world](https://github.com/search?q=EVP_CTRL_CCM_SET_L&type=code).

> also, where do the values for these constants come from?

In isolation, they can be anything as long as they are unique within this header file. The convention we're following is to borrow them from [OpenSSL](https://github.com/openssl/openssl/blob/master/include/openssl/evp.h#L367) (along with the very unfortunate counting decimal numbers with hex literals).

crypto/fipsmodule/cipher/e_aesccm.c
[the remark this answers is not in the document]
WillChilds-Klein: 12 bytes? no, i don't feel too strongly about it.

crypto/fipsmodule/cipher/e_aesccm.c
```diff
@@ -456,3 +494,239 @@ DEFINE_METHOD_FUNCTION(EVP_AEAD, EVP_aead_aes_128_ccm_matter) {
...
+      cipher_ctx->M = 14;
+      cipher_ctx->message_len = 0;
+      return 1;
+    case EVP_CTRL_GET_IVLEN:
+      *(uint32_t *)ptr = CCM_L_TO_NONCE_LEN(cipher_ctx->L);
+      return 1;
+    case EVP_CTRL_AEAD_SET_IVLEN:
+      // The nonce (IV) length is 15-L, compute L here and set it below to "set"
+      // the IV length.
+      return cipher_aes_ccm_ctrl_set_L(cipher_ctx, 15 - arg);
```
WillChilds-Klein, line 645:
```suggestion
      return cipher_aes_ccm_ctrl_set_L(cipher_ctx, CCM_L_TO_NONCE_LEN(arg));
```

crypto/fipsmodule/cipher/e_aesccm.c
```diff
@@ -456,3 +494,239 @@ DEFINE_METHOD_FUNCTION(EVP_AEAD, EVP_aead_aes_128_ccm_matter) {
...
+    cipher_ctx->tag_set = 1;
+  } else {
+    // Decryption path. Compute the plaintext then compute its CBC-MAC.
+    // n.b. The method says encrypt, but it works both ways.
+    if (!ccm128_encrypt(ccm_ctx, ccm_state, &cipher_ctx->ks.ks, out, in, len)) {
+      return -1;
+    }
+    uint8_t computed_tag[EVP_AEAD_AES_CCM_MAX_TAG_LEN] = {0};
+    if (!ccm128_compute_mac(ccm_ctx, ccm_state, &cipher_ctx->ks.ks,
+                            computed_tag, cipher_ctx->M, out, len)) {
```
WillChilds-Klein, line 598: should we `OPENSSL_cleanse(out, len)` before returning here as well?

crypto/fipsmodule/cipher/e_aesccm.c
geedo0: Super pedantic, but this was an intentional choice. Though this works algebraically, this is not technically correct. I named the macro `CCM_L_TO_NONCE_LEN` and this particular line calls for converting a nonce length to L.

crypto/fipsmodule/cipher/e_aesccm.c
geedo0: Yeah, that makes sense.

WillChilds-Klein (approved):

crypto/fipsmodule/cipher/e_aesccm.c
```diff
@@ -456,3 +496,241 @@ DEFINE_METHOD_FUNCTION(EVP_AEAD, EVP_aead_aes_128_ccm_matter) {
...
+
+  if (!cipher_ctx->iv_set || !cipher_ctx->key_set) {
+    return -1;
+  }
+
+  if (!out) {
+    if (!in) {
+      // If |out| and |in| are both NULL, |len| is the total length of the
+      // message which we need to include that in the 0th block of the CBC-MAC.
+      cipher_ctx->message_len = len;
```
torben-hansen, line 543: This risk implicit truncation, no? `len` has higher width than `cipher_text->message_len`.

crypto/fipsmodule/cipher/e_aesccm.c
```diff
@@ -58,28 +58,70 @@
...
+  union {
+    uint64_t align;
+    AES_KEY ks;
+  } ks; // AES key schedule to use
+
+  CCM128_CTX ccm;
+  CCM128_STATE ccm_state;
+
+  // Boolean flags
+  uint32_t key_set;
```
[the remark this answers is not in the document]
WillChilds-Klein: not super important, but similar to [this](https://github.com/aws/aws-lc/pull/1373#discussion_r1445071820). if it's internal, why use more stack space than we need to?

crypto/fipsmodule/cipher/e_aesccm.c
geedo0: That's a great catch! Not only is this an implicit truncation that depends on your architecture, this is actually out of specification. We need to make sure that message_len (and also the AAD len) have a 64-bit integer width.

```
# Here's our range for message length
> The message m, consisting of a string of l(m) octets where 0 <= l(m) < 2^(8*L)

# Here's our range for L
> Valid values of L range between 2 octets and 8 octets

# Subbing that in, we get a max m(l) of
2^(8*8) = 2^64 = 16 EXABYTES

# AAD has a similar max length
> Additional authenticated data a, consisting of a string of l(a) octets where 0 <= l(a) < 2^64
```

crypto/fipsmodule/cipher/e_aesccm.c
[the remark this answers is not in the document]
torben-hansen: 8-bit is okay!

crypto/cipher_extra/cipher_test.cc
```diff
@@ -120,8 +120,12 @@ static const EVP_CIPHER *GetCipher(const std::string &name) {
     return EVP_aes_256_gcm();
   } else if (name == "AES-256-OFB") {
     return EVP_aes_256_ofb();
-  } else if (name == "CHACHA20-POLY1305") {
-    return EVP_chacha20_poly1305();
+  } else if (name == "AES-128-CCM") {
+    return EVP_aes_128_ccm();
+  } else if (name == "AES-192-CCM") {
+    return EVP_aes_192_ccm();
```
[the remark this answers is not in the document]
torben-hansen: Dang. Okay then :(

crypto/fipsmodule/cipher/e_aesccm.c
geedo0: Looked into this more. We're not able to support the maximum message length on 32-bit machines because EVP_CIPHER's [cipher](https://github.com/awslabs/aws-lc/blob/aec07944cd70a425723461edc1543065d442b6d9/crypto/fipsmodule/cipher/internal.h#L182) interface uses size_t for the len parameter. We have logic to truncate the algorithm's maximum length to whatever fits inside a `size_t` [here](https://github.com/awslabs/aws-lc/blob/da93b7ed5aa049c691231622594400713a9dd708/crypto/fipsmodule/cipher/e_aesccm.c#L130). You're right that there's absolutely an implicit truncation bug here. For CCM modes with L>4, running on a 64-bit machine, and messages >= 2^32, we would've done something weird. The fix here is to simply change the message_len type to size_t which will match whatever len value we get from the call.

WillChilds-Klein (approved):

crypto/fipsmodule/cipher/e_aesccm.c
```diff
@@ -456,3 +496,245 @@ DEFINE_METHOD_FUNCTION(EVP_AEAD, EVP_aead_aes_128_ccm_matter) {
...
+    }
+    cipher_ctx->iv_set = 0;
+    cipher_ctx->tag_set = 0;
+    cipher_ctx->len_set = 0;
+    cipher_ctx->ccm_set = 0;
+  }
+  return (int) len;
+}
+
+static void cipher_aes_ccm_cleanup(EVP_CIPHER_CTX *ctx) {}
```
torben-hansen, line 621: Redundant function

crypto/fipsmodule/cipher/e_aesccm.c
```diff
@@ -112,8 +154,8 @@ static int ccm128_init_state(const struct ccm128_context *ctx,
     state->nonce.c[0] |= 0x40;  // Set AAD Flag
   }
   OPENSSL_memcpy(&state->nonce.c[1], nonce, nonce_len);
-  for (unsigned i = 0; i < L; i++) {
-    state->nonce.c[15 - i] = (uint8_t)(plaintext_len >> (8 * i));
+  for (uint32_t i = 0; i < L; i++) {
+    state->nonce.c[15 - i] = (uint8_t)((uint64_t) plaintext_len >> (8 * i));
```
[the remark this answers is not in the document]
geedo0: I spent a ridiculous amount of time figuring out that this explicit `uint64_t` cast is required. If `size_t` is 32-bits you end up shifting out of bounds like you said and that is undefined in the C standard. I observed that this somehow turns into a rotation on some platforms and you end up with the LSB shifted into the 5th byte position 🤷 . Since this routine is building up the 0th block of the MAC by encoding the message length into it, we need to properly encode a 0 for all the higher order bit positions or else the tag calculation gets hosed.

Here's an example I pulled from my notes on what the bad block looked like:

```
# After init state on a 32-bit machine
# This is wrong, we have 0x18 in the 5th byte position for some reason
(gdb) x/16b state
0xd40bfe4:  0x07  0x5a  0x8a  0xa4  0x85  0xc3  0x16  0xe9
0xd40bfec:  0x00  0x00  0x00  0x18  0x00  0x00  0x00  0x18

# Should be
(lldb) x/16b ccm_state
0x7ff24e871b08: 0x07 0x5a 0x8a 0xa4 0x85 0xc3 0x16 0xe9
0x7ff24e871b10: 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x18
```

crypto/fipsmodule/cipher/e_aesccm.c
geedo0: By convention, we've been defining these empty functions just to fill the interface. I can see if the EVP code properly handles a NULL function pointer for it.

geedo0: it's gone 👌

torben-hansen (approved):

WillChilds-Klein (approved):

PR #1384, opened 2024-01-05
Add field-based accessors to SHA_CTX for compatibility

geedo0:
### Issues:
Addresses CryptoAlg-2081

### Description of changes: 

Upstream OpenSSL defines the h_n values as individual fields rather than
an array which BoringSSL introduced. This will improve our compatibility with
code that typically builds against OpenSSL. For example, this is breaking the
strongSwan build when compiling against AWS-LC.

Original BoringSSL source [here](https://github.com/google/boringssl/blob/master/include/openssl/sha.h#L116)

### Call-outs:

These fields were removed from AWS-LC early on in the project by PR #58.
Seemed more incidental rather than intentional, but I don't have the full
context here.

### Testing:

### DIFF
--- a/include/openssl/sha.h
+++ b/include/openssl/sha.h
@@ -97,7 +97,23 @@ OPENSSL_EXPORT void SHA1_Transform(SHA_CTX *sha,
                                    const uint8_t block[SHA_CBLOCK]);
 
 struct sha_state_st {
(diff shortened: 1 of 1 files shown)


geedo0:
This is not going to work due to the C99 requirement.

PR #1397, opened 2024-01-11
Use system allocator for thread-local values

justsmth:
### Issues:
P111627750

### Description of changes: 
* Depending on order destructors are called during process exit, custom allocators might no longer be available.  
* This commit uses the system allocator for all of our thread local values.

### Call-outs:
* Thread locals relating to [error](https://github.com/aws/aws-lc/blob/02e61afe8c9d6cd68ed3fa664ffd0a55312ad35b/crypto/err/err.c#L220) were already using the system allocator.
*  Likewise, the system allocator is used for our [pthread logic managing thread locals](https://github.com/aws/aws-lc/blob/e6e201b9886021eb54f32e3030c438c8e7f1570d/crypto/thread_pthread.c#L157).

### Testing:
* Tested locally.


crypto/fipsmodule/rand/rand.c
```diff
@@ -12,6 +12,8 @@
  * OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
  * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE. */
 
+// Ensure we can't call OPENSSL_malloc.
+#define _BORINGSSL_PROHIBIT_OPENSSL_MALLOC
```
WillChilds-Klein, line 16: should we also define this in err.c and other places the system allocator is used?

WillChilds-Klein (commented):
> Depending on order destructors are called during process exit, custom allocators might no longer be available.

do we know which dtor invalidates the OpenSSL allocator? i'm wondering if, knowing which dtor needs to be called last, there's some way to enforce that that dtor is always called last.

crypto/fipsmodule/rand/rand.c
```diff
@@ -264,7 +266,7 @@ static void rand_thread_state_free(void *state_in) {
   rand_state_fips_clear(state);
 #endif
 
-  OPENSSL_free(state);
+  free(state);
```
torben-hansen, line 269: Should definitely zeroise this memory before dropping the reference to it.

crypto/fipsmodule/rand/rand.c
justsmth: Yeah, it's already defined in [those](https://github.com/aws/aws-lc/blob/02e61afe8c9d6cd68ed3fa664ffd0a55312ad35b/crypto/err/err.c#L110) other [places](https://github.com/aws/aws-lc/blob/e6e201b9886021eb54f32e3030c438c8e7f1570d/crypto/thread_pthread.c#L16).

crypto/fipsmodule/rand/rand.c
```diff
@@ -262,9 +264,11 @@ static void rand_thread_state_free(void *state_in) {
   CRYPTO_STATIC_MUTEX_unlock_write(thread_states_list_lock_bss_get());
 
   rand_state_fips_clear(state);
+#else
+  OPENSSL_memset(state, 0, sizeof(struct rand_thread_state));
```
andrewhop, line 268: Cleanse has additional safety features to avoid the compiler optimizing out the unused result. 
```suggestion
  OPENSSL_cleanse(state, 0, sizeof(struct rand_thread_state));
```

justsmth:
> > Depending on order destructors are called during process exit, custom allocators might no longer be available.
> 
> do we know which dtor invalidates the OpenSSL allocator? i'm wondering if, knowing which dtor needs to be called last, there's some way to enforce that that dtor is always called last.

The concern is for consumers that are using [`CRYPTO_set_mem_functions`](https://github.com/aws/aws-lc/blob/84c5dee7f00efa567d67ca36b466ebf18d7f0234/crypto/mem.c#L140) to inject a custom allocator. At process exit, it's possible that our destructors are being invoked after the consumer's custom allocator (whatever that may be) has been destructed.

crypto/fipsmodule/rand/rand.c
```diff
@@ -262,9 +264,11 @@ static void rand_thread_state_free(void *state_in) {
   CRYPTO_STATIC_MUTEX_unlock_write(thread_states_list_lock_bss_get());
 
   rand_state_fips_clear(state);
+#else
+  OPENSSL_cleanse(state, sizeof(struct rand_thread_state));
```
torben-hansen, line 268: This could really be called for both fips and non-fips code path. But it's fine as-is now, since the drbg state is zeroised. in both paths.

torben-hansen (approved):

andrewhop (approved):

PR #1398, opened 2024-01-11
Update EVP cipher APIs to gracefully handle null EVP_CIPHER_CTX

andrewhop:
### Issues:
Resolves V1187459157

### Description of changes: 
This change does 3 things:
1. All EVP_CIPHER_CTX now start off as poisoned and require calling EVP_CipherInit_ex to cure it
2. All EVP Encrypt/Decrypt Update/Final now check that the EVP_CIPHER_CTX and EVP_CIPHER_CTX->cipher are not null
3. Add the start of reusable safety macros inspired by [s2n-tls](https://github.com/aws/s2n-tls/blob/main/docs/SAFETY-MACROS.md)

### Call-outs:
The __AWS_LC_ENSURE macro uses the do {} while (0) trick to ensure the action is run once, anything passed into the macro doesn't accidentally expand and change the scope, and the compiler enforces you add a `;` after the macro. We use this trick in other macros and all compilers are smart enough to optimize out the jump.

### Testing:
`GUARD_PTR(ctx);` expands to:
```
  do {
    if (!((ctx) != ((void *)0))) {
      ERR_put_error(ERR_LIB_CRYPTO, 0, (3 | 64), "_file_name_", 259);
      return 0;
    }
  } while (0);
```

[Here is an example](https://godbolt.org/z/z99roroKq) with the macro, and here it is implemented as a [traditional if/else](https://godbolt.org/z/E56EYnnW9). Both result in the same code.


crypto/internal.h
```diff
@@ -1149,6 +1149,28 @@ OPENSSL_EXPORT int OPENSSL_vasprintf_internal(char **str, const char *format,
...
+    do {                           \
+        if (!(cond)) {             \
+            action;                \
+        }                          \
+    } while (0)
+
+#define AWS_LC_ERROR 0
+#define AWS_LC_SUCCESS 1
+
+// RESULT_GUARD_PTR checks if |ptr|, if it is null it adds ERR_R_PASSED_NULL_PARAMETER
```
torben-hansen, line 1168: wording

crypto/internal.h
```diff
@@ -1149,6 +1149,28 @@ OPENSSL_EXPORT int OPENSSL_vasprintf_internal(char **str, const char *format,
                                               va_list args, int system_malloc)
     OPENSSL_PRINTF_FORMAT_FUNC(2, 0);
 
+
+// Safety Macros
+
+// Inspired by s2n-tls
+
+// __AWS_LC_ENSURE checks if |cond| is true nothing happens, else |action| is called
+#define __AWS_LC_ENSURE(cond, action) \
```
torben-hansen, line 1158: I ain't the super fan of these macro's. It quickly inflate into tens of macro's to capture different return types/values and different error behaviour (e.g. goto's).

mark as experimental?

torben-hansen (approved):

crypto/internal.h
nebeid, line 1168:
```suggestion
// RESULT_GUARD_PTR checks |ptr|, if it is null it adds ERR_R_PASSED_NULL_PARAMETER
```

crypto/fipsmodule/cipher/cipher.c
```diff
@@ -255,6 +256,7 @@ static int block_remainder(const EVP_CIPHER_CTX *ctx, int len) {
 
 int EVP_EncryptUpdate(EVP_CIPHER_CTX *ctx, uint8_t *out, int *out_len,
                       const uint8_t *in, int in_len) {
+  GUARD_PTR(ctx);
```
nebeid, line 259: Why there are no guards for CipherInit_ex, CipherUpdate and CipherFinal_ex? `ctx` is accessed there right away.

crypto/internal.h
andrewhop: It's in an internal header so we shouldn't have to worry about external customers using it, but I'll mark the section as experimental in case they ever get moved and we forget.

justsmth (dismissed):

crypto/fipsmodule/cipher/cipher.c
```diff
@@ -70,6 +70,7 @@
 
 void EVP_CIPHER_CTX_init(EVP_CIPHER_CTX *ctx) {
   OPENSSL_memset(ctx, 0, sizeof(EVP_CIPHER_CTX));
+  ctx->poisoned = 1;
```
justsmth, line 73: `ctx->poisoned` also needs to be set in `EVP_CIPHER_CTX_new` and `EVP_CIPHER_CTX_cleanup` right below this.

justsmth (changes requested):

justsmth (dismissed):

justsmth (approved):

nebeid (approved):

PR #1407, opened 2024-01-17
Transfer serde

skmcgrail:
### Description of changes: 
Increase test coverage for the TLS transfer feature.


torben-hansen (dismissed):

ssl/ssl_test.cc
```diff
@@ -11999,5 +11999,24 @@ TEST(SSLTest, SSLFileTests) {
...
+  static constexpr size_t INCOMPATIBLE_DER_LEN = 7;
+  static const uint8_t INCOMPATIBLE_DER[INCOMPATIBLE_DER_LEN] = {
+      0x30, 0x05, 0x30, 0x03, 0x02, 0x01, 0x2a};
+
+  bssl::UniquePtr<SSL_CTX> ctx(SSL_CTX_new(TLS_method()));
+  ASSERT_TRUE(ctx);
+
+  ASSERT_FALSE(
+      SSL_from_bytes(INCOMPATIBLE_DER, INCOMPATIBLE_DER_LEN, ctx.get()));
+}
```
andrewhop, line 12019: For V1177861396 is it possible to verify this test is failing due to the version mismatch and not something like INCOMPATIBLE_DER doesn't have enough data to read the state from?

crypto/cipher_extra/aead_test.cc
```diff
@@ -1377,3 +1359,182 @@ TEST(AEADTest, TestGCMSIV256Change16Alignment) {
...
+                                                 params.tag_len, direction));
+
+    CBS cbs;
+    CBS_init(&cbs, INVALID_CIPHER_ID_DER, INVALID_CIPHER_ID_DER_LEN);
+    ASSERT_FALSE(EVP_AEAD_CTX_deserialize_state(ctx.get(), &cbs));
+  }
+}
+
+TEST(EvpAeadCtxSerdeTest, ID) {
+  bool identifiers[AEAD_MAX_ID + 1] = {false};
```
andrewhop, line 1526: What does this identifier do/track? I don't quite understand the logic of when this flips to true/false.

crypto/cipher_extra/aead_test.cc
```diff
@@ -282,24 +282,6 @@ TEST_P(PerAEADTest, TestVector) {
...
-  std::string key_hex("809adab149635b22f0a37de146f17dc8");
-  std::vector<uint8_t> key;
-
-  ASSERT_TRUE(DecodeHex(&key, key_hex));
-  ASSERT_TRUE(EVP_AEAD_CTX_init(ctx.get(), EVP_aead_aes_128_gcm(), key.data(),
-                                key.size(), 16, nullptr));
-
-  ASSERT_EQ(EVP_AEAD_CTX_get_aead_id(ctx.get()), AEAD_AES_128_GCM_ID);
-}
-
```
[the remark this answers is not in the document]
skmcgrail: Not accidental, I just moved the testing to [here](https://github.com/aws/aws-lc/pull/1407/files#diff-ddef2a8fa601d9e2fc6e21fb2baffa581e1a85d55587cda87339dd964b0a3363R1532) as it made more logical sense to be co-located in that test.

crypto/fipsmodule/cipher/internal.h
```diff
@@ -106,6 +106,7 @@ extern "C" {
 #define AEAD_AES_128_CCM_BLUETOOTH_ID 25
 #define AEAD_AES_128_CCM_BLUETOOTH_8_ID 26
 #define AEAD_AES_128_CCM_MATTER_ID 27
+#define AEAD_MAX_ID 27
```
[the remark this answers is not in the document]
skmcgrail: There is no need to enforce a maximum, this is just a misunderstanding of the code. The code explicitly checks [here](https://github.com/aws/aws-lc/blob/5c358103c5df836b9343bf995717b5bc13d5e82f/crypto/fipsmodule/cipher/aead.c#L383-L384) the state information that is being decoded from the bytes has a matching identifier with the `EVP_AEAD_CTX` that the state is being restored into.

Additionally for the unknown identifier case, if we were to ever miss adding an identifier (and matching test-case entry) we would outright fail the serialization call all together intentionally, see [here](https://github.com/aws/aws-lc/blob/5c358103c5df836b9343bf995717b5bc13d5e82f/crypto/fipsmodule/cipher/aead.c#L313-L319)

ssl/ssl_test.cc
skmcgrail: I'll have to assign explicit error codes for this to do so, I've verified in a debugger that this is the reason it is failing. But let me explore just mapping new error types in.

crypto/cipher_extra/aead_test.cc
skmcgrail: What this is doing is creating an array of boolean values equal to the number of AEAD cipher identifiers (determined by the AEAD_MAX_ID which is the highest value currently), and then add one so that we can use `identifiers[0]` to be the unknown identifier which should never be used by an AEAD cipher. We then loop over all the test cases and record `identifiers[EVP_AEAD_CTX_get_aead_id()] = true` If an identifier is already been seen then this test case will fail as that is unexpected. At the end we should expect that we tested all the AEAD cipher identifiers (`identifiers[1] ... identifiers[AEAD_MAX_ID]` must all be `true`), and that `identifiers[0] == false`.

This ensures that:
1. All identifiers have a uniquely assigned identifier number
2. No AEAD algorithm returns the default value `0` aka the `AEAD_UNKNOWN_ID` value.
3. As new AEAD identifiers are added (and thus AEAD_MAX_ID is updated) that they are added to the test matrix here and are not missed in testing.

torben-hansen (dismissed):

andrewhop (approved):

andrewhop (approved):

torben-hansen (approved):

PR #1409, opened 2024-01-18
More test coverage for AWSLC_thread_local_clear

justsmth:
### Issues:
N/A

### Description of changes: 
* Adds unit test coverage for `AWSLC_thread_local_clear`

### Call-outs:
* This function also gets tested by `dynamic_loading_test.c`, but that does not run as part of the `crypto_test` target.

### Testing:
Tested in dev environment on my Mac.

### DIFF
--- a/crypto/thread_test.cc
+++ b/crypto/thread_test.cc
@@ -152,6 +152,8 @@ TEST(ThreadTest, RandState) {
   thread.join();
 }
 
+#if defined(OPENSSL_PTHREADS)
+
 static void thread_task(bool *myFlag) {
   EXPECT_EQ(1, AWSLC_thread_local_clear());
   EXPECT_EQ(1, AWSLC_thread_local_clear());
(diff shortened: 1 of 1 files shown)


crypto/thread_test.cc
```diff
@@ -152,6 +152,34 @@ TEST(ThreadTest, RandState) {
...
 
+static void thread_task(bool *myFlag) {
+  EXPECT_EQ(1, AWSLC_thread_local_clear());
+  EXPECT_EQ(1, AWSLC_thread_local_clear());
+  uint8_t buf[8];
+  EXPECT_EQ(1, RAND_bytes(buf, sizeof(buf)));
+  EXPECT_EQ(1, AWSLC_thread_local_clear());
+  ERR_clear_error();
+  EXPECT_EQ(1, AWSLC_thread_local_clear());
+  EXPECT_EQ(1, AWSLC_thread_local_clear());
```
WillChilds-Klein, line 163: why call `AWSLC_thread_local_clear` twice here?

WillChilds-Klein (approved):

crypto/thread_test.cc
justsmth: The concern is that `AWSLC_thread_local_clear` could leave the thread-local data in an invalid state, causing an (e.g.) seg-fault if the function is called a second time.

andrewhop (approved):

WillChilds-Klein (approved):

PR #1415, opened 2024-01-22
Zero the stack after AVX-512 XTS

pittma:
### Description of changes: 
A small patch to zero out the used stack frames in the AVX-512 XTS implementation.

---

### DIFF
--- a/crypto/fipsmodule/aes/asm/aesni-xts-avx512.pl
+++ b/crypto/fipsmodule/aes/asm/aesni-xts-avx512.pl
@@ -1832,29 +1832,65 @@ ___
   vmovdqu 	 %xmm8,-0x10($output)
 ___
   }
-  $code .= "\n.L_ret_${rndsuffix}:\n";
-  $code .= "mov 	 $GP_STORAGE($TW),%rbx\n";
+
+  {
+  $code .= <<___;
+  .L_ret_${rndsuffix}:
+  mov 	 $GP_STORAGE($TW),%rbx
+  xor    $tmp1,$tmp1
+  mov    $tmp1,$GP_STORAGE($TW)
+  # Zero-out the whole of `%zmm0`.
+  vpxorq %zmm0,%zmm0,%zmm0
+___
+  }
 
   if ($win64) {
-    $code .= "mov 	 $GP_STORAGE + 8*1($TW),%rdi\n";
-    $code .= "mov 	 $GP_STORAGE + 8*2($TW),%rsi\n";
+    $code .= <<___;
+    mov $GP_STORAGE + 8*1($TW),%rdi
+    mov $tmp1,$GP_STORAGE + 8*1($TW)
+    mov $GP_STORAGE + 8*2($TW),%rsi
+    mov $tmp1,$GP_STORAGE + 8*2($TW)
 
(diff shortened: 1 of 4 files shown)


nebeid:
Thank you, Dan @pittma.
I believe we could add:
- clearing the locations on the stack where the registers were backed up, taking into account the extra ones in the case of `win64`. This is done in `aesni-x86_64.pl`
- this change and the suggestion above to the decryption function.

pittma:
As far as I can tell, this CI failure is unrelated. It looks like `HEAD` moved a few hours ago for the tpm2 dependency in this test, and now there is a conflict with the patch:

```
commit 84ead1b7a35362d5fae9b26a0edb9a68eeb6e5c8 (HEAD -> master, origin/master, origin/HEAD)
Author: Juergen Repp <juergen_repp@web.de>
Date:   Tue Jan 23 12:00:45 2024 +0100

    FAPI: Clean command context before execution.
    
    The FAPI command context now is cleared in the async function of
    every command.
    
    Signed-off-by: Juergen Repp <juergen_repp@web.de>
```

This is the error I'm seeing:

```sh
$ git apply /home/dpitt/src/aws-lc/tests/ci/integration/tpm2_tss_patch/aws-lc-tpm2-tss.patch
error: patch failed: test/unit/fapi-eventlog.c:138
error: test/unit/fapi-eventlog.c: patch does not apply
```

torben-hansen:
> As far as I can tell, this CI failure is unrelated. It looks like `HEAD` moved a few hours ago for the tpm2 dependency in this test, and now there is a conflict with the patch:
> 
> ```
> commit 84ead1b7a35362d5fae9b26a0edb9a68eeb6e5c8 (HEAD -> master, origin/master, origin/HEAD)
> Author: Juergen Repp <juergen_repp@web.de>
> Date:   Tue Jan 23 12:00:45 2024 +0100
> 
>     FAPI: Clean command context before execution.
>     
>     The FAPI command context now is cleared in the async function of
>     every command.
>     
>     Signed-off-by: Juergen Repp <juergen_repp@web.de>
> ```
> 
> This is the error I'm seeing:
> 
> ```shell
> $ git apply /home/dpitt/src/aws-lc/tests/ci/integration/tpm2_tss_patch/aws-lc-tpm2-tss.patch
> error: patch failed: test/unit/fapi-eventlog.c:138
> error: test/unit/fapi-eventlog.c: patch does not apply
> ```

Resolved here: https://github.com/aws/aws-lc/pull/1422

nebeid:
I think the comments that used to be there before line 1881 for the special case of zeroising were helpful:
```
# Stack usage isn't divisible by 64, so we use a kmask register to
# only use 48 of the bytes (6 quad-words).
```
The prior comments were good too.

pittma:
> I think the comments that used to be there before line 1881 for the special case of zeroising were helpful:
> 
> ```
> # Stack usage isn't divisible by 64, so we use a kmask register to
> # only use 48 of the bytes (6 quad-words).
> ```
> 
> The prior comments were good too.

Thanks for saying so @nebeid. I'd mostly left them as mnemonics for myself and I began to fear I was the only one! I've added them back.

nebeid (approved):

dkostic (approved):

PR #1419, opened 2024-01-23
Include FIPS mode in OpenSSL_version return value

WillChilds-Klein:
### Issues:
- n/a

### Description of changes: 

This change includes FIPS mode information in the return value of `OpenSSL_version(OPENSSL_VERSION)`. This will allow consumers of CPython (and likely other tools or language runtimes) to detect FIPS mode without calling to an AWS-LC-specific API, which would require either patching or forking the language or tool in question.

### Call-outs:

* we don't modify `OPENSSL_VERSION_TEXT` in `crypto.h` as it's deprecated

### Testing:
* CI testing
* built cpython 3.10 against this before/after:

```
# before change, built against AWS-LC w/ -DFIPS=1
$ ./python -c 'import ssl; print(ssl.OPENSSL_VERSION)'
AWS-LC 1.20.0

# after change, built against AWS-LC w/ -DFIPS=1
$ ./python -c 'import ssl; print(ssl.OPENSSL_VERSION)'
AWS-LC FIPS 1.20.0

# after change, built against AWS-LC w/ -DFIPS=0
$ ./python -c 'import ssl; print(ssl.OPENSSL_VERSION)'
AWS-LC 1.20.0
```

---

### DIFF
--- a/crypto/crypto.c
+++ b/crypto/crypto.c
@@ -115,11 +115,10 @@ void CRYPTO_pre_sandbox_init(void) {
 
 const char *SSLeay_version(int which) { return OpenSSL_version(which); }
 
-#define AWS_LC_VERSION_TEXT AWSLC_VERSION_NAME " " AWSLC_VERSION_NUMBER_STRING
 const char *OpenSSL_version(int which) {
   switch (which) {
     case OPENSSL_VERSION:
-      return AWS_LC_VERSION_TEXT;
+      return AWSLC_VERSION_STRING;
     case OPENSSL_CFLAGS:
       return "compiler: n/a";
     case OPENSSL_BUILT_ON:
--- a/crypto/crypto_test.cc
+++ b/crypto/crypto_test.cc
@@ -22,6 +22,7 @@
 #include <openssl/crypto.h>
 #include <openssl/cipher.h>
 #include <openssl/mem.h>
+#include <openssl/service_indicator.h>
 
 #include <gtest/gtest.h>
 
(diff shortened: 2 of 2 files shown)


andrewhop (commented):
How does the new return value compare to OpenSSL's? Is there any other information people might expect in it?

WillChilds-Klein:
> How does the new return value compare to OpenSSL's?

They include library name, some config info (including FIPSness), major/minor version, and release date. Source for 3.0 [here](https://github.com/openssl/openssl/blob/0f644b96d209443b4566f7e86e3be2568292e75b/include/openssl/opensslv.h.in#L90).

On AL2 that looks like:

```
$ uname -a && openssl version
Linux dev-dsk-childw-1d-8bded7fc.us-east-1.amazon.com 5.10.205-172.804.amzn2int.x86_64 #1 SMP Tue Jan 2 17:35:42 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
OpenSSL 1.0.2k-fips  26 Jan 2017
```

And on Ubuntu 22.04:

```
$ uname -a && openssl version
Linux ip-172-31-89-138 6.2.0-1014-aws #14~22.04.1-Ubuntu SMP Thu Oct  5 22:43:45 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
OpenSSL 3.0.2 15 Mar 2022 (Library: OpenSSL 3.0.2 15 Mar 2022)
```

> Is there any other information people might expect in it?

Perhaps release date, but users can reference our release history on github for a particular version if they need that info so I don't think it's really necessary.

andrewhop (approved):

samuel40791765 (approved):

PR #1423, opened 2024-01-26
Slight tweaks and integration CI to support Bind9

samuel40791765:
### Issues:
Addresses `CryptoAlg-2162`

### Description of changes: 
1. Integration CI dimension for Bind9
2. Resolved "cmocka unit tests" for Bind9
    * Additional `<openssl/asn1.h>` import in `<openssl/objects.h>`: Bind depends on [some ASN1 functions](https://gitlab.isc.org/isc-projects/bind9/-/blob/4330014fce5ab2ac1a6e58cd330762318a34f8d9/lib/dns/rdata.c#L625), but does not directly import the corresponding header. OpenSSL imports the [asn1 header file](https://github.com/openssl/openssl/blob/master/include/openssl/objects.h#L21) in `objects.h` (which Bind is pulling [these symbols from](https://gitlab.isc.org/isc-projects/bind9/-/blob/4330014fce5ab2ac1a6e58cd330762318a34f8d9/lib/dns/rdata.c#L21)), so I've added the header file reference to `objects.h`.
    * `SSL_get_error` error anticipation fixing: There were several failures discovered to be related this, thanks to research done in https://github.com/aws/aws-lc/pull/1333 and dumping out the logs from Bind9 built with OpenSSL and AWS-LC. The issue was pinned down the check implemented in https://github.com/google/boringssl/commit/9a38e924aafa96b4f1d7495ce140ba06b956854f. This check used to exist before the final return of `SSL_get_error` in OpenSSL. BoringSSL moved this earlier in the function with https://github.com/google/boringssl/commit/fcf25833bc4d057377da1ea755f57b2f5126e555. However, much of the functions guards for `i < 0` checks have been removed since OpenSSL 1.1.1, so the early logic no longer applies.
    This check has evolved into `SSL_ERROR_ZERO_RETURN` in our code. Moving the logic further down helps us gain better parity with OpenSSL 1.1.1. Doing so passes the bind test failures for `proxystream_test`, `tls_test`, and `doh_test`. This also happens to help our integration with CPython.
    * We actually already [use SSL_AUTO_RETRY by default](https://github.com/aws/aws-lc/commit/ec48af40a75bbeef2287af8bf3d867e8fb094692) in AWS-LC. The recent change mentioned in the point above surrounding the flag (https://github.com/aws/aws-lc/commit/208327e863e7efab0b8f1f3ce928f9055c5ae185) was just to make some of the errors consistent in CPython when the flag was used. I've reverted the special behavior surrounding it since it should no longer be needed.
    
    *  Assertion for `SSL_set_shutdown`: The assertion was added in https://github.com/aws/aws-lc/commit/63006a913b772d48a364915c3155d51744a8f704, where it’s stated that they didn’t want `SSL_set_shutdown` messing up the state machine. This assertion is causing failures in `tlsdns_test` for Bind9, so it appears that we'll have to remove this to gain better OpenSSL parity. This assertion is non-existent [in OpenSSL](https://github.com/openssl/openssl/blob/a552c23c6502592c1b3c67d93dd7e5ffbe958aa4/ssl/ssl_lib.c#L5181-L5190) today.
    
3. Patch file needed for Bind seems to be slight bug in their build configuration. This was from a fairly recent commit. We can look to contribute this sometime soon.

### Call-outs:
* We don't need `SSL_AUTO_RETRY` anymore to support CPython, so I've removed and regenerated the patch. This simplifies our support for CPython.
* There's still an additional test suite for Bind9 [(system tests)](https://gitlab.isc.org/isc-projects/bind9/-/tree/4330014fce5ab2ac1a6e58cd330762318a34f8d9/bin/tests/system#bind9-system-test-framework) that has failures we'll need to resolve. The plan is to investigate this as a subsequent PR.

### Testing:
New CI dimension 

### DIFF
--- a/ssl/ssl_lib.cc
+++ b/ssl/ssl_lib.cc
@@ -1439,8 +1439,7 @@ int SSL_get_error(const SSL *ssl, int ret_code) {
   }
 
   // An EOF was observed which violates the protocol, and the underlying
-  // transport does not participate in the error queue. If
-  // |SSL_MODE_AUTO_RETRY| is unset, bubble up to the caller.
+  // transport does not participate in the error queue.
   //
   // This was moved earlier in the function by BoringSSL and used to be
   // wrapped with a check for |ret_code| == 0 in OpenSSL 1.0.2. Since
--- a/ssl/ssl_lib.cc
+++ b/ssl/ssl_lib.cc
@@ -1382,23 +1382,6 @@ int SSL_get_error(const SSL *ssl, int ret_code) {
     return SSL_ERROR_SSL;
   }
 
-  if (ret_code == 0) {
-    if (ssl->s3->rwstate == SSL_ERROR_ZERO_RETURN) {
-      return SSL_ERROR_ZERO_RETURN;
-    }
-    // An EOF was observed which violates the protocol, and the underlying
-    // transport does not participate in the error queue. If
-    // |SSL_MODE_AUTO_RETRY| is unset, bubble up to the caller.
-    if ((ssl->ctx->mode & SSL_MODE_AUTO_RETRY) == 0) {
-      return SSL_ERROR_SYSCALL;
-    }
-    // If |SSL_MODE_AUTO_RETRY| is set, proceed if in a retryable state.
-    if (ssl->s3->rwstate != SSL_ERROR_WANT_READ
-            && ssl->s3->rwstate != SSL_ERROR_WANT_WRITE) {
-      return SSL_ERROR_SYSCALL;
-    }
-  }
-
   switch (ssl->s3->rwstate) {
     case SSL_ERROR_PENDING_SESSION:
     case SSL_ERROR_PENDING_CERTIFICATE:
@@ -1453,6 +1436,20 @@ int SSL_get_error(const SSL *ssl, int ret_code) {
 
       break;
     }
+  }
+
+  // An EOF was observed which violates the protocol, and the underlying
+  // transport does not participate in the error queue. If
+  // |SSL_MODE_AUTO_RETRY| is unset, bubble up to the caller.
+  //
+  // This was moved earlier in the function by BoringSSL and used to be
+  // wrapped with a check for |ret_code| == 0 in OpenSSL 1.0.2. Since
+  // OpenSSL 1.1.1, the check for |ret_code| has been removed and we've
+  // moved this back before the final return to gain better parity with
+  // OpenSSL.
+  // See openssl/openssl@8051ab2 for more details.
+  if (ssl->s3->rwstate == SSL_ERROR_ZERO_RETURN) {
+    return SSL_ERROR_ZERO_RETURN;
   }
 
   return SSL_ERROR_SYSCALL;
@@ -2611,10 +2608,6 @@ void SSL_set_quiet_shutdown(SSL *ssl, int mode) {
 int SSL_get_quiet_shutdown(const SSL *ssl) { return ssl->quiet_shutdown; }
 
 void SSL_set_shutdown(SSL *ssl, int mode) {
-  // It is an error to clear any bits that have already been set. (We can't try
-  // to get a second close_notify or send two.)
-  assert((SSL_get_shutdown(ssl) & mode) == SSL_get_shutdown(ssl));
-
   if (mode & SSL_RECEIVED_SHUTDOWN &&
       ssl->s3->read_shutdown == ssl_shutdown_none) {
     ssl->s3->read_shutdown = ssl_shutdown_close_notify;
--- a/ssl/ssl_lib.cc
+++ b/ssl/ssl_lib.cc
@@ -1394,6 +1394,7 @@ int SSL_get_error(const SSL *ssl, int ret_code) {
     case SSL_ERROR_WANT_CERTIFICATE_VERIFY:
     case SSL_ERROR_WANT_RENEGOTIATE:
     case SSL_ERROR_HANDSHAKE_HINTS_READY:
+    case SSL_ERROR_ZERO_RETURN:
       return ssl->s3->rwstate;
 
     case SSL_ERROR_WANT_READ: {
@@ -1436,19 +1437,6 @@ int SSL_get_error(const SSL *ssl, int ret_code) {
 
       break;
     }
-  }
-
-  // An EOF was observed which violates the protocol, and the underlying
-  // transport does not participate in the error queue.
-  //
-  // This was moved earlier in the function by BoringSSL and used to be
-  // wrapped with a check for |ret_code| == 0 in OpenSSL 1.0.2. Since
-  // OpenSSL 1.1.1, the check for |ret_code| has been removed and we've
-  // moved this back before the final return to gain better parity with
-  // OpenSSL.
-  // See openssl/openssl@8051ab2 for more details.
-  if (ssl->s3->rwstate == SSL_ERROR_ZERO_RETURN) {
-    return SSL_ERROR_ZERO_RETURN;
   }
 
   return SSL_ERROR_SYSCALL;
--- a/tests/ci/integration/bind9_patch/bind-fix.patch
+++ b/tests/ci/integration/bind9_patch/bind-fix.patch
@@ -1,0 +1,19 @@
+diff --git a/tests/isc/Makefile.am b/tests/isc/Makefile.am
+index 5cdd915..6ee1935 100644
+--- a/tests/isc/Makefile.am
++++ b/tests/isc/Makefile.am
+@@ -115,10 +115,12 @@ proxyheader_test_SOURCES =	\
+ 	proxyheader_test_data.h
+ 
+ proxystream_test_CPPFLAGS =	\
+-	$(AM_CPPFLAGS)
++	$(AM_CPPFLAGS)			\
++	$(OPENSSL_CFLAGS)
+ 
+ proxystream_test_LDADD =	\
+-	$(LDADD)
++	$(LDADD)				\
++	$(OPENSSL_LIBS)
+ 
+ proxystream_test_SOURCES =	\
+ 	proxystream_test.c \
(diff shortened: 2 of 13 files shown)


ssl/ssl_lib.cc
```diff
@@ -1455,6 +1438,19 @@ int SSL_get_error(const SSL *ssl, int ret_code) {
...
+  //
+  // This was moved earlier in the function by BoringSSL and used to be
+  // wrapped with a check for |ret_code| == 0 in OpenSSL 1.0.2. Since
+  // OpenSSL 1.1.1, the check for |ret_code| has been removed and we've
+  // moved this back before the final return to gain better parity with
+  // OpenSSL.
+  // See openssl/openssl@8051ab2 for more details.
+  if (ssl->s3->rwstate == SSL_ERROR_ZERO_RETURN) {
+    return SSL_ERROR_ZERO_RETURN;
+  }
```
WillChilds-Klein, line 1452: if we move this back up to where it was, do your bind9 tests still pass?

ssl/ssl_lib.cc
```diff
@@ -2611,10 +2607,6 @@ void SSL_set_quiet_shutdown(SSL *ssl, int mode) {
 int SSL_get_quiet_shutdown(const SSL *ssl) { return ssl->quiet_shutdown; }
 
 void SSL_set_shutdown(SSL *ssl, int mode) {
-  // It is an error to clear any bits that have already been set. (We can't try
-  // to get a second close_notify or send two.)
-  assert((SSL_get_shutdown(ssl) & mode) == SSL_get_shutdown(ssl));
-
```
WillChilds-Klein, line 2617: why remove this assertion?

tests/ci/integration/run_bind9_integration.sh
```diff
@@ -0,0 +1,60 @@
...
+SCRATCH_FOLDER="${SRC_ROOT}/BIND9_BUILD_ROOT"
+BIND9_SRC_FOLDER="${SCRATCH_FOLDER}/bind9"
+BIND9_BUILD_FOLDER="${SCRATCH_FOLDER}/bind9-aws-lc"
+BIND9_PATCH_FOLDER=${SRC_ROOT}/"tests/ci/integration/bind9_patch"
+AWS_LC_BUILD_FOLDER="${SCRATCH_FOLDER}/aws-lc-build"
+AWS_LC_INSTALL_FOLDER="${SCRATCH_FOLDER}/aws-lc-install"
+
+function bind9_build() {
+  autoreconf -fi  
+  PKG_CONFIG_PATH="${AWS_LC_INSTALL_FOLDER}/lib/pkgconfig" ./configure --with-openssl="${AWS_LC_INSTALL_FOLDER}" --enable-dnstap --enable-dnsrps --with-cmocka --with-libxml2 --enable-leak-detection
```
WillChilds-Klein, line 27: minor style nit

```suggestion
  PKG_CONFIG_PATH="${AWS_LC_INSTALL_FOLDER}/lib/pkgconfig" ./configure \
      --with-openssl="${AWS_LC_INSTALL_FOLDER}" \
      --enable-dnstap \
      --enable-dnsrps \
      --with-cmocka \
      --with-libxml2 \
      --enable-leak-detection
```

tests/ci/integration/run_bind9_integration.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# Assumes script is executed from the root of aws-lc directory
+SCRATCH_FOLDER="${SRC_ROOT}/BIND9_BUILD_ROOT"
+BIND9_SRC_FOLDER="${SCRATCH_FOLDER}/bind9"
+BIND9_BUILD_FOLDER="${SCRATCH_FOLDER}/bind9-aws-lc"
+BIND9_PATCH_FOLDER=${SRC_ROOT}/"tests/ci/integration/bind9_patch"
+AWS_LC_BUILD_FOLDER="${SCRATCH_FOLDER}/aws-lc-build"
+AWS_LC_INSTALL_FOLDER="${SCRATCH_FOLDER}/aws-lc-install"
+
+function bind9_build() {
```
WillChilds-Klein, line 25: looks like we're building statically. does bind's CLI `--version` include libssl/libcrypto version info? in other tests we try have assertions that the project under test has been correctly linked to AWS-LC. grepping `--version` output is one way to do this if libssl/libcrypto version is in there, another way is to dynamically link then `ldd ... | grep`.

WillChilds-Klein (approved):

ssl/ssl_lib.cc
WillChilds-Klein: gotcha. we can just bring `SSL_ERROR_ZERO_RETURN` into a case clause on L1397. I don't think we need the comments anymore.

samuel40791765: Great catch, just moved it to the switch statement

ssl/ssl_lib.cc
andrewhop: What does "messing up the state machine" mean for AWS-LC? Does OpenSSL avoid messing up their state machine without this assert some other way?

tests/ci/integration/run_bind9_integration.sh
```diff
@@ -0,0 +1,73 @@
...
+
+mkdir -p ${SCRATCH_FOLDER}
+rm -rf ${SCRATCH_FOLDER}/*
+cd ${SCRATCH_FOLDER}
+
+git clone https://gitlab.isc.org/isc-projects/bind9.git ${BIND9_SRC_FOLDER} --depth 1
+mkdir -p ${AWS_LC_BUILD_FOLDER} ${AWS_LC_INSTALL_FOLDER} ${BIND9_BUILD_FOLDER}
+ls
+
+aws_lc_build ${SRC_ROOT} ${AWS_LC_BUILD_FOLDER} ${AWS_LC_INSTALL_FOLDER} -DBUILD_TESTING=OFF -DBUILD_TOOL=OFF -DBUILD_SHARED_LIBS=1
```
andrewhop, line 57:
```suggestion
aws_lc_build ${SRC_ROOT} ${AWS_LC_BUILD_FOLDER} ${AWS_LC_INSTALL_FOLDER} -DBUILD_TESTING=OFF -DBUILD_TOOL=OFF -DBUILD_SHARED_LIBS=1 -DCMAKE_BUILD_TYPE=RelWithDebInfo
```

tests/ci/integration/bind9_patch/bind-fix.patch
```diff
@@ -0,0 +1,19 @@
+diff --git a/tests/isc/Makefile.am b/tests/isc/Makefile.am
+index 5cdd915..6ee1935 100644
+--- a/tests/isc/Makefile.am
++++ b/tests/isc/Makefile.am
+@@ -115,10 +115,12 @@ proxyheader_test_SOURCES =	\
+ 	proxyheader_test_data.h
+ 
+ proxystream_test_CPPFLAGS =	\
+-	$(AM_CPPFLAGS)
++	$(AM_CPPFLAGS)			\
```
andrewhop, line 10: Do you have any idea how this works with OpenSSL and Bind right now?

ssl/ssl_lib.cc
samuel40791765: I don't think OpenSSL has ever avoided this, in 1.1.1 this was even just a simple mode configuration: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/ssl/ssl_lib.c#L4086-L4089
Based on the ancient BoringSSL commit (https://github.com/aws/aws-lc/commit/63006a913b772d48a364915c3155d51744a8f704), the concern was that setting the same flag twice in a row would clear the mode bits and mess up the state.

A lot has changed since then, this commit (https://github.com/aws/aws-lc/commit/fa214e4a18785a4281375b56c00732b470e9cd5b) consolidated the shutdown state so that flags can't be cleared like the original behavior.

tests/ci/integration/run_bind9_integration.sh
samuel40791765: I had consider using Release, but that would "work around" the assertion I removed. Does `RelWithDebInfo` trip on assertions?

samuel40791765: I did a test run and `RelWithDebInfo` doesn't trip over assertions. I'm fine either way, was just looking for 100% coverage of the changes I made here.

andrewhop: Oh.... that's interesting, this makes me want to flip back more of our builds to just debug unless they actually need the speed like curl/mysql.

tests/ci/integration/bind9_patch/bind-fix.patch
andrewhop: Don't we install our package config?

samuel40791765: Yeah we install our binaries with pkgconfig, but without the flags here the test binary isn't finding the right artifacts.
https://github.com/aws/aws-lc/blob/ddc449d4f17ea2621a2e8d36138f2525e05a86a9/CMakeLists.txt#L1137-L1139

tests/ci/integration/run_bind9_integration.sh
samuel40791765: Yeah I'm all for it. Our consumers should be using Release, so we wouldn't be accidentally breaking anyone with new assertions. But it'd be better to see if any integrations are using these code paths.

ssl/ssl_lib.cc
samuel40791765: As sure good call, I'll update the documentation to say that it's ignored. I'm on the same page with you on this isn't much of a concern anymore.

WillChilds-Klein (approved):

tests/ci/integration/bind9_patch/bind-fix.patch
andrewhop: Is there a built in way to tell package config to look for other package config in a different location?

samuel40791765: Yes, the PKG_CONFIG_PATH environment variable was pkgconfig's own recommendation to get around this. I added the comment based off of that.
We use this in all our integration ci scripts that use pkgconfig.

andrewhop (approved):

PR #1424, opened 2024-01-26
Fix Clang-6 FIPS static build issue

skmcgrail:
Clang-6 exhibits a behavior where a filno declaration in a `.file` directive causes it to assume that all file numbers prior to it are declared. This cause errors when later directives define lower numbered filno's.

This only occurs because delocator computes the next largest fileno and injects `.file`/`.loc` directives about the `BORINGSSL_bcm_text_start` and `BORINGSSL_bcm_text_end` symbols. These aren't actually required to have working debugging, so opting to remove this all together.


skmcgrail:
Need to understand the ramifications of this change more. While CI is passing fine, and this resolves the issue with clang-6 it's unclear how to reproduce the issue described in the original commit in BoringSSL that introduced the directives on these labels: https://boringssl-review.googlesource.com/c/boringssl/+/27544

skmcgrail:
> Need to understand the ramifications of this change more. While CI is passing fine, and this resolves the issue with clang-6 it's unclear how to reproduce the issue described in the original commit in BoringSSL that introduced the directives on these labels: https://boringssl-review.googlesource.com/c/boringssl/+/27544

I did some digging into the clang 6 and clang 5 output on x86-64 and was not able to determine / reproduce the Clang behavior that BoringSSL was referring to in this commit.

I used `readelf --debug-dump=info bcm.o` and `readelf --debug-dump=line bcm.o` to review the output object file's DWARF debug information built from the `bcm-delocated.S` file output by the delocator. I did not see anything that stood out regarding for the `BORINGSSL_bcm_text_start` and `BORINGSSL_bcm_text_end` symbols in this debug information. (They were just not present in the debug section information). I think this is sufficient to proceed with mering this pull request, and if need be we can always roll-back this change if required.

samuel40791765 (approved):

samuel40791765 (approved):

torben-hansen (approved):

PR #1426, opened 2024-01-31
Speed up CodeBuild CI

andrewhop:
### Description of changes: 
1. Split OpenSSH into two jobs, one for main branch, and another for 8.9
2. Use RelWithDebInfo build for MySQL
3. Force MySQL to use double the number of jobs to run the tests
4. Update ARM fuzzer to run for a shorter period of time

### Call-outs:
The majority of the MySQL CI time is spent building (~50 minutes) and then running the tests (22 minutes). The build already uses all CPU cores, but the tests seem not take full advantage with the default `parallel=auto` which just uses 1 test per core (orange is CPU usage %, blue is memory usage %): 
image

### Testing:
Locally and in this CI. 

### DIFF
--- a/tests/ci/integration/run_mysql_integration.sh
+++ b/tests/ci/integration/run_mysql_integration.sh
@@ -45,8 +45,8 @@ function mysql_patch_reminder() {
 }
 
 function mysql_build() {
-  cmake ${MYSQL_SRC_FOLDER} -GNinja -DWITH_BOOST=${BOOST_INSTALL_FOLDER} -DWITH_SSL=${AWS_LC_INSTALL_FOLDER} "-B${MYSQL_BUILD_FOLDER}"
-  ninja -C ${MYSQL_BUILD_FOLDER}
+  cmake ${MYSQL_SRC_FOLDER} -GNinja -DWITH_BOOST=${BOOST_INSTALL_FOLDER} -DWITH_SSL=${AWS_LC_INSTALL_FOLDER} "-B${MYSQL_BUILD_FOLDER}" -DCMAKE_BUILD_TYPE=RelWithDebInfo
+  time ninja -C ${MYSQL_BUILD_FOLDER}
   ls -R ${MYSQL_BUILD_FOLDER}
 }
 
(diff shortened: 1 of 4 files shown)


tests/ci/integration/run_mysql_integration.sh
```diff
@@ -88,7 +88,7 @@ main.client_ssl_data_print  : Bug#0002 AWS-LC does not support Stateful session
 main.ssl_cache : Bug#0002 AWS-LC does not support Stateful session resumption (Session Caching).
 main.ssl_cache_tls13 : Bug#0002 AWS-LC does not support Stateful session resumption (Session Caching).
 "> skiplist
-  ./mtr --suite=main --force --parallel=auto --skip-test-list=${MYSQL_BUILD_FOLDER}/mysql-test/skiplist --retry-failure=3 --retry=3 --report-unstable-tests
+  ./mtr --suite=main --force --parallel=$((NUM_CPU_THREADS * 2)) --skip-test-list=${MYSQL_BUILD_FOLDER}/mysql-test/skiplist --retry-failure=3 --retry=3 --report-unstable-tests
```
WillChilds-Klein, line 91: this is already the number available cores, won't doubling it just cause more context switching between processes? is the idea that enough time will be spent waiting on IO that this won't matter?

tests/ci/integration/run_openssh_integration.sh
```diff
@@ -80,24 +79,25 @@ function openssh_run_tests() {
...
 
-# Buld AWS-LC as a shared library
+# Build AWS-LC as a shared library
 aws_lc_build "$SRC_ROOT" "$AWS_LC_BUILD_FOLDER" "$AWS_LC_INSTALL_FOLDER" -DBUILD_TESTING=OFF -DBUILD_TOOL=OFF -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_SHARED_LIBS=1
 install_aws_lc
 
-CODEBUILD_SKIPPED_TESTS="agent-subprocess forwarding multiplex channel-timeout forward-control agent-restrict connection-timeout"
+if [ "$OPENSSH_BRANCH" != "master" ]; then
+  checkout_openssh_branch "$OPENSSH_BRANCH"
+fi
```
WillChilds-Klein, line 91: can you do this checkout unconditionally? if OpenSSH ever changes the name of their default branch, this might break.

andrewhop: Git was complaining about a 'master' branch already existing, I couldn't find a nice option to checkout if it doesn't already exist so i went with this.

tests/ci/integration/run_mysql_integration.sh
andrewhop: The overall build time for this run is 86 minutes. Logs contain `Using parallel: 16` as expected, the test time took `Spent 18062.931 of 1376 seconds executing testcases` I'll try putting this back to auto and see if anything changes. Looking at this resource utilization CPU usage is still only 15-20% so maybe it was always I/O limited.

image

tests/ci/integration/run_openssh_integration.sh
WillChilds-Klein: hm, maybe you could do something like

```
if [ "$OPENSSH_BRANCH != $(git branch --show-current) ]; then
    checkout_openssh_branch "$OPENSSH_BRANCH
fi
```

WillChilds-Klein (approved):

samuel40791765 (approved):

PR #1429, opened 2024-02-01
Change sk_*_find signature to 2-arg for OpenSSL comapat

WillChilds-Klein:
### Description of changes: 

OpenSSL [omits][1] the `|out_index|` parameter from `|sk_*_find|`. This (breaking) change conforms with that interface and exposes our old, 3-arg interface with `|out_index|` as `|sk_*_find_awslc|`.

[1]: https://www.openssl.org/docs/man1.1.1/man3/sk_TYPE_find.html

### Call-outs:
- This is a **breaking change** for consumers using the `|sk_*_find|` function. Consuming code will need to be updated to either use the new `|sk_*_find|` function or to drop the second positional `|out_index| `parameter from their function call(s).

### Testing:
- CI checks

---

### DIFF
--- a/include/openssl/stack.h
+++ b/include/openssl/stack.h
@@ -196,14 +196,17 @@ void sk_SAMPLE_delete_if(STACK_OF(SAMPLE) *sk, sk_SAMPLE_delete_if_func func,
 //
 // If the stack is sorted (see |sk_SAMPLE_sort|), this function uses a binary
 // search. Otherwise it performs a linear search. If it finds a matching
-// element, it writes the index to |*out_index| (if |out_index| is not NULL) and
-// returns one. Otherwise, it returns zero.
+// element, it returns the index of that element. Otherwise, it returns -1.
 //
-// Note this differs from OpenSSL. The type signature is slightly different, and
-// OpenSSL's version will implicitly sort |sk| if it has a comparison function
-// defined.
+// Note this differs from OpenSSL in that OpenSSL's version will implicitly
+// sort |sk| if it has a comparison function defined.
+int sk_SAMPLE_find(const STACK_OF(SAMPLE) *sk, const SAMPLE *p);
+
+// sk_SAMPLE_find_awslc is like sk_SAMPLE_find, but if it finds a matching
+// element, it writes the index to |*out_index| (if |out_index| is not NULL)
+// and returns one. Otherwise, it returns zero.
 int sk_SAMPLE_find(const STACK_OF(SAMPLE) *sk, size_t *out_index,
-                   const SAMPLE *p);
+        const SAMPLE *p);
 
 // sk_SAMPLE_shift removes and returns the first element in |sk|, or NULL if
 // |sk| is empty.
@@ -502,8 +505,15 @@ BSSL_NAMESPACE_END
   /* use 2-arg sk_*_find for OpenSSL compatibility */                          \
   OPENSSL_INLINE int sk_##name##_find(const STACK_OF(name) *sk,                \
                                       constptrtype p) {                        \
-    return OPENSSL_sk_find((const OPENSSL_STACK *)sk, NULL,                    \
-                           (const void *)p, sk_##name##_call_cmp_func);        \
+    const size_t mask = 0xffffffffffffffffUL << (sizeof(int) * 8);             \
+    size_t out_index;                                                          \
+    int ok = OPENSSL_sk_find((const OPENSSL_STACK *)sk, &out_index,            \
+                             (const void *)p, sk_##name##_call_cmp_func);      \
+    /* return -1 if element not found or stack is larger than INT_MAX */       \
+    if (ok == 0 || (out_index & mask)) {                                       \
+      return -1;                                                               \
+    }                                                                          \
+    return (int) (out_index & ~mask);                                          \
   }                                                                            \
                                                                                \
   OPENSSL_INLINE ptrtype sk_##name##_shift(STACK_OF(name) *sk) {               \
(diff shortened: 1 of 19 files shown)


include/openssl/stack.h
```diff
@@ -492,12 +492,20 @@ BSSL_NAMESPACE_END
...
+  OPENSSL_INLINE int sk_##name##_find_awslc(const STACK_OF(name) *sk,          \
                                       size_t *out_index, constptrtype p) {     \
     return OPENSSL_sk_find((const OPENSSL_STACK *)sk, out_index,               \
                            (const void *)p, sk_##name##_call_cmp_func);        \
   }                                                                            \
                                                                                \
+  /* use 2-arg sk_*_find for OpenSSL compatibility */                          \
+  OPENSSL_INLINE int sk_##name##_find(const STACK_OF(name) *sk,                \
+                                      constptrtype p) {                        \
+    return OPENSSL_sk_find((const OPENSSL_STACK *)sk, NULL,                    \
```
torben-hansen, line 505: `sk_<>_find` should return the index now. But does `OPENSSL_sk_find` has changed as well, to not return any index.

include/openssl/stack.h
```diff
@@ -492,12 +495,27 @@ BSSL_NAMESPACE_END
...
                                       size_t *out_index, constptrtype p) {     \
     return OPENSSL_sk_find((const OPENSSL_STACK *)sk, out_index,               \
                            (const void *)p, sk_##name##_call_cmp_func);        \
   }                                                                            \
                                                                                \
+  /* use 2-arg sk_*_find for OpenSSL compatibility */                          \
+  OPENSSL_INLINE int sk_##name##_find(const STACK_OF(name) *sk,                \
+                                      constptrtype p) {                        \
+    const size_t mask = 0xffffffffffffffffUL << (sizeof(int) * 8);             \
+    size_t out_index;                                                          \
```
justsmth, line 509:
```c
size_t out_index = 0;
```

include/openssl/stack.h
```diff
@@ -492,12 +495,29 @@ BSSL_NAMESPACE_END
...
+  OPENSSL_INLINE int sk_##name##_find_awslc(const STACK_OF(name) *sk,          \
                                       size_t *out_index, constptrtype p) {     \
     return OPENSSL_sk_find((const OPENSSL_STACK *)sk, out_index,               \
                            (const void *)p, sk_##name##_call_cmp_func);        \
   }                                                                            \
                                                                                \
+  /* use 2-arg sk_*_find for OpenSSL compatibility */                          \
+  OPENSSL_INLINE int sk_##name##_find(const STACK_OF(name) *sk,                \
+                                      constptrtype p) {                        \
+    const size_t mask = sizeof(size_t) > sizeof(int)                           \
```
justsmth, line 508: The mask should also cover the sign bit, even if `sizeof(size_t) == sizeof(int)`:
```
const size_t mask = ~((size_t) 0)) << (sizeof(int) * 8 - 1);
```

I don't think it's the case for any supported platform that 
`sizeof(size_t) < sizeof(int)`

WillChilds-Klein: > The mask should also cover the sign bit, even if `sizeof(size_t) == sizeof(int)`

`size_t` is always unsigned, no?

> I don't think it's the case for any supported platform that `sizeof(size_t) < sizeof(int)`

I didn't think so either, but am concerned with the `sizeof(size_t) == sizeof(int)` case, hence the hard inequality. Initially I'd written this as `sizeof(size_t) != sizeof(int)`, used `<` to guard against cases I hadn't considered.

In any case, something is wrong with my implementation here. The [`ubuntu1604_gcc5x_x86` test is failing](https://tiny.amazon.com/zhmkmmgs/IsenLink) with a compile error:

```
../include/openssl/stack.h:509:46: error: left shift count >= width of type [-Werror=shift-count-overflow]
         ? (~((size_t) 0)) << (sizeof(int) * 8)                                 \
                                              ^
```

Same with `centos7_gcc4x_x86_64_fips`. I'll look into how earlier versions of GCC handle bit shifting.

WillChilds-Klein: I ended up dropping the masking logic altogether in favor of checking `INT_MAX` and casting.

justsmth (approved):

include/openssl/stack.h
```diff
@@ -492,12 +497,26 @@ BSSL_NAMESPACE_END
...
     return OPENSSL_sk_find((const OPENSSL_STACK *)sk, out_index,               \
                            (const void *)p, sk_##name##_call_cmp_func);        \
   }                                                                            \
                                                                                \
+  /* use 2-arg sk_*_find for OpenSSL compatibility */                          \
+  OPENSSL_INLINE int sk_##name##_find(const STACK_OF(name) *sk,                \
+                                      constptrtype p) {                        \
+    size_t out_index = 0;                                                      \
+    int ok = OPENSSL_sk_find((const OPENSSL_STACK *)sk, &out_index,            \
+                             (const void *)p, sk_##name##_call_cmp_func);      \
```
samuel40791765, line 512: Nit: Can this call `sk_##name##_find_awslc`? No strong preference though

samuel40791765 (approved):

include/openssl/stack.h
WillChilds-Klein: that would work, but IMO we should put any checks, etc. we want shared by both `sk_*_find` and `sk_*_find_awslc` in the concrete function `OPENSSL_sk_find` rather than one of the wrapping macros.

PR #1435, opened 2024-02-06
Add macho parser for use by C inject_hash

billbo-yang:
### Issues:
Addresses CryptoAlg-2240

### Description of changes: 
Adds a macho file parser and tests in anticipation of usage in an upcoming C language replacement for `inject_hash.go`

### Call-outs:
Since this is only meant to be used in a specific use-case, it's reading capabilities are tailored to only read & store information on the parts of the macho file that we're interested in (__test section, __const section, the string table, and the symbol table. We can open another PR to add additional capabilities if the need arises.

### Testing:
Tests pass locally


CMakeLists.txt
```diff
@@ -1025,6 +1025,16 @@ if(BUILD_TESTING)
       add_custom_target(fips_specific_tests_if_any)
     endif()
 
+    # Add macho parser tests if FIPS and on MacOS
+    if(FIPS AND APPLE)
+      add_custom_target(
+        macho_parser_tests
```
torben-hansen, line 1031: Do you actually need a separate target for this? The target is not configured to depend on the what it executes, so in principle it's misconfigured. You can fix that.

Or you can just define a `test_macho_parser` as a dependency to e.g.  `fips_specific_tests_if_any` in `/util/fipstools/inject_hash/tests/CMakeList.txt`. Maybe?

Then you would not need this target and `ninja run_tests` would execute the macho tests.

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+uint32_t find_macho_symbol_index(uint8_t *symbol_table_data, size_t symbol_table_size, uint8_t *string_table_data, size_t string_table_size, const char *symbol_name, uint32_t *base) {
+    char* string_table = NULL;
+    int ret = 0;
+
+    if (symbol_table_data == NULL || string_table_data == NULL) {
+        LOG_ERROR("Symbol and string table pointers cannot be null to find the symbol index");
+        goto end;
+    }
+
+    string_table = malloc(string_table_size);
```
torben-hansen, line 134: Must check return value of `malloc`

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+        goto end;
+    }
+    if (base) {
+        index = index - *base;
+    }
+    ret = index;
+
+end:
+    if (string_table != NULL) {
+        free(string_table);
```
torben-hansen, line 162: `free(NULL)` is just a no-op

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+                goto end;
+            }
+
+            fseek(file, macho->sections[i].offset, SEEK_SET);
+            fread(section_data, 1, macho->sections[i].size, file);
+
+            if (size != NULL) {
+                *size = macho->sections[i].size;
+            }
+            if (offset) {
```
torben-hansen, line 105: You can't decide if you want to use `if (var != NULL)` or `if (var)`? (:

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+
+            ret = section_data;
+            goto end;
+        }
+    }
+
+    LOG_ERROR("Section %s not found in macho file %s", section_name, filename);
+end:
+    if (file != NULL) {
+        fclose(file);
```
torben-hansen, line 117: For everywhere: `free(NULL)` is a no-op.

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+    load_commands = malloc(macho->macho_header.sizeofcmds);
+    fread(load_commands, macho->macho_header.sizeofcmds, 1, file);
+
+    // We're only looking for __text, __const in the __TEXT segment, and the string & symbol tables
+    macho->num_sections = 4;
+    macho->sections = malloc(macho->num_sections * sizeof(section_info));
+
+    uint32_t section_index = 0;
+    for (uint32_t i = 0; i < macho->macho_header.sizeofcmds / BIT_MODIFIER; i += load_commands[i].cmdsize / BIT_MODIFIER) {
+        if (load_commands[i].cmd == LC_SEG) {
```
torben-hansen, line 38: mmmh, your tests just use `LC_SEGMENT_64`?

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+    int ret = 1;
+
+    file = fopen(filename, "rb");
+    if (!file) {
+        LOG_ERROR("Error opening file %s", filename);
+        ret = 0;
+        goto end;
+    }
+
+    fread(&macho->macho_header, sizeof(macho_header), 1, file);
```
torben-hansen, line 22: Check `man fread`. Could add some error handling and sanity check on the file.

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+#include <stdint.h>
+
+#include "common.h"
+#include "macho_parser.h"
+
+// Documentation for the Mach-O structs can be found in macho-o/loader.h and mach-o/nlist.h
+int read_macho_file(const char *filename, machofile *macho) {
+    FILE *file = NULL;
+    load_cmd *load_commands = NULL;
+    int ret = 1;
```
torben-hansen, line 13: From a principal point of view, I prefer to start with assuming an error state. Makes it easier to handle errors without accidentally forgetting to switch to an error state before returning.

util/fipstools/inject_hash/macho_parser/macho_parser.h
```diff
@@ -0,0 +1,46 @@
...
+typedef struct section_64 section_data;
+typedef struct nlist_64 symbol_info;
+
+typedef struct {
+    macho_header macho_header;
+    section_info *sections;
+    uint32_t num_sections;
+} machofile;
+
+int read_macho_file(const char *filename, machofile *macho);
```
torben-hansen, line 37: Document functions?

util/fipstools/inject_hash/macho_parser/macho_parser.h
```diff
@@ -0,0 +1,46 @@
...
+
+typedef struct {
+    char name[16];
+    size_t size;
+    uint32_t offset;
+} section_info;
+
+// Since we only support 64-bit architectures on Apple, we don't need to account for any of the 32-bit structures
+#define LC_SEG LC_SEGMENT_64
+#define BIT_MODIFIER 8
```
torben-hansen, line 22: Something is counting in bytes and other things in bits and you need to go from bits to bytes somewhere?

CMakeLists.txt
```diff
@@ -1025,6 +1025,16 @@ if(BUILD_TESTING)
       add_custom_target(fips_specific_tests_if_any)
     endif()
 
+    # Add macho parser tests if FIPS and on MacOS
+    if(FIPS AND APPLE)
```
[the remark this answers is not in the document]
billbo-yang: Even if it works regardless of FIPS, I don't see why we need the parser other than for FIPS currently. So I don't think there's a point in building it if we're not doing FIPS.

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+    for (uint32_t i = 0; i < macho->num_sections; i++) {
+        if (strcmp(macho->sections[i].name, section_name) == 0) {
+            section_data = malloc(macho->sections[i].size);
+            if (!section_data) {
+                LOG_ERROR("Error allocating memory for section data");
+                goto end;
+            }
+
+            fseek(file, macho->sections[i].offset, SEEK_SET);
+            fread(section_data, 1, macho->sections[i].size, file);
```
justsmth, line 100: Check the return values on these calls.

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+
+void free_macho_file(machofile *macho) {
+    free(macho->sections);
+    free(macho);
+    macho = NULL;
+}
+
+uint8_t* get_macho_section_data(const char *filename, machofile *macho, const char *section_name, size_t *size, uint32_t *offset) {
+    FILE *file = NULL;
+    uint8_t *section_data = NULL;
```
justsmth, line 83: You can move the declaration of `section_data` down to the first place it's assigned to.

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+            goto end;
+        }
+    }
+
+    LOG_ERROR("Section %s not found in macho file %s", section_name, filename);
+end:
+    if (file != NULL) {
+        fclose(file);
+    }
+    if (ret == NULL && section_data != NULL) {
```
justsmth, line 119: This condition doesn't appear to be possible.  (Moving the `section_data` declaration into the for loop should provide more clarity on this.)

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+        free(section_data);
+    }
+    return ret;
+}
+
+uint32_t find_macho_symbol_index(uint8_t *symbol_table_data, size_t symbol_table_size, uint8_t *string_table_data, size_t string_table_size, const char *symbol_name, uint32_t *base) {
+    char* string_table = NULL;
+    int ret = 0;
```
justsmth, line 127: The return value should be a `uint32_t`.

util/fipstools/inject_hash/macho_parser/macho_parser.c
justsmth: There's a macro defining it:
```
#define LC_SEG LC_SEGMENT_64
```

util/fipstools/inject_hash/macho_parser/macho_parser.h
```diff
@@ -0,0 +1,46 @@
...
+
+typedef struct mach_header_64 macho_header;
```
justsmth, line 24: I'm not a fan of using `typedef` to alias library-provided types to custom names. Maintaining/scaling this code might be easier by leaving the type names unchanged.

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+        if (load_commands[i].cmd == LC_SEG) {
+            segment_load_cmd *segment = (segment_load_cmd *)&load_commands[i];
+            if (strcmp(segment->segname, "__TEXT") == 0) {
+                section_data *sections = (section_data *)&segment[1];
+                for (uint32_t j = 0; j < segment->nsects; j++) {
+                    if (strcmp(sections[j].sectname, "__text") == 0 || strcmp(sections[j].sectname, "__const") == 0) {
+                        macho->sections[section_index].offset = sections[j].offset;
+                        macho->sections[section_index].size = sections[j].size;
+                        strcpy(macho->sections[section_index].name, sections[j].sectname);
+                        section_index++;
```
justsmth, line 47: Verify `section_index < 4` after each increment.

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+            section_index++;
+            macho->sections[section_index].offset = symtab->stroff;
+            macho->sections[section_index].size = symtab->strsize;
+            strcpy(macho->sections[section_index].name, "__string_table");
+            section_index++;
+        }
+    }
+
+
+end:
```
justsmth, line 65: Before returning, check that all 4 required sections were found.

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,165 @@
...
+    macho->sections = malloc(macho->num_sections * sizeof(section_info));
+
+    uint32_t section_index = 0;
+    for (uint32_t i = 0; i < macho->macho_header.sizeofcmds / BIT_MODIFIER; i += load_commands[i].cmdsize / BIT_MODIFIER) {
+        if (load_commands[i].cmd == LC_SEG) {
+            segment_load_cmd *segment = (segment_load_cmd *)&load_commands[i];
+            if (strcmp(segment->segname, "__TEXT") == 0) {
+                section_data *sections = (section_data *)&segment[1];
+                for (uint32_t j = 0; j < segment->nsects; j++) {
+                    if (strcmp(sections[j].sectname, "__text") == 0 || strcmp(sections[j].sectname, "__const") == 0) {
```
justsmth, line 43: Instead of incrementing a `section_index` counter, why not just have an explicit index at which "__text", "__const", etc. are stored?

billbo-yang: I think the current implementation is more easily expandable if we ever decide that we're interested in more than just the `__text` and `__const` sections, and I'm unsure if changing it around provides any added benefit to our current use-case.

util/fipstools/inject_hash/macho_parser/macho_parser.c
billbo-yang: I'm under the impression that I need to leave it out of the for loop in order to be able to free it, unless I'm misunderstanding you here?

justsmth: You could have it like this:
```
...
    for (uint32_t i = 0; i < macho->num_sections; i++) {
        if (strcmp(macho->sections[i].name, section_name) == 0) {
            uint8_t *section_data = malloc(macho->sections[i].size); // declaration
            if (section_data == NULL) {
                LOG_ERROR("Error allocating memory for section data");
                goto end;
            }

            fseek(file, macho->sections[i].offset, SEEK_SET);
            bytes_read = fread(section_data, 1, macho->sections[i].size, file);
            if (bytes_read != macho->sections[i].size) {
                free(section_data);                                // free on error
                LOG_ERROR("Error reading section data from file %s", filename);
                goto end;
            }
...
```

util/fipstools/inject_hash/macho_parser/macho_parser.c
justsmth: I think having a known location for each section could make it more robust to potential(?) changes in the order that sections are processed. I don't see it limiting  expansion to handle other sections; sections being added to the processing would likewise be assigned to a specific index.

util/fipstools/inject_hash/macho_parser/macho_parser.h
justsmth: * Based on the description, it seems the name of this macro/constant would more appropriately be `BYTES_PER_WORD`?

billbo-yang: Taking another look at this, I believe I can remove the need for a macro at all

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,218 @@
...
+    macho->sections = malloc(macho->num_sections * sizeof(section_info));
+    if (macho->sections == NULL) {
+        LOG_ERROR("Error allocating memory for macho sections");
+    }
+
+    int text_found = 0;
+    int const_found = 0;
+    int symtab_found = 0;
+
+    for (uint32_t i = 0; i < macho->macho_header.sizeofcmds / sizeof(struct load_command); i += load_commands[i].cmdsize / sizeof(struct load_command)) {
```
justsmth, line 59: This logic looks suspcious. Is `load_commands[i].cmdsize` not always the same as `sizeof(struct load_command)`?  This logic assumes that it can be an exact multiple (>1). A comment here should explain the "why".

Otherwise, if the two values are equal it could be simplified:
```
    const uint32_t num_cmds = macho->macho_header.sizeofcmds / sizeof(struct load_command);
    for (uint32_t i = 0; i < num_cmds; i += 1) {
```

billbo-yang: You're right, that was an oversight on my part - thanks for pointing it out!

billbo-yang: The reason this works is that sizeof(struct load_command) results in 8 bytes (the struct consists of two `uint32_t`s). The documentation in `loader.h` says that `sizeofcmds` actually needs to be a multiple of 8 on 64bit systems, so this should work here.

I'll add a comment and your simplified code.

util/fipstools/inject_hash/macho_parser/tests/macho_tests.h
```diff
@@ -0,0 +1,278 @@
...
+        struct symtab_command test_symtab_command = {
+            .cmd = LC_SYMTAB,
+            .cmdsize = sizeof(struct symtab_command),
+            .symoff = symtab_command_symoff,
+            .nsyms = NUM_SYMS,
+            .stroff = symtab_command_stroff,
+            .strsize = symtab_command_strsize,
+        };
+
+        if (fwrite(&test_header, sizeof(struct mach_header_64), 1, file) != 1) {
```
torben-hansen, line 101: I dunno what the definition are of these structures. But generally `sizeof(struct) != <sum of size of fields in struct>`.

This should work locally here, because you use sizeof(struct) to be both write and read. But in general, this is not the correct way to serialise.

util/fipstools/inject_hash/macho_parser/macho_parser.c
```diff
@@ -0,0 +1,219 @@
...
+    if (macho->sections == NULL) {
+        LOG_ERROR("Error allocating memory for macho sections");
+    }
+
+    int text_found = 0;
+    int const_found = 0;
+    int symtab_found = 0;
+
+    // mach-o/loader.h explains that cmdsize (and by extension sizeofcmds) must be a multiple of 8 on 64-bit systems. struct load_command will always be 8 bytes.
+    for (uint32_t i = 0; i < macho->macho_header.sizeofcmds / sizeof(struct load_command); i += load_commands[i].cmdsize / sizeof(struct load_command)) {
```
torben-hansen, line 60:
```suggestion
    for (size_t i = 0; i < macho->macho_header.sizeofcmds / sizeof(struct load_command); i += load_commands[i].cmdsize / sizeof(struct load_command)) {
```
Idiomatic to use `size_t`. If one can speak of idiomatic C. Same for the other occurrences.

torben-hansen (approved):

justsmth (approved):

torben-hansen (approved):

justsmth (approved):

PR #1437, opened 2024-02-08
OpenBSD 7.4 and 7.5 Support

skmcgrail:
This pull request adds support for OpenBSD [7.4](https://www.openbsd.org/plus74.html) and [7.5](https://www.openbsd.org/plus75.html) releases.

OpenBSD has a release cadence similar to Go:
> When a critical bug is found, the fix will be committed to the -current tree (and made available in [snapshot builds](https://www.openbsd.org/faq/faq5.html#Flavors)) as soon as possible. It will then be backported to the two most recent OpenBSD releases in the form of [errata](https://www.openbsd.org/errata.html) and details will be sent to the [announce](https://lists.openbsd.org/cgi-bin/mj_wwwusr?func=lists-long-full&extra=announce) mailing list.

See https://www.openbsd.org/faq/faq10.html#Patches


crypto/fipsmodule/rand/urandom.c
```diff
@@ -78,6 +78,11 @@
 #endif
 #endif
 
+#if defined(OPENSSL_OPENBSD)
+#define URANDOM_BLOCKS_FOR_ENTROPY
+// OpenBSD does not support getrandom, so we will fallback to /dev/urandom
```
PiotrSikora, line 83: OpenBSD supports [`getentropy()`](https://man.openbsd.org/getentropy), which is supported and preferred in upstream BoringSSL (see: [boringssl@53f09ad241ddf7e37a426a5e3c6dfa6d4dca9706](https://boringssl.googlesource.com/boringssl/+/53f09ad241ddf7e37a426a5e3c6dfa6d4dca9706)), so it looks that it was lost in a merge?

PiotrSikora (changes requested):

crypto/fipsmodule/bcm.c
```diff
@@ -221,6 +221,7 @@ static void BORINGSSL_maybe_set_module_text_permissions(int permission) {
   if (mprotect((void *)page_start,
                ((uintptr_t)BORINGSSL_bcm_text_end) - page_start,
                permission) != 0) {
+    exit(errno);
```
skmcgrail, line 224: Need to remove this is next push, this was just for debugging purposes.

tests/ci/common_posix_setup.sh
```diff
@@ -23,7 +23,7 @@ PLATFORM=$(uname -m)
 
 NUM_CPU_THREADS=''
 KERNEL_NAME=$(uname -s)
-if [[ "${KERNEL_NAME}" == "Darwin" ]]; then
+if [[ "${KERNEL_NAME}" == "Darwin" || "${KERNEL_NAME}" == "OpenBSD" ]]; then
```
justsmth, line 26: This appears to be true for other *BSDs as well, e.g., [FreeBSD](https://man.freebsd.org/cgi/man.cgi?sysctl(8)) and [NetBSD](https://man.netbsd.org/sysctl.8). Could we match instead:
```
if [[ "${KERNEL_NAME}" == "Darwin" || "${KERNEL_NAME}" =~ *BSD ]]; then
```

justsmth:
Could portions of this be broken out into incremental changes that have value independent of our OpenBSD-support? For example, I see value in a PR that migrates our shell scripts to `#!/usr/bin/env bash`, and in a PR that adds explicit `.rodata` and `.text` sections in our Perl/assembly files. This PR is really big and there appear to be several parts that we could merge prior to fully supporting OpenBSD.

skmcgrail:
Should be good to review now. I've collapsed commits and did some reorganization to make this easy to review.

Without the regenerated sources commit this is only a review of [442 additions and 124 deletions.](https://github.com/aws/aws-lc/compare/main...skmcgrail:aws-lc:openbsd~1)

crypto/fipsmodule/rand/urandom.c
```diff
@@ -252,6 +256,13 @@ static void init_once(void) {
   return;
 #endif
 
+#if defined(OPENSSL_OPENBSD)
+  // To get system randomness on OpenBSD we use |arc4random_buf| function
+  // which is recommended to use for C APIs rather then /dev/random.
```
skmcgrail, line 261:
```suggestion
  // which is recommended to use for C APIs rather then /dev/urandom.
```

Leaving this note so that I can fix it after I get other comments.

CMakeLists.txt
```diff
@@ -41,6 +41,17 @@ if(FIPS)
   endif()
 endif()
 
+if(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+  # OpenBSD by defaults links with --execute-only this is problematic for two reasons:
```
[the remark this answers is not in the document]
skmcgrail: The s2n-bignum comment is also applicable to the non-FIPS builds.

justsmth (approved):

justsmth (dismissed):

util/fipstools/test_fips.c
```diff
@@ -214,12 +214,12 @@ int main(int argc, char **argv) {
     goto err;
   }
   printf("About to AES-ECB decrypt ");
-  hexdump(output, out_len);
+  hexdump(output, sizeof(kPlaintext));
   for (size_t j = 0; j < out_len / 16; j++) {
```
torben-hansen, line 218: should this be `out_len`?

util/fipstools/test_fips.c
```diff
@@ -234,19 +234,19 @@ int main(int argc, char **argv) {
...
   AES_ctr128_encrypt(kPlaintext, output, sizeof(kPlaintext), &aes_key, aes_iv, ecount_buf, &num);
   printf("  got ");
-  hexdump(output, out_len);
+  hexdump(output, sizeof(kPlaintext));
 
 
   /* AES-CTR Decryption */
   OPENSSL_memset(aes_iv, 0, sizeof(aes_iv));
   printf("About to AES-CTR Decrypt ");
   hexdump(output, out_len);
```
torben-hansen, line 246: `out_len`?

tests/ci/run_ssl_runner_valgrind_tests.sh
```diff
@@ -1,7 +1,9 @@
-#!/bin/bash -ex
+#!/usr/bin/env bash
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0 OR ISC
 
+set -ex
```
torben-hansen, line 5: you like `set -exo pipefail` or `set -ex` the most? :)

`tests/ci/run_ssl_asan_tests.sh` use the latter, here the former.

tests/ci/run_fips_tests.sh
```diff
@@ -1,17 +1,33 @@
...
 
+function static_linux_supported() {
+  if [[ ("$(uname -s)" == 'Linux'*) && (("$(uname -p)" == 'x86_64'*) || ("$(uname -p)" == 'aarch64'*)) ]]; then
+    return 0
+  fi
+  return 1
+}
+
+function static_openbsd_supported() {
+  if [[ "$(uname -s)" == 'OpenBSD' && "$(uname -p)" == 'x86_64'* ]]; then
```
torben-hansen, line 17: Always clean `'OpenBSD'? (no suffix?)

tests/ci/common_posix_setup.sh
```diff
@@ -23,7 +23,7 @@ PLATFORM=$(uname -m)
 
 NUM_CPU_THREADS=''
 KERNEL_NAME=$(uname -s)
-if [[ "${KERNEL_NAME}" == "Darwin" ]]; then
+if [[ "${KERNEL_NAME}" == "Darwin" || "${KERNEL_NAME}" =~ .*BSD ]]; then
```
torben-hansen, line 26: why needed?

crypto/curve25519/internal.h
```diff
@@ -26,11 +26,11 @@ extern "C" {
...
 // set, s2n-bignum path is capable.
-#if ((defined(OPENSSL_X86_64) &&                                               \
-          !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)) ||                        \
-      defined(OPENSSL_AARCH64)) &&                                             \
-     (defined(OPENSSL_LINUX) || defined(OPENSSL_APPLE)) &&                     \
-     !defined(OPENSSL_NO_ASM)
+#if ((defined(OPENSSL_X86_64) && !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)) || \
+     defined(OPENSSL_AARCH64)) &&                                              \
+    (defined(OPENSSL_LINUX) || defined(OPENSSL_APPLE) ||                       \
+     defined(OPENSSL_OPENBSD)) &&                                              \
```
torben-hansen, line 32: does this work when .text contains data tables in the s2n-bignum impl?

crypto/cmake/crypto-config.cmake
```diff
@@ -1,4 +1,4 @@
-if(WIN32 OR UNIX OR APPLE)
+if(WIN32 OR UNIX)
```
torben-hansen, line 1: how is this related to enabling openbsd?

CMakeLists.txt
```diff
@@ -41,6 +41,17 @@ if(FIPS)
...
+  # 1. The FIPS shared and static builds need to compute the module signature hash by reading the .text section
+  # 2. s2n-bignum x86 assembly sources still have references to static data in the .text section
+  if(NOT BUILD_SHARED_LIBS)
```
torben-hansen, line 48: a'ha you disable it... This would be very nice to have

util/fipstools/test_fips.c
skmcgrail: `out_len` was being abused in a few places across multiple algorithms and was confusing to track. So I intentionally started to strip it out, but I think what is missing here is that the for loop below this change should be fixed.

crypto/cmake/crypto-config.cmake
skmcgrail: It's related to an earlier comment from Justin here `APPLE` and `UNIX` were used in a cmake check, but UNIX covers both, so this was the only other spot where we could simplify. Not BSD specific.

tests/ci/run_ssl_runner_valgrind_tests.sh
justsmth: This is just keeping the same `bash` flags as were already on the script.  B/c this is switching to invoking bash via `#!/usr/bin/env bash` (instead of `#!/bin/bash`) these flags needed to be moved into a `set` statement.

tests/ci/run_fips_tests.sh
skmcgrail: Correct `uname -s` doesn't attach any version information as a suffix.

tests/ci/common_posix_setup.sh
skmcgrail: Similar to Apple, OpenBSD doesn't have a /proc/cpuinfo that you can query to get the number of processor cores. So this uses the same utility available to both macOS and OpenBSD for getting that information.

tests/ci/run_ssl_runner_valgrind_tests.sh
skmcgrail: I didn't add pipefail to the scripts that didn't have it previously but I'll just go back and fix them up to all be the same.

crypto/curve25519/internal.h
skmcgrail: Looks like you figured the answer our later in a comment further down 👍

CMakeLists.txt
skmcgrail: Yeah need to poke that ticket we have with them to determine if there is a technical block for it (maybe related to formal verification?), or if we can just submit a PR for them. But I don't want to block this on that for the time being.

Regardless for FIPS we don't have a good strategy without disabling this, as the premise of our integrity check is to compute the hash of the rodata and text section. With text being execute-only and no read that doesn't work.

util/fipstools/test_fips.c
torben-hansen, line 246: being pedantic, should be?
```suggestion
  hexdump(output, sizeof(kPlaintext));
```

torben-hansen (approved):

justsmth (approved):

PR #1445, opened 2024-02-20
Add BIO_meth_set_write_ex and BIO_meth_set_read_ex

andrewhop:
### Description of changes: 
BIO_meth_set_write_ex and BIO_meth_set_read_ex use the modern callback method signature with a simple success/error return value and an out parameter for the amount of data actually processed. 

### Call-outs:
I used a conversion function so all of our internal code can continue to use the old bread/bwrite methods. I'm interested if people think it's worth flipping our internal use over to the new style for clarity/safety. 

### Testing:
I modified two tests that had some interesting behavior around the read/write and callbacks to test the old and new methods.


crypto/bio/bio.c
```diff
@@ -734,15 +734,57 @@ int (*BIO_meth_get_destroy(const BIO_METHOD *method)) (BIO *) {
...
+static int bwrite_ex_to_briwte_conversion(BIO *bio, const char *input, int requested) {
+  if (requested > INT_MAX) {
+    requested = INT_MAX;
+  }
+  size_t actual = 0;
+  const int success = bio->method->bwrite_ex(bio, input, requested, &actual);
+  if (success <= 0) {
+    return success;
+  }
+  // this is safe because we capped the write to INT_MAX
```
justsmth, line 746: This implies the `actual` value can never be more than `requested`. This is a reasonable expectation, but it's unenforceable. We can't prevent the callback from setting the value of `actual` to a larger value.

crypto/bio/bio.c
```diff
@@ -734,15 +734,57 @@ int (*BIO_meth_get_destroy(const BIO_METHOD *method)) (BIO *) {
...
 }
 
+int BIO_meth_set_write_ex(BIO_METHOD *method, int (*write)(BIO *, const char *, size_t, size_t *)) {
+  method->bwrite_ex = write;
+  method->bwrite = bwrite_ex_to_briwte_conversion;
+  return 1;
+}
+
+static int bread_ex_to_bread_conversion(BIO *bio, char *input, int requested) {
+  if (requested > INT_MAX) {
```
justsmth, line 764: Same as above.

crypto/bio/bio.c
```diff
@@ -734,15 +734,57 @@ int (*BIO_meth_get_destroy(const BIO_METHOD *method)) (BIO *) {
   return method->destroy;
 }
 
+static int bwrite_ex_to_briwte_conversion(BIO *bio, const char *input, int requested) {
+  if (requested > INT_MAX) {
```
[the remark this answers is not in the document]
andrewhop: Good catch, this is a hold over when I was going the other way bwrite to bwrite_ex and had to cap the size_t to an int. This should check for negative values now.

PR #1447, opened 2024-02-21
Use OPENSSL_zalloc in more places

justsmth:
### Description of changes: 
* Use `OPENSSL_zalloc` instead of `OPENSSL_malloc` when it makes subsequent initialization unnecessary. 

### Call-outs:
* If we either define an `OPENSSL_rezalloc` or modify `OPENSSL_realloc` to call `OPENSSL_zalloc`, there's a couple more places where we might avoid subsequent initializations.

### Testing:
Debug/FIPS build succeeded on my Mac.

### DIFF
--- a/crypto/asn1/a_object.c
+++ b/crypto/asn1/a_object.c
@@ -188,11 +188,6 @@ ASN1_OBJECT *ASN1_OBJECT_new(void) {
   if (ret == NULL) {
     return NULL;
   }
-  assert(ret->length == 0);
-  assert(ret->data == NULL);
-  assert(ret->nid == 0);
-  assert(ret->sn == NULL);
-  assert(ret->ln == NULL);
   ret->flags = ASN1_OBJECT_FLAG_DYNAMIC;
   return ret;
 }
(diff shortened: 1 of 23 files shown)


crypto/asn1/a_object.c
```diff
@@ -184,15 +184,15 @@ ASN1_OBJECT *c2i_ASN1_OBJECT(ASN1_OBJECT **out, const unsigned char **inp,
...
-  ret->length = 0;
-  ret->data = NULL;
-  ret->nid = 0;
-  ret->sn = NULL;
-  ret->ln = NULL;
+  assert(ret->length == 0);
+  assert(ret->data == NULL);
+  assert(ret->nid == 0);
+  assert(ret->sn == NULL);
+  assert(ret->ln == NULL);
```
skmcgrail, line 195: We don't consistently `assert` on every structure in this pull request, is this providing us any benefit? I feel like a test-case to demonstrate `OPENSSL_zalloc` behavior would be more appropriate to verify it calls calloc. Is more appropriate? Is `OPENSSL_zalloc` allowed to be overridden with a user provided implementation similar to malloc, is that concern?

justsmth: > We don't consistently assert on every structure in this pull request, is this providing us any benefit?

I converted initializations that were now redundant into (debug) `assert` statements. The benefit is arguably minor: (*) documenting the (now implicit) initial value of these fields, and (*) verifying that my change is preserving its initialized value.

justsmth: In the [latest commit](https://github.com/aws/aws-lc/pull/1447/commits/396ac54acf46209f3dbfb6b42096e61c616ef734), I removed many of these debug asserts.

nebeid (approved):

skmcgrail (approved):

PR #1449, opened 2024-02-21
Add misc. x509 un/lock and set1 functions

WillChilds-Klein:
### Issues:
n/a

### Description of changes: 

The set1 functions are mostly cribbed from [OpenSSL's implementation][1].

[1]: https://github.com/openssl/openssl/blob/4a6f70c03182b421d326831532edca32bcdb3fb1/crypto/x509/x509_lu.c#L506

### Call-outs:
n/a

### Testing:
- CI


crypto/x509/x509_lu.c
```diff
@@ -196,6 +196,16 @@ X509_STORE *X509_STORE_new(void) {
   return NULL;
 }
 
+int X509_STORE_lock(X509_STORE *v) {
+    CRYPTO_MUTEX_lock_write(&v->objs_lock);
```
andrewhop, line 200: Null checks here and down below.

WillChilds-Klein: ah, good catch.

crypto/x509/x509_test.cc
```diff
@@ -5022,12 +5043,41 @@ TEST(X509Test, AddDuplicates) {
...
-  EXPECT_TRUE(X509_STORE_add_cert(store.get(), a.get()));
-  EXPECT_TRUE(X509_STORE_add_cert(store.get(), b.get()));
+
+  // Half the threads add duplicate certs, the other half take a lock and
+  // look them up to exercise un/locking functions.
+  const size_t kNumThreads = 10;
+  std::vector<std::thread> threads;
+  for (size_t i = 0; i < kNumThreads/2; i++) {
+    threads.emplace_back([&] {
+      // Sleep with some jitter to offset thread execution
```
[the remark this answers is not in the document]
WillChilds-Klein: my worry with jitter-less execution is that the adders/getters might be fast enough as to take/release locks while the subsequent threads are spawning, with the threads not getting a chance to contend. maybe this isn't a viable concern; i don't know what the overhead for spawning a thread is like... i don't know, maybe this is overkill and thread spawn overhead is small enough that we can be confident that contention would occur without any sleeps?

the way it's currently written is that half the threads sleep for some random jitter before adding a cert (which [takes then releases a lock internally](https://github.com/aws/aws-lc/blob/main/crypto/x509/x509_lu.c#L329)) while the other half of the threads take a lock immediately, then sleep, then do a lookup and release the lock. my hope is that with the getters all taking locks immediately then sleeping (while holding the lock), it would cause contention with the adders who sleep for a bit before attempting to take a lock.

we can ensure that the two sets of threads always entirely overlap by making the getter lock-holders always sleep for longer than the adders.

include/openssl/x509.h
```diff
@@ -2800,7 +2800,11 @@ OPENSSL_EXPORT int X509_OBJECT_get_type(const X509_OBJECT *a);
 OPENSSL_EXPORT X509 *X509_OBJECT_get0_X509(const X509_OBJECT *a);
 // X509_OBJECT_get0_X509_CRL returns the |X509_CRL| associated with |a|
 OPENSSL_EXPORT X509_CRL *X509_OBJECT_get0_X509_CRL(const X509_OBJECT *a);
+OPENSSL_EXPORT int X509_OBJECT_set1_X509(X509_OBJECT *a, X509 *obj);
+OPENSSL_EXPORT int X509_OBJECT_set1_X509_CRL(X509_OBJECT *a, X509_CRL *obj);
```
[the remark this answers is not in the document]
WillChilds-Klein: me too. poking around a bit, i found the "LIBRARY CONVENTIONS" section [here](https://www.openssl.org/docs/man3.2/man7/ossl-guide-libraries-introduction.html) that explains the set0/set1 convention explicitly.

crypto/x509/x509_test.cc
```diff
@@ -5022,12 +5043,43 @@ TEST(X509Test, AddDuplicates) {
...
+    });
+    threads.emplace_back([&] {
+      uint8_t sleep_buf[1];
+      ASSERT_TRUE(RAND_bytes(sleep_buf, sizeof(sleep_buf)));
+      ASSERT_TRUE(X509_STORE_lock(store.get()));
+      // Sleep after taking the lock to cause contention. Sleep longer than the
+      // adder half of threads to ensure we hold the lock while they contend
+      // for it.
+      std::this_thread::sleep_for(std::chrono::microseconds(11 + (sleep_buf[0] % 5)));
+      EXPECT_TRUE(X509_OBJECT_retrieve_by_subject(
```
justsmth, line 5071: `X509_OBJECT_retrieve_by_subject` is safe to call b/c the function is only being passed a field of the struct, and thus can't acquire its lock. However, invoking the similarly named `X509_STORE_CTX_get_by_subject` would cause a deadlock.

include/openssl/x509.h
```diff
@@ -2800,7 +2800,22 @@ OPENSSL_EXPORT int X509_OBJECT_get_type(const X509_OBJECT *a);
...
+OPENSSL_EXPORT int X509_OBJECT_set1_X509(X509_OBJECT *a, X509 *obj);
+
+// X509_OBJECT_set1_X509_CRL sets CRL |obj| on |a| and uprefs |obj|. As with
+// other set1 methods, |a| does not take ownership of |obj|; the caller is
+// responsible for managing freeing |obj| when appropriate.
+OPENSSL_EXPORT int X509_OBJECT_set1_X509_CRL(X509_OBJECT *a, X509_CRL *obj);
+
 OPENSSL_EXPORT X509_STORE *X509_STORE_new(void);
+// X509_STORE_lock takes a write lock on |v|. return 1 on success, 0 on failure
+OPENSSL_EXPORT int X509_STORE_lock(X509_STORE *v);
```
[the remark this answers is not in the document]
WillChilds-Klein: agreed.

crypto/x509/x509_test.cc
WillChilds-Klein: comment updated.

justsmth (approved):

andrewhop (approved):

PR #1456, opened 2024-02-27
Make the clang-6 bcm-delocated.S directive behavior for start/end symbols dynamic

skmcgrail:
### Description of changes: 
CRT team was seeing failures on Clang-8 with [this](https://github.com/aws/aws-lc/commit/f61870199f1bdfe3182e493231e60ea7243edbcb) change, even though we didn't have failures for clang-8 in our CI. Opting to gate this fix on detecting the clang compiler version being used. CRT reported clang-6 fix did address their issue.

### Testing
I verified that on clang-6 it continues to not emit the directive. Then verified on clang-8 that it correctly defaults to the original behavior of emitting the `.file` directive for these symbols.

### DIFF
--- a/crypto/fipsmodule/CMakeLists.txt
+++ b/crypto/fipsmodule/CMakeLists.txt
@@ -159,8 +159,8 @@ endif()
 # clang-6 (and older) knows how to compile AVX512 assembly instructions,
 # but only if it's given the right flags (e.g. -mavx512*).
 # The flags are not required for any other compiler we are running in the CI.
-if ((CMAKE_ASM_COMPILER_ID MATCHES "Clang" OR CMAKE_ASM_COMPILER MATCHES "clang") AND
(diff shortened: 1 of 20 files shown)


justsmth (approved):

nebeid (approved):
Will we rely on the customer testing the code if our CI didn't fail before the fix?

skmcgrail:
> Will we rely on the customer testing the code if our CI didn't fail before the fix?

I verified that on clang-6 it continues to not emit the directive. Then verified on clang-8 that it correctly defaults to the original behavior of emitting the `.file` directive for these symbols.

nebeid:
> > Will we rely on the customer testing the code if our CI didn't fail before the fix?
> 
> I verified that on clang-6 it continues to not emit the directive. Then verified on clang-8 that it correctly defaults to the original behavior of emitting the `.file` directive for these symbols.

Maybe good to add this test to the Testing section in the PR description.

PR #1462, opened 2024-02-29
Test BORINGSSL_FIPS_abort in CI

oliver-wm:
Hey Andrew set this up to see your CI results, cases where these functions have been hooked or altered in some way. Can you give any more context as to why it's written like this?


andrewhop:
Hi, what are you trying to test/achieve with this change? Are you running into issues with the abort/exit loop?

PR #1473, opened 2024-03-05
Document no-op functions and flags in AWS-LC

samuel40791765:
### Description of changes: 
Update porting guide for AWS-LC. This is the second phase which includes updated function documentation.

### Call-outs:
Originally, this was to include links to the corresponding functions/sections in the new tables as well, but we can't actually provide correct links until we have the correct commit hash in main.
A third phase will link the tables to the corresponding section.

### DIFF
--- a/include/openssl/bio.h
+++ b/include/openssl/bio.h
@@ -880,7 +880,10 @@ OPENSSL_EXPORT int (*BIO_meth_get_puts(const BIO_METHOD *method)) (BIO *, const 
 // General No-op Functions [Deprecated].
 
 // BIO_set_write_buffer_size returns zero.
-OPENSSL_EXPORT OPENSSL_DEPRECATED int BIO_set_write_buffer_size(
+//
+// TODO: Add |OPENSSL_DEPRECATED|. nginx defines -Wno-deprecated-declarations
+//       and depends on this.
+OPENSSL_EXPORT int BIO_set_write_buffer_size(
     BIO *bio, int buffer_size);
 
 
--- a/include/openssl/bio.h
+++ b/include/openssl/bio.h
@@ -881,10 +881,8 @@ OPENSSL_EXPORT int (*BIO_meth_get_puts(const BIO_METHOD *method)) (BIO *, const 
 
 // BIO_set_write_buffer_size returns zero.
 //
-// TODO: Add |OPENSSL_DEPRECATED|. nginx defines -Wno-deprecated-declarations
-//       and depends on this.
-OPENSSL_EXPORT int BIO_set_write_buffer_size(
-    BIO *bio, int buffer_size);
+// TODO: Add |OPENSSL_DEPRECATED|. nginx defines -Werror and depends on this.
+OPENSSL_EXPORT int BIO_set_write_buffer_size(BIO *bio, int buffer_size);
 
 
 // Private functions
--- a/include/openssl/bio.h
+++ b/include/openssl/bio.h
@@ -880,7 +880,8 @@ OPENSSL_EXPORT int (*BIO_meth_get_puts(const BIO_METHOD *method)) (BIO *, const 
 // General No-op Functions [Deprecated].
 
 // BIO_set_write_buffer_size returns zero.
-OPENSSL_EXPORT int BIO_set_write_buffer_size(BIO *bio, int buffer_size);
+OPENSSL_EXPORT OPENSSL_DEPRECATED int BIO_set_write_buffer_size(
+    BIO *bio, int buffer_size);
 
 
 // Private functions
--- a/include/openssl/bio.h
+++ b/include/openssl/bio.h
@@ -862,9 +862,6 @@ OPENSSL_EXPORT const BIO_METHOD *BIO_f_base64(void);
 
 OPENSSL_EXPORT void BIO_set_retry_special(BIO *bio);
 
-// BIO_set_write_buffer_size returns zero.
-OPENSSL_EXPORT int BIO_set_write_buffer_size(BIO *bio, int buffer_size);
-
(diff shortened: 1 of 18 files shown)


justsmth (commented):
Should we use our [OPENSSL_DEPRECATED](https://github.com/aws/aws-lc/blob/b47c6ff69ef57c0489d2ecd4e534a2ba84961e43/include/openssl/base.h#L163) macro for many of these function declarations?

samuel40791765:
> Should we use our [OPENSSL_DEPRECATED](https://github.com/aws/aws-lc/blob/b47c6ff69ef57c0489d2ecd4e534a2ba84961e43/include/openssl/base.h#L163) macro for many of these function declarations?

Good point, this was brought up as something to do, but I hadn't realized we could do it in this PR.

justsmth (dismissed):

justsmth (approved):

include/openssl/bio.h
```diff
@@ -879,6 +876,17 @@ OPENSSL_EXPORT int BIO_meth_set_puts(BIO_METHOD *method,
 // BIO_meth_get_puts returns |puts| function of |method|.
 OPENSSL_EXPORT int (*BIO_meth_get_puts(const BIO_METHOD *method)) (BIO *, const char *);
 
+
+// General No-op Functions [Deprecated].
+
+// BIO_set_write_buffer_size returns zero.
+//
+// TODO: Add |OPENSSL_DEPRECATED|. nginx defines -Wno-deprecated-declarations
```
andrewhop, line 884: Do we have a issue for this and the other TODOs?

include/openssl/conf.h
```diff
@@ -124,33 +124,39 @@ OPENSSL_EXPORT const char *NCONF_get_string(const CONF *conf,
...
 // CONF_modules_finish does nothing.
-OPENSSL_EXPORT void CONF_modules_finish(void);
+OPENSSL_EXPORT OPENSSL_DEPRECATED void CONF_modules_finish(void);
 
-// OPENSSL_config does nothing.
+// OPENSSL_config does nothing. This has been deprecated since OpenSSL 1.1.0.
+//
+// TODO: Add |OPENSSL_DEPRECATED|. nginx defines -Wno-deprecated-declarations
+//       and depends on this.
 OPENSSL_EXPORT void OPENSSL_config(const char *config_name);
```
andrewhop, line 155:
```suggestion
OPENSSL_EXPORT OPENSSL_DEPRECATED void OPENSSL_config(const char *config_name);
```

include/openssl/digest.h
```diff
@@ -271,17 +271,14 @@ OPENSSL_EXPORT int EVP_marshal_digest_algorithm(CBB *cbb, const EVP_MD *md);
 // EVP_MD_unstable_sha3_enable is a no-op as SHA3 is always enabled.
 OPENSSL_EXPORT void EVP_MD_unstable_sha3_enable(bool enable);
 
-// EVP_MD_unstable_sha3_is_enabled always returns true as SHA3 is always enabled.
+// EVP_MD_unstable_sha3_is_enabled always returns true as SHA3 is always
+// enabled.
 OPENSSL_EXPORT bool EVP_MD_unstable_sha3_is_enabled(void);
```
andrewhop, line 276: Should this get moved down to the no-op section?
```suggestion
// EVP_MD_unstable_sha3_is_enabled always returns true as SHA3 is always
// enabled.
OPENSSL_EXPORT OPENSSL_DEPRECATED bool EVP_MD_unstable_sha3_is_enabled(void);
```

include/openssl/digest.h
```diff
@@ -363,6 +352,24 @@ struct env_md_ctx_st {
 } /* EVP_MD_CTX */;
 
 
+// General No-op Functions [Deprecated].
+
+// EVP_MD_CTX_set_flags does nothing. We strongly discourage doing any
+// additional configurations when consuming |EVP_MD_CTX|.
+OPENSSL_EXPORT void EVP_MD_CTX_set_flags(EVP_MD_CTX *ctx, int flags);
```
andrewhop, line 359:
```suggestion
OPENSSL_EXPORT OPENSSL_DEPRECATED void EVP_MD_CTX_set_flags(EVP_MD_CTX *ctx, int flags);
```

include/openssl/thread.h
```diff
@@ -116,7 +116,7 @@ OPENSSL_EXPORT int AWSLC_thread_local_clear(void);
 // calls to |AWSLC_thread_local_clear|.
 OPENSSL_EXPORT int AWSLC_thread_local_shutdown(void);
 
-// Deprecated functions.
+// General No-op Functions [Deprecated].
```
andrewhop, line 119: Can we add OPENSSL_DEPRECATED to all the functions below?

samuel40791765: I had been pondering this since we enable thread safety by default, so it was OK for these functions to do nothing and not alert the consuming application.
I dug into[ the OpenSSL headers](https://github.com/openssl/openssl/blob/4514e02cdfc96589d5e8ab0a08942fafa8e418ae/include/openssl/crypto.h.in#L297) and noticed there are some deprecated functions though, we can mark those to align with upstream.

include/openssl/bio.h
samuel40791765: Good call, I'll create one

include/openssl/conf.h
samuel40791765: I wish we could do this, but nginx defines `-Werror` and will fail the build :(

include/openssl/digest.h
samuel40791765: Hmm yeah we should

andrewhop (approved):

justinwsmith (approved):

justsmth (approved):

PR #1474, opened 2024-03-05
Adding No-op functions required for NodeJS compatability

smittals2:
### Description of changes: 
Added OPENSSL_init and SSL_CTX_set_dh_auto as no-op functions to support compilation with NodeJS. OPENSSL_init does nothing by design and SSL_CTX_set_dh_auto because DH is not supported with SSL in AWS-LC. 


ssl/ssl_lib.cc
```diff
@@ -2760,6 +2760,10 @@ void SSL_CTX_set_tmp_dh_callback(SSL_CTX *ctx,
 void SSL_set_tmp_dh_callback(SSL *ssl, DH *(*cb)(SSL *ssl, int is_export,
                                                  int keylength)) {}
 
+long SSL_CTX_set_dh_auto(SSL_CTX *ctx, int onoff) {
+    return 0;
```
andrewhop, line 2764: Is this always safe? Could customers accidentally think they're turning on FFDH even though AWS-LC doesn't support it at all? Should we require onoff to always be 0?

include/openssl/ssl.h
```diff
@@ -5159,6 +5159,9 @@ OPENSSL_EXPORT void SSL_set_tmp_dh_callback(SSL *ssl,
                                             DH *(*cb)(SSL *ssl, int is_export,
                                                       int keylength));
 
+// SSL_CTX_set_dh_auto does nothing.
```
andrewhop, line 5162: Pending my other comment the pattern here seems to be nothing is for void, this technically does one thing: return 1.
```suggestion
// SSL_CTX_set_dh_auto returns 1.
```

crypto/crypto.c
```diff
@@ -153,4 +153,8 @@ int OPENSSL_init_crypto(uint64_t opts, const OPENSSL_INIT_SETTINGS *settings) {
   return 1;
 }
 
+void OPENSSL_init(void) {
+    return;
+}
```
andrewhop, line 158: NP pattern seems to be empty, I could see some compilers getting tripped up on this. 
```suggestion
void OPENSSL_init(void) {}
```

ssl/ssl_lib.cc
smittals2: https://www.openssl.org/docs/man3.1/man3/SSL_CTX_set_dh_auto.html
From above link:
"All of these functions/macros return 1 for success or 0 on error."

So returning 0 should always be treated as an error, I can update documentation with this info.

crypto/bio/bio_mem.c
```diff
@@ -302,4 +302,4 @@ int BIO_set_mem_buf(BIO *bio, BUF_MEM *b, int take_ownership) {
 
 int BIO_set_mem_eof_return(BIO *bio, int eof_value) {
   return (int)BIO_ctrl(bio, BIO_C_SET_BUF_MEM_EOF_RETURN, eof_value, NULL);
-}
+}
```
andrewhop, line 305: Unrelated change

include/openssl/bio.h
```diff
@@ -447,7 +447,6 @@ OPENSSL_EXPORT int BIO_set_mem_buf(BIO *bio, BUF_MEM *b, int take_ownership);
 // default is -1 so that additional data can be written once exhausted.
 OPENSSL_EXPORT int BIO_set_mem_eof_return(BIO *bio, int eof_value);
 
-
```
andrewhop, line 450: Two blank lines creates a new section in the documentation and should be kept.

include/openssl/ssl.h
andrewhop, line 5162:
```suggestion
// SSL_CTX_set_dh_auto does nothing and returns 0 for error.
```

andrewhop (approved):

skmcgrail (approved):

PR #1475, opened 2024-03-05
Error Codes for NodeJS compatibility added

smittals2:
### Description of changes: 
Added in some error codes from OpenSSL 


include/openssl/err.h
```diff
@@ -339,7 +339,17 @@ enum {
...
+  ERR_LIB_PKCS12,
+  ERR_LIB_DSO=37,
+  ERR_LIB_OSSL_STORE=44,
+  ERR_LIB_FIPS,
+  ERR_LIB_CMS,
+  ERR_LIB_TS,
+  ERR_LIB_CT=50,
+  ERR_LIB_ASYNC,
+  ERR_LIB_KDF,
+  ERR_LIB_SM2
```
WillChilds-Klein, line 352: in light of [this commit](https://github.com/aws/aws-lc/commit/6bdd10042ea5e9015bb1ecf740b76a29ea7c0302), you'll need to convert these into macros. do we know why there are explicit values assigned to some of these enum entries?

WillChilds-Klein (changes requested):

include/openssl/err.h
smittals2: Ah, I'll make that change. 

There were many other error codes in OpenSSL that I have not added here, only those required by NodeJS. By providing explicit values I've tried to maintain that same values for those error codes across libraries. However, the values shouldn't matter so I can also make them numerically ascending if we prefer that

smittals2: https://github.com/openssl/openssl/blob/5677992679b38950c6a0c3775fd57378e1879ba5/include/openssl/err.h.in#L113

include/openssl/dh.h
```diff
@@ -385,4 +385,8 @@ BSSL_NAMESPACE_END
 #define DH_R_INVALID_NID 106
 #define DH_R_INVALID_PARAMETERS 107
 
+// error code for nodejs compatibility
```
WillChilds-Klein, line 388: nit: we might want to remove this comment. other projects we integrate with may depend on this as well, nodejs just happens to be the first one.

WillChilds-Klein (approved):

include/openssl/dh.h
andrewhop, line 388:
```suggestion
```

include/openssl/err.h
```diff
@@ -339,6 +339,16 @@ OPENSSL_EXPORT int ERR_get_next_error_library(void);
...
+#define ERR_LIB_PKCS12          35
+#define ERR_LIB_DSO             36
+#define ERR_LIB_OSSL_STORE      37
+#define ERR_LIB_FIPS            38
+#define ERR_LIB_CMS             39
+#define ERR_LIB_TS              40
+#define ERR_LIB_CT              41
+#define ERR_LIB_ASYNC           42
+#define ERR_LIB_KDF             43
+#define ERR_LIB_SM2             44
```
andrewhop, line 351: Hmmmmmmm, does NodeJS not handle `OPENSSL_NO_SM2` correctly?

smittals2: Not sure but they define mappings to these specific ERR_LIB_* codes, so we need to define them as well.

andrewhop: Where do you see that? A naive search only finds the OpenSSL dependency code and no real usage https://github.com/search?q=repo%3Anodejs%2Fnode%20ERR_LIB_SM2&type=code

smittals2: They create a mapping to those err codes here:
https://github.com/nodejs/node/blob/1abff073921bcb0631602032aef0135bccfaee0d/src/crypto/crypto_util.cc#L501

WillChilds-Klein (approved):

andrewhop (approved):

PR #1476, opened 2024-03-06
Adding OPENSSL_secure_zalloc and BIO_s_secmem

smittals2:
### Description of changes: 
Added BIO_s_secmem and OPENSSL_secure_zalloc. These functions are related to Secure Heaps which aren't supported in AWS-LC. Therefore, these functions use their normal alternatives (BIO_s_mem and OPENSSL_zalloc).


crypto/mem.c
```diff
@@ -329,6 +329,8 @@ size_t CRYPTO_secure_used(void) { return 0; }
 
 void *OPENSSL_secure_malloc(size_t size) { return OPENSSL_malloc(size); }
 
+void *OPENSSL_secure_zalloc(size_t size) { return OPENSSL_zalloc(size); }
```
justsmth, line 332: We've established a precedence with `OPENSSL_secure_malloc` of allocating "unsecured"(?) memory, although I think it would be more proper for these to return `NULL`. 

[`CRYPTO_secure_malloc_init` always returns 0](https://github.com/aws/aws-lc/blob/7b383f4e14be4ed792a48585544887aed536f2c4/crypto/mem.c#L324) to indicate our lack of support for secure heaps.

include/openssl/bio.h
```diff
@@ -447,6 +447,9 @@ OPENSSL_EXPORT int BIO_set_mem_buf(BIO *bio, BUF_MEM *b, int take_ownership);
 // default is -1 so that additional data can be written once exhausted.
 OPENSSL_EXPORT int BIO_set_mem_eof_return(BIO *bio, int eof_value);
 
+// Returns the normal BIO_METHOD |BIO_s_mem|. Deprecated since AWS-LC does not
+// support secure heaps.
+OPENSSL_EXPORT OPENSSL_DEPRECATED const BIO_METHOD *BIO_s_secmem(void);
```
justsmth, line 452: Maybe we should likewise mark `OPENSSL_secure_zalloc`, `OPENSSL_secure_malloc`, etc., as deprecated due to not supporting secure heaps?

crypto/mem.c
samuel40791765: Wouldn't that break any codepaths trying to allocate memory when calling `OPENSSL_secure_malloc` or `OPENSSL_secure_zalloc`? It's true that we don't support "secure heaps" today, but I'm not sure if we want to block making integration easier due to the non-support.

include/openssl/mem.h
```diff
@@ -250,6 +250,9 @@ OPENSSL_EXPORT size_t CRYPTO_secure_used(void);
 // OPENSSL_secure_malloc calls |OPENSSL_malloc|.
 OPENSSL_EXPORT void *OPENSSL_secure_malloc(size_t size);
 
+// OPENSSL_secure_zalloc calls |OPENSSL_zalloc|.
+OPENSSL_EXPORT void *OPENSSL_secure_zalloc(size_t size);
+
 // OPENSSL_secure_clear_free calls |OPENSSL_clear_free|.
 OPENSSL_EXPORT void OPENSSL_secure_clear_free(void *ptr, size_t len);
```
samuel40791765, line 257: Adding some documentation here on what secure heaps are and that we don't support this yet would help.

samuel40791765 (commented):
Not trying to nitpick the wording here. But in the PR description and final commit message, could we replace "un-secure alternatives" with something else? It could make things seem scary and insecure, but in reality we're just pointing to the traditional alternatives.
It'd also help to clarify that these are "no-ops" in the PR title, since we aren't actually implementing secure heap versions of the functions.

crypto/mem.c
smittals2: Returning NULL was my first thought but doing so would break code using these functions and we are creating more instances of de-referencing NULL pointers? We should make a decision on which is more important. However, given that for other functions we "fail" (by returning 0), I think returning NULL ([to indicate failure](https://www.openssl.org/docs/man3.0/man3/CRYPTO_secure_malloc_init.html)) here might be the consistent way forward. We could also leave it as is and improve the documentation provided.

include/openssl/mem.h
```diff
@@ -247,8 +247,16 @@ OPENSSL_EXPORT int CRYPTO_secure_malloc_initialized(void);
...
+// could return arbitrary data from the program's dynamic memory area where sensitive information may be stored.
+// AWS-LC does not support secure heaps. Therefore, |OPENSSL_secure_malloc| and |OPENSSL_secure_zalloc| are
+// implemented as wrappers around their normal counterparts.
+
 // OPENSSL_secure_malloc calls |OPENSSL_malloc|.
-OPENSSL_EXPORT void *OPENSSL_secure_malloc(size_t size);
+OPENSSL_EXPORT OPENSSL_DEPRECATED void *OPENSSL_secure_malloc(size_t size);
+
+// OPENSSL_secure_zalloc calls |OPENSSL_zalloc|.
+OPENSSL_EXPORT OPENSSL_DEPRECATED void *OPENSSL_secure_zalloc(size_t size);
```
andrewhop, line 259: I think it is worth a small doc and plan for this secure heap decisions. I don't like emitting a warning that some people may or not see, but then continuing on with the normal allocations.

include/openssl/mem.h
```diff
@@ -238,20 +238,27 @@ OPENSSL_EXPORT int CRYPTO_set_mem_functions(
   void *(*r)(void *, size_t, const char *, int),
   void (*f)(void *, const char *, int));
 
+// OPENSSL supports the concept of secure heaps to help protect applications from pointer overruns or underruns that
+// could return arbitrary data from the program's dynamic memory area where sensitive information may be stored.
+// AWS-LC does not support secure heaps. Therefore, the following functions are marked as deprecated.
```
samuel40791765, line 243:
```suggestion
// AWS-LC does not support secure heaps. Therefore, the following functions are marked as 
// deprecated. The initialization functions intentionally return zero to indicate that secure
// heaps aren't supported.
// If there is any interest in utilizing "secure heaps" with AWS-LC, cut us an issue at
// https://github.com/aws/aws-lc/issues/new/choose 
```

samuel40791765 (approved):

justsmth (approved):

samuel40791765 (approved):

justsmth (approved):

PR #1479, opened 2024-03-06
Provide an API to turn off blinding for RSA

amirhosv:
### Description of changes: 
JCA allows private RSA keys without public exponent to be used for signing and verifying. The RSA structure is opaque, so we provide an API to allow disabling blinding for RSA keys.

Here are examples in ACCP code that would turn blinding off by accessing the `flags` field of an RSA strucuture:
* https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/csrc/keyutils.cpp#L74
* https://github.com/corretto/amazon-corretto-crypto-provider/blob/main/csrc/java_evp_keys.cpp#L502

### Testing:
+ run_tests


crypto/fipsmodule/rsa/rsa.c
```diff
@@ -1048,6 +1048,12 @@ int RSA_blinding_on(RSA *rsa, BN_CTX *ctx) {
   return 1;
 }
 
+void RSA_blinding_off(RSA *rsa) {
+  if (rsa != NULL) {
+    rsa->flags |= RSA_FLAG_NO_BLINDING;
+  }
```
justsmth, line 1054: I think `RSA_blinding_on` needs to be updated to invert this, even if it's simply updating `rsa->flags` -- that function is currently a no-op.

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -1045,9 +1045,18 @@ int RSA_flags(const RSA *rsa) { return rsa->flags; }
 int RSA_test_flags(const RSA *rsa, int flags) { return rsa->flags & flags; }
 
 int RSA_blinding_on(RSA *rsa, BN_CTX *ctx) {
+  if (rsa != NULL) {
+    rsa->flags &= (~RSA_FLAG_NO_BLINDING);
+  }
```
dkostic, line 1050: this will not work, there is a setup routine that should be done for blinding to be enabled. I suggest leaving the function a no-op.

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -1045,9 +1045,18 @@ int RSA_flags(const RSA *rsa) { return rsa->flags; }
...
   return 1;
 }
 
+void RSA_blinding_off(RSA *rsa) {
```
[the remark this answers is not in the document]
amirhosv: Sounds like a good plan.

dkostic (dismissed):

crypto/fipsmodule/rsa/rsa.c
justsmth: If `rsa_blinding_off_temp_for_accp_compatibility` really does disable blinding and there's no way for bliding to be re-enabled, then `RSA_blinding_on` should return an error (`0`) when it's later called (instead of returning success on the no-op).

amirhosv:
Close and open the PR to restart the CI tasks that were canceled.

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -877,7 +877,13 @@ int RSA_flags(const RSA *rsa) { return rsa->flags; }
 int RSA_test_flags(const RSA *rsa, int flags) { return rsa->flags & flags; }
 
 int RSA_blinding_on(RSA *rsa, BN_CTX *ctx) {
-  return 1;
+  return (rsa != NULL && (rsa->flags & RSA_FLAG_NO_BLINDING) == 0) ? 1 : 0;
```
justsmth, line 880: Nit-pick: Another pair of parenthesis would make this world a little happier:
```C
  return (rsa != NULL && ((rsa->flags & RSA_FLAG_NO_BLINDING) == 0)) ? 1 : 0;
```

justsmth (dismissed):

crypto/fipsmodule/rsa/rsa.c
amirhosv: done

dkostic (approved):

PR #1481, opened 2024-03-08
Update return type for EVP_EncodeUpdate

samuel40791765:
### Description of changes: 
tpm2-tools expects `EVP_EncodeUpdate` to return an integer, instead of being a void function: https://github.com/tpm2-software/tpm2-tools/commit/54a41786245de115c23a5163b631de8e1175940b.

Apparently the return type of this has been updated since OpenSSL 1.1.0: https://github.com/openssl/openssl/commit/cf3404fcc77aaf592c95326cbdd25612a8af6878 It seems worth updating to align with upstream. 

### Call-outs:
N/A

### Testing:
Slightly modified the existing test.

### DIFF
--- a/crypto/base64/base64.c
+++ b/crypto/base64/base64.c
@@ -134,13 +134,13 @@ void EVP_EncodeInit(EVP_ENCODE_CTX *ctx) {
   OPENSSL_memset(ctx, 0, sizeof(EVP_ENCODE_CTX));
 }
 
-void EVP_EncodeUpdate(EVP_ENCODE_CTX *ctx, uint8_t *out, int *out_len,
-                      const uint8_t *in, size_t in_len) {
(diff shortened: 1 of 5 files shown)


crypto/base64/base64.c
```diff
@@ -196,6 +196,8 @@ void EVP_EncodeUpdate(EVP_ENCODE_CTX *ctx, uint8_t *out, int *out_len,
     total = 0;
```
darylmartin100, line 196: We should consider returning a 0 here.  Looking at the OpenSSL implementation, they do return 0 if there is to much output data.

smittals2 (approved):

andrewhop (approved):

smittals2 (approved):

PR #1482, opened 2024-03-08
Added u16 endian loading/storing functions, SSL_CIPHER_find, and SSL_client_hello_get0_ciphers

smittals2:
### Description of changes: 
Added u16 endian loading/storing functions, SSL_CIPHER_find, and SSL_client_hello_get0_ciphers and tested all of them.

### Call-outs:
SSL_client_hello_get0_ciphers uses a new field added to ssl_st: ssl->client_hello_cipher_suites_arr.


ssl/ssl_cipher.cc
```diff
@@ -1446,6 +1446,17 @@ const SSL_CIPHER *SSL_get_cipher_by_value(uint16_t value) {
       ssl_cipher_id_cmp));
 }
 
+const SSL_CIPHER *SSL_CIPHER_find(SSL *ssl, const unsigned char *ptr) {
+  if (ptr) {
```
andrewhop, line 1450:
```suggestion
  if (ptr != nullptr) {
```

ssl/ssl_lib.cc
```diff
@@ -3315,3 +3315,34 @@ int SSL_CTX_set1_curves_list(SSL_CTX *ctx, const char *curves) {
 int SSL_set1_curves_list(SSL *ssl, const char *curves) {
   return SSL_set1_groups_list(ssl, curves);
 }
+
+size_t SSL_client_hello_get0_ciphers(SSL *ssl, const unsigned char **out) {
+  STACK_OF(SSL_CIPHER) *client_cipher_suites = SSL_get_client_ciphers(ssl);
+  if (!client_cipher_suites) {
```
andrewhop, line 3321:
```suggestion
  if (client_cipher_suites == nullptr) {
```

ssl/ssl_lib.cc
```diff
@@ -3315,3 +3315,34 @@ int SSL_CTX_set1_curves_list(SSL_CTX *ctx, const char *curves) {
...
+      return 0;
+  }
+
+  if (out) {
```
andrewhop, line 3325:
```suggestion
  if (out != nullptr) {
```

ssl/ssl_lib.cc
```diff
@@ -3315,3 +3315,34 @@ int SSL_CTX_set1_curves_list(SSL_CTX *ctx, const char *curves) {
...
+    // Called before
+    if(!ssl->client_cipher_suites_arr.empty()) {
+      *out = reinterpret_cast<unsigned char*>(ssl->client_cipher_suites_arr.data());
+      return ssl->client_cipher_suites_arr.size();
+    }
+
+    size_t total_length = sk_SSL_CIPHER_num(client_cipher_suites);
+    assert(ssl->client_cipher_suites_arr.Init(total_length));
```
andrewhop, line 3333: Is assert the right check here? Should this always be a failure?

include/openssl/ssl.h
```diff
@@ -1838,6 +1843,12 @@ OPENSSL_EXPORT const SSL_CIPHER *SSL_get_current_cipher(const SSL *ssl);
 // performed a new handshake.
 OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_client_ciphers(const SSL *ssl);
 
+// SSL_client_hello_get0_ciphers provides access to the client ciphers field from the
+// Client Hello, optionally writing the result to an out pointer. It returns the field
+// length if successful, or 0 if |ssl| is a client or the handshake hasn't occurred yet.
+// Free |*out| after use if provided to the function.
```
andrewhop, line 1849: Should out be freed? Isn't it a pointer to the array in the SSL struct which will be cleaned up automatically?

ssl/ssl_test.cc
```diff
@@ -2379,6 +2379,79 @@ static bool ConnectClientAndServer(bssl::UniquePtr<SSL> *out_client,
...
+  ASSERT_TRUE(SSL_CTX_set_min_proto_version(client_ctx.get(), TLS1_3_VERSION));
+  ASSERT_TRUE(SSL_CTX_set_max_proto_version(client_ctx.get(), TLS1_3_VERSION));
+
+  bssl::UniquePtr<SSL> client, server;
+  ASSERT_TRUE(CreateClientAndServer(&client, &server,
+                                    client_ctx.get(), server_ctx.get()));
+
+  // Handshake not completed, getting ciphers should fail
+  ASSERT_FALSE(SSL_client_hello_get0_ciphers(client.get(), nullptr));
+  ASSERT_FALSE(SSL_client_hello_get0_ciphers(server.get(), nullptr));
```
andrewhop, line 2435: Should this pass in valid out pointers? Can this distinguish between the handshake not being done and the out pointer not being valid?

ssl/ssl_lib.cc
```diff
@@ -3315,3 +3315,34 @@ int SSL_CTX_set1_curves_list(SSL_CTX *ctx, const char *curves) {
...
+
+    // Construct list of cipherIDs
+    for (size_t i = 0; i < sk_SSL_CIPHER_num(client_cipher_suites); i++) {
```
andrewhop, line 3336:
```suggestion
    for (size_t i = 0; i < num_ciphers; i++) {
```

ssl/ssl_lib.cc
```diff
@@ -3315,3 +3315,34 @@ int SSL_CTX_set1_curves_list(SSL_CTX *ctx, const char *curves) {
...
+  }
+
+  if (out) {
+    // Called before
+    if(!ssl->client_cipher_suites_arr.empty()) {
+      *out = reinterpret_cast<unsigned char*>(ssl->client_cipher_suites_arr.data());
+      return ssl->client_cipher_suites_arr.size();
+    }
+
+    size_t total_length = sk_SSL_CIPHER_num(client_cipher_suites);
```
andrewhop, line 3332:
```suggestion
    size_t num_ciphers = sk_SSL_CIPHER_num(client_cipher_suites);
```

ssl/ssl_lib.cc
```diff
@@ -3315,3 +3315,34 @@ int SSL_CTX_set1_curves_list(SSL_CTX *ctx, const char *curves) {
...
+      uint16_t iana_id = SSL_CIPHER_get_protocol_id(cipher);
+
+      CRYPTO_store_u16_be(&ssl->client_cipher_suites_arr[i], iana_id);
+    }
+
+    *out = reinterpret_cast<unsigned char*>(ssl->client_cipher_suites_arr.data());
+  }
+
+  // Return the size
+  return sk_SSL_CIPHER_num(client_cipher_suites);
```
andrewhop, line 3347:
```suggestion
  return num_ciphers;
```

ssl/ssl_lib.cc
```diff
@@ -3315,3 +3315,34 @@ int SSL_CTX_set1_curves_list(SSL_CTX *ctx, const char *curves) {
...
+
+    // Construct list of cipherIDs
+    for (size_t i = 0; i < sk_SSL_CIPHER_num(client_cipher_suites); i++) {
+      const SSL_CIPHER *cipher = sk_SSL_CIPHER_value(client_cipher_suites, i);
+      uint16_t iana_id = SSL_CIPHER_get_protocol_id(cipher);
+
+      CRYPTO_store_u16_be(&ssl->client_cipher_suites_arr[i], iana_id);
+    }
+
+    *out = reinterpret_cast<unsigned char*>(ssl->client_cipher_suites_arr.data());
```
andrewhop, line 3343: Is it safe to do this cast from an array of uint16_t to unsigned chars which are closer to uint8t_? Will this hit any endianness issues?

ssl/ssl_cipher.cc
```diff
@@ -1446,6 +1446,17 @@ const SSL_CIPHER *SSL_get_cipher_by_value(uint16_t value) {
       ssl_cipher_id_cmp));
 }
 
+const SSL_CIPHER *SSL_CIPHER_find(SSL *ssl, const unsigned char *ptr) {
```
andrewhop, line 1449: Is this supposed to use the ssl parameter for anything? Like only return the matching cipher if it's enabled in ssl?

ssl/ssl_cipher.cc
```diff
@@ -1446,6 +1446,17 @@ const SSL_CIPHER *SSL_get_cipher_by_value(uint16_t value) {
...
+  if (ptr) {
+    // Taking first element and left shit by 8 bits to set upper 8 bits.
+    // Bitwise OR with second element to set lower 8 bits of 16 bit integer.
+    uint16_t cipher_id = static_cast<uint16_t>((ptr[0] << 8) | ptr[1]);;
```
andrewhop, line 1453: This just looks a little weird/not common for AWS-LC, can this use one of your new loading functions?

ssl/internal.h
```diff
@@ -4067,6 +4067,11 @@ struct ssl_st {
   // serialized and is only populated if used in a server context.
   bssl::UniquePtr<STACK_OF(SSL_CIPHER)> client_cipher_suites;
 
+  // client_cipher_suites_arr is initialized when |SSL_client_hello_get0_ciphers|
+  // is called with a valid |out| param. It holds the cipher suites offered by the
+  // client from |client_cipher_suites| in an array.
+  bssl::Array<uint16_t> client_cipher_suites_arr;
```
andrewhop, line 4073: Does this need to be reset in SSL_clear?

ssl/ssl_test.cc
```diff
@@ -2379,6 +2379,79 @@ static bool ConnectClientAndServer(bssl::UniquePtr<SSL> *out_client,
...
+
+  // Client calling, should return 0
+  ASSERT_EQ(SSL_client_hello_get0_ciphers(client.get(), nullptr), (size_t) 0);
+
+  const unsigned char *p;
+  const unsigned char expected[] = {0x13, 0x01, 0x13, 0x02, 0x13, 0x03};
+
+  // Get client ciphers and ensure written to out in appropriate format
+  ASSERT_EQ(SSL_client_hello_get0_ciphers(server.get(), &p), (size_t) 3);
+  ASSERT_EQ(OPENSSL_memcmp(expected, p, sizeof(expected)), 0);
```
andrewhop, line 2447: NP: using the `Bytes` class will make nicer error messages if the values aren't the same. This is especially helpful if they are reversed and you're debugging and endianess issue.

ssl/ssl_test.cc
```diff
@@ -2379,6 +2379,79 @@ static bool ConnectClientAndServer(bssl::UniquePtr<SSL> *out_client,
...
+  // Handshake not completed, getting ciphers should fail
+  ASSERT_FALSE(SSL_client_hello_get0_ciphers(client.get(), nullptr));
+  ASSERT_FALSE(SSL_client_hello_get0_ciphers(server.get(), nullptr));
+
+  ASSERT_TRUE(CompleteHandshakes(client.get(), server.get()));
+
+  // Client calling, should return 0
+  ASSERT_EQ(SSL_client_hello_get0_ciphers(client.get(), nullptr), (size_t) 0);
+
+  const unsigned char *p;
```
andrewhop, line 2442: You don't free p, which makes sense, but this conflicts with how I read the API documentation.

include/openssl/ssl.h
```diff
@@ -1462,6 +1462,11 @@ DEFINE_CONST_STACK_OF(SSL_CIPHER)
 // https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4.
 OPENSSL_EXPORT const SSL_CIPHER *SSL_get_cipher_by_value(uint16_t value);
 
+// SSL_CIPHER_FIND re-casts |ptr| to uint16_t type and calls |SSL_get_cipher_by_value|.
```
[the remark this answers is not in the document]
smittals2: I'll modify the documentation here so it makes more sense. The |ptr| variable is a two element char array which is passed into SSL_CIPHER_FIND. In AWS-LC, we have a SSL_get_cipher_by_value function which takes in a uint16_t. So the plumbing is two byte array ptr into SSL_CIPHER_FIND -> cast to uint16_t -> Pass into SSL_get_cipher_by_value.

ssl/ssl_lib.cc
smittals2: If out is not null (so caller is requesting writing the ciphers to out), array initialization would only fail if there is a memory allocation error. So if this fails, error should be raised? We could also choose to return 0?

include/openssl/ssl.h
smittals2: fixed, from previous implementation of the function

ssl/ssl_test.cc
smittals2: fixed

ssl/ssl_lib.cc
smittals2: Based on my research this should be generally safe. When casting to unsigned char *, the uint16_t is treated as an array of bytes. So the difference may just be it takes 2 bytes of memory, 2 elements of the array. Since we're using CRYPTO_store_u16_be to explicitly specify order, the endianness should be correct.

ssl/ssl_cipher.cc
smittals2: In OpenSSL, the get cipher by value function is stored in ssl->method. For AWS-LC, we have a function defined separately which SSL_CIPHER_find is wrapping. So in our case, it's essentially just there because of the function signature.

ssl/ssl_cipher.cc
smittals2: yes we can, made the fix

ssl/ssl_test.cc
smittals2: Thanks for the tip!

ssl/internal.h
smittals2: Looking at the function, since stack is being reset it would make sense to reset the array as well - added.

ssl/ssl_lib.cc
andrewhop: Yeah if memory allocation fails this function should gracefully handle it and return 0

ssl/ssl_lib.cc
```diff
@@ -3315,3 +3316,34 @@ int SSL_CTX_set1_curves_list(SSL_CTX *ctx, const char *curves) {
...
+  STACK_OF(SSL_CIPHER) *client_cipher_suites = SSL_get_client_ciphers(ssl);
+  if (client_cipher_suites == nullptr) {
+      return 0;
+  }
+
+  size_t num_ciphers = sk_SSL_CIPHER_num(client_cipher_suites);
+  if (out != nullptr) {
+    // Called before
+    if(!ssl->client_cipher_suites_arr.empty()) {
+      *out = reinterpret_cast<unsigned char*>(ssl->client_cipher_suites_arr.data());
```
andrewhop, line 3330: NP: I think this would be an interesting sanity test to check in our tests but not necessarily fail in production.
```suggestion
      assert(ssl->client_cipher_suites_arr.size() == num_ciphers);
      *out = reinterpret_cast<unsigned char*>(ssl->client_cipher_suites_arr.data());
```

ssl/ssl_cipher.cc
justsmth: Even though we don't use the `ssl` parameter, should we still require that it not be null to provide behavior more similar to OpenSSL?

andrewhop (dismissed):

ssl/ssl_lib.cc
```diff
@@ -3315,3 +3316,37 @@ int SSL_CTX_set1_curves_list(SSL_CTX *ctx, const char *curves) {
...
+    // Called before
+    if(!ssl->client_cipher_suites_arr.empty()) {
+      assert(ssl->client_cipher_suites_arr.size() == num_ciphers);
+      *out = reinterpret_cast<unsigned char*>(ssl->client_cipher_suites_arr.data());
+      return ssl->client_cipher_suites_arr.size();
+    }
+
+    if(!ssl->client_cipher_suites_arr.Init(num_ciphers)) {
+       return 0;
+    }
```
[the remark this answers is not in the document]
smittals2: That function returns false and doesn't throw an error so I've added one here!

ssl/ssl_cipher.cc
smittals2: added a condition to check

ssl/ssl_lib.cc
```diff
@@ -3315,3 +3316,37 @@ int SSL_CTX_set1_curves_list(SSL_CTX *ctx, const char *curves) {
...
+size_t SSL_client_hello_get0_ciphers(SSL *ssl, const unsigned char **out) {
+  STACK_OF(SSL_CIPHER) *client_cipher_suites = SSL_get_client_ciphers(ssl);
+  if (client_cipher_suites == nullptr) {
+      return 0;
+  }
+
+  size_t num_ciphers = sk_SSL_CIPHER_num(client_cipher_suites);
+  if (out != nullptr) {
+    // Called before
+    if(!ssl->client_cipher_suites_arr.empty()) {
```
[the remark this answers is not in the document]
smittals2: Good point, made the change!

ssl/ssl_lib.cc
```diff
@@ -3315,3 +3316,35 @@ int SSL_CTX_set1_curves_list(SSL_CTX *ctx, const char *curves) {
...
+  if (client_cipher_suites == nullptr) {
+      return 0;
+  }
+
+  size_t num_ciphers = sk_SSL_CIPHER_num(client_cipher_suites);
+  if (out != nullptr) {
+    // Hasn't been called before
+    if(ssl->client_cipher_suites_arr.empty()) {
+      if (!ssl->client_cipher_suites_arr.Init(num_ciphers)) {
+        OPENSSL_PUT_ERROR(SSL, ERR_R_MALLOC_FAILURE);
```
andrewhop, line 3331: You don't need to add this error here, Array.Init() calls OpenSSL_malloc which already adds an error if the system malloc fails. 
```suggestion
```

justsmth: `OPENSSL_malloc` [reports the error](https://github.com/aws/aws-lc/blob/637afcda8f495ade222e29e5e2e79ff770bf0211/crypto/mem.c#L201) against the `CRYPTO` library.
```
  OPENSSL_PUT_ERROR(CRYPTO, ERR_R_MALLOC_FAILURE);
```

There are [many other places](https://github.com/search?q=repo%3Aaws%2Faws-lc+path%3A%2F%5Essl%5C%2F%2F+ERR_R_MALLOC_FAILURE&type=code) that we also report this error for SSL.

justsmth (approved):

ssl/ssl_lib.cc
justsmth: (Whoops, I deleted my previous comment by accident.)

We set the `ERR_R_MALLOC_FAILURE` for `SSL` in several other places as well: https://github.com/search?q=repo%3Aaws%2Faws-lc+path%3A%2Fssl%2F+ERR_R_MALLOC_FAILURE&type=code

smittals2: If malloc is already doing so, is there a specific reason for this redundancy? Maybe to ensure that the trace for the error stack is more detailed?

justsmth: Malloc reports the error against the "CRYPTO" module, not the "SSL" module as is done here. That seems to be the pattern but I'm. It sure how consistent it is.

smittals2: For clearer error logging, I'll leave this here

andrewhop: TIL

andrewhop (approved):

PR #1486, opened 2024-03-11
Add ASN1_i2d_bio and ASN1_i2d_bio_of back

samuel40791765:
### Description of changes: 
tmp2-tools defines some of it's own ASN1 structures and [depends](https://github.com/tpm2-software/tpm2-tools/blob/beaecabac2e9c1f26afcb5658c2cdf244c6ad8cc/tools/misc/tpm2_certifyX509certutil.c#L51) on some removed functions & macros.
The removal of these were spread across https://github.com/aws/aws-lc/commit/792c1dc43eb888470b2ecb162be2acc5e0e1d61b and https://github.com/aws/aws-lc/commit/419144adce049b5341bd94d355c52d099eac56e3. 

### Call-outs:
N/A

### Testing:
Parsing and verifying the contents against our existing ASN1 functions.

### DIFF
--- a/crypto/asn1/a_i2d_fp.c
+++ b/crypto/asn1/a_i2d_fp.c
@@ -54,6 +54,8 @@
  * copied and put under another distribution licence
  * [including the GNU Public Licence.] */
 
+#include <assert.h>
+
 #include <openssl/asn1.h>
 
 #include <openssl/bio.h>
@@ -69,25 +71,32 @@ int ASN1_i2d_bio(i2d_of_void *i2d, BIO *out, void *in) {
 
   int size = i2d(in, NULL);
   if (size <= 0) {
+    OPENSSL_PUT_ERROR(ASN1, ERR_R_MALLOC_FAILURE);
     return 0;
   }
 
-  char *buffer = (char *)OPENSSL_malloc(size);
+  unsigned char *buffer = (unsigned char *)OPENSSL_malloc(size);
   if (buffer == NULL) {
     OPENSSL_PUT_ERROR(ASN1, ERR_R_MALLOC_FAILURE);
     return 0;
   }
 
-  unsigned char *outp = (unsigned char *)buffer;
-  i2d(in, &outp);
+  unsigned char *outp = buffer;
+  int ret = i2d(in, &outp);
+  if (ret < 0 || ret > size) {
+    OPENSSL_PUT_ERROR(ASN1, ASN1_R_BUFFER_TOO_SMALL);
+    return 0;
+  }
 
-  int written = 0, idx=0;
+  int idx=0;
   for (;;) {
-    written = BIO_write(out, &(buffer[idx]), size);
+    int written = BIO_write(out, &(buffer[idx]), size);
+    assert(written <= size);
     if (written == size) {
       break;
     }
     if (written <= 0) {
+      OPENSSL_PUT_ERROR(ASN1, ASN1_R_BUFFER_TOO_SMALL);
       OPENSSL_free(buffer);
       return 0;
     }
--- a/crypto/asn1/a_i2d_fp.c
+++ b/crypto/asn1/a_i2d_fp.c
@@ -61,6 +61,43 @@
 #include <openssl/mem.h>
 
 
+int ASN1_i2d_bio(i2d_of_void *i2d, BIO *out, void *in) {
+  if (i2d == NULL || out == NULL || in == NULL) {
+    OPENSSL_PUT_ERROR(ASN1, ERR_R_PASSED_NULL_PARAMETER);
+    return 0;
+  }
+
+  int size = i2d(in, NULL);
+  if (size <= 0) {
+    return 0;
+  }
+
+  char *buffer = (char *)OPENSSL_malloc(size);
+  if (buffer == NULL) {
+    OPENSSL_PUT_ERROR(ASN1, ERR_R_MALLOC_FAILURE);
+    return 0;
+  }
+
+  unsigned char *outp = (unsigned char *)buffer;
+  i2d(in, &outp);
+
+  int written = 0, idx=0;
+  for (;;) {
+    written = BIO_write(out, &(buffer[idx]), size);
+    if (written == size) {
+      break;
+    }
+    if (written <= 0) {
+      OPENSSL_free(buffer);
+      return 0;
+    }
+    idx += written;
+    size -= written;
+  }
+  OPENSSL_free(buffer);
+  return 1;
+}
+
 int ASN1_item_i2d_fp(const ASN1_ITEM *it, FILE *out, void *x) {
   BIO *b = BIO_new_fp(out, BIO_NOCLOSE);
   if (b == NULL) {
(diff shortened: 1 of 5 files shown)


skmcgrail (dismissed):

crypto/asn1/a_i2d_fp.c
```diff
@@ -61,6 +61,43 @@
...
+  unsigned char *outp = (unsigned char *)buffer;
+  i2d(in, &outp);
+
+  int written = 0, idx=0;
+  for (;;) {
+    written = BIO_write(out, &(buffer[idx]), size);
+    if (written == size) {
+      break;
+    }
+    if (written <= 0) {
```
justsmth, line 90: * Should we call `OPENSSL_PUT_ERROR` here?
* `BIO_write` returning `0` is not an error, it just means that no bytes were written. How should we handle that?

crypto/asn1/a_i2d_fp.c
```diff
@@ -61,6 +61,43 @@
...
+    OPENSSL_PUT_ERROR(ASN1, ERR_R_PASSED_NULL_PARAMETER);
+    return 0;
+  }
+
+  int size = i2d(in, NULL);
+  if (size <= 0) {
+    return 0;
+  }
+
+  char *buffer = (char *)OPENSSL_malloc(size);
```
justsmth, line 75: Why not use `unsigned char *` as the type for `buffer` and avoid the need for the subsequent cast?

crypto/asn1/a_i2d_fp.c
```diff
@@ -61,6 +61,43 @@
...
+    return 0;
+  }
+
+  unsigned char *outp = (unsigned char *)buffer;
+  i2d(in, &outp);
+
+  int written = 0, idx=0;
+  for (;;) {
+    written = BIO_write(out, &(buffer[idx]), size);
+    if (written == size) {
```
justsmth, line 87: Should we debug assert that `written <= size`?

crypto/asn1/a_i2d_fp.c
```diff
@@ -61,6 +61,43 @@
...
 
 
+int ASN1_i2d_bio(i2d_of_void *i2d, BIO *out, void *in) {
+  if (i2d == NULL || out == NULL || in == NULL) {
+    OPENSSL_PUT_ERROR(ASN1, ERR_R_PASSED_NULL_PARAMETER);
+    return 0;
+  }
+
+  int size = i2d(in, NULL);
+  if (size <= 0) {
```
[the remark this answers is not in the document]
samuel40791765: * Added 
* I'm not exactly sure if there's a good answer for a maximum size. I'd also be wary of putting an upper bound on here. That would require triaging on how these legacy APIs are being consumed and putting any bound here would essentially be diverging from the upstream's original function behavior.
https://github.com/openssl/openssl/blob/f08be096517f9bdae8a9d1d837748237db4d13a9/crypto/asn1/a_i2d_fp.c#L40-L42

crypto/asn1/a_i2d_fp.c
samuel40791765: * Added
* Even though returning 0 isn't an actual error from `BIO_write`, this still means that the bytes weren't all properly written to `out`. The previous block of code checks if all `size` bytes was written to `bio`.

crypto/asn1/a_i2d_fp.c
samuel40791765: Added

crypto/asn1/a_i2d_fp.c
```diff
@@ -61,6 +61,43 @@
...
+  }
+
+  char *buffer = (char *)OPENSSL_malloc(size);
+  if (buffer == NULL) {
+    OPENSSL_PUT_ERROR(ASN1, ERR_R_MALLOC_FAILURE);
+    return 0;
+  }
+
+  unsigned char *outp = (unsigned char *)buffer;
+  i2d(in, &outp);
```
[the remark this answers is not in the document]
samuel40791765: Sure, added a check for both

crypto/asn1/a_i2d_fp.c
samuel40791765: Consolidating this block to use `BIO_write_all` instead, like much of the other related APIs are using.

crypto/asn1/a_i2d_fp.c
```diff
@@ -54,13 +54,45 @@
...
+  unsigned char *buffer = (unsigned char *)OPENSSL_malloc(size);
+  if (buffer == NULL) {
+    OPENSSL_PUT_ERROR(ASN1, ERR_R_MALLOC_FAILURE);
+    return 0;
+  }
+
+  unsigned char *outp = buffer;
+  int ret = i2d(in, &outp);
+  if (ret < 0 || ret > size) {
+    OPENSSL_PUT_ERROR(ASN1, ASN1_R_BUFFER_TOO_SMALL);
```
justsmth, line 87: Free buffer before return.

crypto/asn1/a_i2d_fp.c
```diff
@@ -54,13 +54,45 @@
...
 
+int ASN1_i2d_bio(i2d_of_void *i2d, BIO *out, void *in) {
+  if (i2d == NULL || out == NULL || in == NULL) {
+    OPENSSL_PUT_ERROR(ASN1, ERR_R_PASSED_NULL_PARAMETER);
+    return 0;
+  }
+
+  int size = i2d(in, NULL);
+  if (size <= 0) {
+    OPENSSL_PUT_ERROR(ASN1, ERR_R_MALLOC_FAILURE);
```
justsmth, line 74: Is this the right error to use?

samuel40791765: Changed to INTERNAL_ERROR

justsmth (approved):

crypto/bio/bio.c
```diff
@@ -240,7 +240,9 @@ int BIO_write_all(BIO *bio, const void *data, size_t len) {
   const uint8_t *data_u8 = data;
   while (len > 0) {
     int ret = BIO_write(bio, data_u8, len > INT_MAX ? INT_MAX : (int)len);
+    assert(ret <= (int)len);
```
justsmth, line 243: NP: The current assert would fail when `len > INT_MAX`.
```
    const int write_len = ((len > INT_MAX) ? INT_MAX : (int)len);
    int ret = BIO_write(bio, data_u8, write_len);
    assert(ret <= write_len);
```

samuel40791765: Would it? If `len > INT_MAX` was true, `BIO_write` would return a maximum of `INT_MAX` bytes if all were written. 
 In my mind it's `ret <= len < INT_MAX`.
I don't like the existing `int` cast, so I'll take your change though.

justsmth (approved):

crypto/bio/bio.c
samuel40791765: ahh that cast behavior was the logic I was overlooking. Thanks for clearing things up!

skmcgrail (approved):

PR #1492, opened 2024-03-13
Support for MinGW on Windows

justinwsmith:
### Issues
* Related to: aws/aws-lc-rs#370
* CryptoAlg-1323

### Description of changes: 
* Attempting to build using gcc/mingw surfaced a few minor issues.
* Adding build/testing of Windows MinGW to our CI tests

### Testing
All tests passed on my Windows Server 2022 Datacenter EC2 instance.


ssl/ssl_test.cc
```diff
@@ -12112,10 +12112,10 @@ TEST(SSLTest, SSLFileTests) {
...
 
 #if defined(OPENSSL_WINDOWS)
-  char rsa_pem_filename[L_tmpnam];
-  char ecdsa_pem_filename[L_tmpnam];
-  ASSERT_EQ(tmpnam_s(rsa_pem_filename, sizeof(rsa_pem_filename)), 0);
-  ASSERT_EQ(tmpnam_s(ecdsa_pem_filename, sizeof(ecdsa_pem_filename)), 0);
+  char rsa_pem_filename[FILENAME_MAX];
+  char ecdsa_pem_filename[FILENAME_MAX];
+  ASSERT_EQ(tmpnam_s(rsa_pem_filename, FILENAME_MAX), 0);
+  ASSERT_EQ(tmpnam_s(ecdsa_pem_filename, FILENAME_MAX), 0);
```
andrewhop, line 12118: Interesting, why does this need to change to `FILENAME_MAX`? Does `sizeof(rsa_pem_filename)` return the wrong length?

.github/workflows/mingw.yml
```diff
@@ -0,0 +1,54 @@
...
+    branches: [ '*' ]
+  push:
+    branches: [ '*' ]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+jobs:
+  mingw:
+    if: github.repository == 'aws/aws-lc'
```
andrewhop, line 13: Should we add this to our other jobs? Is there a reason not to let people run this CI on there forks?

.github/workflows/mingw.yml
```diff
@@ -0,0 +1,54 @@
...
+            CMAKE_SYSTEM_NAME=Windows \
+            CMAKE_SYSTEM_PROCESSOR=x86_64 \
+            CMAKE_BUILD_TOOL=C:/ProgramData/chocolatey/lib/mingw/tools/install/mingw64/bin/ninja.exe \
+            CMAKE_FIND_ROOT_PATH=C:/ProgramData/chocolatey/lib/mingw/tools/install/mingw64 \
+            CMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER \
+            CMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY \
+            CMAKE_FIND_ROOT_PATH_MODE_INCLUDE=ONLY \
+      - name: Build Project
+        run: cmake --build ./build --target all
+      - name: Run crypto_test
```
andrewhop, line 41: Why not run all the tests with `ninja run_tests`? Which also does it all in parallel

third_party/jitterentropy/jitterentropy-base-user.h
```diff
@@ -97,7 +97,7 @@
 
 #ifdef __x86_64__
 # define DECLARE_ARGS(val, low, high)    unsigned long low, high
-# define EAX_EDX_VAL(val, low, high)     ((low) | (high) << 32)
+# define EAX_EDX_VAL(val, low, high)     (((uint64_t)low) | ((uint64_t)high) << 32)
```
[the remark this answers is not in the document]
justsmth: On some platforms "unsigned long" is only 4 byte (32-bits). The compiler was warning about left-shifting an "unsigned long" value by 32-bits.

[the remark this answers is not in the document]
justsmth: This change is making the cast to uint64_t explicit, when it was being done implicitly.

ssl/ssl_test.cc
justsmth: The mingw header files were using a value of 12, which was too small and causing the function to fail.

.github/workflows/mingw.yml
justsmth: Yeah, we can use that instead.

ssl/ssl_test.cc
justsmth: Microsoft's documentation says that the parameter is the size of the buffer "in characters", which should be the same but when it was failing I decided to be pedantic with it.

third_party/jitterentropy/jitterentropy-base-user.h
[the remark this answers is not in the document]
justsmth: > Should we just update the types in DECLARE_ARGS?

This is "third party" source, so I was trying to minimize the change to only what's necessary to fix the compile error.

.github/workflows/mingw.yml
andrewhop: Isn't it only expensive if they open PR's to their own repo? Which raises the question why is this running on pushes?

.github/workflows/mingw.yml
```diff
@@ -0,0 +1,42 @@
+name: MinGW
+on:
+  pull_request:
+    branches: [ '*' ]
+  push:
+    branches: [ '*' ]
```
andrewhop, line 6: Why pushes? This also seems inconsistent in our other GitHub workflows.

andrewhop (commented):
Everything looks good, just the one question on push or not push.

.github/workflows/mingw.yml
justsmth: 6 of the 8 workflows run on pushes. Running on "push" allows it to run after a PR has been merged to main with a rebase.

skmcgrail (approved):

third_party/jitterentropy/jitterentropy-base-user.h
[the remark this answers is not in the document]
andrewhop: Fair, in a future update we can take a new version of jitter which correctly declares these as uint64_t https://github.com/smuellerDD/jitterentropy-library/blob/master/jitterentropy-base-user.h#L100-L103.

skmcgrail (approved):

andrewhop (approved):

PR #1496, opened 2024-03-15
Various minor functions to support mysql 8.3

samuel40791765:
### Issues:
Addresses `V1073444663`

### Description of changes: 
MySQL 8.3 depends on some minor symbols that wrap around our existing functionality. 
1. `sk_X509_unshift`: https://github.com/mysql/mysql-server/blob/mysql-cluster-8.3.0/storage/ndb/src/common/util/NodeCertificate.cpp#L1167-L1173
2. `X509_REQ_get0_pubkey`: https://github.com/mysql/mysql-server/blob/824e2b4064053f7daf17d7f3f84b7a3ed92e5fb4/storage/ndb/src/common/util/NodeCertificate.cpp#L526
3. `SSL_CTX_set1_cert_store`: https://github.com/mysql/mysql-server/blob/824e2b4064053f7daf17d7f3f84b7a3ed92e5fb4/storage/ndb/src/common/util/TlsKeyManager.cpp#L204

There are more symbols incoming, these only encapsulate the less complicated ones.

### Call-outs:
N/A

### Testing:
Build still needs to be fixed for MySQL 8.3, along with a couple more symbols.

### DIFF
--- a/crypto/x509/x509_req.c
+++ b/crypto/x509/x509_req.c
@@ -78,6 +78,7 @@ X509_NAME *X509_REQ_get_subject_name(const X509_REQ *req) {
 
 EVP_PKEY *X509_REQ_get_pubkey(X509_REQ *req) {
   if ((req == NULL) || (req->req_info == NULL)) {
+    OPENSSL_PUT_ERROR(X509, ERR_R_PASSED_NULL_PARAMETER);
     return NULL;
   }
   return (X509_PUBKEY_get(req->req_info->pubkey));
@@ -85,6 +86,7 @@ EVP_PKEY *X509_REQ_get_pubkey(X509_REQ *req) {
 
 EVP_PKEY *X509_REQ_get0_pubkey(X509_REQ *req) {
   if ((req == NULL) || (req->req_info == NULL)) {
+    OPENSSL_PUT_ERROR(X509, ERR_R_PASSED_NULL_PARAMETER);
     return NULL;
   }
   return (X509_PUBKEY_get0(req->req_info->pubkey));
--- a/crypto/x509/x509_req.c
+++ b/crypto/x509/x509_req.c
@@ -83,6 +83,13 @@ EVP_PKEY *X509_REQ_get_pubkey(X509_REQ *req) {
   return (X509_PUBKEY_get(req->req_info->pubkey));
 }
 
+EVP_PKEY *X509_REQ_get0_pubkey(X509_REQ *req) {
+  if ((req == NULL) || (req->req_info == NULL)) {
+    return NULL;
+  }
+  return (X509_PUBKEY_get0(req->req_info->pubkey));
+}
+
 int X509_REQ_check_private_key(X509_REQ *x, EVP_PKEY *k) {
   EVP_PKEY *xk = NULL;
   int ok = 0;
--- a/ssl/ssl_x509.cc
+++ b/ssl/ssl_x509.cc
@@ -160,13 +160,15 @@ BSSL_NAMESPACE_BEGIN
 // installed. Calling an X509-based method on an |ssl| with a different method
 // will likely misbehave and possibly crash or leak memory.
 static void check_ssl_x509_method(const SSL *ssl) {
-  assert(ssl == NULL || ssl->ctx->x509_method == &ssl_crypto_x509_method);
+  assert(ssl == NULL || (ssl != NULL && ssl->ctx != NULL &&
+                         ssl->ctx->x509_method == &ssl_crypto_x509_method));
 }
 
 // check_ssl_ctx_x509_method acts like |check_ssl_x509_method|, but for an
 // |SSL_CTX|.
 static void check_ssl_ctx_x509_method(const SSL_CTX *ctx) {
-  assert(ctx == NULL || ctx->x509_method == &ssl_crypto_x509_method);
+  assert(ctx == NULL ||
+         (ctx != NULL && ctx->x509_method == &ssl_crypto_x509_method));
 }
 
 // x509_to_buffer returns a |CRYPTO_BUFFER| that contains the serialised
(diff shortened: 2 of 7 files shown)


crypto/x509/x509_req.c
```diff
@@ -83,6 +83,13 @@ EVP_PKEY *X509_REQ_get_pubkey(X509_REQ *req) {
   return (X509_PUBKEY_get(req->req_info->pubkey));
 }
 
+EVP_PKEY *X509_REQ_get0_pubkey(X509_REQ *req) {
+  if ((req == NULL) || (req->req_info == NULL)) {
```
justsmth, line 87: Set an error?
```
OPENSSL_PUT_ERROR(X509, ERR_R_PASSED_NULL_PARAMETER);
```

ssl/ssl_x509.cc
```diff
@@ -765,6 +765,14 @@ void SSL_CTX_set_cert_store(SSL_CTX *ctx, X509_STORE *store) {
   ctx->cert_store = store;
 }
 
+void SSL_CTX_set1_cert_store(SSL_CTX *ctx, X509_STORE *store) {
+  check_ssl_ctx_x509_method(ctx);
```
justsmth, line 769: `check_ssl_ctx_x509_method` succeeds when `ctx` is null, [it only ensures](https://github.com/aws/aws-lc/blob/2352c50a6b5b95c12aed8b92c5881def3218a646/ssl/ssl_x509.cc#L169) when it's not null that `ctx-.x509_method` is set correctly.

ssl/ssl_x509.cc
```diff
@@ -765,6 +765,14 @@ void SSL_CTX_set_cert_store(SSL_CTX *ctx, X509_STORE *store) {
...
+  if (store != nullptr) {
+    X509_STORE_up_ref(store);
+  }
+  SSL_CTX_set_cert_store(ctx, store);
```
justsmth, line 773: There should be a check in [SSL_CTX_set_cert_store](https://github.com/aws/aws-lc/blob/2352c50a6b5b95c12aed8b92c5881def3218a646/ssl/ssl_x509.cc#L764) that `ctx` is not null prior to dereferencing it.

justsmth: Actually, the check for `ctx == NULL` needs to occur prior to `X509_STORE_up_ref`, so it should be here as well.

justsmth: These functions return `void` so there's no way to inform the caller of the failure.  We should have `assert` statements, so at least for DEBUG builds consumers can be alerted to the failure.
```
assert(ctx != NULL);
```

ssl/ssl_x509.cc
samuel40791765: Yes it's also called at the start of every publicly accessible function in this file.

ssl/ssl_x509.cc
samuel40791765: added assertion to `check_ssl_ctx_x509_method`

ssl/ssl_x509.cc
```diff
@@ -160,13 +160,15 @@ BSSL_NAMESPACE_BEGIN
 // installed. Calling an X509-based method on an |ssl| with a different method
 // will likely misbehave and possibly crash or leak memory.
 static void check_ssl_x509_method(const SSL *ssl) {
-  assert(ssl == NULL || ssl->ctx->x509_method == &ssl_crypto_x509_method);
+  assert(ssl == NULL || (ssl != NULL && ssl->ctx != NULL &&
```
skmcgrail, line 163: Should we be using nullptr over NULL as we rework code? I notice further down we use nullptr for the check rather then NULL.

samuel40791765: I'm on team nullptr as well for C++. I got sucked into following the original format for this, but I'll make the change 👍

ssl/ssl_x509.cc
```diff
@@ -160,13 +160,15 @@ BSSL_NAMESPACE_BEGIN
...
+  assert(ssl == nullptr || (ssl != nullptr && ssl->ctx != nullptr &&
+                            ssl->ctx->x509_method == &ssl_crypto_x509_method));
 }
 
 // check_ssl_ctx_x509_method acts like |check_ssl_x509_method|, but for an
 // |SSL_CTX|.
 static void check_ssl_ctx_x509_method(const SSL_CTX *ctx) {
-  assert(ctx == NULL || ctx->x509_method == &ssl_crypto_x509_method);
+  assert(ctx == nullptr ||
+         (ctx != nullptr && ctx->x509_method == &ssl_crypto_x509_method));
```
justsmth, line 171: I don't think this change actually did anything -- the `ctx->x509_method == &ssl_crypto_x509_method` was already only being evaluated when `ctx != nullptr`.

ssl/ssl_x509.cc
```diff
@@ -160,13 +160,15 @@ BSSL_NAMESPACE_BEGIN
 // installed. Calling an X509-based method on an |ssl| with a different method
 // will likely misbehave and possibly crash or leak memory.
 static void check_ssl_x509_method(const SSL *ssl) {
-  assert(ssl == NULL || ssl->ctx->x509_method == &ssl_crypto_x509_method);
+  assert(ssl == nullptr || (ssl != nullptr && ssl->ctx != nullptr &&
+                            ssl->ctx->x509_method == &ssl_crypto_x509_method));
```
justsmth, line 164: The `ssl != nullptr` is not needed. 

Should this fail when `ssl->ctx == nullptr`? Should this be:
```
  assert(ssl == nullptr || 
            ssl->ctx == nullptr ||
                ssl->ctx->x509_method == &ssl_crypto_x509_method));
```
?

samuel40791765: I'm not 100% sure and I'll just remove this change since this is getting out of scope of this particular PR. Asserts against null checks in libssl are definitely nice, but we should open this as another task and resolve these all at once.

ssl/ssl_x509.cc
samuel40791765: Ahh nice catch, I’ll just add the asserts in the designated functions we’re handling

justsmth (approved):

smittals2 (approved):

skmcgrail (approved):

PR #1497, opened 2024-03-15
Remove build patch for bind9

samuel40791765:
### Description of changes: 
Build patch for Bind was submitted upstream, so we don't need the patch anymore. We now build with tip of main of Bind9 without any patches.
* https://gitlab.isc.org/isc-projects/bind9/-/issues/4639

### Call-outs:
N/A

### Testing:
CI


tests/ci/integration/run_bind9_integration.sh
```diff
@@ -34,14 +34,6 @@ function bind9_build() {
   make -j ${NUM_CPU_THREADS} -k all
 }
 
-# TODO: Remove this when we make an upstream contribution.
-function bind9_patch() {
```
justsmth, line 38: Could we leave this logic in place in case it's ever needed in the future?  Simply removing the patch file would make this a no-op.

samuel40791765: Ideally we shouldn't be needing patches to build with integrations. In this case, it was a build fix for Bind which was unrelated to AWS-LC.
If we really need to patch for some new feature Bind9 implements that we don't wish to support, it should be quite easy to add the original logic in place. We already have existing code in other files that do similar things. I just got the idea that we could extract these "patch" functions into a general bash function though.

justsmth (approved):

WillChilds-Klein (approved):

PR #1504, opened 2024-03-21
add support for X509_get_signature_info

samuel40791765:
### Issues:
Resolves `CryptoAlg-2264`

### Description of changes: 
MySQL 8.2/8.3 depends on [`X509_get_signature_info`](https://github.com/mysql/mysql-server/blob/mysql-cluster-8.2.0/storage/ndb/src/common/util/NodeCertificate.cpp#LL1290). This symbol originated from an OpenSSL 1.1.1 commit (https://github.com/openssl/openssl/commit/786dd2c22c71081492e209d93beee3ff4fe66357) implementing support for "custom signature parameters".
The commit mainly allows the user to set and retrieve information about the signature type. There's also a custom ASN1 method handler that OpenSSL implemented in this commit.
We don't support custom `EVP_PKEY` ASN1 methods and setting your own signature information doesn't make much sense without the custom methods, so I've only implemented support for retrieving the signature type. MySQL only depends on retrieving the signature info, so this fits our use case.

### Call-outs:
`CryptoAlg-2393` was cut to track RSA PSS support in libssl, which was uncovered in the postgres CI.

### Testing:
New test


crypto/x509/x509_set.c
```diff
@@ -238,3 +239,67 @@ const X509_ALGOR *X509_get0_tbs_sigalg(const X509 *x) {
...
+  if (sec_bits != NULL) {
+    *sec_bits = sig_info->sec_bits;
+  }
+  if (flags != NULL) {
+    *flags = sig_info->flags;
+  }
+  return (sig_info->flags & X509_SIG_INFO_VALID) != 0;
+}
+
+int X509_get_signature_info(X509 *x509, int *digest_nid, int *pubkey_nid,
```
skmcgrail, line 260: Should check that `x509 != NULL` ?

crypto/x509/x509_set.c
```diff
@@ -238,3 +239,67 @@ const X509_ALGOR *X509_get0_tbs_sigalg(const X509 *x) {
...
+                            int *sec_bits, uint32_t *flags) {
+  if (!X509_check_purpose(x509, -1, -1)) {
+    OPENSSL_PUT_ERROR(X509, X509_V_ERR_INVALID_PURPOSE);
+    return 0;
+  }
+  return X509_SIG_INFO_get(&x509->sig_info, digest_nid, pubkey_nid, sec_bits,
+                           flags);
+}
+
+int x509_init_signature_info(X509 *x509) {
```
skmcgrail, line 270: Don't recall if we were only adding the check on external functions first, but if we are also adding to same checks to internal functions then should this validated `x509 != NULL`?

crypto/x509/x509_set.c
```diff
@@ -238,3 +239,67 @@ const X509_ALGOR *X509_get0_tbs_sigalg(const X509 *x) {
...
+  }
+  if (flags != NULL) {
+    *flags = sig_info->flags;
+  }
+  return (sig_info->flags & X509_SIG_INFO_VALID) != 0;
+}
+
+int X509_get_signature_info(X509 *x509, int *digest_nid, int *pubkey_nid,
+                            int *sec_bits, uint32_t *flags) {
+  if (!X509_check_purpose(x509, -1, -1)) {
```
skmcgrail, line 262: It seems like the intent of this is to just trigger `x509v3_cache_extensions`, as it will bail out of the function afterwards due to `id == -1`.  Would it be clearer to just call the cache function?

crypto/x509/x509_set.c
samuel40791765: This did come up a long while back, I think the final consensus was to cover external functions. The justification being that there would be NULL checks in the external functions calling these internal functions, so the extra check would be overkill and add additional bloat.

crypto/x509/x509_set.c
samuel40791765: Ahh nice catch, I was just following the original openssl implementation, but that makes sense. I'll also add [this comment](https://github.com/openssl/openssl/blame/master/crypto/x509/v3_purp.c#L83C3-L84) from OpenSSL to `X509_check_purpose`

samuel40791765: It seems like Postgres is also [dependent](https://github.com/postgres/postgres/blob/0075d78947e3800c5a807f48fd901f16db91101b/src/backend/libpq/be-secure-openssl.c#L1470-L1479) on this function and the new behavior is causing the postgres CI to fail..
I'm guessing that `X509_get_signature_info` is just a information retrieval function, so the cert is allowed to "invalid extensions".. which explains why [OpenSSL's implementation](https://github.com/openssl/openssl/blob/master/crypto/x509/x509_set.c#L204-L209) didn't check the return value. It's a bit of a pain, but I'll have to change this to be less error prone

samuel40791765: Nvm, the postgres CI was failing since the initial change. I'll have to figure out what's going on here

crypto/x509/x509_set.c
```diff
@@ -238,3 +240,72 @@ const X509_ALGOR *X509_get0_tbs_sigalg(const X509 *x) {
 X509_PUBKEY *X509_get_X509_PUBKEY(const X509 *x509) {
   return x509->cert_info->key;
 }
+
+static int X509_SIG_INFO_get(const X509_SIG_INFO *sig_info, int *digest_nid,
+                             int *pubkey_nid, int *sec_bits, uint32_t *flags) {
+  if (digest_nid != NULL) {
```
WillChilds-Klein, line 246: do we need a null check for `sig_info`?

crypto/x509/internal.h
```diff
@@ -139,10 +139,21 @@ typedef struct {
 // an |X509_NAME|.
 DECLARE_ASN1_FUNCTIONS(X509_CINF)
 
+struct x509_sig_info_st {
+  // NID of message digest.
+  int digest_nid;
+  // NID of public key algorithm.
+  int pubkey_nid;
+  // Security bits.
+  int sec_bits;
```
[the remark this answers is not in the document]
samuel40791765: Mainly so.. I was on the fence on tweaking this to `size_t`, but OpenSSL considers [the final output is still "valid"](https://github.com/openssl/openssl/commit/786dd2c22c71081492e209d93beee3ff4fe66357#diff-7b0a55e945112d8fee45ff63ec6b110acb696d8a8eb5fae013fff3a1aef49e7eR214-R223) when  `digest_nid` is `NID_undef`. When this happens, `sec_bits` is defined as `-1`. There are some examples of these circumstances in the tests I wrote.

crypto/x509/x509_set.c
samuel40791765: This is an external function in OpenSSL that we might end up exposing some day, so sure.

WillChilds-Klein (approved):

crypto/x509/x509_set.c
WillChilds-Klein: I see all CI checks have passed. What ended up being the issue with the postgres checks?

samuel40791765: I forgot to mention I cut `CryptoAlg-2393` to track PSS support in libssl. I'll add that in the commit/PR description.

skmcgrail (approved):

PR #1511, opened 2024-03-26
allow empty lists in SSL_CTX_set_ciphersuites

samuel40791765:
### Description of changes: 
MySQL deprecated the usage of weaker ciphers in their test suite: https://github.com/mysql/mysql-server/commit/bc14366ba3d1127ee6b1e0c84c21033e01b67a36 This brought an additional gap to light, where OpenSSL allows the setting of empty strings in the Ciphersuite string configuration. This is known behavior [since 1.1.1](https://github.com/openssl/openssl/blame/OpenSSL_1_1_1/ssl/ssl_ciph.c#L1311-L1315) and still [exists today](https://github.com/openssl/openssl/blob/master/ssl/ssl_ciph.c#L1348C27-L1353).

I've reconfigured the behavior for `SSL_[CTX]_set_cipher_list` which is our OpenSSL compat layer, but kept the same behavior for `SSL_[CTX]_set_strict_cipher_list`.

### Call-outs:
N/A

### Testing:
Test tweaking

### DIFF
--- a/ssl/ssl_cipher.cc
+++ b/ssl/ssl_cipher.cc
@@ -1336,9 +1336,11 @@ bool ssl_create_cipher_list(UniquePtr<SSLCipherPreferenceList> *out_cipher_list,
 
   *out_cipher_list = std::move(pref_list);
 
-  // Configuring an empty cipher list is an error but still updates the
-  // output.
-  if (sk_SSL_CIPHER_num((*out_cipher_list)->ciphers.get()) == 0) {
+  // Configuring an empty cipher list is an error when |strict| is true, but
+  // still updates the output. When otherwise, OpenSSL explicitly allows an
+  // empty list.
+  if ((strict || (*rule_str != '\0')) &&
+      sk_SSL_CIPHER_num((*out_cipher_list)->ciphers.get()) == 0) {
     OPENSSL_PUT_ERROR(SSL, SSL_R_NO_CIPHER_MATCH);
     return false;
   }
--- a/ssl/ssl_test.cc
+++ b/ssl/ssl_test.cc
@@ -5758,15 +5758,16 @@ TEST(SSLTest, EmptyCipherList) {
   // Configuring the empty cipher list with |SSL_CTX_set_cipher_list|
   // succeeds.
   EXPECT_TRUE(SSL_CTX_set_cipher_list(ctx.get(), ""));
-
   // The cipher list is updated to empty.
+  EXPECT_EQ(0u, sk_SSL_CIPHER_num(SSL_CTX_get_ciphers(ctx.get())));
+
+  // Configuring the empty cipher list with |SSL_CTX_set_ciphersuites|
+  // also succeeds.
+  EXPECT_TRUE(SSL_CTX_set_ciphersuites(ctx.get(), ""));
   EXPECT_EQ(0u, sk_SSL_CIPHER_num(SSL_CTX_get_ciphers(ctx.get())));
 
   // Configuring the empty cipher list with |SSL_CTX_set_strict_cipher_list|
   // fails.
-  EXPECT_FALSE(SSL_CTX_set_strict_cipher_list(ctx.get(), ""));
-  ERR_clear_error();
-
   EXPECT_FALSE(SSL_CTX_set_strict_cipher_list(ctx.get(), ""));
   ERR_clear_error();
 
--- a/ssl/ssl_test.cc
+++ b/ssl/ssl_test.cc
@@ -548,8 +548,6 @@ static const char *kBadRules[] = {
   "[+RSA]",
   // Unknown directive.
   "@BOGUS",
-  // Empty cipher lists error at SSL_CTX_set_cipher_list.
-  "",
   "BOGUS",
   // COMPLEMENTOFDEFAULT is empty.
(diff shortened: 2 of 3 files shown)


ssl/ssl_cipher.cc
```diff
@@ -1336,9 +1336,11 @@ bool ssl_create_cipher_list(UniquePtr<SSLCipherPreferenceList> *out_cipher_list,
 
   *out_cipher_list = std::move(pref_list);
 
-  // Configuring an empty cipher list is an error but still updates the
-  // output.
-  if (sk_SSL_CIPHER_num((*out_cipher_list)->ciphers.get()) == 0) {
+  // Configuring an empty cipher list is an error with |strict| but still
+  // updates the output. When otherwise, OpenSSL explicitly allows an empty
+  // list.
+  if ((strict || (*rule_str != '\0')) &&
```
smittals2, line 1342: NP: comment slight confusing, maybe just "Otherwise, OpenSSL explicitly..."

smittals2 (approved):

ssl/ssl_cipher.cc
samuel40791765: "When otherwise" was meant when "strict is not true", so I don't think that clarifies things better. I'll change the previous sentence and see if that helps

ssl/ssl_test.cc
```diff
@@ -5757,12 +5755,19 @@ TEST(SSLTest, EmptyCipherList) {
...
+  // The cipher list is updated to empty.
+  EXPECT_EQ(0u, sk_SSL_CIPHER_num(SSL_CTX_get_ciphers(ctx.get())));
+
+  // Configuring the empty cipher list with |SSL_CTX_set_strict_cipher_list|
+  // fails.
+  EXPECT_FALSE(SSL_CTX_set_strict_cipher_list(ctx.get(), ""));
   ERR_clear_error();
 
-  // Configuring the empty cipher list fails.
-  EXPECT_FALSE(SSL_CTX_set_ciphersuites(ctx.get(), ""));
```
dlenski, line 5765: Instead of deleting this check _for failure_… should it be replaced with an explicit check for _success_ of setting the (TLSv1.3) ciphersuite list to an empty string?

samuel40791765: Explicit check for success for `SSL[_CTX]_set_cipher_list` is on L5758-5763 in the new change :) 
But I just realized that I forgot about `SSL[_CTX]_set_ciphersuites`...

ssl/ssl_test.cc
```diff
@@ -5758,18 +5758,19 @@ TEST(SSLTest, EmptyCipherList) {
...
   // succeeds.
   EXPECT_TRUE(SSL_CTX_set_cipher_list(ctx.get(), ""));
-
   // The cipher list is updated to empty.
   EXPECT_EQ(0u, sk_SSL_CIPHER_num(SSL_CTX_get_ciphers(ctx.get())));
 
+  // Configuring the empty cipher list with |SSL_CTX_set_ciphersuites|
+  // also succeeds.
+  EXPECT_TRUE(SSL_CTX_set_ciphersuites(ctx.get(), ""));
+  EXPECT_EQ(0u, sk_SSL_CIPHER_num(SSL_CTX_get_ciphers(ctx.get())));
```
dlenski, line 5767: Thanks, LGTM with this change, and should detect any regressions in how MySQL+MariaDB use this :raised_hands:

dlenski (approved):

ssl/ssl_test.cc
dlenski: Thanks, this is good.

This _doesn't actually verify_ the pattern used by MySQL+MariaDB :sweat_smile: , which is:

1. Do *both* `SSL_CTX_set_cipher_list("non empty list for TLSv1.2")` *and* `SSL_CTX_set_ciphersuites("empty list for TLSv1.3")`
2. Expect that there will be _only_ TLSv1.2 (or earlier) ciphers available for new connections, and no TLSv1.3 ciphersuites

As described in https://github.com/openssl/openssl/issues/13704#issuecomment-748921021, it seems like it's quite complicated to get a list of _only_ the TLSv1.3 ciphersuites or _only_ the TLSv1.2 ciphers :frowning_face:

samuel40791765: We have an integration build with mysql that runs against the tests that consume this behavior. We're in the process of updating to 8.3
* https://github.com/aws/aws-lc/pull/1496

ssl/ssl_cipher.cc
```diff
@@ -1336,9 +1336,11 @@ bool ssl_create_cipher_list(UniquePtr<SSLCipherPreferenceList> *out_cipher_list,
 
   *out_cipher_list = std::move(pref_list);
 
-  // Configuring an empty cipher list is an error but still updates the
-  // output.
-  if (sk_SSL_CIPHER_num((*out_cipher_list)->ciphers.get()) == 0) {
+  // Configuring an empty cipher list is an error when |strict| is true, but
```
torben-hansen, line 1339: Please update function descriptionS as well.

torben-hansen (approved):

smittals2 (approved):

PR #1521, opened 2024-04-11
Ignore march for FIPS static and add gcc-11 FIPS CI

samuel40791765:
### Issues:
Addresses `P125414272`

### Description of changes: 
Some downstream consumers were defining `march`, which doesn't mesh well with the static FIPS build. This disables the option for now to improve usability. This may cause them to miss out on some potential CPU optimizations from the compiler, but this is better than having them break in unanticipated situations.

This change also adds a new CI dimension for gcc-10 with non-fips/fips.

### Call-outs:
Support for actually supporting `march` environments is tracked in `CryptoAlg-2396`

### Testing:
New CI for gcc-10

### DIFF
--- a/crypto/fipsmodule/CMakeLists.txt
+++ b/crypto/fipsmodule/CMakeLists.txt
@@ -308,6 +308,12 @@ if(FIPS_DELOCATE)
     message(FATAL_ERROR "Can't set both delocate and shared mode for FIPS build")
   endif()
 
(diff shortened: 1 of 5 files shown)


samuel40791765:
Closing this PR as we await more data on:
1. How much optimizations are we potentially causing consuming applications to miss out on?
2. Is this flag clearing global or just specific to `bcm.o`? We were under the impression that this is specific, but we should make sure we're not causing consuming application code to be unoptimized.

PR #1526, opened 2024-04-11
Define OPENSSL_NO_TLS_PHA, typedef PSK callback signatures

WillChilds-Klein:
### Description of changes: 

This commit defines a new configuration macro `OPENSSL_NO_TLS_PHA`. This macro is meant to be used by consuming applications to detect the fact that we (or other `libssl`s) don't support TLSv1.3's post-handshake authentication (PHA). We then use this macro in place of `OPENSSL_IS_AWSLC` to detect PHA support in our CPython patch.

We also enable PSK in our CPython patch and create two PSK-related callback function signatures [defined by OpenSSL][1] and used by CPython.

Finally, we fix the now-executed PSK tests in CPython.

[1]: https://www.openssl.org/docs/man1.1.1/man3/SSL_psk_client_cb_func.html

### Call-outs:
n/a

### Testing:
- CI
---


include/openssl/opensslconf.h
```diff
@@ -48,6 +48,7 @@
 #define OPENSSL_NO_MD2
 #define OPENSSL_NO_MDC2
 #define OPENSSL_NO_OCB
+#define OPENSSL_NO_PHA
```
samuel40791765, line 51: Should we name this to something clearer? Maybe something like 
```suggestion
#define OPENSSL_NO_TLS1_3_PHA
```

include/openssl/ssl.h
```diff
@@ -3593,6 +3593,14 @@ OPENSSL_EXPORT void SSL_set_psk_client_callback(
                              unsigned max_identity_len, uint8_t *psk,
                              unsigned max_psk_len));
 
+// SSL_psk_client_cb_func defines a function signature for the client callback.
```
samuel40791765, line 3611:
```suggestion
// SSL_psk_client_cb_func defines a function signature for |SSL(_CTX)_psk_client_cb_func|
```

Should we change `SSL(_CTX)_psk_client_cb_func` to just use this function signature? Same comments with `SSL_psk_server_cb_func` as well.

include/openssl/opensslconf.h
WillChilds-Klein: this makes sense today, but PHA will likely be present in TLS versions >1.3 as well

samuel40791765: Sure that makes sense. How about something like `OPENSSL_NO_TLS_PHA`?
The list of `OPENSSL_NO_*` macros tend to be individual algorithms and protocols. Since `PHA`'s a concept defined within the TLS protocol, I feel like there's a need to be more specific.

include/openssl/ssl.h
WillChilds-Klein: yep, that makes sense. updated function signatures for relevant functions.

include/openssl/opensslconf.h
WillChilds-Klein: agreed. updated.

samuel40791765 (dismissed):

samuel40791765 (approved):

util/doc.config
```diff
@@ -63,6 +63,7 @@
   },{
     "Name": "SSL implementation",
     "Headers": [
+      "include/openssl/opensslconf.h",
```
samuel40791765, line 66: Does this belong under "SSL implementation"? I'm not sure which of the other categories would make more sense though...

WillChilds-Klein: that's a good question, `opensslconf.h`seems to span both `libcrypto` and `libssl`. what do you think @andrewhop?

WillChilds-Klein: hm, so it looks like available sections are

- Low-level infrastructure
- Low-level crypto primitives
- Crypto interfaces
- Legacy ASN.1 and X.509 implementation
- SSL implementation

`opensslconf.h` contains configuration macros pertaining to SSL/TLS, low-level primitives, obscure protocols (e.g. trusted timestamping), etc. it's a bit of a grab bag, but I think out of the available sections "SSL implementation" is most appropriate.

I suppose we could also break it out into a new "configuration" section.

andrewhop: This seems fine for now, thanks for getting this in.

andrewhop (approved):

lilpoozie2005 (approved):

PR #1530, opened 2024-04-15
MinGW: mitigate potential abort on rwlocks using PTHREAD_RWLOCK_INITIALIZER

justinwsmith:
### Description of changes: 
* On some platforms/pthreads-implementations, there are [race conditions](https://github.com/ldx/winpthreads/blob/fecf6566606c9111628441ed2426e7efdc5faf0e/src/rwlock.c#L161) on the first use of rwlocks initialized with `PTHREAD_RWLOCK_INITIALIZER`.
* This change allows for a re-attempt on lock acquisition when it initially fails on MinGW, The current logic aborts the process when this occurs.
* A test is added to attempt to reproduce the race-condition, although success of it can be a false-negative for such a bug.


crypto/thread_pthread.c
```diff
@@ -65,13 +65,47 @@ void CRYPTO_MUTEX_cleanup(CRYPTO_MUTEX *lock) {
 }
 
 void CRYPTO_STATIC_MUTEX_lock_read(struct CRYPTO_STATIC_MUTEX *lock) {
-  if (pthread_rwlock_rdlock(&lock->lock) != 0) {
+  int result = pthread_rwlock_rdlock(&lock->lock);
+  if (result != 0) {
+#ifdef __MINGW32__
```
torben-hansen, line 70: Factor this out into its own function? e.g. `mingw_fallback_retry()`.

skmcgrail (approved):

skmcgrail (approved):

crypto/thread_pthread.c
```diff
@@ -64,14 +66,46 @@ void CRYPTO_MUTEX_cleanup(CRYPTO_MUTEX *lock) {
...
+  } while(result == EINVAL && attempt_num < MAX_ATTEMPTS);
+  return result;
+}
+#endif
+
 void CRYPTO_STATIC_MUTEX_lock_read(struct CRYPTO_STATIC_MUTEX *lock) {
-  if (pthread_rwlock_rdlock(&lock->lock) != 0) {
+  const int result = pthread_rwlock_rdlock(&lock->lock);
+  if (result != 0) {
+#ifdef __MINGW32__
```
torben-hansen, line 90: Take it or leave it suggestion: Can avoid the last two set of if/defs if calling `rwlock_EINVAL_fallback_retry()` on every invocation. Making it an inline no-op if not inside mingw, will optimise it out, I suppose.

torben-hansen (approved):

skmcgrail (approved):

torben-hansen (approved):

PR #1538, opened 2024-04-18
Windows build with clang-cl; CI for Windows/ARM64 build

justsmth:
## Issues
Needed for: https://github.com/aws/aws-lc-rs/issues/376

## Description
* Updated `CMakeLists.txt` to support a Windows build using `clang-cl`.
* Adds CI build/test of Windows/x64 using clang-cl.
* Adds CI build of Windows/ARM64 using clang-cl. (Currently there are no runners available for testing Windows/ARM64 executables.)

## Callout
* This change removes debug information by default from the "Release" build type. Alternatively, one may retain debug information in the artifacts by adding the  `-ggdb` build flag via **CMAKE_C_FLAGS** or by using the "Relwithdebinfo" or "Debug" build types.


smittals2 (approved):

CMakeLists.txt
```diff
@@ -333,7 +333,37 @@ if(GCC OR CLANG)
...
+      set(C_CXX_FLAGS "${C_CXX_FLAGS} -Wno-nonportable-system-include-path")
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-reserved-identifier")
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-shorten-64-to-32")
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-sign-conversion")
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unreachable-code-break")
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-tautological-value-range-compare")
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unreachable-code-return")
+      set(C_CXX_FLAGS "${C_CXX_FLAGS} -Wno-unsafe-buffer-usage")
+      set(C_CXX_FLAGS "${C_CXX_FLAGS} -Wno-unused-macros")
+      set(C_CXX_FLAGS "${C_CXX_FLAGS} -Wno-used-but-marked-unused")
```
samuel40791765, line 362: Is there a reason why we're appending to the flags like this? These seem to be all over the place and it'd be cleaner if we could define these like our other flags.
Ex:
```
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-bad-function-cast -Wno-cast-function-type-strict -Wno-declaration-after-statement -Wno-disabled-macro-expansion ...")
```

CMakeLists.txt
```diff
@@ -333,7 +333,37 @@ if(GCC OR CLANG)
...
+    else()
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fvisibility=hidden")
```
samuel40791765, line 364: Is the intent to remove this for Windows/MSVC? Won't this expose all hidden symbols when built with MSVC?

CMakeLists.txt
justsmth: This format makes it easier to disable specific flags (i.e., by commenting out a line) when working to correct these warnings. For example, I've got a local branch where I've started addressing `-Wno-nonportable-system-include-path`. There are a few other Warnings that appear to be quite fixable with just a little effort.  I plan to periodically post PRs that address such warnings.

CMakeLists.txt
justsmth: This flag is not supported by `clang-cl` so it must be excluded.

CMakeLists.txt
samuel40791765: Sure, if this does makes things easier I guess there's no harm keeping it. Should we leave a TODO comment above this that we'll get to these turned off warnings and group them together? And for now, could we at least group the `CMAKE_C_FLAGS` and `C_CXX_FLAGS` in the same sections?

CMakeLists.txt
samuel40791765: Is there a way to differentiate `clang-cl` with MSVC? I'm not sure if we want to disable this for all of Windows entirely, do we understand the repercussions of turning this flag off for MSVC?

justsmth: This block is unique to `clang-cl` since it satisfies both the `CLANG` and `MSVC` -- this block doesn't affect "normal (`cl`) MSVC builds.

samuel40791765: Ahhhh thanks for clearing things up there. I didn't know `clang-cl` was Clang/MSVC hybrid

smittals2 (dismissed):

samuel40791765 (dismissed):

samuel40791765 (approved):

skmcgrail (approved):

PR #1539, opened 2024-04-18
Use CodeBuild batch matrix to test all major releases of GCC between 8 and 13

andrewhop:
### Description of changes: 
This uses pre-built docker images that contain various versions of GCC that AWS ECR mirrors from docker hub. The [CodeBuild batch build matrix](https://docs.aws.amazon.com/codebuild/latest/userguide/batch-build.html#batch_build_matrix) should result in 2 * 6 jobs being run. Later once we have confidence in these images we can remove some of the manually configured  jobs below. 

### Call-outs:
For some reason [ECR only has 8-13](https://gallery.ecr.aws/docker/library/gcc) while [Docker Hub goes back to 4](https://hub.docker.com/_/gcc/tags?page=&page_size=&ordering=-last_updated&name=). I decided not to use Docker Hub because of rate limiting concerns.  

### Testing:
This is a CI change, I expect to find some minor issues with some of the GCC versions. 


andrewhop:
> BATCH_SPEC_VALIDATION_ERROR: BatchSpec has invalid values.. 1 batch type is expected at a time for buildBatchArn

Well that's a bummer. I am going to try deleting all fo the build-list just to see what happens.

PR #1542, opened 2024-04-20
Changed SSL_client_hello_get0_ciphers to align with OpenSSL behavior

smittals2:
### Description of changes: 
This PR aligns SSL_client_hello_get0_ciphers with OpenSSL's behavior. Previously, it returned only the ciphers supported by our library. Now, it offers all ciphersuites from the client hello, including unsupported ones.

1. New field added to SSL struct, |all_client_cipher_suites|
2. |ssl_parse_client_cipher_list| modified to also populate |ssl->all_client_cipher_suites|. Function signature also modified to take in an SSL object
3. Rewrote | SSL_client_hello_get0_ciphers|
4. Changed documentation to better callout the differences between which ciphersuites are returned (supported or unsupported)
5. Added tests for TLS 1.2 and 1.3


ssl/handshake_server.cc
```diff
@@ -291,13 +293,20 @@ bool ssl_parse_client_cipher_list(
...
+    // Storing only supported ciphers
     const SSL_CIPHER *c = SSL_get_cipher_by_value(cipher_suite);
     if (c != NULL && !sk_SSL_CIPHER_push(sk.get(), c)) {
       OPENSSL_PUT_ERROR(SSL, SSL_R_ERROR_IN_RECEIVED_CIPHER_LIST);
       return false;
     }
   }
 
+  // Store raw bytes for cipher suites offered
+  char * tmp = (char *) OPENSSL_malloc(all_ciphers_len + 1);
```
justsmth, line 305: * Is extending it with a null byte and the re-computation of `all_ciphers_len` necessary?
* Could this just use `OPENSSL_memdup`:
```c++
  char* tmp =  static_cast<char *>(OPENSSL_memdup(
      client_hello->cipher_suites, client_hello->cipher_suites_len));
```
* How should we handle `tmp` being NULL after malloc?

ssl/internal.h
```diff
@@ -4064,14 +4065,15 @@ struct ssl_st {
   bssl::UniquePtr<SSL_SESSION> session;
 
   // client_cipher_suites contains cipher suites offered by the client during
-  // the handshake, with preference order maintained. This field is NOT
-  // serialized and is only populated if used in a server context.
+  // the handshake and supported by this library, with preference order maintained.
```
justsmth, line 4068: NP: "... that are supported by this library."

ssl/ssl_lib.cc
```diff
@@ -3311,33 +3310,16 @@ int SSL_set1_curves_list(SSL *ssl, const char *curves) {
 }
 
 size_t SSL_client_hello_get0_ciphers(SSL *ssl, const unsigned char **out) {
-  STACK_OF(SSL_CIPHER) *client_cipher_suites = SSL_get_client_ciphers(ssl);
-  if (client_cipher_suites == nullptr) {
+  if (SSL_get_client_ciphers(ssl) == nullptr) {
       return 0;
   }
+  const char * ciphers = ssl->all_client_cipher_suites.get();
+  if (ciphers == nullptr) {
```
justsmth, line 3317: It shouldn't possible for this to be null if the logic is correct, right?  Could we put an assert here to make such a failure more obvious during testing:
```c++
assert!(ciphers != nullptr);
```

ssl/ssl_lib.cc
```diff
@@ -3311,33 +3310,16 @@ int SSL_set1_curves_list(SSL *ssl, const char *curves) {
...
-    }
-
-    assert(ssl->client_cipher_suites_arr.size() == num_ciphers);
-    *out = reinterpret_cast<unsigned char*>(ssl->client_cipher_suites_arr.data());
+    *out = reinterpret_cast<const unsigned char*>(ciphers);
   }
-
-  // Return the size
-  return num_ciphers;
+  return strlen(ciphers);
```
justsmth, line 3324: ahh, the size+1 allocation was needed b/c the value is being interpreted as a null-terminated string here. I think it'd be better to explicitly store the size of this array and use that here.

ssl/ssl_lib.cc
```diff
@@ -3085,7 +3085,6 @@ int SSL_clear(SSL *ssl) {
   }
 
   ssl->client_cipher_suites.reset();
-  ssl->client_cipher_suites_arr.Reset();
```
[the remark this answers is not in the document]
smittals2: Yep I had removed it when I changed arr to char ptr but forgot to put it back in :)

ssl/handshake_server.cc
smittals2: As long as we treat it as a string and return strlen in SSL_client_hello_get0_ciphers, it's probably best to explicitly set the null char and compute length. CBS_data(&cipher_suites) returns the relevant data as a char pointer but doesn't zero out the memory after it/has a null char at the end. Using memdup works to copy the data but there is no guarantee of there being a null char at the end. We would need to allocate another byte and then explicitly set it. 

To handle tmp being NULL, I'll add a failure since this allocation shouldn't fail in the handshake.

ssl/internal.h
```diff
@@ -4064,14 +4065,15 @@ struct ssl_st {
...
 
-  // client_cipher_suites_arr is initialized when |SSL_client_hello_get0_ciphers|
-  // is called with a valid |out| param. It holds the cipher suites offered by the
-  // client from |client_cipher_suites| in an array.
-  bssl::Array<uint16_t> client_cipher_suites_arr;
+  // all_client_cipher_suites contains the raw bytes for cipher suites offered by the
+  // client during the handshake (including those not supported by this library), with
+  // preference order maintained. This field is NOT serialized and is only populated
+  // if used in a server context. This field may contain cipher IDs that are invalid.
+  bssl::UniquePtr<char> all_client_cipher_suites;
```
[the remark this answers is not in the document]
smittals2: Is there any advantage to that? Way I see it is another 2 bytes of overhead over the lifetime of the connection but I guess it's a small enough amount to not matter?

ssl/ssl_lib.cc
smittals2: Yep, made the change!

ssl/ssl_lib.cc
justsmth: Do we know that the cipher content cannot contain a 0-byte?

smittals2: Hmmm that's an interesting case. If the cipher data contains valid cipher IDs, then we know. But I guess a client could hypothetically send invalid cipher suites that contain the null byte - investigating.

andrewhop: This is a problem from the OpenSSL API, because lots of cipher suites include a 0 byte for the first part of the ID, and the all null cipher suite is 0x0, 0x0 https://datatracker.ietf.org/doc/html/rfc5246#appendix-A.5. However reading https://www.openssl.org/docs/man1.1.1/man3/SSL_client_hello_get0_ciphers.html more closely it also returns the length of the array, so I don't think we need to null terminate this string because OpenSSL also isn't treating it as a string. But we do need to track the length.

smittals2: So we can directly return the underlying buffer and don't have to worry about any scrap data at the end that was never zeroed out. It becomes the caller's responsibility to read the right number of bytes using the return value of get0_ciphers?

ssl/handshake_server.cc
```diff
@@ -291,13 +291,19 @@ bool ssl_parse_client_cipher_list(
...
+    // Storing only supported ciphers
     const SSL_CIPHER *c = SSL_get_cipher_by_value(cipher_suite);
     if (c != NULL && !sk_SSL_CIPHER_push(sk.get(), c)) {
       OPENSSL_PUT_ERROR(SSL, SSL_R_ERROR_IN_RECEIVED_CIPHER_LIST);
       return false;
     }
   }
 
+  // Store raw bytes for cipher suites offered
+  ssl->all_client_cipher_suites.reset(static_cast<char *>(OPENSSL_memdup(
```
justsmth, line 303: Should this be stored earlier in the function (prior to parsing) in case parsing fails?  Is there a use-case where a user would want to inspect/log the client hello cipher suites message to diagnose a failure?

ssl/ssl_lib.cc
andrewhop: Yes, that's how I'm reading the OpenSSL docs.

include/openssl/ssl.h
```diff
@@ -1849,16 +1849,18 @@ OPENSSL_EXPORT int SSL_get_extms_support(const SSL *ssl);
 // not been negotiated yet.
 OPENSSL_EXPORT const SSL_CIPHER *SSL_get_current_cipher(const SSL *ssl);
 
-// SSL_get_client_ciphers returns stack of ciphers offered by the client during
-// a handshake. If the |ssl| is a client or the handshake hasn't occurred yet,
-// NULL is returned. The stack of ciphers IS NOT de/serialized, so NULL will
-// also be returned for deserialized or transported |ssl|'s that haven't yet
+// SSL_get_client_ciphers returns the stack of ciphers offered by the client during
+// a handshake and that are supported by this library. If the |ssl| is a client or the
```
andrewhop, line 1853: Is it the ones supported by AWS-LC, or the potentially matching ciphers with what the server accepts?

justsmth (approved):

ssl/internal.h
```diff
@@ -4064,14 +4065,19 @@ struct ssl_st {
...
-  // is called with a valid |out| param. It holds the cipher suites offered by the
-  // client from |client_cipher_suites| in an array.
-  bssl::Array<uint16_t> client_cipher_suites_arr;
+  // all_client_cipher_suites contains the raw bytes for cipher suites offered
+  // by the client during the handshake (including those not supported by this
+  // library), with preference order maintained. This field may contain
+  // cipher IDs that are invalid.
+  bssl::UniquePtr<char> all_client_cipher_suites;
+
+  // Field length of ciphers in client hello
```
andrewhop, line 4079: NP: it's worth calling out 2^16 is the maximum size allowed by the RFC https://datatracker.ietf.org/doc/html/rfc8446#section-4.1.2 and this is separate from the SSLBuffer 2^16 limit.

ssl/ssl_lib.cc
```diff
@@ -3311,33 +3311,14 @@ int SSL_set1_curves_list(SSL *ssl, const char *curves) {
...
-    }
-
-    assert(ssl->client_cipher_suites_arr.size() == num_ciphers);
-    *out = reinterpret_cast<unsigned char*>(ssl->client_cipher_suites_arr.data());
+    *out = reinterpret_cast<const unsigned char*>(ciphers);
   }
-
-  // Return the size
-  return num_ciphers;
+  return ssl->all_client_cipher_suites_len;
```
andrewhop, line 3323: Nice this is much simpler.

ssl/ssl_test.cc
```diff
@@ -2411,13 +2411,75 @@ TEST(SSLTest, FindingCipher) {
...
+  // This should fail, but should be able to inspect client ciphers still
+  ASSERT_FALSE(CompleteHandshakes(client.get(), server.get()));
+
+  // Client calling, should return 0
+  ASSERT_EQ(SSL_client_hello_get0_ciphers(client.get(), nullptr), (size_t) 0);
+
+  const unsigned char expected_cipher_bytes[] = {0x13, 0x02, 0x13, 0x03};
+  const unsigned char *p = nullptr;
+
+  // Expected size is 2 bytes more than |expected_cipher_bytes| to account for grease value
```
andrewhop, line 2452: This is a useful comment because it looks weird without the context of grease is the extra two.

ssl/ssl_lib.cc
smittals2: Got it, that will be the new behavior after this PR!

ssl/ssl_test.cc
```diff
@@ -2411,13 +2411,75 @@ TEST(SSLTest, FindingCipher) {
...
+  const unsigned char *tmp = nullptr;
+  // Handshake not completed, getting ciphers should fail
+  ASSERT_FALSE(SSL_client_hello_get0_ciphers(client.get(), &tmp));
+  ASSERT_FALSE(SSL_client_hello_get0_ciphers(server.get(), &tmp));
+  ASSERT_FALSE(tmp);
+
+  // This should fail, but should be able to inspect client ciphers still
+  ASSERT_FALSE(CompleteHandshakes(client.get(), server.get()));
+
+  // Client calling, should return 0
```
andrewhop, line 2446: NP: comments like this don't add any value, this is what the test is doing on the next line.
```suggestion
```

include/openssl/ssl.h
smittals2: No cipher choosing is occurring yet. SSL_get_client_ciphers returns client_cipher_suites which is the parsed variant of all_client_cipher_suites (added in this PR). This parsed variant contains all cipher suites offered from the client that are supported by our library (checking against server supported suites doesn't occur yet). client_cipher_suites is then used to pick a cipher that is supported both by the server and client.

justsmth (approved):

andrewhop (approved):

PR #1543, opened 2024-04-20
Support vpinsrq in delocater

torben-hansen:
### Issues:

P125414272

### Description of changes: 

Instead of generating portable code, gcc can be configured with e.g. `march=cpu-type` that allows it to generate code using instructions from instruction sets supported on up-to `cpu-type`.

In one of these cases, we for example saw:
```
[ 62%] Generating bcm-delocated.S
error while processing "\tvpinsrq\t$1, EVP_PKEY_CTX_dup@GOTPCREL(%rip), %xmm1, %xmm0\n" on line 27676: "GOT access must be source operand, instrOther"
```
In this case, the compiler tries to be smart and get the necessary memory addresses for `EVP_PKEY_CTX_free` and `EVP_PKEY_CTX_dup` to populate `EVP_MD_pctx_ops ` e.g.:
```
    vmovq   EVP_PKEY_CTX_free@GOTPCREL(%rip), %xmm1
    vpinsrq $1, EVP_PKEY_CTX_dup@GOTPCREL(%rip), %xmm1, %xmm0
    vmovdqa %xmm0, -16(%rbp)
```
This fails because `vpinsrq` is not supported in the delocater.

This PR adds support for `vpinsrq` an instruction from the set `AVX512DQ`. I don't think there are other 4-argument instructions, for where a GOT reloc can be emitted, that we need to support right now. This also makes the implementation a tad easier, because we do not need to cater for the relocation being either the first or second argument - it can only be the second.

Otherwise, the implementation follows the one for the type `instrThreeArg`.

### Testing:

Tested this in an environment with conditions that triggers the missing support for `vpinsrq`. Before it failed, after this PR is applied, build and tests succeeded.

### DIFF
--- a/util/fipstools/delocate/delocate.go
+++ b/util/fipstools/delocate/delocate.go
@@ -1189,6 +1189,8 @@ const (
 	instrMemoryVectorCombine
 	// instrThreeArg merges two sources into a destination in some fashion.
 	instrThreeArg
+	// instrFourArg merges three sources into a destination in some fashion.
+	instrFourArg
 	// instrCompare takes two arguments and writes outputs to the flags register.
 	instrCompare
 	instrOther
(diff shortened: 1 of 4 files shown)


dkostic:
nit: `vpinsrq` was introduced in AVX ISA.

PR #1546, opened 2024-04-23
Basic GH CI build/test with full range of gcc/clang

justsmth:
### Description of changes: 
* Adds basic CI build/test with all GH available GCC/Clang compilers on Ubuntu.

### Call-outs:
* This a supplement to the more exhaustive testing done in CodeBuild for many (but not all) of the same compilers.

### DIFF
--- a/.github/workflows/actions-ci.yml
+++ b/.github/workflows/actions-ci.yml
@@ -136,6 +136,104 @@ jobs:
           echo ${env:SDEROOT}
           .\tests\ci\run_windows_tests.bat "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x64 true
 
+  gcc-9-13-sanity:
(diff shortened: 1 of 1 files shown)


samuel40791765:
My understanding was that we used Codebuild for basic tests on linux with different versions of gcc/clang. Did we want to introduce another gcc test dimension via Github Actions?

samuel40791765 (approved):
We synced up on how we plan to use GHA for compiler tests on linux. The plan is to use GHA for basic sanity tests on different compilers and continue using Codebuild for more exhaustive tests on different compilers.
Ideally we would use succeeding the sanity test in GHA to kick off the exhaustive tests in Codebuild. But this would need more rewriting of the existing cdk script and it's unknown if Codebuild supports being started off a successful Github Actions run.

skmcgrail (approved):

PR #1551, opened 2024-04-25
Remove comments about overread for entropy generation

fabrice102:
### Description of changes: 

AWS-LC does not overread entropy sources in FIPS mode. This commit is to clarify this and remove/rephrase comments mentioning overread.

### Call-outs:

Only changes documentations & comments. No code change.

### Testing:

Only changes documentations & comments. No code change.

### DIFF
--- a/crypto/fipsmodule/FIPS.md
+++ b/crypto/fipsmodule/FIPS.md
@@ -50,7 +50,7 @@ FIPS requires that RNG state be zeroed when the process exits. In order to imple
 
 By default, entropy is sourced using a "Passive" method where the specific entropy source depends on the OE. Seeding and reseeding material for the DRBG is sourced from the specific entropy source.
 
-In FIPS mode, when the CPU Jitter entropy source is used, we do a 3x oversampling (the CPU Jitter default setting). Note that the CPU Jitter entropy source is not the default entropy source (see subsection below).
+In FIPS mode, when the CPU Jitter entropy source is used, we do a 3x oversampling (the CPU Jitter default setting). Note that the CPU Jitter is not the default entropy source (see subsection below).
 
 ### Modify entropy source - not recommended
 
--- a/crypto/fipsmodule/FIPS.md
+++ b/crypto/fipsmodule/FIPS.md
@@ -40,7 +40,7 @@ The DRBG state is kept in a thread-local structure and is seeded using the confi
 
 The CTR-DRBG is reseeded every 4096 calls to `RAND_bytes`. Thus the process will randomly crash about every 2¹³⁵ calls.
 
-The FIPS PRNGs allow “additional input” to be fed into a given call. We use this feature to be as robust as possible to state duplication from process forks and VM copies: for every call we read 32 bytes of “additional data” from the entropy source (without overread) which means that cloned states will diverge at the next call to `RAND_bytes`. This is called “prediction resistance” by FIPS, but we do *not* claim this property in a FIPS context because we don't implement it the way they want.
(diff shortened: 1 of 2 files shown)


crypto/fipsmodule/FIPS.md
```diff
@@ -40,7 +40,7 @@ The DRBG state is kept in a thread-local structure and is seeded using the confi
 
 The CTR-DRBG is reseeded every 4096 calls to `RAND_bytes`. Thus the process will randomly crash about every 2¹³⁵ calls.
 
-The FIPS PRNGs allow “additional input” to be fed into a given call. We use this feature to be as robust as possible to state duplication from process forks and VM copies: for every call we read 32 bytes of “additional data” from the entropy source (without overread) which means that cloned states will diverge at the next call to `RAND_bytes`. This is called “prediction resistance” by FIPS, but we do *not* claim this property in a FIPS context because we don't implement it the way they want.
+The FIPS PRNGs allow “additional input” to be fed into a given call. We use this feature to be as robust as possible to state duplication from process forks and VM copies: for every call we read 32 bytes of “additional data” from the entropy source which means that cloned states will diverge at the next call to `RAND_bytes`. This is called “prediction resistance” by FIPS, but we do *not* claim this property in a FIPS context because we don't implement it the way they want.
```
torben-hansen, line 43: Only true for Intel CPUs.

fabrice102: What do you think of the following:

```suggestion
The FIPS PRNGs allow “additional input” to be fed into a given call. We use this feature to be as robust as possible to state duplication from process forks and VM copies: on Intel CPUs with RDRAND, for every call we read 32 bytes of “additional data” from RDRAND, which means that cloned states will diverge at the next call to `RAND_bytes`. This is called “prediction resistance” by FIPS, but we do *not* claim this property in a FIPS context because we don't implement it the way they want.
```

torben-hansen (dismissed):

crypto/fipsmodule/FIPS.md
```diff
@@ -50,8 +50,6 @@ FIPS requires that RNG state be zeroed when the process exits. In order to imple
 
 By default, entropy is sourced using a "Passive" method where the specific entropy source depends on the OE. Seeding and reseeding material for the DRBG is sourced from the specific entropy source.
 
-In FIPS mode, each of the entropy sources is subject to a 10× overread. That is, when *n* bytes of entropy are needed, *10n* bytes will be read from the entropy source and XORed down to *n* bytes.
```
nebeid, line 53: Based on https://github.com/aws/aws-lc/blob/7962f75d9e836cd5ad84beed3385add9157c8a30/crypto/fipsmodule/rand/rand.c#L141 I think we can say that, when Jitter is used, we do a 3x oversampling (the Jitter default setting).

fabrice102: I've added a comment about it and clarified that the CPU Jitter is not the default entropy source.

crypto/fipsmodule/FIPS.md
```diff
@@ -50,7 +50,7 @@ FIPS requires that RNG state be zeroed when the process exits. In order to imple
...
+In FIPS mode, when the CPU Jitter entropy source is used, we do a 3x oversampling (the CPU Jitter default setting). Note that the CPU Jitter entropy source is not the default entropy source (see subsection below).
```
nebeid, line 53:
```suggestion
In FIPS mode, when the CPU Jitter entropy source is used, we do a 3x oversampling (the CPU Jitter default setting). Note that the CPU Jitter is not the default entropy source (see subsection below).
```
Nit: a suggestion, in order to reduce repetition.

torben-hansen (approved):

nebeid (approved):

PR #1554, opened 2024-04-26
Add dependency to python3-six in github action grpc

fabrice102:
### Description of changes: 

Fix Github actions to work with Ubuntu 24.04

Ubuntu 24.04 does not provide the Python package six by default, contrary to previous Ubuntu versions. This causes issues with the integrations:grpc action.

### Call-outs:

N/A

### Testing:

N/A (just a change in dependencies in integrations tests)

### DIFF
--- a/.github/workflows/actions-ci.yml
+++ b/.github/workflows/actions-ci.yml
@@ -35,6 +35,9 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          brew install ninja golang
       - name: Build ${{ env.PACKAGE_NAME }}
         run: |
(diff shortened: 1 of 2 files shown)


dkostic (dismissed):

torben-hansen (approved):

samuel40791765 (approved):
Thanks! Seems like Github is making the transition to use ARM for `macos-latest`.. https://github.com/actions/runner-images/issues/9741#issuecomment-2075569220

Note on `macos-latest` and `macos-latest-xlarge`: the CI build might be slightly slower if we use `macos-latest`, but it's all good if the total run is around an hour. Seems like Github is using `large` and `xlarge` to differentiate x86 and arm, so the extra distinction is nice. But no strong preferences on my side.

andrewhop (approved):

PR #1563, opened 2024-04-29
Add integration script and CI for ruby 3.1 and 3.2

samuel40791765:
### Description of changes: 
We've finalized our support for Ruby 3.1 and 3.2. This adds the patch and integration CI to test against it.

Ruby 3.1 patch contents are as following:
1. Slight logic to properly detect AWS-LC in FIPS mode.
2. Differences in emitted error messages
3. During signature verification, an invalid ASN.1 input raises an error in Ruby/OpenSSL and an actual verification failure returns false. AWS-LC simply returns false for all cases.
4. OpenSSL allows invalid DH parameters to be parsed successfully and invalidates it in a subsequent call to `params_ok?`. AWS-LC simply disallows invalid parameters to be parsed.
5. AWS-LC does not support the serialization of custom/explicit curves. Explicit curves are highly impractical to validate from a security standpoint and have been the source of many CVEs. See the following issues for more details on why explicit curves are discouraged.
    * https://github.com/openssl/openssl/issues/9286
    * https://github.com/openssl/openssl/issues/20119
6. AWS-LC does not support `BN::CONSTTIME`. See https://github.com/aws/aws-lc/commit/0a211dfe91588d2986a8735e1969dd9202a8b025 for more details. 
7. AWS-LC does not support DHE ciphersuites in SSL connections. 5 tests have been adjusted accordingly.
8. Changes in pkcs12 are pieces of https://github.com/ruby/ruby/commit/63e9eaa5bef3fdb1f621d81eab6578fd4cb0f78d in upstream Ruby. This test was testing with the wrong number of parameters. AWS-LC does not support the "old MSIE extension" mentioned in the commit.
9. Changes in `test/openssl/test_pkey_rsa.rb` are pieces of https://github.com/ruby/ruby/commit/2e5680d304a9cf9a6a2ba582091af6719e839351 in upstream Ruby. It accounts for RSA operations disallowed in FIPS mode in Ruby's RSA tests. The commit diff is a bit too large to muddle with the changes in this CR. It's also fairly recent and not directly applicable to Ruby 3.1's version of the file, so I've only taken pieces that ensure we're not losing coverage.

Ruby 3.2 patch contents are nearly identical to the points mentioned above. Only additional thing to note is AWS-LC does key checks while parsing EC Keys and disallows invalid keys to be parsed. This is similar to the DH params discrepancy described above in point 4.

### Testing:
CI


tests/ci/integration/run_ruby_integration.sh
```diff
@@ -0,0 +1,76 @@
...
+RUBY_PATCH_FOLDER="${SRC_ROOT}/tests/ci/integration/ruby_patch"
+AWS_LC_BUILD_FOLDER="${SCRATCH_FOLDER}/aws-lc-build"
+AWS_LC_INSTALL_FOLDER="${SCRATCH_FOLDER}/aws-lc-install"
+
+function ruby_build() {
+    local branch=${1}
+    pushd ${branch}
+    ./autogen.sh
+    mkdir -p build && cd build
+    ../configure --with-openssl-dir=${AWS_LC_INSTALL_FOLDER} \
```
WillChilds-Klein, line 33: minor nit: you can use the `-C` flag for make to specify a directory instead of `cd`ing into `build`

tests/ci/integration/run_ruby_integration.sh
```diff
@@ -0,0 +1,76 @@
...
+                 --with-openssl-lib=${AWS_LC_INSTALL_FOLDER}/lib \
+                 --with-openssl-include=${AWS_LC_INSTALL_FOLDER}/include
+    make -j ${NUM_CPU_THREADS}
```
WillChilds-Klein, line 36: after the build (and tests), we should be able to ensure that we're linked against AWS-LC with

```
$ ruby -e 'require "openssl"; puts OpenSSL::OPENSSL_VERSION'
OpenSSL 1.0.2k  26 Jan 2017
```

WillChilds-Klein (commented):
(sorry for lurking on a draft)

tests/ci/integration/run_ruby_integration.sh
```diff
@@ -0,0 +1,85 @@
...
+    ./configure --disable-install-doc \
+                 --prefix="$PWD/install" \
+                 --with-openssl-dir=${AWS_LC_INSTALL_FOLDER}
+    make V=1 -j ${NUM_CPU_THREADS}
+    make install
+    # Check that AWS-LC was used.
+    ./install/bin/ruby -e 'require "openssl"; puts OpenSSL::OPENSSL_VERSION' | grep -q "AWS-LC" && echo "AWS-LC found!" || exit 1
+
+    #TODO: add more relevant tests here
+    make test-all TESTS="test/openssl/*.rb"
```
WillChilds-Klein, line 42: I may be misunderstanding this, but from this PR's most recent CI run, it [looks like](https://github.com/aws/aws-lc/actions/runs/12248575984/job/34168439237?pr=1563#step:4:4549) a different executable than `./install/bin/ruby` is used to execute the tests, namely `miniruby`:

```

test-all
  Run options: 
    --seed=24196
    "--ruby=./miniruby -I./lib -I. -I.ext/common  ./tool/runruby.rb --extout=.ext  -- --disable-gems"
    --excludes-dir=./test/excludes
    --name=!/memory_leak/
```

could we patch a test into the OpenSSL module unit tests that asserts `OpenSSL::OPENSSL_VERSION.include? 'AWS-LC'`?

tests/ci/integration/run_ruby_integration.sh
```diff
@@ -0,0 +1,85 @@
...
+# Assumes script is executed from the root of aws-lc directory
+SCRATCH_FOLDER="${SRC_ROOT}/RUBY_BUILD_ROOT"
+RUBY_SRC_FOLDER="${SCRATCH_FOLDER}/ruby-src"
+RUBY_PATCH_FOLDER="${SRC_ROOT}/tests/ci/integration/ruby_patch"
+AWS_LC_BUILD_FOLDER="${SCRATCH_FOLDER}/aws-lc-build"
+AWS_LC_INSTALL_FOLDER="${SCRATCH_FOLDER}/aws-lc-install"
+
+function ruby_build() {
+    local branch=${1}
+    pushd ${branch}
```
justsmth, line 30: It would be more robust to not assume the current directory, but fully qualify the path:
```
pushd "${RUBY_SRC_FOLDER}/${branch}"
```

tests/ci/integration/run_ruby_integration.sh
```diff
@@ -0,0 +1,85 @@
...
+
+mkdir -p ${SCRATCH_FOLDER}
+rm -rf ${SCRATCH_FOLDER}/*
+cd ${SCRATCH_FOLDER}
+
+mkdir -p ${AWS_LC_BUILD_FOLDER} ${AWS_LC_INSTALL_FOLDER}
+aws_lc_build ${SRC_ROOT} ${AWS_LC_BUILD_FOLDER} ${AWS_LC_INSTALL_FOLDER} -DBUILD_TESTING=OFF -DFIPS=1
+
+mkdir -p ${RUBY_SRC_FOLDER}
+pushd ${RUBY_SRC_FOLDER}
```
justsmth, line 77: This directory change isn't needed if the functions below use fully qualified paths.

tests/ci/integration/run_ruby_integration.sh
```diff
@@ -0,0 +1,85 @@
...
+    pushd ${branch}
+    ./autogen.sh
+    mkdir -p install
+    ./configure --disable-install-doc \
+                 --prefix="$PWD/install" \
+                 --with-openssl-dir=${AWS_LC_INSTALL_FOLDER}
+    make V=1 -j ${NUM_CPU_THREADS}
+    make install
+    # Check that AWS-LC was used.
+    ./install/bin/ruby -e 'require "openssl"; puts OpenSSL::OPENSSL_VERSION' | grep -q "AWS-LC" && echo "AWS-LC found!" || exit 1
```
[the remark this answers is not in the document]
samuel40791765: Plans are to do so internally. But now that you mention it, I should do an ldd check here to assert we're using the right libcrypto installation.

tests/ci/integration/run_ruby_integration.sh
samuel40791765: Not a test, but we do have a version detection mechanism in the patch:
```
+ def aws_lc?(major = nil, minor = nil, fix = nil)
+    version = OpenSSL::OPENSSL_VERSION.scan(/AWS-LC (\d+)\.(\d+)\.(\d+).*/)[0]
+    return false unless version
+    !major || (version.map(&:to_i) <=> [major, minor, fix]) >= 0
+  end
```
It's used throughout the tests to skip across various AWS-LC specific discrepancies. I'm hesitant to add a specific test that asserts AWS-LC. Ultimate end goal is to run this patch down to 0 if upstream is willing to take us. I also think the various bash checks we do and the mentioned version detection should be sufficient.

samuel40791765: `miniruby` appears to be minimal lighter version of ruby used in their build. The ruby binary we're asserting in this script is the actual ruby binary that the general consumer uses.
* https://github.com/ruby/ruby/wiki/Developer-How-To#make-targets

.github/workflows/integrations.yml
```diff
@@ -257,3 +257,15 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install OS Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get -y --no-install-recommends install cmake gcc ninja-build golang make autoconf ruby libyaml-dev
+      - uses: actions/checkout@v3
+      - name: Build AWS-LC, build ruby, run tests
+        run: |
+          ./tests/ci/integration/run_ruby_integration.sh ruby_3_1 ruby_3_2
```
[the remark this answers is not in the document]
WillChilds-Klein: today, our python integration tests [split version dimension concerns](https://github.com/aws/aws-lc/blob/main/.github/workflows/integrations.yml#L113-L150) into two buckets:

- main (which we expect to break with some nonzero cadence)
- supported versions (which we expect to be more stable, and over which we parameterize other concerns like FIPS and interaction with 3p modules)

the thought here is that we let upstream `main` run free and wild without increasing CI dimensions, but see breakages and get ahead of them by either modifying our main CI patch file (thus documenting current incompatibilities, perhaps upstreaming a patch) or making attendant compatibility-oriented changes to AWS-LC.

for python, we expect the non-main patch set ("*-releases") to be more stable, so we codify individual release versions and test against those across multiple dimensions (for now, just FIPS and aws-crt's backing lib for python).

a similar strategy might be useful for the ruby integration as well.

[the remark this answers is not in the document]
samuel40791765: Sure that makes perfect sense. I'll move these back under the same dimension and break main out into it's own once we have support for that.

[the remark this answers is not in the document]
justsmth: With only two branches, testing them together in the same job is not much of an issue. However as we scale to support for more Ruby versions, having visibility on the specific branch for which tests are failing (when some failing versions might be known, but other failures new/unknown) will make it easier to determine whether action needs to be taken during a PR review.

```
  ruby-releases:
    if: github.repository_owner == 'aws'
    strategy:
      fail-fast: false
      matrix:
        fips:
          - "0"
          - "1"
        branch:
          - "ruby_3_2"
          - "ruby_3_1"
    runs-on: ubuntu-latest
    name: Ruby releases (FIPS=${{ matrix.fips}})
    steps:
      - name: Install OS Dependencies
        run: |
          sudo apt-get update
          sudo apt-get -y --no-install-recommends install cmake gcc ninja-build golang make autoconf ruby libyaml-dev
      - uses: actions/checkout@v3
      - name: Build AWS-LC, build ruby, run tests
        run: |
          ./tests/ci/integration/run_ruby_integration.sh ${{ matrix.branch }}
...
```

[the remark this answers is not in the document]
samuel40791765: Like Will mentioned, I'd expect the releases to be much more stable. Nothing really changes on these branches, especially in terms of OpenSSL functionality. Any commit that does get backported to previous branches would most likely appear on main first as well. The build logs are already also pretty clear on which specific branch is being tested.
If I'm wrong and these Ruby versions have more churn than expected, we could split up the dimension into separate dimensions. But from what I've observed, I think we can still follow the same strategy as Python to minimize noise and spinning up additional runners for now.

justsmth (dismissed):

tests/ci/integration/run_ruby_integration.sh
```diff
@@ -0,0 +1,85 @@
...
+
+    #TODO: add more relevant tests here
```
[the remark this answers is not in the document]
WillChilds-Klein: ok, do we have a SIM or something to track the follow-up?

tests/ci/integration/run_ruby_integration.sh
WillChilds-Klein: > miniruby appears to be minimal lighter version of ruby used in their build

right. i think we need some way to assert that the `miniruby` used to run the tests is actually linked against AWS-LC. otherwise, we might be testing against an OpenSSL-backed `miniruby`, which is meaningless to us.

samuel40791765: Added an assertion for miniruby

WillChilds-Klein (approved):

justsmth (approved):

PR #1568, opened 2024-05-01
Add support for BIO_read/write_ex; Update MySQL CI to 8.4;

samuel40791765:
### Description of changes: 
MySQL 8.4 has been released, this updates our CI to run against the latest version. This time we only need two missing APIs from OpenSSL.
* `BIO_read_ex`
* `BIO_write_ex`

### Call-outs:
* Removed disabled tests were removed from mysql's test suite, so I removed them from `mysql_run_tests`
* Original patch needed to be updated, new patch is needed for `int` to `size_t` mismatch for `BIO_pending`
     * I considered switching our `BIO_pending` to `int` instead, but our version has been using the `size_t` signature for a decade now, so this doesn't seem easily convertible.

### Testing:
* New tests for `BIO_read/write_ex`
* Update MySQL build for 8.4

### DIFF
--- a/crypto/bio/bio.c
+++ b/crypto/bio/bio.c
@@ -198,7 +198,7 @@ int BIO_read_ex(BIO *bio, void *data, size_t data_len, size_t *read_bytes) {
   }
 
   int ret = BIO_read(bio, data, read_len);
-  if (ret > 0) {
+  if (ret >= 0) {
     *read_bytes = ret;
     return 1;
   } else {
--- a/crypto/bio/bio.c
+++ b/crypto/bio/bio.c
@@ -198,10 +198,11 @@ int BIO_read_ex(BIO *bio, void *data, size_t data_len, size_t *read_bytes) {
   }
 
   int ret = BIO_read(bio, data, read_len);
-  *read_bytes = ret;
   if (ret > 0) {
+    *read_bytes = ret;
     return 1;
   } else {
+    *read_bytes = 0;
     return 0;
   }
 }
@@ -268,12 +269,15 @@ int BIO_write_ex(BIO *bio, const void *data, size_t data_len, size_t *written_by
   }
 
   int ret = BIO_write(bio, data, write_len);
-  if (written_bytes != NULL) {
-    *written_bytes = ret;
-  }
   if (ret > 0) {
+    if (written_bytes != NULL) {
+      *written_bytes = ret;
+    }
     return 1;
   } else {
+    if (written_bytes != NULL) {
+      *written_bytes = 0;
+    }
     return 0;
   }
 }
--- a/crypto/bio/bio.c
+++ b/crypto/bio/bio.c
@@ -186,6 +186,26 @@ int BIO_read(BIO *bio, void *buf, int len) {
   return ret;
 }
 
+int BIO_read_ex(BIO *bio, void *data, size_t data_len, size_t *read_bytes) {
+  if (bio == NULL || read_bytes == NULL) {
+    OPENSSL_PUT_ERROR(BIO, BIO_R_NULL_PARAMETER);
+    return 0;
+  }
+
+  int read_len = (int)data_len;
+  if (data_len > INT_MAX) {
+    read_len = INT_MAX;
+  }
+
+  int ret = BIO_read(bio, data, read_len);
+  *read_bytes = ret;
+  if (ret > 0) {
+    return 1;
+  } else {
+    return 0;
+  }
+}
+
 int BIO_gets(BIO *bio, char *buf, int len) {
   if (bio == NULL || bio->method == NULL || bio->method->bgets == NULL) {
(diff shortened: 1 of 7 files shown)


crypto/bio/bio.c
```diff
@@ -186,6 +186,26 @@ int BIO_read(BIO *bio, void *buf, int len) {
...
+    return 0;
+  }
+
+  int read_len = (int)data_len;
+  if (data_len > INT_MAX) {
+    read_len = INT_MAX;
+  }
+
+  int ret = BIO_read(bio, data, read_len);
+  *read_bytes = ret;
```
skmcgrail, line 201: Should this check to make sure that `BIO_read` didn't return a negative number first prior to setting `read_bytes` to that value? Otherwise you are potentially assigning a negative number in this case to an unsigned type. Then it looks like a weird scenario of "read X bytes but also had an error".

Maybe only update this value to `ret` in this `> 0` branch, otherwise set the value to `0` in the failure case.

crypto/bio/bio.c
```diff
@@ -236,6 +256,28 @@ int BIO_write(BIO *bio, const void *in, int inl) {
...
+
+  int write_len = (int)data_len;
+  if (data_len > INT_MAX) {
+    write_len = INT_MAX;
+  }
+
+  int ret = BIO_write(bio, data, write_len);
+  if (written_bytes != NULL) {
+    *written_bytes = ret;
+  }
```
skmcgrail, line 273: Similar to comment above about `ret` being a negative value.

crypto/bio/bio.c
```diff
@@ -236,6 +256,28 @@ int BIO_write(BIO *bio, const void *in, int inl) {
   return ret;
 }
 
+int BIO_write_ex(BIO *bio, const void *data, size_t data_len, size_t *written_bytes) {
+  if (bio == NULL) {
```
skmcgrail, line 260: Missing `written_bytes` null check

samuel40791765: It's a bit misaligned, but unlike `BIO_read_ex`, `written_bytes` is allowed to be NULL in OpenSSL. I've also documented this in the function description.

crypto/bio/bio.c
samuel40791765: ahh good catch, I'll make the change.

skmcgrail (dismissed):

crypto/bio/bio_test.cc
```diff
@@ -1063,3 +1063,40 @@ TEST(BIOTest, InvokeConnectCallback) {
...
+  ASSERT_TRUE(bio);
+
+  size_t written = 0;
+  ASSERT_TRUE(BIO_write_ex(bio.get(), "abcdef", 6, &written));
+  EXPECT_EQ(written, (size_t)6);
+
+  char buf[32];
+  size_t read = 0;
+  ASSERT_TRUE(BIO_read_ex(bio.get(), buf, sizeof(buf), &read));
+  EXPECT_GT(read, (size_t)0);
```
andrewhop, line 1078: Can this check for equality? This should be sizeof(buf) right?

andrewhop (commented):
Do we need to do anything to support https://github.com/aws/aws-lc/pull/1333 and handle https://github.com/openssl/openssl/issues/8208?

samuel40791765 (commented):
> Do we need to do anything to support https://github.com/aws/aws-lc/pull/1333 and handle https://github.com/openssl/openssl/issues/8208?

1. For https://github.com/aws/aws-lc/pull/1333, I'm not sure how `SSL_MODE_AUTO_RETRY` comes into play here since these are BIO APis we're adding here. But I've reverted the change in https://github.com/aws/aws-lc/commit/171ee7a65467c29f161a4a39f3ebe543f5f41789 since it wasn't needed anymore. `SSL_MODE_AUTO_RETRY` is also default behavior which we've documented.

2. For https://github.com/openssl/openssl/issues/8208, really good catch I didn't notice this issue.. Luckily nobody other than MySQL 8.4 is dependent on us for this yet, but it'd be better to implement the right behavior from the get go.
My only concern would be folks depending on the non-existent eof support `BIO_read_ex` has right now. But judging from the comments in the thread it seems like `BIO_read_ex` is unusable right now because of this, so it seems unlikely. I'll implement the suggestion from https://github.com/openssl/openssl/issues/8208#issuecomment-700753683 and document the discrepancy. Will also add tests.

samuel40791765:
We decided to pivot and align with OpenSSL's original behavior instead. Misaligning the behavior will only create more gaps with OpenSSL/AWS-LC with no real additional benefit. This will only create more confusion further down the road with future integration projects. It's also very unlikely that OpenSSL will change the current behavior of `BIO_read_ex` for the sake of backwards compatibility.
I've added documentation to discourage usage of the API and linked to the original OpenSSL issue.

andrewhop (approved):

skmcgrail (approved):

PR #1569, opened 2024-05-01
ppc64le: support OPENSSL_ppccap ENV variable

justsmth:
### Issues:
CryptoAlg-2425

### Description of changes: 
* Architectures for which we have custom assembly implementations need to provide a mechanism for the environment to alter the CPU capabilities that are detected at runtime.
* This PR adds support for setting a `OPENSSL_ppccap` environment variable for this purpose.

### Testing:
* Without setting the environment variable: 55ms
```
[ RUN      ] AESTest.TestVectors
[       OK ] AESTest.TestVectors (55 ms)
```
* After setting `export OPENSSL_ppccap=0x0000000000000000`: 216ms
```
[ RUN      ] AESTest.TestVectors
[       OK ] AESTest.TestVectors (216 ms)
```

### DIFF
--- a/crypto/fipsmodule/cpucap/cpu_intel.c
+++ b/crypto/fipsmodule/cpucap/cpu_intel.c
@@ -133,6 +133,8 @@ static void handle_cpu_env(uint32_t *out, const char *in) {
   const int or = in[0] == '|';
   const int skip_first_byte = invert || or;
   const int hex = in[skip_first_byte] == '0' && in[skip_first_byte+1] == 'x';
+  uint32_t intelcap0 = out[0];
+  uint32_t intelcap1 = out[1];
 
   int sscanf_result;
   uint64_t v;
@@ -146,15 +148,31 @@ static void handle_cpu_env(uint32_t *out, const char *in) {
     return;
   }
 
+  uint32_t reqcap0 = (uint32_t)(v & UINT32_MAX);
+  uint32_t reqcap1 = (uint32_t)(v >> 32);
+
+  // Detect if the user is trying to use the environment variable to set
+  // a capability that is _not_ available on the CPU.
+  // The case of invert cannot enable an unexisting capability;
+  // it can only disable an existing one.
+  if (!invert && (intelcap0 || intelcap1)) {
+    if((~intelcap0 & reqcap0) || (~intelcap1 & reqcap1)) {
+      fprintf(stderr,
+              "Fatal Error: HW capability found: 0x%02X 0x%02X, but HW capability requested: 0x%02X 0x%02X.\n",
+              intelcap0, intelcap1, reqcap0, reqcap1);
+      abort();
+    }
+  }
+
   if (invert) {
-    out[0] &= ~v;
-    out[1] &= ~(v >> 32);
+    out[0] &= ~reqcap0;
+    out[1] &= ~reqcap1;
   } else if (or) {
-    out[0] |= v;
-    out[1] |= (v >> 32);
+    out[0] |= reqcap0;
+    out[1] |= reqcap1;
   } else {
-    out[0] = v;
-    out[1] = v >> 32;
+    out[0] = reqcap0;
+    out[1] = reqcap1;
   }
 }
 
--- a/crypto/fipsmodule/cpucap/cpu_intel.c
+++ b/crypto/fipsmodule/cpucap/cpu_intel.c
@@ -156,7 +156,8 @@ static void handle_cpu_env(uint32_t *out, const char *in) {
   // The case of invert cannot enable an unexisting capability;
   // it can only disable an existing one.
   if (!invert && (intelcap0 || intelcap1)) {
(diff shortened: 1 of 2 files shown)


crypto/fipsmodule/cpucap/cpu_ppc64le.c
```diff
@@ -24,11 +24,56 @@
 #define PPC_FEATURE2_HAS_VCRYPTO 0x02000000
 #endif
 
+static void handle_cpu_env(unsigned long *out, const char *in) {
```
torben-hansen, line 27: Could put this function implementation in an internal header file instead of duplicating?

jvdsn:
I can't really test it because I don't have a ppc platform, but the code looks like what I expected.

crypto/fipsmodule/cpucap/cpu_ppc64le.c
justsmth: The existing implementations are very similar but not quite compatible with each other:
* Intel: https://github.com/aws/aws-lc/blob/33637ff1765823bef584c7959a5ef84e6f7bbc9f/crypto/fipsmodule/cpucap/cpu_intel.c#L131
* ARM64: https://github.com/aws/aws-lc/blob/33637ff1765823bef584c7959a5ef84e6f7bbc9f/crypto/fipsmodule/cpucap/cpu_aarch64.c#L8

This implementation does appear to be compatible with "Intel", except the "Intel" implementation assumes little-endian while this one should also work for big-endian.

nebeid: Do we want to handle in intel and ppc the case of trying to enable a non-existent capability as in arm https://github.com/aws/aws-lc/blob/33637ff1765823bef584c7959a5ef84e6f7bbc9f/crypto/fipsmodule/cpucap/cpu_aarch64.c#L27

skmcgrail:
Does this need any any special handling in the delocater? I feel like armcap and ia32 both have special delocate logic handling. Not sure if FIPS is in play for this architecture in question.

justsmth:
> Does this need any any special handling in the delocater? I believe `OPENSSL_armcap_P` and `OPENSSL_ia32cap_P` both have special delocate logic handling. Not sure if FIPS is in play for this architecture in question.

Since this change is only updating the implementation of `OPENSSL_cpuid_setup` so it can be affected by an environment variable, I don't anticipate it having any impact on the delocater. There are no changes to where/how `OPENSSL_ppc64le_hwcap2` is stored.

crypto/fipsmodule/cpucap/cpu_ppc64le.c
justsmth: > Do we want to handle in intel and ppc the case of trying to enable a non-existent capability as in arm.

This would invalidate the "OR" use-case [here](https://github.com/aws/aws-lc/blob/33637ff1765823bef584c7959a5ef84e6f7bbc9f/crypto/fipsmodule/cpucap/cpu_intel.c#L152) unless it's likewise safe to ignore the existing `out` values when they are `0`.

justsmth: Added rejection logic in latest commit.

crypto/fipsmodule/cpucap/cpu_intel.c
```diff
@@ -133,28 +133,47 @@ static void handle_cpu_env(uint32_t *out, const char *in) {
...
 
+  uint32_t reqcap0 = (uint32_t)(v & UINT32_MAX);
+  uint32_t reqcap1 = (uint32_t)(v >> 32);
+
+  // Detect if the user is trying to use the environment variable to set
+  // a capability that is _not_ available on the CPU.
+  // The case of invert cannot enable an unexisting capability;
+  // it can only disable an existing one.
+  if (!invert && (intelcap0 || intelcap1)) {
+    // Still allow RDRAND bit to be set for testing
```
torben-hansen, line 159: I think you mean that this reserved bit indicates Intel cpu?

justsmth: Ahh.. you're right!  I update the comment.

nebeid: This is not very clear to me. We could have disabled the RDRAND, e.g. in https://github.com/aws/aws-lc/blob/5cd1b450cdddafa080f40ba666df5fdfd162eaab/crypto/fipsmodule/cpucap/cpu_intel.c#L220-L225 and still allow the environment variable to set it?

justsmth: The RDRAND (ecx) bit is different.  The bit we're allowing to be set is the one that indicated "intel", which is the flag unset by this line: https://github.com/aws/aws-lc/blob/5cd1b450cdddafa080f40ba666df5fdfd162eaab/crypto/fipsmodule/cpucap/cpu_intel.c#L249

nebeid (approved):

PR #1574, opened 2024-05-02
Design for support of HMAC precomputed keys

fabrice102:
### Issues:
N/A

### Description of changes: 

This commit implements an optimization of HMAC described in NIST-FIPS-198-1 (Section 6) and in RFC2104 (Section 4).

Instead of using the original key, it is possible to use a precomputed key consisting of two intermediate hashing results. Using this precomputed key removes at least two calls to the hash compression function when computing a HMAC value.

The first commit from May 06, 2024 is an initial design. As such, only SHA-256 is supported and all other hashes are disabled from HMAC. This makes some unit tests fail. Once this design is validated, we will add back the other hash functions in additional commits.

Changes:
* May 6, 2024: this PR was split in two. The ability to export state of SHA-256 was moved to https://github.com/aws/aws-lc/pull/1577

### Call-outs:

* This commit adds a few fields to the `methods` field of the HMAC context. This slightly increases the size of the HMAC_context (by 24 bytes on a 64-bit architecture).

### Testing:

Testing was added in two places:

* the C++ test harness based on Google Test (file `crypto/hmac_extra/hmac_test.cc`)
* the ACVP harness (file `util/fipstools/acvp/modulewrapper/modulewrapper.cc`)

To allow partial testing, we updated hmac_tests.txt to only contain the HMAC-SHA256 tests. The original file is hmac_tests_full.txt

Testing can be done in the CMake build folder using:

* `crypto/crypto_test --gtest_filter='*InitAndGetStateSHA256*:HMACTest.TestVectors'` for the new/changed unit tests
* `ninja acvp_tests` for the ACVP tests (failure is expected for anything using HMAC-xx where xx is not SHA-256)


include/openssl/hmac.h
```diff
@@ -151,6 +151,80 @@ OPENSSL_EXPORT int HMAC_CTX_copy_ex(HMAC_CTX *dest, const HMAC_CTX *src);
...
+
+// HMAC_Init_from_precomputed_key sets up an initialised |HMAC_CTX| to use |md|
+// as the hash function and |precomputed_key| as the precomputed key (see
+// HMAC_get_precomputed_key). For a non-initial call, |md| may be NULL, in which
+// case the previous hash function will be used. If the hash function has not
+// changed and |precomputed_key| is NULL, |ctx| reuses the previous key. It
+// returns one on success or zero on failure.
+//
+// It is possible to mix and match HMAC_Init_ex and
+// HMAC_Init_from_precomputed_key.
```
nebeid, line 210:
```suggestion
// It is possible to call HMAC_Init_ex after HMAC_Init_from_precomputed_key
// and vice-versa when the input key is NULL.
```
Is "vice-versa" correct? I understand that from the example below. Is it only when the key is NULL?

include/openssl/hmac.h
```diff
@@ -151,6 +151,80 @@ OPENSSL_EXPORT int HMAC_CTX_copy_ex(HMAC_CTX *dest, const HMAC_CTX *src);
...
+// If precomputed_key=NULL and md=previous MD or NULL,
+// and if HMAC_Init_ex or HMAC_Init_from_precomputed_key was successfully called
+// before, then
+//   HMAC_Init_ex(ctx, NULL, precomputed_key_len, md, engine=NULL)
+// is equivalent to
+//   HMAC_Init_from_precomputed_key(ctx, NULL, precomputed_key_len, md)
+//
+// Note: Contrary to HMAC_Init_ex, it is not possible to use the empty key with
+//   that function. Passing NULL to |precomputed_key| is only allowed on a
```
nebeid, line 219:
```suggestion
//   that function on an initial call. Passing NULL to |precomputed_key| is only allowed on a
```
Is that what is meant? Because it's allowed on a non-initial call as in the case of HMAC_Init_ex, right?

nebeid (commented):
Some comments on the header files. Will continue reviewing the implementation.

include/openssl/hmac.h
```diff
@@ -151,6 +151,80 @@ OPENSSL_EXPORT int HMAC_CTX_copy_ex(HMAC_CTX *dest, const HMAC_CTX *src);
...
+// been initialized via HMAC_Init_*, and neither HMAC_Update nor HMAC_Final
+// must have been called after the last HMAC_Init_ex. It returns one on success
+// and zero on programmer error.
+// After a successful call to HMAC_set_precomputed_key_export, HMAC_Update and
+// HMAC_Final will fail.
+//
+// Note: The main reason for this function is to satisfy FIPS assertion AS09.16,
+// since HMAC_get_precomputed_key returns key material (i.e., a CSP in NIST
+// terminology).
+OPENSSL_EXPORT int HMAC_set_precomputed_key_export(HMAC_CTX *ctx);
```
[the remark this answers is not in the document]
fabrice102: I like better the second option (`HMAC_get_precomputed_key` can just set `out_len` to the correct value when out is NULL.)
I can implement it if you think it's better than having a separate function.

include/openssl/hmac.h
```diff
@@ -151,6 +151,80 @@ OPENSSL_EXPORT int HMAC_CTX_copy_ex(HMAC_CTX *dest, const HMAC_CTX *src);
...
+
+// HMAC_precompute_size returns the size, in bytes, of the HMAC precomputed key
```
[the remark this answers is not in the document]
fabrice102: Yes, it's a better name. Once we decide whether we keep this function or not (according to comment above), I'll make the change.

include/openssl/hmac.h
```diff
@@ -151,6 +151,80 @@ OPENSSL_EXPORT int HMAC_CTX_copy_ex(HMAC_CTX *dest, const HMAC_CTX *src);
...
+// |HMAC_Init_ex|.
+OPENSSL_EXPORT size_t HMAC_precomputed_key_size(const HMAC_CTX *ctx);
+
+// HMAC_get_precomputed_key writes the precomputed key to |out| and sets
+// |*out_len| to the length of the result. On entry, the function
+// HMAC_set_precomputed_key_export must have been called on |ctx|. Furthermore,
+// |out| must contain at least |HMAC_precomputed_key_size| bytes of space. An
+// output size of |HMAC_MAX_PRECOMPUTED_KEY_SIZE| will always be large enough.
+// After a successful call to HMAC_get_precomputed_key, the context need to be
+// set up again by HMAC_Init_*, like after a call to HMAC_Final.
```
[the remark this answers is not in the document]
fabrice102: I meant `HMAC_Init_ex` or `HMAC_Init_from_precomputed_key`. We can also spell them out too.

include/openssl/hmac.h
fabrice102: Yes, and "vice-versa" is correct.
And it also works when the input key is not NULL. In that case, the key is changed.

include/openssl/hmac.h
fabrice102: Indeed, it's not completely clear. What I mean is the following:
HMAC can be set with an empty key.

This can be done using HMAC_Init and a key=NULL.

HMAC cannot be set with an empty precomputed key. This would actually not make sense from a correctness point of view even.

Here is a suggestion for the wording:

```
// Note: Contrary to input keys to HMAC_Init_ex, which can be the empty key, 
//   an input precomputed key (for HMAC_Init_from_precomputed_key) 
//   cannot be the empty key. The notion of empty precomputed key is not
//   well-defined.
//   Passing NULL to |precomputed_key| is only allowed on a
```

include/openssl/digest.h
```diff
@@ -171,6 +171,12 @@ OPENSSL_EXPORT int EVP_DigestUpdate(EVP_MD_CTX *ctx, const void *data,
 // at least this much space.
 #define EVP_MAX_MD_SIZE 64  // SHA-512 is the longest so far.
 
+// EVP_MAX_MD_CHAINING_LENGTH is the largest chaining length supported, in
+// bytes. This constant is only for Merkle-Damgard-based hashed functions
+// like SHA-1, SHA-2, and MD5.
+// This constant is only used internally by HMAC.
+#define EVP_MAX_MD_CHAINING_LENGTH 64  // SHA-512 has the longest chaining length so far
```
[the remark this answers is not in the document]
fabrice102: The chaining length is not necessarily equal to the digest length: for example SHA-224 chaining length is 32 bytes, not 24 bytes.

The maximum chaining length is currently equal to the maximum digest length, but this is not guaranteed in the future. For example, if NIST creates SHA2-768 using a chaining length of 128 bytes, then the maximum digest length would be 96 bytes but the maximum chaining length would be 128 bytes. Another example is if NIST adds a SHA-4 with digest length 128 bytes which does not use a Merkle-Damgard construction. Then, the maximum chaining length would stay at 64 bytes, while the maximum digest length would increase to 128 bytes.

include/openssl/hmac.h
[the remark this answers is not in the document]
fabrice102: Thinking more about it, I think there is a concrete difference between having a separate function to get the size or not:

* with a separate function (like in the current code), there is no restriction when the function to get the size can be called (as long as the md has been initialized)
* when merging this function with `HMAC_set_precomputed_key_export` or `HMAC_get_precomputed_key`, the size can only be retrieved in a specific state (i.e., just after initialization or just after setting the key export flag respectively).

Having a separate function gives a bit more of flexibility. But I don't know if it really matters to users of the library.

nebeid (commented):
I'm continuing to review hmac.c; I was focused on the state transition and will continue reviewing the new functions.

include/openssl/digest.h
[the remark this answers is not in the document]
fabrice102: Do you mean keeping the name `EVP_MAX_MD_CHAINING_LENGTH`?

include/openssl/hmac.h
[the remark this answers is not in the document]
fabrice102: That's a good point. I'll switch to this state logic.

include/openssl/digest.h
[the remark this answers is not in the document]
nebeid: Sorry, Fabrice, I didn't write what I was thinking. I wanted to suggest `EVP_MAX_MD_STATE_SIZE` instead.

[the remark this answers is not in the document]
nebeid: On a second thought, `state_size` may be confusing because the state such as `struct sha256_state_st` for example in `sha.h` contains other values. If "chaining_length" is the correct term used in standards or elsewhere, we can keep it. I suggest you explain it here as the internal state of the MD algorithm or something of the sort.

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -62,18 +62,24 @@
...
-typedef int (*HashUpdate)(void *, const void*, size_t);
-typedef int (*HashFinal)(uint8_t *, void*);
+typedef int (*HashUpdate)(void *, const void *, size_t);
+typedef int (*HashFinal)(uint8_t *, void *);
+typedef int (*HashInitFromState)(void *, const uint8_t *, uint64_t);
+typedef int (*HashGetState)(void *, uint8_t *, uint64_t *);
 
 struct hmac_methods_st {
   const EVP_MD* evp_md;
+  size_t chaining_length;  // chaining length in bytes
```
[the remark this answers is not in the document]
nebeid: Please see my most recent comment on `EVP_MAX_MD_CHAINING_LENGTH` in digest.h revisiting my preference.

crypto/fipsmodule/sha/internal.h
```diff
@@ -92,6 +103,26 @@ void sha512_block_data_order(uint64_t *state, const uint8_t *in,
 #define KECCAK1600_ASM
 #endif
 
+// SHA256_Init_from_state is a low-level function that initializes |sha| with a
+// custom state. |h| is the hash state in big endian. |n| is the number of bits
+// processed at this point. It must be a multiple of |SHA256_CBLOCK*8|.
+// It returns one on success and zero on programmer error.
+// External users should never use directly this function.
```
nebeid, line 110:
```suggestion
// This function is for internal use only and should never be directly called.
```

crypto/fipsmodule/sha/internal.h
```diff
@@ -92,6 +103,26 @@ void sha512_block_data_order(uint64_t *state, const uint8_t *in,
...
+    SHA256_CTX *sha, const uint8_t h[SHA256_CHAINING_LENGTH], uint64_t n);
+
+// SHA256_get_state is a low-level function that exports the hash state in big
+// endian into |out_n| and the number of bits processed at this point in
+// |out_n|. SHA256_Final must not have been called before (otherwise results
+// are not guaranteed). Furthermore, the number of bytes processed by
+// SHA256_Update must be a multiple of the block length |SHA256_CBLOCK|
+// (otherwise it fails). It returns one on success and zero on programmer
+// error.
+// External users should never use directly this function.
```
nebeid, line 121: Same suggestion.

crypto/fipsmodule/sha/sha256.c
```diff
@@ -163,14 +198,41 @@ static int sha256_final_impl(uint8_t *out, size_t md_len, SHA256_CTX *c) {
   return 1;
 }
 
-int SHA256_Final(uint8_t out[SHA256_DIGEST_LENGTH], SHA256_CTX *c) {
+int SHA256_Final(uint8_t out[SHA256_CHAINING_LENGTH], SHA256_CTX *c) {
```
nebeid, line 201: Why is this change needed? If not needed, we prefer not to change the function signature between its declaration in sha.h and its definition here.

nebeid (commented):
After you make changes, Fabrice, I'll spend more time on the new implementations in sha256.c and the tests.

crypto/fipsmodule/sha/internal.h
```diff
@@ -92,6 +103,26 @@ void sha512_block_data_order(uint64_t *state, const uint8_t *in,
 #define KECCAK1600_ASM
 #endif
 
+// SHA256_Init_from_state is a low-level function that initializes |sha| with a
+// custom state. |h| is the hash state in big endian. |n| is the number of bits
+// processed at this point. It must be a multiple of |SHA256_CBLOCK*8|.
+// It returns one on success and zero on programmer error.
+// External users should never use directly this function.
+OPENSSL_EXPORT int SHA256_Init_from_state(
+    SHA256_CTX *sha, const uint8_t h[SHA256_CHAINING_LENGTH], uint64_t n);
```
nebeid, line 112: Do we need to pass the value `n` or we can assume it's exactly one block that was processed? Will there be other use-cases? Which is more error prone if we call this function internally elsewhere; passing it or assuming it?

crypto/fipsmodule/sha/sha256.c
fabrice102: Good catch! This is a mistake. This should not have been changed.

crypto/fipsmodule/sha/internal.h
fabrice102: For HMAC precompute, `n` will indeed always be 1. I made this function slightly more generic than needed for HMAC precompute in case we need it for other purposes. Also, I thought it would be a bit strange to have a function that only work for `n=1`. (I was not even sure how to name it in that case: `SHA256_Init_from_state_after_one_block`?)

include/openssl/digest.h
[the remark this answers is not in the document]
fabrice102: It is the term used by openssl/AWS LC there:
https://github.com/aws/aws-lc/blob/4e54dd8363396f257d7a2317c48101e18170e6fb/crypto/fipsmodule/digest/md32_common.h#L78-L80
But it's true I have not really found any other place where it's used.

[Wikipedia](https://en.wikipedia.org/wiki/SHA-2) uses "internal state size".
But the issue with "state size" is that it may be confused with the size of `struct sha256_state_st` which is bigger than just the chaining length.

[FIPS 180-4](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf) uses the term "working variables", but "working variables length" would be very confusing.

So at the end of the day, I think chaining length might be the best option.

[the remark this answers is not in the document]
nebeid: Thanks, Fabrice. I agree to keeping it as "chaining_length" with the added explanation here similarly to md32_common.h: "this is the internal state size without truncation (e.g. 32 for SHA-224 and SHA-256, 64 for SHA-384 and SHA-512)". Does this sound reasonable?

include/openssl/hmac.h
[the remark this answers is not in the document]
fabrice102: >  If it's given a non-NULL out, it checks that out_len is not less than a certain minimum (otherwise an error is returned)

What is the purpose of testing that out_len is "not less than a certain minimum (otherwise an error is returned)"?
I tried to look at a few functions with a `*out_len` in AWS-LC and none of them checked `out_len` before setting it to the actual output length.

include/openssl/hmac.h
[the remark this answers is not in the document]
fabrice102: I've removed the function following the comments below.

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -393,6 +477,125 @@ void HMAC_CTX_reset(HMAC_CTX *ctx) {
...
+  // state, so we lock the state here. Technically this is not really needed,
+  // because the functions we call should not update the indicator state.
+  // But this is safer.
+  FIPS_service_indicator_lock_state();
+  int result = 0;
+
+  // Initialize i_ctx from the state stored in the first part of precomputed_key
+  // Recall that i_ctx is the state of the hash function after processing
+  // one block (ipad xor keyOrHashedKey)
+  if (!methods->init_from_state(&ctx->i_ctx, precomputed_key, block_size * 8)) {
```
nebeid, line 575: Do we need to zero out `i_ctx` and `o_ctx` first? Or maybe zero out the `data` and `num` members of them in the implementation of `init_from_state`?

nebeid (commented):
I will review the tests shortly.

crypto/fipsmodule/hmac/hmac.c
fabrice102: Yes, `init_from_state` starts by zeroing out everything:
https://github.com/aws/aws-lc/pull/1574/files#diff-45faa6f87b7148d297f13d80c973233a2dbfc4670d75621f402f729a9b109981R110

include/openssl/hmac.h
```diff
@@ -206,17 +210,18 @@ OPENSSL_EXPORT int HMAC_get_precomputed_key(HMAC_CTX *ctx, uint8_t *out,
 // changed and |precomputed_key| is NULL, |ctx| reuses the previous key. It
 // returns one on success or zero on failure.
 //
-// It is possible to mix and match HMAC_Init_ex and
-// HMAC_Init_from_precomputed_key.
-// If precomputed_key=NULL and md=previous MD or NULL,
+// |HMAC_Init_ex| and |HMAC_Init_from_precomputed_key| are interchangeable when
```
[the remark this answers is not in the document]
fabrice102: Looks nicer indeed.

include/openssl/hmac.h
```diff
@@ -151,6 +151,76 @@ OPENSSL_EXPORT int HMAC_CTX_copy_ex(HMAC_CTX *dest, const HMAC_CTX *src);
...
+// HMAC_get_precomputed_key can directly be called again with a non-null |out|.
+// But HMAC_Update and HMAC_Final will still fail.
+//
+// If |out| is not NULL, |out| must contain at least |out_len| bytes of space
+// (as output by HMAC_get_precomputed_key). An output size of
+// |HMAC_MAX_PRECOMPUTED_KEY_SIZE| will always be large enough. After a
+// successful call to HMAC_get_precomputed_key with a non-NULL |out|, the
+// context can directly be used for computing an HMAC, as if it was set up by
+// HMAC_Init_*. In particular, in particular, HMAC_Update and HMAC_Final can be
+// called. Further calls to HMAC_get_precomputed_key will fail.
```
nebeid, line 186:
```suggestion
// successful call to HMAC_get_precomputed_key with a non-NULL |out|, the
// |ctx| is unchanged and HMAC_Update and HMAC_Final can be called.
```

include/openssl/hmac.h
```diff
@@ -151,6 +151,76 @@ OPENSSL_EXPORT int HMAC_CTX_copy_ex(HMAC_CTX *dest, const HMAC_CTX *src);
...
+//
+// The function returns one on success and zero on programmer error.
```
nebeid, line 188: I think we can remove "programmer" here and on l.163.

include/openssl/hmac.h
nebeid, line 186: I think "Further calls ..." is not needed because it was mentioned above that it needs to be preceded by `HMAC_set_precomputed_key_export`.

include/openssl/hmac.h
```diff
@@ -151,6 +151,76 @@ OPENSSL_EXPORT int HMAC_CTX_copy_ex(HMAC_CTX *dest, const HMAC_CTX *src);
...
+// HMAC_Final will fail.
+//
+// Note: The main reason for this function is to satisfy FIPS assertion AS09.16,
+// since HMAC_get_precomputed_key returns key material (i.e., a CSP in NIST
+// terminology).
+OPENSSL_EXPORT int HMAC_set_precomputed_key_export(HMAC_CTX *ctx);
+
+// HMAC_get_precomputed_key writes the precomputed key to |out| and sets
+// |*out_len| to the length of the result. On entry, the function
+// HMAC_set_precomputed_key_export must have been called on |ctx|.
```
nebeid, line 174:
```suggestion
// |HMAC_set_precomputed_key_export| must have been called on |ctx|.
```

crypto/fipsmodule/hmac/hmac.c
nebeid: Sorry, missed that, thank you.

include/openssl/hmac.h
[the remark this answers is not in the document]
nebeid: This is an example I found https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/evp/p_hkdf.c#L123.
It's a good idea also to put `OPENSSL_PUT_ERROR` as in that example where it makes sense to let the user know what went wrong.

include/openssl/hmac.h
fabrice102: I used the term "programmer error" to mean that the error can only come from calling wrongly the function but not from any other source of error (e.g., out-of-memory, non-authenticated ciphertext if it was decryption, ...). This is how I understood the term there: https://github.com/aws/aws-lc/blob/37ba0e2efc68407e7b20c71a7f4332846a76a7fb/include/openssl/sha.h#L198

But apparently `sha.h` is the only place where the term is used this way.

Should we just remove "programmer"? Or use another term to indicate the same idea?

include/openssl/hmac.h
fabrice102: Actually, the `|ctx|` is changed to allow calls to `HMAC_Update` and `HMAC_Final`.
Indeed, after calling `HMAC_set_precomputed_key_export`, the `|ctx|` is put in a state that does not allow calling `HMAC_Update` and `HMAC_Final`.

include/openssl/hmac.h
nebeid: Yeah, I think we can drop "programmer".

include/openssl/hmac.h
nebeid: Thanks, Fabrice. Then may be we can just shorten it to
```Suggestion
// successful call to HMAC_get_precomputed_key with a non-NULL |out|, the
// context can be directly used for computing an HMAC using |HMAC_Update| and |HMAC_Final|.
```

fabrice102: Committing this in the next commit. But should we also mention 
```
Further calls to |HMAC_get_precomputed_key| will fail.
```

include/openssl/hmac.h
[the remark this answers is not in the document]
fabrice102: I've modified the behavior to match the above.

crypto/fipsmodule/sha/internal.h
fabrice102: Is it ok for me to resolve this comment?

util/fipstools/acvp/modulewrapper/modulewrapper.cc
```diff
@@ -1855,12 +1856,30 @@ static bool TDES_CBC(const Span<const uint8_t> args[], ReplyCallback write_reply
 template <const EVP_MD *HashFunc()>
 static bool HMAC(const Span<const uint8_t> args[], ReplyCallback write_reply) {
   const EVP_MD *const md = HashFunc();
+
+  // Normal HMAC computation
```
nebeid, line 1860:
```suggestion
  // Approved HMAC computation
```

util/fipstools/acvp/modulewrapper/modulewrapper.cc
```diff
@@ -1855,12 +1856,30 @@ static bool TDES_CBC(const Span<const uint8_t> args[], ReplyCallback write_reply
...
   uint8_t digest[EVP_MAX_MD_SIZE];
   unsigned digest_len;
   if (::HMAC(md, args[1].data(), args[1].size(), args[0].data(), args[0].size(),
              digest, &digest_len) == nullptr) {
     return false;
   }
+
+  // HMAC computation with precomputed keys
```
nebeid, line 1868:
```suggestion
  // HMAC computation with precomputed keys
  // The purpose of this call is to test |HMAC_set_precomputed_key_export| and
  // |HMAC_get_precomputed_key|, which are called by |HMAC_with_precompute|.
```

crypto/fipsmodule/hmac/internal.h
```diff
@@ -0,0 +1,28 @@
...
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
+// HMAC_with_precompute is only used in FIPS ACVP harness, in order to test
+// the computation of HMAC using precomputed keys (internally). It should not
+// be used for any other purposes as it outputs the same results as |HMAC| and is
+// slower than |HMAC|.
+// This function does not update the FIPS service indicator.
+OPENSSL_EXPORT uint8_t *HMAC_with_precompute(const EVP_MD *evp_md,
```
nebeid, line 18: A suggestion, maybe call it `HMAC_with_precomputed_key`?

crypto/hmac_extra/hmac_test.cc
```diff
@@ -227,20 +235,71 @@ TEST(HMACTest, TestVectors) {
...
+    ASSERT_FALSE(HMAC_get_precomputed_key(ctx2.get(), precomputed_key, &precomputed_key_len_out));
+
+    // Get the precomputed key length for later use
+    // And test the precomputed key size is at most HMAC_MAX_PRECOMPUTED_KEY_SIZE
+    ASSERT_TRUE(HMAC_set_precomputed_key_export(ctx.get()));
+    size_t precomputed_key_len;
+    HMAC_get_precomputed_key(ctx.get(), nullptr, &precomputed_key_len);
+    ASSERT_LE(precomputed_key_len, (size_t) HMAC_MAX_PRECOMPUTED_KEY_SIZE);
+
+    // Test that at this point, the context can be used with HMAC_Update
```
nebeid, line 260:
```suggestion
    // Test that at this point, the context cannot be used with HMAC_Update
```

crypto/hmac_extra/hmac_test.cc
```diff
@@ -227,20 +235,71 @@ TEST(HMACTest, TestVectors) {
...
+    OPENSSL_memset(mac.get(), 0, expected_mac_len); // Clear the prior correct answer
+
+    // Test that an HMAC_CTX may be reset with the same key and a null md
     ASSERT_TRUE(HMAC_Init_ex(ctx.get(), nullptr, 0, nullptr, nullptr));
     ASSERT_TRUE(HMAC_Update(ctx.get(), input.data(), input.size()));
     ASSERT_TRUE(HMAC_Final(ctx.get(), mac.get(), &mac_len));
     EXPECT_EQ(Bytes(output), Bytes(mac.get(), mac_len));
     OPENSSL_memset(mac.get(), 0, expected_mac_len);  // Clear the prior correct answer
 
+    // Same test but using the Init_with_precompute instead
```
nebeid, line 296:
```suggestion
    // Same test but using the Init_from_precomputed_key instead
```
Would you like to change to this in all comments?

crypto/fipsmodule/hmac/internal.h
fabrice102: The reason I called it `HMAC_with_precompute` was to indicate it does precompute internally but does not take as input a precomputed key.

fabrice102:
I've added back all the hash functions.
I've also added tests to the service indicator for HMAC with precomputed key.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -1338,6 +1338,37 @@ TEST_P(HMACServiceIndicatorTest, HMACTest) {
...
+                    ctx.get(), precomputed_key, precomputed_key_len, digest)));
+  EXPECT_EQ(approved, AWSLC_NOT_APPROVED);
+  CALL_SERVICE_AND_CHECK_APPROVED(approved,
+                                  ASSERT_TRUE(HMAC_Update(ctx.get(), kPlaintext, sizeof(kPlaintext))));
+  EXPECT_EQ(approved, AWSLC_NOT_APPROVED);
+  CALL_SERVICE_AND_CHECK_APPROVED(approved,
+                                  ASSERT_TRUE(HMAC_Final(ctx.get(), mac.data(), &mac_len)));
+  EXPECT_EQ(approved, test.expect_approved);
+  EXPECT_EQ(Bytes(test.expected_digest, expected_mac_len),
+            Bytes(mac.data(), mac_len));
```
nebeid, line 1371: Looks good. I think we need to check that `HMAC_with_precompute` sets the service indicator to `AWSLC_NOT_APPROVED`.

crypto/digest_extra/digest_test.cc
```diff
@@ -477,45 +478,93 @@ TEST(DigestTest, TransformBlocks) {
...
-// FIXME: Need to implement this for all hash functions used by HMAC
-TEST(DigestTest, InitAndGetStateSHA256) {
-  const size_t nb_blocks = 10;
-  const size_t block_size = SHA256_CBLOCK;
-  uint8_t data[block_size * nb_blocks];
-  for (size_t i = 0; i < sizeof(data); i++) {
-    data[i] = i*3;
+// DIGEST_TEST_InitAndGetState_Body expands to the body of the various
+// InitAndGetState* tests below.
+// Because EXPECT_* do not output the proper line when used in a macro,
```
nebeid, line 483: Maybe you can try `SCOPED_TRACE(hash1)` (or (ctx1)) https://github.com/google/googletest/blob/main/docs/advanced.md#adding-traces-to-assertions, not sure if it will help.

fabrice102: The issue we have is that `EXPECT_*` would log the line where the macro is called/expanded, rather than the line in the definition of the macro. `SCOPED_TRACE` would unfortunately not helped in that case. I'm updating the comment in the code to make it clearer.

crypto/fipsmodule/sha/sha512.c
```diff
@@ -141,6 +141,63 @@ int SHA512_256_Init(SHA512_CTX *sha) {
...
+  sha->md_len = md_len;
+
+  const size_t out_words = SHA512_CHAINING_LENGTH / 8;
+  for (size_t i = 0; i < out_words; i++) {
+    sha->h[i] = CRYPTO_load_u64_be(h);
+    h += 8;
+  }
+
+  sha->Nh = n >> 32;
+  sha->Nl = n & 0xffffffff;
```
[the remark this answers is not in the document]
fabrice102: Oh, I completely missed that `Nl/Nh` are `uint64_t` for SHA-512 as opposed to `uint32_t` everywhere else. Good catch! Since `n` is `uint64_t`, we can just set `Nl` to `n`. I'm adding comments to the `*_get_state` function to indicate that they can only be called if fewer than 2^64 bits (i.e., 2^61 bytes) have been processed.

crypto/fipsmodule/sha/sha512.c
```diff
@@ -353,6 +410,48 @@ static int sha512_final_impl(uint8_t *out, size_t md_len, SHA512_CTX *sha) {
...
+    return 0;
+  }
+
+  const size_t out_words = SHA512_CHAINING_LENGTH / 8;
+  for (size_t i = 0; i < out_words; i++) {
+    CRYPTO_store_u64_be(out_h, ctx->h[i]);
+    out_h += 8;
+  }
+
+  *out_n = (((uint64_t)ctx->Nh) << 32) + ctx->Nl;
```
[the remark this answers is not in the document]
fabrice102: `*out_n` is currently `uint64_t`, so the first fix is good. And I'm adding a test at the beginning of the function that makes it fail if `Nh != 0`

crypto/fipsmodule/sha/sha512.c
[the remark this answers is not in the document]
fabrice102: I'll add a unit test to catch this regression on Monday.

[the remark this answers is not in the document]
fabrice102: I've added the unit test.

crypto/fipsmodule/sha/sha512.c
```diff
@@ -141,6 +141,63 @@ int SHA512_256_Init(SHA512_CTX *sha) {
...
+  OPENSSL_memset(sha, 0, sizeof(SHA512_CTX));
+  sha->md_len = md_len;
+
+  const size_t out_words = SHA512_CHAINING_LENGTH / 8;
+  for (size_t i = 0; i < out_words; i++) {
+    sha->h[i] = CRYPTO_load_u64_be(h);
+    h += 8;
+  }
+
+  sha->Nh = n >> 32;
```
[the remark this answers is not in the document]
fabrice102: Since `n` is `uint64_t`, I've changed to `sha->Nl = n`.

crypto/digest_extra/digest_test.cc
```diff
@@ -527,44 +528,176 @@ TEST(DigestTest, TransformBlocks) {
...
-// those as tests.
 
-TEST(DigestTest, InitAndGetStateMD5) {
-  DIGEST_TEST_InitAndGetState_Body(MD5, MD5_CTX, MD5_CBLOCK);
+// DIGEST_TEST_InitAndGetStateLarge_Body expands to the body of the various
+// InitAndGetStateLarge* tests below.
+// Compared to DIGEST_TEST_InitAndGetStateBasic_Body, it allows testing
+// hashing an arbitrary amount of data.
+// Of particular interest is the choice of NB_1000_BLOCKS to be such that
+// the number of hash bits n > 2^32, therefore overflowing uint32_t (used for
```
nebeid, line 547: I understand that you're trying to overflow `Nl` so that `Nh` gets updated and the hash continues operation seamlessly and successfully. If that's the case, I think whenever overflow is mentioned, here and further below, it may need to be clarified that it's not an error condition, just the normal operation.

crypto/digest_extra/digest_test.cc
```diff
@@ -527,44 +528,176 @@ TEST(DigestTest, TransformBlocks) {
...
+    }                                                                          \
+                                                                               \
+    /* Compute the hash of the NB_1000_BLOCKS+2 times the data_1000_blocks for \
+     * the baseline */                                                         \
+    HASH_CTX ctx1;                                                             \
+    EXPECT_TRUE(HASH_NAME##_Init(&ctx1)) << "init 1";                          \
+    for (size_t i = 0; i < NB_1000_BLOCKS + 2; i++) {                          \
+      EXPECT_TRUE(HASH_NAME##_Update(&ctx1, data_1000_blocks,                  \
+                                     sizeof(data_1000_blocks)))                \
+          << "update 1 - i=" << i;                                             \
```
nebeid, line 569: I don't remember if between quotes of a string you can concatenate `HASH_NAME##` so that the output messages are more specific (here and the macro above).

crypto/fipsmodule/sha/internal.h
```diff
@@ -126,19 +108,18 @@ void sha512_block_data_order(uint64_t *state, const uint8_t *in,
 // SHA1_Init_from_state is a low-level function that initializes |sha| with a
```
nebeid, line 108: I think the repetition of the comments on the functions below can be reduced in the same way as with the macros. That would be more readable, I believe.

crypto/digest_extra/digest_test.cc
fabrice102: Because the macro is expanded into the actual test function, we already can know where it comes from (as the function name is included in the output).
Concretely, here is an actual output when forcing an error:

```
$ ./crypto/crypto_test --gtest_filter='DigestTest.InitAndGetState*'
Note: Google Test filter = DigestTest.InitAndGetState*
[==========] Running 16 tests from 1 test suite.
[----------] 16 tests from DigestTest
[ RUN      ] DigestTest.InitAndGetStateMD5Basic
[...]/digest_test.cc:624: Failure
Value of: MD5_get_state(&ctx2, state_h, &state_n)
  Actual: true
Expected: false
get_state 2b
[  FAILED  ] DigestTest.InitAndGetStateMD5Basic (0 ms)
```

nebeid (dismissed):

crypto/fipsmodule/sha/internal.h
```diff
@@ -92,6 +105,54 @@ void sha512_block_data_order(uint64_t *state, const uint8_t *in,
...
+// SHAx_get_state is a low-level function that exports the hash state in big
+// endian into |out_n| and the number of bits processed at this point in
+// |out_n|. |SHAx_Final| must not have been called before (otherwise results
+// are not guaranteed). Furthermore, the number of bytes processed by
+// |SHAx_Update| must be a multiple of the block length |SHAy_CBLOCK| and
+// must be less than 2^61 (otherwise it fails). See comment above about
+// SHAx_Init_from_state for the definition of SHAy.
+// This function returns one on success and zero on error.
+// This function is for internal use only and should never be directly called.
+OPENSSL_EXPORT int SHA1_get_state(
```
torben-hansen, line 139: Does not need to be exported symbols(?)

fabrice102: Unfortunately, we need to export them for the tests to work, e.g., https://github.com/aws/aws-lc/pull/1574/files#diff-a486879d22cdbac91336850b85c6f1c88b0725a1115b3d2b7674326925c17225R486-R490

torben-hansen: Ahh you generate test cases by macro. Fair.
My lazy ctrl+F search didn't catch that.

include/openssl/hmac.h
```diff
@@ -203,4 +270,12 @@ BSSL_NAMESPACE_END
 
 #endif
 
+
+// Errors
+
+#define HMAC_R_MISSING_PARAMETERS 100
+#define HMAC_R_BUFFER_TOO_SMALL 102
+#define HMAC_R_SET_PRECOMPUTED_KEY_EXPORT_NOT_CALLED 103
+#define HMAC_R_NOT_CALLED_JUST_AFTER_INIT 104
```
skmcgrail, line 279: It's not documented well, but I think you will want to run `go run ./util/make_errors.go hmac` from the root of the repository after adding these. This will generate a `crypto/err/hmac.errordata` file that will be visible if you do a `git status`. It may also reformat the hmac.h header here as well.

You will then want to add that file to the list [here](https://github.com/aws/aws-lc/blob/main/crypto/CMakeLists.txt#L285).

fabrice102: Thanks! Done.

crypto/fipsmodule/sha/internal.h
```diff
@@ -92,6 +105,54 @@ void sha512_block_data_order(uint64_t *state, const uint8_t *in,
...
+    SHA512_CTX *sha, const uint8_t h[SHA384_CHAINING_LENGTH], uint64_t n);
+OPENSSL_EXPORT int SHA512_Init_from_state(
+    SHA512_CTX *sha, const uint8_t h[SHA512_CHAINING_LENGTH], uint64_t n);
+OPENSSL_EXPORT int SHA512_224_Init_from_state(
+    SHA512_CTX *sha, const uint8_t h[SHA512_224_CHAINING_LENGTH], uint64_t n);
+OPENSSL_EXPORT int SHA512_256_Init_from_state(
+    SHA512_CTX *sha, const uint8_t h[SHA512_256_CHAINING_LENGTH], uint64_t n);
+
+// SHAx_get_state is a low-level function that exports the hash state in big
+// endian into |out_n| and the number of bits processed at this point in
```
nebeid, line 131:
```suggestion
// endian into |out_h| and the number of bits processed at this point in
```

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -402,6 +488,143 @@ void HMAC_CTX_reset(HMAC_CTX *ctx) {
...
+  if (*out_len < actual_out_len) {
+    OPENSSL_PUT_ERROR(HMAC, HMAC_R_BUFFER_TOO_SMALL);
+    return 0;
+  }
+  *out_len = actual_out_len;
+
+  uint64_t i_ctx_n;
+  uint64_t o_ctx_n;
+
+  const int ok = ctx->methods->get_state(&ctx->i_ctx, out, &i_ctx_n) &&
```
nebeid, line 533: The windows arm64 build https://github.com/aws/aws-lc/actions/runs/9992803035/job/27618683893?pr=1574 is complaining about `o_ctx_n` not being initialised on l.544 whenever `&&` is false. So maybe initialising to 0 the values on l.530 and 531 would fix that.

skmcgrail:
```
  In file included from D:\a\aws-lc\aws-lc\crypto\fipsmodule\bcm.c:116:
D:\a\aws-lc\aws-lc\crypto\fipsmodule/hmac/hmac.c(533,18): error : variable 'o_ctx_n' is used uninitialized whenever '&&' condition is false [-Werror,-Wsometimes-uninitialized] [D:\a\aws-lc\aws-lc\build\crypto\fipsmodule\fipsmodule.vcxproj]
D:\a\aws-lc\aws-lc\crypto\fipsmodule/hmac/hmac.c(544,28): message : uninitialized use occurs here [D:\a\aws-lc\aws-lc\build\crypto\fipsmodule\fipsmodule.vcxproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.20348.0\ucrt\assert.h(40,17): message : expanded from macro 'assert' [D:\a\aws-lc\aws-lc\build\crypto\fipsmodule\fipsmodule.vcxproj]
D:\a\aws-lc\aws-lc\crypto\fipsmodule/hmac/hmac.c(533,18): message : remove the '&&' if its condition is always true [D:\a\aws-lc\aws-lc\build\crypto\fipsmodule\fipsmodule.vcxproj]
D:\a\aws-lc\aws-lc\crypto\fipsmodule/hmac/hmac.c(531,19): message : initialize the variable 'o_ctx_n' to silence this warning [D:\a\aws-lc\aws-lc\build\crypto\fipsmodule\fipsmodule.vcxproj]
```

Is the error message being seen in the clang-clmsbuild for x64 and x64_arm64 in Windows

crypto/fipsmodule/hmac/hmac.c
fabrice102: I've initialized `o_ctx_n` and left a comment about why it's done.
(I think build should pass even if `i_ctx_n` but I'm waiting for the CI/CD to go through to be certain.)

skmcgrail (approved):

nebeid (approved):

PR #1581, opened 2024-05-08
More minor symbols for Ruby support

samuel40791765:
### Issues:
Addresses `CryptoAlg-1717`

### Description of changes: 
This implements two minor symbols and a few no-op flags for Ruby support.

New symbols:
* PKCS12_new
* CONF_get1_default_config_file

No-op flags:
* SSL_OP_CRYPTOPRO_TLSEXT_BUG
* SSL_OP_SAFARI_ECDHE_ECDSA_BUG
* SSL_OP_TLSEXT_PADDING

### Call-outs:
All of these no-ops are following a precedent in AWS-LC. We may have to have a discussion on whether to support CONF modules further down the line, but exposing this as a no-op for now.

### Testing:
N/A

### DIFF
--- a/crypto/conf/conf.c
+++ b/crypto/conf/conf.c
@@ -642,6 +642,18 @@ int CONF_modules_load_file(const char *filename, const char *appname,
   return 1;
 }
 
+char *CONF_get1_default_config_file(void) {
+  const char *temp = "No support for Config files in AWS-LC.";
+  size_t temp_len = strlen(temp);
+
+  char *ret = (char *)OPENSSL_malloc(temp_len);
+  if(ret == NULL) {
+    OPENSSL_PUT_ERROR(CONF, ERR_R_MALLOC_FAILURE);
+  }
+  OPENSSL_memcpy(ret, temp, temp_len);
+  return ret;
+}
+
 void CONF_modules_free(void) {}
 
 void CONF_modules_unload(int all) {}
--- a/crypto/conf/conf.c
+++ b/crypto/conf/conf.c
@@ -649,6 +649,7 @@ char *CONF_get1_default_config_file(void) {
   char *ret = (char *)OPENSSL_malloc(temp_len);
   if(ret == NULL) {
     OPENSSL_PUT_ERROR(CONF, ERR_R_MALLOC_FAILURE);
+    return NULL;
   }
   OPENSSL_memcpy(ret, temp, temp_len);
   return ret;
--- a/crypto/conf/conf.c
+++ b/crypto/conf/conf.c
@@ -643,15 +643,7 @@ int CONF_modules_load_file(const char *filename, const char *appname,
 }
 
 char *CONF_get1_default_config_file(void) {
-  const char *temp = "No support for Config files in AWS-LC.";
-  size_t temp_len = strlen(temp) + 1;
-
-  char *ret = (char *)OPENSSL_malloc(temp_len);
-  if(ret == NULL) {
-    return NULL;
-  }
-  OPENSSL_memcpy(ret, temp, temp_len);
-  return ret;
+  return OPENSSL_strdup("No support for Config files in AWS-LC.");
 }
 
 void CONF_modules_free(void) {}
--- a/crypto/conf/conf.c
+++ b/crypto/conf/conf.c
@@ -644,7 +644,7 @@ int CONF_modules_load_file(const char *filename, const char *appname,
 
 char *CONF_get1_default_config_file(void) {
   const char *temp = "No support for Config files in AWS-LC.";
-  size_t temp_len = strlen(temp);
+  size_t temp_len = strlen(temp) + 1;
 
   char *ret = (char *)OPENSSL_malloc(temp_len);
   if(ret == NULL) {
--- a/crypto/conf/conf.c
+++ b/crypto/conf/conf.c
@@ -648,7 +648,6 @@ char *CONF_get1_default_config_file(void) {
 
   char *ret = (char *)OPENSSL_malloc(temp_len);
   if(ret == NULL) {
-    OPENSSL_PUT_ERROR(CONF, ERR_R_MALLOC_FAILURE);
     return NULL;
   }
   OPENSSL_memcpy(ret, temp, temp_len);
--- a/crypto/pkcs8/pkcs8_x509.c
+++ b/crypto/pkcs8/pkcs8_x509.c
@@ -741,7 +741,7 @@ struct pkcs12_st {
 
 PKCS12 *d2i_PKCS12(PKCS12 **out_p12, const uint8_t **ber_bytes,
                    size_t ber_len) {
-  PKCS12 *p12 = OPENSSL_malloc(sizeof(PKCS12));
+  PKCS12 *p12 = PKCS12_new();
   if (!p12) {
     return NULL;
   }
@@ -1328,7 +1328,7 @@ PKCS12 *PKCS12_create(const char *password, const char *name,
     goto err;
   }
 
-  ret = OPENSSL_malloc(sizeof(PKCS12));
+  ret = PKCS12_new();
   if (ret == NULL ||
       !CBB_finish(&cbb, &ret->ber_bytes, &ret->ber_len)) {
     OPENSSL_free(ret);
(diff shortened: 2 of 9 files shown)


crypto/conf/conf.c
```diff
@@ -642,6 +642,18 @@ int CONF_modules_load_file(const char *filename, const char *appname,
   return 1;
 }
 
+char *CONF_get1_default_config_file(void) {
+  const char *temp = "No support for Config files in AWS-LC.";
+  size_t temp_len = strlen(temp);
+
+  char *ret = (char *)OPENSSL_malloc(temp_len);
+  if(ret == NULL) {
+    OPENSSL_PUT_ERROR(CONF, ERR_R_MALLOC_FAILURE);
```
WillChilds-Klein, line 651: suggest returning early here to guard against segfault on L653

crypto/pkcs8/pkcs8_x509.c
```diff
@@ -1342,6 +1342,10 @@ PKCS12 *PKCS12_create(const char *password, const char *name,
   return ret;
 }
 
+PKCS12 *PKCS12_new(void) {
+  return OPENSSL_zalloc(sizeof(PKCS12));
```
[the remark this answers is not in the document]
samuel40791765: `OpenSSL_zalloc` isn't strictly necessary here, but I thought it'd be nice to get in the habit of zeroizing memory whenever we allocate. There's generally a negligible performance impact in our experience of zeroization, but I don't really have a strong preference here.

`OPENSSL_free` is a no-op when encountering a nullptr. I don't think we should allocate `ber_bytes` in `PKCS12_new` yet. `ber_bytes` is not allocatable until the data to be written is known and `PKCS12_new` is merely creating the `PKCS12` structure.
For coverage, `PKCS12_free` is used [as a deleter](https://github.com/aws/aws-lc/blob/main/include/openssl/pkcs8.h#L246) extensively in the [PKCS12 tests](https://github.com/aws/aws-lc/blob/8ae155b195d101f02a712e9b4bb41bba7affb0da/crypto/pkcs8/pkcs12_test.cc#L86).

crypto/conf/conf.c
samuel40791765: Ah my bad, forgot about that...

crypto/pkcs8/pkcs8_x509.c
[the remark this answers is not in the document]
samuel40791765: I switched back to `OPENSSL_malloc` to follow what some of the [original PKCS code](https://github.com/aws/aws-lc/blob/8ae155b195d101f02a712e9b4bb41bba7affb0da/crypto/pkcs8/pkcs8.c#L459) was doing. Also added a small test for `PKCS12_new` :)

crypto/conf/conf.c
andrewhop: You don't need to put the malloc failure message, [OPENSSL_malloc](https://github.com/aws/aws-lc/blob/main/crypto/mem.c#L201-L202) does that for you now.

samuel40791765: Ah sure, too much hygiene on my end

crypto/conf/conf.c
```diff
@@ -642,6 +642,18 @@ int CONF_modules_load_file(const char *filename, const char *appname,
   return 1;
 }
 
+char *CONF_get1_default_config_file(void) {
+  const char *temp = "No support for Config files in AWS-LC.";
+  size_t temp_len = strlen(temp);
```
WillChilds-Klein, line 647: do we need `+1` for null terminator?

samuel40791765: Good catch, just added.

crypto/conf/conf.c
```diff
@@ -642,6 +642,18 @@ int CONF_modules_load_file(const char *filename, const char *appname,
   return 1;
 }
 
+char *CONF_get1_default_config_file(void) {
+  const char *temp = "No support for Config files in AWS-LC.";
+  size_t temp_len = strlen(temp) + 1;
+
+  char *ret = (char *)OPENSSL_malloc(temp_len);
```
WillChilds-Klein, line 649: note for other reviewers: OpenSSL [also](https://github.com/openssl/openssl/blob/0285160ffa3b8c2b5491222243042593808298c4/crypto/conf/conf_mod.c#L698) alloc's in this method and leaves `free`ing up to the caller.

WillChilds-Klein (dismissed):

crypto/conf/conf.c
andrewhop, line 649: When you create a compile time constant string like this is it guaranteed that temp is null terminated so strlen will work as expected? Also when you malloc temp_len there might not be a 0 at the end so after the memcpy there could be garbage at the end too?

docs/porting/configuration-differences.md
```diff
@@ -188,6 +188,21 @@ The following table contains the differences in libssl configuration options AWS
   <p><span>NO-OP</span></p>
   </td>
  </tr>
+ <tr>
+  <td>
+  <p><span>
+    <a href="">
```
andrewhop, line 194: Missing link to docs.

include/openssl/conf.h
```diff
@@ -139,6 +139,10 @@ OPENSSL_EXPORT const char *NCONF_get_string(const CONF *conf,
 OPENSSL_EXPORT OPENSSL_DEPRECATED int CONF_modules_load_file(
     const char *filename, const char *appname, unsigned long flags);
 
+// CONF_get1_default_config_file returns a fixed dummy string. AWS-LC is defined
+// to have no config file options.
```
andrewhop, line 143: Is returning null safer? Would customer code now accidentally attempt to open/read/write to a invalid file path and crash?

include/openssl/ssl.h
```diff
@@ -5650,6 +5658,18 @@ OPENSSL_EXPORT int SSL_set1_curves_list(SSL *ssl, const char *curves);
 // AWS-LC
 #define SSL_OP_NO_SSLv3 0
 
+// SSL_OP_SAFARI_ECDHE_ECDSA_BUG is OFF by default in AWS-LC. Turning this ON in
+// OpenSSL lets the application not prefer ECDHE-ECDSA ciphers when the client
+// appears to be Safari on OSX.
+//
+// Note: OS X 10.8..10.8.3 broke support for ECDHE-ECDSA ciphers.
```
andrewhop, line 5665: Missing line about this is not supported in AWS-LC like the other 0 defines.

include/openssl/ssl.h
```diff
@@ -5650,6 +5658,18 @@ OPENSSL_EXPORT int SSL_set1_curves_list(SSL *ssl, const char *curves);
...
+#define SSL_OP_SAFARI_ECDHE_ECDSA_BUG 0
+
+// SSL_OP_TLSEXT_PADDING is OFF by default in AWS-LC. Turning this ON in OpenSSL
+// adds a padding extension to ensure the ClientHello size is never between 256
+// and 511 bytes in length. This is needed as a workaround for F5 terminators.
```
andrewhop, line 5670: Same.

crypto/conf/conf.c
samuel40791765: For the first question, yes.
* https://en.cppreference.com/w/c/language/string_literal
* https://stackoverflow.com/questions/47288881/are-c-constant-character-strings-always-null-terminated

For the second question, I'm not entirely sure what you mean by 0 coming into play. The additional `+ 1` is to account for the null terminator. This follows a pattern [in OpenSSL](https://github.com/openssl/openssl/blob/0285160ffa3b8c2b5491222243042593808298c4/crypto/conf/conf_mod.c#L697) and [our codebase](https://github.com/aws/aws-lc/blob/94e91d96d311404a0eaaca6039bfd7e6577c6dff/crypto/mem.c#L382) as well.

Good news I discovered `OPENSSL_strdup` while fiddling with this. I'll call this instead of all the duplicate hocus pocus here.

include/openssl/conf.h
WillChilds-Klein: hm, unguarded `strlen(path)` is unfortunate. that would segfault if `path` is `NULL`.

samuel40791765: Yeah I forgot to mention that this was the initial forcing function to return a string. Populating a value for the user to read into seemed like the better call than the potential segfault.

andrewhop (approved):

WillChilds-Klein (approved):

lilpoozie2005 (approved):

PR #1582, opened 2024-05-08
Revert "Trim some unused XN_FLAG_* values"

samuel40791765:
This reverts commit b6082af174881079096aad9661e505be6ba0c5e7.

### Description of changes: 
Ruby [is dependent](https://github.com/ruby/ruby/blob/ruby_3_1/ext/openssl/ossl_x509name.c#L596) on `XN_FLAG_MULTILINE`. We removed the flag in b6082af174881079096aad9661e505be6ba0c5e7 during the upstream merge, this reverts the change.

### Call-outs:
N/A

### Testing:
Original tests

### DIFF
--- a/include/openssl/x509.h
+++ b/include/openssl/x509.h
@@ -2281,12 +2281,25 @@ OPENSSL_EXPORT int X509_REQ_print_fp(FILE *fp, X509_REQ *req);
 // XN_FLAG_FN_SN uses the attribute type's short name, when available.
 #define XN_FLAG_FN_SN 0
 
+// XN_FLAG_FN_LN uses the attribute type's long name, when available.
+#define XN_FLAG_FN_LN (1 << 21)
+
+// XN_FLAG_FN_OID always prints attribute types as OIDs.
+#define XN_FLAG_FN_OID (2 << 21)
+
+// XN_FLAG_FN_NONE skips printing field names.
+#define XN_FLAG_FN_NONE (3 << 21)
+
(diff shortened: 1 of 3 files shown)


include/openssl/x509.h
```diff
@@ -2281,13 +2281,26 @@ OPENSSL_EXPORT int X509_REQ_print_fp(FILE *fp, X509_REQ *req);
...
 #define XN_FLAG_FN_SN 0
 
+// XN_FLAG_FN_LN uses the attribute type's long name, when available.
+#define XN_FLAG_FN_LN (1 << 21)
+
+// XN_FLAG_FN_OID always prints attribute types as OIDs.
+#define XN_FLAG_FN_OID (2 << 21)
+
+// XN_FLAG_FN_NONE skips printing field names.
+#define XN_FLAG_FN_NONE (3 << 21)
```
WillChilds-Klein, line 2291: looks like these [aren't used in ruby](https://github.com/search?q=repo%3Aruby%2Fruby+XN_FLAG_&type=code) or for defining `XN_FLAG_MULTILINE`. can we add another commit to this PR removing them? perhaps with a note explaining the gap in bit shift values.

samuel40791765: Sure I have no objection to that.

justinwsmith (approved):

include/openssl/x509.h
WillChilds-Klein: Agreed that it's not necessary. Probably doesn't provide material benefit either.

PR #1584, opened 2024-05-09
OpenVPN error codes, SSL_get_peer_signature_* funcs, and first patch file

smittals2:
### Issues
CryptoAlg-2477
CryptoAlg-2478

### Description of changes: 
1. Added three new aliases for AES-128-GCM, AES-192-GCM, and AES-256-GCM as required by OpenVPN 
2. Added an error and function code. Error code points to our equivalent of the OpenSSL PKCS12_R_MAC_VERIFY_FAILURE error
3. Added SSL_get_peer_signature_nid and SSL_get_peer_signature_type_nid. No new state is stored for these in the SSL connection.
4. Generated a patch file to use OpenVPNs internal ASSERT function instead of OpenSSL_assert, SSL_get_ciphers instead of SSL_get1_supported_ciphers, and point SSL_get_peer_tmp_key to SSL_get_server_tmp_key (which we define as a no-op)
5. Added pkcs8.h to doc.config

### Call-outs:
The patch file is temporary and will likely change with future PRs

### Testing:
1. SSL_get_peer_signature_nid and SSL_get_peer_signature_type_nid are tested in a TLS 1.2 connection (with renegotiation) and in a TLS 1.3 connection (with mTLS configured to test both the client and server side). 


ssl/ssl_lib.cc
```diff
@@ -3047,6 +3047,37 @@ uint16_t SSL_get_peer_signature_algorithm(const SSL *ssl) {
...
 }
 
+int SSL_get_peer_signature_nid(const SSL *ssl, int *psig_nid) {
+  uint16_t sig_alg = SSL_get_peer_signature_algorithm(ssl);
+  if (sig_alg == 0) {
+    return 0;
+  }
+
+  const EVP_MD *digest_type = SSL_get_signature_algorithm_digest(sig_alg);
+  if (digest_type == nullptr) {
```
geedo0, line 3057: nit: The rest of this file uses `NULL`

ssl/ssl_lib.cc
```diff
@@ -3047,6 +3047,37 @@ uint16_t SSL_get_peer_signature_algorithm(const SSL *ssl) {
...
+    return 0;
+  }
+
+  *psig_nid = EVP_MD_nid(digest_type);
```
geedo0, line 3061: Is there an error case to check for like `EVP_PKEY_NONE` below?

geedo0 (commented):
LGTM, just some style nits to fix.

ssl/ssl_lib.cc
smittals2: Error case is checked when SSL_get_signature_algorithm_digest is called. So there should always be a valid mapping for any given EVP_MD object (EVP_MD_nid calls EVP_md_type which just returns md->type).

geedo0 (dismissed):

skmcgrail (changes requested):

include/openssl/ssl.h
```diff
@@ -4898,6 +4898,16 @@ OPENSSL_EXPORT int SSL_get_shutdown(const SSL *ssl);
 // peer. If not applicable, it returns zero.
 OPENSSL_EXPORT uint16_t SSL_get_peer_signature_algorithm(const SSL *ssl);
 
+// SSL_get_peer_signature_nid sets |psig_nid| to the NID of the digest used by
+// the peer to sign TLS messages. Returns one on success and zero on failure.
```
andrewhop, line 4902:
```suggestion
// the peer to sign their TLS messages. Returns one on success and zero on
// failure.
```

include/openssl/ssl.h
```diff
@@ -4898,6 +4898,16 @@ OPENSSL_EXPORT int SSL_get_shutdown(const SSL *ssl);
...
+OPENSSL_EXPORT int SSL_get_peer_signature_nid(const SSL *ssl, int *psig_nid);
+
+// SSL_get_peer_signature_type_nid sets |psigtype_nid| to the signature type
+// used by the peer to sign TLS messages. The signature type is the NID of the
```
andrewhop, line 4906:
```suggestion
// used by the peer to sign their TLS messages. The signature type is the NID of the
```

ssl/ssl_lib.cc
```diff
@@ -3047,6 +3047,45 @@ uint16_t SSL_get_peer_signature_algorithm(const SSL *ssl) {
   return session->peer_signature_algorithm;
 }
 
+int SSL_get_peer_signature_nid(const SSL *ssl, int *psig_nid) {
+  if(psig_nid == NULL) {
+    return 0;
+  }
```
andrewhop, line 3053: This is still a style preference but for future reference we have:
```suggestion
  GUARD_PTR(psig_nid);
```

skmcgrail (dismissed):

geedo0 (approved):

skmcgrail (approved):

lilpoozie2005 (approved):

PR #1587, opened 2024-05-10
Add integration tests for OpenSSL-linking 3p modules

WillChilds-Klein:
### Description of changes: 

This change expands our python integration tests to load and exercise libcrypto-linking 3rd party modules. Cases covered include statically linked AWS-LC (default AWS CRT) and dynamically linked OpenSSL (AWS CRT, PyOpenSSL, PyCA). We also test these integrations with python's AWS-LC built in FIPS mode.

### Call-outs:

Due to the pre-release nature of python 3.13 and 3.14, there are a few caveats for later versions: 3.12 introduced [a change](https://github.com/python/cpython/issues/95299) to drop `setuptools` from default virtual environments, causing installation issues for PyCA and PyOpenSSL on 3.13+. To work around this, we allow installation failure for those dependencies on newer (currently both 3.13 and 3.14 are pre-release) and exit early from the relevant test cases. We also encountered some issues installing CRT bindings from source on 3.12+, so for those versions we install the pre-compiled wheel module from PyPI, which uses AWS-LC under the hood.

### Testing:
- CI

### DIFF
--- a/tests/ci/integration/python_tests/test_cryptography.py
+++ b/tests/ci/integration/python_tests/test_cryptography.py
@@ -18,4 +18,4 @@ assert pt == f.decrypt(f.encrypt(pt))
 
 version = cryptography.hazmat.backends.openssl.backend.openssl_version_text()
 assert "OpenSSL" in version, f"PyCA didn't link OpenSSL: {version}"
-assert "AWS-LC" not in version, f"PyOpenSSL linked AWS-LC: {version}"
+assert "AWS-LC" not in version, f"PyCA linked AWS-LC: {version}"
--- a/tests/ci/integration/python_tests/test_cryptography.py
+++ b/tests/ci/integration/python_tests/test_cryptography.py
@@ -1,7 +1,7 @@
 import sys
 assert sys.version_info.major == 3, 'Only python 3 supported'
-if sys.version_info.minor == 14:
-    print("Fernet import currently broken on mainline py release canddiate")
+if sys.version_info.minor >= 13:
+    print("Fernet import currently broken on python release candidates >= 3.13")
     print("Returning early for now, need to check in on this post-release")
     sys.exit()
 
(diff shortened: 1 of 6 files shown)


justsmth (dismissed):

tests/ci/integration/run_python_integration.sh
```diff
@@ -91,8 +176,6 @@ function python_run_tests() {
 #     authentication portion of that protocol.
 #   - Modify the ssl module's backing C code to account for AWS-LC's divergent
 #     function signature and return value for |sk_SSL_CIPHER_find|
-#
-# TODO: Remove these patches when we make an upstream contribution.
```
samuel40791765, line 95: Nit: Are all of the call outs above still applicable? I was under the impression that we finished tweaking some of the places in AWS-LC for better adoption.

tests/ci/integration/python_tests/test_cryptography.py
```diff
@@ -0,0 +1,21 @@
...
+
+# exercise simple round trip, then assert that PyCA has linked OpenSSL
+k = Fernet.generate_key()
+f = Fernet(k)
+pt = b"hello world"
+assert pt == f.decrypt(f.encrypt(pt))
+
+version = cryptography.hazmat.backends.openssl.backend.openssl_version_text()
+assert "OpenSSL" in version, f"PyCA didn't link OpenSSL: {version}"
+assert "AWS-LC" not in version, f"PyOpenSSL linked AWS-LC: {version}"
```
samuel40791765, line 21: Not sure if it should be: 
```suggestion
assert "AWS-LC" not in version, f"PyCA linked AWS-LC: {version}"
```

tests/ci/integration/run_python_integration.sh
WillChilds-Klein: They're still applicable for older versions of CPython, namely 3.10. Unfortunately 3.10 is now [only taking security patches](https://github.com/python/cpython/pull/117229#issuecomment-2041163376).

samuel40791765: ahh understood

samuel40791765 (approved):

justsmth (approved):

PR #1588, opened 2024-05-10
Fix NTP integration test in the case of multiple occurrences of the package

nebeid:
Fix the NTP integration test, a continuation of the fix in https://github.com/aws/aws-lc/pull/1548.

When there were multiple `.tar.gz` packages, the test was trying `wget` with
```
wget -q https://downloads.nwtime.org/ntp//ntp-4.2.8p17.tar.gz ntp-4.2.8p18-RC1.tar.gz
```
and was failing.


nebeid:
Superseded by https://github.com/aws/aws-lc/pull/1589

PR #1591, opened 2024-05-13
add back ASN1_dup with tests

samuel40791765:
### Issues:
Addresses `CryptoAlg-1717`

### Description of changes: 
`ASN1_dup` was removed in https://github.com/aws/aws-lc/commit/419144adce049b5341bd94d355c52d099eac56e3 in favor of `ASN1_Item_dup`. This shouldn't normally be called directly, but Ruby happens to consume the API[ in several instances](https://github.com/search?q=repo%3Aruby%2Fruby%20ASN1_dup&type=code).
I've optimized the function to allocate memory with `i2d` directly, instead of the [ancient OpenSSL allocate](https://github.com/aws/aws-lc/commit/419144adce049b5341bd94d355c52d099eac56e3#diff-46888fde23260e66fc474ee241cefde06f5af3ec79856c2c45a491103e91739eL72-L78) and pass into behavior.
Also added some tests for verification along with a do-not-use warning.

### Call-outs:
N/A

### Testing:
New tests

### DIFF
--- a/crypto/asn1/a_dup.c
+++ b/crypto/asn1/a_dup.c
@@ -59,6 +59,26 @@
 #include <openssl/err.h>
 #include <openssl/mem.h>
 
+void *ASN1_dup(i2d_of_void *i2d, d2i_of_void *d2i, void *input) {
+  if (i2d == NULL || d2i == NULL || input == NULL) {
+    OPENSSL_PUT_ERROR(ASN1, ERR_R_PASSED_NULL_PARAMETER);
+    return NULL;
+  }
+
+  // Size and allocate |buf|.
+  unsigned char *buf = NULL;
+  int buf_len = i2d(input, &buf);
+  if (buf == NULL) {
+    return NULL;
+  }
+
+  // |buf| needs to be converted to |const| to be passed in.
+  const unsigned char *temp_input = buf;
+  char *ret = d2i(NULL, &temp_input, buf_len);
+  OPENSSL_free(buf);
+  return ret;
+}
+
 // ASN1_ITEM version of dup: this follows the model above except we don't
 // need to allocate the buffer. At some point this could be rewritten to
 // directly dup the underlying structure instead of doing and encode and
--- a/crypto/asn1/a_dup.c
+++ b/crypto/asn1/a_dup.c
@@ -68,7 +68,7 @@ void *ASN1_dup(i2d_of_void *i2d, d2i_of_void *d2i, void *input) {
   // Size and allocate |buf|.
   unsigned char *buf = NULL;
   int buf_len = i2d(input, &buf);
-  if (buf == NULL) {
+  if (buf == NULL || buf_len < 0) {
     return NULL;
   }
 
--- a/crypto/asn1/asn1_test.cc
+++ b/crypto/asn1/asn1_test.cc
@@ -2412,6 +2412,51 @@ TEST(ASN1Test, LargeString) {
 #endif
 }
 
+TEST(ASN1Test, ASN1Dup) {
+  const uint8_t *tag = kTag128;
+  bssl::UniquePtr<ASN1_TYPE> asn1(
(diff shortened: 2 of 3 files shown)


crypto/asn1/a_dup.c
```diff
@@ -59,6 +59,26 @@
...
+void *ASN1_dup(i2d_of_void *i2d, d2i_of_void *d2i, void *input) {
+  if (i2d == NULL || d2i == NULL || input == NULL) {
+    OPENSSL_PUT_ERROR(ASN1, ERR_R_PASSED_NULL_PARAMETER);
+    return NULL;
+  }
+
+  // Size and allocate |buf|.
+  unsigned char *buf = NULL;
+  int buf_len = i2d(input, &buf);
+  if (buf == NULL) {
```
skmcgrail, line 71: Does this indicate a malloc failure then? Should put an ERR_* in the queue?

crypto/asn1/asn1_test.cc
```diff
@@ -2412,6 +2412,70 @@ TEST(ASN1Test, LargeString) {
 #endif
 }
 
+
+// Wrapper functions are needed to get around Control Flow Integrity Sanitizers.
```
justsmth, line 2416: Would implementing/using the `D2I_OF` and `I2D_OF_const` macros make these functions unnecessary. I tried this change locally, and the CFI build passed.

justsmth: https://github.com/aws/aws-lc/commit/74a1acd3666c007e6ae15a6090e076bf05fe3fd0

samuel40791765: I took a closer look and noticed that the definition of `CHECKED_I2D_OF` is misaligned from [OpenSSL's ](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/include/openssl/asn1.h#L264-L265) (which doesn't take a `const`).
I tried changing ours to align, but it's not applicable.. This will cause issues for consumers looking to depend on the same version from OpenSSL, so I've changed ours to align with theirs. The macro's still very handy, so I've renamed and moved it to the `asn1_test` file.

skmcgrail (approved):

justsmth (approved):

samuel40791765:
Apparently the CFI is still failing with the macros:
```
/codebuild/output/src1171782177/src/github.com/aws/aws-lc/crypto/asn1/a_dup.c:70:17: runtime error: control flow integrity check for type 'int (const void *, unsigned char **)' failed during indirect function call
(/codebuild/output/src1171782177/src/github.com/aws/aws-lc/test_build_dir/crypto/crypto_test+0x3afd5e0): note: i2d_ASN1_TYPE defined here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /codebuild/output/src1171782177/src/github.com/aws/aws-lc/crypto/asn1/a_dup.c:70:17 in 
```

I reverted to run [4a5be1e](https://github.com/aws/aws-lc/pull/1591/commits/4a5be1e45173d1615c2d5ba19d0b228d16a22af3) against the CFI sanitizers, but apparently it doesn't actually pass the CFI tests either. I'll need to revert [4a5be1e](https://github.com/aws/aws-lc/pull/1591/commits/4a5be1e45173d1615c2d5ba19d0b228d16a22af3) and [0cd2716](https://github.com/aws/aws-lc/pull/1591/commits/0cd271638a7d9dad46a7977f94ee142252bfac66) to remove the macros and get the CFI build passing.

I think the main issue is that `i2d_of_void` and `d2i_of_void` expect a `void` parameter, but the typical `i2d`/`d2i` functions have a specific type assigned as the parameter. Upon closer reexamination, the macros don't seem to solve this issue. The two CI failures in the commits may show more details.

justsmth (approved):

skmcgrail (approved):

PR #1597, opened 2024-05-15
Only emit deprecation warnings for no-ops in debug

samuel40791765:
### Description of changes: 
More consumers than anticipated define `-Werror` and the no-op symbols we've marked as deprecated are breaking builds that define so. The original concept was great for finding integrations that consume no-op symbols and debugging them, so we'd like to keep that.
This change only marks no-op symbols as deprecated in debug builds. This unblocks our release build to work with projects that define `-Werror` and lets us keep the behavior that was great for debugging.

### Call-outs:
I didn't go the route of using `NDEBUG` since it's not always defined for downstream release builds. If it's not defined in the environment, the deprecation warnings will be emitted for no-op builds and break `-Werror` consuming applications that don't have `NDEBUG` defined.

### Testing:
CI

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -730,6 +730,10 @@ if(CONSTANT_TIME_VALIDATION)
   add_definitions(-DNDEBUG)
 endif()
 
+if(CMAKE_BUILD_TYPE_LOWER STREQUAL "debug")
+  add_definitions(-DAWSLC_DEBUG_BUILD)
+endif()
+
 # CMake's iOS support uses Apple's multiple-architecture toolchain. It takes an
 # architecture list from CMAKE_OSX_ARCHITECTURES, leaves CMAKE_SYSTEM_PROCESSOR
 # alone, and expects all architecture-specific logic to be conditioned within
--- a/include/openssl/base.h
+++ b/include/openssl/base.h
@@ -182,6 +182,22 @@ extern "C" {
 
 #endif
 
+#if defined(AWSLC_DEBUG_BUILD)
+
+// AWSLC_NOOP is used to mark a no-op symbol in AWS-LC as deprecated
+// in debug builds. This allows potential reliance on these no-ops symbols
+// to be more discoverable when migrating to AWS-LC, but does not block
+// release builds that may define "-Werror".
+// Refer to "docs/porting/functionality-differences.md" for more details
+// on no-op symbols in AWS-LC.
+#define AWSLC_NOOP __attribute__((__deprecated__))
+
+#else
+
+#define AWSLC_NOOP
+
+#endif
+
 
 #if defined(__GNUC__) || defined(__clang__)
 // MinGW has two different printf implementations. Ensure the format macro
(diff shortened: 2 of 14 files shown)


include/openssl/base.h
```diff
@@ -182,6 +182,22 @@ extern "C" {
...
 
+#if defined(AWSLC_DEBUG_BUILD)
+
+// AWSLC_NOOP is used to mark a no-op symbol in AWS-LC as deprecated
+// in debug builds. This allows potential reliance on these no-ops symbols
+// to be more discoverable when migrating to AWS-LC, but does not block
+// release builds that may define "-Werror".
+// Refer to "docs/porting/functionality-differences.md" for more details
+// on no-op symbols in AWS-LC.
+#define AWSLC_NOOP __attribute__((__deprecated__))
```
justsmth, line 193: You could use `OPENSSL_DEPRECATED` here and leverage the logic above this which determines the appropriate way to indicate that for the current environment.

include/openssl/base.h
```diff
@@ -182,6 +182,22 @@ extern "C" {
 
 #endif
 
+#if defined(AWSLC_DEBUG_BUILD)
```
justsmth, line 185: Were we to automatically detect debug/release builds (see other comment), I think we should still have an explicit `AWSLC_DEBUG_BUILD` macro for those that would like it for their release builds.
```C++
#if defined(DEBUG) || defined(_DEBUG) || !defined(NDEBUG)
   #define AWSLC_DEBUG_BUILD 1
#endif

#if defined(AWSLC_DEBUG_BUILD)
...
```

include/openssl/base.h
```diff
@@ -182,6 +182,22 @@ extern "C" {
...
+// in debug builds. This allows potential reliance on these no-ops symbols
+// to be more discoverable when migrating to AWS-LC, but does not block
+// release builds that may define "-Werror".
+// Refer to "docs/porting/functionality-differences.md" for more details
+// on no-op symbols in AWS-LC.
+#define AWSLC_NOOP __attribute__((__deprecated__))
+
+#else
+
+#define AWSLC_NOOP
```
justsmth, line 197: Instead of naming this to narrowly apply to "no-op" functions, we could use a name that might be applied to other similar situations. Perhaps a name like `AWSLC_DEBUG_DEPRECATED`?
```C++
#if defined(AWSLC_DEBUG_BUILD)
#define AWSLC_DEBUG_DEPRECATED OPENSSL_DEPRECATED
#else
#define AWSLC_DEBUG_DEPRECATED
#endif
```

CMakeLists.txt
```diff
@@ -730,6 +730,10 @@ if(CONSTANT_TIME_VALIDATION)
   add_definitions(-DNDEBUG)
 endif()
 
+if(CMAKE_BUILD_TYPE_LOWER STREQUAL "debug")
```
justsmth, line 733: (See other comment.) 

There are certain common macros that are specific to debug/release builds that we could use to automatically detect a debug build.
```C++
#if defined(DEBUG) || defined(_DEBUG) || !defined(NDEBUG)
...
```

PR #1607, opened 2024-05-24
NIST.SP.800-56Cr2 One-Step Key Derivation

skmcgrail:
### Description of changes: 
This pull request implements the One-Step Key Derivation function defined in [Section 4 of NIST.SP.800-56Cr2](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Cr2.pdf#%5B%7B%22num%22%3A52%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C70%2C720%2C0%5D)

Here we implement two of the three variants: option 1 (hash based) and option 2 (hmac based).

The abbreviation `SSKDF` is used in the public functions due to the extensive usage of the term within OpenSSL when referring to this algorithm. `SSKDF == "Single-Step Key Derivation Function"`.

### Call-outs:
* We will need to add appropriate service indicator logic ahead of our next FIPS certification round.
* There is some non-trivial amount of work required to wire-up the KDA-OneStep ACVP test vectors into our current Go ACVP setup and modulewrapper. I started going down this path, but there is some refactoring that requires to decouple some of the JSON structure from the current KDA-HKDF expectations. I'd rather keep this PR to the implementation and some confidence test vectors I ported from various other implementation sources.

### Testing:
See the test vectors provided in [sskdf.txt](https://github.com/skmcgrail/aws-lc/blob/sskdf/crypto/fipsmodule/kdf/test/sskdf.txt) which I sourced from a number of locations as documented in that file. This currently provides coverage of the KDF functions for SHA-1, SHA-224, SHA-256, SHA-384, and SHA-512. SHA-3 algorithms will also get added coverage once we hook into the ACVP system.


crypto/fipsmodule/kdf/internal.h
```diff
@@ -0,0 +1,29 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include <openssl/digest.h>
+#include <openssl/hmac.h>
+
+typedef struct {
+  void *data;
```
dkostic, line 8: I guess this is `void *` because it can point to `SSKDF_VARIANT_DIGEST_CTX` or `SSKDF_VARIANT_HMAC_CTX`?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+  if (!md_ctx) {
+    goto err;
+  }
+
+  ret = 1;
+  variant_ctx->digest = digest;
+  variant_ctx->ctx = md_ctx;
+  ctx->data = variant_ctx;
+
+  return ret;
```
dkostic, line 38: nit: you can delete this return.

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+}
+
+
+
+static void SSKDF_VARIANT_DIGEST_CTX_cleanup(SSKDF_VARIANT_CTX *ctx) {
+  if (!ctx) {
+    return;
+  }
+  SSKDF_VARIANT_DIGEST_CTX *variant_ctx = (SSKDF_VARIANT_DIGEST_CTX *)ctx->data;
+  if (!ctx->data) {
```
dkostic, line 54: shouldn't this check come before the line above?

crypto/fipsmodule/kdf/internal.h
```diff
@@ -0,0 +1,29 @@
...
+} SSKDF_VARIANT_CTX;
+
+typedef struct {
+  const EVP_MD *digest;
+  EVP_MD_CTX *ctx;
```
dkostic, line 13: nit: `md_ctx` instead of `ctx` would help with readability of the code where this is being used.

crypto/fipsmodule/kdf/internal.h
```diff
@@ -0,0 +1,29 @@
...
+} SSKDF_VARIANT_DIGEST_CTX;
+
+typedef struct {
+  HMAC_CTX *ctx;
```
dkostic, line 17: nit: hmac_ctx instead of ctx would help with readability of the code where this is being used.

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+
+static int SSKDF_VARIANT_HMAC_CTX_init(SSKDF_VARIANT_CTX *ctx,
+                                       const EVP_MD *digest,
+                                       const uint8_t *salt, size_t salt_len) {
+  SSKDF_VARIANT_HMAC_CTX *variant_ctx = NULL;
+  HMAC_CTX *hmac_ctx = NULL;
+
+  int ret = 0;
+
+  if (!ctx || !digest) {
```
dkostic, line 70: `salt` can be NULL?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+
+  if (!HMAC_Init_ex(hmac_ctx, salt, salt_len, digest, NULL)) {
+    goto err;
+  }
+
+  ret = 1;
+  variant_ctx->ctx = hmac_ctx;
+  ctx->data = variant_ctx;
+
+  return ret;
```
dkostic, line 92: nit: you can delete this return.

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+  return EVP_MD_size(variant_ctx->digest);
+}
+
+static int sskdf_variant_digest_compute(SSKDF_VARIANT_CTX *ctx, uint8_t *out,
+                                        size_t out_len,
+                                        const uint8_t counter[4],
+                                        const uint8_t *secret,
+                                        size_t secret_len, const uint8_t *info,
+                                        size_t info_len) {
+  if (!ctx || !ctx->data || !out || !counter || !secret) {
```
dkostic, line 130: `info` can be NULL?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+  if (!ctx || !ctx->data) {
+    return 0;
+  }
+  SSKDF_VARIANT_DIGEST_CTX *variant_ctx = (SSKDF_VARIANT_DIGEST_CTX *)ctx->data;
+  return EVP_MD_size(variant_ctx->digest);
+}
+
+static int sskdf_variant_digest_compute(SSKDF_VARIANT_CTX *ctx, uint8_t *out,
+                                        size_t out_len,
+                                        const uint8_t counter[4],
```
dkostic, line 126: why 4?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+  if (!variant_ctx->ctx || !variant_ctx->digest) {
+    return 0;
+  }
+
+  if (!EVP_MD_CTX_reset(variant_ctx->ctx) ||
+      !EVP_DigestInit_ex(variant_ctx->ctx, variant_ctx->digest, NULL) ||
+      !EVP_DigestUpdate(variant_ctx->ctx, &counter[0], 4) ||
+      !EVP_DigestUpdate(variant_ctx->ctx, secret, secret_len) ||
+      !EVP_DigestUpdate(variant_ctx->ctx, info, info_len) ||
+      !EVP_DigestFinal(variant_ctx->ctx, out, NULL)) {
```
dkostic, line 145: should we check if `out_len` is large enough for `EVP_DigestFinal` to write its result to?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+    return 0;
+  }
+  return HMAC_size(variant_ctx->ctx);
+}
+
+static int sskdf_variant_hmac_compute(SSKDF_VARIANT_CTX *ctx, uint8_t *out,
+                                      size_t out_len, const uint8_t counter[4],
+                                      const uint8_t *secret, size_t secret_len,
+                                      const uint8_t *info, size_t info_len) {
+  if (!ctx || !ctx->data || !out || !counter || !secret) {
```
dkostic, line 167: info can be NULL?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+
+  if (!variant_ctx->ctx) {
+    return 0;
+  }
+
+  if (!HMAC_Init_ex(variant_ctx->ctx, NULL, 0, NULL, NULL) ||
+      !HMAC_Update(variant_ctx->ctx, &counter[0], 4) ||
+      !HMAC_Update(variant_ctx->ctx, secret, secret_len) ||
+      !HMAC_Update(variant_ctx->ctx, info, info_len) ||
+      !HMAC_Final(variant_ctx->ctx, out, NULL)) {
```
dkostic, line 181: should we check if `out_len` is large enough for `EVP_DigestFinal` to write its result to?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+  out->compute = sskdf_variant_hmac_compute;
+};
+
+static int SSKDF(const SSKDF_VARIANT *variant, SSKDF_VARIANT_CTX *ctx,
+                 uint8_t *out_key, size_t out_len, const uint8_t *secret,
+                 size_t secret_len, const uint8_t *info, size_t info_len) {
+  int ret = 0;
+
+  if (!ctx) {
+    abort();
```
dkostic, line 208: that escalated quickly :) why not just return 0?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  if (!out_key || out_len == 0 || !secret || secret_len == 0) {
+    goto err;
+  }
+
+  size_t h_len = variant->size(ctx);
```
dkostic, line 215: can you add a comment what does `h_len` represent? or use a more descriptive name?

dkostic, line 215: should we check if `variant != NULL` before dereferencing?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+  if (!h_len) {
```
dkostic, line 216: nit
```suggestion
  if (h_len == 0) {
```

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+    goto err;
+  }
+  assert(h_len <= EVP_MAX_MD_SIZE);
+
+  uint64_t n = (out_len + h_len - 1) / h_len;
+
+  if (out_len + h_len < out_len || n > UINT32_MAX) {
+    goto err;
+  }
```
dkostic, line 225: can you please add some documentation explaining what are we trying to do here?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+
+  // Validate that the sum of the fixed info lengths won't overflow a size_t
+  if (4 > SIZE_MAX - secret_len || secret_len > SIZE_MAX - 4 ||
```
dkostic, line 228: why do we need both conditions?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+
+  // Now cap the approved algorithms to not exceed the max_H_inputBits for
+  // hash(x) or HMAC-hash(salt, x). UINT32_MAX is a sufficient max value. See
+  // Section 4.2 Table 1 and 2
+  // https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Cr2.pdf
+  if (4 + secret_len + info_len > UINT32_MAX) {
+    goto err;
+  }
+
+  // TODO: Abstract buffer size, if we ever need to support KMAC this could be
```
dkostic, line 241:
```suggestion
  // TODO(awslc): Abstract buffer size, if we ever need to support KMAC this could be
```

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+  // variable. Currently sufficient for HMAC and digest variants
+  uint8_t previous[EVP_MAX_MD_SIZE];
+
+  size_t done = 0;
+
+  for (uint32_t i = 0; i < n; i++) {
+    uint8_t counter[4];  // counter == 4 bytes
```
dkostic, line 248: we should probably give a name to the 4 and add an explanation why it's 4.

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+    size_t todo;
+
+    CRYPTO_store_u32_be(&counter[0], i + 1);
```
dkostic, line 251: why big endian?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+static int SSKDF(const SSKDF_VARIANT *variant, SSKDF_VARIANT_CTX *ctx,
+                 uint8_t *out_key, size_t out_len, const uint8_t *secret,
+                 size_t secret_len, const uint8_t *info, size_t info_len) {
+  int ret = 0;
+
+  if (!ctx) {
+    abort();
+  }
+
+  if (!out_key || out_len == 0 || !secret || secret_len == 0) {
```
dkostic, line 211: `info` can be NULL?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+}
+
+
+DEFINE_METHOD_FUNCTION(SSKDF_VARIANT, SSKDF_VARIANT_HMAC) {
+  memset(out, 0, sizeof(SSKDF_VARIANT));
+  out->size = sskdf_variant_hmac_size;
+  out->compute = sskdf_variant_hmac_compute;
+};
+
+static int SSKDF(const SSKDF_VARIANT *variant, SSKDF_VARIANT_CTX *ctx,
```
dkostic, line 202: this function could use more comments documenting what's being done or referencing the SP.

crypto/fipsmodule/kdf/internal.h
```diff
@@ -0,0 +1,29 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include <openssl/digest.h>
+#include <openssl/hmac.h>
+
```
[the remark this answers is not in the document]
skmcgrail: Wasn't sure when we used capitalized struct names and when we didn't. Wasn't sure if that was reserved for structures that are typedef'd or if it was structures that were public only. No preference here though, I can lower case them.

crypto/fipsmodule/kdf/internal.h
skmcgrail: That is correct!

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: Good point this should either come first, or cast the pointer then check that the cast'ed pointer is not null. Likely check before cast so will move this up.

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+  if (!ctx) {
+    return;
+  }
+  SSKDF_VARIANT_HMAC_CTX *hmac_ctx = (SSKDF_VARIANT_HMAC_CTX *)ctx->data;
+  if (!hmac_ctx) {
+    return;
+  }
+  HMAC_CTX_free(hmac_ctx->ctx);
+  OPENSSL_free(hmac_ctx);
+  ctx->data = NULL;
```
darylmartin100, line 113: Nit:  I would consider just a memset of the structure.  Its effectively the same thing but easier to read.

crypto/fipsmodule/kdf/sskdf.c
darylmartin100: nit:  0 == h_len.  You will get a compilation error if you type 0 = h_len vs h_len = 0.

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,314 @@
...
+
+  size_t h_len = variant->size(ctx);
+  if (!h_len) {
+    goto err;
+  }
+  assert(h_len <= EVP_MAX_MD_SIZE);
+
+  uint64_t n = (out_len + h_len - 1) / h_len;
+
+  if (out_len + h_len < out_len || n > UINT32_MAX) {
```
darylmartin100, line 223: nit:  you already check above if h_len == 0 so its impossible for out_len + h_len < out_len.

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: That would then cause the structures to be unallcoated below, which is not the desired behavior.

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: Yes per the NIST specification salt can be NULL, and in that case the default key will be an all-zero byte string whose bit length matches the bit length of the input block of the hash function. In our case HMAC_Init_ex does exactly this behavior if the key provided is NULL.

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: The early return is required here, otherwise the initalized data structures would get free'd below, which would be unexpected.

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: Per the NIST specification one-step key derivation uses big-endian 4-byte usnigned integer counter. I chose to keep that contract explicit in these compute computation functions by being explicit about the length of the array.

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: This a programming implementation error on our part if this occurs. We are responsible for allocating the `SSKDF_VARIANT` structure within the implementation our public functions (inside `SSKDF_DIGEST` and `SSKDF_HMAC`). So the user can never pass it in themselves as a NULL value. So this is just a sanity check for ourselves. I can change it but a user would never really hit this case unless a series of failures occured in our coding of the implementation.

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: I was borrowijg an optimization decision we take advantage of in `HKDF_expand` which is that we've sized `out` to be exactly `EVP_MAX_MD_SIZE` so we know the output will be no more then that value. We could technically add an assert here, but we also already check this exact condition in the `SSHKDF` function that orchestrates the specific compute variant functions. Here size outputs the size that `EVP_DigestFinal` or `HMAC_Final` will output:
```
size_t h_len = variant->size(ctx);
  if (!h_len) {
    goto err;
  }
  assert(h_len <= EVP_MAX_MD_SIZE);
```

So I think our bases are covered here already and additional checking would be redundant.

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: NIST doesn't have any explicit statement that is must be non-NULL / non-zero length, nor does the algorithm explicitly require a length check on it:

> FixedInfo – A bit string of context-specific data that is appropriate for the relying key-establishment scheme

Additionally the OpenSSL implementation also supports a NULL info length which is why I opted to allow it for comparability with that implementation.

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: See the above comment where I addressed this for EVP_DigestFinal.

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: Defined so by the specification:
> Initialize a big-endian 4-byte unsigned integer counter as 0x00000000, corresponding to a 32-bit binary representation of the number zero.

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: See above comment about info being NULL to be compatible with the OpenSSL implementation, and not explicitly required by NIST to be non-NULL.

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,336 @@
...
+  // 2. If reps > (2^32 - 1), then output an error indicator and exit this
+  // process without performing the remaining actions
+  if (n > UINT32_MAX) {
+    goto err;
+  }
+
+  // Validate that the sum of the fixed info lengths won't overflow a size_t
+  // which will very based on 32-bit or 64-bit platforms. We will further clamp
+  // this value below.
+  if (info_len > SIZE_MAX - 4 - secret_len) {
```
dkostic, line 237: it's not clear to me why we have this and the next check? why not test only with `UINT32_MAX`?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,336 @@
...
+    goto err;
+  }
+
+  // NIST.SP.800-56Cr2: "4. If counter || Z || FixedInfo is more than
+  // max_H_inputBits bits long, then output an error indicator and exit this
+  // process without performing any of the remaining actions"
+  //
+  // UINT32_MAX is a sufficient max value to satisfy this requirement. See
+  // Section 4.2 Table 1 and 2
+  // https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Cr2.pdf
```
dkostic, line 247: can you say a bit more on how you determine this based on those tables?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,336 @@
...
+    goto err;
+  }
+
+  // TODO(awslc): Abstract buffer size, if we ever need to support KMAC this
+  // could be variable. Currently sufficient for HMAC and digest variants
+  uint8_t previous[EVP_MAX_MD_SIZE];
+
+  size_t done = 0;
+
+  // NIST.SP.800-56Cr2: "Initialize a big-endian 4-byte unsigned integer counter
```
dkostic, line 258: is this a leftover? as is, it doesn't refer to any code.

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,336 @@
...
+  // UINT32_MAX is a sufficient max value to satisfy this requirement. See
+  // Section 4.2 Table 1 and 2
+  // https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Cr2.pdf
+  if (4 + secret_len + info_len > UINT32_MAX) {
+    goto err;
+  }
+
+  // TODO(awslc): Abstract buffer size, if we ever need to support KMAC this
+  // could be variable. Currently sufficient for HMAC and digest variants
+  uint8_t previous[EVP_MAX_MD_SIZE];
```
dkostic, line 254: can we move this within the for loop for clarity? and maybe call it something else, `out_key_i` or similar?

include/openssl/kdf.h
```diff
@@ -25,12 +25,32 @@ extern "C" {
...
+                                   size_t secret_len, const char *label,
+                                   size_t label_len, const uint8_t *seed1,
+                                   size_t seed1_len, const uint8_t *seed2,
+                                   size_t seed2_len);
+
+// SSKDF_digest computes the One-step (Single-step) key derivation using the
+// provided digest algorithm as the backing PRF.
+//
+// Returns a 1 on success, otherwise returns 0.
+OPENSSL_EXPORT int SSKDF_digest(uint8_t *out_key, size_t out_len,
```
nebeid, line 39: It would be good to describe the arguments in relation to the standard.
I believe the standard allows _L_ to be an arbitrary length in bits, not necessarily divisible by 8. But I think we don't neither does OpenSSL. I recommend we state that.
Same for `SSKDF_hmac`.

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,336 @@
...
+  if (output_size == 0) {
+    goto err;
+  }
+  assert(output_size <= EVP_MAX_MD_SIZE);
+
+  // NIST.SP.800-56Cr2:
+  // 1. If L > 0, then set reps = ceil(L / H_outputBits); otherwise, output an
+  // error indicator and exit this process without performing the remaining
+  // actions
+  uint64_t n = (out_len + output_size - 1) / output_size;
```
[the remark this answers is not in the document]
skmcgrail: The digest algorithms we all support are multiples of eight, from my investigation and knowledge we don't support a EVP_MD that falls outside that. I don't think this is an issue, as similar FIPS specifications (like our existing KDF in feedback mode) are also specified by bits rather then bytes.

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: The original intention of this check (and the original variation which broke out the check per variable) is to ensure that summing the `4 + secret_len + info_len` doesn't overflow size_t. Just checking that the sum is less then UINT32_MAX would not catch that conidtion. If the lengths are sufficiently large then the total length can wrap around and it would go uncaught.

> Unsigned integer arithmetic is always performed modulo 2n where n is the number of bits in that particular integer. E.g. for unsigned int, adding one to UINT_MAX gives ​0​, and subtracting one from ​0​ gives UINT_MAX. 
See https://en.cppreference.com/w/c/language/operator_arithmetic

I doubted myself when you initially made this comment, and collapsed it down to this one statement, but on review again I think validating this is important for correctly validating the later condition required in step 4 of the specification.

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: Let me revisit this with our internal customers first to see what the max derived output length they would reasonably require. It seems suspect that a single invocation of the SSKDF is going to ask for 32 GiB of output key material in one-shot with a given shared secret and info string?

OpenSSL's implementation currently places the following limitations that we might wish to impose to have matching capatability between our implementations:
`secret_len <= 1<<30`, `info_len <= 1<<30`, and also limits the requested derivation length to be `derived_key_len <= 1<<30`.

We could relax this restriction, but that could also result in issues in some hypothetical key agreement scheme where the party using AWS-LC is able to arrived to a derived key with larger inputs, while the party using OpenSSL would fail. Hypothetical of course, but such a scenario could arrise if we choose to differ. Regardless I think the call out here is that the values here, and those by OpenSSL are sufficient enough to satisfy the NIST requirements.

skmcgrail: Some proof that the OpenSSL limits are sufficient to cap the inputs to satisfy the NIST table 2 and 3:

```C
#include <stddef.h>
#include <stdint.h>

#define OPENSSL_SSKDF_MAX (1 << 30)
#define NIST_MAX_APPROXIMATION (1UL << 60)

int check_overflow(size_t a, size_t b, size_t c) {
  if (a > SIZE_MAX - b || a + b > SIZE_MAX - c) {
    return 0;
  }
  return 1;
}

int main() {
  size_t z = OPENSSL_SSKDF_MAX;
  size_t info_len = OPENSSL_SSKDF_MAX;
  size_t counter = UINT32_MAX;

  if (!check_overflow(z, info_len, counter)) {
    return 1;
  }

  if (z + info_len + counter > NIST_MAX_APPROXIMATION) {
    return 2;
  }

  return 0;
}
```

This program returns 0 if you compile and execute it, thus showing the OpenSSL imposed limits are sufficient to appropriately cap using the smallest max input limit opposed by the tables (max `2^64 - 512` in bits). As the sum of the max secret length, max info length, and max counter will not exceed `2^60` which is the nearest power of 2 smaller then `(2^64 - 512)/8`.

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,336 @@
...
+  HMAC_CTX_free(variant_ctx->hmac_ctx);
+  OPENSSL_free(variant_ctx);
+  ctx->data = NULL;
+}
+
+static size_t sskdf_variant_digest_output_size(sskdf_variant_ctx *ctx) {
+  if (!ctx || !ctx->data) {
+    return 0;
+  }
+  sskdf_variant_digest_ctx *variant_ctx = (sskdf_variant_digest_ctx *)ctx->data;
```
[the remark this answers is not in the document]
skmcgrail: ctx->data is already checked above which is the same check, all this is doing is casting the type from void.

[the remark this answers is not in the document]
skmcgrail: I believe the other version is a miss on my part. `ctx->data` is already checked on line 111 above to verify the address is not null. There is no need to check it again after casting it from a void type.

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -0,0 +1,336 @@
...
+      !HMAC_Final(variant_ctx->hmac_ctx, out, &written) || out_len != written) {
+    return 0;
+  }
+
+  return 1;
+}
+
+
+DEFINE_METHOD_FUNCTION(sskdf_variant, sskdf_variant_digest) {
+  memset(out, 0, sizeof(sskdf_variant));
```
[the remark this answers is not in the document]
skmcgrail: It's technically not necessary, I can remove it.

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -174,6 +175,9 @@ static int sskdf_variant_hmac_compute(sskdf_variant_ctx *ctx, uint8_t *out,
 
   uint32_t written;
 
+  // NIST.SP.800-56Cr2: Step 6.2 HMAC-hash(salt, counter || secret || info)
+  // Note: |variant_ctx->hmac_ctx| is already initalized with the salt during
+  // it's initial construction.
```
nebeid, line 180: Nit:
```Suggestion
  // its initial construction.
```

nebeid (approved):

dkostic (approved):

crypto/fipsmodule/kdf/sskdf.c
skmcgrail: Will make this documentation correction in the KDF counter PR.

lilpoozie2005 (approved):

PR #1617, opened 2024-06-03
Script for creating compilation database

justsmth:
### Description of changes: 
* A [compilation database](https://clang.llvm.org/docs/JSONCompilationDatabase.html) is a JSON file that helps facilitate code assistance within a code/text editor.  This is typically via the editor's integration with [Clangd/LSP](https://clangd.llvm.org/).  It is supported by a variety of editors: [VSCode](https://code.visualstudio.com/docs/cpp/faq-cpp#_how-do-i-get-intellisense-to-work-correctly), [CLion](https://www.jetbrains.com/help/clion/compilation-database.html), [Neovim](https://neovim.io/), [Helix](https://helix-editor.com/) among others.
* This script simplifies the creation of this database for AWS-LC: `compile_commands.json`
* This adds the `compile_commands.json` file to `.gitignore`.

### Testing
* I tested the script on both Linux and macOS.
```
❯ ls -lh *.json
-rw-rw-r-- 1 justsmth justsmth 456K Jun  3 14:04 compile_commands.json
```

### DIFF
--- a/util/build_compilation_database.sh
+++ b/util/build_compilation_database.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -ex
+
+BASE_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}/" )/.." &> /dev/null && pwd )
+
+TMP_DIR=`mktemp -d`
+echo ${TMP_DIR}
+AWS_LC_BUILD="${TMP_DIR}/AWS-LC-BUILD"
+
+MY_CMAKE_FLAGS=("-GNinja" "-DCMAKE_BUILD_TYPE=Debug" "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON")
+
+mkdir -p "${AWS_LC_BUILD}"
+
+cmake "${BASE_DIR}" -B "${AWS_LC_BUILD}" ${MY_CMAKE_FLAGS[@]}
+
+cmake --build "${AWS_LC_BUILD}" --target all
+
+cp "${AWS_LC_BUILD}"/compile_commands.json "${BASE_DIR}"/
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ symbols.txt
 .fleet/
 .cache/
 /CMakePresets.json
+/compile_commands.json


util/build_compilation_database.sh
```diff
@@ -0,0 +1,19 @@
...
+
+set -ex
+
+BASE_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}/" )/.." &> /dev/null && pwd )
+
+TMP_DIR=`mktemp -d`
+echo ${TMP_DIR}
+AWS_LC_BUILD="${TMP_DIR}/AWS-LC-BUILD"
+
+MY_CMAKE_FLAGS=("-GNinja" "-DCMAKE_BUILD_TYPE=Debug" "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON")
```
nebeid, line 11: The description page says in https://clang.llvm.org/docs/JSONCompilationDatabase.html#supported-systems that "Ninja is in the works"; is it supported by now?

justsmth: Apparently so. The [CMake documentation for the `CMAKE_EXPORT_COMPILE_COMMANDS` flag](https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html) says:
> This option is implemented only by [Makefile Generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#makefile-generators) and [Ninja Generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#ninja-generators). It is ignored on other generators.

WillChilds-Klein, line 11: looks like this export option was added in CMake 2.8.5. do we care about versions older than that? if so, suggest checking CMake version and exiting w/ descriptive error, else this option will be ignored and the script will silently fail to produce the compilation DB.

WillChilds-Klein (approved):

util/build_compilation_database.sh
justsmth: AWS-LC won't build with CMake prior to 3.2.  We've got CI to help ensure we don't regress on that here: https://github.com/aws/aws-lc/blob/main/.github/workflows/cmake.yml

nebeid (approved):

PR #1619, opened 2024-06-03
Prevent non-constant-time code in Kyber-R3 and ML-KEM implementation

geedo0:
### Issues:
Resolves V1399146249

### Description of changes: 
A branch can be introduced by the compiler from the `poly_frommsg` code. This has been reported for Clang versions
15-18, x86_64, for the following compiler flags:
- `-Os`
- `-O1`
- `-O2 -fno-vectorize`
- `-O3 -fno-vectorize`

An upstream fix has been published [here](https://github.com/pq-crystals/kyber/commit/9b8d30698a3e7449aeb34e62339d4176f11e3c6c) which we are deciding not to take directly. Due to the way AWS-LC compiles the Kyber reference code, it may not have the
desired effect. We instead opt to use our own constant-time implementations for performing conditional bit-masks which
have much stronger guarantees around portability.

### Call-outs:

- An interesting observation is that the old code would *always* produce SIMD
instructions despite passing in `-fno-vectorize` and technically never produced
the non-constant-time branching described.
- This new code for performing the conditional bit-masks seems to completely
avoid SIMD now. The fact that this is more controlled/predicted is an improvement in my book.

### Testing:

- Compiled the source on a MacOS, x86_65, Clang-15 machine.
- Test compiled -Os, -O2 -fno-vectorize cases.
- Disassembled the `poly_frommsg` function and checked for any
non-constant-time-behavior
- Disassembly is below for the reviewer

```
# -Os
pqcrystals_kyber512_ref_poly_frommsg:
  pushq  %rbp
  movq   %rsp, %rbp
  xorl   %eax, %eax
  xorl   %ecx, %ecx
  movzbl (%rsi,%rax), %edx
  btl    %ecx, %edx
  movl   $0x0, %edx
  sbbq   %rdx, %rdx
  andl   $0x681, %edx               ; imm = 0x681
  movw   %dx, (%rdi,%rcx,2)
  incq   %rcx
  cmpq   $0x8, %rcx
  jne    0x1005774ad                ; <+8> at poly.c:177:37
  incq   %rax
  addq   $0x10, %rdi
  cmpq   $0x20, %rax
  jne    0x1005774ab                ; <+6> at poly.c:167
  popq   %rbp
  retq

# -O2 -fno-vectorize
pqcrystals_kyber512_ref_poly_frommsg:
  pushq  %rbp                     
  movq   %rsp, %rbp               
  addq   $0xe, %rdi               
  xorl   %eax, %eax               
  nopw   (%rax,%rax)              
  movzbl (%rsi,%rax), %ecx        
  andl   $0x1, %ecx               
  negq   %rcx                     
  andl   $0x681, %ecx               ; imm = 0x681
  movw   %cx, -0xe(%rdi)          
  movzbl (%rsi,%rax), %ecx        
  shlb   $0x6, %cl                
  sarb   $0x7, %cl                
  movsbq %cl, %rcx                
  andl   $0x681, %ecx               ; imm = 0x681
  movw   %cx, -0xc(%rdi)          
  movzbl (%rsi,%rax), %ecx        
  shlb   $0x5, %cl                
  sarb   $0x7, %cl                
  movsbq %cl, %rcx                
  andl   $0x681, %ecx               ; imm = 0x681
  movw   %cx, -0xa(%rdi)          
  movzbl (%rsi,%rax), %ecx        
  shlb   $0x4, %cl                
  sarb   $0x7, %cl                
  movsbq %cl, %rcx                
  andl   $0x681, %ecx               ; imm = 0x681
  movw   %cx, -0x8(%rdi)          
  movzbl (%rsi,%rax), %ecx        
  shlb   $0x3, %cl                
  sarb   $0x7, %cl                
  movsbq %cl, %rcx                
  andl   $0x681, %ecx               ; imm = 0x681
  movw   %cx, -0x6(%rdi)          
  movzbl (%rsi,%rax), %ecx        
  shlb   $0x2, %cl                
  sarb   $0x7, %cl                
  movsbq %cl, %rcx                
  andl   $0x681, %ecx               ; imm = 0x681
  movw   %cx, -0x4(%rdi)          
  movzbl (%rsi,%rax), %ecx        
  addb   %cl, %cl                 
  sarb   $0x7, %cl                
  movsbq %cl, %rcx                
  andl   $0x681, %ecx               ; imm = 0x681
  movw   %cx, -0x2(%rdi)          
  movzbl (%rsi,%rax), %ecx        
  sarb   $0x7, %cl                
  movsbq %cl, %rcx                
  andl   $0x681, %ecx               ; imm = 0x681
  movw   %cx, (%rdi)              
  incq   %rax                     
  addq   $0x10, %rdi              
  cmpq   $0x20, %rax              
  jne    0x100331690                ; <+16> [inlined] constant_time_msb_w + 6 at internal.h:325:18
  popq   %rbp                     
  retq                            
  nopw   (%rax,%rax) 
```


crypto/kyber/pqcrystals_kyber_ref_common/poly.c
```diff
@@ -164,16 +166,17 @@ void poly_frombytes(poly *r, const uint8_t a[KYBER_POLYBYTES])
...
 
 #if (KYBER_INDCPA_MSGBYTES != KYBER_N/8)
 #error "KYBER_INDCPA_MSGBYTES must be equal to KYBER_N/8 bytes!"
 #endif
 
   for(i=0;i<KYBER_N/8;i++) {
     for(j=0;j<8;j++) {
-      mask = -(int16_t)((msg[i] >> j)&1);
-      r->coeffs[8*i+j] = mask & ((KYBER_Q+1)/2);
+      mask = constant_time_lt_w(0, (msg[i] >> j) & 1);
```
dkostic, line 177: can we please use `constant_time_is_zero_w` since it makes more sense in this context?

crypto/kyber/pqcrystals_kyber_ref_common/poly.c
```diff
@@ -164,16 +166,17 @@ void poly_frombytes(poly *r, const uint8_t a[KYBER_POLYBYTES])
...
+      r->coeffs[8*i+j] = (int16_t) constant_time_select_w(mask,
```
dkostic, line 178: can we add a comment that the conversion to `int16_t` works only because the possible values (`(KYBER_Q+1)/2`, 0) are such that no error is caused by the cast? maybe add an assert as well?

dkostic:
we should do the same for ML-KEM?

crypto/kyber/pqcrystals_kyber_ref_common/poly.c
geedo0: Sure, but we end up needing to introduce some inversion logic external to `constant_time_*`. i.e. We have to either invert the output mask or we invert the input bit. I don't think it adds risk to our compiler output, but just calling it out.

crypto/kyber/pqcrystals_kyber_ref_common/poly.c
```diff
@@ -164,16 +169,21 @@ void poly_frombytes(poly *r, const uint8_t a[KYBER_POLYBYTES])
...
 
 #if (KYBER_INDCPA_MSGBYTES != KYBER_N/8)
 #error "KYBER_INDCPA_MSGBYTES must be equal to KYBER_N/8 bytes!"
 #endif
 
   for(i=0;i<KYBER_N/8;i++) {
     for(j=0;j<8;j++) {
-      mask = -(int16_t)((msg[i] >> j)&1);
-      r->coeffs[8*i+j] = mask & ((KYBER_Q+1)/2);
+      mask = ~constant_time_is_zero_w((msg[i] >> j) & 1);
```
dkostic, line 180: nit/micro optimization: no need for `~` if you reverse the order of the inputs in `constant_time_select_w`

crypto/ml_kem/ml_kem_ipd_ref_common/poly.c
```diff
@@ -167,16 +172,21 @@ void poly_frombytes(poly *r, const uint8_t a[KYBER_POLYBYTES])
...
 
 #if (KYBER_INDCPA_MSGBYTES != KYBER_N/8)
 #error "KYBER_INDCPA_MSGBYTES must be equal to KYBER_N/8 bytes!"
 #endif
 
   for(i=0;i<KYBER_N/8;i++) {
     for(j=0;j<8;j++) {
-      mask = -(int16_t)((msg[i] >> j)&1);
-      r->coeffs[8*i+j] = mask & ((KYBER_Q+1)/2);
+      mask = ~constant_time_is_zero_w((msg[i] >> j) & 1);
```
dkostic, line 183: nit/micro optimization: no need for ~ if you reverse the order of the inputs in constant_time_select_w

crypto/kyber/pqcrystals_kyber_ref_common/poly.c
geedo0: I like it, that's much clearer, will add in the next rev.

dkostic (approved):

nebeid (approved):

PR #1620, opened 2024-06-04
Fix SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR behavior

samuel40791765:
### Description of changes: 
This was discovered when taking https://github.com/google/boringssl/commit/5b3dc49c1271554f73b976c2c625600d6bd912b0 during the upstream merge.
`ERR_clear_error` is being called more eagerly with the new change, which led us to discover that `SSLTest.BuildCertChain` was actually testing against an error code propagated onto the stack by the previous call to `SSL_CTX_build_cert_chain`.

Upon further examination, we weren't propagating an error when calling `SSL_CTX_build_cert_chain` with `SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR` . The correct behavior should be to push an error onto the stack regardless.

### Call-outs:
N/A

### Testing:
Slight test tweaks

### DIFF
--- a/ssl/ssl_x509.cc
+++ b/ssl/ssl_x509.cc
@@ -1061,12 +1061,13 @@ static int ssl_build_cert_chain(CERT *cert, X509_STORE *cert_store, int flags) {
 
   bool ignore_error = false;
   if (X509_verify_cert(store_ctx.get()) <= 0) {
+    OPENSSL_PUT_ERROR(SSL, SSL_R_CERTIFICATE_VERIFY_FAILED);
+    ERR_add_error_data(2, "Verify error:",
+                       X509_verify_cert_error_string(
+                           X509_STORE_CTX_get_error(store_ctx.get())));
+
     // Fail if |SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR| is not set.
-    if(!is_flag_set(flags, SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR)) {
-      OPENSSL_PUT_ERROR(SSL, SSL_R_CERTIFICATE_VERIFY_FAILED);
-      ERR_add_error_data(2, "Verify error:",
-                         X509_verify_cert_error_string(
-                             X509_STORE_CTX_get_error(store_ctx.get())));
+    if (!is_flag_set(flags, SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR)) {
       return 0;
     }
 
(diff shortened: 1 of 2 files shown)


ssl/ssl_x509.cc
```diff
@@ -1061,12 +1061,13 @@ static int ssl_build_cert_chain(CERT *cert, X509_STORE *cert_store, int flags) {
...
+                       X509_verify_cert_error_string(
+                           X509_STORE_CTX_get_error(store_ctx.get())));
+
     // Fail if |SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR| is not set.
-    if(!is_flag_set(flags, SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR)) {
-      OPENSSL_PUT_ERROR(SSL, SSL_R_CERTIFICATE_VERIFY_FAILED);
-      ERR_add_error_data(2, "Verify error:",
-                         X509_verify_cert_error_string(
-                             X509_STORE_CTX_get_error(store_ctx.get())));
+    if (!is_flag_set(flags, SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR)) {
```
andrewhop, line 1070: Is this the right spot to put this? I think SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR is supposed to interact with SSL_BUILD_CHAIN_FLAG_CLEAR_ERROR down below:
>SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR ignores any errors during verification: if flag SSL_BUILD_CHAIN_FLAG_CLEAR_ERROR is also set verification errors are cleared from the error queue.

samuel40791765: Hmm I still think it is. 
This if statement checks if `SSL_BUILD_CHAIN_FLAG_IGNORE_ERROR` is non-existent. If it does not exist, we return `0` here since we chose not to ignore errors.

If it does exist, we ignore errors and continue the code path below. If `SSL_BUILD_CHAIN_FLAG_CLEAR_ERROR` exists, we clear errors from the queue.

Both code blocks are not executed if the previous `X509_verify_cert` call was successful though. I have tests in `ssl/ssl_test.cc` [from L5219 to L5235 ](https://github.com/aws/aws-lc/pull/1620/files#diff-39cb80e1b51b8f538bf9667decca8d5f25849138ec5623ddece489113dc3c601R5219-L5234)trying to verify the different flag behaviors.

torben-hansen (approved):

andrewhop (approved):

PR #1624, opened 2024-06-05
Add support for NETSCAPE_SPKI_print

samuel40791765:
### Issues:
Resolves `CryptoAlg-1717`

### Description of changes: 
Ruby [consumes NETSCAPE_SPKI_print](https://github.com/ruby/ruby/blob/0b31986909eaccb6b5f8f99f9ab2a458170a21af/ext/openssl/ossl_ns_spki.c#L167) for debugging purposes. This adds support for the symbol for easier integration.

### Call-outs:
N/A

### Testing:
Expected output test

### DIFF
--- a/crypto/x509/x509_test.cc
+++ b/crypto/x509/x509_test.cc
@@ -7618,6 +7618,7 @@ TEST(X509Test, SPKIPrint) {
   const uint8_t *data;
   size_t data_len;
   ASSERT_TRUE(BIO_mem_contents(bio.get(), &data, &data_len));
+  ASSERT_GT(data_len, strlen(expected_certificate_string));
   std::string print(reinterpret_cast<const char *>(data),
                     strlen(expected_certificate_string));
   EXPECT_EQ(print, expected_certificate_string);
--- a/crypto/x509/x509_test.cc
+++ b/crypto/x509/x509_test.cc
@@ -7583,3 +7583,42 @@ TEST(X509Test, PublicKeyCache) {
   key2.reset(X509_PUBKEY_get(pub));
   EXPECT_FALSE(key2);
 }
+
+TEST(X509Test, SPKIPrint) {
+  bssl::UniquePtr<BIO> bio(BIO_new(BIO_s_mem()));
+  ASSERT_TRUE(bio);
+  bssl::UniquePtr<NETSCAPE_SPKI> spki(NETSCAPE_SPKI_new());
+  ASSERT_TRUE(spki);
+
+  bssl::UniquePtr<EVP_PKEY> key = PrivateKeyFromPEM(kP256Key);
+  EXPECT_TRUE(NETSCAPE_SPKI_set_pubkey(spki.get(), key.get()));
+  EXPECT_TRUE(NETSCAPE_SPKI_sign(spki.get(), key.get(), EVP_sha256()));
+
+  std::string challenge = "challenge string";
+  ASSERT_TRUE(ASN1_STRING_set(spki.get()->spkac->challenge, challenge.data(),
+                              challenge.size()));
+
+  EXPECT_TRUE(NETSCAPE_SPKI_print(bio.get(), spki.get()));
+
+  // The contents of the signature is printed last but it's randomized,
+  // so we only check the expected output before that.
+  static const char expected_certificate_string[] = R"(Netscape SPKI:
+  Public Key Algorithm: id-ecPublicKey
+    Public-Key: (P-256)
+    pub:
+        04:e6:2b:69:e2:bf:65:9f:97:be:2f:1e:0d:94:8a:
+        4c:d5:97:6b:b7:a9:1e:0d:46:fb:dd:a9:a9:1e:9d:
+        dc:ba:5a:01:e7:d6:97:a8:0a:18:f9:c3:c4:a3:1e:
+        56:e2:7c:83:48:db:16:1a:1c:f5:1d:7e:f1:94:2d:
+        4b:cf:72:22:c1
+  Challenge String: challenge string
+  Signature Algorithm: ecdsa-with-SHA256
+  )";
+
+  const uint8_t *data;
+  size_t data_len;
+  ASSERT_TRUE(BIO_mem_contents(bio.get(), &data, &data_len));
+  std::string print(reinterpret_cast<const char *>(data),
+                    strlen(expected_certificate_string));
+  EXPECT_EQ(print, expected_certificate_string);
+}
--- a/crypto/x509/x509spki.c
+++ b/crypto/x509/x509spki.c
@@ -134,7 +134,8 @@ char *NETSCAPE_SPKI_b64_encode(NETSCAPE_SPKI *spki) {
 }
 
 int NETSCAPE_SPKI_print(BIO *out, NETSCAPE_SPKI *spki) {
-  if (out == NULL || spki == NULL) {
+  if (out == NULL || spki == NULL || spki->spkac->pubkey == NULL ||
+      spki->sig_algor == NULL || spki->signature == NULL) {
     OPENSSL_PUT_ERROR(X509, ERR_R_PASSED_NULL_PARAMETER);
     return 0;
   }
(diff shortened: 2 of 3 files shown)


WillChilds-Klein (commented):
can we get away with just defining no-op'ing this? given that this is used for debugging, i'm wondering whether this would actually impact any tests or programmatic functionality required by applications.

samuel40791765:
> can we get away with just defining no-op'ing this? 

We could consider it, but it did seem harmless to incorporate given that we have support for all the surrounding functions within the module: https://github.com/ruby/ruby/blob/ruby_2_7/ext/openssl/ossl_ns_spki.c#L378-L405

Great question on potential impact, I did some more digging. The only concern would be us failing [this Ruby test](https://github.com/ruby/ruby/blob/ruby_2_7/test/openssl/test_ns_spki.rb#L35) which asserts that the inputted BIO has at least some contents, but we can pass a short string into BIO to prevent that from happening.

My preference would be supporting the entire module with no caveats, but I'm open to any other concerns with us supporting this.

crypto/x509/x509_test.cc
```diff
@@ -7312,3 +7312,43 @@ TEST(X509Test, ParamInheritance) {
...
+        dc:ba:5a:01:e7:d6:97:a8:0a:18:f9:c3:c4:a3:1e:
+        56:e2:7c:83:48:db:16:1a:1c:f5:1d:7e:f1:94:2d:
+        4b:cf:72:22:c1
+  Challenge String: challenge string
+  Signature Algorithm: ecdsa-with-SHA256
+  )";
+
+  const uint8_t *data;
+  size_t data_len;
+  ASSERT_TRUE(BIO_mem_contents(bio.get(), &data, &data_len));
```
WillChilds-Klein, line 7349: nit: should we assert that `strlen(expected_certificate_string) >= data_len`?

crypto/x509/x509spki.c
```diff
@@ -131,3 +132,45 @@ char *NETSCAPE_SPKI_b64_encode(NETSCAPE_SPKI *spki) {
...
+  X509_PUBKEY_get0_param(&spkioid, NULL, NULL, NULL, spki->spkac->pubkey);
+  BIO_printf(out, "  Public Key Algorithm: %s\n",
+             (OBJ_obj2nid(spkioid) == NID_undef)
+                 ? "UNKNOWN"
+                 : OBJ_nid2ln(OBJ_obj2nid(spkioid)));
+  EVP_PKEY *pkey = X509_PUBKEY_get0(spki->spkac->pubkey);
+  if (pkey == NULL) {
+    BIO_printf(out, "  Unable to load public key\n");
+  } else {
+    EVP_PKEY_print_public(out, pkey, 4, NULL);
```
WillChilds-Klein, line 154: what's the significance of `4` here?

crypto/x509/x509spki.c
```diff
@@ -131,3 +132,45 @@ char *NETSCAPE_SPKI_b64_encode(NETSCAPE_SPKI *spki) {
...
+int NETSCAPE_SPKI_print(BIO *out, NETSCAPE_SPKI *spki) {
+  if (out == NULL || spki == NULL) {
+    OPENSSL_PUT_ERROR(X509, ERR_R_PASSED_NULL_PARAMETER);
+    return 0;
+  }
+  BIO_printf(out, "Netscape SPKI:\n");
+
+  // Print out public key algorithm and contents.
+  ASN1_OBJECT *spkioid;
+  X509_PUBKEY_get0_param(&spkioid, NULL, NULL, NULL, spki->spkac->pubkey);
```
WillChilds-Klein, line 145: are we sure that `spki->spkac` is non-null?

crypto/x509/x509spki.c
```diff
@@ -131,3 +132,45 @@ char *NETSCAPE_SPKI_b64_encode(NETSCAPE_SPKI *spki) {
...
+  }
+
+  ASN1_IA5STRING *chal = spki->spkac->challenge;
+  if (chal->length != 0) {
+    BIO_printf(out, "  Challenge String: %.*s\n", chal->length, chal->data);
+  }
+
+  // Print out signature algorithm and contents.
+  BIO_printf(out, "  Signature Algorithm: %s",
+             (OBJ_obj2nid(spki->sig_algor->algorithm) == NID_undef)
```
WillChilds-Klein, line 164: are we sure that `spki->sig_algor` is non-null?

crypto/x509/x509spki.c
```diff
@@ -131,3 +132,45 @@ char *NETSCAPE_SPKI_b64_encode(NETSCAPE_SPKI *spki) {
...
+                 ? "UNKNOWN"
+                 : OBJ_nid2ln(OBJ_obj2nid(spki->sig_algor->algorithm)));
+  for (int i = 0; i < spki->signature->length; i++) {
+    if ((i % 18) == 0) {
+      BIO_write(out, "\n      ", 7);
```
WillChilds-Klein, line 169: nit: suggest pulling the string out into a `const` and calling `strlen()` on it in the `BIO_write` invocation

or perhaps consider using `BIO_printf`

crypto/x509/x509spki.c
```diff
@@ -131,3 +132,45 @@ char *NETSCAPE_SPKI_b64_encode(NETSCAPE_SPKI *spki) {
...
+  if (chal->length != 0) {
+    BIO_printf(out, "  Challenge String: %.*s\n", chal->length, chal->data);
+  }
+
+  // Print out signature algorithm and contents.
+  BIO_printf(out, "  Signature Algorithm: %s",
+             (OBJ_obj2nid(spki->sig_algor->algorithm) == NID_undef)
+                 ? "UNKNOWN"
+                 : OBJ_nid2ln(OBJ_obj2nid(spki->sig_algor->algorithm)));
+  for (int i = 0; i < spki->signature->length; i++) {
```
WillChilds-Klein, line 167: are we sure that `spki->signature` is non-null?

crypto/x509/x509spki.c
```diff
@@ -131,3 +132,45 @@ char *NETSCAPE_SPKI_b64_encode(NETSCAPE_SPKI *spki) {
...
+  // Print out public key algorithm and contents.
+  ASN1_OBJECT *spkioid;
+  X509_PUBKEY_get0_param(&spkioid, NULL, NULL, NULL, spki->spkac->pubkey);
+  BIO_printf(out, "  Public Key Algorithm: %s\n",
+             (OBJ_obj2nid(spkioid) == NID_undef)
+                 ? "UNKNOWN"
+                 : OBJ_nid2ln(OBJ_obj2nid(spkioid)));
+  EVP_PKEY *pkey = X509_PUBKEY_get0(spki->spkac->pubkey);
+  if (pkey == NULL) {
+    BIO_printf(out, "  Unable to load public key\n");
```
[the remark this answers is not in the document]
samuel40791765: Yes this function is just printing out the contents of the struct. The absence of a public key doesn't indicate that the print is unsuccessful, it just means that there isn't one set.
We could make the argument that the wording could be better though, but I didn't want any unnecessary misalignments from upstream behavior.

crypto/x509/x509_test.cc
samuel40791765: I'll add the assertion, but It should be the other way around since the actual `data` has an additional random signature appended at the end.

crypto/x509/x509spki.c
samuel40791765: Yes the seemingly arbitrary numbers throughout this code are all indentations

crypto/x509/x509spki.c
samuel40791765: Added assertions for all the required non-null parameters at the top

crypto/x509/x509spki.c
samuel40791765: Great point, I automatically assumed 7 was an indentation.. Changed to use `BIO_printf` instead

crypto/x509/x509spki.c
```diff
@@ -131,3 +132,45 @@ char *NETSCAPE_SPKI_b64_encode(NETSCAPE_SPKI *spki) {
   OPENSSL_free(der_spki);
   return b64_str;
 }
+
+int NETSCAPE_SPKI_print(BIO *out, NETSCAPE_SPKI *spki) {
+  if (out == NULL || spki == NULL || spki->spkac->pubkey == NULL ||
```
justsmth, line 137: I think this should be checking `spki->spkac == NULL` instead of `spki->spkac->pubkey`.

justsmth (approved):

WillChilds-Klein (approved):

PR #1637, opened 2024-06-14
Add EVP_md_null and SSL_set_ciphersuites

WillChilds-Klein:
### Issues:
Addresses CryptoAlg-2464

### Description of changes: 

This commit adds two functions used by OpenLDAP 2.5.17+. Both functions are very straightforward.

From OpenSSL's [docs][1] for `EVP_md_null`:

>A "null" message digest that does nothing: i.e. the hash it returns is
of zero length.

From OpenSSL's [docs][2] for `SSL_set_ciphersuites`:

> SSL_set_ciphersuites() is the same as SSL_CTX_set_ciphersuites()
except it configures the ciphersuites for ssl.

[1]: https://www.openssl.org/docs/man1.1.1/man3/EVP_md_null.html
[2]: https://www.openssl.org/docs/man1.1.1/man3/SSL_set_ciphersuites.html

### Call-outs:
- n/a

### Testing:
- new integration test
- increased unit test coverage

---

### DIFF
--- a/crypto/digest_extra/digest_extra.c
+++ b/crypto/digest_extra/digest_extra.c
@@ -282,15 +282,15 @@ static void null_update(EVP_MD_CTX *ctx, const void *data, size_t count) {}
 static void null_final(EVP_MD_CTX *ctx, unsigned char *md) {}
 
 static const EVP_MD evp_md_null = {
-    NID_undef,
-    0,
-    0,
-    null_init,
-    null_update,
-    null_final,
-    0,
-    0,
-    NULL,
+  NID_undef,
+  0,
+  0,
+  null_init,
+  null_update,
+  null_final,
+  0,
+  sizeof(EVP_MD_CTX),
+  NULL,
 };
 
 const EVP_MD *EVP_md_null(void) { return &evp_md_null; }
--- a/crypto/digest_extra/digest_extra.c
+++ b/crypto/digest_extra/digest_extra.c
@@ -274,3 +274,23 @@ static const EVP_MD evp_md_blake2b256 = {
 };
 
 const EVP_MD *EVP_blake2b256(void) { return &evp_md_blake2b256; }
+
+static void null_init(EVP_MD_CTX *ctx) {}
+
+static void null_update(EVP_MD_CTX *ctx, const void *data, size_t count) {}
+
+static void null_final(EVP_MD_CTX *ctx, unsigned char *md) {}
+
+static const EVP_MD evp_md_null = {
+    NID_undef,
+    0,
+    0,
+    null_init,
+    null_update,
+    null_final,
+    0,
+    0,
+    NULL,
+};
+
+const EVP_MD *EVP_md_null(void) { return &evp_md_null; }
--- a/ssl/ssl_lib.cc
+++ b/ssl/ssl_lib.cc
@@ -2221,6 +2221,10 @@ int SSL_CTX_set_ciphersuites(SSL_CTX *ctx, const char *str) {
                                 true /* only configure TLSv1.3 ciphers */);
 }
 
+int SSL_set_ciphersuites(SSL *ssl, const char *str) {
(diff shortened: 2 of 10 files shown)


crypto/digest_extra/digest_extra.c
smittals2, line 1: Might be a dumb question but OpenSSL defined the null_{init, update, final} with int return values (they always return 1 for success). Is there a reason we don't/don't need to?

smittals2 (approved):

ssl/ssl_lib.cc
```diff
@@ -2221,6 +2221,10 @@ int SSL_CTX_set_ciphersuites(SSL_CTX *ctx, const char *str) {
                                 true /* only configure TLSv1.3 ciphers */);
 }
 
+int SSL_set_ciphersuites(SSL *ssl, const char *str) {
+  return SSL_CTX_set_ciphersuites(ssl->ctx.get(), str);
```
samuel40791765, line 2225: This seems a bit strange since I was under the impression that `SSL` objects are created from `SSL_CTX`s, so it doesn't seem quite right for `SSL_set_ciphersuites` to be configuring contents from the parent object.

[Our `SSL_CONFIG`](https://github.com/aws/aws-lc/blob/34d5e8ffd8df2424c7ef0f86108fca2c6869d349/ssl/internal.h#L3164-L3170) is sort of like OpenSSL's `SSL_CONNECTION` and contents about the `cipher_list` are saved there as well. I think we should configure the `cipher_list` in `SSL_CONFIG` instead, since it's more specific to the `SSL` structure. [`SSL(_CTX)_set_cipher_list`](https://github.com/aws/aws-lc/blob/34d5e8ffd8df2424c7ef0f86108fca2c6869d349/ssl/ssl_lib.cc#L2188-L2214) and `SSL(_CTX)_set_strict_cipher_list` may serve as good examples.

crypto/digest_extra/digest_extra.c
WillChilds-Klein: Not a dumb question. We define the function signature for `EVP_MD` init, update, and finish functions [here](https://github.com/aws/aws-lc/blame/34d5e8ffd8df2424c7ef0f86108fca2c6869d349/crypto/fipsmodule/digest/internal.h#L78-L85) and OpenSSL does so [here](https://github.com/openssl/openssl/blob/d5412c94a399d3923b2dec2431ead60288d857c8/include/crypto/evp.h#L264-L266). As you note, ours have `void` return type while OpenSSL's return `int`. BoringSSL made this change [about 10 years ago](https://github.com/aws/aws-lc/commit/2c3679210e77f825782317e6b5d4ffc7eae4b81b) for the simple reason that hash functions (or at least the hash functions they supported at the time) cannot fail like e.g. authenticated cipher decryption can. An implication of this choice of `void` return type is that the init/update/finish functions cannot do any sort of memory allocation or other non-algorithmic operation that has a failure mode.

I'm not sure why OpenSSL decided to allow for failure in `EVP_MD` init/update/finish...

ssl/ssl_lib.cc
WillChilds-Klein: Ah, good catch. This should be just like `SSL_set_cipher_list` and `SSL_set_strict_cipher_list` but for TLSv1.3 and not strict. Will fix.

crypto/digest_extra/digest_extra.c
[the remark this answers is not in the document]
WillChilds-Klein: It looks to be used for detecting weak/missing hash function specification. OpenLDAP [will default](https://github.com/openldap/openldap/blob/aa31963e44e8e89bf7752a39d19c78b22c3711e8/libraries/libldap/tls_o.c#L1091) a digest function used internally to SHA256 if the specified NID is invalid, unspecified, or broken (e.g. MD5).

samuel40791765 (approved):

smittals2 (approved):

PR #1639, opened 2024-06-17
Remove dead tail code from (non-SHA3) AES-GCM AArch64 kernel

hanno-becker:
On AArch64 systems without support for EOR3, assembly kernels `aes_gcm_enc_kernel` and `aes_gcm_dec_kernel` from `aesv8-gcm-armv8.pl` are used for the bulk of AES-GCM processing. These kernels have dedicated tail code for handling inputs whose size is not a multiple of the block size (16 bytes).

However, the unique call-sites for `aes_gcm_enc_kernel` and `aes_gcm_dec_kernel` in `gcm.c` only invoke them with data of size a multiple of 16 bytes: See the masking here [here](https://github.com/aws/aws-lc/blob/98735a2f6723ba984a18b2f79e05173a61e0f869/crypto/fipsmodule/modes/gcm.c#L154) and [here](https://github.com/aws/aws-lc/blob/98735a2f6723ba984a18b2f79e05173a61e0f869/crypto/fipsmodule/modes/gcm.c#L191). This renders the tail code in `aesv8-gcm-armv8.pl` dead.

Simply removing the truncation to 16-byte aligned data in `gcm.c` -- that is, attempting to let `aes_gcm_{dec,enc}_kernel` process the entire data -- leads to tests failing. It is not clear to me why that is, and in particular the tail code could be faulty. OpenSSL seems to behave similarly and call the AArch64 AES-GCM kernels for block-sized data only.

This PR removes the dead tail code from the non-SHA3 AES-GCM kernels `aes_gcm_enc_kernel` and `aes_gcm_dec_kernel`. In a first commit, the code is annotated to explain the effect of the tail code in case of block-aligned data. In the second commit, the tail code is removed.

It seems that a similar change can be made for the AES-GCM kernels leveraging SHA3 instructions, but is not attempted here. (The primary motivation for this change is to facilitate @pennyannn's verification work, which focuses on the non-SHA3 kernels).

### DIFF
--- a/crypto/fipsmodule/modes/asm/aesv8-gcm-armv8.pl
+++ b/crypto/fipsmodule/modes/asm/aesv8-gcm-armv8.pl
@@ -849,31 +849,8 @@ aes_gcm_enc_kernel:
 	eor     $acc_mb, $acc_mb, $rk4v.16b                      // GHASH final-1 block - mid
 	eor     $acc_lb, $acc_lb, $rk3                           // GHASH final-1 block - low
 .Lenc_blocks_less_than_1:                                        // blocks left <= 1
-	and     $bit_length, $bit_length, #127                   // bit_length %= 128
-                                                                 // (NOTE: == 0 for block-sized data)
-	mvn     $rkN_l, xzr                                      // rkN_l = 0xffffffffffffffff
-	sub     $bit_length, $bit_length, #128                   // bit_length -= 128
-                                                                 // (NOTE: == -128 for block-sized data)
-	neg     $bit_length, $bit_length                         // bit_length = 128 - #bits in input (in range [1,128])
-                                                                 // (NOTE: == 128 for block-sized data)
-	ld1     { $rk0}, [$output_ptr]                           // load existing bytes where the possibly partial last block is to be stored
-                                                                 // NOTE: Not needed for block-sized data
-	mvn     $rkN_h, xzr                                      // rkN_h = 0xffffffffffffffff
-	and     $bit_length, $bit_length, #127                   // bit_length %= 128
-                                                                 // (NOTE: == 0 for block-sized data)
-	lsr     $rkN_h, $rkN_h, $bit_length                      // rkN_h is mask for top 64b of last block
-                                                                 // NOTE: == 0xFFFFFFFF for block-sized data
-	cmp     $bit_length, #64                                 // NOTE: Compare 0 < 64 for block-sized data ==> lt == TRUE
-	csel    $input_l0, $rkN_l, $rkN_h, lt                    // NOTE: 0xFFFFFFFF for block-sized data
-	csel    $input_h0, $rkN_h, xzr, lt                       // NOTE: 0xFFFFFFFF for block-sized data
-	fmov    $ctr0d, $input_l0                                // ctr0b is mask for last block
-	fmov    $ctr0.d[1], $input_h0                            // NOTE: ctr0 == 0xFFFFFFFFFFFFFFFF (all 1's) for block-sized data
-	and     $res1b, $res1b, $ctr0b                           // possibly partial last block has zeroes in highest bits
-                                                                 // NOTE: no-op for block-sized data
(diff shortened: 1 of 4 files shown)


hanno-becker:
@torben-hansen Did I understand correctly that the CI failures are unrelated to this PR?

hanno-becker:
**Review note**: Make sure to review commit https://github.com/aws/aws-lc/pull/1639/commits/12ac5a25e8c96198883062c6b2b3ad5ecb367982 separately as it provides some explanation of why the lines that are removed are indeed dispensable.

dkostic (dismissed):
not material for this change, but I'm really curious why are the tests failing when you try to process non-block aligned data with the removed code.

crypto/fipsmodule/modes/asm/aesv8-gcm-armv8.pl
```diff
@@ -849,31 +849,8 @@
 	eor     $acc_mb, $acc_mb, $rk4v.16b                      // GHASH final-1 block - mid
 	eor     $acc_lb, $acc_lb, $rk3                           // GHASH final-1 block - low
 .Lenc_blocks_less_than_1:                                        // blocks left <= 1
```
nebeid, line 851: Do we change these labels so they're not misleading?

nebeid (commented):
Thanks, @hanno-becker, for this contribution. Just a nit comment.

crypto/fipsmodule/modes/asm/aesv8-gcm-armv8.pl
hanno-becker: @nebeid Can you elaborate? I think the number of blocks to be processed is unchanged, or?

nebeid: The label suggests we're processing a partial block, which was the code that was removed, right?
On the other hand, the comment says `blocks left <=1`, so maybe (if I understand correctly)
```suggestion
.Lenc_blocks_eq_1:                                        // blocks left = 1
```

hanno-becker: Yes, you are right! I'll change it.

hanno-becker: https://github.com/aws/aws-lc/pull/1639/commits/bef148935a93b4c40c27c6fb5bba480830f0beea

crypto/fipsmodule/modes/asm/aesv8-gcm-armv8.pl
nebeid, line 1: I think remains to change the comments according to the new labels.

hanno-becker: Done

hanno-becker:
> @dkostic not material for this change, but I'm really curious why are the tests failing when you try to process non-block aligned data with the removed code.

It turns out that the tail code isn't buggy, it just presents a different interface to AES: For non-block aligned data, the tail code pads the data with 0 in the last block and computes AES + GHASH. That's different from the incremental behaviour of functions like `CRYPTO_gcm128_encrypt_ctr32`: When those find tail data smaller than a block, they just stash it in the AES context, until (a) further data completes the block, or (b) the entire computation is concluded, forcing the padding + final GHASH.

Rewriting the tail code assembly to suit the incremental AES interface does not seem trivial, so I'd sugges one revisits this alongside the integration of the clean and SLOTHY-optimized AES-GCM at a later stage.

nebeid (approved):

dkostic (approved):

hanno-becker:
Tested on Graviton2:
```
mkdir build && cd build
cmake -DCMAKE_BUILD_TYPE=Release .. && make -j8
./crypto/crypto_test
./ssl/ssl_test
```
All successful.

lilpoozie2005 (approved):

PR #1640, opened 2024-06-17
Snapsafe-type uniqueness breaking event detection

justsmth:
### Issues:
* Resolves: CryptoAlg-2505

### Description of changes: 
* Add support for Snapsafe-type uniqueness breaking event detection on Linux using SysGenID (`/dev/sysgenid`). Adds 4 new (internal-use) functions:
  * `int CRYPTO_get_snapsafe_generation(uint32_t *snapsafe_generation_number);`
  * `int CRYPTO_get_snapsafe_active(void);`
  * `int CRYPTO_get_snapsafe_supported(void);`
  * `const char *CRYPTO_get_sysgenid_path(void);`

### Call-outs:
New build flags added:
* `TEST_SYSGENID_PATH` -- a temporary file path that will be used for testing. The user executing the test must have permission read/write to this path and it must be created prior to running the tests.

### Testing:
* Added new "SnapsafeGenerationTest" test-suite. The test logic differs by platform.


crypto/fipsmodule/rand/rand.c
```diff
@@ -395,7 +403,10 @@ void RAND_bytes_with_additional_data(uint8_t *out, size_t out_len,
     // entropy is used. This can be expensive (one read per |RAND_bytes| call)
     // and so is disabled when we have fork detection, or if the application has
     // promised not to fork.
-    if (fork_generation != 0 || fork_unsafe_buffering) {
+    // snapsafe_status is only 0 when the kernel has snapsafe support, but it
+    // failed to initialize. Otherwise, snapsafe_status is 1.
+    if ((snapsafe_status != 0 && fork_generation != 0) ||
+        fork_unsafe_buffering) {
       OPENSSL_memset(additional_data, 0, sizeof(additional_data));
```
andrewhop, line 410: I realize you didn't add this but you are just continuing the pattern: why the memset 0? Even if we have fork protection or snapsafe why not just mix in whatever uninitialized value was in memory?

crypto/fipsmodule/rand/rand.c
```diff
@@ -539,6 +554,12 @@ void RAND_bytes_with_additional_data(uint8_t *out, size_t out_len,
 
   OPENSSL_cleanse(additional_data, 32);
 
+  CRYPTO_get_snapsafe_generation(&snapsafe_generation_id);
```
andrewhop, line 557: Should this check the return value of CRYPTO_get_snapsafe_generation?

crypto/fipsmodule/rand/snapsafe_detect.c
```diff
@@ -0,0 +1,153 @@
...
+DEFINE_BSS_GET(const char *, sgc_file_path)
+DEFINE_BSS_GET(volatile uint32_t *, sgc_addr)
+DEFINE_BSS_GET(int, snapsafety_state)
+
+/* aws_snapsafe_check_kernel_support returns 1 if the special sysgenid device
+ * file exists and 0 otherwise. */
+static int aws_snapsafe_check_kernel_support(void) {
+  /* This file-exist method is generally brittle. But for our purpose, this
+   * should be more than fine. */
+  if (access(*sgc_file_path_bss_get(), F_OK) != 0) {
```
andrewhop, line 30: Should this be `R_OK` in the unlikely case it exists but we can't access it?

crypto/fipsmodule/rand/snapsafe_detect.c
```diff
@@ -0,0 +1,153 @@
...
+
+  /* While not totally obvious from man pages, mmap actually only allocates
+   * memory on page size boundaries. So just give it that hint. */
+  long page_size = sysconf(_SC_PAGESIZE);
+  if (page_size <= 0) {
+    *snapsafety_state_bss_get() = SNAPSAFETY_STATE_FAILED_INITIALISE;
+    return;
+  }
+
+  if (*sgc_addr_bss_get() == NULL) {
```
andrewhop, line 58: Isn't this always true since this function sets it to null on line 43? Does it make sense to check if sgc_addr_bss_get is null at the start of the function so this is a lazy init that only happens once?

crypto/fipsmodule/rand/snapsafe_detect.c
```diff
@@ -0,0 +1,153 @@
...
+    *sgc_addr_bss_get() = addr;
+
+    *snapsafety_state_bss_get() = SNAPSAFETY_STATE_SUCCESS_INITIALISE;
+  }
+}
+
+static uint32_t aws_snapsafe_read_sgn(void) {
+  if (*snapsafety_state_bss_get() == SNAPSAFETY_STATE_SUCCESS_INITIALISE) {
+    return **sgc_addr_bss_get();
+  }
```
andrewhop, line 92: Just double checking are we reading the right part of the mmap data? From the documentation:
> The driver supports ``PROT_READ, MAP_SHARED`` mmaps of a single page
> in size. The first 4 bytes of the mapped page will contain an
>  up-to-date u32 copy of the system generation counter.

So the pointer points to the page of memory and the counter is at address -> address + 3, not something like an endian issue and the counter is on the other side of the page.

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,121 @@
...
+
+typedef struct sgn_test_s {
+  void* addr;
+  size_t pgsize;
+} sgn_test_s;
+
+int init_sgn_file(sgn_test_s* sgn_test);
+int init_sgn_file(sgn_test_s* sgn_test) {
+  const char* sgc_file_path = AWSLC_SYSGENID_PATH;
+  int fd_sgn = open(sgc_file_path, O_CREAT | O_RDWR | O_APPEND, (mode_t)0600);
```
andrewhop, line 25: What is 0600? Cna that use any of the other flag values to make it clearer?

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,121 @@
...
+    return 0;
+  }
+  return 1;
+}
+
+TEST(SnapsafeGenerationTest, SysGenIDretrievalTesting) {
+  sgn_test_s sgn_test;
+  if(1 != init_sgn_file(&sgn_test)) {
+    FAIL();
+  }
```
andrewhop, line 69: ASSERTS will fail and stop executing the test and is our normal test convention.
```suggestion
  ASSERT_TRUE(init_sgn_file(&sgn_test)));
```

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,121 @@
...
+    0x1000005, // 2^24 + 5
+    0xFFFFFFFF // 2^32 - 1
+  };
+
+  for (size_t i = 0; i < NUMBER_OF_TEST_VALUES; i++) {
+    // Exercise all bytes of the 32-bit generation number.
+    unsigned int new_sysgenid_value_hint = test_sysgenid_values[i];
+    ASSERT_TRUE(set_sgn(&sgn_test, new_sysgenid_value_hint));
+    ASSERT_TRUE(CRYPTO_get_snapsafe_generation(&current_snapsafe_gen_num));
+    ASSERT_EQ(new_sysgenid_value_hint, current_snapsafe_gen_num);
```
andrewhop, line 96: NP: this could use an EXPECT_EQ so all values would be executed and you could see if it's always wrong or just wrong for some of them:
```suggestion
    EXPECT_EQ(new_sysgenid_value_hint, current_snapsafe_gen_num);
```

crypto/fipsmodule/rand/snapsafe_detect.c
justsmth: For this function, the goal is to determine whether the `/dev/sysgenid` exists regardless of whether we are able to fully access it. We take the existence of `/dev/sysgenid` as implying that the system is capable of "snapshots", and thus we should take extra precaution w.r.t. entropy collection.

crypto/fipsmodule/rand/rand.c
justsmth: Good point.  We only really care about this when snapsafe detection is active.  (Although when it's inactive, the generation value wouldn't change anyways.) I'll make this logic also conditioned on `CRYPTO_get_snapsafe_generation` returning 1.

crypto/fipsmodule/rand/rand.c
justsmth: The memset to 0 is done to avoid collecting extra data/entropy via a slow syscall/rdrand invocation unless it's required. So this change essentially makes it required when snapsafe is supported but not initialized properly.

crypto/fipsmodule/rand/snapsafe_detect.c
justsmth: As long as we're reading the first (i.e., smallest-address) 4 bytes of the page and interpreting them as an (endian-appropriate) u32 value, the logic should be correct.

crypto/fipsmodule/rand/snapsafe_detect_test.cc
justsmth: I can change it to [`S_IRUSR | S_IWUSR`](https://en.cppreference.com/w/cpp/filesystem/perms) instead.

crypto/fipsmodule/rand/rand.c
andrewhop: Well that's just what this branch does, you could avoid collecting slow system entropy *and* avoid calling memset. Nothing I see in the rest of this function depends on this additional_data being all 0 if there is no system entropy to collect.

BUILDING.md
```diff
@@ -222,3 +222,15 @@ range of unit tests, as well as running valgrind and SDE tests. Building without
...
 More information on this can be found in [INCORPORATING.md](/INCORPORATING.md).
+
+# Snapsafe Detection
+
+AWS-LC supports Snapsafe-type uniqueness breaking event detection 
+on Linux using SysGenID (https://lkml.org/lkml/2021/3/8/677). This mechanism 
+is used for security hardening. If a SysGenID interface is not found, then the 
+mechanism is ignored. 
+
+## Snapsafety Prerequisites
```
torben-hansen, line 233:
```suggestion
## Snapsafe Prerequisites
```

CMakeLists.txt
```diff
@@ -91,6 +91,22 @@ install(DIRECTORY include/openssl
...
+  add_definitions(-DAWSLC_SNAPSAFE_TESTING=1)
+  message(STATUS "Setting AWSLC_SYSGENID_PATH=${TEST_SYSGENID_PATH}")
+  add_definitions(-DAWSLC_SYSGENID_PATH=\"${TEST_SYSGENID_PATH}\")
+
+  if (SYSGENID_PATH)
+    message(FATAL_ERROR "Cannot set both TEST_SYSGENID_PATH and SYGENID_PATH")
+  endif()
+endif()
+
+if (SYSGENID_PATH)
```
torben-hansen, line 105: Is this necessary? You can define `AWSLC_SYSGENID_PATH` in your build if you want to.

crypto/fipsmodule/bcm.c
```diff
@@ -128,6 +128,7 @@
 #include "pbkdf/pbkdf.c"
 #include "rand/ctrdrbg.c"
 #include "rand/fork_detect.c"
+#include "rand/snapsafe_detect.c"
```
torben-hansen, line 131: Order alphabetically

util/build_compilation_database.sh
```diff
@@ -12,7 +12,7 @@ MY_CMAKE_FLAGS=("-GNinja" "-DCMAKE_BUILD_TYPE=Debug" "-DCMAKE_EXPORT_COMPILE_COM
 
 mkdir -p "${AWS_LC_BUILD}"
 
-cmake "${BASE_DIR}" -B "${AWS_LC_BUILD}" ${MY_CMAKE_FLAGS[@]}
+cmake "${BASE_DIR}" -B "${AWS_LC_BUILD}" ${MY_CMAKE_FLAGS[@]} ${@}
```
torben-hansen, line 15: looks a bit unrelated?

util/all_tests.json
```diff
@@ -51,6 +51,13 @@
     "skip_valgrind": true,
     "target_arch": "x86"
   },
+  {
+    "comment": "Run snapsafe detection test suite with fork detection disabled and without Intel RDRAND",
+    "cmd": ["crypto/crypto_test", "--fork_unsafe_buffering", "--gtest_filter=SnapsafeGenerationTest.*"],
+    "env": ["OPENSSL_ia32cap=~0x4000000000000000"],
+    "skip_valgrind": true,
+    "target_arch": "x86"
```
torben-hansen, line 59: should try Arm as well

tests/ci/run_posix_tests.sh
```diff
@@ -39,7 +39,14 @@ fi
...
 # Lightly verify that uncommon build options does not break the build. Fist
 # define a list of typical build options to verify the special build option with
-build_options_to_test=("" "-DBUILD_SHARED_LIBS=1" "-DCMAKE_BUILD_TYPE=Release" "-DBUILD_SHARED_LIBS=1 -DCMAKE_BUILD_TYPE=Release")
+TEST_SYSGENID_PATH=$(mktemp)
+build_options_to_test=(
+    ""
+    "-DBUILD_SHARED_LIBS=1"
+    "-DCMAKE_BUILD_TYPE=Release"
+    "-DBUILD_SHARED_LIBS=1 -DCMAKE_BUILD_TYPE=Release"
+    "-DTEST_SYSGENID_PATH='${TEST_SYSGENID_PATH}'"
```
torben-hansen, line 48: does this pass through memory testers?

crypto/fipsmodule/rand/rand.c
```diff
@@ -100,6 +100,10 @@ struct rand_thread_state {
   // fork-unsafe buffering was enabled.
   int fork_unsafe_buffering;
 
+  // snapsafe_generation_id is non-zero when active. When the value changes,
+  // entropy must be collected.
+  uint32_t snapsafe_generation_id;
```
torben-hansen, line 105: could replicate `fork_generation` and simply name it `snapsafe_generation`?

crypto/fipsmodule/rand/rand.c
```diff
@@ -100,6 +100,10 @@ struct rand_thread_state {
   // fork-unsafe buffering was enabled.
   int fork_unsafe_buffering;
 
+  // snapsafe_generation_id is non-zero when active. When the value changes,
+  // entropy must be collected.
```
torben-hansen, line 104:
```suggestion
  // the drbg state must be reseeded.
```

crypto/fipsmodule/rand/rand.c
```diff
@@ -539,6 +553,14 @@ void RAND_bytes_with_additional_data(uint8_t *out, size_t out_len,
 
   OPENSSL_cleanse(additional_data, 32);
 
+  if (1 == CRYPTO_get_snapsafe_active()) {
```
torben-hansen, line 556: Hmm this could be replaced by:
```suggestion
  if (1 == CRYPTO_get_snapsafe_generation(&snapsafe_generation_id)) {
```
and just always do the test. If snapsafe is not supported in runtime, the test will always pass.

Current impl saves a few cycles, which I guess is your motivation?

crypto/fipsmodule/rand/rand.c
```diff
@@ -539,6 +553,14 @@ void RAND_bytes_with_additional_data(uint8_t *out, size_t out_len,
...
+    CRYPTO_get_snapsafe_generation(&snapsafe_generation_id);
+    if (snapsafe_generation_id != state->snapsafe_generation_id) {
+      // Unexpected change to snapsafe generation id.
```
torben-hansen, line 559: maybe needs a bit of elaboration.

torben-hansen (commented):
Still need to review `crypto/fipsmodule/rand/snapsafe_*`

tests/ci/run_posix_tests.sh
torben-hansen: In fact, this doesn't actually run any tests AFAIK

justsmth: I think all of Linux-based our CodeBuild jobs use it.

crypto/fipsmodule/rand/snapsafe_detect.c
```diff
@@ -0,0 +1,140 @@
...
+#include <unistd.h>
+#include "../delocate.h"
+
+/* Snapsafety state */
+#define SNAPSAFETY_STATE_FAILED_INITIALISE 0x00
+#define SNAPSAFETY_STATE_SUCCESS_INITIALISE 0x01
+#define SNAPSAFETY_STATE_NOT_SUPPORTED 0x02
+
+DEFINE_STATIC_ONCE(aws_snapsafe_init)
+DEFINE_BSS_GET(const char *, sgc_file_path)
```
andrewhop, line 21: Does this need to be a variable? Can the two places that use the file path use AWSLC_SYSGENID_PATH directly?

andrewhop (commented):
Core change looks good to me, ship it pending urandom test fix.

crypto/fipsmodule/rand/snapsafe_detect.c
justsmth: Good point.  There's no need for a variable.

CMakeLists.txt
justsmth: Ok, I'll remove this build option.

util/all_tests.json
justsmth: Ok, I'll just remove the `target_arch`.

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,119 @@
...
+    ASSERT_TRUE(set_sgn(&sgn_test, new_sysgenid_value_hint));
+    ASSERT_TRUE(CRYPTO_get_snapsafe_generation(&current_snapsafe_gen_num));
+    EXPECT_EQ(new_sysgenid_value_hint, current_snapsafe_gen_num);
+  }
+}
+#elif defined(OPENSSL_LINUX)
+TEST(SnapsafeGenerationTest, SysGenIDretrievalLinux) {
+  uint32_t current_snapsafe_gen_num = 0xffffffff;
+  ASSERT_TRUE(CRYPTO_get_snapsafe_generation(&current_snapsafe_gen_num));
+  if(CRYPTO_get_snapsafe_supported()) {
```
torben-hansen, line 101:
```suggestion
  if (CRYPTO_get_snapsafe_supported()) {
```

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,119 @@
...
+  }
+  return 1;
+}
+
+TEST(SnapsafeGenerationTest, SysGenIDretrievalTesting) {
+  sgn_test_s sgn_test;
+  ASSERT_TRUE(init_sgn_file(&sgn_test));
+
+  if(1 != set_sgn(&sgn_test, 0)) {
+    FAIL();
```
torben-hansen, line 70: can't use `ASSERT_TRUE`?

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  ASSERT_TRUE(CRYPTO_get_snapsafe_supported());
+  ASSERT_TRUE(CRYPTO_get_snapsafe_active());
+
+  uint32_t current_snapsafe_gen_num = 0;
+  ASSERT_TRUE(set_sgn(&sgn_test, 7));
+  ASSERT_TRUE(CRYPTO_get_snapsafe_generation(&current_snapsafe_gen_num));
+  ASSERT_EQ((unsigned int)7, current_snapsafe_gen_num);
```
torben-hansen, line 79:
```suggestion
  ASSERT_EQ((unsigned int) 7, current_snapsafe_gen_num);
```

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,119 @@
...
+
+#define NUMBER_OF_TEST_VALUES 5
+
+typedef struct sgn_test_s {
+  void* addr;
+  size_t pgsize;
+} sgn_test_s;
+
+int init_sgn_file(sgn_test_s* sgn_test);
+int init_sgn_file(sgn_test_s* sgn_test) {
```
torben-hansen, line 23: can be static.

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,119 @@
...
+  close(fd_sgn);
+
+  sgn_test->addr = addr;
+  sgn_test->pgsize = pgsize;
+
+  return 1;  
+}
+
+int set_sgn(const sgn_test_s* sgn_test, uint32_t val);
+int set_sgn(const sgn_test_s* sgn_test, uint32_t val) {
```
torben-hansen, line 57: static?

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,119 @@
...
+typedef struct sgn_test_s {
+  void* addr;
+  size_t pgsize;
+} sgn_test_s;
+
+int init_sgn_file(sgn_test_s* sgn_test);
+int init_sgn_file(sgn_test_s* sgn_test) {
+  const char* sgc_file_path = AWSLC_SYSGENID_PATH;
+  int fd_sgn = open(sgc_file_path, O_CREAT | O_RDWR | O_APPEND, S_IRUSR | S_IWUSR);
+  if (fd_sgn == -1) {
```
torben-hansen, line 26: can use the `ASSERT_` and `EXPECT_` in these utility functions as well.

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,119 @@
...
+    return 0;
+  }
+  if (0 != lseek(fd_sgn, 0, SEEK_SET)) {
```
torben-hansen, line 29: missing newline before if-statement.

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,119 @@
...
+  
+  uint32_t set_sgn = 0;
+  if(0 >= write(fd_sgn, &set_sgn, sizeof(unsigned int))) {
+    return 0;
+  }
+  long page_size = sysconf(_SC_PAGESIZE);
+  if (page_size <= 0) {
+    return 0;
+  }
+  size_t pgsize = (size_t)page_size;
```
torben-hansen, line 41:
```suggestion
  size_t pgsize = (size_t) page_size;
```

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,119 @@
...
+  int fd_sgn = open(sgc_file_path, O_CREAT | O_RDWR | O_APPEND, S_IRUSR | S_IWUSR);
+  if (fd_sgn == -1) {
+    return 0;
+  }
+  if (0 != lseek(fd_sgn, 0, SEEK_SET)) {
+    return 0;
+  }
+  
+  uint32_t set_sgn = 0;
+  if(0 >= write(fd_sgn, &set_sgn, sizeof(unsigned int))) {
```
torben-hansen, line 34: just a suggestion: could zeroise the entire memory area

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,119 @@
...
+    return 0;
+  }
+  size_t pgsize = (size_t)page_size;
+
+  void* addr = mmap(NULL, pgsize, PROT_WRITE, MAP_SHARED, fd_sgn, 0);
+  if (addr == MAP_FAILED) {
+    return 0;
+  }
+
+  close(fd_sgn);
```
torben-hansen, line 48: not an issue for the testing. But the file descriptor leaks if the four previous if-statements are true.

torben-hansen: Our static analyser might complain after merging this in though...

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,119 @@
...
+
+#if defined(OPENSSL_LINUX) && defined(AWSLC_SNAPSAFE_TESTING)
+#include <fcntl.h>
+#include <cstring>
+#include <sys/mman.h>
+
+#define NUMBER_OF_TEST_VALUES 5
+
+typedef struct sgn_test_s {
+  void* addr;
```
torben-hansen, line 18:
```suggestion
  void *addr;
```

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,119 @@
...
+  sgn_test->addr = addr;
+  sgn_test->pgsize = pgsize;
+
+  return 1;  
+}
+
+int set_sgn(const sgn_test_s* sgn_test, uint32_t val);
+int set_sgn(const sgn_test_s* sgn_test, uint32_t val) {
+  memcpy(sgn_test->addr, &val, sizeof(unsigned int));
+  if(0 != msync(sgn_test->addr, sgn_test->pgsize, MS_SYNC)) {
```
torben-hansen, line 59: a'ha I think I missed this step back when I tried making this file-mocking method work....

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,119 @@
...
+
+  sgn_test->addr = addr;
+  sgn_test->pgsize = pgsize;
+
+  return 1;  
+}
+
+int set_sgn(const sgn_test_s* sgn_test, uint32_t val);
+int set_sgn(const sgn_test_s* sgn_test, uint32_t val) {
+  memcpy(sgn_test->addr, &val, sizeof(unsigned int));
```
torben-hansen, line 58:
```suggestion
  memcpy(sgn_test->addr, &val, sizeof(uint32_t));
```

crypto/fipsmodule/rand/rand.c
```diff
@@ -100,6 +101,10 @@ struct rand_thread_state {
   // fork-unsafe buffering was enabled.
   int fork_unsafe_buffering;
 
+  // snapsafe_generation_id is non-zero when active. When the value changes,
```
torben-hansen, line 104:
```suggestion
  // snapsafe_generation is non-zero when active. When the value changes,
```

crypto/fipsmodule/rand/snapsafe_detect_test.cc
justsmth: Will fix.

crypto/fipsmodule/rand/snapsafe_detect_test.cc
justsmth: But if we do that we would be unable to cleanup resources, e.g., close the file. I could use `FAIL()` instead of returning 0?

torben-hansen: OK that makes sense.

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,118 @@
...
+    return 0;
+  }
+
+  if (0 != fsync(fd_sgn)) {
+    return 0;
+  }
+
+  void* my_addr = mmap(nullptr, sizeof(uint32_t), PROT_WRITE, MAP_SHARED, fd_sgn, 0);
+  if (my_addr == MAP_FAILED) {
+    return 0;
```
WillChilds-Klein, line 42: nit: `close(fd_sgn)` before returning?

crypto/fipsmodule/rand/rand.c
```diff
@@ -382,6 +387,9 @@ void RAND_bytes_with_additional_data(uint8_t *out, size_t out_len,
   const uint64_t fork_generation = CRYPTO_get_fork_generation();
   const int fork_unsafe_buffering = rand_fork_unsafe_buffering_enabled();
 
+  uint32_t snapsafe_generation = 0;
+  int snapsafe_status = CRYPTO_get_snapsafe_generation(&snapsafe_generation);
```
WillChilds-Klein, line 391: what if this is 0 after the call to `CRYPTO_get_snapsafe_generation`? is that a case we need to handle as on L558 or retry?

crypto/fipsmodule/rand/snapsafe_detect_test.cc
```diff
@@ -0,0 +1,118 @@
...
+  if (fd_sgn == -1) {
+    return 0;
+  }
+
+  if (0 != lseek(fd_sgn, 0, SEEK_SET)) {
+    close(fd_sgn);
+    return 0;
+  }
+
+  if (0 != fsync(fd_sgn)) {
```
WillChilds-Klein, line 36: nit: why call `fsync` here? we haven't buffered any data

justsmth: Good point. Will remove.

crypto/fipsmodule/rand/rand.c
justsmth: It's expected that `snapsafe_generation` will likely be 0 after the call to `CRYPTO_get_snapsafe_generation` -- in fact, that'll be true for any environment that doesn't have the `/dev/sysgenid` setup in the kernel.  What we care about is whether the `snapsafe_generation` has *changed* since the last time we re-seeded the DRBG.

crypto/fipsmodule/rand/snapsafe_detect.h
```diff
@@ -0,0 +1,59 @@
...
+// SysGenID interface. Otherwise, it returns 0.
+OPENSSL_EXPORT int CRYPTO_get_snapsafe_supported(void);
+
+// CRYPTO_get_sysgenid_path returns the path used for the SysGenId interface.
+OPENSSL_EXPORT const char *CRYPTO_get_sysgenid_path(void);
+
+#if defined(OPENSSL_LINUX) && defined(AWSLC_SNAPSAFE_TESTING)
+// HAZMAT_init_sysgenid_file should only be used for testing. It creates and
+// initializes the sysgenid path indicated by AWSLC_SYSGENID_PATH.
+// On success, it returns 1.  Otherwise, returns 0.
```
torben-hansen, line 51:
```suggestion
// On success, it returns 1. Otherwise, returns 0.
```

crypto/internal.h
```diff
@@ -115,6 +115,8 @@
 #include <openssl/stack.h>
 #include <openssl/thread.h>
 
+#include "fipsmodule/rand/snapsafe_detect.h"
```
torben-hansen, line 118: what's this needed for?

justsmth: I think it was so that gtest main could call the HAZMAT function.

crypto/fipsmodule/rand/rand.c
torben-hansen: Could still resolve this typo.

torben-hansen (approved):

torben-hansen (approved):

dkostic (approved):

PR #1642, opened 2024-06-18
Make DH_check consistent with OpenSSL

dkostic:
### Issues:
CryptoAlg-2490

### Description of changes: 
Previously, `DH_check` allowed only primes with certain
properties when the generator was equal to 2 or 5. We remove
this requirement to:
  - be more consistent with OpenSSL that doesn't have these
    checks. As a consequence, it was possible to generate DH
    parameters with OpenSSL that wouldn't pass AWS-LC `DH_check`.
  - not break customers who use one of the standard DH groups
    from RFC's 3526 and 7919 and call `DH_check` on it (like s2n-tls).

### Call-outs:

### Testing:

### DIFF
--- a/crypto/fipsmodule/dh/check.c
+++ b/crypto/fipsmodule/dh/check.c
@@ -150,7 +150,6 @@ int DH_check(const DH *dh, int *out_flags) {
   // should hold.
   int ok = 0, r, q_good = 0;
   BN_CTX *ctx = NULL;
-  BN_ULONG l;
   BIGNUM *t1 = NULL, *t2 = NULL;
 
   ctx = BN_CTX_new();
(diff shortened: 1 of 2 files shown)


crypto/fipsmodule/dh/check.c
```diff
@@ -203,24 +202,6 @@ int DH_check(const DH *dh, int *out_flags) {
     if (!BN_is_one(t2)) {
       *out_flags |= DH_CHECK_INVALID_Q_VALUE;
     }
-  } else if (BN_is_word(dh->g, DH_GENERATOR_2)) {
-    l = BN_mod_word(dh->p, 24);
```
darylmartin100, line 207: You should update the documentation at the top of this function as I think it's out of date now.  I would also add a small note (that you put in the PR) in the header file to explain why the checks are they way they are.

crypto/fipsmodule/dh/check.c
```diff
@@ -203,24 +202,6 @@ int DH_check(const DH *dh, int *out_flags) {
...
-    }
-    if (l != 11) {
-      *out_flags |= DH_CHECK_NOT_SUITABLE_GENERATOR;
-    }
-  } else if (BN_is_word(dh->g, DH_GENERATOR_5)) {
-    l = BN_mod_word(dh->p, 10);
-    if (l == (BN_ULONG)-1) {
-      goto err;
-    }
-    if (l != 3 && l != 7) {
```
WillChilds-Klein, line 219: what's significant about these values or 11 (mod 24) for `g=2`? why did we previously restrict `p` to these values for relevant `g`?

WillChilds-Klein: Details [here](https://wiki.openssl.org/index.php/Diffie-Hellman_parameters#Validating_Parameters).

WillChilds-Klein (approved):

skmcgrail (approved):

lilpoozie2005 (approved):

PR #1643, opened 2024-06-18
Patch for OpenVPN certificate setting behavioral difference

smittals2:
### Description of changes: 
SSL_CTX_set_client_CA_list in AWS-LC creates a copy and frees the passed in parameter while OpenSSL stores a reference and leaves the passed in parameter untouched. The function contract states it is taking ownership of the passed in parameter and consuming applications should not expect to have continued access to the passed in parameter. However, OpenVPN code accesses the parameter immediately after an invocation to SSL_CTX_set_client_CA_list. This worked when building with OpenSSL but in AWS-LC it does not. 

Changes: 
1. A small patch to account for this functionality difference in OpenVPN code. 
2. Changes to the documentation for SSL_CTX_set_client_CA_list & SSL_set_client_CA_list to make their behavior clearer
3. A callout in the porting guide

### Testing:
N/A


andrewhop (approved):

tests/ci/integration/openvpn_patch/aws-lc-openvpn-cert.patch
samuel40791765, line 1: Might be a dumb question, but does this patch work with OpenSSL? If it does, maybe we could consider contributing it upstream, seems like a relatively simple change they'd be willing to take.

samuel40791765 (approved):

tests/ci/integration/openvpn_patch/aws-lc-openvpn-cert.patch
smittals2: Yes this would work with OpenSSL as well and the plan is to upstream. I've added it in for now while we start the process and get them to take it!

lilpoozie2005 (approved):

PR #1644, opened 2024-06-19
NIST SP 800-108r1-upd1: KDF Counter Implementation

skmcgrail:
### Description of changes: 
This pull request implements the KDF in Counter Mode
defined in [Section 4 of NIST.SP.800-108r1-upd1](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-108r1-upd1.pdf#%5B%7B%22num%22%3A77%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C70%2C300%2C0%5D)

The abbreviation `KBKDF` stands for Key-based key derivation function. `ctr` is abbreviation for counter.

### Call-outs:
* We will need to add appropriate service indicator logic ahead of our
next FIPS certification round.
* We will need to wire-up ACVP tests separately.

### Testing:
Test vectors are provided [here](https://github.com/aws/aws-lc/pull/1644/files#diff-d0cd06e99fdc733df70e2aa730e973c5ed0ab73ff851c33b1f2e5f44beb8a82d)

By submitting this pull request, I confirm that my contribution is made
under the terms of the Apache 2.0 license and the ISC license.


skmcgrail:
As a note for the reviewers. I'm still on the fence with the proposed `KBKDF_ctr` for the one-shot function name. It may be clearer to call it `KBKDF_counter` or even `KBKDF_hmac_counter` (as the PRF function could be implemented with KMAC if we so ever wished). Open to suggestions / discussions around the name choice.

include/openssl/kdf.h
```diff
@@ -85,6 +85,27 @@ OPENSSL_EXPORT int SSKDF_hmac(uint8_t *out_key, size_t out_len,
...
+// using the provided key derivation key |secret| and fixed info |info|.
+// |info| or |info_len| may be zero-length.
+//
+// This implementation adheres to the algorithm specified in Section 4.1 of the
+// NIST Special Publication 800-108 Revision 1 Update 1 published on August
+// 2022. The parameters relevant to the specification are as follows:
+// * |out_len|, |secret_len|, and |info_len| are specified in bytes
+// * |out_len| is analogous to |L| in the specification.
+// * |r| the length of the binary representation of the counter |i|
+// . referred to by the specification is 32-bit in this implementation.
```
nebeid, line 98:
```suggestion
// * |r| the length of the binary representation of the counter |i|
// referred to by the specification. |r| is 32 bits in this implementation.
```

crypto/fipsmodule/kdf/kbkdf.c
```diff
@@ -0,0 +1,92 @@
...
+    // When we are finished clear the temporary buffer to cleanse key material
+    // from stack.
+    if (done == out_len) {
+      OPENSSL_cleanse(&out_key_i[0], EVP_MAX_MD_SIZE);
+    }
+  }
+
+  ret = 1;
+
+err:
```
nebeid, line 89: I believe that in case of error, we should clear `out_key`.

crypto/fipsmodule/kdf/test/kbkdf_counter.txt
nebeid, line 1: what's the source of the test vectors here and in sskdf.txt?

skmcgrail: The ones here are sourced from our internal Java implementation, which you can find linked in the original design documented.

The ones that were used for the single-step implementation were sources from that same package, OpenSSL, and the vectors found in this third-party [project](https://github.com/patrickfav/singlestep-kdf) as an additional compatibility check.

include/openssl/kdf.h
```diff
@@ -85,6 +85,27 @@ OPENSSL_EXPORT int SSKDF_hmac(uint8_t *out_key, size_t out_len,
                               const uint8_t *info, size_t info_len,
                               const uint8_t *salt, size_t salt_len);
 
+// KBKDF_ctr derives keying material using the KDF counter mode algorithm,
```
nebeid, line 88: > As a note for the reviewers. I'm still on the fence with the proposed `KBKDF_ctr` for the one-shot function name. It may be clearer to call it `KBKDF_counter` or even `KBKDF_hmac_counter` (as the PRF function could be implemented with KMAC if we so ever wished). Open to suggestions / discussions around the name choice.

I think your suggestion to add "hmac" to the name is a good one, especially since OpenSSL has implementation for cmac as well. `KBKDF_ctr_hmac` is what I think of. A similar combination is used in [`EVP_aead_aes_128_ctr_hmac_sha256`](https://github.com/aws/aws-lc/blob/b2bbd9c287d3fff4e209892d342ea3b0d562477e/include/openssl/aead.h#L135).

Also, I suggest you expand the acronym to Key-Based Key Derivation Function or Key-Based KDF somewhere as was done with SSKDF.

nebeid (approved):

crypto/fipsmodule/kdf/kbkdf.c
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  if (!HMAC_Init_ex(hmac_ctx, secret, secret_len, digest, NULL)) {
+    goto err;
+  }
+
+  // Determine the length of the output in bytes of a single invocation of the
+  // HMAC function.
+  size_t h_output_bytes = HMAC_size(hmac_ctx);
+  if (!h_output_bytes || h_output_bytes <= 0) {
```
dkostic, line 30: `!h_output_bytes` is superficial here, `h_output_bytes <= 0` covers that case too

crypto/fipsmodule/kdf/kbkdf.c
```diff
@@ -0,0 +1,95 @@
...
+    goto err;
+  }
+
+  // NIST.SP.800-108r1-upd1: Step 1:
+  // Determine how many output chunks are required to produce the requested
+  // output length |out_len|. This determines how many times the variant compute
+  // function will be called to output key material.
+  uint64_t n = (out_len + h_output_bytes - 1) / h_output_bytes;
```
dkostic, line 38: just curious, `out_len` and `h_output_bytes` are `size_t`. If `size_t` is 32 bits wide, when will cast to `uint64_t` happen?

crypto/fipsmodule/kdf/sskdf.c
```diff
@@ -278,6 +278,9 @@ static int SSKDF(const sskdf_variant *variant, sskdf_variant_ctx *ctx,
   ret = 1;
 
 err:
+  if(ret <= 0) {
+    OPENSSL_cleanse(out_key, out_len);
+  }
```
dkostic, line 283: looks like this is seg-faulting the unit test?

crypto/fipsmodule/kdf/kbkdf.c
skmcgrail: Fixed this to compute `n` as a size_t. Then the `uint32_t i` in the for loop will be lifted through implicit conversion when comparing to `n` to size_t for the loop condition expression.

crypto/fipsmodule/kdf/kbkdf.c
```diff
@@ -27,15 +27,15 @@ int KBKDF_ctr_hmac(uint8_t *out_key, size_t out_len, const EVP_MD *digest,
   // Determine the length of the output in bytes of a single invocation of the
   // HMAC function.
   size_t h_output_bytes = HMAC_size(hmac_ctx);
-  if (!h_output_bytes || h_output_bytes <= 0) {
+  if (h_output_bytes <= 0) {
```
nebeid, line 30: can `h_output_bytes` be `< 0` being unsigned?

crypto/fipsmodule/kdf/kbkdf.c
dkostic: I think that won't work for a 32-bit build where `size_t` is 32 bits wide. For example, `out_len` might be equal to `UINT32_MAX`, and `n` won't be computed correctly then (and also the condition `n > UINT32_MAX` wouldn't do anything). I think `n` must be `uint64_t`, I was just wondering if we have to explicitly cast to `uint64_t` any of the variables in the expression:
```
(out_len + h_output_bytes - 1) / h_output_bytes;
```

also, maybe worth thinking about adding a test case for this.

nebeid (approved):

lilpoozie2005 (approved):

PR #1645, opened 2024-06-19
add support for EC_POINT_bn2point

samuel40791765:
### Issues:
Addresses `CryptoAlg-1715`

### Description of changes: 
Ruby happens to [consume this](https://github.com/ruby/ruby/blob/ruby_3_1/ext/openssl/ossl_pkey_ec.c#L1167-L1171). This is deprecated in OpenSSL 3.0 and I've marked it likewise in AWS-LC as well. Although Ruby has generally moved off of deprecated 3.0 APIs, they still depend on this and later versions like 3.2 and 3.3 have not removed support for this either.

### Call-outs:
N/A

### Testing:
Tests built upon original tests for `EC_POINT_point2bn`.

### DIFF
--- a/crypto/ec_extra/ec_asn1.c
+++ b/crypto/ec_extra/ec_asn1.c
@@ -614,3 +614,44 @@ BIGNUM *EC_POINT_point2bn(const EC_GROUP *group, const EC_POINT *point,
 
   return ret;
 }
+
+EC_POINT *EC_POINT_bn2point(const EC_GROUP *group,
+                            const BIGNUM *bn, EC_POINT *point, BN_CTX *ctx) {
+  EC_POINT *ret = NULL;
+
+  // Allocate buffer and length.
+  size_t buf_len = BN_num_bytes(bn);
+  if (buf_len == 0) {
+    buf_len = 1;
+  }
+  uint8_t *buf = OPENSSL_malloc(buf_len);
+  if (buf == NULL) {
+    return NULL;
+  }
+
+  if (BN_bn2bin_padded(buf, buf_len,bn) < 0) {
+    goto end;
+  }
+
+  // Allocate new |EC_POINT| if |point| is NULL. Otherwise, use |point|.
+  if (point == NULL) {
+    ret = EC_POINT_new(group);
+    if (ret == NULL) {
+      goto end;
+    }
+  } else {
+    ret = point;
+  }
+
+  if (!EC_POINT_oct2point(group, ret, buf, buf_len, ctx)) {
+    if (ret != point) {
+      // Free the newly allocated |EC_POINT| on failure.
+      EC_POINT_free(ret);
+    }
+    goto end;
+  }
+
+end:
+  OPENSSL_free(buf);
+  return ret;
+}
--- a/crypto/ec_extra/ec_asn1.c
+++ b/crypto/ec_extra/ec_asn1.c
@@ -617,10 +617,15 @@ BIGNUM *EC_POINT_point2bn(const EC_GROUP *group, const EC_POINT *point,
 
 EC_POINT *EC_POINT_bn2point(const EC_GROUP *group, const BIGNUM *bn,
                             EC_POINT *point, BN_CTX *ctx) {
+  if (group == NULL || bn == NULL) {
+    OPENSSL_PUT_ERROR(EC, ERR_R_PASSED_NULL_PARAMETER);
+    return NULL;
+  }
+
   // Allocate buffer and length.
   size_t buf_len = BN_num_bytes(bn);
   if (buf_len == 0) {
-    // See https://github.com/openssl/openssl/issues/10258
+    // See https://github.com/openssl/openssl/issues/10258.
     buf_len = 1;
   }
   uint8_t *buf = OPENSSL_malloc(buf_len);
--- a/crypto/ec_extra/ec_asn1.c
+++ b/crypto/ec_extra/ec_asn1.c
@@ -615,13 +615,12 @@ BIGNUM *EC_POINT_point2bn(const EC_GROUP *group, const EC_POINT *point,
   return ret;
 }
 
-EC_POINT *EC_POINT_bn2point(const EC_GROUP *group,
-                            const BIGNUM *bn, EC_POINT *point, BN_CTX *ctx) {
-  EC_POINT *ret = NULL;
-
+EC_POINT *EC_POINT_bn2point(const EC_GROUP *group, const BIGNUM *bn,
+                            EC_POINT *point, BN_CTX *ctx) {
   // Allocate buffer and length.
   size_t buf_len = BN_num_bytes(bn);
   if (buf_len == 0) {
+    // See https://github.com/openssl/openssl/issues/10258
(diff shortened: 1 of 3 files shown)


include/openssl/ec.h
```diff
@@ -393,14 +393,21 @@ OPENSSL_EXPORT int EC_GROUP_set_generator(EC_GROUP *group,
...
-// point into the X9.62 form given by |form| then interpretting it as a BIGNUM.
+// point into the X9.62 form given by |form| then interpreting it as a BIGNUM.
 // On success, it returns the BIGNUM pointer supplied or, if |ret| is NULL,
 // allocates and returns a fresh |BIGNUM|. On error, it returns NULL. The |ctx|
 // argument may be used if not NULL.
 OPENSSL_EXPORT OPENSSL_DEPRECATED BIGNUM *EC_POINT_point2bn(
     const EC_GROUP *group, const EC_POINT *point, point_conversion_form_t form,
     BIGNUM *ret, BN_CTX *ctx);
 
+// EC_POINT_bn2point is like |EC_POINT_point2bn|, but converts a |BIGNUM| to an
```
dkostic, line 404: this doesn't make sense (nor does the EC_POINT_point2bn documentation :)). An EC point consists of (at least) two coordinates x and y, which are usually of type BIGNUM. So what's the relation between `bn` and `point`?

crypto/ec_extra/ec_asn1.c
```diff
@@ -614,3 +614,44 @@ BIGNUM *EC_POINT_point2bn(const EC_GROUP *group, const EC_POINT *point,
...
   return ret;
 }
+
+EC_POINT *EC_POINT_bn2point(const EC_GROUP *group,
+                            const BIGNUM *bn, EC_POINT *point, BN_CTX *ctx) {
+  EC_POINT *ret = NULL;
+
+  // Allocate buffer and length.
+  size_t buf_len = BN_num_bytes(bn);
+  if (buf_len == 0) {
```
dkostic, line 624: why?? isn't this an error or something?

crypto/ec_extra/ec_asn1.c
```diff
@@ -614,3 +614,44 @@ BIGNUM *EC_POINT_point2bn(const EC_GROUP *group, const EC_POINT *point,
...
+    if (ret == NULL) {
+      goto end;
+    }
+  } else {
+    ret = point;
+  }
+
+  if (!EC_POINT_oct2point(group, ret, buf, buf_len, ctx)) {
+    if (ret != point) {
+      // Free the newly allocated |EC_POINT| on failure.
```
dkostic, line 648: I'm not sure I understand the logic around `ret`/`point`. can we implement that in a simpler way?

crypto/ec_extra/ec_asn1.c
samuel40791765: Good question, I found the issue/commit on why this was done, but forgot to provide the link: https://github.com/openssl/openssl/issues/10258

I added it as a comment in the code.

crypto/ec_extra/ec_asn1.c
samuel40791765: The API design around this wasn't the best. The user can provide an `EC_POINT` with `point` and use that to save the intended value. If it's not provided, OpenSSL/AWS-LC allocates a value for you. Basically we want to be careful not to free the `EC_POINT` if the user has provided one, but we should be freeing the one we allocated.

I rearranged a code a bit to make more readable, do lmk if it's still a bit confusing

include/openssl/ec.h
samuel40791765: I see what you're saying now, OpenSSL also deprecated these functions for [this reason](https://github.com/openssl/openssl/blob/b76a6c26a254b4cc428275fc0ced56759dd5088a/doc/man7/ossl-guide-migration.pod?plain=1#L1686-L1689).

I've reworded the function documentation to describe the actual relation (there's not really much of one, it's just another conversion of the serialized output).

crypto/ec_extra/ec_asn1.c
```diff
@@ -614,3 +614,47 @@ BIGNUM *EC_POINT_point2bn(const EC_GROUP *group, const EC_POINT *point,
 
   return ret;
 }
+
+EC_POINT *EC_POINT_bn2point(const EC_GROUP *group, const BIGNUM *bn,
+                            EC_POINT *point, BN_CTX *ctx) {
+  // Allocate buffer and length.
+  size_t buf_len = BN_num_bytes(bn);
```
dkostic, line 621: should we do NULL checks on function arguments before dereferencing?

crypto/ec_extra/ec_asn1.c
dkostic: can we add a test for this case?

include/openssl/ec.h
```diff
@@ -392,15 +392,31 @@ OPENSSL_EXPORT int EC_GROUP_set_generator(EC_GROUP *group,
                                           const BIGNUM *cofactor);
 
 
-// EC_POINT_point2bn converts an |EC_POINT| to a |BIGNUM| by serializing the
-// point into the X9.62 form given by |form| then interpretting it as a BIGNUM.
+// EC_POINT_point2bn serialises |point| into the X9.62 form given by |form|
```
dkostic, line 395: what is this format exactly? like if the point is `(x, y)` what should the serialized output be?

include/openssl/ec.h
```diff
@@ -392,15 +392,31 @@ OPENSSL_EXPORT int EC_GROUP_set_generator(EC_GROUP *group,
...
     const EC_GROUP *group, const EC_POINT *point, point_conversion_form_t form,
     BIGNUM *ret, BN_CTX *ctx);
 
+// EC_POINT_bn2point is like |EC_POINT_point2bn|, but takes the |BIGNUM|
+// representation and de-serialises it back to an |EC_POINT|. On success,
+// it returns the EC_POINT pointer supplied or, if |ret| is NULL, allocates and
+// returns a fresh |EC_POINT|. On error, it returns NULL. The |ctx| argument
+// may be used if not NULL.
+//
+// Note: |EC_POINT| serialization formats are not individual big-endian
```
dkostic, line 414: I still don't understand how is this supposed to work at all :)
A point is `(x, y)` with `x` and `y` big integers. So how do you get from `|bn|` to `|point|`?

crypto/ec_extra/ec_asn1.c
samuel40791765: Added

include/openssl/ec.h
dkostic: oh boy, I get it now. terrible API, but what to do... I like the updated documentation.

dkostic (approved):

include/openssl/ec.h
samuel40791765: Yeah everything about the function really throws people off unless you look into what it's actually doing... Thanks for iterating on this, I think it's in a much better state now.

smittals2 (approved):

lilpoozie2005 (approved):

PR #1665, opened 2024-06-25
Improve gcc-4.8 support/testing

justsmth:
### Description of changes: 
* Pre-generated assembly was not usable with certain older compilers/assemblers such as gcc-4.8.
  * This change injects preprocessor macros to guard the portions of ADX/AVX2/AVX-512 assembly that would break GCC-4.8, allowing GCC-4.8 to now use the pre-generated assembly.

### Call-out
* The pre-generated assembly is still not usable on the oldest supported compilers (e.g., gcc-4.1)

### Testing:
* Added tests that build using the pregenerated assembly code.

### DIFF
--- a/.github/docker_images/gcc-4.8/awslc_build.sh
+++ b/.github/docker_images/gcc-4.8/awslc_build.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+set -ex -o pipefail
+
+echo "Building with GCC Version: $(gcc --version)"
+
+BUILD_DIR=$(mktemp -d)
+SRC_DIR="${SRC_DIR:-/awslc}"
+
+pushd "${BUILD_DIR}"
+
+cmake "${SRC_DIR}" "-DDISABLE_GO=ON" "-DDISABLE_PERL=ON" "-DMY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX=1"
+make -j 4 ssl
+
+popd # ${BUILD_DIR}
--- a/crypto/fipsmodule/aes/asm/aesni-xts-avx512.pl
+++ b/crypto/fipsmodule/aes/asm/aesni-xts-avx512.pl
@@ -1460,7 +1460,10 @@ ___
 
   my $rndsuffix = &random_string();
 
-  $code .= ".text\n";
+  $code .= <<___;
+#ifndef MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX
+.text
+___
 
   {
   $code.=<<___;
@@ -3108,6 +3111,7 @@ ___
     .byte  0xff, 0xff, 0xff, 0xff, 0xff
 
 .text
+#endif
 ___
 } else {
     $code .= <<___;
--- a/crypto/fipsmodule/bn/asm/x86_64-mont.pl
+++ b/crypto/fipsmodule/bn/asm/x86_64-mont.pl
@@ -67,7 +67,7 @@ open OUT,"| \"$^X\" \"$xlate\" $flavour \"$output\"";
 # versions, but BoringSSL is intended to be used with pre-generated perlasm
 # output, so this isn't useful anyway.
 $addx = 1;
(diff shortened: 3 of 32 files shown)


crypto/fipsmodule/aes/asm/aesni-xts-avx512.pl
```diff
@@ -1460,7 +1460,10 @@
 
   my $rndsuffix = &random_string();
 
-  $code .= ".text\n";
+  $code .= <<___;
+#ifndef MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX
```
justsmth, line 1464: This is the start of code generation for the `if` block that started way up at [line 49](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/aes/asm/aesni-xts-avx512.pl#L49)

crypto/fipsmodule/modes/gcm.c
```diff
@@ -335,7 +337,7 @@ void CRYPTO_gcm128_setiv(GCM128_CONTEXT *ctx, const AES_KEY *key,
   ctx->ares = 0;
   ctx->mres = 0;
 
-#if defined(GHASH_ASM_X86_64)
+#if defined(GHASH_ASM_X86_64) && !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX)
```
dkostic, line 340: `GHASH_ASM_X86_64` itself is guarded here: https://github.com/aws/aws-lc/blob/88faaa28922a584b9df111cf2ead7f16fa6f6b41/crypto/fipsmodule/modes/internal.h#L296-L297

crypto/fipsmodule/bn/asm/x86_64-mont.pl
```diff
@@ -105,8 +105,10 @@
 	jb	.Lmul_enter
 ___
 $code.=<<___ if ($addx);
+#ifndef MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX
```
nebeid, line 108: I have similar questions here and in other files where `$addx` is controlled by the flag, what is the need to add the macros?

crypto/fipsmodule/aes/asm/aesni-xts-avx512.pl
```diff
@@ -1460,7 +1460,10 @@
 
   my $rndsuffix = &random_string();
 
-  $code .= ".text\n";
+  $code .= <<___;
```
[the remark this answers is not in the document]
justsmth: It's only needed so that the pre-generated assembly code is usable with GCC 4.8.x: https://github.com/aws/aws-lc/pull/1665/files#diff-f797f62cb82976f63cec2e5936234848fab2e19af04010edc1945c822d291cf0R7

The `aesni-gcm-avx512` assembly file is unconditionally included in the build for x86-64:
https://github.com/aws/aws-lc/blob/bf992e705fa475abcff9e9f2a137babbd0399164/crypto/fipsmodule/CMakeLists.txt#L30

crypto/fipsmodule/bn/asm/x86_64-mont.pl
justsmth: As above, it's only needed so that the pre-generated assembly can be used with GCC 4.8.x.

.github/docker_images/gcc-4.8/awslc_build.sh
```diff
@@ -0,0 +1,18 @@
...
+set -ex -o pipefail
+
+echo "Building with GCC Version: $(gcc --version)"
+
+BUILD_DIR=$(mktemp -d)
+SRC_DIR="${SRC_DIR:-/awslc}"
+
+pushd "${BUILD_DIR}"
+
+cmake "${SRC_DIR}" "-DDISABLE_GO=ON" "-DDISABLE_PERL=ON" "-DMY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX=1"
```
nebeid, line 15: Do we want to test also with Perl and Go available, or it's enough we're doing that in our CI?

justsmth: Right, we already have several tests in our CI that test GCC 4.8 with Go and Perl enabled. The focus of this test is to verify that we can build with GCC 4.8 and the pre-generated assembly. (I'm not sure that there is a Go version able to run our Go code that is compatible with the glibc version available on that docker image.)

nebeid (approved):

dkostic (approved):

lilpoozie2005 (approved):

PR #1667, opened 2024-06-26
Add test to ensure sequence numbers are allowed to increase by more than one

maddeleine:
### Issues:
P131226031

### Description of changes: 
EVP_aead_aes_128_gcm_tls13() and EVP_aead_aes_256_gcm_tls13() are currently required to encrypt with a monotonically-increasing sequence number. This sequence number however is allowed to increase by more than 1 per encrypt call. This PR adds a test to codify this behavior.
### Testing:
Adds test.


BUILDING.md
```diff
@@ -20,7 +20,7 @@ If in doubt, use the most recent stable version of each build tool.
     `PERL_EXECUTABLE`.
     * To build without Perl (not recommended) see [this section.](#using-pre-generated-build-files)
 
-  * [Go](https://golang.org/dl/) 1.18 or later is required. If not found by
+  * [Go](https://golang.org/dl/) 1.20 or later is required. If not found by
```
justsmth, line 23: go 1.20 is required for running the tests, but I believe the project can still be built using go 1.18. Perhaps we could set go 1.20 as the minimum only when `BUILD_TESTING` is enabled?

These other places would also need to be updated:
* https://github.com/aws/aws-lc/blob/b5e3681ff4dd984a5bab223c676752e954c1b3f7/cmake/go.cmake#L25
* https://github.com/aws/aws-lc/blob/b5e3681ff4dd984a5bab223c676752e954c1b3f7/go.mod#L4-L3

crypto/cipher_extra/aead_test.cc
```diff
@@ -1364,6 +1364,48 @@ TEST(AEADTest, TestGCMSIV256Change16Alignment) {
...
+TEST(AEADTest, TestMonotonicityCheck) {
+
+  static const uint8_t kEvpAeadCtxKey[32] = {0};
+
+  // Only the tls13() ciphers have monotonicity checks
+  struct {
+    const EVP_AEAD *cipher;
+    const size_t key_len;
+  } ctx[] = { { EVP_aead_aes_128_gcm_tls13(), 16},
+              { EVP_aead_aes_256_gcm_tls13(), 32} };
```
andrewhop, line 1376: Instead of hardcoding these lengths you can get them using `EVP_AEAD_key_length` and `EVP_AEAD_nonce_length`, and if it's just a list of aeads I'd rename it aeads_to_test.

crypto/cipher_extra/aead_test.cc
```diff
@@ -1364,6 +1364,48 @@ TEST(AEADTest, TestGCMSIV256Change16Alignment) {
...
+
+  for (int i = 0; i < 2; i++) {
+    const EVP_AEAD *cipher = ctx[i].cipher;
+    EVP_AEAD_CTX *encrypt_ctx =
+      (EVP_AEAD_CTX *)malloc(sizeof(EVP_AEAD_CTX) + 8);
```
andrewhop, line 1381: You should use our nice c++ helper classes. The above test is doing weird stuff to test for alignment/ABI isseus and shouldn't be used as a template for "normal" tests. 
```suggestion
    bssl::ScopedEVP_AEAD_CTX encrypt_ctx;
```

crypto/cipher_extra/aead_test.cc
```diff
@@ -1364,6 +1364,48 @@ TEST(AEADTest, TestGCMSIV256Change16Alignment) {
...
+    uint8_t plaintext[16] = {0};
+    uint8_t ciphertext[32] = {0};
+    size_t out_len = 0;
+
+    // Checks that sequence numbers are allowed to increment by more than one
+    // as long as monotonicity is preserved. Here the implicit IV is presumed
+    // to be a zero-filled array. That lets us update the nonce value directly
+    // with an increasing sequence number.
+    for (size_t sequence_num = 0; sequence_num <= 255; sequence_num+=10) {
+      nonce[last_byte] = sequence_num;
```
andrewhop, line 1399: NP: this should probably use uint8_t, currently it looks a little weird to assign a size_t to a byte array.

crypto/cipher_extra/aead_test.cc
```diff
@@ -1364,6 +1364,48 @@ TEST(AEADTest, TestGCMSIV256Change16Alignment) {
...
+      ASSERT_TRUE(EVP_AEAD_CTX_seal(encrypt_ctx, ciphertext, &out_len,
+                                    sizeof(ciphertext), nonce, sizeof(nonce), plaintext,
+                                    sizeof(plaintext), nullptr /* ad */, 0));
+    }
+
+    free(encrypt_ctx);
```
andrewhop, line 1405: Not needed with the smart pointer. 
```suggestion
```

crypto/cipher_extra/aead_test.cc
```diff
@@ -1364,6 +1364,48 @@ TEST(AEADTest, TestGCMSIV256Change16Alignment) {
...
+
+  // Only the tls13() ciphers have monotonicity checks
+  struct {
+    const EVP_AEAD *cipher;
+    const size_t key_len;
+  } ctx[] = { { EVP_aead_aes_128_gcm_tls13(), 16},
+              { EVP_aead_aes_256_gcm_tls13(), 32} };
+
+  for (int i = 0; i < 2; i++) {
+    const EVP_AEAD *cipher = ctx[i].cipher;
```
andrewhop, line 1379:
```suggestion
    for (EVP_AEAD cipher : aeads_to_test) {
```

crypto/cipher_extra/aead_test.cc
maddeleine: 🤔 Then I end up with this compiler error.
```
/awslc/crypto/cipher_extra/aead_test.cc:1390:49: error: comparison is always true due to limited range of data type [-Werror=type-limits]
 1390 |     for (uint8_t sequence_num = 0; sequence_num <= 255; sequence_num+=10) {
      |                                    ~~~~~~~~~~~~~^~~~~~
```

crypto/cipher_extra/aead_test.cc
```diff
@@ -1364,6 +1364,38 @@ TEST(AEADTest, TestGCMSIV256Change16Alignment) {
...
+    // Checks that sequence numbers are allowed to increment by more than one
+    // as long as monotonicity is preserved. Here the implicit IV is presumed
+    // to be a zero-filled array. That lets us update the nonce value directly
+    // with an increasing sequence number.
+    for (size_t sequence_num = 0; sequence_num <= 255; sequence_num+=10) {
+      nonce[last_byte] = sequence_num;
+      ASSERT_TRUE(EVP_AEAD_CTX_seal(encrypt_ctx.get(), ciphertext, &out_len,
+                                    sizeof(ciphertext), nonce, sizeof(nonce), plaintext,
+                                    sizeof(plaintext), nullptr /* ad */, 0));
+    }
```
andrewhop, line 1395: While we're hear it's nice to test that the monotonicity check is doing something.
```suggestion
    }
    nonce[last_byte] = 0;
    ASSERT_FALSE(EVP_AEAD_CTX_seal(encrypt_ctx.get(), ciphertext, &out_len,
                               sizeof(ciphertext), nonce, sizeof(nonce), plaintext,
                               sizeof(plaintext), nullptr /* ad */, 0));
```

crypto/cipher_extra/aead_test.cc
andrewhop: Ahhh yeah didn't think about that, this works then.

justsmth (approved):

andrewhop (approved):

lilpoozie2005 (approved):

PR #1670, opened 2024-06-28
Add support for parsing ECPKParameter PEM files

samuel40791765:
### Issues:
Addresses `CryptoAlg-1715`

### Description of changes: 
Ruby depends on reading/writing PEM files that are defined as `EcpkParameters ` (See [RFC 3279](https://datatracker.ietf.org/doc/html/rfc3279#section-2.3.5) for more details).`EcpkParameters` actually just refer to potential different ways of serializing EC Groups and we already have basic support for this. OpenSSL's implementation of `d2i/i2d_ECParameters` actually [calls `d2i/i2d_ECPKParameters`](https://github.com/openssl/openssl/blob/fbd6609bb21b125c9454d07c484d166a33b4815b/crypto/ec/ec_asn1.c#L1088-L1114) internally.

* I've borrowed some of the logic for `d2i/i2d_ECParameters` and worked them into the new functions. Apparently we only support parsing named curves or explicitly encoded versions of them, so I've documented that and followed the same pattern. No new parsing logic has been added for `EC_GROUP`s. 
* We do not support writing custom curves to any ASN.1 output and this PR continues that trend.

### Call-outs:
* Code change in `crypto/pem/pem_lib.c` was taken from OpenSSL: https://github.com/openssl/openssl/commit/205957405d08ef199e6ab654e333a627bbca9ccc It was worth taking, so the call to `PEM_write_bio` could be cleaner.

### Testing:
* Tests for named curves and a explicitly encoded versions of P256. The same PEM file works fine with OpenSSL as well
* We don't support writing custom curves, so I have a test to make sure.

### DIFF
--- a/crypto/ec_extra/ec_asn1.c
+++ b/crypto/ec_extra/ec_asn1.c
@@ -484,9 +484,7 @@ EC_KEY *d2i_ECParameters(EC_KEY **out_key, const uint8_t **inp, long len) {
     return NULL;
   }
 
-  CBS cbs;
-  CBS_init(&cbs, *inp, (size_t)len);
-  const EC_GROUP *group = EC_KEY_parse_parameters(&cbs);
+  EC_GROUP *group = d2i_ECPKParameters(NULL, inp, len);
   if (group == NULL) {
     return NULL;
   }
@@ -501,8 +499,28 @@ EC_KEY *d2i_ECParameters(EC_KEY **out_key, const uint8_t **inp, long len) {
     EC_KEY_free(*out_key);
     *out_key = ret;
   }
-  *inp = CBS_data(&cbs);
   return ret;
+}
+
+EC_GROUP *d2i_ECPKParameters(EC_GROUP **out_group, const uint8_t **inp,
+                             long len) {
+  if (len < 0) {
+    return NULL;
+  }
+
+  CBS cbs;
+  CBS_init(&cbs, *inp, (size_t)len);
+  EC_GROUP *group = EC_KEY_parse_parameters(&cbs);
+  if (group == NULL) {
+    return NULL;
+  }
+
+  if (out_group != NULL) {
+    EC_GROUP_free(*out_group);
+    *out_group = group;
+  }
+  *inp = CBS_data(&cbs);
+  return group;
 }
 
 int i2d_ECParameters(const EC_KEY *key, uint8_t **outp) {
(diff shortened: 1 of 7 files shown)


crypto/ec_extra/ec_asn1.c
```diff
@@ -501,19 +499,47 @@ EC_KEY *d2i_ECParameters(EC_KEY **out_key, const uint8_t **inp, long len) {
...
 }
 
+EC_GROUP *d2i_ECPKParameters(EC_GROUP **out_group, const uint8_t **inp,
+                             long len) {
+  if (len < 0) {
+    return NULL;
+  }
+
+  CBS cbs;
+  CBS_init(&cbs, *inp, (size_t)len);
```
skmcgrail, line 512: Check `inp != NULL` prior to first dereference.

skmcgrail (approved):

crypto/pem/pem_test.cc
```diff
@@ -187,5 +187,81 @@ TEST(PEMTest, WriteReadECPem) {
   const BIGNUM* orig_priv_key = EC_KEY_get0_private_key(ec_key.get());
   const BIGNUM* read_priv_key = EC_KEY_get0_private_key(ec_key_read.get());
   ASSERT_EQ(0, BN_cmp(orig_priv_key, read_priv_key));
+}
+
+const char *kPemECPARAMETERS =
```
dkostic, line 192: how did you generate these?

dkostic (approved):

crypto/pem/pem_test.cc
samuel40791765: `kPemECPARAMETERS` was more easily generate-able. Running the following command on the system's openssl gives you the output.
```
aws-lc on  main [?] via △ v3.20.0 via  v1.21.3 
❯ openssl ecparam -name secp521r1
-----BEGIN EC PARAMETERS-----
BgUrgQQAIw==
-----END EC PARAMETERS-----
```

`kPemExplictECPARAMETERS` was slightly harder, but not too complicated. I took the contents of the explicit P256 curve in [our test file](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/ec/ec_test.cc#L50-L78) and wrote it out with `PEM_write_bio(..., PEM_STRING_ECPARAMETERS, ...)`.

lilpoozie2005 (approved):

PR #1686, opened 2024-07-02
add ECPKParameters_print as no-op

samuel40791765:
### Issues:
Resolves `CryptoAlg-1715`

### Description of changes: 
Ruby [consumes](https://github.com/ruby/ruby/blob/ruby_3_1/ext/openssl/ossl_pkey_ec.c#L1071-L1088) `ECPKParameters_print`. The entire function is a bit convoluted and not really beneficial unless you're using custom curves. We can expose this as a no-op until there's a more proper use case for this.

### Call-outs:
N/A

### Testing:
N/A

### DIFF
--- a/include/openssl/ec.h
+++ b/include/openssl/ec.h
@@ -475,6 +475,9 @@ OPENSSL_EXPORT OPENSSL_DEPRECATED unsigned char *EC_GROUP_get0_seed(
 OPENSSL_EXPORT OPENSSL_DEPRECATED size_t
 EC_GROUP_get_seed_len(const EC_GROUP *group);
 
+// ECPKParameters_print returns 1.
+OPENSSL_EXPORT OPENSSL_DEPRECATED int ECPKParameters_print(BIO *bio, const EC_GROUP *group,
+                                               int offset );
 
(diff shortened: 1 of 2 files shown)


include/openssl/ec.h
```diff
@@ -460,6 +460,9 @@ OPENSSL_EXPORT OPENSSL_DEPRECATED unsigned char *EC_GROUP_get0_seed(
 OPENSSL_EXPORT OPENSSL_DEPRECATED size_t
 EC_GROUP_get_seed_len(const EC_GROUP *group);
 
+// ECPKParameters_print returns 1.
```
WillChilds-Klein, line 463: nit: you might want to call out that this doesn't actually print anything.

WillChilds-Klein (approved):

include/openssl/ec.h
samuel40791765: No problem, I was following the pattern we had for other no-ops. But the pattern is a bit too simplistic to be honest.

dkostic (approved):

WillChilds-Klein (approved):

lilpoozie2005 (approved):

PR #1687, opened 2024-07-02
Enabling DIT flag in AArch64.

nebeid:
### Issues:
Addresses CryptoAlg-2501

### Description of changes: 
- Add a compile flag to enable it.
- DIT APIs and macro are available for external use.
- Add an option `-dit` to the speed test that's available in a build that used the compile flag. 
Enabling it makes the benchmarks closer to not having DIT enabled. 
This is to suggest that the macro can be used by AWS-LC callers in their own scope 
around multiple calls to the library to improve the performance since, 
in that case, DIT would not switch on and off at the entry/exit of the library functions.

### Call-outs:
N/A

### Testing:
- A test is added to CI with the compile flag that enables DIT set to ON.
- Tested with printouts the value of the DIT when it's expected to be turned 1 or 0.


CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ option(ENABLE_DILITHIUM "Enable Dilithium signatures in the EVP API" OFF)
 option(DISABLE_PERL "Disable Perl for AWS-LC" OFF)
 option(DISABLE_GO "Disable Go for AWS-LC" OFF)
 option(ENABLE_FIPS_ENTROPY_CPU_JITTER "Enable FIPS entropy source: CPU Jitter" OFF)
+option(ENABLE_DATA_INDEPENDENT_TIMING "Enable Data-Independent Timing (DIT) flag" OFF)
```
dkostic, line 26: maybe a bit more specific?
```suggestion
option(ENABLE_DATA_INDEPENDENT_TIMING_AARCH64 "Enable Data-Independent Timing (DIT) flag" OFF)
```

CMakeLists.txt
```diff
@@ -801,6 +802,11 @@ else()
   set(ARCH "generic")
 endif()
 
+#if(ENABLE_DATA_INDEPENDENT_TIMING AND ARCH EQUAL "aarch64")
```
dkostic, line 805: leftover?

crypto/fipsmodule/cpucap/cpu_aarch64.c
```diff
@@ -49,4 +49,44 @@ void handle_cpu_env(uint32_t *out, const char *in) {
...
+//	Op1 (3 for DIT) , Op2 (5 for DIT) encodes the PSTATE field modified and defines the constraints.
+//	CRm = Imm4 (#0 or #1 below)
+//	Rt = 0x1f
+uint64_t armv8_enable_dit(void) {
+  uint64_t original_dit = armv8_get_dit();
+  if (CRYPTO_is_ARMv8_DIT_capable() && original_dit != 1) {
+    // Encoding of "msr dit, #1"
+    __asm__ volatile(".long 0xd503415f");
+  }
+  return original_dit;
```
dkostic, line 81: not sure I follow the logic behind returning the old value?

crypto/fipsmodule/cpucap/cpu_aarch64.c
```diff
@@ -49,4 +49,44 @@ void handle_cpu_env(uint32_t *out, const char *in) {
...
+// barriers and CLREX, and PSTATE access", ARM DDI 0487 J.a, system instructions
+// for accessing PSTATE fields have the following encoding
+// and C5.2.4 DIT, Data Independent Timing:
+//	Op0 = 0, CRn = 4
+//	Op1 (3 for DIT) , Op2 (5 for DIT) encodes the PSTATE field modified and defines the constraints.
+//	CRm = Imm4 (#0 or #1 below)
+//	Rt = 0x1f
+uint64_t armv8_enable_dit(void) {
+  uint64_t original_dit = armv8_get_dit();
+  if (CRYPTO_is_ARMv8_DIT_capable() && original_dit != 1) {
```
dkostic, line 77: do we have to check `original_dit != 1`?

include/openssl/crypto.h
```diff
@@ -84,6 +84,28 @@ OPENSSL_EXPORT int CRYPTO_has_broken_NEON(void);
...
 
+// Data-Independent Timing (DIT) on AArch64
+
+#if defined(OPENSSL_AARCH64) && !defined(OPENSSL_WINDOWS) && defined(MAKE_DIT_AVAILABLE)
+// (TODO): See if we can detect the DIT capability in Windows environment
+uint64_t armv8_get_dit(void);
+uint64_t armv8_enable_dit(void);
+void armv8_restore_dit(volatile uint64_t *original_dit);
+
+#define ENABLE_DIT_ALL                      \
```
dkostic, line 95: seems like this is unused?

include/openssl/crypto.h
```diff
@@ -84,6 +84,28 @@ OPENSSL_EXPORT int CRYPTO_has_broken_NEON(void);
 OPENSSL_EXPORT int CRYPTO_needs_hwcap2_workaround(void);
 #endif  // OPENSSL_ARM && OPENSSL_LINUX && !OPENSSL_STATIC_ARMCAP
 
+// Data-Independent Timing (DIT) on AArch64
```
dkostic, line 87: this has to be in the public header?

include/openssl/crypto.h
```diff
@@ -84,6 +84,28 @@ OPENSSL_EXPORT int CRYPTO_has_broken_NEON(void);
...
+uint64_t armv8_get_dit(void);
+uint64_t armv8_enable_dit(void);
+void armv8_restore_dit(volatile uint64_t *original_dit);
+
+#define ENABLE_DIT_ALL                      \
+  volatile uint64_t _dit_orig                \
+          OPENSSL_UNUSED = armv8_enable_dit();
+
+
+#define ENABLE_DIT_AUTO_DISABLE                      \
```
dkostic, line 100: I think having both enable and disable in the name is a bit confusing. can we do `SET_DIT_AUTO_DISABLE` or something else?

include/openssl/crypto.h
```diff
@@ -84,6 +84,28 @@ OPENSSL_EXPORT int CRYPTO_has_broken_NEON(void);
...
+  volatile uint64_t _dit_restore_orig                \
+         __attribute__((cleanup(armv8_restore_dit))) \
+          OPENSSL_UNUSED = armv8_enable_dit();
```
dkostic, line 103: can you explain a bit what's this construction doing? I'm a bit lost tbh.

include/openssl/crypto.h
dkostic: scratch that, I just read the PR description.

include/openssl/crypto.h
```diff
@@ -84,6 +84,28 @@ OPENSSL_EXPORT int CRYPTO_has_broken_NEON(void);
...
+void armv8_restore_dit(volatile uint64_t *original_dit);
+
+#define ENABLE_DIT_ALL                      \
+  volatile uint64_t _dit_orig                \
+          OPENSSL_UNUSED = armv8_enable_dit();
+
+
+#define ENABLE_DIT_AUTO_DISABLE                      \
+  volatile uint64_t _dit_restore_orig                \
+         __attribute__((cleanup(armv8_restore_dit))) \
```
torben-hansen, line 102: What if a function with `ENABLE_DIT_AUTO_DISABLE` use another function with `ENABLE_DIT_AUTO_DISABLE` as a subroutine? Then as soon as the latter returns, the DIT will be reset for the remaining scope of the former (well unless it calls other functions with `ENABLE_DIT_AUTO_DISABLE`)?

include/openssl/crypto.h
nebeid: From our offline conversation:
`_dit_restore_orig` on l.101 which is the current DIT value on scope entry is output by `armv8_enable_dit()`. On scope exit, it is passed to `armv8_restore_dit()`. To illustrate:
```
func() {
  // ENABLE_DIT_AUTO_DISABLE does the following:
  // This first part is done in armv8_enable_dit()
  old_dit_state <- read old dit state;
  if (old_dit_state != 1) {
   SET_DIT
  }
  ...
  // scope exit, calls armv8_restore_dit()
  if (old_dit_state != 1) UNSET_DIT;
}
```

include/openssl/crypto.h
nebeid: I tried to explain it in the reply to the comment below by dkostic.

include/openssl/crypto.h
nebeid: Done.

include/openssl/crypto.h
nebeid: Yes, thanks dkostic. Removed.

crypto/fipsmodule/cpucap/cpu_aarch64.c
nebeid: Removed. Thanks.

crypto/fipsmodule/cpucap/cpu_aarch64.c
nebeid: It's needed to be able to restore to it on scope exit as illustrated in the other response.

CMakeLists.txt
nebeid: Removed. Thanks.

crypto/fipsmodule/cpucap/cpu_aarch64.c
```diff
@@ -49,4 +49,44 @@ void handle_cpu_env(uint32_t *out, const char *in) {
   }
 }
 
+#if defined(MAKE_DIT_AVAILABLE) && !defined(OPENSSL_WINDOWS)
+// "DIT" is not recognised as a register name by clang-10 (at least)
+// Register's encoded name is from e.g.
+// https://github.com/ashwio/arm64-sysreg-lib/blob/d421e249a026f6f14653cb6f9c4edd8c5d898595/include/sysreg/dit.h#L286
+#define DIT_REGISTER s3_3_c4_c2_5
+
+uint64_t armv8_get_dit(void) {
```
dkostic, line 58: I think this function can be static since it's not used outside this file?

crypto/fipsmodule/cpucap/cpu_aarch64.c
```diff
@@ -49,4 +49,44 @@ void handle_cpu_env(uint32_t *out, const char *in) {
...
+// barriers and CLREX, and PSTATE access", ARM DDI 0487 J.a, system instructions
+// for accessing PSTATE fields have the following encoding
+// and C5.2.4 DIT, Data Independent Timing:
+//	Op0 = 0, CRn = 4
+//	Op1 (3 for DIT) , Op2 (5 for DIT) encodes the PSTATE field modified and defines the constraints.
+//	CRm = Imm4 (#0 or #1 below)
+//	Rt = 0x1f
+uint64_t armv8_enable_dit(void) {
+  uint64_t original_dit = armv8_get_dit();
+  if (CRYPTO_is_ARMv8_DIT_capable()) {
```
dkostic, line 77: you can do:
```
if (CRYPTO_is_ARMv8_DIT_capable()) {
  uint64_t original_dit = armv8_get_dit();
  __asm__ volatile(".long 0xd503415f");
  return original_dit;
} else {
  return 0;
}
```
This way you can remove `if (CRYPTO_is_ARMv8_DIT_capable())` from `armv8_get_dit` so it doesn't get called twice.

crypto/fipsmodule/cpucap/cpu_aarch64.c
```diff
@@ -49,4 +49,44 @@ void handle_cpu_env(uint32_t *out, const char *in) {
...
+// Register's encoded name is from e.g.
+// https://github.com/ashwio/arm64-sysreg-lib/blob/d421e249a026f6f14653cb6f9c4edd8c5d898595/include/sysreg/dit.h#L286
+#define DIT_REGISTER s3_3_c4_c2_5
+
+uint64_t armv8_get_dit(void) {
+  uint64_t val = 0;
+  if (CRYPTO_is_ARMv8_DIT_capable()) {
+    __asm__ volatile("mrs %0, s3_3_c4_c2_5" : "=r" (val));
+  }
+  return val >> 24;
```
dkostic, line 63: if the intention is to extract bit 24 only then we should do: `(val >> 24) & 1`?

crypto/fipsmodule/cpucap/cpu_aarch64.c
nebeid: Done. Thanks.

crypto/fipsmodule/cpucap/cpu_aarch64.c
nebeid: Done.

crypto/fipsmodule/cpucap/cpu_aarch64.c
nebeid: Done, thanks. I had missed that because in the printout tests, it printed as 1.

dkostic (dismissed):

AGSaidi:
I don't see any documentation about how, if all the DIT twiddling cause perf degradation, at a higher level in the program DIT could be enabled. Is it worth documentating that somewhere to try and define an API for that that other crypto libs would also adopt?

crypto/fipsmodule/dh/dh.c
```diff
@@ -116,20 +117,39 @@ void DH_free(DH *dh) {
...
 }
 
-unsigned DH_bits(const DH *dh) { return BN_num_bits(dh->p); }
+unsigned DH_bits(const DH *dh) {
+  SET_DIT_AUTO_DISABLE;
+  return BN_num_bits(dh->p);
+}
 
-const BIGNUM *DH_get0_pub_key(const DH *dh) { return dh->pub_key; }
+const BIGNUM *DH_get0_pub_key(const DH *dh) {
```
torben-hansen, line 125: Needed when simply returning a pointer?

crypto/fipsmodule/dh/dh.c
```diff
@@ -97,6 +97,7 @@ DH *DH_new_by_nid(int nid) {
 }
```
torben-hansen, line 97: missed `calculate_rfc7919_DH_from_p()`?

torben-hansen: Ahh I  guess those are public parameters...

crypto/fipsmodule/ec/ec.c
```diff
@@ -881,6 +881,7 @@ int ec_point_mul_scalar_public_batch(const EC_GROUP *group, EC_JACOBIAN *r,
```
torben-hansen, line 881: `EC_POINT_mul()` maybe?

tool/speed.cc
```diff
@@ -2584,6 +2584,7 @@ static bool parseStringVectorToIntegerVector(
 }
 
 bool Speed(const std::vector<std::string> &args) {
+  //SET_DIT_AUTO_DISABLE;
```
torben-hansen, line 2587: You can add this as an option to the speed tool

crypto/fipsmodule/tls/kdf.c
```diff
@@ -58,7 +58,7 @@
 #include <openssl/mem.h>
 
 #include "../../internal.h"
-
+#include "../service_indicator/internal.h"
```
torben-hansen, line 61: why include?

nebeid:
> I don't see any documentation about how, if all the DIT twiddling cause perf degradation, at a higher level in the program DIT could be enabled. Is it worth documentating that somewhere to try and define an API for that that other crypto libs would also adopt?

Thank you @AGSaidi , I added the documentation to BUILDING.md and to crypto.h.

torben-hansen (approved):

dkostic (approved):

PR #1691, opened 2024-07-08
Add partially deinterleaved version of non-EOR3 AES-GCM kernels

hanno-becker:
**Depends on:** #1639 

**Context:** On AArch64 platforms without EOR3 (e.g. Graviton2), AWS-LC leverages the assembly kernels aes_gcm_enc_kernel and aes_gcm_dec_kernel for fast AES-GCM computations. These kernels are highly interleaved in consideration of instruction latencies, which drastically improves performance even on out of order cores like Graviton2. However, the interleaving also impedes an intuitive understanding of what is happening, and thereby hardens modification and verification efforts.

**Change:** To facilitate auditability and verification, this commit adds a separate source file which contains partially deinterleaved versions of the non-EOR3 AES-GCM assembly kernels. The interleaving is 'partial' in the sense that, within a control flow block, instructions are freely reordered, but not across control flow boundaries. In particular, software pipelining is still present in the de-interleaved versions.

There is further potential for clarifying the de-interleaved code, for example by (a) the use of macros, or (b) de-interleaving across loop boundaries. Potential changes along those lines are left for future commits.

The de-interleaved versions are provided in a separate file `aesv8-gcm-armv8-deinterleaved.pl`.

### DIFF
--- a/crypto/fipsmodule/CMakeLists.txt
+++ b/crypto/fipsmodule/CMakeLists.txt
@@ -88,6 +88,7 @@ if(ARCH STREQUAL "aarch64")
 
     aesv8-armx.${ASM_EXT}
     aesv8-gcm-armv8.${ASM_EXT}
+    aesv8-gcm-armv8-deinterleaved.${ASM_EXT}
     aesv8-gcm-armv8-unroll8.${ASM_EXT}
     armv8-mont.${ASM_EXT}
     bn-armv8.${ASM_EXT}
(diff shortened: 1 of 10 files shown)


darylmartin100:
This shouldn't be going into our mainline, can you please either submit this to a fork/branch that you maintain?  Alternatively, this could be put into the s2n-bignum repo until we are ready to consider putting this into our mainline.

PR #1697, opened 2024-07-10
Fix Windows/ARM64 assembly build

justsmth:
### Issues:
* Addresses: https://github.com/aws/aws-lc-rs/issues/453

### Description of changes: 
* When building for Windows/ARM64 with the "Visual Studio" generator (instead of Ninja) the "*.S" assembly source files for libraries are ignored.
* This change provides a custom command to assemble the require objects files and adds them to the list of sources for the library.

### Call-outs:
* See related aws-lc-rs PR: https://github.com/aws/aws-lc-rs/pull/452

### Testing
* I've added CI tests to build for Windows using the "Visual Studio" generator.


CMakeLists.txt
```diff
@@ -9,6 +9,10 @@ if(MSVC)
   set(CMAKE_GENERATOR_CC cl)
 endif()
 
+if(ARCH STREQUAL "aarch64" AND CMAKE_GENERATOR MATCHES "Visual Studio" AND NOT "${CMAKE_VS_PLATFORM_TOOLSET}" MATCHES "ClangCL")
```
andrewhop, line 12: Is this issue aarch64 specific? Should this also check for arch "arm"?

crypto/CMakeLists.txt
```diff
@@ -1,3 +1,47 @@
+
+# Function to handle assembly files for ARM64 targets when using MSBuild (the "Visual Studio" generator)
+# This function is necessary because MSBuild ignores ARM64 assembly file dependencies
+function(msbuild_aarch64_asm)
+  set(options "")
+  set(oneValueArgs TARGET OUTPUT_OBJECTS)
+  set(multiValueArgs ASM_FILES)
+  
+  cmake_parse_arguments(MSBUILD_AARCH64_ASM "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGV})
```
andrewhop, line 9: TIL about cmake_parse_arguments, very neat.

I think cmake_parse_arguments was only added in CMake 3.5. Based on our CI we test back to 3.2 on Linux. Is it possible to implement msbuild_aarch64_asm without using cmake_parse_arguments?

.github/workflows/windows-alt.yml
```diff
@@ -63,13 +63,56 @@ jobs:
...
+            CMAKE_SYSTEM_NAME=Windows \
+            CMAKE_SYSTEM_PROCESSOR=ARM64 \
+            CMAKE_BUILD_TYPE=Release \
+      - name: Build Project
+        run: cmake --build ./build --target all_tests
+      - if: ${{ matrix.target  == 'x64' }}
+        name: Run crypto_test
+        # MSVC places the tests executables in a different location.
+        # With MSVC, the "run_tests" target cannot locate the tests.
+        # We run "crypto_test" as a sanity check.
```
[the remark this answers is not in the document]
justsmth: MSVC inserts the build type (e.g., "Debug" or "Release") into the path right above the binary.  So instead of [`"crypto/crypto_test"`](https://github.com/aws/aws-lc/blob/main/util/all_tests.json#L17), the path would need to be `"crypto/Debug/crypto_test"` or `"crypto/Release/crypto_test"`.

There might be a way to alter the artifact path whenever we see that the Generator matches "Visual Studio", but it's not been an issue prior to now. So I don't think it's worth the effort to pursue it.

crypto/CMakeLists.txt
justsmth: It was available in older CMake versions, it just required including the [CMakeParseArguments](https://cmake.org/cmake/help/latest/module/CMakeParseArguments.html#module:CMakeParseArguments) module prior to 3.5.

andrewhop: Is that included automatically? Or something you can add?

CMakeLists.txt
justsmth: Yeah, I suspect Windows/ARM32 assembly would also need special handling, but don't know that we would even generate proper assembly for it. (I didn't consider Windows/ARM32 to be something had support for.)  We can revisit supporting that platform if anyone requests.

crypto/CMakeLists.txt
justsmth: Yeah, I'll add it.

skmcgrail (approved):

util/build_compilation_database.sh
```diff
@@ -12,8 +12,8 @@ MY_CMAKE_FLAGS=("-GNinja" "-DCMAKE_BUILD_TYPE=Debug" "-DCMAKE_EXPORT_COMPILE_COM
 
 mkdir -p "${AWS_LC_BUILD}"
 
-cmake "${BASE_DIR}" -B "${AWS_LC_BUILD}" ${MY_CMAKE_FLAGS[@]} "${@}"
+cmake "${BASE_DIR}" -B "${AWS_LC_BUILD}" "${MY_CMAKE_FLAGS[@]}" "${@}"
 
-cmake --build "${AWS_LC_BUILD}" --target all
+cmake --build "${AWS_LC_BUILD}" -j 4 --target all_tests
```
WillChilds-Klein, line 17: nit: why 4 here? should we use `$(nproc)` or [the integration tests' logic](https://github.com/aws/aws-lc/blob/1fbf58437a32da4008774eed92530b787d210b1e/tests/ci/common_posix_setup.sh#L24)?

CMakeLists.txt
```diff
@@ -366,10 +370,10 @@ if(GCC OR CLANG)
     else()
       set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wextra")
       set(C_CXX_FLAGS "${C_CXX_FLAGS} -Wall -fvisibility=hidden -fno-common")
+      set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wno-newline-eof")
```
WillChilds-Klein, line 373: not a concern, but why don't we want this for MSVC?

.github/workflows/windows-alt.yml
```diff
@@ -63,13 +63,53 @@ jobs:
...
+        uses: threeal/cmake-action@v1.3.0
+        with:
+          options: |
+            CMAKE_GENERATOR_PLATFORM=ARM64 \
+            CMAKE_SYSTEM_NAME=Windows \
+            CMAKE_SYSTEM_PROCESSOR=ARM64 \
+            CMAKE_BUILD_TYPE=Release \
+      - name: Build Project
+        run: cmake --build ./build --target all_tests
+      - if: ${{ matrix.target  == 'x64' }}
```
WillChilds-Klein, line 109: do we not want to run the tests on `x64_arm64`?

util/build_compilation_database.sh
justsmth: B/c it provides an easy way to perform the build, I was running the script on Windows.  It's not really relevant to this PR otherwise.  It seemed to be defaulting to 0 with "Visual Studio" generator, so I changed it to 4 which provides a reasonable level of parallelism for most hosts.

CMakeLists.txt
justsmth: It's apparently not a valid parameter on MSVC -- I was getting warning due to it during the build.

.github/workflows/windows-alt.yml
justsmth: IFAICT, we can't test `arm64` executables on the `x86` Windows hosts provided by GitHub.

WillChilds-Klein: ah, yep. that would do it.

WillChilds-Klein (approved):

lilpoozie2005 (approved):

PR #1703, opened 2024-07-12
add support for OCSP_SINGLERESP functions

samuel40791765:
### Issues:
Addresses `CryptoAlg-2420`

### Description of changes: 
Add support for ASN1 and X509v3 extension functions for `OCSP_SINGLERESP`. These are all consumed in Ruby's OCSP module.

### Call-outs:
N/A

### Testing:
New tests for the extension functions and `OCSP_SINGLERESP` creation.


crypto/ocsp/ocsp_extension.c
```diff
@@ -37,6 +37,19 @@ X509_EXTENSION *OCSP_BASICRESP_get_ext(OCSP_BASICRESP *bs, int loc) {
   return X509v3_get_ext(bs->tbsResponseData->responseExtensions, loc);
 }
 
+int OCSP_SINGLERESP_add_ext(OCSP_SINGLERESP *sresp, X509_EXTENSION *ex,
```
andrewhop, line 40: Missing null checks.

samuel40791765: Yeah I was on the fence for these, they're single calls against existing functions that already have null checks. 
https://github.com/aws/aws-lc/blob/34aceaf04b1ff0499104a73e67c7d4b36551f1f8/crypto/x509/x509_v3.c#L67-L124

Should we add them here? I'm ok either way.

WillChilds-Klein: if there's no way for a NULL pointer passed in by the caller to segfault, then IMO we can skip redundant null checks.

WillChilds-Klein (commented):
other than null checks, this PR looks good to me

WillChilds-Klein (approved):

andrewhop (approved):

PR #1705, opened 2024-07-15
Add OpenVPN to CI

smittals2:
### Description of changes: 
1. Added patch for TLS1_PRF to use alternative codepath in OpenVPN (for versions <= 2.6.x)
2. integration script and workflow for Github CI

### Call-outs:
- We anchor to tip of the minor version 2.6.x. We can not support OpenVPN versions > 2.6 as tip of main has removed the alternate codepath for TLS1_PRF that we currently rely on. This change should be pulled into the next minor version of OpenVPN and will break AWS-LC compatibility. 
- Management interface is disabled when building openVPN, this will be changed in a later PR once we add in support. 


tests/ci/integration/run_openvpn_integration.sh
```diff
@@ -0,0 +1,83 @@
...
+  for patchfile in $(find -L "${OPENVPN_PATCH_BUILD_FOLDER}" -type f -name '*.patch'); do
+    echo "Apply patch $patchfile..."
+    patch -p1 --quiet -i "$patchfile"
+  done
+}
+
+function openvpn_run_tests() {
+  # Explicitly running as sudo and passing in LD_LIBRARY_PATH as some OpenVPN
+  # tests run as sudo and LD_LIBRARY_PATH doesn't get inherited.
+  sudo LD_LIBRARY_PATH="${AWS_LC_INSTALL_FOLDER}/lib" make check
```
WillChilds-Klein, line 66: why does this require root permissions?

tests/ci/integration/run_openvpn_integration.sh
```diff
@@ -0,0 +1,83 @@
...
+}
+
+git clone https://github.com/OpenVPN/openvpn.git ${OPENVPN_SRC_FOLDER}
+
+# anchoring to specific commit of OpenVPN, currently not compatible with
+# versions of OpenVPN > 2.6.10
+cd ${OPENVPN_SRC_FOLDER} && git checkout b25c6d7
```
WillChilds-Klein, line 73: how was this commit selected? did you consider specifying [a release tag](https://github.com/OpenVPN/openvpn/tags) instead?

tests/ci/integration/run_openvpn_integration.sh
smittals2: One of the integration tests (t_net.sh) uses sudo to modify iptables and conduct tests. The LD_LIBRARY_PATH set in the un-privileged environment does not get inherited by root user when this test is executed. Therefore, we need to explicitly pass in LD_LIBRARY_PATH as root user when running the tests to ensure the correct libcrypto is being used. 

Test link: https://github.com/OpenVPN/openvpn/blob/856065b2eb37286c389550593472bf180bc5be9d/tests/t_net.sh#L54

tests/ci/integration/run_openvpn_integration.sh
smittals2: This commit was picked so it was right before a new release was created. That is a good callout, I will change this to use a specific release tag instead :)

smittals2: We now pin to tip of minor version 2.6.x instead of any specific commit. This should check compatibility with any further 2.6.x changes.

andrewhop (approved):

tests/ci/integration/run_openvpn_integration.sh
```diff
@@ -0,0 +1,83 @@
...
+  OPENSSL_LIBS="-L/${AWS_LC_INSTALL_FOLDER}/lib -lssl -lcrypto" \
+  ./configure \
+    --prefix="$OPENVPN_BUILD_PREFIX" \
+    --exec-prefix="$OPENVPN_BUILD_EPREFIX" \
+    --with-crypto-library=openssl \
+    --with-openssl-engine=no \
+    --disable-management
+
+  make
+  make install
```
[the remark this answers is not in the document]
WillChilds-Klein: huh, i didn't know previously that `-j` works without explicit arguments. hopefully this will hold true for our targeted testing platforms; i'm confident in that. thank @andrewhop :)

tests/ci/integration/run_openvpn_integration.sh
WillChilds-Klein: makes sense for GitHub CI where integration tests are run in isolation and everything is ephemeral. i'm a bit worried about idempotency when running these on dev machines. today, we [don't use `sudo`](https://github.com/search?q=repo%3Aaws%2Faws-lc+path%3Atests%2Fci%2Fintegration%2F*+sudo&type=code) for integration tests. i hope openvpn is configurable enough be able to run these tests on configurable (i.e. unprivileged) ports. if not, there are [ways](https://github.com/WillChilds-Klein/pq-tls-benchmark/blob/master/emulation-exp/code/kex/setup.sh#L47) to reasonably guard against network state snafus.

let's sync up on this offline, i'm happy to work through it with you :)

WillChilds-Klein: mm, reading through t_net.sh a bit more, it looks like they're careful enough to clean up state and don't `set -e` so early exit shouldn't be an issue. i think we're good here.

tests/ci/integration/openvpn_patch/aws-lc-openvpn2-6-x.patch
```diff
@@ -29,10 +38,29 @@ index c9fa7196..a48ef391 100644
...
 +++ b/src/openvpn/ssl_openssl.c
-@@ -2314,7 +2314,7 @@ show_available_tls_ciphers_list(const char *cipher_list,
+@@ -1658,8 +1658,10 @@ tls_ctx_load_ca(struct tls_root_ctx *ctx, const char *ca_file,
+             sk_X509_INFO_pop_free(info_stack, X509_INFO_free);
+         }
+ 
++        int cnum;
+         if (tls_server)
+         {
++            cnum = sk_X509_NAME_num(cert_names);
```
WillChilds-Klein, line 51: not a blocker, but why is this change needed? looking at the [source](https://github.com/OpenVPN/openvpn/blob/release/2.6/src/openvpn/ssl_openssl.c#L1620), the `cert_names` stack doesn't appear to be modified between these assignments of `cnum`.

WillChilds-Klein (approved):

tests/ci/integration/openvpn_patch/aws-lc-openvpn2-6-x.patch
smittals2: Once SSL_CTX_set_client_CA_list is called (line 52), in OpenSSL the passed in cert_names stack remains untouched but AWS-LC resets it. So later when sk_X509_NAME_num is called (previously line 59), we get a 0 length in AWS-LC which causes an error below when cnum != added (line 60). Therefore, we obtain the length of the stack before calling SSL_CTX_set_client_CA_list. This patch doesn't break behavior for OpenSSL either.

lilpoozie2005 (approved):

PR #1706, opened 2024-07-15
Add support to detect Neoverse V2 cores

andrewhop:
### Description of changes: 
Graviton 4 uses Neoverse V2 cores which we were previously not detecting which resulted in poor performing implementations being used. It still produced the correct answer, it just didn't take full advantage of the CPU's capabilities.

### Call-outs:
I thought about combining the ARM CPU capability flag for Neoverse V1 and V2 cores into one flag but we might have a future usecase that would need specific handling even though right now they behave the same. 

We also need to test Apple M2 and M3 CPUs which will probably also want the same optimizations as M1. This also might make sense to combine into a single Apple "M" ARM capability flag.

### Testing:
Built and ran locally. On a Graviton 4 instance:
|Algorithm|Before|After|
|---|---|---|
|RSA 2048 sign|929.6 ops/sec|1,397.5 ops/sec|
|ECDH P-384|3,541.2 ops/sec|3,744.8 ops/sec|
|ECDH P-521|1,885.1 ops/sec|2,406.7 ops/sec
|AES 256 GCM 16 bytes|204.4 MB/s|203.9 MB/s|
|AES 256 GCM 16kb bytes|4,500.5 MB/s|6,019.0 MB/s|

### DIFF
--- a/crypto/fipsmodule/cpucap/cpu_aarch64_linux.c
+++ b/crypto/fipsmodule/cpucap/cpu_aarch64_linux.c
@@ -80,11 +80,14 @@ void OPENSSL_cpuid_setup(void) {
   // is supported. As of Valgrind 3.21 trying to read from that register will
   // cause Valgrind to crash.
   if (hwcap & kCPUID) {
-    // Check if the CPU model is Neoverse V1,
+    // Check if the CPU model is Neoverse V1 or V2,
     // which has a wide crypto/SIMD pipeline.
     OPENSSL_arm_midr = armv8_cpuid_probe();
     if (MIDR_IS_CPU_MODEL(OPENSSL_arm_midr, ARM_CPU_IMP_ARM, ARM_CPU_PART_V1)) {
       OPENSSL_armcap_P |= ARMV8_NEOVERSE_V1;
+    }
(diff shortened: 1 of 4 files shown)


dkostic (approved):

crypto/fipsmodule/cpucap/cpu_aarch64_linux.c
```diff
@@ -80,12 +80,15 @@ void OPENSSL_cpuid_setup(void) {
...
   // cause Valgrind to crash.
   if (hwcap & kCPUID) {
-    // Check if the CPU model is Neoverse V1,
+    // Check if the CPU model is Neoverse V1 or V2,
     // which has a wide crypto/SIMD pipeline.
     OPENSSL_arm_midr = armv8_cpuid_probe();
     if (MIDR_IS_CPU_MODEL(OPENSSL_arm_midr, ARM_CPU_IMP_ARM, ARM_CPU_PART_V1)) {
       OPENSSL_armcap_P |= ARMV8_NEOVERSE_V1;
     }
+    if (MIDR_IS_CPU_MODEL(OPENSSL_arm_midr, ARM_CPU_IMP_ARM, ARM_CPU_PART_V2)) {
```
torben-hansen, line 89: Pedantically... `elseif`...

torben-hansen (approved):

lilpoozie2005 (approved):

PR #1708, opened 2024-07-15
Support utility OCSP request functions

samuel40791765:
### Issues:
Addresses `CryptoAlg-2420`

### Description of changes: 
There are a few easier `OCSP_REQUEST` utility functions Ruby uses. We initially didn't support these as they made more sense as supporting functions for OCSP responders. 

`OCSP_id_issuer_cmp` was also pulled out of `internal.h` since it's needed.

### Call-outs:
N/A

### Testing:
Brief tests inserted in existing test suites.

### DIFF
--- a/crypto/ocsp/ocsp_server.c
+++ b/crypto/ocsp/ocsp_server.c
@@ -10,13 +10,17 @@
 // SPDX-License-Identifier: Apache-2.0 OR ISC
 // Modifications Copyright Amazon.com, Inc. or its affiliates.
 
+#include "../internal.h"
 #include "internal.h"
 
 int OCSP_request_onereq_count(OCSP_REQUEST *req) {
+  GUARD_PTR(req);
+  GUARD_PTR(req->tbsRequest);
   return (int)sk_OCSP_ONEREQ_num(req->tbsRequest->requestList);
 }
 
 OCSP_ONEREQ *OCSP_request_onereq_get0(OCSP_REQUEST *req, int i) {
+  GUARD_PTR(req);
   return sk_OCSP_ONEREQ_value(req->tbsRequest->requestList, i);
 }
 
@@ -42,6 +46,7 @@ int OCSP_id_get0_info(ASN1_OCTET_STRING **nameHash, ASN1_OBJECT **algor,
 }
 
 int OCSP_request_is_signed(OCSP_REQUEST *req) {
+  GUARD_PTR(req);
   if (req->optionalSignature != NULL) {
     return 1;
   }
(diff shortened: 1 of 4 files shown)


crypto/ocsp/ocsp_server.c
```diff
@@ -34,6 +41,13 @@ int OCSP_id_get0_info(ASN1_OCTET_STRING **nameHash, ASN1_OBJECT **algor,
   return 1;
 }
 
+int OCSP_request_is_signed(OCSP_REQUEST *req) {
+  if (req->optionalSignature != NULL) {
```
andrewhop, line 45:
```suggestion
  GUARD_PTR(req);
  if (req->optionalSignature != NULL) {
```

crypto/ocsp/ocsp_server.c
```diff
@@ -12,6 +12,13 @@
 
 #include "internal.h"
 
+int OCSP_request_onereq_count(OCSP_REQUEST *req) {
+  return (int)sk_OCSP_ONEREQ_num(req->tbsRequest->requestList);
```
andrewhop, line 16:
```suggestion
  GUARD_PTR(req);
  return (int)sk_OCSP_ONEREQ_num(req->tbsRequest->requestList);
```

crypto/ocsp/ocsp_server.c
```diff
@@ -12,6 +12,13 @@
...
+}
+
+OCSP_ONEREQ *OCSP_request_onereq_get0(OCSP_REQUEST *req, int i) {
+  return sk_OCSP_ONEREQ_value(req->tbsRequest->requestList, i);
```
andrewhop, line 20: Hmmmmm might need a new macro that returns null, if you use GUARD_PTR which returns zero (which is basically null) does the compiler get mad?

samuel40791765: Seems like it works! Also verified that returning 0 is equivalent to NULL.

justsmth (dismissed):

crypto/ocsp/ocsp_server.c
```diff
@@ -10,8 +10,19 @@
...
 
+int OCSP_request_onereq_count(OCSP_REQUEST *req) {
+  GUARD_PTR(req);
+  GUARD_PTR(req->tbsRequest);
+  return (int)sk_OCSP_ONEREQ_num(req->tbsRequest->requestList);
+}
+
+OCSP_ONEREQ *OCSP_request_onereq_get0(OCSP_REQUEST *req, int i) {
+  GUARD_PTR(req);
+  return sk_OCSP_ONEREQ_value(req->tbsRequest->requestList, i);
```
andrewhop, line 24: Missing the second check on tbsRequest

samuel40791765: Added

justsmth (approved):

andrewhop (approved):

PR #1711, opened 2024-07-17
add support for OCSP_copy_nonce

samuel40791765:
### Issues:
Addresses `CryptoAlg-2420`

### Description of changes: 
Ruby consumes `OCSP_copy_nonce`, which copies the nonce from the request to the OCSP response.
OpenSSL's `OCSP_copy_nonce` directly appends the new OCSP nonce to the list of extensions in the response. This allows multiple OCSP nonces to exist the same response since the existing nonces are not deleted. It's a bit strange to allow multiple nonces to exist in a single response. To make things stranger, `OCSP_check_nonce` directly takes the first OCSP nonce in the list to compare, so any newer  nonces are ignored. 
We ultimately decided to follow OpenSSL's faulty implementation and document/test the behavior, there doesn't seem to be much value in diverging from the original and will only introduce more unanticipated behavioral changes.

### Call-outs:
N/A

### Testing:
Extend upon original OCSP nonce tests

### DIFF
--- a/crypto/ocsp/ocsp_test.cc
+++ b/crypto/ocsp/ocsp_test.cc
@@ -1393,8 +1393,17 @@ TEST_P(OCSPNonceTest, OCSPNonce) {
       t.nonce_check_status == OCSP_NONCE_BOTH_ABSENT) {
     EXPECT_EQ(OCSP_copy_nonce(basicResponse.get(), ocspRequest.get()), 2);
     EXPECT_EQ(OCSP_check_nonce(ocspRequest.get(), basicResponse.get()),
-            t.nonce_check_status);
+              t.nonce_check_status);
   } else {
+    // OpenSSL's implementation of |OCSP_copy_nonce| keeps the original nonce in
+    // |resp| at the start of the list. We have to remove the old nonce prior to
+    // copying the new one over.
+    int resp_idx = OCSP_BASICRESP_get_ext_by_NID(basicResponse.get(),
+                                                 NID_id_pkix_OCSP_Nonce, -1);
+    if (resp_idx >= 0) {
+      bssl::UniquePtr<X509_EXTENSION> old_resp_ext(
+          OCSP_BASICRESP_delete_ext(basicResponse.get(), resp_idx));
+    }
     EXPECT_EQ(OCSP_copy_nonce(basicResponse.get(), ocspRequest.get()), 1);
     EXPECT_EQ(OCSP_check_nonce(ocspRequest.get(), basicResponse.get()),
               OCSP_NONCE_EQUAL);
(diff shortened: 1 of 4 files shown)


crypto/ocsp/ocsp_test.cc
```diff
@@ -1385,6 +1385,16 @@ TEST_P(OCSPNonceTest, OCSPNonce) {
   }
   EXPECT_EQ(OCSP_check_nonce(ocspRequest.get(), basicResponse.get()),
             t.nonce_check_status);
+
+  // Check that nonce copying from |req| to |bs| also works as expected.
+  if (t.nonce_check_status == OCSP_NONCE_RESPONSE_ONLY ||
+      t.nonce_check_status == OCSP_NONCE_BOTH_ABSENT) {
+    EXPECT_EQ(OCSP_copy_nonce(basicResponse.get(), ocspRequest.get()), 2);
```
WillChilds-Klein, line 1392: what do we expect `OCSP_check_nonce` to be after this copying no-op?

samuel40791765: The check's behavior will be identical to L1386 since there's no nonce to be copied. It seemed a bit redundant, but I can add the extra check here for parity across the if else statements.

include/openssl/ocsp.h
```diff
@@ -194,6 +194,11 @@ OPENSSL_EXPORT int OCSP_request_add1_nonce(OCSP_REQUEST *req,
 //  but aren't equal.
 OPENSSL_EXPORT int OCSP_check_nonce(OCSP_REQUEST *req, OCSP_BASICRESP *bs);
 
+// OCSP_copy_nonce copies the nonce value (if any) from |req| to |resp|. Returns
+// 1 on success and 0 on failure. If the optional nonce value does not exist in
+// |req|, we return 2 instead.
+OPENSSL_EXPORT int OCSP_copy_nonce(OCSP_BASICRESP *resp, OCSP_REQUEST *req);
```
WillChilds-Klein, line 200: nit: should we document that the nonce extension is prepended?

WillChilds-Klein (dismissed):

include/openssl/ocsp.h
samuel40791765: Ah yes we should

samuel40791765: The more I think about it, I think the correct behavior should be replacing any existing nonce in `resp` instead of prepending/appending it to the list of extensions. OpenSSL appends the nonce to the list, but it doesn't really make sense for a `resp` to have two nonces.
My initial concern was that someone may have noticed OpenSSL's double nonce behavior and they're removing the old extra nonce themselves after the call to `OCSP_copy_nonce`. This might cause our enforced singular nonce to be removed unintentionally. But it's an edge case trying to get around OpenSSL's buggy behavior, enforcing the right behavior might be the better call here.

skmcgrail (dismissed):

include/openssl/ocsp.h
samuel40791765: After discussing with @WillChilds-Klein, ultimately decided to align with OpenSSL's implementation. There's not too much benefit in misaligning here to be fair and creates more unanticipated behavioral changes.

skmcgrail (approved):

PR #1714, opened 2024-07-18
X509toolcomparison

ecdeye:
Description:
Refactored existing x509 tests
Added OpenSSL comparison tests and CI script


tool-openssl/x509_test.cc
```diff
@@ -60,325 +58,328 @@ X509* CreateAndSignX509Certificate() {
...
+  char csr_path[PATH_MAX];
+  char out_path_tool[PATH_MAX];
+  char out_path_openssl[PATH_MAX];
+  char signkey_path[PATH_MAX];
+  bssl::UniquePtr<X509> x509;
+  bssl::UniquePtr<X509_REQ> csr;
+  const char* tool_executable_path;
+  const char* openssl_executable_path;
+  std::string tool_output_str;
+  std::string openssl_output_str;
```
samuel40791765, line 326: Just a minor comment/question, would it be a lot to make the string variable types consistent? It seems like these are all used as `std::string` in the end.

samuel40791765 (commented):
Just some minor comments, but the tests are looking great! Really like how we're testing 5 flavors of OpenSSL and the way the test harness makes it easy to configure.
The new test class abstractions are awesome too, I hadn't realized this was a thing.

tool-openssl/x509_test.cc
```diff
@@ -60,325 +58,328 @@ X509* CreateAndSignX509Certificate() {
...
-// Test x509 -in and -out
-TEST(X509Test, X509ToolTest) {
-  char in_path[PATH_MAX];
-  char out_path[PATH_MAX];
+class X509Test : public ::testing::Test {
+public:
+  static char in_path[PATH_MAX];
+  static char csr_path[PATH_MAX];
+  static char out_path[PATH_MAX];
+  static char signkey_path[PATH_MAX];
```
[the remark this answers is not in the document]
ecdeye: This is leftover from when I was trying different test implementations, removing

tool-openssl/x509_test.cc
ecdeye: `size_t createTempFILEpath(char buffer[PATH_MAX]);` from test_util.cc is why the files were char type, which variables specifically are you referring to?

smittals2 (approved):

tool-openssl/x509_test.cc
samuel40791765: I guess I was asking about all of these since a `char` array/pointer is analogous to a `std::string`. We could keep the original types for the ones that use `createTempFILEpath` if that's the case.
How about `tool_executable_path` and `openssl_executable_path`? I see that those are both converted to `std::string` before being used.

samuel40791765 (approved):

smittals2 (approved):

lilpoozie2005 (approved):

PR #1715, opened 2024-07-19
ec2-test-framework enhancements and graviton 4 testing 

samuel40791765:
### Issues:
Resolves `P131897680`

### Description of changes: 
1. Some of our ec2 instances from the ec2-test-framework were failing to be properly stopped. This adds logic to the job pruner to stop hanging instances.
2. Also parallelizing the test runs. GV2 is significantly slower which causes the run to be much longer. This also makes it much easier for us to extend testing against new test scripts. 
3. Also added graviton 4 testing to the ec2-test-framework.

### Call-outs:
N/A

### Testing:
Local fork CI run: https://github.com/samuel40791765/aws-lc/pull/33

### DIFF
--- a/tests/ci/cdk/cdk/codebuild/ec2_test_framework_omnibus.yaml
+++ b/tests/ci/cdk/cdk/codebuild/ec2_test_framework_omnibus.yaml
@@ -32,3 +32,30 @@ batch:
           EC2_INSTANCE_TYPE: "c6g.2xlarge"
           ECR_DOCKER_TAG: "amazonlinux-2023_clang-15x_sanitizer"
           TARGET_TEST_SCRIPT: "./tests/ci/run_fips_tests.sh"
+
+    # Actual tests are ran on an Graviton4 ec2 instance via SSM Commands.
+    - identifier: graviton4_tests_asan
+      buildspec: ./tests/ci/codebuild/common/run_ec2_target.yml
+      env:
+        type: LINUX_CONTAINER
+        privileged-mode: false
+        compute-type: BUILD_GENERAL1_SMALL
+        image: 620771051181.dkr.ecr.us-west-2.amazonaws.com/aws-lc-docker-images-linux-x86:ubuntu-20.04_clang-7x-bm-framework_latest
+        variables:
+          EC2_AMI: "ami-0c29a2c5cf69b5a9c"
+          EC2_INSTANCE_TYPE: "r8g.2xlarge"
+          ECR_DOCKER_TAG: "amazonlinux-2023_clang-15x_sanitizer"
+          TARGET_TEST_SCRIPT: "./tests/ci/run_posix_sanitizers.sh"
+
+    - identifier: graviton4_tests_fips
+      buildspec: ./tests/ci/codebuild/common/run_ec2_target.yml
+      env:
+        type: LINUX_CONTAINER
+        privileged-mode: false
+        compute-type: BUILD_GENERAL1_SMALL
+        image: 620771051181.dkr.ecr.us-west-2.amazonaws.com/aws-lc-docker-images-linux-x86:ubuntu-20.04_clang-7x-bm-framework_latest
+        variables:
+          EC2_AMI: "ami-0c29a2c5cf69b5a9c"
+          EC2_INSTANCE_TYPE: "r8g.2xlarge"
+          ECR_DOCKER_TAG: "amazonlinux-2023_clang-15x_sanitizer"
+          TARGET_TEST_SCRIPT: "./tests/ci/run_fips_tests.sh"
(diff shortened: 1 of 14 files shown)


tests/ci/cdk/cdk/components.py
```diff
@@ -37,11 +37,28 @@ def __init__(self, scope: Construct, id: str, *, project: codebuild.IProject) ->
...
                                 ],
                                 resources=[project.project_arn]))
 
+        if ec2_permissions:
+            lambda_function.add_to_role_policy(
+                iam.PolicyStatement(effect=iam.Effect.ALLOW,
+                                actions=[
+                                    "ec2:TerminateInstances",
+                                ],
+                                resources=["arn:aws:ec2:{}:{}:instance/*".format(AWS_REGION, AWS_ACCOUNT)]))
```
andrewhop, line 50: Can this be further scoped down to only allow terminating some instances based on the tag https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_ec2-start-stop-tags.html.

tests/ci/lambda/src/bin/purge-stale-builds.rs
```diff
@@ -119,6 +124,58 @@ async fn handle(_event: LambdaEvent<Value>) -> Result<(), Error> {
...
+    if ec2_describe_response.is_err() {
+        log::info!("No IAM Permissions to DescribeInstances");
+        return Ok(());
+    }
+
+    let mut ec2_hanging_instances: Vec<String> = vec![];
+    for reservation in ec2_describe_response.unwrap().reservations() {
+        log::info!("Checking Instance {:?}", reservation.instances());
+        for instance in reservation.instances() {
+            // Prune ec2 instances hanging more than 120 minutes.
```
andrewhop, line 155: Can this also handle cleaning up duplicate instances in the case of second push to the same PR creating a new job and the results of the first job being unused?

tests/ci/cdk/cdk/codebuild/ec2_test_framework_omnibus.yaml
```diff
@@ -18,3 +18,17 @@ batch:
           EC2_AMI: "ami-0c29a2c5cf69b5a9c"
           EC2_INSTANCE_TYPE: "c6g.2xlarge"
           ECR_DOCKER_TAG: "amazonlinux-2023_clang-15x_sanitizer"
+          TARGET_TEST_SCRIPT: "./tests/ci/run_posix_sanitizers.sh"
+
+    - identifier: graviton2_tests_fips
```
andrewhop, line 23: While we're here can you add a Graviton 4 instance as well?

tests/ci/lambda/src/bin/purge-stale-builds.rs
skmcgrail: To add to this, we should probably use tags on the instance to indicate the metadata about what pull request its related to, what the commit head is when you launch it with run-instances, and tag it with something to indicate that it is part of the ec2-test-framework. Then your filter here can be to look for that indicator tag, and then you can use the other tags with the PR number, head commit etc to determine if an instance should be nuked cause its an old push.

samuel40791765: Thanks for the feedback Sean! Yeah it would be a bit more to support this, but doable. 
I was planning on adding the `CODEBUILD_SOURCE_VERSION`(which contains the commit hash) [to the ec2 tags](https://github.com/aws/aws-lc/blob/main/tests/ci/run_ec2_test_framework.sh#L37), then [reuse a bit of your PR commit logic above](https://github.com/aws/aws-lc/blob/main/tests/ci/lambda/src/bin/purge-stale-builds.rs#L68-L73) to terminate the right instances.
I guess that coincides with your proposal, do let me know if there's any part of the puzzle I'm missing though.

skmcgrail (approved):

tests/ci/cdk/cdk/components.py
samuel40791765: Took me a while, but I was tripping over a self inflicted bug with the wrong tag name... I have the policy working now.

tests/ci/cdk/cdk/codebuild/ec2_test_framework_omnibus.yaml
samuel40791765: Added

tests/ci/lambda/src/bin/purge-stale-builds.rs
samuel40791765: I got things working and the ec2 instances are now pruned at the same pace as our Codebuild jobs.

skmcgrail (approved):

andrewhop (approved):

PR #1716, opened 2024-07-22
Check at runtime that the tool is loading the same libcrypto it was built with

andrewhop:
### Description of changes: 
The tool CMakeList was already checking /lib and /lib64 for the different versions of OpenSSH. However, at runtime we set the LD_LIBRARY_PATH to only /lib which doesn't have the OpenSSL 3+ artifacts. The benchmark tool was then finding the system install of OpenSSL which happened to be 3.something and close enough to work. This change is inspired by a similar mechanism OpenSSL's own tool uses to check build/runtime dependencies. 

### Callouts
This uses SSLeay which seems to be the best common way to check at runtime all the library versions we test against. 

### Testing:
Tested locally changing the expected version and observing the tool fail. 


tool/benchmark.cc
```diff
@@ -2,8 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0 OR ISC
 
 #include "internal.h"
+#include <openssl/opensslv.h>
 
 int main(int argc, char **argv) {
+  unsigned long build_version = OPENSSL_VERSION_NUMBER;
+  unsigned long runtime_version = SSLeay();
```
WillChilds-Klein, line 9: are you just concerned with version numbers within a libcrypto "family" (e.g. compiled against OpenSSL 1.1, linked against OpenSSL 3.0) or are you also concerned about accidentally compiling/linking across "families" (e.g. compiled against OpenSSL 1.1, linked against AWS-LC)? the version number may conflict across "families", so you may want to use `OPENSSL_VERSION_TEXT` and `OPENSSL_version` if cross-family linkage is a concern. we [implemented](https://github.com/awslabs/aws-c-cal/pull/186#discussion_r1612497699) a similar check in the CRT recently.

tests/ci/run_benchmark_build_tests.sh
```diff
@@ -107,9 +107,9 @@ LD_LIBRARY_PATH="${install_dir}/aws-lc-fips-2021-10-20/lib" "${BUILD_ROOT}/tool/
 LD_LIBRARY_PATH="${install_dir}/aws-lc-fips-2022/lib" "${BUILD_ROOT}/tool/aws-lc-fips-2022" -timeout_ms 10
 LD_LIBRARY_PATH="${install_dir}/openssl-${openssl_1_0_2_branch}/lib" "${BUILD_ROOT}/tool/open102" -timeout_ms 10
 LD_LIBRARY_PATH="${install_dir}/openssl-${openssl_1_1_1_branch}/lib" "${BUILD_ROOT}/tool/open111" -timeout_ms 10
-LD_LIBRARY_PATH="${install_dir}/openssl-${openssl_3_1_branch}/lib" "${BUILD_ROOT}/tool/open31" -timeout_ms 10
-LD_LIBRARY_PATH="${install_dir}/openssl-${openssl_3_2_branch}/lib" "${BUILD_ROOT}/tool/open32" -timeout_ms 10
-LD_LIBRARY_PATH="${install_dir}/openssl-${openssl_master_branch}/lib" "${BUILD_ROOT}/tool/openmaster" -timeout_ms 10
+LD_LIBRARY_PATH="${install_dir}/openssl-${openssl_3_1_branch}/lib64" "${BUILD_ROOT}/tool/open31" -timeout_ms 10
```
WillChilds-Klein, line 110: we no longer want to check `/lib` at all? do we run these benchmarks on systems with a smaller address space?

WillChilds-Klein (approved):

tests/ci/run_benchmark_build_tests.sh
andrewhop: This isn't how we actually run the benchmarks, this is just a quick sanity test to ensure everything can build and run in our CI.

samuel40791765 (approved):

PR #1719, opened 2024-07-23
Fix for select point from table in ec_nistp scalar_mul

dkostic:
Previous change unified and simplified selecting a point from table by using constant_time_select_entry_from_table_8 function. Unfortunatelly, on x86 this function is significantly slower than constant_time_select_entry_from_table_w. This commit fixes the performance regression.

Measured on Intel(R) Xeon(R) Platinum 8488C:
P-384 scalar_mul from 4091 to 5461 ops/s,
P-521 scalar_mul from 2520 to 3731 ops/s.

### Call-outs:

### Testing:

### DIFF
--- a/crypto/fipsmodule/ec/ec_nistp.c
+++ b/crypto/fipsmodule/ec/ec_nistp.c
@@ -360,16 +360,27 @@ static void generate_table(const ec_nistp_meth *ctx,
   }
 }
 
-// Writes to xyz_out the idx-th point from table in constant-time.
-static void select_point_from_table(const ec_nistp_meth *ctx,
-                                    ec_nistp_felem_limb *xyz_out,
-                                    const ec_nistp_felem_limb *table,
-                                    const size_t idx) {
-  size_t entry_size = 3 * ctx->felem_num_limbs * sizeof(ec_nistp_felem_limb);
+// Writes to out the idx-th point from table in constant-time.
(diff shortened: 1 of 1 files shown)


crypto/fipsmodule/ec/ec_nistp.c
```diff
@@ -360,16 +360,27 @@ static void generate_table(const ec_nistp_meth *ctx,
...
-  size_t entry_size = 3 * ctx->felem_num_limbs * sizeof(ec_nistp_felem_limb);
-
-  constant_time_select_entry_from_table_8(
-          (uint8_t*)xyz_out, (uint8_t*)table,
+// Writes to out the idx-th point from table in constant-time.
+static inline void select_point_from_table(const ec_nistp_meth *ctx,
+                                           ec_nistp_felem_limb *out,
+                                           const ec_nistp_felem_limb *table,
+                                           const size_t idx,
+                                           const size_t projective) {
```
samuel40791765, line 368: Is the only possible value being only `1` right now also temporary?

dkostic: yes, this will be useful for a coming change where the function will be called with `projective` set to 0.

samuel40791765 (approved):

andrewhop (approved):

PR #1721, opened 2024-07-25
Implement BIO_puts and add callback function support to BIO_puts,gets,ctrl

kexgaber:
### Issues: 
CryptoAlg-2463

### Description of changes: 
Currently, `BIO_puts` delegates to the public `BIO_write` API, which may result in unexpected behavior should applications rely on them being distinguishable (e.g., if a callback has certain logic for when write is called versus when puts is called). This change implements BIO_puts as its own function should this method be defined in the application, otherwise, it defaults to calling the "write" function of the type of BIO instantiated.

This also expands on the work done in commit https://github.com/aws/aws-lc/commit/51316840ef473c1e78c0d2a104bc346cda1b81f7 to add support for `callback_ex`s in `BIO_puts`, `BIO_gets`, and `BIO_ctrl`.  If no custom callback_ex function is defined, the default behavior remains unchanged.

### Call-outs:
N/A

### Testing:
- A custom BIO is defined to test cases where BIO_puts is defined
- Additional test cases were added which use a mock callback created in previous testing to test the added callback functionality of BIO_gets, BIO_puts, and BIO_ctrl


crypto/bio/bio.c
```diff
@@ -212,17 +212,26 @@ int BIO_gets(BIO *bio, char *buf, int len) {
     OPENSSL_PUT_ERROR(BIO, BIO_R_UNSUPPORTED_METHOD);
     return -2;
   }
+  int ret = 0;
+  if (HAS_CALLBACK(bio)) {
+    ret = (int)bio->callback_ex(bio,BIO_CB_GETS,buf,len,0, 0L, 1L, NULL);
```
skmcgrail, line 217:
```suggestion
    ret = (int)bio->callback_ex(bio, BIO_CB_GETS, buf, len, 0, 0L, 1L, NULL);
```

crypto/bio/bio_test.cc
```diff
@@ -1095,3 +1095,133 @@ TEST(BIOTest, ReadWriteEx) {
...
+  bssl::UniquePtr<BIO> bio(BIO_new(BIO_s_mem()));
+  ASSERT_TRUE(bio);
+
+  // Test basic puts and read  
+  uint8_t buf[32];
+  EXPECT_EQ(12, BIO_puts(bio.get(), "hello world\n"));
+  EXPECT_EQ(12, BIO_read(bio.get(), buf, sizeof(buf)));
+}
+
+TEST(BIOTest, TestPutsGetsCtrlCallbacks) {
```
skmcgrail, line 1110: Since this ended up getting length with the different assertions I would split this into 3 different test functions covering puts, gets, and ctrl.

crypto/bio/bio.c
```diff
@@ -299,13 +308,42 @@ int BIO_write_all(BIO *bio, const void *data, size_t len) {
...
-  return BIO_write(bio, in, (int)len);
+  // Check for bwrites here since we use that if bputs is NULL
+  if (bio == NULL || bio->method == NULL || (bio->method->bwrite == NULL &&
+                                            bio->method->bputs == NULL)) {
+    OPENSSL_PUT_ERROR(BIO, BIO_R_UNSUPPORTED_METHOD);
+    return -2;
+  }
+  int ret = 0;
+  if(HAS_CALLBACK(bio)) {
+    ret = (int)bio->callback_ex(bio, BIO_CB_PUTS, in, len, 0, 0L, 1L, NULL);
```
skmcgrail, line 325: For matching behavior we should probably pass `0` here for `len`, as that is the OpenSSL behavior, even though we technically know the length in our emulation.

https://github.com/openssl/openssl/blob/07e4d7f4747005e3ce56423182ad047eb05d8e16/crypto/bio/bio_lib.c#L513

crypto/bio/bio.c
```diff
@@ -299,13 +308,42 @@ int BIO_write_all(BIO *bio, const void *data, size_t len) {
...
+  }
+  if (bio->method->bputs != NULL) {
+    ret = bio->method->bputs(bio, in);
+  } else {
+    ret = bio->method->bwrite(bio, in, len);
+  }
+  if (ret > 0) {
+    bio->num_write += ret;
+  }
+  ret = call_bio_callback_with_processed(bio, BIO_CB_PUTS | BIO_CB_RETURN,
```
skmcgrail, line 343: Same base as above here for the `len` argument.

https://github.com/openssl/openssl/blob/07e4d7f4747005e3ce56423182ad047eb05d8e16/crypto/bio/bio_lib.c#L532

crypto/bio/bio.c
```diff
@@ -299,13 +308,42 @@ int BIO_write_all(BIO *bio, const void *data, size_t len) {
 }
 
 int BIO_puts(BIO *bio, const char *in) {
-  size_t len = strlen(in);
+  const size_t len = strlen(in);
```
skmcgrail, line 311: You should be able to move this into else code block on lines 337-339 if you implement the other feedback below.

crypto/bio/bio.c
```diff
@@ -299,13 +308,42 @@ int BIO_write_all(BIO *bio, const void *data, size_t len) {
...
   if (len > INT_MAX) {
     // |BIO_write| and the return value both assume the string fits in |int|.
     OPENSSL_PUT_ERROR(BIO, ERR_R_OVERFLOW);
     return -1;
   }
-  return BIO_write(bio, in, (int)len);
+  // Check for bwrites here since we use that if bputs is NULL
+  if (bio == NULL || bio->method == NULL || (bio->method->bwrite == NULL &&
+                                            bio->method->bputs == NULL)) {
```
skmcgrail, line 319: Can you add a test case that exercises this branch of code?

crypto/bio/bio_test.cc
```diff
@@ -1095,3 +1095,184 @@ TEST(BIOTest, ReadWriteEx) {
...
+  ASSERT_EQ(0, BIO_puts(bio.get(), "hello world"));
+
+  // Test setting new method
+  bssl::UniquePtr<BIO_METHOD> method(BIO_meth_new(0, nullptr));
+  ASSERT_TRUE(method);
+  ASSERT_TRUE(BIO_meth_set_puts(
+    method.get(), [](BIO *b, const char *in) -> int {
+      return 100;
+    }
+  ));
```
skmcgrail, line 1147: Should assert after this call that method->bputs is not null. That will verify that set_puts actually updated the value and just didn't just return 1.

skmcgrail (approved):

andrewhop (approved):

skmcgrail (approved):

lilpoozie2005 (approved):

PR #1726, opened 2024-07-29
Remove custom PKCS7 ASN1 functions, add new structs

WillChilds-Klein:
### Issues:
Resolves #CryptoAlg-2493

### Description of changes: 

This PR is the first in a series implementing Ruby's supported subset of the PKCS7 standard. To do this, we remove AWS-LC's customized ASN.1 serialization logic and delegate to `asn1.h` and `asn1t.h`. We also update the various `PKCS_type_is_*` no-op functions to actually check the input's type.

The PKCS7 tests currently contain a test signedData structure that is encoded using BER. Because our `ASN1_get_object` currently [disallows](https://github.com/aws/aws-lc/blob/8080ce32b118746475c718d14a609bd1325166e1/crypto/asn1/asn1_lib.c#L136) indefinite-length BER, we need to modify `d2i_PKCS7` to detect and convert BER (if present) into DER before parsing. This allows us to retain backwards compatibility with BER-encoded PKCS7 objects. Please see the appendix for more discussion on this topic.

The next PR in this series will implement the various getters, setters, and allocation functions needed to support CRruby with testing for each. Please see PR #1780 for an idea of what that will look like.

### Call-outs:

Source edits conform to OpenSSL's implementation, inlining some of OpenSSL's common subroutines.

Prior struct definitions in public `pkcs7.h` are preserved for backwards compatibility and expanded for the rest of this series. **Whether we can make these existing definitions opaque should be a point of discussion**. New structure definitions are made opaque.

Also, because AWS-LC previously implemented BER/DER parsing semi-manually and cached the encoded value in PKCS7 structs, there may exist performance differences due to the new lack of caching. Similarly, BER parsing is now slightly less efficient because we need to make 2 passes over the input (once to convert from BER to DER, once to parse the DER).

### Testing:
- existing unit tests
- new tests are added in PR #1780 (patch coverage 81% including this PR's commits)

---

### Appendix: Indefinite-length BER in PKCS7

TL;DR -- Reading RFC 2315, it looks like `encrypted` is the only content type should be affected, but the RFC imposes no restrictions around how the top-level ContentInfo SEQUENCE shall be encoded.

Support for indefinite-length BER was [removed](https://boringssl-review.googlesource.com/c/boringssl/+/50287) (see also [here](https://issues.chromium.org/issues/42290221)) in the general ASN.1 parser macros (although BoringSSL has retained support for it in their minimal PKCS7 implementation, as it used the CBS API for parsing).

Per [section 5](https://datatracker.ietf.org/doc/html/rfc2315#section-5) of RFC 2315 defining PKCS7:

```
   The syntax is general enough to support many different content types.
   This document defines six: data, signed data, enveloped data,
   signed-and-enveloped data, digested data, and encrypted data. ...

   There are two classes of content types: base and enhanced.  Content
   types in the base class contain "just data," with no cryptographic
   enhancements. Presently, one content type is in this class, the data
   content type. ...

   The document is designed such that the enhanced content types can be
   prepared in a single pass using indefinite-length BER encoding, and
   processed in a single pass in any BER encoding. Single-pass operation
   is especially helpful if content is stored on tapes, or is "piped"
   from another process. One of the drawbacks of single-pass operation,
   however, is that it is difficult to output a DER encoding in a single
   pass, since the lengths of the various components may not be known in
   advance. Since DER encoding is required by the signed-data, signed-
   and-enveloped data, and digested-data content types, an extra pass
   may be necessary when a content type other than data is the inner
   content of one of those content types.
```

Since indefinite length is forbidden in DER and `data` is not an enhanced content type, the last paragraph excerpted above implies that only `enveloped` and `encrypted` content types can utilize indefinite BER. However, [section 7]() states:

```
        2.   When a ContentInfo value is the inner content of
             signed-data, signed-and-enveloped-data, or digested-data
             content, a message-digest algorithm is applied to the
             contents octets of the DER encoding of the content field.
             When a ContentInfo value is the inner content of
             enveloped-data or signed-and-enveloped-data content, a
             content-encryption algorithm is applied to the contents
             octets of a definite-length BER encoding of the content
             field.
```

So, it would seem that the only content type that can be encoded in indefinite length BER is the `encrypted` type. 

Interestingly, the NSS certificate previously included in our PKCS7 tests contains a `signed` content encoded in indefinite-length BER (note that length value `0x80` [indicates](https://en.wikipedia.org/wiki/X.690#Indefinite_form) indefinite length).

Screenshot 2024-07-30 at 1 05 29 PM

Unfortunately, this appears to be a valid PKCS7 object per the RFC because the RFC only imposes definite-length encoding on _the contents that are operated on cryptographically_, not the top-level structures themselves.

---


include/openssl/pkcs7.h
```diff
@@ -139,11 +126,44 @@ typedef struct {
...
+    STACK_OF(PKCS7_RECIP_INFO) *recipientinfo;
+    STACK_OF(X509_ALGOR) *md_algs;
+    PKCS7_ENC_CONTENT *enc_data;
+    STACK_OF(X509) *cert;
+    STACK_OF(X509_CRL) *crl;
+    STACK_OF(PKCS7_SIGNER_INFO) *signer_info;
+};
+
+DECLARE_ASN1_FUNCTIONS(PKCS7)
+DECLARE_ASN1_FUNCTIONS(PKCS7_ISSUER_AND_SERIAL)
```
samuel40791765, line 151: Ran into something similar with the OCSP ASN1 functions, most consumers consuming the module only work with the top-level structures' ASN.1 functions and don't necessarily consume the ASN.1 functions of the lower-level structures.
I think we could move some of these ASN1 declarations  to `crypto/pkcs7/internal.h`. By only exposing the more necessary top level structures, we could minimize the number of APIs we expose by quite a lot.
Same thing also applies to `PKCS7_ENC_CONTENT` and maybe `PKCS7_SIGN_ENVELOPE` / `PKCS7_DIGEST`/`PKCS7_ENVELOPE`/`PKCS7_ENCRYPT`/`PKCS7_SIGNED`/?

Note: Did a quick search, Ruby doesn't directly consume the ASN.1 functions of the above mentioned structures.

include/openssl/base.h
```diff
@@ -349,6 +349,13 @@ typedef struct md4_state_st MD4_CTX;
 typedef struct md5_state_st MD5_CTX;
 typedef struct ocsp_req_ctx_st OCSP_REQ_CTX;
 typedef struct ossl_init_settings_st OPENSSL_INIT_SETTINGS;
+typedef struct pkcs7_digest_st PKCS7_DIGEST;
+typedef struct pkcs7_enc_content_st PKCS7_ENC_CONTENT;
+typedef struct pkcs7_encrypt_st PKCS7_ENCRYPT;
+typedef struct pkcs7_envelope_st PKCS7_ENVELOPE;
+typedef struct pkcs7_issuer_and_serial_st PKCS7_ISSUER_AND_SERIAL;
```
samuel40791765, line 356: `PKCS7_ISSUER_AND_SERIAL` seems to only be nested internally within the opaque `PKCS7` structures. Maybe we could move the `typedef` to `internal.h` and not expose this externally?

Note: Did a search and Ruby doesn't consume this directly. A lot of these are intertwined with non-opaque structures, so looks like we'll have to expose most.

crypto/pkcs7/pkcs7_x509.c
```diff
@@ -236,133 +236,41 @@ int PKCS7_bundle_CRLs(CBB *out, const STACK_OF(X509_CRL) *crls) {
...
-    return NULL;
-  }
-  *inp = CBS_data(&cbs);
-  if (out != NULL) {
-    PKCS7_free(*out);
-    *out = ret;
-  }
-  return ret;
+int i2d_PKCS7_bio(BIO *bio, const PKCS7 *p7) {
+  return ASN1_item_i2d_bio(ASN1_ITEM_rptr(PKCS7), bio, (void *) p7);
```
samuel40791765, line 248: Is the `void *` cast necessary? Only asking because I don't see the same pattern in other calls to `ASN1_item_d2i_bio`

crypto/pkcs7/pkcs7_asn1.c
```diff
@@ -0,0 +1,186 @@
...
+} ASN1_ADB_END(PKCS7, 0, type, 0, &p7default_tt, NULL);
+
+ASN1_SEQUENCE(PKCS7) = {
+    ASN1_SIMPLE(PKCS7, type, ASN1_OBJECT),
+    ASN1_ADB_OBJECT(PKCS7)
+} ASN1_SEQUENCE_END(PKCS7)
+
+IMPLEMENT_ASN1_ALLOC_FUNCTIONS(PKCS7)
+
+PKCS7 *d2i_PKCS7(PKCS7 **a, const unsigned char **in, long len) {
```
samuel40791765, line 28: Just a minor nit, would it be worth adding a TODO comment here stating that we want to translate all of this CBB/CBS in some distant future

crypto/pkcs7/pkcs7_asn1.c
```diff
@@ -0,0 +1,186 @@
...
+
+ASN1_SEQUENCE(PKCS7_ISSUER_AND_SERIAL) = {
+        ASN1_SIMPLE(PKCS7_ISSUER_AND_SERIAL, issuer, X509_NAME),
+        ASN1_SIMPLE(PKCS7_ISSUER_AND_SERIAL, serial, ASN1_INTEGER)
+} ASN1_SEQUENCE_END(PKCS7_ISSUER_AND_SERIAL)
+
+IMPLEMENT_ASN1_FUNCTIONS(PKCS7_ISSUER_AND_SERIAL)
+
+/* Minor tweak to operation: free up X509 */
+static int ri_cb(int operation, ASN1_VALUE **pval, const ASN1_ITEM *it,
```
samuel40791765, line 97: Interesting callback mechanism, should we also take the chance to improve on OpenSSL's static function naming clearer for here and below?

Minor nit for documentation style for here and below:
```suggestion
// Minor tweak to operation: free up X509.
static int recip_info_cb(int operation, ASN1_VALUE **pval, const ASN1_ITEM *it,
```

samuel40791765 (commented):
Really like the way of handling the backwards compatibility retainment with BER-encoded PKCS7 objects

include/openssl/pkcs7.h
WillChilds-Klein: great idea. i'll move all the sub-structure function declarations into `internal.h`.

include/openssl/base.h
WillChilds-Klein: yep, agreed.

include/openssl/pkcs7.h
WillChilds-Klein: was able to internalize all of the sub-struct ASN.1 declarations, as well as every sub-struct except for SIGNED and SIGNER_INFO. SIGNED is referenced directly in some pre-existing tests, and SIGNER_INFO is part of SIGNED's definiton.

crypto/pkcs7/pkcs7_asn1.c
WillChilds-Klein: I don't think so. Personally, I prefer to keep committed TODO's much more specific and narrower in scope.

crypto/pkcs7/pkcs7_asn1.c
WillChilds-Klein: one goal i had in this was to preserve as much similarity as possible. between this series and OpenSSL's code. that said, this is just (static) function name and comment style, so yeah that works.

crypto/pkcs7/internal.h
```diff
@@ -22,6 +22,84 @@ extern "C" {
...
+DECLARE_ASN1_FUNCTIONS(PKCS7_SIGNER_INFO)
+DECLARE_ASN1_FUNCTIONS(PKCS7_ENC_CONTENT)
+DECLARE_ASN1_FUNCTIONS(PKCS7_ENCRYPT)
+DECLARE_ASN1_FUNCTIONS(PKCS7_ENVELOPE)
+DECLARE_ASN1_FUNCTIONS(PKCS7_DIGEST)
+DECLARE_ASN1_FUNCTIONS(PKCS7_SIGN_ENVELOPE)
+
+DEFINE_STACK_OF(PKCS7)
+DEFINE_STACK_OF(PKCS7_RECIP_INFO)
+DEFINE_STACK_OF(PKCS7_SIGNER_INFO)
```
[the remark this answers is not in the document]
WillChilds-Klein: we didn't `DEFINE_STACK_OF(PKCS7)` previously, so i like the idea of keeping it private until/if we need to expose it.

crypto/pkcs7/internal.h
```diff
@@ -22,6 +22,84 @@ extern "C" {
...
 
 
+typedef struct pkcs7_issuer_and_serial_st PKCS7_ISSUER_AND_SERIAL;
+typedef struct pkcs7_enc_content_st PKCS7_ENC_CONTENT;
+typedef struct pkcs7_recip_info_st PKCS7_RECIP_INFO;
+
+DECLARE_ASN1_FUNCTIONS(PKCS7_ISSUER_AND_SERIAL)
+DECLARE_ASN1_FUNCTIONS(PKCS7_RECIP_INFO)
+DECLARE_ASN1_FUNCTIONS(PKCS7_SIGNED)
+DECLARE_ASN1_FUNCTIONS(PKCS7_SIGNER_INFO)
```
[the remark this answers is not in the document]
WillChilds-Klein: ah, thanks.

include/openssl/pkcs7.h
```diff
@@ -105,27 +106,18 @@ OPENSSL_EXPORT int PKCS7_get_PEM_CRLs(STACK_OF(X509_CRL) *out_crls,
...
+// needed to build cryptography.io and CRuby.
+
+typedef struct pkcs7_st PKCS7;
+typedef struct pkcs7_signed_st PKCS7_SIGNED;
+typedef struct pkcs7_envelope_st PKCS7_ENVELOPE;
+typedef struct pkcs7_sign_envelope_st PKCS7_SIGN_ENVELOPE;
+typedef struct pkcs7_digest_st PKCS7_DIGEST;
+typedef struct pkcs7_encrypt_st PKCS7_ENCRYPT;
+typedef struct pkcs7_recip_info_st PKCS7_RECIP_INFO;
+typedef struct pkcs7_signer_info_st PKCS7_SIGNER_INFO;
```
samuel40791765, line 118: Nit: Should we keep this in `base.h` or do we like it here?

samuel40791765 (dismissed):

include/openssl/pkcs7.h
WillChilds-Klein: yeah, that makes more sense.

samuel40791765 (approved):

crypto/pkcs7/internal.h
```diff
@@ -22,6 +22,79 @@ extern "C" {
...
+DECLARE_ASN1_FUNCTIONS(PKCS7_SIGNED)
+DECLARE_ASN1_FUNCTIONS(PKCS7_ENC_CONTENT)
+DECLARE_ASN1_FUNCTIONS(PKCS7_ENCRYPT)
+DECLARE_ASN1_FUNCTIONS(PKCS7_ENVELOPE)
+DECLARE_ASN1_FUNCTIONS(PKCS7_DIGEST)
+DECLARE_ASN1_FUNCTIONS(PKCS7_SIGN_ENVELOPE)
+
+DEFINE_STACK_OF(PKCS7)
+
+struct pkcs7_sign_envelope_st {
```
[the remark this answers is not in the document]
WillChilds-Klein: yes, good call.

crypto/pkcs7/pkcs7_asn1.c
```diff
@@ -0,0 +1,186 @@
+#include <openssl/pkcs7.h>
```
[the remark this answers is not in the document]
WillChilds-Klein: thanks, will fix.

crypto/pkcs7/pkcs7_asn1.c
```diff
@@ -0,0 +1,186 @@
...
+ASN1_ADB_TEMPLATE(p7default) = ASN1_EXP_OPT(PKCS7, d.other, ASN1_ANY, 0);
+
+ASN1_ADB(PKCS7) = {
+        ADB_ENTRY(NID_pkcs7_data, ASN1_EXP_OPT(PKCS7, d.data, ASN1_OCTET_STRING, 0)),
+        ADB_ENTRY(NID_pkcs7_signed, ASN1_EXP_OPT(PKCS7, d.sign, PKCS7_SIGNED, 0)),
+        ADB_ENTRY(NID_pkcs7_enveloped, ASN1_EXP_OPT(PKCS7, d.enveloped, PKCS7_ENVELOPE, 0)),
+        ADB_ENTRY(NID_pkcs7_signedAndEnveloped, ASN1_EXP_OPT(PKCS7, d.signed_and_enveloped, PKCS7_SIGN_ENVELOPE, 0)),
+        ADB_ENTRY(NID_pkcs7_digest, ASN1_EXP_OPT(PKCS7, d.digest, PKCS7_DIGEST, 0)),
+        ADB_ENTRY(NID_pkcs7_encrypted, ASN1_EXP_OPT(PKCS7, d.encrypted, PKCS7_ENCRYPT, 0))
+} ASN1_ADB_END(PKCS7, 0, type, 0, &p7default_tt, NULL);
```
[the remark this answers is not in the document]
WillChilds-Klein: to use a rough C analogy, it's kind of like a union (`PKCS7`) of structs (`PKCS7_SIGNED`, other `SEQUENCE`s). ADB stands for `ANY DEFINED BY`, which is like ASN.1's `ANY` type, but restricted to a defined subset of possible types. there's a good explanation in sections 5.2 and 5.3 [here](https://luca.ntop.org/Teaching/Appunti/asn1.html).

samuel40791765 (approved):

andrewhop (approved):

Issue #1723, opened 2024-07-25
Abysmal multi-thread performance due to C11 atomics not being used by default


wtarreau:
### Problem:

We at HAProxy have been testing various TLS libraries, and were shocked to notice that AWS-LC had significantly worse performance than OpenSSL 1.1.1 on client-side and TLS resumption and was almost as bad as OpenSSL-3.x(!) in threaded environments due to extreme use of pthread locks. We first noticed this about a year ago with v1.19 but didn't have time to investigate further by then. Now we've assigned more time to this and tested v1.29.0 and v1.32.0 (both of which show strictly identical performance on this point).

In order to overcome a part of OpenSSL's slowness, we had already integrated a pthread lock emulation in haproxy, that exposes much lighter locks ("lorw" for "low overhead rwlock") and that libraries can use transparently. So it was easy for us to try that as well with AWS-LC. It helped quite a bit but the performance collapsed again past ~40 threads and remained significantly below openssl.

A perf flamegraph showed that it was always the refcounting functions that were calling pthread_rwlock. See attached flamegraphs below, first one with standard pthread rwlocks, second one with low overhead rwlocks:

![aws-lc-1 29](https://github.com/user-attachments/assets/1f5c2d35-4463-4d2a-8331-8347ecb17bd9)
![aws-lc-1 29-lorw](https://github.com/user-attachments/assets/11cf9b52-ffa4-4020-83b1-ed276dac0949)

Looking at the code revealed that aside this heavy pthread-based implementation, there is an alternate implementation for C11 compatible compilers, using much cheaper C11 atomics based on fetch-and-add and friends.

We finally found that `CMakeLists.txt` forces `-std=c99` by default when the `CMAKE_C_STANDARD` variable is not set, which forces the fallback to the slow implementation. Just building with `-DCMAKE_C_STANDARD=11` completely solved the problem, making AWS-LC much faster than OpenSSL. Now here's the comparison of the performance of the different libs + locking mechanisms, running on an r8g-8xlarge (Graviton 4, 64 cores):

![graviton4-64t-locking-comparison](https://github.com/user-attachments/assets/70568080-ebbc-45ab-b830-78a84c7b6e65)

### Solution:

[...]

PR #1608, opened 2024-05-28
Fixes for building with `-pedantic`

justsmth:
### Issues:
Relates to: aws/aws-lc-rs#425

### Description of changes: 
* Fix compiler errors that occur when building the "crypto" & "ssl" targets with `-pedantic`.
* Adds CI jobs `gcc-14-pedantic` and `clang-18-pedantic`.

### Call-outs:
* This only verifies that the "crypto" and "ssl" targets can be built.  These are the two targets required by downstream packages, such as `aws-lc-sys`.
*  The `-Wno-c99-c11-compat` (GCC) and `-Wno-c11-extensions` (Clang) are to avoid a pedantic warning related to our use of `_Generic`:
```
/home/justsmth/repos/aws-lc/crypto/asn1/../internal.h:1099:4: error: ISO C99 does not support ‘_Generic’ [-Werror=pedantic]
 1099 |   (_Generic((x),                                    \
      |    ^~~~~~~~
```
* The `-Wno-newline-eof` is needed b/c Clang complains about a missing newline at the end of the s2n-bignum assembly files:
```
/home/justsmth/repos/aws-lc/third_party/s2n-bignum/include/_internal_s2n_bignum.h:17:7: warning: no newline at end of file [-Wnewline-eof]
   17 | #endif
      |       ^
```
* The `-Wno-overlength-strings` parameter was needed to avoid a failure due to string literals in `err-data.c`:
```
...
/tmp/tmp.SO3vhKEAo4/AWS-LC-BUILD/crypto/err_data.c:1603:5: error: string length ‘16178’ is greater than the length ‘4095’ ISO C99 compilers are required to support [-Werror=overlength-strings]
 1603 |     "";
      |     ^~
...
```
* Building our tests with `-pedantic` will still fail due largely to our tests using binary literals:
```
...
Files/crypto_test.dir/endian_test.cc.o -MF crypto/CMakeFiles/crypto_test.dir/endian_test.cc.o.d -o crypto/CMakeFiles/crypto_test.dir/endian_test.cc.o -c /home/justsmth/repos/aws-lc/crypto/endian_test.cc
/home/justsmth/repos/aws-lc/crypto/endian_test.cc:69:20: error: binary constants are a C++14 feature or GCC extension [-Werror]
   69 |   uint32_t value = 0b00000010000000000000000000000;
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #1729, opened 2024-07-31
Enable C11 automatically if the compiler supports it

andrewhop:
### Issues:
Resolves https://github.com/aws/aws-lc/issues/1723

### Description of changes: 
https://github.com/aws/aws-lc/issues/1723 pointed out the default AWS-LC behavior can be very slow because we enable C99 by default. The user has to opt into the better atomics by specifying -DCMAKE_C_STANDARD=11 and they probably won't find that option. This change turns C11 on by default if the compiler supports it.

Add a new benchmark to test our ref count performance.

### Call-outs:
If users do not change their CMake options those with modern compilers will get C11 atomics automatically, legacy customers will continue with the pthread implementation. Both legacy and modern users can continue to manually control this by setting the CMAKE_C_STANDARD option. This option is only available to CMake >= 3.1 customers. 

### Testing:
Tried different options and verified the build.ninja included the expected `-std=gnu99` or `-std=gnu11`. Performance numbers on an Graviton 4 R8g show a 1.6 times increase for single threaded uncontested locking performance, and **33.0 times** increase in heavily contested locking performance. depending on the use case real world applications will see less of an increase. Each thread attempts to increment the same `CRYPTO_refcount_t` 1,000 times:
C99 (previous performance):
```
cmake -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_STANDARD=99 ../
./tool/bssl speed -filter CRYPTO_refcount_inc -timeout 10
Did 225000 CRYPTO_refcount_inc 1000 iterations with 1 threads operations in 10043601us (22402.3 ops/sec)
Did 78861 CRYPTO_refcount_inc 1000 iterations with 2 threads operations in 10008608us (7879.3 ops/sec)
Did 26448 CRYPTO_refcount_inc 1000 iterations with 4 threads operations in 10032376us (2636.3 ops/sec)
Did 8262 CRYPTO_refcount_inc 1000 iterations with 8 threads operations in 10035344us (823.3 ops/sec)
Did 2520 CRYPTO_refcount_inc 1000 iterations with 16 threads operations in 10048296us (250.8 ops/sec)
Did 684 CRYPTO_refcount_inc 1000 iterations with 32 threads operations in 10079403us (67.9 ops/sec)
Did 218 CRYPTO_refcount_inc 1000 iterations with 64 threads operations in 10015418us (21.8 ops/sec)
```

C11 (current performance):
```
cmake -GNinja -DCMAKE_BUILD_TYPE=Release ../ 
./tool/bssl speed -filter CRYPTO_refcount_inc -timeout 10
Did 372000 CRYPTO_refcount_inc 1000 iterations with 1 threads operations in 10002129us (37192.1 ops/sec)
Did 268000 CRYPTO_refcount_inc 1000 iterations with 2 threads operations in 10022289us (26740.4 ops/sec)
Did 168000 CRYPTO_refcount_inc 1000 iterations with 4 threads operations in 10022843us (16761.7 ops/sec)
Did 74943 CRYPTO_refcount_inc 1000 iterations with 8 threads operations in 10100478us (7419.7 ops/sec)
Did 32550 CRYPTO_refcount_inc 1000 iterations with 16 threads operations in 10037521us (3242.8 ops/sec)
Did 15246 CRYPTO_refcount_inc 1000 iterations with 32 threads operations in 10065910us (1514.6 ops/sec)
Did 7208 CRYPTO_refcount_inc 1000 iterations with 64 threads operations in 10042097us (717.8 ops/sec)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,16 @@ if(CMAKE_VERSION VERSION_GREATER "3.1.0")
   endif()
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
   if (NOT DEFINED CMAKE_C_STANDARD)
-    set(CMAKE_C_STANDARD 99)
+    try_compile(
+            RESULT
+            ${PROJECT_BINARY_DIR}
+            SOURCES "${CMAKE_CURRENT_LIST_DIR}/tests/compiler_features_tests/simple_main.c"
+            C_STANDARD 11)
+    if(RESULT)
+      set(CMAKE_C_STANDARD 11)
+    else()
+      set(CMAKE_C_STANDARD 99)
+    endif()
   endif ()
   set(CMAKE_C_STANDARD_REQUIRED ON)
 endif ()
@@ -365,8 +374,11 @@ if(GCC OR CLANG)
   # Note clang-cl is odd and sets both CLANG and MSVC. We base our configuration
   # primarily on our normal Clang one.
   if (NOT CMAKE_C_STANDARD)
+    message(STATUS "Adding CMAKE_C_FLAGS -std=c99")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99")
-  endif ()
+  else()
+    message(STATUS "Setting CMAKE_C_STANDARD=${CMAKE_C_STANDARD}")
+  endif()
 
   # TODO(CryptoAlg-759): enable '-Wpedantic' if awslc has to follow c99 spec.
   if(CLANG OR (GCC AND CMAKE_C_COMPILER_VERSION VERSION_GREATER "4.1.3"))
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,7 +75,7 @@ if (NOT DEFINED CMAKE_C_STANDARD)
   try_compile(
           RESULT
           ${PROJECT_BINARY_DIR}
-          SOURCES "${CMAKE_CURRENT_LIST_DIR}/tests/compiler_features_tests/c11_atomics.c"
+          SOURCES "${CMAKE_CURRENT_LIST_DIR}/tests/compiler_features_tests/c11.c"
           COMPILE_DEFINITIONS -c -std=c11)
   if(RESULT)
     set(CMAKE_C_STANDARD 11)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,16 +73,7 @@ if(CMAKE_VERSION VERSION_GREATER "3.1.0")
   endif()
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
   if (NOT DEFINED CMAKE_C_STANDARD)
-    try_compile(
-            RESULT
-            ${PROJECT_BINARY_DIR}
-            SOURCES "${CMAKE_CURRENT_LIST_DIR}/tests/compiler_features_tests/simple_main.c"
-            CMAKE_FLAGS "-DCMAKE_C_FLAGS=\"-std=c11\"")
-    if(RESULT)
-      set(CMAKE_C_STANDARD 11)
-    else()
-      set(CMAKE_C_STANDARD 99)
-    endif()
+    set(CMAKE_C_STANDARD 99)
   endif ()
   set(CMAKE_C_STANDARD_REQUIRED ON)
 endif ()
@@ -374,11 +365,8 @@ if(GCC OR CLANG)
   # Note clang-cl is odd and sets both CLANG and MSVC. We base our configuration
   # primarily on our normal Clang one.
   if (NOT CMAKE_C_STANDARD)
-    message(STATUS "Adding CMAKE_C_FLAGS -std=c99")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99")
-  else()
-    message(STATUS "Setting CMAKE_C_STANDARD=${CMAKE_C_STANDARD}")
-  endif()
+  endif ()
 
   # TODO(CryptoAlg-759): enable '-Wpedantic' if awslc has to follow c99 spec.
   if(CLANG OR (GCC AND CMAKE_C_COMPILER_VERSION VERSION_GREATER "4.1.3"))
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,16 @@ if(CMAKE_VERSION VERSION_GREATER "3.1.0")
   endif()
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
   if (NOT DEFINED CMAKE_C_STANDARD)
-    set(CMAKE_C_STANDARD 99)
+    try_compile(
+            RESULT
+            ${PROJECT_BINARY_DIR}
+            SOURCES "${CMAKE_CURRENT_LIST_DIR}/tests/compiler_features_tests/simple_main.c"
+            CMAKE_FLAGS "-DCMAKE_C_FLAGS=\"-std=c11\"")
+    if(RESULT)
+      set(CMAKE_C_STANDARD 11)
+    else()
+      set(CMAKE_C_STANDARD 99)
+    endif()
   endif ()
   set(CMAKE_C_STANDARD_REQUIRED ON)
 endif ()
@@ -365,8 +374,11 @@ if(GCC OR CLANG)
   # Note clang-cl is odd and sets both CLANG and MSVC. We base our configuration
   # primarily on our normal Clang one.
   if (NOT CMAKE_C_STANDARD)
+    message(STATUS "Adding CMAKE_C_FLAGS -std=c99")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99")
-  endif ()
+  else()
+    message(STATUS "Setting CMAKE_C_STANDARD=${CMAKE_C_STANDARD}")
+  endif()
 
   # TODO(CryptoAlg-759): enable '-Wpedantic' if awslc has to follow c99 spec.
   if(CLANG OR (GCC AND CMAKE_C_COMPILER_VERSION VERSION_GREATER "4.1.3"))
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,7 @@ if(CMAKE_VERSION VERSION_GREATER "3.1.0")
             RESULT
             ${PROJECT_BINARY_DIR}
             SOURCES "${CMAKE_CURRENT_LIST_DIR}/tests/compiler_features_tests/simple_main.c"
-            C_STANDARD 11)
+            CMAKE_FLAGS "-DCMAKE_C_FLAGS=\"-std=c11\"")
     if(RESULT)
       set(CMAKE_C_STANDARD 11)
     else()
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -355,8 +364,11 @@ if(GCC OR CLANG)
   # Note clang-cl is odd and sets both CLANG and MSVC. We base our configuration
   # primarily on our normal Clang one.
   if (NOT CMAKE_C_STANDARD)
+    message(STATUS "Adding CMAKE_C_FLAGS -std=c99")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99")
```
nebeid, line 368: Do we want to try and compile with c11 here too?

justsmth (dismissed):

CMakeLists.txt
```diff
@@ -68,7 +68,16 @@ if(CMAKE_VERSION VERSION_GREATER "3.1.0")
...
+    try_compile(
+            RESULT
+            ${PROJECT_BINARY_DIR}
+            SOURCES "${CMAKE_CURRENT_LIST_DIR}/tests/compiler_features_tests/simple_main.c"
+            CMAKE_FLAGS "-DCMAKE_C_FLAGS=\"-std=c11\"")
+    if(RESULT)
+      set(CMAKE_C_STANDARD 11)
+    else()
+      set(CMAKE_C_STANDARD 99)
+    endif()
```
justsmth, line 80: NP: This block is conditioned on `CMAKE_VERSION VERSION_GREATER "3.1.0"` but I don't think that's necessary for testing compiler C11 compatibility. (We could still set the `CMAKE_C_STANDARD` and `CMAKE_CXX_STANDARD` values even if they will be ignored by CMake.)

Instead we could condition the `set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99")` line below (and a C11 analog) on the `CMAKE_VERSION` being at least 3.1.

justsmth:
Should `simple_main.c` instead be a little more targetted for our usage of atomics. I'd like to make a similar change for aws-lc-rs, but ran into the following [compile error](https://github.com/aws/aws-lc-rs/actions/runs/10182686006/job/28165868488?pr=475#step:8:1267) when compiling for `arm-linux-androideabi` (using [Clang 14.0.6](https://github.com/aws/aws-lc-rs/actions/runs/10182686006/job/28165868488?pr=475#step:8:1066)).

```
  /home/runner/work/aws-lc-rs/aws-lc-rs/aws-lc-sys/aws-lc/crypto/refcount_c11.c:39:23: error: large atomic operation may incur significant performance penalty; the access size (4 bytes) exceeds the max lock-free size (0  bytes) [-Werror,-Watomic-alignment]
    uint32_t expected = atomic_load(count);
                        ^
```

justsmth:
I think we'll need to add a condition on [`ATOMIC_INT_LOCK_FREE == 2`](https://en.cppreference.com/w/c/atomic/ATOMIC_LOCK_FREE_consts) to the macro definition [here](https://github.com/aws/aws-lc/blob/9431f99ce2a2da97c01de213664f8699b56dbaa0/crypto/internal.h#L571)
```c
#if !defined(OPENSSL_C11_ATOMIC) && defined(OPENSSL_THREADS) &&   \
    !defined(__STDC_NO_ATOMICS__) && defined(__STDC_VERSION__) && \
    __STDC_VERSION__ >= 201112L && ATOMIC_INT_LOCK_FREE == 2
#define OPENSSL_C11_ATOMIC 
#endif
```

CMakeLists.txt
andrewhop: This is kind of a weird fallback branch that should only happen for cmake versions between 3.0 and 3.1 which don't support CMAKE_C_STANDARD.

andrewhop:
> Should `simple_main.c` instead be a little more targetted for our usage of atomics. I'd like to make a similar change for aws-lc-rs, but ran into the following [compile error](https://github.com/aws/aws-lc-rs/actions/runs/10182686006/job/28165868488?pr=475#step:8:1267) when compiling for `arm-linux-androideabi` (using [Clang 14.0.6](https://github.com/aws/aws-lc-rs/actions/runs/10182686006/job/28165868488?pr=475#step:8:1066)).
> 
> ```
>   /home/runner/work/aws-lc-rs/aws-lc-rs/aws-lc-sys/aws-lc/crypto/refcount_c11.c:39:23: error: large atomic operation may incur significant performance penalty; the access size (4 bytes) exceeds the max lock-free size (0  bytes) [-Werror,-Watomic-alignment]
>     uint32_t expected = atomic_load(count);
>                         ^
> ```

Hmmmm, it looks like our atomic version can't be enabled with just C11 then, I think we need two checks and flags:
1. Use C99 or C11
2. If we're using C11 can we use atomics

wtarreau:
> Hmmmm, it looks like our atomic version can't be enabled with just C11 then, I think we need two checks and flags:
> 
>     1. Use C99 or C11
> 
>     2. If we're using C11 can we use atomics

Andrew, please excuse me for insisting, but as I already suggested, why not just rely on the compiler version in the source code instead of all the cmake machinery ? The atomics are not dependent on C11 but on (clang || gcc >= 4.7). That's much easier IMHO to just fix the condition in the code itself rather than risk to break stuff by adding complex version detection. It would even allow to keep the -std=c99 for now if it turns out to have some uses.

andrewhop:
> The atomics are not dependent on C11 but on (clang || gcc >= 4.7).

Ahhh, I missed that part, doing an experiment locally confirms that it does compile as expected, let's see if it gets through all the CI compilers.

andrewhop:
Just using atomics if the compiler supports it does not work because we enabled `-Wno-c99-c11-compat`/`-Wno-c11-extensions` to fix a pedantic error in https://github.com/aws/aws-lc/pull/1608. I agree with the spirit of that error message: if we're setting the standard to C99 we shouldn't be using C11 features. I will try and think of a more sane way to enable C11 in code/cmake.

wtarreau:
OK, I can understand, of course!

And the other option, why not simply *drop* the forced -std=c99 ? In which case exactly is it required to manually force and older version of the standard when 99% of the modern compilers will automatically default to the newer ones ?

andrewhop:
Hit a fun issue, [GCC 4.8 actually doesn't support atomics](https://github.com/aws/aws-lc/actions/runs/10204862260/job/28234472435?pr=1729#step:4:108) even though it supports C11. This seems to be a [known issue](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58016), you might want to increase your check to >= [4.9 which is when support and the header file got included](https://gcc.gnu.org/gcc-4.9/changes.html).

wtarreau:
I'm using them there and I can assure you they work pretty well:
```
2$ cat atomic.c
unsigned atomic_inc(unsigned *a)
{
        return __atomic_fetch_add(a, 1, __ATOMIC_SEQ_CST);
}
-bash-4.2$ gcc -Os -c atomic.c 
-bash-4.2$ objdump  -dr atomic.o

atomic.o:     file format elf64-x86-64


Disassembly of section .text:

0000000000000000 <atomic_inc>:
   0:   b8 01 00 00 00          mov    $0x1,%eax
   5:   f0 0f c1 07             lock xadd %eax,(%rdi)
   9:   c3                      retq   
-bash-4.2$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/lto-wrapper
Target: x86_64-redhat-linux
Configured with: ../configure --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.redhat.com/bugzilla --enable-bootstrap --enable-shared --enable-threads=posix --enable-checking=release --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-linker-hash-style=gnu --enable-languages=c,c++,objc,obj-c++,java,fortran,ada,go,lto --enable-plugin --enable-initfini-array --disable-libgcj --with-isl=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/isl-install --with-cloog=/builddir/build/BUILD/gcc-4.8.5-20150702/obj-x86_64-redhat-linux/cloog-install --enable-gnu-indirect-function --with-tune=generic --with-arch_32=x86-64 --build=x86_64-redhat-linux
Thread model: posix
gcc version 4.8.5 20150623 (Red Hat 4.8.5-44) (GCC) 
```
What might possibly be missing is stdatomic or some defines, but I still don't understand why these would be needed at all. I've never been aware of them and have been using atomics  on gcc and clang for many years now.

I really think that simply dropping -std=c99 which breaks the test and triggers your warning, and checking on gcc version should do the trick. I don't understand why there seems to be some complicated checks that strive to force certain version and later to check them, this might have been inherited from older legacy code maybe, but I really feel that there's some form of inconsistency there that is self-feeding and not necessary at all.

wtarreau:
Or maybe if you want, for non-gcc compilers that might require stdatomics, you could do something like this:
```
# if !defined(__ATOMIC_RELAXED) && defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
#include <stdatomic.h>
#endif
```
I.e. it wlil consider the C standard version for compilers that do not define the ATOMIC macros and that might require stdatomic.h (if there are any at all, but given that gcc and clang have them builtin, that would only be for possibly other ones). E.g. that's gcc 4.8 below:
```
$ gcc -dM -xc -E - < /dev/null |grep __ATOMIC
#define __ATOMIC_ACQUIRE 2
#define __ATOMIC_HLE_RELEASE 131072
#define __ATOMIC_HLE_ACQUIRE 65536
#define __ATOMIC_RELAXED 0
#define __ATOMIC_CONSUME 1
#define __ATOMIC_SEQ_CST 5
#define __ATOMIC_ACQ_REL 4
#define __ATOMIC_RELEASE 3
```

andrewhop:
Poking around we need stdatomic.h to get access to `ATOMIC_LONG_LOCK_FREE` to handle this issue https://github.com/aws/aws-lc/pull/1729#issuecomment-2260789014.

andrewhop:
> I really think that simply dropping -std=c99 which breaks the test and triggers your warning, and checking on gcc version should do the trick

We have some active customers that still rely on very old versions of GCC and C99. Trust me I am very excited for the day when they update and we can drop a lot of this complicated build options.

wtarreau:
Then maybe it's easier to detect older compilers and force std=c99 only for them if that's the purpose ?

wtarreau:
Regrading stdatomic.h, on gcc you have the ATOMIC_LONG_LOCK_FREE definition by default without including anything. If stdatomic is needed for clang, what about including only with clang, since all of its versions support C99 ?

andrewhop:
I figured out why I didn't see any performance improvement on some platforms with this change:
* Thread T1 is spinning up N other threads to call CRYPTO_refcount_inc X times
* When a new thread T2 is created it immediately starts calling CRYPTO_refcount_inc X times
*By the time the T1 had created the next thread T3, T2 was already done so there was no contention
The solution was increasing X from 100 to 1,000 so it takes longer for each thread to finish. TLDR yes C11 atomics are much faster than pthread

justsmth (approved):

smittals2 (approved):

justsmth (approved):

nebeid (approved):

lilpoozie2005 (approved):

PR #1732, opened 2024-08-01
add OCSP_response_create and OCSP_basic_add1_status

samuel40791765:
### Issues:
Addresses `CryptoAlg-2420`

### Description of changes: 
Add support for `OCSP_response_create` and `OCSP_basic_add1_status`.  As we've noted before, these functions are used as OCSP responder functions and Ruby happens to depend on them in their OCSP module.
This builds upon the existing ASN.1 infrastructure we already have for creating OCSP structures. The two APIs allow for the user to specify relevant fields while creating the OCSP structures.

### Call-outs:
New error codes were introduced, 1 from OpenSSL (`OCSP_R_NO_REVOKED_TIME`) and 1 new code that is used when an invalid field value is used (`OCSP_R_UNKNOWN_FIELD_VALUE`).

### Testing:
New tests for both function


include/openssl/ocsp.h
```diff
@@ -21,6 +21,33 @@ extern "C" {
...
+//        aACompromise            (10) }
+//
+// Reason Code RFC: https://www.rfc-editor.org/rfc/rfc5280#section-5.3.1
+#define OCSP_REVOKED_STATUS_UNSPECIFIED 0
+#define OCSP_REVOKED_STATUS_KEYCOMPROMISE 1
+#define OCSP_REVOKED_STATUS_CACOMPROMISE 2
+#define OCSP_REVOKED_STATUS_AFFILIATIONCHANGED 3
+#define OCSP_REVOKED_STATUS_SUPERSEDED 4
+#define OCSP_REVOKED_STATUS_CESSATIONOFOPERATION 5
+#define OCSP_REVOKED_STATUS_CERTIFICATEHOLD 6
```
justsmth, line 47: Should we go ahead an define a constant for `7`? 
```c++
// The value 7 is not an assigned to a status
#define OCSP_UNASSIGNED_REVOKED_STATUS 7
```
Since this is the "internal" header, consumers won't have visibility to these constants.

samuel40791765: I had to move these to the external header, so they'll have visibility to the rest. But I'll add this to the internal header, definitely better than just using the magic number 7.

crypto/ocsp/ocsp_server.c
```diff
@@ -70,3 +70,148 @@ int OCSP_basic_add1_cert(OCSP_BASICRESP *resp, X509 *cert) {
...
+    return NULL;
+  }
+
+  OCSP_SINGLERESP *single = OCSP_SINGLERESP_new();
+  if (single == NULL) {
+    goto err;
+  }
+
+  // Init |resp->tbsResponseData->responses| if NULL.
+  if (resp->tbsResponseData->responses == NULL) {
```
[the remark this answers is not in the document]
samuel40791765: good catch, added another `GUARD_PTR`

justsmth (approved):

crypto/ocsp/ocsp_server.c
```diff
@@ -70,3 +70,149 @@ int OCSP_basic_add1_cert(OCSP_BASICRESP *resp, X509 *cert) {
   X509_up_ref(cert);
   return 1;
 }
+
+OCSP_SINGLERESP *OCSP_basic_add1_status(OCSP_BASICRESP *resp, OCSP_CERTID *cid,
+                                        int status, int revoked_reason,
+                                        ASN1_TIME *revoked_time,
+                                        ASN1_TIME *this_update,
+                                        ASN1_TIME *next_update) {
+  GUARD_PTR(resp);
```
torben-hansen, line 79: That this works is basically an implicit application of the C99 spec. Not very descriptive.

crypto/ocsp/ocsp_server.c
```diff
@@ -70,3 +70,149 @@ int OCSP_basic_add1_cert(OCSP_BASICRESP *resp, X509 *cert) {
...
+err:
+  OCSP_SINGLERESP_free(single);
+  return NULL;
+}
+
+OCSP_RESPONSE *OCSP_response_create(int status, OCSP_BASICRESP *bs) {
+  if (status < OCSP_RESPONSE_STATUS_SUCCESSFUL ||
+      status > OCSP_RESPONSE_STATUS_UNAUTHORIZED ||
+      // 4 is not a valid response status code.
+      status == 4) {
```
torben-hansen, line 187: another magic value

crypto/ocsp/ocsp_server.c
```diff
@@ -70,3 +70,149 @@ int OCSP_basic_add1_cert(OCSP_BASICRESP *resp, X509 *cert) {
...
+  if (bs == NULL) {
+    // |bs| is allowed to be NULL.
+    return rsp;
+  }
+
+  rsp->responseBytes = OCSP_RESPBYTES_new();
+  if (rsp->responseBytes == NULL) {
+    goto err;
+  }
+  rsp->responseBytes->responseType = OBJ_nid2obj(NID_id_pkix_OCSP_basic);
```
torben-hansen, line 208: At some point, not now, RFC references would be nice to have to quickly being able to review whether this is the correctly assigned value or not.

crypto/ocsp/ocsp_test.cc
```diff
@@ -644,6 +647,84 @@ TEST(OCSPTest, BasicAddCert) {
             cert.get());
 }
 
+TEST(OCSPTest, BasicAddStatus) {
```
torben-hansen, line 650: Any cherry-picked checks on the second mode of `OCSP_basic_add1_status()`?
> // 2. If |status| has the value of either |V_OCSP_CERTSTATUS_GOOD| or
// |V_OCSP_CERTSTATUS_UNKNOWN|, |revoked_reason| and |revoked_time| are ignored.

samuel40791765: It's ignored, so it'll still succeed so I hadn't thought of adding an additional check here. But will add.

crypto/ocsp/ocsp_server.c
samuel40791765: I remember this but had thought this was part of another PR haha.. Fixed!

crypto/ocsp/ocsp_server.c
samuel40791765: I can change this back to a regular null check if we prefer that for this sort

crypto/ocsp/ocsp_server.c
samuel40791765: No worries at all, I happen to have the RFC reference to this at hand.

torben-hansen (dismissed):

justsmth (dismissed):

justsmth (approved):

lilpoozie2005 (approved):

PR #1742, opened 2024-08-05
add support and tests for OCSP_basic_sign

samuel40791765:
### Issues:
Addresses `CryptoAlg-2420`

### Description of changes: 
`OCSP_basic_sign` is mandatory for OCSP responders as the RFC specifies that all OCSP response SHALL be digitally signed. This is different from `OCSP_request_sign` since signing the OCSP request was optional.
`OCSP_basic_sign` isn't too complex however, the API does a basic ASN.1 sign along with setting some fields. The slight complications involved are with the additional flag customizations exposed by OpenSSL. Ruby also exposes these flags, so I've taken them in and documented them. I also took the chance to rephrase the documentation around `OCSP_request_sign`.

### Call-outs:
N/A

### Testing:
Reused the certs and keys used for testing `OCSP_request_sign`. What's different from `OCSP_request_sign` is that we create our own `OCSP_BASICRESP` instead of loading it from a file. Creating one from scratch makes more sense as all OCSP response files already have an existing signature.


crypto/ocsp/ocsp_test.cc
```diff
@@ -954,6 +952,141 @@ TEST_P(OCSPRequestTest, OCSPRequestSign) {
...
+          .c_str()));
+  ASSERT_TRUE(rsa);
+  ASSERT_TRUE(EVP_PKEY_set1_RSA(pkey.get(), rsa.get()));
+
+  // Call |OCSP_basic_sign| with no flags and check the expected output.
+  EXPECT_TRUE(OCSP_basic_sign(basic_response.get(), signer_cert.get(),
+                              pkey.get(), EVP_sha256(), additional_cert.get(),
+                              0));
+  EXPECT_TRUE(
+      ASN1_TIME_check(basic_response.get()->tbsResponseData->producedAt));
```
andrewhop, line 1052: Can this check anything else about the time? Such as it's within the current second/minute?

crypto/ocsp/ocsp_test.cc
```diff
@@ -954,6 +952,141 @@ TEST_P(OCSPRequestTest, OCSPRequestSign) {
...
+            0);
+
+  // Check expected effects of |OCSP_NOTIME|.
+  basic_response.reset(OCSP_BASICRESP_new());
+  ASSERT_TRUE(basic_response);
+  EXPECT_TRUE(OCSP_basic_sign(basic_response.get(), signer_cert.get(),
+                              pkey.get(), EVP_sha256(), additional_cert.get(),
+                              OCSP_NOTIME));
+  EXPECT_FALSE(
+      ASN1_TIME_check(basic_response.get()->tbsResponseData->producedAt));
```
andrewhop, line 1070: What does NOTIME do? Does it put a zero or null into the response?

crypto/ocsp/ocsp_test.cc
```diff
@@ -954,6 +952,141 @@ TEST_P(OCSPRequestTest, OCSPRequestSign) {
...
+
+  // Check expected effects of |OCSP_RESPID_KEY|.
+  basic_response.reset(OCSP_BASICRESP_new());
+  ASSERT_TRUE(basic_response);
+  EXPECT_TRUE(OCSP_basic_sign(basic_response.get(), signer_cert.get(),
+                              pkey.get(), EVP_sha256(), additional_cert.get(),
+                              OCSP_RESPID_KEY));
+  EXPECT_EQ(basic_response.get()->tbsResponseData->responderId->type,
+            V_OCSP_RESPID_KEY);
```
andrewhop, line 1079: Besides just the type can we check that the right value is here?

crypto/ocsp/ocsp_test.cc
```diff
@@ -954,6 +952,141 @@ TEST_P(OCSPRequestTest, OCSPRequestSign) {
...
+  ASSERT_TRUE(EVP_PKEY_set1_RSA(pkey.get(), rsa.get()));
+
+  // Call |OCSP_basic_sign| with no flags and check the expected output.
+  EXPECT_TRUE(OCSP_basic_sign(basic_response.get(), signer_cert.get(),
+                              pkey.get(), EVP_sha256(), additional_cert.get(),
+                              0));
+  EXPECT_TRUE(
+      ASN1_TIME_check(basic_response.get()->tbsResponseData->producedAt));
+  EXPECT_EQ(basic_response.get()->tbsResponseData->responderId->type,
+            V_OCSP_RESPID_NAME);
```
andrewhop, line 1054: Can we check the value?

crypto/ocsp/ocsp_test.cc
samuel40791765: Added a time check for a 2 hour time frame

crypto/ocsp/ocsp_test.cc
```diff
@@ -954,6 +952,141 @@ TEST_P(OCSPRequestTest, OCSPRequestSign) {
...
+                        .c_str())
+            .c_str()));
+    ASSERT_TRUE(ecdsa);
+    ASSERT_TRUE(EVP_PKEY_set1_EC_KEY(pkey.get(), ecdsa.get()));
+  }
+
+  // Do the actual sign.
+  EXPECT_EQ(OCSP_basic_sign(basic_response.get(), signer_cert.get(), pkey.get(),
+                            t.dgst, additional_cert.get(), 0),
+            t.expected_sign_status);
```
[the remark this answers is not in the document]
andrewhop: I meant does this test actually test any of the behavior of the function or just the return code. As an extreme example this test would pass if the function looked like:
```
int OCSP_basic_sign(OCSP_BASICRESP *resp, X509 *signer, EVP_PKEY *key,
                    const EVP_MD *dgst, STACK_OF(X509) *certs,
                    unsigned long flags) {
    return resp != nullptr;
}
```

Using OCSP_basic_verify ensures it was actually signed, or that both sign and verify are broken equally which is less likely.

crypto/ocsp/ocsp_server.c
```diff
@@ -70,3 +70,125 @@ int OCSP_basic_add1_cert(OCSP_BASICRESP *resp, X509 *cert) {
...
+  if (byKey == NULL) {
+    return 0;
+  }
+
+  if (!ASN1_OCTET_STRING_set(byKey, digest, SHA_DIGEST_LENGTH)) {
+    ASN1_OCTET_STRING_free(byKey);
+    return 0;
+  }
+  respid->type = V_OCSP_RESPID_KEY;
+  respid->value.byKey = byKey;
```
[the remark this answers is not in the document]
andrewhop: Does the existing OCSP_RESPID already cleanup `byKey` if it is set?

crypto/ocsp/ocsp_test.cc
```diff
@@ -954,6 +952,141 @@ TEST_P(OCSPRequestTest, OCSPRequestSign) {
...
+                                          std::string(t.signer_cert) + ".pem")
+                                  .c_str())
+                      .c_str()));
+  ASSERT_TRUE(signer_cert);
+  bssl::UniquePtr<STACK_OF(X509)> additional_cert(CertChainFromPEM(
+      GetTestData(std::string("crypto/ocsp/test/aws/ca_cert.pem").c_str())
+          .c_str()));
+  ASSERT_TRUE(additional_cert);
+  bssl::UniquePtr<EVP_PKEY> pkey(EVP_PKEY_new());
+  if (std::string(t.private_key) == "server_key") {
```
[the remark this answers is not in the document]
andrewhop: Fine waiting for now.

crypto/ocsp/ocsp_test.cc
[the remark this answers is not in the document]
samuel40791765: Sure, but there's [a rather high criteria](https://github.com/aws/aws-lc/pull/1742/files#diff-b9ae1425ebda6cdd3cfe54e89fdf3feb3f368a34b662d701bf38186d98ed039cR113-R191) for `OCSP_basic_sign` to actually return `1` on success considering all the checks and hurdles it has to go through. 
This parameterized test also has coverage with success cases when signing with the right certs and multiple error cases when attempting to sign with certificate key mismatches: https://github.com/aws/aws-lc/pull/1742/files#diff-b21542ac76bf0a11ad1ad8fdd01d9eab463a9fa814e94b5e494233732a8db602R962-R975
Here's a link to [the coverage report](https://app.codecov.io/gh/aws/aws-lc/pull/1742?src=pr&el=tree&filepath=crypto%2Focsp%2Focsp_server.c&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws) if we want to see which operations have been properly checked for this test.

I've added `OCSP_basic_verify` so we could check that the signature can be verified, but still I don't think this test would easily pass even without the additional verification check.

andrewhop (approved):

skmcgrail (approved):

lilpoozie2005 (approved):

PR #1749, opened 2024-08-07
EVP_PKEY_get0 implementation

justsmth:
### Issues:
Addresses CryptoAlg-1716

### Description of changes: 
* Provide a working implementation of the (deprecated) function `EVP_PKEY_get0`.

### DIFF
--- a/crypto/fipsmodule/evp/evp.c
+++ b/crypto/fipsmodule/evp/evp.c
@@ -581,12 +581,11 @@ int EVP_PKEY_CTX_get_signature_md(EVP_PKEY_CTX *ctx, const EVP_MD **out_md) {
 
 void *EVP_PKEY_get0(const EVP_PKEY *pkey) {
   SET_DIT_AUTO_DISABLE;
-  // Node references, but never calls this function, so for now we return NULL.
-  // If other projects require complete support, call |EVP_PKEY_get0_RSA|, etc.,
-  // rather than reading |pkey->pkey.ptr| directly. This avoids problems if our
-  // internal representation does not match the type the caller expects from
-  // OpenSSL.
-  return NULL;
+  GUARD_PTR(pkey);
+  if (pkey->type == EVP_PKEY_NONE) {
+    return NULL;
+  }
+  return pkey->pkey.ptr;
 }
 
 void OpenSSL_add_all_algorithms(void) {}
--- a/crypto/fipsmodule/evp/evp.c
+++ b/crypto/fipsmodule/evp/evp.c
@@ -582,10 +582,15 @@ int EVP_PKEY_CTX_get_signature_md(EVP_PKEY_CTX *ctx, const EVP_MD **out_md) {
 void *EVP_PKEY_get0(const EVP_PKEY *pkey) {
   SET_DIT_AUTO_DISABLE;
   GUARD_PTR(pkey);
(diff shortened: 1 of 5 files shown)


include/openssl/evp.h
```diff
@@ -1142,17 +1142,17 @@ OPENSSL_EXPORT EVP_PKEY *EVP_PKEY_new_mac_key(int type, ENGINE *engine,
...
-// functions instead.
+// EVP_PKEY_get0 returns the consumed key. The type of value returned differs
+// depending on the type of the |EVP_PKEY|.
 //
-// Note: In OpenSSL, the returned type will be different depending on the type
-//       of |EVP_PKEY| consumed. This leads to misuage very easily and has been
-//       deprecated as a no-op to avoid so.
+// This function is provided only for compatibility with OpenSSL.
+// Prefer the use the typed |EVP_PKEY_get0_*| functions instead.
 OPENSSL_EXPORT OPENSSL_DEPRECATED void *EVP_PKEY_get0(const EVP_PKEY *pkey);
```
WillChilds-Klein, line 1152: hm, should we be using `OPENSSL_DEPRECATED` elsewhere? i'm seeing [some nominally deprecated functions](https://github.com/aws/aws-lc/blob/5fd0f085dbc4030b309e6572893f7d5ee074f935/include/openssl/pkcs7.h#L104) in pkcs7.h that don't use this.

crypto/fipsmodule/evp/evp.c
```diff
@@ -543,12 +543,11 @@ int EVP_PKEY_CTX_get_signature_md(EVP_PKEY_CTX *ctx, const EVP_MD **out_md) {
 }
 
 void *EVP_PKEY_get0(const EVP_PKEY *pkey) {
-  // Node references, but never calls this function, so for now we return NULL.
-  // If other projects require complete support, call |EVP_PKEY_get0_RSA|, etc.,
-  // rather than reading |pkey->pkey.ptr| directly. This avoids problems if our
-  // internal representation does not match the type the caller expects from
-  // OpenSSL.
-  return NULL;
+  GUARD_PTR(pkey);
```
WillChilds-Klein, line 546: not a request for a change, just an observation -- coverity shows this PR as having 100% patch coverage, but i don't see a test case for `EVP_PKEY_get0(NULL)`. i guess that implies that macro-generated code isn't included in coverage data?

WillChilds-Klein (approved):

crypto/fipsmodule/evp/evp.c
justsmth: I added a test case for `EVP_PKEY_get0(NULL)` in the latest commit.

include/openssl/evp.h
justsmth: Our use of `OPENSSL_DEPRECATED` has been discussed a few times. Generally, unless there's a security concern, we're hesitant to add this designation to existing functions as it could potentially break consumers' builds.

In this case, the function was already marked `OPENSSL_DEPRECATED` so I just left it in place.

samuel40791765, line 1152: Coverage and all looks good, should we document what type is returned for each kind of `EVP_PKEY`?
Although we don't consumers using this API, it still seems better to know what's actually being returned.

samuel40791765 (approved):

WillChilds-Klein (approved):

PR #1751, opened 2024-08-07
Add EVP_PKEY_asn1_* functions

justsmth:
### Issues:
Addresses CryptoAlg-1716

### Description of changes: 
* Implementations for `EVP_PKEY_asn1_get0_info`, `EVP_PKEY_asn1_find_str`, `EVP_PKEY_asn1_find`, `EVP_PKEY_asn1_get0` and `EVP_PKEY_asn1_get_count`

### Testing:
Added two new tests.


crypto/evp_extra/p_hmac_asn1.c
```diff
@@ -121,6 +121,11 @@ const EVP_PKEY_ASN1_METHOD hmac_asn1_meth = {
     EVP_PKEY_HMAC,
     {0xff} /* placeholder oid */,
     0 /* oid_len */,
+
+    EVP_PKEY_HMAC,
+    "HMAC",
+    "AWS-LC HMAC method",
```
samuel40791765, line 127: OpenSSL seemed to remove the `hmac_asn1_meth` in favor of providers from 1.1.1 -> 3.0, so this is non-existent today.
* https://github.com/openssl/openssl/commit/2cf765e5a20762df1442c80cd2afc99e8bb1b823

1.1.1 defines this as "OpenSSL HMAC method" however, should we change this to align? 
* https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/hmac/hm_ameth.c#L95-L101

crypto/fipsmodule/evp/internal.h
```diff
@@ -77,14 +78,17 @@ extern "C" {
...
 
-typedef struct evp_pkey_asn1_method_st EVP_PKEY_ASN1_METHOD;
 typedef struct evp_pkey_method_st EVP_PKEY_METHOD;
 
 struct evp_pkey_asn1_method_st {
   int pkey_id;
   uint8_t oid[11];
   uint8_t oid_len;
 
+  int pkey_base_id;
```
samuel40791765, line 88: It seems like all values of `pkey_base_id` are identical to `pkey_id`. Can we omit this field and continue using `pkey_id` instead?

For a bit more context, `pkey_base_id` was part of pointing the historically different NID references to the same key types (`EVP_PKEY_RSA`/`EVP_PKEY_RSA2` and `EVP_PKEY_DSA1-4`). We don't have support for all the historical cruft, so I think we can get away with just the singular field.
* https://github.com/openssl/openssl/blob/8781087a69934cf45e201ac425d593b0e12a1538/crypto/rsa/rsa_ameth.c#L1009-L1010
* https://github.com/openssl/openssl/blob/8781087a69934cf45e201ac425d593b0e12a1538/crypto/dsa/dsa_ameth.c#L521-L541

include/openssl/evp.h
```diff
@@ -928,6 +928,44 @@ OPENSSL_EXPORT EVP_PKEY *EVP_PKEY_kem_new_raw_key(int nid,
...
+// ASN1 functions
+
+typedef struct evp_pkey_asn1_method_st EVP_PKEY_ASN1_METHOD;
+
+// EVP_PKEY_asn1_get_count returns the number of available
+// |EVP_PKEY_ASN1_METHOD| structures.
+OPENSSL_EXPORT int EVP_PKEY_asn1_get_count(void);
+
+// EVP_PKEY_asn1_get0
+// Returns a pointer to an EVP_PKEY_ASN1_METHOD structure. |idx| is the index
```
samuel40791765, line 940:
```suggestion
// EVP_PKEY_asn1_get0 returns a pointer to an EVP_PKEY_ASN1_METHOD structure. |idx| is the index
```

include/openssl/evp.h
```diff
@@ -928,6 +928,44 @@ OPENSSL_EXPORT EVP_PKEY *EVP_PKEY_kem_new_raw_key(int nid,
...
+// |pe| is ignored.
+// |name| is the name of the key type to find, e.g, "RSA" or "EC".
+// |len| is the length of the name.
+OPENSSL_EXPORT const EVP_PKEY_ASN1_METHOD *EVP_PKEY_asn1_find_str(
+    ENGINE **_pe, const char *name, int len);
+
+// EVP_PKEY_asn1_get0_info retrieves information about an |EVP_PKEY_ASN1_METHOD|
+// structure. |ppkey_id| is a pointer to get the key type identifier.
+// |pkey_base_id| is a pointer to get the base key type.
+// |ppkey_flags| is not supported.  Value is set to 0 is pointer is not |NULL|.
```
samuel40791765, line 960:
```suggestion
// |ppkey_flags| is not supported.  Value is set to 0 if pointer is not |NULL|.
```

samuel40791765 (dismissed):

include/openssl/evp.h
```diff
@@ -928,6 +928,45 @@ OPENSSL_EXPORT EVP_PKEY *EVP_PKEY_kem_new_raw_key(int nid,
 // to the secret key in |key|.
 OPENSSL_EXPORT int EVP_PKEY_kem_check_key(EVP_PKEY *key);
 
+// ASN1 functions
+
+typedef struct evp_pkey_asn1_method_st EVP_PKEY_ASN1_METHOD;
```
WillChilds-Klein, line 933: nit: should we move this to base.h with [the other `evp_pkey_*_st` typedefs](https://github.com/aws/aws-lc/blob/c28d0cb27759b09eb305a98f822c0223eeb2c727/include/openssl/base.h#L345)?

crypto/evp_extra/evp_asn1.c
```diff
@@ -575,3 +575,70 @@ int i2d_EC_PUBKEY(const EC_KEY *ec_key, uint8_t **outp) {
...
+  if (len < 0) {
+    return NULL;
+  }
+  // OPENSSL_strnlen returns an i, where str[i] == 0
+  const size_t name_len = OPENSSL_strnlen(name, len);
+
+  for (size_t i = 0; i < (size_t)EVP_PKEY_asn1_get_count(); i++) {
+    const EVP_PKEY_ASN1_METHOD *ameth = EVP_PKEY_asn1_get0(i);
+
+    const size_t pem_str_len = OPENSSL_strnlen(ameth->pem_str, 16);
```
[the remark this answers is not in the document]
justsmth: When you look through all of the `EVP_PKEY_ASN1_METHOD` that are defined by our library, the longest one is "DILITHIUM3" which has length 10, so a better value to use here would be 11.  I'll set a const for this value and add a comment.

WillChilds-Klein (approved):

samuel40791765 (approved):

PR #1752, opened 2024-08-07
Updating erroneous documentation for BIO_get_mem_data and subsequent usage

smittals2:
### Description of changes: 
Coverity scan flagged the usage of BIO_get_mem_data. The documentation for this function was incorrect. The documentation stated it would return the length of the data or 0 for failure, but the return value could in fact be negative (this function is a macro to BIO_ctrl which may return -2).

BIO_get_mem_data is subsequently used in ocsp_http.c without a check for the case of -2 which may lead to unexpected behavior. This PR updates the documentation for BIO_get_mem_data and replaces BIO_get_mem_data usage in ocsp_http.c with BIO_mem_contents which has less edge cases. 


crypto/bio/bio_mem.c
```diff
@@ -287,7 +287,9 @@ int BIO_mem_contents(const BIO *bio, const uint8_t **out_contents,
   }
 
   b = (BUF_MEM *)bio->ptr;
-  *out_contents = (uint8_t *)b->data;
+  if (out_contents != NULL) {
+    *out_contents = (uint8_t *)b->data;
+  }
   *out_len = b->length;
```
andrewhop, line 293: `out_len` needs the same check before using the pointer.

crypto/ocsp/ocsp_http.c
```diff
@@ -176,12 +177,20 @@ int OCSP_REQ_CTX_nbio(OCSP_REQ_CTX *rctx) {
 
       OPENSSL_FALLTHROUGH;
     case OHS_ASN1_WRITE_INIT:
-      rctx->asn1_len = BIO_get_mem_data(rctx->mem, NULL);
+      if(!BIO_mem_contents(rctx->mem, NULL, &tmp_data_len)) {
+        rctx->state = OHS_ERROR;
+        return 0;
+      }
+      rctx->asn1_len = (unsigned long)tmp_data_len;
```
andrewhop, line 184: size_t's size is platform dependent, but will be the same size or smaller than unsigned long so this is safe.

crypto/ocsp/ocsp_http.c
```diff
@@ -176,12 +177,20 @@ int OCSP_REQ_CTX_nbio(OCSP_REQ_CTX *rctx) {
...
       rctx->state = OHS_ASN1_WRITE;
 
       OPENSSL_FALLTHROUGH;
     case OHS_ASN1_WRITE:
-      data_len = BIO_get_mem_data(rctx->mem, &data);
+      if(!BIO_mem_contents(rctx->mem, &data, &tmp_data_len)) {
+        rctx->state = OHS_ERROR;
+        return 0;
+      }
+      data_len = (int)tmp_data_len;
```
andrewhop, line 193: For all of these int casts in the incredibly unlikely event that the bio holds INT_MAX + 1 bytes this will underflow and could do something weird.

crypto/ocsp/ocsp_http.c
```diff
@@ -308,7 +324,7 @@ int OCSP_REQ_CTX_nbio(OCSP_REQ_CTX *rctx) {
         }
         data++;
         rctx->asn1_len = 0;
-        for (int i = 0; i < data_len; i++) {
+        for (int i = 0; i < (int)data_len; i++) {
```
samuel40791765, line 327:
```suggestion
        for (size_t i = 0; i < data_len; i++) {
```

samuel40791765 (approved):

samuel40791765 (approved):

andrewhop (approved):

PR #1753, opened 2024-08-08
Add Alpine-Linux-x86 to GitHub Actions CI

kexgaber:
### Issues:
CryptoAlg-621
CryptoAlg-2262

### Description of changes: 

This adds a new job to the General CI GitHub Actions workflow to build and test AWS-LC on Alpine Linux (x86_64 architecture), using both Clang and GCC compilers. This ensures AWS-LC remains compatible with musl-based systems, such as Alpine. 

### Call-outs:
-  When ARM-powered runners are made generally available, this job can be expanded to also target the arm64 architecture.

### Testing:
This new job runs the existing tests found in `tests/ci/run_posix_tests.sh` and `tests/ci/run_fips_tests.sh`


skmcgrail (approved):

.github/docker_images/alpine-linux/entry.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andrewhop, line 3: Instead of a boolean can this take the path to the script to run, so in the future if we wanted to add other tests we could without needing to add more parameters to this script, e.g. Alpine + Address sanitizer.

.github/workflows/actions-ci.yml
```diff
@@ -462,6 +462,30 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build Docker Image
+        working-directory: .github/docker_images/alpine-linux
+        run: |
+          docker build -t alpine_linux ${{ matrix.compiler }} .
+      - name: Run tests
+        run: |
+          docker run -v "${{ github.workspace }}:/awslc" \
+          alpine_linux ${{ matrix.fips }}
```
andrewhop, line 487: Actually instead of an entry script can we do it all here with the Docker run command?

kexgaber: Good idea, implementing this now :)

chipitsine:
I'd suggest to use more native way like this one

https://github.com/haproxy/haproxy/blob/master/.github/workflows/musl.yml#L13-L14

justsmth (approved):

skmcgrail (approved):

PR #1755, opened 2024-08-09
add basic support for dgst hmac in tool

samuel40791765:
### Issues:
Resolves `CryptoAlg-2602`

### Description of changes: 
OpenSSL's hmac command line tool uses `dgst` with a `-hmac` option. Supporting the entire functionality of `dgst` will take a bit more work, but we can support `-hmac` with the default `SHA256` for the time being.

### Call-outs:
Rearranged some of the header file function calls to make things cleaner

### Testing:
Basic functionality tests against single and multiple /tmp files.

### DIFF
--- a/tests/ci/run_openssl_comparison_tests.sh
+++ b/tests/ci/run_openssl_comparison_tests.sh
@@ -39,10 +39,7 @@ declare -A openssl_branches=(
 export AWSLC_TOOL_PATH="${BUILD_ROOT}/tool-openssl/openssl"
 for branch in "${!openssl_branches[@]}"; do
   export OPENSSL_TOOL_PATH="${install_dir}/openssl-${branch}/bin/openssl"
-  LD_LIBRARY_PATH="${install_dir}/openssl-${branch}/${openssl_branches[$branch]}"
-  for test in X509ComparisonTest RSAComparisonTest MD5ComparisonTest; do
-      echo "Running ${test} against OpenSSL ${branch}"
-      LD_LIBRARY_PATH="${install_dir}/openssl-${branch}/${openssl_branches[$branch]}" "${BUILD_ROOT}/tool-openssl/tool_openssl_test" --gtest_filter=${test}.*
-  done
+  echo "Running ${test} against OpenSSL ${branch}"
+  LD_LIBRARY_PATH="${install_dir}/openssl-${branch}/${openssl_branches[$branch]}" "${BUILD_ROOT}/tool-openssl/tool_openssl_test"
 done
 
--- a/tool-openssl/dgst.cc
+++ b/tool-openssl/dgst.cc
@@ -1,0 +1,135 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include <fcntl.h>
+#include <iostream>
+
+#include <openssl/evp.h>
+#include <openssl/hmac.h>
+#include "internal.h"
+
+// MD5 command currently only supports stdin
+static const argument_t kArguments[] = {
+    {"-help", kBooleanArgument, "Display option summary"},
+    {"-hmac", kOptionalArgument,
+     "Create a hashed MAC with the corresponding key"},
+    {"", kOptionalArgument, ""}};
+
+static bool dgst_file_op(const std::string &filename, const EVP_MD *digest,
+                         const std::string &hmac_key) {
+  ScopedFD scoped_fd = OpenFD(filename.c_str(), O_RDONLY | O_BINARY);
+  int fd = scoped_fd.get();
+
+  static const size_t kBufSize = 8192;
+  std::unique_ptr<uint8_t[]> buf(new uint8_t[kBufSize]);
+
+  if (!hmac_key.empty()) {
+    bssl::ScopedHMAC_CTX ctx;
+    if (!HMAC_Init_ex(ctx.get(), hmac_key.data(), hmac_key.size(), digest,
+                      nullptr)) {
+      fprintf(stderr, "Failed to initialize HMAC_Init_ex.\n");
+      return false;
+    }
+
+    // Update |buf| from file continuously.
+    for (;;) {
+      size_t n;
+      if (!ReadFromFD(fd, &n, buf.get(), kBufSize)) {
+        fprintf(stderr, "%s: No such file or directory\n", filename.c_str());
+        return false;
+      }
+
+      if (n == 0) {
+        break;
+      }
+
+      if (!HMAC_Update(ctx.get(), buf.get(), n)) {
+        fprintf(stderr, "Failed to update HMAC.\n");
+        return false;
+      }
+    }
+
+    const unsigned expected_mac_len = EVP_MD_size(digest);
+    std::unique_ptr<uint8_t[]> mac(new uint8_t[expected_mac_len]);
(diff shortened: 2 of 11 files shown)


tool-openssl/dgst_test.cc
```diff
@@ -0,0 +1,99 @@
...
+                    input_file2 + +" > " + out_path_openssl;
+
+  RunCommandsAndCompareOutput(awslc_command, openssl_command, out_path_awslc,
+                              out_path_openssl, awslc_output_str,
+                              openssl_output_str);
+
+  awslc_hash = GetHash(awslc_output_str);
+  openssl_hash = GetHash(openssl_output_str);
+  awslc_hash = trim(awslc_hash);
+  openssl_hash = trim(openssl_hash);
```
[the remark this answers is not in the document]
andrewhop: Also how does this work if there are two files getting hashed? Is this only comparing the output of one of them?

tool-openssl/dgst.cc
```diff
@@ -0,0 +1,135 @@
...
+
+    if (!arg.empty() && arg[0] == '-') {
+      const std::string option = arg.substr(1);
+      if (option == "help") {
+        PrintUsage(kArguments);
+        return false;
+      } else if (option == "hmac") {
+        // Read next argument as key string.
+        it++;
+        hmac_key_string = *it;
```
justsmth, line 103: Verify that there actually is a next argument: `it != args.end()`.

tool-openssl/dgst.cc
```diff
@@ -0,0 +1,135 @@
...
+  std::string hmac_key_string;
+
+  auto it = args.begin();
+  while (it != args.end()) {
+    const std::string &arg = *it;
+    if (!arg.empty() && arg[0] != '-') {
+      // Any input without a '-' prefix is parsed as a file. This
+      // also marks the end of any option input.
+      while (it != args.end()) {
+        file_inputs.push_back(*it);
```
justsmth, line 89: Verify that `*it` is not empty: `!(*it).empty()`

tool-openssl/dgst.cc
```diff
@@ -0,0 +1,135 @@
...
+                      nullptr)) {
+      fprintf(stderr, "Failed to initialize HMAC_Init_ex.\n");
+      return false;
+    }
+
+    // Update |buf| from file continuously.
+    for (;;) {
+      size_t n;
+      if (!ReadFromFD(fd, &n, buf.get(), kBufSize)) {
+        fprintf(stderr, "%s: No such file or directory\n", filename.c_str());
```
justsmth, line 38: NP: There was an error reading from the file, which doesn't necessarily mean that there's "no such file". (Could you even read from a directory?)

tool-openssl/dgst.cc
```diff
@@ -0,0 +1,135 @@
...
+    const unsigned expected_mac_len = EVP_MD_size(digest);
+    std::unique_ptr<uint8_t[]> mac(new uint8_t[expected_mac_len]);
+    unsigned mac_len;
+    if (!HMAC_Final(ctx.get(), mac.get(), &mac_len)) {
+      fprintf(stderr, "Failed to finalize HMAC.\n");
+      return false;
+    }
+
+    // Print HMAC output.
+    fprintf(stdout, "HMAC-%s(%s)= ", EVP_MD_get0_name(digest),
```
justsmth, line 61: NP: This would display the full path.  Is that expected?

tool-openssl/dgst.cc
```diff
@@ -0,0 +1,135 @@
...
+        return false;
+      } else if (option == "hmac") {
+        // Read next argument as key string.
+        it++;
+        hmac_key_string = *it;
+      } else {
+        fprintf(stderr, "Unknown option '%s'.\n", option.c_str());
+        return false;
+      }
+    } else {
```
justsmth, line 108: I think this else block is only hit when `arg.empty()` -- I'm not sure the error message below this applies.

tool-openssl/dgst.cc
```diff
@@ -0,0 +1,135 @@
...
+      fprintf(stderr, "Unknown option '%s'.\n", arg.c_str());
+      return false;
+    }
+
+    // Increment while loop.
+    it++;
+  }
```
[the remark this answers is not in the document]
justsmth: It looks like `ParseKeyValueArguments` doesn't support an arbitrary list of file arguments as is needed by this.(?)

tool-openssl/dgst.cc
samuel40791765: Oh this was to align with OpenSSL's error messaging, I don't really enjoy it either.. I'm conflicted on compatibility everywhere or better error messages in general. 
I can change this to my original error message which also prints the `errno`. It's not really a 1 way door and we can change to align if we get complaints.

tool-openssl/dgst.cc
samuel40791765: Yes that's expected. OpenSSL's output example:
```
aws-lc on  main [?] via △ v3.20.0 via  v1.21.3 
❯ openssl dgst $PWD/../bio_test
SHA256(/Users/sachiang/Documents/repo/aws-lc/../bio_test)= 28176a1f8de724adeb381febacf90595715b55a15294c14c24a3e5f96ced04dc
```

tool-openssl/dgst.cc
samuel40791765: I was under the impression that that was already checked after the pointer's assigned to `arg`?

tool-openssl/dgst.cc
samuel40791765: The intention was to hit this if `arg[0] != '-'`. But true, that it doesn't really apply if `arg.empty()`. Will reconfigure this.

tool-openssl/dgst.cc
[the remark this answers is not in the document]
samuel40791765: Yes, `ParseKeyValueArguments` expects a map alignment of arguments and will error if there isn't a 1:1 mapping. This conflicts with how the `dgst` tool works since it can take arbitrary list of arguments at the end.
`DigestSum` from the `bssl` tool also has to deal with this and implements a similar pattern:
https://github.com/aws/aws-lc/blob/main/tool/digest.cc#L331-L334

tool-openssl/dgst_test.cc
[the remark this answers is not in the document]
samuel40791765: This was due to different OpenSSL versions having different outputs, but now that you mention it I don't think it applies here. Will change.
Also a reminder that I forgot to add these tests to the test script.

justsmth (approved):

tests/ci/run_openssl_comparison_tests.sh
```diff
@@ -39,10 +39,7 @@ declare -A openssl_branches=(
 export AWSLC_TOOL_PATH="${BUILD_ROOT}/tool-openssl/openssl"
 for branch in "${!openssl_branches[@]}"; do
   export OPENSSL_TOOL_PATH="${install_dir}/openssl-${branch}/bin/openssl"
-  LD_LIBRARY_PATH="${install_dir}/openssl-${branch}/${openssl_branches[$branch]}"
-  for test in X509ComparisonTest RSAComparisonTest MD5ComparisonTest; do
```
andrewhop, line 43: Nice, this is much simpler to add more tests in the future automatically.

andrewhop (approved):

PR #1756, opened 2024-08-09
Fix for BIO_gets and update documentation

smittals2:
### Description of changes: 
BIO_gets takes a len/size parameter which is of type int and can be negative. Previously, this param was passed into callback_ex without a check for len being positive. This is an issue since callback_ex defines this size/len param as a size_t var. Passing in a negative number to size_t could lead to unexpected behavior. 
A similar issue exists in BIO_read and BIO_write

This PR moves the check for len to before calling the callback and updates documentation for BIO_gets, BIO_read, and BIO_write.

### DIFF
--- a/crypto/bio/bio.c
+++ b/crypto/bio/bio.c
@@ -212,6 +212,10 @@ int BIO_gets(BIO *bio, char *buf, int len) {
     OPENSSL_PUT_ERROR(BIO, BIO_R_UNSUPPORTED_METHOD);
     return -2;
   }
+  if (len <= 0) {
+    return 0;
+  }
+
   int ret = 0;
   if (HAS_CALLBACK(bio)) {
     ret = (int)bio->callback_ex(bio, BIO_CB_GETS, buf, len, 0, 0L, 1L, NULL);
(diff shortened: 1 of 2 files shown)


skmcgrail (dismissed):

andrewhop (dismissed):

darylmartin100 (approved):

crypto/bio/bio.c
```diff
@@ -212,6 +212,10 @@ int BIO_gets(BIO *bio, char *buf, int len) {
     OPENSSL_PUT_ERROR(BIO, BIO_R_UNSUPPORTED_METHOD);
     return -2;
   }
+  if (len <= 0) {
+    return 0;
+  }
```
kexgaber, line 217: We might also want to consider applying this change to some of the other BIO functions where `len` is used in the callback (i.e. BIO_read and BIO_write)

smittals2: Thanks for the heads up, made the changes for those as well :)

andrewhop (approved):

skmcgrail (approved):

PR #1757, opened 2024-08-09
Basic code flow for new randomness generation

torben-hansen:
### Description of changes: 

Implements the basic randomness generation code path. Dependencies are now:
```
             +-----+                   
      +----->|ube.c|<--------+         
      |      +-----+         |         
      |                      |         
      |                      |         
+-----+----+         +-----------------+
|new_rand.c|+------->|entropy_sources.c|
+----------+         +-----------------+
```
randomness generation is implemented in `new_rand.c`. It now depends on `entropy_sources.c` for entropy needs. Eventually, these will depend on `ube.c` that will provide services to protect volatile memory. `ube.c` is not implemented yet, but e.g. `rand_ensure_ctr_drbg_uniquness()` will depend on it.

### Call-outs:

The entropy source is just a fake one right now. Later we will replace it with a re-implementation of the existing sources and with new sources.

The new randomness generation API is prefixed with `NR_PREFIX`. But to minimise the number of changes required later. Some new files use the existing API `RAND_bytes()`, etc. For example `new_rand_test.cc`. `new_rand_prefix.h` will do the necessary prefixing.

The prefix builds were too complicated to get working with this inline prefixing of the new rand API. So, just compile out compilation units if prefix builds are used.

### Testing:

The testing is also just a very basic functionality test. Later we will implement test parity with the existing implementation.


crypto/fipsmodule/rand/new_rand.c
```diff
@@ -0,0 +1,290 @@
...
+
+    if (!CTR_DRBG_generate(&state->drbg, out, todo, pred_resistance,
+          first_pred_resistance_len)) {
+      abort();
+    }
+
+    out += todo;
+    out_len -= todo;
+    state->generate_calls_since_seed++;
+    first_pred_resistance_len = 0;
```
andrewhop, line 222: Ohhhh, I think I see what your comment is talking about: if we need to call the ctr drbg multiple times for a single call to RAND_bytes we only use prediction resistance on the first call.

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -0,0 +1,290 @@
...
+      pred_resistance[i] ^= user_pred_resistance[i];
+    }
+    first_pred_resistance_len = RAND_PRED_RESISTANCE_LEN;
+  }
+
+  assert(first_pred_resistance_len == 0 ||
+         first_pred_resistance_len == RAND_PRED_RESISTANCE_LEN);
+
+  // Iterate CTR-DRBG generate until we generated |out_len| bytes of randomness.
+  while (out_len > 0) {
```
andrewhop, line 195: Can this be simplified? From my reading of CTR_DRBG_generate it either generates the requested amount of bytes or returns 0 for error, and then you abort down on L216.

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -0,0 +1,290 @@
...
+// Retire and replace call sites with RAND_bytes_with_user_prediction_resistance
+int NR_PREFIX(RAND_bytes_with_additional_data)(uint8_t *out, size_t out_len,
+  const uint8_t user_pred_resistance[RAND_PRED_RESISTANCE_LEN]) {
+  
+  RAND_bytes_private(out, out_len, user_pred_resistance,
+    RAND_USE_USER_PRED_RESISTANCE);
+  return 1;
+}
+
+int RAND_bytes_with_user_prediction_resistance(uint8_t *out, size_t out_len,
```
andrewhop, line 268: Isn't this the same as RAND_bytes_with_additional_data?

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -0,0 +1,290 @@
...
+  const uint8_t user_pred_resistance[RAND_PRED_RESISTANCE_LEN]) {
+  
+  RAND_bytes_private(out, out_len, user_pred_resistance,
+    RAND_USE_USER_PRED_RESISTANCE);
+  return 1;
+}
+
+int NR_PREFIX(RAND_bytes)(uint8_t *out, size_t out_len) {
```
andrewhop, line 276: Shouldn't this use a different "public" random bytes? If everything uses the "private" one it's not exactly private.

crypto/fipsmodule/rand/new_rand_prefix.h
```diff
@@ -0,0 +1,21 @@
...
+
+#ifndef OPENSSL_HEADER_CRYPTO_RAND_LOCAL_PREFIX_INTERNAL_H
+#define OPENSSL_HEADER_CRYPTO_RAND_LOCAL_PREFIX_INTERNAL_H
+
+#include <openssl/base.h>
+
+#define NR_CONCAT(x,y) x##_##y
+#define NR_PREFIX(x) NR_CONCAT(new_rand,x)
+
+#endif // OPENSSL_HEADER_CRYPTO_RAND_LOCAL_PREFIX_INTERNAL_H
```
andrewhop, line 12: Doesn't this need to be at the end or you could end up with multiple definitions of the below aliases?

crypto/fipsmodule/rand/new_rand_prefix.h
```diff
@@ -0,0 +1,21 @@
...
+
+#if defined(COMPILATION_UNIT_NR_PREFIX)
+
+#define RAND_bytes_with_additional_data NR_PREFIX(RAND_bytes_with_additional_data)
```
andrewhop, line 16: Does this actually cause everything in aws-lc to flip over to the new random code? This could be potentially confusing to debug what version is being used based on when this header file gets included some code could use the real/old version, and some code could use the new version.

crypto/fipsmodule/rand/new_rand_test.cc
```diff
@@ -0,0 +1,43 @@
...
+
+#include "new_rand_internal.h"
+
+
+// TODO
+// Remove when promoting to default
+#if !defined(BORINGSSL_PREFIX)
+
+#define COMPILATION_UNIT_NR_PREFIX
+#include "new_rand_prefix.h"
```
andrewhop, line 16: Why not just have this file call the new random functions directly?

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -0,0 +1,290 @@
...
+// |*pred_resistance_len| is set to 0 if no prediction resistance source is
+// available and |RAND_PRED_RESISTANCE_LEN| otherwise.
+static void rand_maybe_get_ctr_drbg_pred_resistance(
+  struct rand_thread_local_state *state,
+  uint8_t pred_resistance[RAND_PRED_RESISTANCE_LEN],
+  size_t *pred_resistance_len) {
+
+  *pred_resistance_len = 0;
+
+  if (state->entropy_source.prediction_resistance != NULL) {
```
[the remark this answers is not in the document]
torben-hansen: I was having the assumption that `state` couldn't be `NULL`. It is ensured because every code path must path through state allocation code.

But now I modified code to be more paranoid at different places.

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -0,0 +1,290 @@
...
+    state->entropy_source.prediction_resistance(pred_resistance);
+    *pred_resistance_len = RAND_PRED_RESISTANCE_LEN;
+  }
+}
+
+// rand_get_ctr_drbg_seed_entropy source entropy for seeding and reseeding the
+// CTR-DRBG state. Firstly, |seed| is filled with |CTR_DRBG_ENTROPY_LEN| bytes
+// from the seed source from the |state| entropy source. Secondly, if
+// available, |CTR_DRBG_ENTROPY_LEN| bytes is filled into
```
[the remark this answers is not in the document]
torben-hansen: Because the source for the personalization string != the source for the prediction resistance.

And to see, we need the seed source and the personalization string source. I might make the personalization string source required at some point. But dunno yet.

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -0,0 +1,290 @@
...
+  const uint8_t user_pred_resistance[RAND_PRED_RESISTANCE_LEN],
+  int use_user_pred_resistance) {
+
+  // Ensure CTR-DRBG state is unique.
+  if (rand_ensure_ctr_drbg_uniquness(state, out_len) != 1) {
+    rand_ctr_drbg_reseed(state);
+  }
+
+  // If a prediction resistance source is available, use it.
+  // Prediction resistance is only used on first invocation of the CTR-DRBG,
```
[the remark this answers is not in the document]
torben-hansen: Not necessary IMO.

Firstly, don't want to use the same unpredictability twice. Secondly, adding in new unpredictability for forward security in the middle of generating the output doesn't really add any value IMO.

Secondly, the output of multiple CTR-DRBG generate executions will be concatenated and returned to the caller. By injecting some unpredictability before the first execution we ensure forward security: that is, given the secret state at some time `T`, you can't recover output from the CTR-DRBG for time `> T` i.e. in the future.

crypto/fipsmodule/rand/new_rand.c
torben-hansen: I don't see how because the generate function can output at most `CTR_DRBG_MAX_GENERATE_LENGTH` per invocation. So, we must loop over invocations to satisfy `out_len` that could be higher.

We can simplify if we restrict `RAND_bytes()` and friends to only allow `out_len <= CTR_DRBG_MAX_GENERATE_LENGTH`.

crypto/fipsmodule/rand/new_rand.c
torben-hansen: Yes, correct. But I want to correct the name.

The `additional_data` was not really used for "additional data" in the sense of SP800-90. It was used as prediction resistance.

crypto/fipsmodule/rand/new_rand.c
torben-hansen: I'm still pondering. Later I want to add another, independent state, such that we can have a real public/private API. But which exported function should have the "private" state and which should have the "public" state? You can view it two ways:

1. Every output of `RAND_bytes` is assumed to come from "private". You have to actively select you want some public output.
2. Every output of `RAND_bytes` is assumed to be "public". 

I am probably going for (2). I will have to rename some stuff later. But that is just renaming some internal functions.

Firstly, let's say we pick (2). Then if an application want to catagorise into public/private, then they need to change all function calls that generate private output. You can miss some call sites. Similarly, if we pick (1), you can still miss call sites when migrating; it's just the reverse where you forgot some call sites that should have been made public. It's sorta equivalent.

Secondly, everyone are already using `RAND_bytes()` in their code. We are not going to change this API, ever. Hence, to avoid having to add yet another function "RAND_bytes_public"-something, we can make use of `RAND_priv_bytes`.

In fact, it's better not to think about this as public/private. A better way to think about it, conceptually, is that you have two independent states from which you can generate randomness. Applications can then do the catagorisation they want. But since public/private is already used. We are stuck with that naming.

crypto/fipsmodule/rand/new_rand_prefix.h
torben-hansen: No, see your second comment for this file.

crypto/fipsmodule/rand/new_rand_prefix.h
torben-hansen: I can see why this file is a bit confusing. It's needed such that I can use the existing function names in e.g. testing code.

The `OPENSSL_HEADER_CRYPTO_RAND_LOCAL_PREFIX_INTERNAL_H` part ensures I can use the prefix macros where I need themm.

Now, note that you have to actively define `COMPILATION_UNIT_NR_PREFIX` for something to happen. I will use this part in compilation units where I need to write e.g. `RAND_bytes` many times. For example in tests. When we flip over the default to the new randomness generation I won't have to go around and change names at 1000 places. I just delete this prefix-file from the compilation unit. It is also a way where I can flip over existing tests to use my new implementation during development.

I don't use the part `COMPILATION_UNIT_NR_PREFIX` everywhere because of how our FIPS module build works. To ensure that my stuff works for our fips build I need to add e.g. `new_rand.c` to `bcm.c`. But after pre-processing that is just one giant file. Therefore, I can't even use the same struct type names. Even static functions with the same name would result in redefinition errors.

The prefix build also make things complicated. Because the prefix macros are included everywhere.

crypto/fipsmodule/rand/new_rand_test.cc
torben-hansen: See comment for `new_rand_prefix.h`.

crypto/fipsmodule/rand/entropy/internal.h
```diff
@@ -0,0 +1,32 @@
...
+#endif
+
+// I could make these array types!
+struct entropy_source {
+  int is_initialized;
+  int (*initialize)(void);
+  int (*cleanup)(void);
+  int (*seed)(uint8_t seed[CTR_DRBG_ENTROPY_LEN]);
+  int (*personalization_string)(uint8_t personalization_string[CTR_DRBG_ENTROPY_LEN]);
+  int (*prediction_resistance)(uint8_t pred_resistance[RAND_PRED_RESISTANCE_LEN]);
```
skmcgrail, line 22: Might be easier to grawk this if these are prefixed to indicate that these are getters (i.e. `get_*`?

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -0,0 +1,290 @@
...
+// rand_get_ctr_drbg_seed_entropy source entropy for seeding and reseeding the
+// CTR-DRBG state. Firstly, |seed| is filled with |CTR_DRBG_ENTROPY_LEN| bytes
+// from the seed source from the |state| entropy source. Secondly, if
+// available, |CTR_DRBG_ENTROPY_LEN| bytes is filled into
+// |personalization_string| sourced from the personalization string source from
+// the entropy source in |state|.
+//
+// |*personalization_string_len| is set to 0 if no personalization string source
+// is available and |CTR_DRBG_ENTROPY_LEN| otherwise.
+static void rand_get_ctr_drbg_seed_entropy(struct entropy_source *entropy_source,
```
skmcgrail, line 84:
```suggestion
static void rand_get_ctr_drbg_seed_entropy(struct entropy_source *state,
```

To match your doc string.

skmcgrail (dismissed):

crypto/fipsmodule/rand/entropy/internal.h
```diff
@@ -0,0 +1,32 @@
...
+  int (*randomize)(void);
+};
```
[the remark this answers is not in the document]
andrewhop: At first I was going to push for everything should be handled in the PR that adds it, but I see this isn't going to main so I'll allow it since I imagine some of this will be changing over time.

crypto/fipsmodule/rand/entropy/internal.h
```diff
@@ -0,0 +1,32 @@
...
+
+void get_entropy_source(struct entropy_source *entropy_source);
```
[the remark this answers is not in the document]
andrewhop: I don't see the return type of get_entropy_source changing, did you forgot to push the latest revision?

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -0,0 +1,290 @@
...
+// For UBE.
+static int rand_ensure_ctr_drbg_uniquness(struct rand_thread_local_state *state,
+  size_t out_len) {
+
+  return 1;
+}
+
+// rand_maybe_get_ctr_drbg_pred_resistance maybe fills |pred_resistance| with
+// |RAND_PRED_RESISTANCE_LEN| bytes. The bytes are sourced from the prediction
+// resistance source from the entropy source in |state|, if such a source has
```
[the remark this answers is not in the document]
andrewhop: Can you find a reference for that and add it to this documentation so we can track why we don't always have prediction resistance.

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -0,0 +1,290 @@
...
+// been configured.
+//
+// |*pred_resistance_len| is set to 0 if no prediction resistance source is
+// available and |RAND_PRED_RESISTANCE_LEN| otherwise.
+static void rand_maybe_get_ctr_drbg_pred_resistance(
+  struct rand_thread_local_state *state,
+  uint8_t pred_resistance[RAND_PRED_RESISTANCE_LEN],
+  size_t *pred_resistance_len) {
```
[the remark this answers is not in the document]
andrewhop: Right, but isn't that always the same as what I asked about initially? It's either zero or RAND_PRED_RESISTANCE_LEN, nothing else.

crypto/fipsmodule/rand/new_rand.c
andrewhop: Are you planning on making this a public API? Will that be even more confusing for future customers? They'll have to pick from public/private random sources, and prediction resistance or additional data. How should they figure out what to use?

crypto/fipsmodule/rand/new_rand.c
andrewhop: Eagerly awaiting this change, have you talked to any of our customers (such as s2n-tls) what they would expect or prefer?

crypto/fipsmodule/rand/new_rand.c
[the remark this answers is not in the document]
andrewhop: Is that assumption documented anywhere? Fun thing to think about: how can we codify and enforce that assumption? Is this a use case for something like CBMC or some other new safety tool?

crypto/fipsmodule/rand/new_rand.c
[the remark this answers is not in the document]
torben-hansen: I think this should be documented where entropy sources are configured. I haven't implemented that part yet. But I will add plentiful of reasoning there.

This part of the code is agnostic to which sources are configured. It's concerned with managing the ctr-drbg state that generates output to users.

crypto/fipsmodule/rand/new_rand.c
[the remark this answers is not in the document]
torben-hansen: I see what you mean. I am overloading `first_pred_resistance_len` which is the variable that is given as the `pred_resistance_len` argument: it both acts like a boolean (true if value is greater than 0) and carries the length.

It's to avoid annoying if-statements at the call-site of `rand_maybe_get_ctr_drbg_pred_resistance`. `CTR_DRBG_generate` takes a length argument, not a boolean, so we would have to convert a boolean into a length parameter. `CTR_DRBG_generate` is exported, so can't change the interface.

It also makes it easier to later implemented prediction resistance lengths that is decided by `rand_maybe_get_ctr_drbg_pred_resistance`.

crypto/fipsmodule/rand/new_rand.c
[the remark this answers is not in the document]
torben-hansen: I hadn't documented it. It's no longer an assumptions so leaving it out.

It could be covered by some tool that verifies that there are no UB's. Out of scope for now. But maybe later down the line.

crypto/fipsmodule/rand/new_rand.c
torben-hansen: Changed description instead.

crypto/fipsmodule/rand/new_rand.c
torben-hansen: `RAND_bytes` and `RAND_priv_bytes` exists in OpenSSL and are not deprecated. So, I think it would make most sense to stick to those. They are pretty natural to use for e.g. s2n-tls.

crypto/fipsmodule/rand/new_rand.c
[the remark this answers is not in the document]
andrewhop: Ahhhh makes sense.

crypto/fipsmodule/rand/new_rand.c
andrewhop: Makes sense to use those names, I was thinking about the choice between open 1 and 2, and what (if anything) in AWS-LC's own use of RAND_bytes should be migrated to RAND_priv_bytes.

andrewhop (approved):

crypto/fipsmodule/rand/new_rand.c
torben-hansen: > (if anything) in AWS-LC's own use of RAND_bytes should be migrated to RAND_priv_bytes.

We can definitely look at that later.

skmcgrail (approved):

PR #1762, opened 2024-08-13
Allow aarch64 CPUID capability check for all Linux platforms

skmcgrail:
### Issues:
Addresses P147929287

### Description of changes: 
Removes the `OPENSSL_ANDROID` guard around the aarch64 CPUID capability check and allows it for all aarch64 linux platforms. This is now inline with the original commit taken from main: https://github.com/aws/aws-lc/commit/24cb055cb48ae15b9b924ec939f4b993380723f7

### DIFF
--- a/crypto/fipsmodule/cpucap/cpu_aarch64_linux.c
+++ b/crypto/fipsmodule/cpucap/cpu_aarch64_linux.c
@@ -97,9 +97,7 @@ void OPENSSL_cpuid_setup(void) {
   static const unsigned long kSHA256 = 1 << 6;
   static const unsigned long kSHA512 = 1 << 21;
   static const unsigned long kSHA3 = 1 << 17;
-#if defined(OPENSSL_ANDROID)
   static const unsigned long kCPUID = 1 << 11;
-#endif
 
   uint64_t OPENSSL_arm_midr = 0;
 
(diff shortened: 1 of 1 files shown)


crypto/fipsmodule/cpucap/cpu_aarch64_linux.c
```diff
@@ -133,19 +131,14 @@ void OPENSSL_cpuid_setup(void) {
   // Before calling armv8_cpuid_probe and reading from MIDR_EL1 check that it
   // is supported. As of Valgrind 3.21 trying to read from that register will
   // cause Valgrind to crash.
-
-#if defined(OPENSSL_ANDROID)
   if (hwcap & kCPUID) {
```
justsmth, line 134: Why does `kCPUID` even need to be here? It's a `static const` value initialized above but otherwise doesn't appear to be used?

skmcgrail: I'm not following, it's used here `f (hwcap & kCPUID) {`? This is how we have it structured on main, so if you think the use of constants should be cleaned up we should take it up there first.

torben-hansen: it's not a boolean above. it's the bitwise and `&` :)

torben-hansen (approved):

justsmth (approved):

crypto/fipsmodule/cpucap/cpu_aarch64_linux.c
justsmth: 🤦‍♂️

darylmartin100 (approved):

PR #1763, opened 2024-08-13
ML-KEM refactor

dkostic:
### Issues:
CryptoAlg-2614

### Description of changes: 
Previous to this change, ML-KEM was implemented such that
the code for a parameter set was selected by defining the correct
C pre-processor flag (for example, if you want to compile the code
for ML-KEM 512 parameter set you would `#define KYBER_K 2`).
The consequence of this was that we had to compile the code
three times for three ML-KEM parameter sets. We do this by adding
a C file for each parameter set where we first define the corresponding
`KYBER_K` value and then include all the ML-KEM C files. Besides being
an awkward way to handle multiple parameter sets, this will not work
for the FIPS module where we bundle all C files inside `bcm.c` and
compile it as a single compilation unit.

In this change we refactor ML-KEM by parametrizing all functions that
depend on values that are specific to a parameter set, i.e., that directly
or indirectly depend on the value of `KYBER_K`. To do this, in `params.h`
we define a structure that holds those ML-KEM parameters and functions
that initialize a given structure with values corresponding to a parameter
set. This structure can then be passed to every function that requires it.
For example, `polyvec_ntt` function was previously implemented like this:
```
void polyvec_ntt(polyvec *r) {
  unsigned int i;
  for(i=0;i<KYBER_K;i++)
    poly_ntt(&r->vec[i]);
}
```
We now change that to:
```
void polyvec_ntt(ml_kem_params *params, polyvec *r) {
  unsigned int i;
  for(i=0;i<params->k;i++)
    poly_ntt(&r->vec[i]);
}
```
Of course, now we have to change all callers of `polyvec_ntt` to also
have `ml_kem_params` as an input argument, and then callers of the
caller, etc. These changes bubble up to the highest level API defined
in `kem.h` where we now have:
```
int crypto_kem_keypair(ml_kem_params *params, ...);
int crypto_kem_enc(ml_kem_params * params, ...);
int crypto_kem_dec(ml_kem_params * params, ...);
int crypto_kem_keypair_derand(ml_kem_params *params, ...);
int crypto_kem_enc_derand(ml_kem_params * params, ...);
```
Then we can easily implement these functions for specific parameter sets,
which can be seen in `ml_kem.c` file. For example:
```
int ml_kem_512_ipd_keypair(...) {
  ml_kem_params params;
  return ml_kem_keypair_ref(&params, ...);
}
int ml_kem_768_ipd_keypair(...) {
  ml_kem_params params;
  ml_kem_768_params_init(&params);
  return ml_kem_keypair_ref(&params, ...);
}
```

### Call-outs:

### Testing:


crypto/ml_kem/ml_kem_ipd_ref_common/params.h
```diff
@@ -2,56 +2,86 @@
...
-#endif
-
+// The only defined parameters are those that don't depend
+// on the parameter set. All other parameters are specified
+// in ml_kem_params structure that is unique for each parameter
+// set (ML-KEM 512/768/1024).
 #define KYBER_N 256
 #define KYBER_Q 3329
 
 #define KYBER_SYMBYTES 32   /* size in bytes of hashes, and seeds */
```
torben-hansen, line 14: wrong comment style

crypto/ml_kem/ml_kem_ipd_ref_common/params.h
```diff
@@ -2,56 +2,86 @@
...
 #define KYBER_SSBYTES  32   /* size in bytes of shared key */
 
 #define KYBER_POLYBYTES		384
```
torben-hansen, line 17: indentation

crypto/ml_kem/ml_kem.c
```diff
@@ -12,89 +23,119 @@
 int ml_kem_512_ipd_keypair_deterministic(uint8_t *public_key  /* OUT */,
                                          uint8_t *secret_key  /* OUT */,
                                          const uint8_t *seed  /* IN */) {
-  return ml_kem_512_ref_keypair_derand(public_key, secret_key, seed);
+  ml_kem_params params;
+  ml_kem_512_params_init(&params);
```
torben-hansen, line 27: Functions are specific to a parameter set, so don't need a function that construct the parameter on every invocation, You can simply define a static object for the 3 parameter sets and use that? No need to construct them for every invocation.

torben-hansen: Note that if you carry this into the fips module you would need to use the BSS allocation machinery.

crypto/ml_kem/ml_kem_ipd_ref_common/cbd.c
```diff
@@ -105,17 +101,15 @@ static void cbd3(poly *r, const uint8_t buf[3*KYBER_N/4])
...
-#endif
+  assert((params->eta1 == 2) || (params->eta1 == 3));
+  if (params->eta1 == 2) {
+    cbd2(r, buf);
+  } else {
+    cbd3(r, buf);
+  }
 }
 
 void poly_cbd_eta2(poly *r, const uint8_t buf[KYBER_ETA2*KYBER_N/4])
```
torben-hansen, line 115: Don't need pre-processing in this function

crypto/ml_kem/ml_kem_ipd_ref_common/params.h
```diff
@@ -2,56 +2,86 @@
...
+// for static allocation.
+#define KYBER_K_MAX (4)
+#define KYBER_ETA1_MAX (3)
+#define KYBER_POLYCOMPRESSEDBYTES_MAX    (160)
+#define KYBER_POLYVECCOMPRESSEDBYTES_MAX (4 * 352)
+
+#define KYBER_INDCPA_BYTES_MAX    (KYBER_POLYVECCOMPRESSEDBYTES_MAX + KYBER_POLYCOMPRESSEDBYTES_MAX)
+#define KYBER_CIPHERTEXTBYTES_MAX (KYBER_INDCPA_BYTES_MAX)
+
+#define KYBER_NAMESPACE(s) ml_kem_##s##_ref
```
torben-hansen, line 85: we still have to carry around this namespace thing?

I suspect it might be quite problematic for the prefix build....

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.c
```diff
@@ -37,12 +38,13 @@ static void pack_pk(uint8_t r[KYBER_INDCPA_PUBLICKEYBYTES],
 *              - uint8_t *seed: pointer to output seed to generate matrix A
 *              - const uint8_t *packedpk: pointer to input serialized public key
 **************************************************/
-static void unpack_pk(polyvec *pk,
-                      uint8_t seed[KYBER_SYMBYTES],
-                      const uint8_t packedpk[KYBER_INDCPA_PUBLICKEYBYTES])
+static void unpack_pk(ml_kem_params *params,
+                      polyvec *pk,
+                      uint8_t *seed,
```
torben-hansen, line 43: why not keep as array type?

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.c
```diff
@@ -37,12 +38,13 @@ static void pack_pk(uint8_t r[KYBER_INDCPA_PUBLICKEYBYTES],
 *              - uint8_t *seed: pointer to output seed to generate matrix A
 *              - const uint8_t *packedpk: pointer to input serialized public key
 **************************************************/
-static void unpack_pk(polyvec *pk,
-                      uint8_t seed[KYBER_SYMBYTES],
-                      const uint8_t packedpk[KYBER_INDCPA_PUBLICKEYBYTES])
+static void unpack_pk(ml_kem_params *params,
+                      polyvec *pk,
```
torben-hansen, line 42: const?

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.c
```diff
@@ -19,12 +19,13 @@
 *              polyvec *pk: pointer to the input public-key polyvec
 *              const uint8_t *seed: pointer to the input public seed
 **************************************************/
-static void pack_pk(uint8_t r[KYBER_INDCPA_PUBLICKEYBYTES],
+static void pack_pk(ml_kem_params *params,
+                    uint8_t *r,
```
torben-hansen, line 23: const?

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.c
```diff
@@ -53,9 +55,9 @@ static void unpack_pk(polyvec *pk,
 * Arguments:   - uint8_t *r: pointer to output serialized secret key
 *              - polyvec *sk: pointer to input vector of polynomials (secret key)
 **************************************************/
-static void pack_sk(uint8_t r[KYBER_INDCPA_SECRETKEYBYTES], polyvec *sk)
+static void pack_sk(ml_kem_params *params, uint8_t *r, polyvec *sk)
```
torben-hansen, line 58: `r` const?

torben-hansen: Could review rest of this file for more of these.

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.c
```diff
@@ -200,40 +202,41 @@ void gen_matrix(polyvec *a, const uint8_t seed[KYBER_SYMBYTES], int transposed)
 *              - const uint8_t *coins: pointer to input randomness
 *                             (of length KYBER_SYMBYTES bytes)
 **************************************************/
-void indcpa_keypair_derand(uint8_t pk[KYBER_INDCPA_PUBLICKEYBYTES],
-                           uint8_t sk[KYBER_INDCPA_SECRETKEYBYTES],
-                           const uint8_t coins[KYBER_SYMBYTES])
+void indcpa_keypair_derand(ml_kem_params *params,
+                           uint8_t *pk,
+                           uint8_t *sk,
+                           const uint8_t *coins)
```
torben-hansen, line 208: can't be array type?

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.c
```diff
@@ -307,18 +311,22 @@ void indcpa_enc(uint8_t c[KYBER_INDCPA_BYTES],
...
+                const uint8_t *c,
+                const uint8_t *sk)
 {
   polyvec b, skpv;
   poly v, mp;
 
-  unpack_ciphertext(&b, &v, c);
-  unpack_sk(&skpv, sk);
+  // work-around for gcc-12 which complains that skpv may be used uninitialized.
+  memset(&skpv, 0, sizeof(polyvec));
```
torben-hansen, line 323:
```suggestion
  OPENSSL_memset(&skpv, 0, sizeof(polyvec));
```

crypto/ml_kem/ml_kem_ipd_ref_common/params.h
torben-hansen: actually, for the prefix comment. not really if it's only on non-exported functions.

crypto/ml_kem/ml_kem.c
dkostic: yeah, I decided to use a function instead of static object to avoid having to do the BSS machinery when we move the code to the FIPS module.

crypto/ml_kem/ml_kem_ipd_ref_common/cbd.c
dkostic: I wanted to minimize changes in this PR and change only what is necessary so I skipped things like this. I'd really rather leave this as is and refactor/reimplement once we get more time allocated for it.

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.c
dkostic: I wanted to minimize changes in this PR and change only what is necessary so I skipped things like this. I'd really rather leave this as is and refactor/reimplement once we get more time allocated for it.

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.c
dkostic: miss. fixed it.

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.c
dkostic: fixed.

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.c
dkostic: done

crypto/ml_kem/ml_kem_ipd_ref_common/params.h
dkostic: the whole implementation is using wrong comment style :) I'd rather leave such change for some other refactoring.

crypto/ml_kem/ml_kem_ipd_ref_common/params.h
```diff
@@ -2,56 +2,86 @@
...
+  size_t indcpa_bytes;
+  size_t public_key_bytes;
+  size_t secret_key_bytes;
+  size_t ciphertext_bytes;
+} ml_kem_params;
+
+static void ml_kem_params_init(ml_kem_params *params, size_t k) {
+  assert((k == 2) || (k == 3) || (k == 4));
+
+  size_t eta1 = (k == 2) ? 3 : 2;
```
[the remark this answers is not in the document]
dkostic: I wasn't completely sure if this is always a well defined behavior:
```
params->a = X;
params->b = params->a * Y;
...
```
so to be safe I just used temp variables :)

crypto/ml_kem/ml_kem_ipd_ref_common/params.h
```diff
@@ -2,56 +2,86 @@
...
+  params->poly_vec_compressed_bytes = poly_vec_compressed_bytes;
+  params->indcpa_public_key_bytes = indcpa_public_key_bytes;
+  params->indcpa_secret_key_bytes = indcpa_secret_key_bytes;
+  params->indcpa_bytes = indcpa_bytes;
+  params->public_key_bytes = public_key_bytes;
+  params->secret_key_bytes = secret_key_bytes;
+  params->ciphertext_bytes = ciphertext_bytes;
+}
+
+OPENSSL_UNUSED static void ml_kem_512_params_init(ml_kem_params *params) {
```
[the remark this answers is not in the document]
dkostic: done.

crypto/ml_kem/ml_kem_ipd_ref_common/params.h
```diff
@@ -2,56 +2,86 @@
...
+}
+OPENSSL_UNUSED static void ml_kem_1024_params_init(ml_kem_params *params) {
+  ml_kem_params_init(params, 4);
+}
+
+// We define max values for some parameters because they are used
+// for static allocation.
+#define KYBER_K_MAX (4)
+#define KYBER_ETA1_MAX (3)
+#define KYBER_POLYCOMPRESSEDBYTES_MAX    (160)
```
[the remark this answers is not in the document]
dkostic: I wanted to minimize changes in this PR and change only what is necessary so I skipped things like this. I'd really rather leave this as is and refactor/reimplement once we get more time allocated for it.

crypto/ml_kem/ml_kem_ipd_ref_common/params.h
dkostic: unfortunately we still have to do it, to make the prefix build happy actually. I forgot now what exactly was the issue but I did try to remove the namespace macro and it failed to build the prefix build properly.

torben-hansen (approved):

crypto/ml_kem/ml_kem_ipd_ref_common/cbd.c
```diff
@@ -105,17 +103,15 @@ static void cbd3(poly *r, const uint8_t buf[3*KYBER_N/4])
...
-#elif KYBER_ETA1 == 3
-  cbd3(r, buf);
-#else
-#error "This implementation requires eta1 in {2,3}"
-#endif
+  assert((params->eta1 == 2) || (params->eta1 == 3));
+  if (params->eta1 == 2) {
+    cbd2(r, buf);
+  } else {
+    cbd3(r, buf);
```
sgmenda-aws, line 113: This `else` is risky because release builds may remove the above `assert` so this code path may be reached by code with `eta1 != 3`. Might be worth replacing with `else if (params->eta1 == 3)` and do a separate `else` that is a no-op.

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.c
```diff
@@ -19,12 +22,13 @@
 *              polyvec *pk: pointer to the input public-key polyvec
 *              const uint8_t *seed: pointer to the input public seed
 **************************************************/
-static void pack_pk(uint8_t r[KYBER_INDCPA_PUBLICKEYBYTES],
+static void pack_pk(ml_kem_params *params,
+                    uint8_t *r,
                     polyvec *pk,
-                    const uint8_t seed[KYBER_SYMBYTES])
+                    const uint8_t *seed)
```
sgmenda-aws, line 28: Isn't `seed` always `KYBER_SYMBYTES = 32` bytes long, irrespective of parameter set? And we use exactly `KYBER_SYMBYTES` bytes of `seed`, so seems worth keeping the constraint.

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.c
```diff
@@ -157,15 +162,15 @@ static unsigned int rej_uniform(int16_t *r,
 **************************************************/
 #define GEN_MATRIX_NBLOCKS ((12*KYBER_N/8*(1 << 12)/KYBER_Q + XOF_BLOCKBYTES)/XOF_BLOCKBYTES)
 // Not static for benchmarking
-void gen_matrix(polyvec *a, const uint8_t seed[KYBER_SYMBYTES], int transposed)
+void gen_matrix(ml_kem_params *params, polyvec *a, const uint8_t *seed, int transposed)
```
sgmenda-aws, line 165: Same comment as above for `seed`.

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.c
```diff
@@ -253,45 +259,46 @@ void indcpa_keypair_derand(uint8_t pk[KYBER_INDCPA_PUBLICKEYBYTES],
...
 **************************************************/
-void indcpa_enc(uint8_t c[KYBER_INDCPA_BYTES],
-                const uint8_t m[KYBER_INDCPA_MSGBYTES],
-                const uint8_t pk[KYBER_INDCPA_PUBLICKEYBYTES],
-                const uint8_t coins[KYBER_SYMBYTES])
+void indcpa_enc(ml_kem_params *params,
+                uint8_t *c,
+                const uint8_t *m,
+                const uint8_t *pk,
+                const uint8_t *coins)
```
sgmenda-aws, line 266: ` poly_getnoise_*` assumes `coins` is `KYBER_SYMBYTES`, so might be worth keeping the constraint.

crypto/ml_kem/ml_kem_ipd_ref_common/params.c
```diff
@@ -0,0 +1,41 @@
+#include <openssl/base.h>
+
+#include "params.h"
+
+static void ml_kem_params_init(ml_kem_params *params, size_t k) {
+  assert((k == 2) || (k == 3) || (k == 4));
```
sgmenda-aws, line 6: release builds may ignore this `assert`, leading to unexpected data for the below conditionals.

sgmenda-aws (approved):
lgtm with minor changes

crypto/ml_kem/ml_kem_ipd_ref_common/cbd.c
WillChilds-Klein: ^ ore move the assert to that `else` clause.

sgmenda-aws: Wouldn't the `assert` still be a no-op in release builds?

crypto/ml_kem/ml_kem_ipd_ref_common/polyvec.c
```diff
@@ -87,45 +87,46 @@ void polyvec_compress(uint8_t r[KYBER_POLYVECCOMPRESSEDBYTES], const polyvec *a)
...
-      t[3] = (a[4] >> 1) | ((uint16_t)a[ 5] << 7);
-      t[4] = (a[5] >> 4) | ((uint16_t)a[ 6] << 4);
-      t[5] = (a[6] >> 7) | ((uint16_t)a[ 7] << 1) | ((uint16_t)a[8] << 9);
-      t[6] = (a[8] >> 2) | ((uint16_t)a[ 9] << 6);
-      t[7] = (a[9] >> 5) | ((uint16_t)a[10] << 3);
-      a += 11;
-
-      for(k=0;k<8;k++)
-        r->vec[i].coeffs[8*j+k] = ((uint32_t)(t[k] & 0x7FF)*KYBER_Q + 1024) >> 11;
+  assert((params->poly_vec_compressed_bytes == params->k * 352) ||
```
WillChilds-Klein, line 94: nit: this code would be a bit more legible if we did e.g. `#define K_MLKEM_768 352` for the 3 security levels

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.h
```diff
@@ -6,22 +6,25 @@
 #include "polyvec.h"
 
 #define gen_matrix KYBER_NAMESPACE(gen_matrix)
-void gen_matrix(polyvec *a, const uint8_t seed[KYBER_SYMBYTES], int transposed);
```
WillChilds-Klein, line 9: +1 to sanketh's earlier comment -- seed size is invariant over parameters, right? why can't we define this as an array rather than a pointer?

WillChilds-Klein (approved):
code changes look good, no blockers. per sanketh's earlier comments, can we keep parameter-invariant buffers as byte arrays rather than pointers?

crypto/ml_kem/ml_kem_ipd_ref_common/indcpa.h
dkostic: yes, will fix in https://github.com/aws/aws-lc/pull/1775

PR #1766, opened 2024-08-14
refactor md5 tool with dgst and fix stdin behavior

samuel40791765:
### Issues:
Addresses `CryptoAlg-2602`

### Description of changes: 
Our original md5 stdin behavior was acting weird. The issue seems to be that we weren't actually parsing all of the input from `stdin`.
```
ubuntu in 🌐 ip-172-31-63-232 in aws-lc on  fix-md5-tool [=$+?] via △ v3.16.3 via 🐹 v1.20.4 
❯ echo "test" | openssl md5 
(stdin)= d8e8fca2dc0f896fd7cb4cb0031ba249

ubuntu in 🌐 ip-172-31-63-232 in aws-lc on  fix-md5-tool [=$+?] via △ v3.16.3 via 🐹 v1.20.4 
❯ echo "test" | ./test_build_dir/tool-openssl/openssl md5 
(stdin)= 098f6bcd4621d373cade4e832627b4f6
```
The original implementation used `getline` which only reads one line. This worked with the original test since the file contents were sent into `stdin` with `<`, but this doesn't work correctly with other `stdin` behaviors such as `echo test | openssl md5`.
The correct and more useful way to handle `stdin` is to read it from [the `0` file descriptor.](https://en.wikipedia.org/wiki/File_descriptor) This is how OpenSSL/BoringSSL reads from files.

1. This change fixes the behavior by rebuilding `md5` onto our `dgst` implementation. This is identical to how OpenSSL factors the specific hashing tool names. `md5` now uses `dgst` under the hood and inherits the file hashing behavior we implemented for `dgst hmac`.
2. `stdin` was missing from `dgst`, so that was implemented as well. `stdin` now works for both `openssl dgst -hmac ...` and `openssl md5`. Corresponding tests have also been added to verify.
3. Code was structured so that we can easily extend new digest support for the openssl tool with just a few lines. I considered adding it along with this PR, but didn't want to muddle it with the new logic.

FIxed version:
```
ubuntu in 🌐 ip-172-31-63-232 in aws-lc on  fix-md5-tool [$?] via △ v3.16.3 via 🐹 v1.20.4 
❯ echo "test_fix" | openssl md5
(stdin)= 1312e69e9dc6ec31b51af24fc42ebdea

ubuntu in 🌐 ip-172-31-63-232 in aws-lc on  fix-md5-tool [$?] via △ v3.16.3 via 🐹 v1.20.4 
❯ echo "test_fix" | ./test_build_dir/tool-openssl/openssl md5
(stdin)= 1312e69e9dc6ec31b51af24fc42ebdea

```

### Call-outs:
Deleted the md5 files since most logic was moved under `dgst`.

### Testing:
1. New tests for `hmac stdin`
2. New tests for `md5 stdin` and `md5 files`.


tool-openssl/dgst.cc
```diff
@@ -58,27 +57,79 @@ static bool dgst_file_op(const std::string &filename, const EVP_MD *digest,
...
+      fprintf(stdout, "(%s)= ", filename.c_str());
+    };
     for (size_t i = 0; i < expected_mac_len; i++) {
       fprintf(stdout, "%02x", mac[i]);
     }
     fprintf(stdout, "\n");
     return true;
   }
+  // Run Hash operations.
+  else if (digest) {
```
justsmth, line 75: NP: The if block returns unconditionally, so the `else` is not needed.

tool-openssl/dgst.cc
```diff
@@ -16,14 +16,13 @@ static const argument_t kArguments[] = {
      "Create a hashed MAC with the corresponding key"},
     {"", kOptionalArgument, ""}};
 
-static bool dgst_file_op(const std::string &filename, const EVP_MD *digest,
-                         const char *hmac_key, const size_t hmac_key_len) {
-  ScopedFD scoped_fd = OpenFD(filename.c_str(), O_RDONLY | O_BINARY);
-  int fd = scoped_fd.get();
-
+static bool dgst_file_op(const std::string &filename, const int fd,
```
justsmth, line 19: This feels like two functions crammed into one with almost no common code between them (only the first two lines?).  I would have separate `hmac_file_op` and `dgst_file_op` functions.

samuel40791765: Great point! I had been fixated on finding a way to reuse the operations for `dgstTool` and `md5Tool`..
OpenSSL had muddled these together in a similar pattern here and I followed that, will change to make more sense.

justsmth (approved):

andrewhop (approved):

justsmth (approved):

andrewhop (approved):

PR #1771, opened 2024-08-15
Tighten up experimental pointer guard macro

torben-hansen:
### Description of changes: 

Tighten descriptions and ensure return value assumptions can't just be changed.

### DIFF
--- a/crypto/crypto_test.cc
+++ b/crypto/crypto_test.cc
@@ -27,6 +27,25 @@
 #include <gtest/gtest.h>
 #include "test/test_util.h"
 
+static int AWS_LC_ERROR_return(void) {
+  GUARD_PTR(NULL);
+  return 1;
+}
+
+static int AWS_LC_SUCCESS_return(void) {
+  char non_null_ptr[1];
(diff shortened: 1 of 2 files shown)


crypto/crypto_test.cc
```diff
@@ -27,6 +27,25 @@
...
+  return 1;
+}
+
+static int AWS_LC_SUCCESS_return(void) {
+  char non_null_ptr[1];
+  GUARD_PTR(non_null_ptr);
+  return 1;
+}
+
+TEST(CryptoTest, SafetyMacro) {
```
samuel40791765, line 41: We've noticed that `GUARD_PTR` also works great for functions that return a pointer and they'll return `NULL` when checking pointers.
ex:
```
// Will return `NULL`.
static void *AWS_LC_ERROR_return(void) {
  GUARD_PTR(NULL);
  return 1;
}
```
Should we consider adding a test for that?

samuel40791765 (approved):

justsmth (approved):

PR #1690, opened 2024-07-05
Make aes_hw_ctr32_encrypt_blocks handle len=0 correctly

nebeid:
### Issues:
Resolves #CryptoAlg-2498

### Description of changes: 
`aes_hw_ctr32_encrypt_blocks` encrypts (and writes to the output) 2 blocks when the input length is 0 blocks in the case of AArch64 and 1 block in the case of x86_64 and x86.
The function is guarded wherever it's called by checks that `len != 0`.
This change fixes this behaviour without taxing the performance.

### Testing:
Tested the performance on Graviton3 and on Mac x86_64.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

PR #1774, opened 2024-08-16
Improving corner-case instruction flow in `aes_hw_ctr32_encrypt_blocks`

nebeid:
### Issues:
Addresses #CryptoAlg-2498

### Description of changes: 
The change in https://github.com/aws/aws-lc/pull/1690 showed a 0.01 usec more in EVP-AES-256-CTR on Graviton4. This change optimises the sequence of operations to make use of instruction fusion in the sequence CMP + B.cond.; that is, the `b.le` instruction is placed after the cmp instruction, also avoiding executing unneeded instructions if the jump is to be taken. 

### Testing:
Tested the performance on Graviton4.

### DIFF
--- a/crypto/fipsmodule/aes/asm/aesv8-armx.pl
+++ b/crypto/fipsmodule/aes/asm/aesv8-armx.pl
@@ -897,9 +897,9 @@ $code.=<<___;
 
 .Lctr32_tail:
 	cmp			$len,#1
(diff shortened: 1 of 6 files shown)


andrewhop:
Need to update the pre-generated sources, besides that looks good.

andrewhop (approved):

dkostic (approved):

lilpoozie2005 (approved):

PR #1778, opened 2024-08-19
add support for OCSP_request_verify

samuel40791765:
### Issues:
Resolves `CryptoAlg-2420`

### Description of changes: 
`OCSP_request_verify` is the last OCSP API needed for Ruby. OCSP request signatures are optional according to the OCSP RFC, so this is less used than `OCSP_basic_verify`. Much of the logic & flags are reused across each other however. A significant portion of the logic was also reworked without changing the behavior.
Documentation was also rewritten slightly so that it could be applied to both OCSP verification APIs.

### Call-outs:
We don't support `OCSP_NOSIGS` for `OCSP_basic_verify` either, so I've left that out of here for the time being.

### Testing:
New tests


skmcgrail (approved):

crypto/ocsp/ocsp_verify.c
```diff
@@ -388,3 +391,109 @@ int OCSP_basic_verify(OCSP_BASICRESP *bs, STACK_OF(X509) *certs, X509_STORE *st,
...
+    if (signer != NULL) {
+      *psigner = signer;
+      return SIGNER_IN_OCSP_CERTS;
+    }
+  }
+
+  signer = X509_find_by_subject(certs, nm);
+  if (signer != NULL) {
+    *psigner = signer;
+    return SIGNER_IN_TRUSTED_CERTS;
```
torben-hansen, line 414: trusted certs are in the trust store, no? only `OCSP_NOCHAIN` will mean these are "trusted" from a validation perspective.

torben-hansen: And `OCSP_TRUSTOTHER` I guess.

crypto/ocsp/ocsp_verify.c
```diff
@@ -388,3 +391,109 @@ int OCSP_basic_verify(OCSP_BASICRESP *bs, STACK_OF(X509) *certs, X509_STORE *st,
...
+  if (ASN1_item_verify(ASN1_ITEM_rptr(OCSP_REQINFO),
+                       req->optionalSignature->signatureAlgorithm,
+                       req->optionalSignature->signature, req->tbsRequest,
+                       skey) <= 0) {
+    OPENSSL_PUT_ERROR(OCSP, OCSP_R_SIGNATURE_FAILURE);
+    return 0;
+  }
+
+  // Set up |ctx| and start doing the actual verification.
+  X509_STORE_CTX *ctx = X509_STORE_CTX_new();
```
torben-hansen, line 467: memory alloc only really needed if `!IS_OCSP_FLAG_SET(flags, OCSP_NOVERIFY)` is true. But I can also see how it makes it more readable.

include/openssl/ocsp.h
```diff
@@ -54,26 +54,26 @@ extern "C" {
...
+// search for the signing certificate by default, unless |OCSP_NOINTERN| is set.
 #define OCSP_NOINTERN 0x2
-// OCSP_NOCHAIN is for |OCSP_basic_verify|. All certificates in |certs| and in
-// |bs| are considered as untrusted certificates for the construction of the
-// validation path for the signer certificate unless the |OCSP_NOCHAIN| flag is
-// set.
+// OCSP_NOCHAIN is for |OCSP_basic_verify| and |OCSP_request_verify|. All
+// certificates in both |certs| and in |bs|/|req| are considered as untrusted
+// certificates for the construction of the validation path for the signer
+// certificate by default, unless |OCSP_NOCHAIN| is set.
```
torben-hansen, line 64: > unless |OCSP_NOCHAIN| is set

Just checking: the negation implied by "unless" means that the `certs` and all "intern" certs in `bs/req` are not added for the purpose of chain building. It doesn't mean negating "untrusted" to "trusted.

samuel40791765: Yeah this could be clearer. `OCSP_NOCHAIN` means that certificates in `certs` and `bs/req` aren't considered when building the chain.
But on closer examination, `certs` is never considered in `OCSP_request_verify` so the above only applies to `OCSP_basic_verify`. Will change.

crypto/ocsp/ocsp_verify.c
samuel40791765: Yeah I worded this from the perspective of using the `OCSP_TRUSTOTHER` flag.
`OCSP_NOCHAIN` means that the certs provided outside of the trust store are not considered when building the chain.

crypto/ocsp/ocsp_verify.c
samuel40791765: Yeah I originally had this in the bracket and was trying to avoid the `goto end` statements. But pivoted since the freeing was a bit much.

torben-hansen (approved):

skmcgrail (approved):

PR #1780, opened 2024-08-20
Add various PKCS7 getters and setters

WillChilds-Klein:
### Issues:
Resolves CryptoAlg-2482

### Description of changes: 

Following up PR 1726, this PR adds various getter/setter functions needed to support ruby and provides test coverage for them. We also add test coverage for the various ASN.1 functions and `PKCS7_dup` added in PR 1726.

This commit borrows a sample PKCS7 SignedData from https://lapo.it/asn1js/ that, unlike other test fixtures, contains SignerInfos. That sample object has been hand-edited to reduce its size (~10KiB of an inner EncapsulatedContentInfo object were consumed by filler). You can see a parsed illustration of the new fixture [here](https://lapo.it/asn1js/#MIIDVAYJKoZIhvcNAQcCoIIDRTCCA0ECAQExDzANBglghkgBZQMEAgEFADCCACUGCSqGSIb3DQEHAaCCABYEggASSW5pemlvIGNvbnRlbnV0by4KoIIBcjCCAW4wggEUoAMCAQICEGBc0tUmdbOSpaqbArJqVWYwCgYIKoZIzj0EAwIwDzENMAsGA1UEAwwEVGVzdDAeFw0xODA3MTYxNDU2MzVaFw0xOTA3MTYxNDU2MzVaMA8xDTALBgNVBAMMBFRlc3QwWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAAQkTLbMHGkh_0nYviY-C3zQWihlWoRsgl7K5uzmmmohxOTyICTA6fTkdJyYoa3yX5DebvlIK2cYg6cOtLernwZDo1IwUDAOBgNVHQ8BAf8EBAMCBPAwHQYDVR0OBBYEFMPAhN97BA2wOK9RjOOX9uwg1ibmMB8GA1UdIwQYMBaAFMPAhN97BA2wOK9RjOOX9uwg1ibmMAoGCCqGSM49BAMCA0gAMEUCIQDeYB5XPa-1m8VR1Y4-e57aBhLdARKAWiIXtzR1m4hEFwIgZ8P95geA1BwdejuQKR89OcTcLyBtzLovmCwGtnwJsjIxggGKMIIBhgIBATAjMA8xDTALBgNVBAMMBFRlc3QCEGBc0tUmdbOSpaqbArJqVWYwDQYJYIZIAWUDBAIBBQCggfcwGAYJKoZIhvcNAQkDMQsGCSqGSIb3DQEHATAcBgkqhkiG9w0BCQUxDxcNMTgwNzE2MTQ1NjM1WjAqBgkqhkiG9w0BCTQxHTAbMA0GCWCGSAFlAwQCAQUAoQoGCCqGSM49BAMCMC8GCSqGSIb3DQEJBDEiBCByTFG7522gWvsgy-jrA3za4a_XExJdLcE9VS2p9ELSTTBgBgsqhkiG9w0BCRACLzFRME8wTTBLBCC6vAhDTFgmcwHgBoYdJ-yhJWcOeSeXJI52pXF6W_mTwjAnMBOkETAPMQ0wCwYDVQQDDARUZXN0AhBgXNLVJnWzkqWqmwKyalVmMAoGCCqGSM49BAMCBEcwRQIhAPEEMC6rtCjsyOcc4uOPbL7YSKjporXV2hmTPkhcg8xxAiAMu0Cgt7my4EiSBiL7KG_RRpXyTPZQvapAuIX4TDS7uQ).

The next PR in this series will implement PKCS7's BIO-based  data initialization and finalization functions.

### Call-outs:
- none

### Testing:
- CI

---


crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,349 @@ int pkcs7_add_signed_data(CBB *out,
...
+        break;
+    default:
+        OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_WRONG_CONTENT_TYPE);
+        return 0;
+    }
+
+
+    obj = p7i->digest_alg->algorithm;
+    /* If the digest is not currently listed, add it */
+    j = 0;
```
darylmartin100, line 372: nit:  I would flip this to a boolean and give it a better name than j

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,349 @@ int pkcs7_add_signed_data(CBB *out,
...
+    if (!sk_PKCS7_SIGNER_INFO_push(signer_sk, p7i)) {
+        return 0;
+    }
+    return 1;
+}
+
+ASN1_TYPE *PKCS7_get_signed_attribute(const PKCS7_SIGNER_INFO *si, int nid) {
+    for (size_t i = 0; i < sk_X509_ATTRIBUTE_num(si->auth_attr); i++) {
+        X509_ATTRIBUTE *attr = sk_X509_ATTRIBUTE_value(si->auth_attr, i);
+        ASN1_OBJECT *obj = X509_ATTRIBUTE_get0_object(attr);
```
darylmartin100, line 413: Null check on attr and obj?

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,349 @@ int pkcs7_add_signed_data(CBB *out,
...
+int PKCS7_SIGNER_INFO_set(PKCS7_SIGNER_INFO *p7i, X509 *x509, EVP_PKEY *pkey,
+                          const EVP_MD *dgst) {
+    /* We now need to add another PKCS7_SIGNER_INFO entry */
+    if (!p7i || !dgst || !pkey || !dgst)
+        return 0;
+    if (!ASN1_INTEGER_set(p7i->version, 1))
+        return 0;
+    if (!X509_NAME_set(&p7i->issuer_and_serial->issuer,
+                       X509_get_issuer_name(x509)))
+        return 0;
```
darylmartin100, line 441: nit: The indentation of the if statement and the lack of braces made me think everything below the return was unreachable code.

WillChilds-Klein: yeah, this is hold-over from OpenSSL's (ad-hoc) style conventions. we should be using explicit braces more liberally.

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,349 @@ int pkcs7_add_signed_data(CBB *out,
 
   return CBB_flush(out);
 }
+
+int PKCS7_set_type(PKCS7 *p7, int type)
+{
+    ASN1_OBJECT *obj;
+
+    obj = OBJ_nid2obj(type);    /* will not fail */
```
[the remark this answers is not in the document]
WillChilds-Klein: thanks, fixed.

crypto/pkcs7/pkcs7.c
WillChilds-Klein: i don't think all of our supported compilers supports `C`'s `bool` type, or at least i don't see other usages of it in the `C` portions of our codebase. agreed that `j` is not a good name though.

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,349 @@ int pkcs7_add_signed_data(CBB *out,
...
+        return 0;
+    }
+    return 1;
+}
+
+int PKCS7_content_new(PKCS7 *p7, int type)
+{
+    PKCS7 *ret = NULL;
+
+    if ((ret = PKCS7_new()) == NULL)
```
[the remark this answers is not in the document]
WillChilds-Klein: sure

crypto/pkcs7/pkcs7.c
WillChilds-Klein: `attr` shouldn't be null, but `X509_ATTRIBUTE_get0_object` does a null check on it anyway. `OBJ_obj2nid` does a null check on `obj`.

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,385 @@ int pkcs7_add_signed_data(CBB *out,
...
+        sk = p7->d.enveloped->recipientinfo;
+        break;
+    default:
+        OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_WRONG_CONTENT_TYPE);
+        return 0;
+    }
+
+    if (!sk_PKCS7_RECIP_INFO_push(sk, ri))
+        return 0;
+    return 1;
```
samuel40791765, line 365: Nit:
```suggestion
    if (!sk_PKCS7_RECIP_INFO_push(sk, ri)) {
        return 0;
    }
    return 1;
```
Also same switch statement optimization above

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,385 @@ int pkcs7_add_signed_data(CBB *out,
...
+            || (alg->parameter = ASN1_TYPE_new()) == NULL) {
+            X509_ALGOR_free(alg);
+            OPENSSL_PUT_ERROR(PKCS7, ERR_R_ASN1_LIB);
+            return 0;
+        }
+        /*
+         * If there is a constant copy of the ASN1 OBJECT in libcrypto, then
+         * use that.  Otherwise, use a dynamically duplicated copy
+         */
+        nid = OBJ_obj2nid(obj);
```
samuel40791765, line 412: We can move the initialization down. 
```suggestion
        int nid = OBJ_obj2nid(obj);
```

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,385 @@ int pkcs7_add_signed_data(CBB *out,
...
+        return p7->d.signed_and_enveloped->signer_info;
+    }
+    return NULL;
+}
+
+int PKCS7_SIGNER_INFO_set(PKCS7_SIGNER_INFO *p7i, X509 *x509, EVP_PKEY *pkey,
+                          const EVP_MD *dgst) {
+    /* We now need to add another PKCS7_SIGNER_INFO entry */
+    if (!p7i || !dgst || !pkey || !dgst) {
+        return 0;
```
samuel40791765, line 460: Should we push ERR_R_PASSED_NULL_PARAMETER onto the stack here? Same for the other function parameter null checks else where

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,385 @@ int pkcs7_add_signed_data(CBB *out,
...
+        }
+    }
+
+    X509_up_ref(x509);
+    p7i->cert = x509;
+
+    return 1;
+}
+
+void PKCS7_SIGNER_INFO_get0_algs(PKCS7_SIGNER_INFO *si, EVP_PKEY **pk,
```
samuel40791765, line 559: Should we do a NULL check for `si`?

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,385 @@ int pkcs7_add_signed_data(CBB *out,
...
+    if (pdig) {
+        *pdig = si->digest_alg;
+    }
+    if (psig) {
+        *psig = si->digest_enc_alg;
+    }
+}
+
+void PKCS7_RECIP_INFO_get0_alg(PKCS7_RECIP_INFO *ri, X509_ALGOR **penc)
+{
```
samuel40791765, line 574: NULL check for `ri`?

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,385 @@ int pkcs7_add_signed_data(CBB *out,
...
+        return 0;
+    }
+
+    ASN1_INTEGER_free(p7i->issuer_and_serial->serial);
+    if (!(p7i->issuer_and_serial->serial =
+          ASN1_INTEGER_dup(X509_get0_serialNumber(x509)))) {
+        return 0;
+    }
+
+    pkey = X509_get0_pubkey(x509);
```
samuel40791765, line 537:
```suggestion
    EVP_PKEY *pkey = X509_get0_pubkey(x509);
```

include/openssl/pkcs7.h
```diff
@@ -183,6 +183,75 @@ OPENSSL_EXPORT PKCS7 *d2i_PKCS7_bio(BIO *bio, PKCS7 **out);
...
+
+// PKCS7_RECIP_INFO_set attaches |x509| to |p7i| and increments |x509|'s
+// reference count. It returns 1 on success and 0 on failure or if |x509|'s
+// public key not usable for encryption.
+OPENSSL_EXPORT int PKCS7_RECIP_INFO_set(PKCS7_RECIP_INFO *p7i, X509 *x509);
+
+// PKCS7_SIGNER_INFO_set attaches the other parameters to |p7i|, returning 1 on
+// success and 0 on error or if specified parameters are inapplicable to
+// signing. Only EC, DH, and RSA |pkey|s are supported. |pkey|'s reference
+// count is incremented, but neither |x509|'s nor |dgst|'s is.
```
samuel40791765, line 203:
```suggestion
// count is incremented and assigned to `p7i`.
```
It might help to mention this. The original wording makes it seem like `pkey` needs to be explicitly freed twice.

samuel40791765 (commented):
1. A lot of the PKCS7 setters/getters seem to only apply to applicable or relevant types. Should we add which specific types are relevant in the function documentation? 
2. Nit: There are some remnants of OpenSSL's comment style in `crypto/pkcs7/pkcs7.c` that we should switch up. It'd also help to run `clang-format` on the file since the brackets and spacing are a bit inconsistent.

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,385 @@ int pkcs7_add_signed_data(CBB *out,
 
   return CBB_flush(out);
 }
+
+int PKCS7_set_type(PKCS7 *p7, int type) {
+    ASN1_OBJECT *obj;
+
+    obj = OBJ_nid2obj(type);
```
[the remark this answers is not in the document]
WillChilds-Klein: sure

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,385 @@ int pkcs7_add_signed_data(CBB *out,
...
+    return 1;
+}
+
+int PKCS7_set_cipher(PKCS7 *p7, const EVP_CIPHER *cipher)
+{
+    int i;
+    PKCS7_ENC_CONTENT *ec;
+
+    i = OBJ_obj2nid(p7->type);
+    switch (i) {
```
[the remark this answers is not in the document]
WillChilds-Klein: sure

include/openssl/pkcs7.h
```diff
@@ -183,6 +183,75 @@ OPENSSL_EXPORT PKCS7 *d2i_PKCS7_bio(BIO *bio, PKCS7 **out);
...
+
+// PKCS7_content_new allocates a new PKCS7 and adds it to |p7| as content. It
+// returns 1 on success and 0 on failure.
+OPENSSL_EXPORT int PKCS7_content_new(PKCS7 *p7, int nid);
+
+// PKCS7_set_cipher sets |cipher| on |p7| for applicable types of |p7|. It
+// returns 1 on success and 0 on failure.
+OPENSSL_EXPORT int PKCS7_set_cipher(PKCS7 *p7, const EVP_CIPHER *cipher);
+
+// PKCS7_set_content sets |p7_data| as content on |p7| for applicaple types of
```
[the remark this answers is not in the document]
WillChilds-Klein: sure, makes sense.

crypto/pkcs7/pkcs7.c
WillChilds-Klein: sure, makes sense

crypto/pkcs7/pkcs7.c
WillChilds-Klein: sure

crypto/pkcs7/pkcs7.c
WillChilds-Klein: sure

WillChilds-Klein:
@samuel40791765 --

> 1. A lot of the PKCS7 setters/getters seem to only apply to applicable or relevant types. Should we add which specific types are relevant in the function documentation?

sure, i can do that.

>     2. Nit: There are some remnants of OpenSSL's comment style in `crypto/pkcs7/pkcs7.c` that we should switch up. It'd also help to run `clang-format` on the file since the brackets and spacing are a bit inconsistent.

that's reasonable, although I'd like to do it in a separate PR that only applies `clang-format`.

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,384 @@ int pkcs7_add_signed_data(CBB *out,
...
+        /*
+         * If there is a constant copy of the ASN1 OBJECT in libcrypto, then
+         * use that.  Otherwise, use a dynamically duplicated copy
+         */
+        int nid = OBJ_obj2nid(obj);
+        if (nid != NID_undef) {
+            alg->algorithm = OBJ_nid2obj(nid);
+        } else {
+            alg->algorithm = OBJ_dup(obj);
+        }
```
justsmth, line 407: Although there's no `const` in its signature, it looks like `OBJ_nid2obj` always returns a builtin or static value that should not be freed.(?)

How does `X509_ALGOR_free` know whether or not to free `alg->algorithm`?

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,384 @@ int pkcs7_add_signed_data(CBB *out,
...
+
+STACK_OF(PKCS7_SIGNER_INFO) *PKCS7_get_signer_info(PKCS7 *p7) {
+    if (p7 == NULL || p7->d.ptr == NULL) {
+        OPENSSL_PUT_ERROR(PKCS7, ERR_R_PASSED_NULL_PARAMETER);
+        return NULL;
+    } else if (PKCS7_type_is_signed(p7)) {
+        return p7->d.sign->signer_info;
+    } else if (PKCS7_type_is_signedAndEnveloped(p7)) {
+        return p7->d.signed_and_enveloped->signer_info;
+    }
```
justsmth, line 444: * Why not a `switch (OBJ_obj2nid(p7->type))` here as we do in other functions?
* We check for `p7 == NULL` here -- should we be doing that in other similar functions?

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,384 @@ int pkcs7_add_signed_data(CBB *out,
...
+    case NID_pkcs7_signedAndEnveloped:
+        p7->type = obj;
+        p7->d.signed_and_enveloped = PKCS7_SIGN_ENVELOPE_new();
+        if (p7->d.signed_and_enveloped == NULL) {
+            return 0;
+        }
+        if (!ASN1_INTEGER_set(p7->d.signed_and_enveloped->version, 1)) {
+            return 0;
+        }
+        p7->d.signed_and_enveloped->enc_data->content_type = OBJ_nid2obj(NID_pkcs7_data);
```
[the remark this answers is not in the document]
justsmth: Ok, for known values like `NID_pkcs7_data` it shouldn't fail as long as [kObjects](https://github.com/aws/aws-lc/blob/e0cc91b199bf816ee5add58d4efdcae5b552ac02/crypto/obj/obj_dat.h#L7235-L8909) contains a value for it.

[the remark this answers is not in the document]
WillChilds-Klein: right. i went through the other usages of OBJ_nid2obj in this PR where caller can determine the input. all of them either handled the failures case or were called around a `const EVP_MD *`, which should always result in non-null result.

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,384 @@ int pkcs7_add_signed_data(CBB *out,
...
+        p7->d.encrypted->enc_data->content_type = OBJ_nid2obj(NID_pkcs7_data);
+        break;
+    default:
+        OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_UNSUPPORTED_CONTENT_TYPE);
+        return 0;
+    }
+    return 1;
+}
+
+int PKCS7_set_cipher(PKCS7 *p7, const EVP_CIPHER *cipher)
```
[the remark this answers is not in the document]
WillChilds-Klein: yes, good call, we should leave input unmodified in the failure case. i'll move up the cipher check to be the first thing in this function.

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,384 @@ int pkcs7_add_signed_data(CBB *out,
...
+        break;
+    case NID_pkcs7_digest:
+        PKCS7_free(p7->d.digest->contents);
+        p7->d.digest->contents = p7_data;
+        break;
+    case NID_pkcs7_data:
+    case NID_pkcs7_enveloped:
+    case NID_pkcs7_signedAndEnveloped:
+    case NID_pkcs7_encrypted:
+    default:
```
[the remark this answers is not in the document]
WillChilds-Klein: i don't think so, it's redundant. here and elsewhere, i'll remove fall-throughs to the `default` case.

crypto/pkcs7/pkcs7.c
WillChilds-Klein: > Why not a switch (OBJ_obj2nid(p7->type)) here as we do in other functions?

sure, i can use the case/switch  style here.

> We check for p7 == NULL here -- should we be doing that in other similar functions?

i don't see a good reason not to. i've gone through functions added in this PR and added param null checks.

crypto/pkcs7/pkcs7.c
WillChilds-Klein: the `X509_ALGOR` sequence is [defined](https://github.com/aws/aws-lc/blob/3571a86680b191986b6657c4a16383386a091ae3/crypto/x509/x_algor.c#L67) as:

```
ASN1_SEQUENCE(X509_ALGOR) = {
    ASN1_SIMPLE(X509_ALGOR, algorithm, ASN1_OBJECT),
    ASN1_OPT(X509_ALGOR, parameter, ASN1_ANY),
} ASN1_SEQUENCE_END(X509_ALGOR)
```

with the `algorithm` member of type `ASN1_OBJECT`, which tells the macro system to use `ASN1_OBJECT_free` when implementing `X509_ALGOR_free`. `ASN1_OBJECT_free` is [smart enough](https://github.com/aws/aws-lc/blob/3571a86680b191986b6657c4a16383386a091ae3/crypto/asn1/a_object.c#L199-L209) to check `asn1_obj_st`'s [flags](https://github.com/aws/aws-lc/blob/3571a86680b191986b6657c4a16383386a091ae3/crypto/asn1/internal.h#L117) for whether it was dynamically allocated\*, and no-op if not.

\* as you mention [`OBJ_nid2obj` ](https://github.com/aws/aws-lc/blob/3571a86680b191986b6657c4a16383386a091ae3/crypto/obj/obj.c#L354) consults [`kObjects`](https://github.com/aws/aws-lc/blob/main/crypto/obj/obj_dat.h#L7235) to look up a static object to return. all members of `kObjects` have flag (last struct member) [of 0](https://github.com/aws/aws-lc/blob/3571a86680b191986b6657c4a16383386a091ae3/crypto/asn1/internal.h#L117), so none of the [`ASN1_OBJECT_FLAG_DYNAMIC_*` flags](https://github.com/aws/aws-lc/blob/3571a86680b191986b6657c4a16383386a091ae3/crypto/asn1/internal.h#L100) get set.

crypto/pkcs7/pkcs7_x509.c
```diff
@@ -428,3 +429,78 @@ PKCS7 *PKCS7_sign(X509 *sign_cert, EVP_PKEY *pkey, STACK_OF(X509) *certs,
...
+    if (*sk == NULL) {
+        OPENSSL_PUT_ERROR(PKCS7, ERR_R_CRYPTO_LIB);
+        return 0;
+    }
+
+    if (!sk_X509_insert(*sk, x509, 0)) {
+        X509_free(x509);
+        return 0;
+    }
+    X509_up_ref(x509);
```
samuel40791765, line 466: 1. Minor question: Is there a reason why we insert `x509` as the first element on the stack (contrary to  `PKCS7_add_crl`)? I notice [OpenSSL 1.1.1](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/pkcs7/pk7_lib.c) does as a push onto the stack for both.
2. I think we should get rid of the calls to `X509_free(x509);` here and `X509_CRL_free(crl);` on L504 below. This exists in OpenSSL because the reference count was incremented before the call.
But now that we've moved the reference call below, this could cause unintentional freeing of the user's `x509/crl` if the call fails.

crypto/pkcs7/pkcs7_test.cc
```diff
@@ -1044,3 +1129,212 @@ hJTbHtjEDJ7BHLC/CNUhXbpyyu1y
...
+    PKCS7_SIGNER_INFO_get0_algs(p7si, &pkey, &pdig, &psig);
+    ASSERT_FALSE(pkey); // no attached pkey
+    ASSERT_TRUE(psig);
+    ASSERT_TRUE(pdig);
+
+    bssl::UniquePtr<PKCS7> p7_dup(PKCS7_dup(p7.get()));
+    ASSERT_TRUE(p7_dup);
+    EXPECT_TRUE(PKCS7_type_is_signed(p7_dup.get()));
+
+    p7_der = kPKCS7SignedWithSignerInfo;
```
samuel40791765, line 1211: Minor nit: is this necessary, seems like the pointer value hasn't changed.

crypto/pkcs7/pkcs7.c
samuel40791765, line 1: One last nit, it'd be best to swap out OpenSSL's comment style for ours in this file.

WillChilds-Klein: this should be covered by the `clang-format` follow-up PR.

crypto/pkcs7/pkcs7_test.cc
WillChilds-Klein: are you asking about the `PKCS7_dup` call on L1207? if so, that's meant to exercise `PKCS7_dup` and assert that a.) `PKCS7_dup` duplicates the object without memory leaks and b.) that the duplicated object is of the expected content type.

if asking about the `p7_der` assignment on L1211, then yes, and nice catch :) it's intentional. `d2i_*` functions will increment the input reference during parsing (see comment [here](https://github.com/aws/aws-lc/blob/f88ef5dff488c3a36f65390ff5a0a8e92e3d941b/crypto/pkcs7/pkcs7_asn1.c#L55)). this is why we assign a local pointer to the static address and reset it for relevant tests. OpenSSL documents the generic behavior in [this type-specific man page](https://docs.openssl.org/1.1.1/man3/d2i_X509).

> d2i_TYPE() attempts to decode len bytes at *ppin. If successful a pointer to the TYPE structure is returned and *ppin is incremented to the byte following the parsed data.

thanks for the call-out, this needs to be documented in a comment.

include/openssl/pkcs7.h
```diff
@@ -183,6 +183,77 @@ OPENSSL_EXPORT PKCS7 *d2i_PKCS7_bio(BIO *bio, PKCS7 **out);
...
+
+// PKCS7_set_cipher sets |cipher| on |p7| for applicable types of |p7|. It
+// returns 1 on success and 0 on failure or if |p7| is not an applicable type:
+// envelopedData and signedAndEnvelopedData.
+OPENSSL_EXPORT int PKCS7_set_cipher(PKCS7 *p7, const EVP_CIPHER *cipher);
+
+// PKCS7_set_content sets |p7_data| as content on |p7| for applicable types of
+// |p7|: signedData and digestData. It frees any existing content on |p7|,
+// returning 1 on success and 0 on failure.
+OPENSSL_EXPORT int PKCS7_set_content(PKCS7 *p7, PKCS7 *p7_data);
```
justsmth, line 238: NP: Should we document that `p7_data` is allowed to be NULL? (Is this a viable use-case?)

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,415 @@ int pkcs7_add_signed_data(CBB *out,
...
+        return p7->d.signed_and_enveloped->signer_info;
+    default:
+        return NULL;
+    }
+}
+
+int PKCS7_SIGNER_INFO_set(PKCS7_SIGNER_INFO *p7i, X509 *x509, EVP_PKEY *pkey,
+                          const EVP_MD *dgst) {
+    /* We now need to add another PKCS7_SIGNER_INFO entry */
+    if (!p7i || !dgst || !pkey || !dgst) {
```
justsmth, line 482: It looks like  you have a typo -- this skips `x509` and has `dgst` twice.
```
if (!p7i || !x509 || !pkey || !dgst) {
```

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,415 @@ int pkcs7_add_signed_data(CBB *out,
...
+    ASN1_OBJECT *obj = OBJ_nid2obj(type);
+    if (obj == NULL) {
+        OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_UNSUPPORTED_CONTENT_TYPE);
+        return 0;
+    }
+
+    switch (type) {
+    case NID_pkcs7_signed:
+        p7->type = obj;
+        p7->d.sign = PKCS7_SIGNED_new();
```
justsmth, line 212: Prior to the `p7->d.sign = PKCS7_SIGNED_new()` assignment, do we need to check whether a value is already assigned and free it if so? (Likewise for the other cases.)

crypto/pkcs7/pkcs7_x509.c
WillChilds-Klein: 1. i'm not sure how that got in there, it looks like both [mainline](https://github.com/openssl/openssl/blob/master/crypto/pkcs7/pk7_lib.c#L298) and [1.0.2](https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/crypto/pkcs7/pk7_lib.c#L347) use `sk_X509_push` as well. i've updated this PR accordingly.
2. agreed.

crypto/pkcs7/pkcs7.c
WillChilds-Klein: yes, thanks.

crypto/pkcs7/pkcs7.c
WillChilds-Klein: yes, nice catch. i'll update this and the tests.

include/openssl/pkcs7.h
WillChilds-Klein: OpenSSL 1.1.1 [allows it](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/pkcs7/pk7_lib.c#L80), so I did so here as well. I can document that in the comment.

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,421 @@ int pkcs7_add_signed_data(CBB *out,
...
+            return 0;
+        }
+        hnid = OBJ_obj2nid(alg1->algorithm);
+        if (hnid == NID_undef) {
+            return 0;
+        } else if (!OBJ_find_sigid_by_algs(&snid, hnid, EVP_PKEY_id(pkey))) {
+            return 0;
+        } else if (!X509_ALGOR_set0(alg2, OBJ_nid2obj(snid), V_ASN1_UNDEF, NULL)) {
+            return 0;
+        }
```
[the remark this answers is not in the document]
WillChilds-Klein: sure

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,421 @@ int pkcs7_add_signed_data(CBB *out,
...
+     */
+    ASN1_INTEGER_free(p7i->issuer_and_serial->serial);
+    if (!(p7i->issuer_and_serial->serial =
+          ASN1_INTEGER_dup(X509_get0_serialNumber(x509)))) {
+        return 0;
+    }
+
+    /* lets keep the pkey around for a while */
+    EVP_PKEY_up_ref(pkey);
+    p7i->pkey = pkey;
```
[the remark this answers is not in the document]
WillChilds-Klein: OpenSSL [does not](https://github.com/openssl/openssl/blob/8af4c02ea952ca387691c4a077c260ba045fe285/crypto/pkcs7/pk7_lib.c#L339), so it's possible that caller's would be surprised by us diverging from that behavior and freeing their resources.

crypto/pkcs7/pkcs7.c
```diff
@@ -191,3 +194,421 @@ int pkcs7_add_signed_data(CBB *out,
...
+        X509_ALGOR *alg;
+        PKCS7_RECIP_INFO_get0_alg(p7i, &alg);
+        if (!X509_ALGOR_set0(alg, OBJ_nid2obj(NID_rsaEncryption),
+                                   V_ASN1_NULL, NULL)) {
+            return 0;
+        }
+    }
+
+    X509_up_ref(x509);
+    p7i->cert = x509;
```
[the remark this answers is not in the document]
WillChilds-Klein: OpenSSL [does not](https://github.com/openssl/openssl/blob/8af4c02ea952ca387691c4a077c260ba045fe285/crypto/pkcs7/pk7_lib.c#L617), so it's possible that caller's would be surprised by us diverging from that behavior and freeing their resources.

include/openssl/pkcs7.h
```diff
@@ -183,6 +183,77 @@ OPENSSL_EXPORT PKCS7 *d2i_PKCS7_bio(BIO *bio, PKCS7 **out);
...
+
+// PKCS7_RECIP_INFO_get0_alg sets |*penc| to |ri|'s key encryption algorithm,
+// if present.
+OPENSSL_EXPORT void PKCS7_RECIP_INFO_get0_alg(PKCS7_RECIP_INFO *ri,
+                                              X509_ALGOR **penc);
+
+// PKCS7_SIGNER_INFO_get0_algs sets all of, if present: |*pk| to |si|'s key,
+// |*pdig| to |si|'s digest angorithm, and |*psig| to |si|'s signature
+// algorithm.
+OPENSSL_EXPORT void PKCS7_SIGNER_INFO_get0_algs(PKCS7_SIGNER_INFO *si,
```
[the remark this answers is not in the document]
WillChilds-Klein: sure, will do.

crypto/pkcs7/pkcs7.c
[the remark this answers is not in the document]
WillChilds-Klein: discussed offline. given that we've decided to make the `PKCS7_*_INFO` structs opaque, caller has less ability to track and free the PKEY themselves. this was mostly likely an oversight in openssl's implementation, so we free the pkey here.

crypto/pkcs7/pkcs7.c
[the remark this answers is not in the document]
WillChilds-Klein: discussed offline. given that we've decided to make the `PKCS7_*_INFO` structs opaque, caller has less ability to track and free the PKEY themselves. this was mostly likely an oversight in openssl's implementation, so we free the cert here.

justsmth (approved):

crypto/pkcs7/pkcs7_test.cc
samuel40791765: Ah yes I was asking about `p7_der` on L1211. It seemed like the `d2i` functions did some pointer advancements, but the `const` pointer made me question myself if anything had actually changed. Thanks for the clarification!!

samuel40791765 (approved):

PR #1783, opened 2024-08-20
Adding a runtime dis/enabler of DIT Capability on AArch64.

nebeid:
### Issues:
Addresses #CryptoAlg-2503

### Description of changes: 
- Provide runtime functions that mask out (and back in) the DIT CPU capability by clearing (setting) an additional bit in `OPENSSL_armcap_P`. This mechanism was chosen for the following reasons:
  - It does not require an additional global variable.
  - It avoids extra checks on the path of setting/resetting the DIT bit.
  - It avoids re-evaluating the CPU capability if we were to clear the DIT capability bit itself. That latter bit is now left intact.
There were write locks added around changing `OPENSSL_armcap_P`. However, Thread Sanitizer warned about data race possibilities when trying to run a test with concurrent threads where one disables DIT at runtime and the other tries to check for the capability. Therefore, they are documented with a warning to use them only in initialization contexts.

- Make the DIT functions (enable/disable and set/restore) available regardless of whether the build flag `DENABLE_DATA_INDEPENDENT_TIMING_AARCH64=ON` was used or not.
  - If the build flag was not used, then the DIT flag is not set and reset with every function and the instructions used for setting it and resetting after checking the capability are omitted and don't incur extra cost.
  - The user now has the choice, regardless of the build flag, to place `armv8_set/restore_dit` in the user's code. 

### Call-outs:
- The (external) API `armv8_enable_dit` is renamed to `armv8_set_dit`. 
- `armv8_enable_dit` now means enable back the capability at runtime.

### Testing:
- The new functions `armv8_disable_dit()` and `arm_enable_dit()` were placed in `Speed()`, the benchmarking function, and their effects were confirmed on Apple M2.
- There are also thread tests to confirm that the CPU DIT bit is context-switched at the thread level. The runtime dis/enabler is at the process scope because it manipulates `OPENSSL_armcap_P`.


crypto/fipsmodule/cpucap/cpu_aarch64_apple.c
```diff
@@ -99,11 +99,9 @@ void OPENSSL_cpuid_setup(void) {
     OPENSSL_armcap_P |= ARMV8_APPLE_M1;
   }
 
-#if defined(MAKE_DIT_AVAILABLE)
   if (has_hw_feature("hw.optional.arm.FEAT_DIT")) {
-    OPENSSL_armcap_P |= ARMV8_DIT;
+    OPENSSL_armcap_P |= (ARMV8_DIT | ARMV8_DIT_ALLOWED);
   }
-#endif  // MAKE_DIT_AVAILABLE
```
justsmth, line 105: Why remove the `MAKE_DIT_AVAILABLE` guard?

nebeid: To enable the CPU capability check regardless of the build flag. 
I had missed to save the same change in `cpu_aarch64_liunx.c` (that's what I forced pushed).
Now, the build flag is only responsible for making the macro at the beginning of the APIs effective.

justsmth: Ok, what the `MAKE_DIT_AVAILABLE` build flag does now is no longer to "make DIT available". Instead it determines whether our internal `SET_DIT_AUTO_DISABLE` macro is a no-op or not.  

It seems like a better name for the flag might be something like `AUTO_ENABLE_DISABLE_SET_DIT`?

CMakeLists.txt
```diff
@@ -31,7 +31,8 @@ option(ENABLE_DILITHIUM "Enable Dilithium signatures in the EVP API" OFF)
 option(DISABLE_PERL "Disable Perl for AWS-LC" OFF)
 option(DISABLE_GO "Disable Go for AWS-LC" OFF)
 option(ENABLE_FIPS_ENTROPY_CPU_JITTER "Enable FIPS entropy source: CPU Jitter" OFF)
-option(ENABLE_DATA_INDEPENDENT_TIMING_AARCH64 "Enable Data-Independent Timing (DIT) flag on Arm64" OFF)
+option(ENABLE_DATA_INDEPENDENT_TIMING_AARCH64 "Enable automatic setting/resetting Data-Independent Timing
```
justsmth, line 34: * Should we also change the name here to `AUTO_ENABLE_DATA_INDEPENDENT_TIMING`?
* Should we make this build flag platform-agnostic, so we can re-use it for any platform (not just AARCH64) that requires special handling for DIT?

WillChilds-Klein: regarding the second bullet point, if we do that we need to be clear that this is only available on aarch64, either through documentation or perhaps aborting if it's set to ON on x86.

tool/speed.cc
```diff
@@ -2682,10 +2681,12 @@ bool Speed(const std::vector<std::string> &args) {
     }
   }
 #if defined(DIT_OPTION)
+  armv8_disable_dit(); // disable DIT capability at run-time
```
WillChilds-Klein, line 2684: why do we need this call?

tool/speed.cc
```diff
@@ -2545,8 +2544,8 @@ static const argument_t kArguments[] = {
     {
         "-dit",
```
WillChilds-Klein, line 2545: what happens if this flag is set on non-aarch64 platform?

include/openssl/arm_arch.h
```diff
@@ -90,7 +90,15 @@
 #define ARMV8_NEOVERSE_V2 (1 << 14)
 
 // ARMV8_DIT indicates support for the Data-Independent Timing (DIT) flag.
-#define ARMV8_DIT (1 << 14)
+#define ARMV8_DIT (1 << 15)
```
WillChilds-Klein, line 93: this indicates that DIT is supported on the current CPU, right? if so, suggest making the name more descriptive like `ARMV8_DIT_SUPPORTED`

include/openssl/crypto.h
```diff
@@ -86,33 +86,48 @@ OPENSSL_EXPORT int CRYPTO_needs_hwcap2_workaround(void);
 
 // Data-Independent Timing (DIT) on AArch64
 
-#if defined(OPENSSL_AARCH64) && !defined(OPENSSL_WINDOWS) && defined(MAKE_DIT_AVAILABLE)
+#if defined(OPENSSL_AARCH64) && !defined(OPENSSL_WINDOWS)
 // (TODO): See if we can detect the DIT capability in Windows environment
 
-// armv8_enable_dit sets the DIT flag to 1 and returns its original value
+// armv8_set_dit sets the DIT flag to 1 and returns its original value
```
WillChilds-Klein, line 92: which DIT flag? we have two now.

include/openssl/crypto.h
```diff
@@ -86,33 +86,48 @@ OPENSSL_EXPORT int CRYPTO_needs_hwcap2_workaround(void);
...
 // before it was called.
-uint64_t armv8_enable_dit(void);
+OPENSSL_EXPORT uint64_t armv8_set_dit(void);
```
WillChilds-Klein, line 94: do we need to make the caller track this? can we store it ourselves as a thread-local variable?

crypto/fipsmodule/cpucap/cpu_aarch64.c
```diff
@@ -82,11 +82,20 @@ uint64_t armv8_enable_dit(void) {
 }
 
 void armv8_restore_dit(volatile uint64_t *original_dit) {
-  if (CRYPTO_is_ARMv8_DIT_capable() && *original_dit != 1) {
+  if (*original_dit != 1 && CRYPTO_is_ARMv8_DIT_capable()) {
```
WillChilds-Klein, line 85:
```suggestion
  if (original_dit != NULL && *original_dit != 1 && CRYPTO_is_ARMv8_DIT_capable()) {
```

crypto/fipsmodule/cpucap/cpu_aarch64.c
```diff
@@ -82,11 +82,20 @@ uint64_t armv8_enable_dit(void) {
...
     // Encoding of "msr dit, #0"
     __asm__ volatile(".long 0xd503405f");
   }
 }
-#endif  // MAKE_DIT_AVAILABLE && !OPENSSL_WINDOWS
+
+void armv8_disable_dit(void) {
+  OPENSSL_armcap_P &= ~ARMV8_DIT_ALLOWED;
```
WillChilds-Klein, line 92: here and for `armv8_enable_dit`, is this operation thread-safe? not sure what the `HIDDEN` macro expands to, but it's [not clear](https://github.com/aws/aws-lc/blob/c28d0cb27759b09eb305a98f822c0223eeb2c727/crypto/fipsmodule/cpucap/cpucap.c#L84) that `OPENSSL_armcap_P` is thread-local.

WillChilds-Klein (commented):
I see changes to `cpu_aarch64_linux.c` and `cpu_aarch64_apple.c`. do we need to udpate `cpu_aarch64_win.c` as well?

CMakeLists.txt
justsmth: Isn't the current assumption on x86-64 (and other platforms) that the *timing* of our encryption operations are (or are expected to be) *data-independent*? So enabling it elsewhere is essentially a no-op for the build.

WillChilds-Klein: > Isn't the current assumption on x86-64 (and other platforms) ....

Oh, I wasn't aware of this. If so, agreed.

crypto/fipsmodule/cpucap/cpu_aarch64.c
torben-hansen: `OPENSSL_armcap_P` is a global variable of type `uint32_t` which is shared a cross the entire address space. 

Firstly, since the memory is shared it doesn't allow a thread to control whether it wants to enable DIT mode or not; its choice can be overwritten by any other thread.

Secondly, multiple writes/reads can now occur concurrently on `OPENSSL_armcap_P` since it's mostly just referenced directly and there therefore aren't any synchonisation. This was fine because it was only really readable after the initial construction of the memory content (the initialisation of the arm capability vector). This PR breaks that assumption.

Finally, is this a "threading issue". Well, yes and no. One thread can read the value while another writes to it. The writing operation only mutate a single byte of the 32-bit value. So, the value when read is either the bit set or not. It won't be a bogus value. So, no. On the other hand, from the first point I made above, any thread can mutate the value. So one thread can enable, while another immediately afterwards can disable the DIT. But that might just be what we can build - DIT is a process global CPU control no?

I think we weakly considered `OPENSSL_armcap_P` nominally `const` after the initial initialisation. It might not be the ideal memory to store this information.

WillChilds-Klein: >  from the first point I made above, any thread can mutate the value. So one thread can enable, while another immediately afterwards can disable the DIT.

yes, this is my concern.

> But that might just be what we can build - DIT is a process global CPU control no?

My understanding from nevine's description is that it can be configured per-thread. I suppose we could leave it up to the caller to not call these functions from >1 thread, and document this limitation very clearly. I don't think this would be an issue for ACCP, as we'd just set or unset DIT once on initialization, but it could be a concern for other consumers.

include/openssl/crypto.h
justsmth: IMO, `armv8_set_dit` and `armv8_restore_dit` should be internal-only implementation details. They shouldn't be declared in this file.

BUILDING.md
```diff
@@ -255,19 +257,34 @@ It is also expected to disable the Data Memory-dependent Prefetcher
...
-    ./tool/bssl speed -dit
\ No newline at end of file
+setting and resetting the DIT flag. If it remains set over many calls,
+the effect can be largely mitigated. Hence, the macro can be inserted
+in the caller's application at the beginning of the code scope that
+makes repeated calls to AWS-LC cryptographic functions.
+
+Alternatively, the functions that are invoked in the macro,
+`armv8_set_dit` and `armv8_restore_dit`, can be placed at the
+beginning and the end of the code section, respectively. These
```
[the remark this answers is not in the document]
nebeid: We could, but I still have them used in `Speed()` in speed.cc when the `-dit` option is chosen, which is mentioned in the rest of the paragraph. It's to illustrate the effect of keeping CPU DIT bit as 1 vs. flipping it back and forth (the default). This was added in the original change. I was thinking to keep it and mention that the functions are internal. What do you think?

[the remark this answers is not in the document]
nebeid: Thanks, Will. I modified this part. Let me know if it reads better now.

crypto/evp_extra/p_dh_asn1.c
```diff
@@ -15,6 +15,7 @@
 
 #include "internal.h"
 #include "../internal.h"
+#include "../fipsmodule/cpucap/internal.h"
```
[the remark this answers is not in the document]
nebeid: Not sure why it's not needed in other files, I'm assuming the file is included via other headers because it has the capabilities checks. I didn't verify though.

include/openssl/crypto.h
```diff
@@ -86,33 +86,28 @@ OPENSSL_EXPORT int CRYPTO_needs_hwcap2_workaround(void);
 
 // Data-Independent Timing (DIT) on AArch64
 
-#if defined(OPENSSL_AARCH64) && !defined(OPENSSL_WINDOWS) && defined(MAKE_DIT_AVAILABLE)
+#if defined(OPENSSL_AARCH64) && !defined(OPENSSL_WINDOWS)
```
justsmth, line 89: * Instead of "disallowing" Windows, we should just explicitly "allow" Linux and Apple. There are other OS's, like [FreeBSD](https://www.freebsd.org/platforms/arm/), for which the `ARMV8_DIT_ALLOWED` macro can not be used properly.(?)
* Perhaps we need a meta-macro that we can use in the places that we guard the DIT functionality?
```
#define AARCH64_DIT_SUPPORTED (defined(OPENSSL_AARCH64) && (defined(OPENSSL_LINUX) || defined(OPENSSL_APPLE))
```

crypto/fipsmodule/cpucap/cpu_aarch64.c
```diff
@@ -49,16 +49,21 @@ void handle_cpu_env(uint32_t *out, const char *in) {
   }
 }
 
-#if defined(MAKE_DIT_AVAILABLE) && !defined(OPENSSL_WINDOWS)
+#if !defined(OPENSSL_WINDOWS)
```
justsmth, line 52: See other comment.

tool/speed.cc
```diff
@@ -85,8 +86,7 @@ static inline void *align_pointer(void *ptr, size_t alignment) {
 }
 #endif
 
-#if defined(OPENSSL_IS_AWSLC) && defined(OPENSSL_AARCH64) && !defined(OPENSSL_WINDOWS) && \
-    defined(MAKE_DIT_AVAILABLE)
+#if defined(OPENSSL_IS_AWSLC) && defined(OPENSSL_AARCH64) && !defined(OPENSSL_WINDOWS)
```
justsmth, line 89: See other comment.

crypto/fipsmodule/cpucap/internal.h
```diff
@@ -251,11 +251,51 @@ OPENSSL_INLINE int CRYPTO_is_ARMv8_wide_multiplier_capable(void) {
...
+  return (OPENSSL_armcap_P & (ARMV8_DIT | ARMV8_DIT_ALLOWED)) ==
+    (ARMV8_DIT | ARMV8_DIT_ALLOWED);
 }
 
+// This function is used only for testing; hence, not inlined
+OPENSSL_EXPORT int CRYPTO_is_ARMv8_DIT_capable_for_testing(void);
+
 #endif  // OPENSSL_ARM || OPENSSL_AARCH64
 
+#if defined(OPENSSL_AARCH64) && !defined(OPENSSL_WINDOWS)
```
justsmth, line 263: See other comment.

crypto/fipsmodule/cpucap/cpu_aarch64_dit_test.cc
```diff
@@ -0,0 +1,177 @@
...
+
+#include "internal.h"
+
+#if defined(OPENSSL_THREADS)
+#include <chrono>
+#include <thread>
+#endif
+
+#if defined(OPENSSL_AARCH64) && !defined(OPENSSL_STATIC_ARMCAP) && \
+    !defined(OPENSSL_WINDOWS)
```
justsmth, line 15: See other comment.

WillChilds-Klein (approved):

PR #1785, opened 2024-08-20
Replace ECDSA_METHOD with EC_KEY_METHOD and add the associated API

smittals2:
### Description of changes: 
This is the second pull request in a series of three ([prev PR](https://github.com/aws/aws-lc/pull/1776)). This PR primarily focuses on refactoring and enhancing the EC_KEY_METHOD API. The key changes are as follows:

1. **Removal of ECDSA_METHOD**: ECDSA_METHOD support, including the associated ENGINE functions, has been fully removed and replaced by EC_KEY_METHOD. This aligns with the transition seen in OpenSSL 1.1.1 and was also driven by customer requirements.
2. **Hiding Struct definition**: The EC_KEY_METHOD struct definition is now internal. Unlike OpenSSL, where static allocation was possible, AWS-LC now supports only dynamic allocation of this struct.
3. **Expansion of EC_KEY_METHOD Functions**: Added various EC_KEY_METHOD_* functions to create, free, get/set, and manipulate the EC_KEY_METHOD struct.

### Testing:
Tested with various fields initialized in EC_KEY_METHOD to ensure that custom functionality is correctly invoked.


crypto/fipsmodule/ec/ec_key.c
```diff
@@ -562,3 +567,123 @@ void *EC_KEY_get_ex_data(const EC_KEY *d, int idx) {
...
+  }
+
+  // Setting this field is currently not supported by AWS-LC
+  if(sign_setup) {
+    OPENSSL_PUT_ERROR(EC, ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
+    abort();
+  }
+
+  meth->sign = sign;
+  meth->sign_sig = sign_sig;
```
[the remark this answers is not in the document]
smittals2: Yep, the METHOD structs are zero initialized and meth pointers are "NULL" anyway. This is checked in all wrapping use cases before actually calling the custom code.

crypto/fipsmodule/ec/ec_key.c
```diff
@@ -562,3 +567,123 @@ void *EC_KEY_get_ex_data(const EC_KEY *d, int idx) {
...
+  if(meth == NULL) {
+    OPENSSL_PUT_ERROR(EC, ERR_R_PASSED_NULL_PARAMETER);
+    return;
+  }
+
+  // Setting this field is currently not supported by AWS-LC
+  if(sign_setup) {
+    OPENSSL_PUT_ERROR(EC, ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
+    abort();
+  }
```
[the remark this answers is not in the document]
smittals2: - Discussed offline. Static asserts would not work to determine the nullness of a runtime value. Putting an error on the queue and returning could lead to erroneous usage if the error queue is not checked. Aborting, while abrupt, is documented and will provide a clearer indication of unsupported functionality. 
- We define some _NOT_SUPPORTED_ ERR types but they are library specific (EVP, DH, [ENGINE](https://github.com/smittals2/aws-lc/blob/3c7ba297d6150cbe65b636c507e97abeeb65822f/include/openssl/engine.h#L114), etc.). We could define one for this use case if needed.

include/openssl/ec_key.h
```diff
@@ -356,6 +328,82 @@ OPENSSL_EXPORT EC_KEY *o2i_ECPublicKey(EC_KEY **out_key, const uint8_t **inp,
...
+#define ECDSA_FLAG_OPAQUE 1
+
+// EC_KEY_get_default_method returns a reference to the default
+// |EC_KEY| implementation. The returned |EC_KEY_METHOD| object is
+// zero-initialized. This is different from OpenSSL which returns function
+// pointers to the default implementations within the |EC_KEY_METHOD| struct.
+// We do not do this to make it easier for the compiler/linker to drop unused
+// functions. The wrapper functions will select the appropriate
+// |ec_key_default_*| implementation.
+OPENSSL_EXPORT const EC_KEY_METHOD *EC_KEY_get_default_method(void);
```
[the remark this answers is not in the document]
smittals2: Valid points - I'll update documentation :)

Some context on OpenBSD documentation:
This API has another function EC_KEY_set_default_method that allows consumers to change the default method that future EC_KEY's will be created with. This is why the documentation for EC_KEY_get_default_method in OpenBSD indicates it returns the method "used for creation of new EC_KEY objects in the future". In our case, since we are not supporting this functionality yet, the default method is always returned. We can always change the internal implementation details later should we choose to support setting the default method.

include/openssl/ec_key.h
```diff
@@ -356,6 +328,82 @@ OPENSSL_EXPORT EC_KEY *o2i_ECPublicKey(EC_KEY **out_key, const uint8_t **inp,
...
+OPENSSL_EXPORT EC_KEY_METHOD *EC_KEY_METHOD_new(const EC_KEY_METHOD *eckey_meth);
+
+// EC_KEY_METHOD_free frees the memory associated with |eckey_meth|
+OPENSSL_EXPORT void EC_KEY_METHOD_free(EC_KEY_METHOD *eckey_meth);
+
+// EC_KEY_set_method sets |meth| on |ec|. Currently, AWS-LC does not support
+// setting the compute_key, copy, keygen, set_group, set_private, set_public,
+// verify, verify_sig fields. These must be NULL or the function will fail.
+// Returns zero on failure and one on success.
+OPENSSL_EXPORT int EC_KEY_set_method(EC_KEY *ec, const EC_KEY_METHOD *meth);
```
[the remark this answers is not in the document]
smittals2: Since we don't currently provide setters for these operations, it wouldn't be possible to set those fields. I can change the documentation to be clearer about this - I think leaving the check in there should be okay just to make sure those fields are NULL until we support them.

crypto/fipsmodule/ec/internal.h
```diff
@@ -745,6 +745,48 @@ const EC_METHOD *EC_GFp_nistp521_method(void);
...
+    int (*set_group)(EC_KEY *key, const EC_GROUP *group);
+    int (*set_private)(EC_KEY *key, const BIGNUM *priv_key);
+    int (*set_public)(EC_KEY *key, const EC_POINT *pub_key);
+    int (*keygen)(EC_KEY *key);
+    int (*compute_key)(unsigned char **out, size_t *out_len,
+                       const EC_POINT *pub_key, const EC_KEY *ecdh);
+    int (*verify)(int type, const uint8_t *digest, int digest_len,
+                  const uint8_t *sig, unsigned int sig_len, EC_KEY *eckey);
+    int (*verify_sig)(const uint8_t *digest, int digest_len,
+                      const ECDSA_SIG *sig, EC_KEY *eckey);
```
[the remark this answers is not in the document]
smittals2: In case we ever make the structs public to match OpenSSL and allow static allocation of the structs, it would be easier to match OpenSSL. We don't necessarily need the fields that we don't provide setters for yet. I have removed compute_key, keygen, verify, and verify_sig for now and updated documentation to reflect the above point.

tests/ci/integration/openvpn_patch/aws-lc-openvpn2-6-x.patch
```diff
@@ -38,10 +38,22 @@ index c9fa7196..a48ef391 100644
...
-@@ -1658,8 +1658,10 @@ tls_ctx_load_ca(struct tls_root_ctx *ctx, const char *ca_file,
+@@ -1431,7 +1431,11 @@ tls_ctx_use_external_ec_key(struct tls_root_ctx *ctx, EVP_PKEY *pkey)
+ 
+     /* Among init methods, we only need the finish method */
+     EC_KEY_METHOD_set_init(ec_method, NULL, openvpn_extkey_ec_finish, NULL, NULL, NULL, NULL);
++#ifdef OPENSSL_IS_AWSLC
++    EC_KEY_METHOD_set_sign(ec_method, ecdsa_sign, NULL, ecdsa_sign_sig);
++#else
+     EC_KEY_METHOD_set_sign(ec_method, ecdsa_sign, ecdsa_sign_setup, ecdsa_sign_sig);
++#endif
```
justsmth, line 52: How does our lack of support for "sign_setup" not end up breaking OpenVPN's usage here?

crypto/fipsmodule/ec/ec_key.c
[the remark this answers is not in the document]
smittals2: Using the static asserts in a header file to prevent compilation with consumers that set specific fields to Non-NULL values worked. PR has been updated with relevant changes.

include/openssl/ec_key.h
```diff
@@ -356,6 +328,115 @@ OPENSSL_EXPORT EC_KEY *o2i_ECPublicKey(EC_KEY **out_key, const uint8_t **inp,
...
+                                        unsigned int *siglen,
+                                        const BIGNUM *k_inv,
+                                        const BIGNUM *r, EC_KEY *eckey),
+                                int (*sign_setup)(EC_KEY *eckey, BN_CTX *ctx_in,
+                                        BIGNUM **kinvp, BIGNUM **rp),
+                                ECDSA_SIG *(*sign_sig)(const uint8_t *digest,
+                                         int digest_len,
+                                         const BIGNUM *in_kinv,
+                                         const BIGNUM *in_r,
+                                         EC_KEY *eckey));
```
justsmth, line 401: * Since `EC_KEY_METHOD_set_sign_impl` is our own function, there's no need for it to accept the parameters (like `sign_setup`) that it doesn't support.
* Is `EC_KEY_METHOD_set_sign_impl` a function we want public?  Is the `_impl` suffix a pattern that we want to establish as a precedent if we face this type of issue in the future? Would a `_awslc` suffix be better?

include/openssl/ec_key.h
```diff
@@ -356,6 +328,115 @@ OPENSSL_EXPORT EC_KEY *o2i_ECPublicKey(EC_KEY **out_key, const uint8_t **inp,
...
+OPENSSL_EXPORT int EC_KEY_set_method(EC_KEY *ec, const EC_KEY_METHOD *meth);
+
+// EC_KEY_get_method returns the |EC_KEY_METHOD| object associated with |ec|.
+OPENSSL_EXPORT const EC_KEY_METHOD *EC_KEY_get_method(const EC_KEY *ec);
+
+#ifdef __cplusplus
+#define IS_NULL(ptr) ((ptr) == nullptr)
+#else
+#define IS_NULL(ptr) ((ptr) == NULL)
+#endif
```
justsmth, line 383: * This name `IS_NULL` seems likely to conflict with a macro defined by a consumer. What about `AWSLC_IS_NULL`?
* Why not just use `OPENSSL_STATIC_ASSERT` in the places it's needed below. Although discouraged, I believe the `NULL` macro is still available with C++.

tests/ci/integration/openvpn_patch/aws-lc-openvpn2-6-x.patch
```diff
@@ -38,10 +38,22 @@ index c9fa7196..a48ef391 100644
...
+index 23e76231..cab58ff4 100644
 --- a/src/openvpn/ssl_openssl.c
 +++ b/src/openvpn/ssl_openssl.c
-@@ -1658,8 +1658,10 @@ tls_ctx_load_ca(struct tls_root_ctx *ctx, const char *ca_file,
+@@ -1431,7 +1431,11 @@ tls_ctx_use_external_ec_key(struct tls_root_ctx *ctx, EVP_PKEY *pkey)
+ 
+     /* Among init methods, we only need the finish method */
+     EC_KEY_METHOD_set_init(ec_method, NULL, openvpn_extkey_ec_finish, NULL, NULL, NULL, NULL);
++#ifdef OPENSSL_IS_AWSLC
++    EC_KEY_METHOD_set_sign(ec_method, ecdsa_sign, NULL, ecdsa_sign_sig);
```
justsmth, line 49: I would go ahead and have this directly call `EC_KEY_METHOD_set_sign_impl` (or whatever we end up calling it).

tests/ci/integration/openvpn_patch/aws-lc-openvpn2-6-x.patch
smittals2: Their sign_setup function, ecdsa_sign_setup, does nothing (returns 1). We want to upstream this change so they stop setting that field altogether. In OpenSSL, the only way this custom sign_setup functionality is invoked is 
1) Through ECDSA_sign_setup (which we don't support)
2) Through normal signing workflows (like EVP_PKEY_sign or ECDSA_sign) but only when custom functionality is provided only for sign_setup (and not for sign, or sign_sig). This is not OpenVPNs use case.

tests/ci/integration/openvpn_patch/aws-lc-openvpn2-6-x.patch
smittals2: We want to upstream this change so it's likely best to use the OpenSSL API. If we are unable to do so/for internal use cases, happy to make that change.

include/openssl/ec_key.h
smittals2: Removed the parameters and changed the suffix to _awslc. I don't think it matters if our variant is public or not, it should be okay for consumers to use either function.

include/openssl/ec_key.h
smittals2: Got it, changed to use OPENSSL_STATIC_ASSERT to support C99. Got rid of the IS_NULL macro, NULL does exist in C++ and should be okay for consumers to use.

crypto/fipsmodule/ec/internal.h
```diff
@@ -745,6 +745,46 @@ const EC_METHOD *EC_GFp_nistp521_method(void);
...
+    // not set by default.
+    int flags;
+
+    // AWS-LC currently does not support these fields, must be NULL
+    int (*sign_setup)(EC_KEY *eckey, BN_CTX *ctx_in, BIGNUM **k_inv,
+                      BIGNUM **r);
+    int (*copy)(EC_KEY *dest, const EC_KEY *src);
+    int (*set_group)(EC_KEY *key, const EC_GROUP *group);
+    int (*set_private)(EC_KEY *key, const BIGNUM *priv_key);
+    int (*set_public)(EC_KEY *key, const EC_POINT *pub_key);
```
justsmth, line 785: This is an opaque struct, so I don't see a reason to keep these field here for OpenSSL compatibility. If you don't think they should be deleted, could we comment them out?

smittals2: Got it, I just think it's good to have them since we offer setters for this functionality (although we assert people can't set these fields). I think commenting it out is a good compromise.

justsmth (dismissed):

include/openssl/ec_key.h
```diff
@@ -356,6 +328,98 @@ OPENSSL_EXPORT EC_KEY *o2i_ECPublicKey(EC_KEY **out_key, const uint8_t **inp,
...
+
+// EC_KEY_OpenSSL returns a reference to the default |EC_KEY| implementation.
+// The returned |EC_KEY_METHOD| object is statically allocated. The application
+// should not free this struct.
+//
+// This struct is also zero-initialized. This is different from OpenSSL which
+// returns function pointers to the default implementations within the
+// |EC_KEY_METHOD| struct. We do not do this to make it easier for the
+// compiler/linker to drop unused functions. The wrapper functions will select
+// the appropriate |ec_key_default_*| implementation.
```
nebeid, line 354: - I'm not sure why we decided not to initialise with our default ECDSA implementation if that was possible.
- Where is an example of the wrapper functions selecting the appropriate |ec_key_default_*| implementation?

crypto/fipsmodule/ecdsa/ecdsa.c
nebeid, line 1: I will review the changes in this file in my next review. I want to understand whether it interferes with the FIPS default implementation.

include/openssl/ec_key.h
```diff
@@ -356,6 +328,98 @@ OPENSSL_EXPORT EC_KEY *o2i_ECPublicKey(EC_KEY **out_key, const uint8_t **inp,
...
+
+// EC_KEY_METHOD_set_init_awslc sets the |init| and |finish| pointers on |meth|.
+OPENSSL_EXPORT void EC_KEY_METHOD_set_init_awslc(EC_KEY_METHOD *meth,
+                                            int (*init)(EC_KEY *key),
+                                            void (*finish)(EC_KEY *key));
+
+
+// EC_KEY_METHOD_set_init sets function pointers on |meth|. AWS-LC
+// currently only supports setting the |init| and |finish| fields. |copy|,
+// |set_group|, |set_private|, and |set_public| cannot be set yet and must
```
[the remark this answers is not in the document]
smittals2: I agree it's weird to have these in the EC_KEY_METHOD struct. The language is written as such to leave it "open". While we don't have plans to support that functionality, we may need to based on consumer demand in the future. The comment indicates that and leaves room for that change in the future should we need it.

crypto/fipsmodule/ec/internal.h
```diff
@@ -766,7 +811,7 @@ struct ec_key_st {
 
   CRYPTO_refcount_t references;
 
-  const ECDSA_METHOD *ecdsa_meth;
+  const EC_KEY_METHOD *eckey_method;
```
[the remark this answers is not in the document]
smittals2: Name of the field shouldn't matter, ec_key_st is an opaque struct (defined in an internal header) and consumers cannot directly access this field without using the appropriate EC_KEY_METHOD funcs.

include/openssl/ec_key.h
```diff
@@ -356,6 +328,98 @@ OPENSSL_EXPORT EC_KEY *o2i_ECPublicKey(EC_KEY **out_key, const uint8_t **inp,
 OPENSSL_EXPORT int i2o_ECPublicKey(const EC_KEY *key, unsigned char **outp);
 
 
+// EC_KEY_METHOD
+
+// ECDSA_FLAG_OPAQUE specifies that this EC_KEY_METHOD does not expose its key
+// material. This may be set if, for instance, it is wrapping some other crypto
+// API, like a platform key store. Use |EC_KEY_METHOD_set_flag| to set
+// this flag on an |EC_KEY_METHOD|.
+// This was supported in ECDSA_METHOD previously and is not set by default.
```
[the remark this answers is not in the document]
smittals2: Updated comments

crypto/fipsmodule/ec/internal.h
```diff
@@ -745,6 +745,51 @@ const EC_METHOD *EC_GFp_nistp521_method(void);
 // x86-64 optimized P256. See http://eprint.iacr.org/2013/816.
 const EC_METHOD *EC_GFp_nistz256_method(void);
 
+
+// EC_KEY METHOD
```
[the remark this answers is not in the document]
smittals2: not intentional, made the change :)

crypto/fipsmodule/ec/internal.h
```diff
@@ -745,6 +745,51 @@ const EC_METHOD *EC_GFp_nistp521_method(void);
...
+    // AWS-LC doesn't support custom values for EC_KEY operations
+    // as of now. |k_inv| and |r| must be NULL parameters.
+    // The |type| parameter is ignored in OpenSSL, we pass in zero for it
+    int (*sign)(int type, const uint8_t *digest, int digest_len,
+                uint8_t *sig, unsigned int *siglen, const BIGNUM *k_inv,
+                const BIGNUM *r, EC_KEY *eckey);
+
+    // AWS-LC doesn't support custom values for EC_KEY operations
+    // as of now. |k_inv| and |r| must be NULL parameters.
+    ECDSA_SIG *(*sign_sig)(const uint8_t *digest, int digest_len,
```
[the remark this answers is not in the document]
smittals2: Yup that's exactly it. sign writes it to a param that's passed in while sign_sig returns the signature in a ECDSA_SIG struct. I've updates the documentation for the two pointers to indicate where they are called.

include/openssl/ec_key.h
smittals2: This was a previous change in our library but it doesn't impact the usage or behavior of this struct/API. I have updated the documentation to be clearer about what it means for a default operation to be called. 

In the following link, the wrapper function ECDSA_sign checks for the existence of a custom implementation. In the default case, the ec_key_meth->sign pointer is NULL and so the code continues to execute the rest of the sign functionality (the default implementation). Otherwise, the custom functionality is invoked. ([link](https://github.com/aws/aws-lc/blob/70c6dfd09da529810cc9efebf32dbd9c25c3868a/crypto/fipsmodule/ecdsa/ecdsa.c#L379))

crypto/fipsmodule/ecdsa/ecdsa.c
nebeid: I took another look at the file and your explanation here https://github.com/aws/aws-lc/pull/1785#discussion_r1737469547 helped, @smittals2:
Previously we had `ECDSA_sign` support a custom implementation via 
https://github.com/aws/aws-lc/blob/0f782c257169c76d32b845d580116bd565258bff/crypto/fipsmodule/ecdsa/ecdsa.c#L379-L382

which by default was null and it continued to call `ECDSA_do_sign`.

Now we have also `ECDSA_do_sign` support a custom implementation via `eckey->eckey_method->sign_sig` which will default to null and the flow continues with the original implementation that is FIPS validated.

I'm convinced now that it's alright.

nebeid (approved):
Thank you for updating the documentation. It is clearer to me now.

justsmth (approved):

PR #1789, opened 2024-08-21
Use _Static_assert in refcount_c11.c to support old compilers that don't support the macro static_assert

andrewhop:
### Description of changes: 
Some other old compilers don't support the convince macro `static_assert` but do support the C11 keyword `_Static_assert` even though both should be in [assert.h](https://en.cppreference.com/w/c/error/static_assert). 

### Testing:
Built all old platforms internally. 

### DIFF
--- a/crypto/refcount_c11.c
+++ b/crypto/refcount_c11.c
@@ -26,12 +26,12 @@
 
 
 // See comment above the typedef of CRYPTO_refcount_t about these tests.
-static_assert(alignof(CRYPTO_refcount_t) == alignof(_Atomic CRYPTO_refcount_t),
+_Static_assert(alignof(CRYPTO_refcount_t) == alignof(_Atomic CRYPTO_refcount_t),
               "_Atomic alters the needed alignment of a reference count");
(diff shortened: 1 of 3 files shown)


crypto/refcount_c11.c
```diff
@@ -26,12 +26,12 @@
 
 
 // See comment above the typedef of CRYPTO_refcount_t about these tests.
-static_assert(alignof(CRYPTO_refcount_t) == alignof(_Atomic CRYPTO_refcount_t),
+_Static_assert(alignof(CRYPTO_refcount_t) == alignof(_Atomic CRYPTO_refcount_t),
```
justsmth, line 29: We use [OPENSSL_STATIC_ASSERT](https://github.com/aws/aws-lc/blob/7d3101fb08892c464bdf1ca3a4d4cbea5347820a/crypto/refcount_win.c#L23) other places for this same purpose. It's defined here: https://github.com/aws/aws-lc/blob/7d3101fb08892c464bdf1ca3a4d4cbea5347820a/include/openssl/type_check.h#L67-L98

justsmth (approved):

smittals2 (approved):

lilpoozie2005 (approved):

PR #1790, opened 2024-08-22
Refactor RSA_METHOD and expand API

smittals2:
### Issues
`CryptoAlg-2482
`
### Description of changes: 
This PR is the last in a series of three (prev PRs: [1st](https://github.com/aws/aws-lc/pull/1776), [2nd](https://github.com/aws/aws-lc/pull/1785)) to refactor and expand ENGINE, EC_KEY_METHOD and RSA_METHOD. This PR: 
1. Expands the RSA_METHOD struct and changes the function signatures for encrypt, decrypt, sign_raw, and verify_raw to match OpenSSL 1.1.1.
2. Makes the struct definition for RSA_METHOD internal. This means static allocation of this struct is no longer possible. 
3. Adds various functions to get/set, create, free, and manipulate RSA_METHOD

### Callout: 
This PR includes the commits for the other 2 PRs. They should be reviewed and merged before this one. 

### Testing:
Tested with various fields initialized in RSA_METHOD to ensure that custom functionality is correctly invoked.


include/openssl/rsa.h
```diff
@@ -215,6 +215,73 @@ OPENSSL_EXPORT int RSA_set0_factors(RSA *rsa, BIGNUM *p, BIGNUM *q);
...
+                         int (*pub_dec) (int max_out, const uint8_t *from,
+                                         uint8_t *to, RSA *rsa,
+                                         int padding));
+
+// RSA_meth_set_pub_enc sets |pub_enc| on |meth|. The |pub_enc| function
+// should return the number of bytes written to the object |to| or -1 for error.
+OPENSSL_EXPORT int RSA_meth_set_pub_enc(RSA_METHOD *meth,
+                         int (*pub_enc) (int max_out, const uint8_t *from,
+                                         uint8_t *to, RSA *rsa,
+                                         int padding));
```
justsmth, line 277: For the "RSA_meth_set_*" functions we should document the return value of the function as well as the return value (and parameters) of its function parameter.

tests/ci/integration/run_openssh_integration.sh
```diff
@@ -52,7 +52,7 @@ function openssh_build() {
   # The RSA_meth_XXX functions are not implemented by AWS-LC, and the implementation provided by OpenSSH also doesn't compile for us.
   # Fortunately, these functions are only needed for pkcs11 support, which is disabled for our build.
   # See: https://github.com/openssh/openssh-portable/pull/385
-  export CFLAGS="-DAWS_LC_INTERNAL_IGNORE_BN_SET_FLAGS=1 -DHAVE_RSA_METH_FREE=1 -DHAVE_RSA_METH_DUP=1 -DHAVE_RSA_METH_SET1_NAME=1 -DHAVE_RSA_METH_SET_PRIV_ENC=1 -DHAVE_RSA_METH_SET_PRIV_DEC=1"
+  export CFLAGS="-DAWS_LC_INTERNAL_IGNORE_BN_SET_FLAGS=1 -DHAVE_RSA_METH_FREE=1 -DHAVE_RSA_METH_DUP=1 -DHAVE_RSA_METH_SET1_NAME=1 -DHAVE_RSA_METH_GET_FINISH=1 -DHAVE_RSA_METH_SET_PRIV_ENC=1 -DHAVE_RSA_METH_SET_PRIV_DEC=1 -DHAVE_RSA_METH_SET_FINISH=1 "
   ./configure --with-ssl-dir="${AWS_LC_INSTALL_FOLDER}" --prefix="${OPENSSH_INSTALL_FOLDER}" --disable-pkcs11
```
justsmth, line 56: How close to full support of OpenSSH can we get with this PR?  Could we possibly enable pkcs11 support?

tests/ci/integration/openvpn_patch/aws-lc-openvpn2-6-x.patch
```diff
@@ -38,10 +38,22 @@ index c9fa7196..a48ef391 100644
...
-@@ -1658,8 +1658,10 @@ tls_ctx_load_ca(struct tls_root_ctx *ctx, const char *ca_file,
+@@ -1431,7 +1431,11 @@ tls_ctx_use_external_ec_key(struct tls_root_ctx *ctx, EVP_PKEY *pkey)
+ 
+     /* Among init methods, we only need the finish method */
+     EC_KEY_METHOD_set_init(ec_method, NULL, openvpn_extkey_ec_finish, NULL, NULL, NULL, NULL);
++#ifdef OPENSSL_IS_AWSLC
++    EC_KEY_METHOD_set_sign(ec_method, ecdsa_sign, NULL, ecdsa_sign_sig);
++#else
+     EC_KEY_METHOD_set_sign(ec_method, ecdsa_sign, ecdsa_sign_setup, ecdsa_sign_sig);
++#endif
```
justsmth, line 52: Instead of using the `OPENSSL_IS_AWSLC` macro, could we have it always pass NULL for that parameter?  Would that cause a failure when building against OpenSSL?
```
EC_KEY_METHOD_set_sign(ec_method, ecdsa_sign, NULL, ecdsa_sign_sig);
```

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -448,12 +448,144 @@ int RSA_set0_crt_params(RSA *rsa, BIGNUM *dmp1, BIGNUM *dmq1, BIGNUM *iqmp) {
...
+
+  meth->app_data = app_data;
+  return 1;
+}
+
 static int rsa_sign_raw_no_self_test(RSA *rsa, size_t *out_len, uint8_t *out,
                                      size_t max_out, const uint8_t *in,
                                      size_t in_len, int padding) {
   SET_DIT_AUTO_DISABLE;
   if (rsa->meth->sign_raw) {
```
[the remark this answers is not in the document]
smittals2: rsa->meth will never be NULL. It will either point to our default zero-initialized struct or to a custom one provided by the consumer. RSA_set_method does not allow for a NULL method to bet set and RSA_new always initializes new keys with a reference to the default struct. I'll add the check anyway just so we're more protected from future changes :)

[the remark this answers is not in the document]
smittals2: We have to include these checks in the internal functions. For example, rsa_sign_raw_no_self_test (our internal function) is called by the public RSA_sign_raw function. But that internal function is also called in the RSA_sign workflow (since sign wraps the lower level sign_raw similar to how in EC_KEY sign wrapped sign_sig).  If custom sign functionality is provided, RSA_sign will directly call that custom functionality. But in the case that custom sign functionality is not provided but custom sign_raw is, the RSA_sign workflow should call the lower level custom sign_raw which is only possible when this check is in the internal function.

tests/ci/integration/openvpn_patch/aws-lc-openvpn2-6-x.patch
smittals2: Eventually when we upstream, the goal is to get OpenVPN code to have the "LC variant" with NULL passed in. This shouldn't cause an error in OpenSSL either. It's written this way just for the patch

crypto/fipsmodule/rsa/internal.h
```diff
@@ -69,6 +69,49 @@ extern "C" {
...
+                    int padding);
+    // Set via |RSA_meth_set_pub_dec|
+    int (*verify_raw)(int max_out, const uint8_t *in, uint8_t *out, RSA *rsa,
+                      int padding);
+    // Set via |RSA_meth_set_priv_dec|
+    int (*decrypt)(int max_out, const uint8_t *in, uint8_t *out, RSA *rsa,
+                   int padding);
+    // Set via |RSA_meth_set_pub_enc|
+    int (*encrypt)(int max_out, const uint8_t *in, uint8_t *out, RSA *rsa,
+                   int padding);
```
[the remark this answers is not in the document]
smittals2: The names are a little misleading. Private key encryption and public key decryption aren't really "normal operations" for RSA keys, sign_raw and verify_raw are clearer in indicating what the operation is doing. Since the function pointers are part of an internal struct, the names don't really matter - I picked ones that are clearer to understand. We can always change this if we decide to make the struct public and want to maintain OpenSSL compatibility - I've documented this in a clearer way.

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -881,6 +1013,11 @@ int RSA_flags(const RSA *rsa) {
   return rsa->flags;
 }
 
+void RSA_set_flags(RSA *rsa, int flags) {
+  SET_DIT_AUTO_DISABLE;
+  rsa->flags |= flags;
+}
```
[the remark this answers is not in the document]
smittals2: It looks like similar functions like RSA_flags or RSA_test_flags have been written under the assumption that the RSA object will not be NULL. I think we should just error and return, this doesn't seem like something a program should crash over.

include/openssl/rsa.h
smittals2: The return values for the setters are documented at the top of this section (since it's the same for each function). I have updated documentation with return values for the passed in func pointers.

tests/ci/integration/run_openssh_integration.sh
smittals2: I haven't done a full deep dive, but OpenSSH still needs some other functions part of the RSA_METHOD API like RSA_meth_get_finish or RSA_meth_dup. To support PKCS11, some more work needs to be done.

justsmth: I'm curious how much additional work this would actually be.  The two functions you mention (`RSA_meth_get_finish` and `RSA_meth_dup`) seem rather trivial to implement.

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -470,7 +602,8 @@ int RSA_sign_raw(RSA *rsa, size_t *out_len, uint8_t *out, size_t max_out,
 
 unsigned RSA_size(const RSA *rsa) {
   SET_DIT_AUTO_DISABLE;
-  size_t ret = rsa->meth->size ? rsa->meth->size(rsa) : rsa_default_size(rsa);
+  size_t ret = (rsa->meth && rsa->meth->size) ?
```
justsmth, line 605: I suppose we should allow the SEGFAULT if `rsa` is NULL?

crypto/engine/engine.c
```diff
@@ -50,7 +50,10 @@ int ENGINE_set_RSA(ENGINE *engine, const RSA_METHOD *method) {
 }
 
 const RSA_METHOD *ENGINE_get_RSA(const ENGINE *engine) {
```
justsmth, line 52: This file doesn't have any test coverage.

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -448,12 +448,144 @@ int RSA_set0_crt_params(RSA *rsa, BIGNUM *dmp1, BIGNUM *dmq1, BIGNUM *iqmp) {
   return 1;
 }
 
+RSA_METHOD *RSA_meth_new(const char *name, int flags) {
+  RSA_METHOD *meth = OPENSSL_zalloc(sizeof(*meth));
+
+  if (meth != NULL) {
```
nebeid, line 454: I suggest checking for null and putting the error first in the code flow as we usually see in other functions (including the one right after).

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -878,12 +1011,32 @@ int rsa_private_transform(RSA *rsa, uint8_t *out, const uint8_t *in,
 
 int RSA_flags(const RSA *rsa) {
   SET_DIT_AUTO_DISABLE;
-  return rsa->flags;
+  if (rsa) {
```
nebeid, line 1014: I suggest toggling the check logic as my comment above on `RSA_meth_new`.

include/openssl/rsa.h
```diff
@@ -215,6 +215,88 @@ OPENSSL_EXPORT int RSA_set0_factors(RSA *rsa, BIGNUM *p, BIGNUM *q);
...
+OPENSSL_EXPORT int RSA_set_method(RSA *rsa, const RSA_METHOD *meth);
+
+// RSA_get_method returns the |RSA_METHOD| object associated with |rsa|.
+OPENSSL_EXPORT const RSA_METHOD *RSA_get_method(const RSA *rsa);
+
+// RSA_meth_free frees the memory associated with |meth|
+OPENSSL_EXPORT void RSA_meth_free(RSA_METHOD *meth);
+
+// RSA_METHOD setters
+// The following functions set the corresponding fields on |meth|. Returns one
```
nebeid, line 240: Nit:
```suggestion
// The following functions set the corresponding fields on |meth|. They return one
```

nebeid (commented):
Were the patch files removed?

crypto/fipsmodule/rsa/rsa.c
smittals2: Ahhh sorry I guess I can expand the check. A consumer shouldn't really be calling this on a NULL key but I guess the less jagged edges the better.

include/openssl/rsa.h
```diff
@@ -215,6 +215,88 @@ OPENSSL_EXPORT int RSA_set0_factors(RSA *rsa, BIGNUM *p, BIGNUM *q);
...
+// encrypted data or a negative number on error.
+OPENSSL_EXPORT int RSA_meth_set_pub_enc(RSA_METHOD *meth,
+                         int (*pub_enc) (int max_out, const uint8_t *from,
+                                         uint8_t *to, RSA *rsa,
+                                         int padding));
+
+// RSA_meth_set0_app_data sets |app_data| on |meth|. Although set0 functions
+// generally take ownership in AWS-LC, to maintain OpenSSL compatibility,
+// this function does not. It is the consumers responsibility to free
+// |app_data| before |meth| is freed.
```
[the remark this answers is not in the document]
smittals2: I thought since app_data lives on meth, it should be freed before the overarching object with the reference is freed. But I guess it wouldn't matter as long as the consumer continues to hold a reference to the app_data they set. I'll change the documentation to generally say it's consumer responsibility. I guess, it doesn't necessarily have to happen before.

include/openssl/engine.h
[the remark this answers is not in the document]
smittals2: Good call out, updated the documentation :)

include/openssl/rsa.h
```diff
@@ -215,6 +215,88 @@ OPENSSL_EXPORT int RSA_set0_factors(RSA *rsa, BIGNUM *p, BIGNUM *q);
 OPENSSL_EXPORT int RSA_set0_crt_params(RSA *rsa, BIGNUM *dmp1, BIGNUM *dmq1,
                                        BIGNUM *iqmp);
 
+// RSA_METHOD functions
```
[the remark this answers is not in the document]
smittals2: RSA_METHOD wasn't deprecated before so I am leaving it as so. Since EC_KEY_METHOD was a completely new feature, we made the conscious decision to put in a deprecated section.

include/openssl/rsa.h
```diff
@@ -215,6 +215,88 @@ OPENSSL_EXPORT int RSA_set0_factors(RSA *rsa, BIGNUM *p, BIGNUM *q);
...
+// is called in |RSA_free| before freeing the key. |finish| should return
+// one on success and zero on failure.
+OPENSSL_EXPORT int RSA_meth_set_finish(RSA_METHOD *meth,
+                                       int (*finish) (RSA *rsa));
+
+// RSA_meth_set_priv_dec sets |priv_dec| on |meth|. The |priv_dec| function
+// should return the number of bytes written to the object |to| or -1 for error.
+// |priv_dec| should decrypt |max_out| bytes at |from| using the private key
+// |rsa| and store the plaintext in |to|. |priv_dec| should return the size of
+// the recovered plaintext or a negative number on error.
```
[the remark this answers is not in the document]
smittals2: No we will not need to designate another negative number. Generally OpenSSL uses -1 for error. Since this if for the custom implementations, we can be more generic and say a negative number indicates error/. Our checks align with this (we check if return value is less than 0). This would encompass the -1 case as well.

include/openssl/rsa.h
```diff
@@ -215,6 +215,88 @@ OPENSSL_EXPORT int RSA_set0_factors(RSA *rsa, BIGNUM *p, BIGNUM *q);
...
+// RSA_METHOD functions
+
+// RSA_get_default_method returns a zero initialized |RSA_METHOD| object. The
+// wrapper functions will select the appropriate |rsa_default_*| implementation.
+OPENSSL_EXPORT const RSA_METHOD *RSA_get_default_method(void);
+
+// RSA_meth_new returns a zero-initialized |RSA_METHOD| object. It sets
+// |flags| on the object. The |name| parameter is currently ignored and
+// part of the function signature for OpenSSL compatibility.
+OPENSSL_EXPORT RSA_METHOD *RSA_meth_new(const char *name, int flags);
```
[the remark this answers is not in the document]
smittals2: Yea we don't really need this functionality around flags yet so I am not implementing the associated functions either. Currently, only |RSA_FLAG_OPAQUE| is supported since it was offered before,  but there isn't necessarily a way to use it unless it's created with that flag in RSA_meth_new. Since this a previously offered flag, the necessary caveats/alternate codepaths already exist to accomodate this flag. we can always add this in later should we need it.

smittals2:
> Were the patch files removed?
No we still need some patches to work with OpenVPN

crypto/fipsmodule/rsa/internal.h
```diff
@@ -69,6 +69,53 @@ extern "C" {
...
+struct rsa_meth_st {
+    void *app_data;
+
+    int (*init)(RSA *rsa);
+    int (*finish)(RSA *rsa);
+
+    // size returns the size of the RSA modulus in bytes.
+    size_t (*size)(const RSA *rsa);
+
+    int (*sign)(int type, const uint8_t *m, unsigned int m_length,
```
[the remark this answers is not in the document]
smittals2: Ahh this was an old field we had. We don't need it but since it's already plumbed through I'll add a setter for it and update the documentation.

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -878,12 +1026,32 @@ int rsa_private_transform(RSA *rsa, uint8_t *out, const uint8_t *in,
...
   SET_DIT_AUTO_DISABLE;
+  if (rsa == NULL) {
+    OPENSSL_PUT_ERROR(RSA, ERR_R_PASSED_NULL_PARAMETER);
+    return 0;
+  }
+
   return rsa->flags;
 }
 
+void RSA_set_flags(RSA *rsa, int flags) {
```
justsmth, line 1037: NP: Do we also need `RSA_clear_flags`?

justsmth (approved):

include/openssl/rsa.h
```diff
@@ -294,10 +295,20 @@ OPENSSL_EXPORT int RSA_meth_set_pub_enc(RSA_METHOD *meth,
 // RSA_meth_set0_app_data sets |app_data| on |meth|. Although set0 functions
 // generally take ownership in AWS-LC, to maintain OpenSSL compatibility,
 // this function does not. It is the consumers responsibility to free
-// |app_data| before |meth| is freed.
+// |app_data|.
 OPENSSL_EXPORT int RSA_meth_set0_app_data(RSA_METHOD *meth, void *app_data);
 
 
+// RSA_meth_set_sign sets |sign| on |meth|. The function |sign| should return
```
nebeid, line 302: Do we need more documentation for |sign|? Sould we say that type is a hash_nid and it should create a PKCS1 prefix as in the usage justsmth pointed to https://github.com/aws/aws-lc/blame/612c200ca1ff0f4068fcd47d4869035d16e08bb0/crypto/fipsmodule/rsa/rsa.c#L818? I didn't dive into the implementation to provide an accurate documentation, tbh.

include/openssl/rsa.h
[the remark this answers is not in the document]
nebeid: Do we want to shorten the documentation and avoid duplication here and in the next 3 as suggested above?

crypto/fipsmodule/rsa/internal.h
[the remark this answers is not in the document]
smittals2: Yep, that is currently the only place the custom sign functionality is invoked

crypto/fipsmodule/rsa/rsa.c
```diff
@@ -878,12 +1011,32 @@ int rsa_private_transform(RSA *rsa, uint8_t *out, const uint8_t *in,
...
-  return rsa->flags;
+  if (rsa) {
+    return rsa->flags;
+  }
+
+  OPENSSL_PUT_ERROR(RSA, ERR_R_PASSED_NULL_PARAMETER);
+  return 0;
+}
+
+void RSA_set_flags(RSA *rsa, int flags) {
```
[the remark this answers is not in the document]
smittals2: This isn't impacted by the RSA_METHOD work, just updated the error handling for NULL rsa keys. I can add in testing here if needed, but that would be a separate chunk of work (to understand all the RSA key flags, how they impact behavior, and then test around that).

crypto/fipsmodule/rsa/rsa.c
smittals2: RSA related functionality, don't need as of now

include/openssl/rsa.h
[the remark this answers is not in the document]
smittals2: Only the second sentence is duplicated across the comments. There are slight differences otherwise in what each func should be doing. I think it's okay to leave it like this, might make it more confusing to have two separate documentation sections for these funcs.

tests/ci/integration/run_openssh_integration.sh
smittals2: messaged offline :)

nebeid (approved):

justsmth (approved):

PR #1793, opened 2024-08-22
Add KDA OneStep (SSKDF_digest and SSKDF_hmac) to FIPS indicator

skmcgrail:
### Description of changes: 
Adds the FIPS service indicator APIs to `SSKDF_digest` and `SSKDF_hmac`.

**NOTE**: AWS-LC doesn't currently support using SHA3-* digest algorithms with our HMAC APIs. So that is why it is missing from the approved digest list for `SSKDF_hmac`.

### DIFF
--- a/crypto/fipsmodule/service_indicator/service_indicator.c
+++ b/crypto/fipsmodule/service_indicator/service_indicator.c
@@ -494,7 +494,6 @@ void TLSKDF_verify_service_indicator(const EVP_MD *dgst, const char *label,
 //
 // Sourced from NIST.SP.800-56Cr2 Section 7: Selecting Hash Functions and MAC Algorithms
 // https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Cr2.pdf
-
 void SSKDF_digest_verify_service_indicator(const EVP_MD *dgst) {
   switch (dgst->type) {
     case NID_sha1:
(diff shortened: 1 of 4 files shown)


nebeid (dismissed):

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -487,6 +487,61 @@ void TLSKDF_verify_service_indicator(const EVP_MD *dgst, const char *label,
...
+      break;
+  }
+}
+
+// "Whenever a hash function is employed (including as the primitive used by HMAC), an
+// approved hash function shall be used. FIPS 180 and FIPS 202 specify approved hash
+// functions"
+//
+// * FIPS 180 covers the SHA-1 and SHA-2* family of algorithms
+// * FIPS 202 covers the SHA3-* family of algorithms (Note: AWS-LC does not currently support SHA-3 with HMAC)
```
WillChilds-Klein, line 525: do we think we will support HMAC-SHA3 at some point? if so, is there any cost to adding relevant NIDs below now so we don't forget?

WillChilds-Klein (dismissed):

crypto/fipsmodule/service_indicator/service_indicator.c
skmcgrail: I'm going to hold off. Even if we add it in the future we won't have it as an approved algorithm in the next round of certification if we don't test it with the lab. So it would be incorrect for the indicator to be set if that was the case and could cause issues.

WillChilds-Klein (approved):

nebeid (approved):

PR #1800, opened 2024-08-23
Update x509 tool to write all output to common BIO which is a file or stdout

andrewhop:
### Description of changes: 
Previously only some output followed the user's request using `-out` to direct the output to a file. This change updates everything to use a common `BIO` which is `stdout` by default or the user provided file.

### Testing:
Added new tests that test writing to a file directly with `-out` we have existing tests that cover writing to `stdout` (and then getting redirected to a file with `>`) works as expected.

### DIFF
--- a/tool-openssl/x509.cc
+++ b/tool-openssl/x509.cc
@@ -272,7 +272,7 @@ bool X509Tool(const args_list_t &args) {
       }
     }
 
-    if (!noout && !checkend && !modulus) {
+    if (!noout && !checkend) {
       if (!WriteSignedCertificate(x509.get(), output_bio, out_path)) {
         return false;
       }
--- a/tool-openssl/x509.cc
+++ b/tool-openssl/x509.cc
@@ -72,7 +72,7 @@ bool X509Tool(const args_list_t &args) {
   GetString(&in_path, "-in", "", parsed_args);
   GetBoolArgument(&req, "-req", parsed_args);
   GetString(&signkey_path, "-signkey", "", parsed_args);
-  GetString(&out_path, "-out", "", parsed_args);
+  GetString(&out_path, "-out", "stdout", parsed_args);
   GetBoolArgument(&noout, "-noout", parsed_args);
   GetBoolArgument(&dates, "-dates", parsed_args);
   GetBoolArgument(&modulus, "-modulus", parsed_args);
--- a/tool-openssl/x509.cc
+++ b/tool-openssl/x509.cc
@@ -72,7 +72,7 @@ bool X509Tool(const args_list_t &args) {
   GetString(&in_path, "-in", "", parsed_args);
   GetBoolArgument(&req, "-req", parsed_args);
   GetString(&signkey_path, "-signkey", "", parsed_args);
-  GetString(&out_path, "-out", "stdout", parsed_args);
+  GetString(&out_path, "-out", "", parsed_args);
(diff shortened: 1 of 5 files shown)


tool-openssl/x509.cc
```diff
@@ -279,15 +272,8 @@ bool X509Tool(const args_list_t &args) {
...
-    if (!out_path.empty()) {
-      if (!WriteSignedCertificate(x509.get(), out_path)) {
-        return false;
-      }
-    }
-
-    if (!noout && !in_path.empty() && !checkend && parsed_args.count("-out")==0) {
-      bssl::UniquePtr<BIO> bio_out(BIO_new_fp(stdout, BIO_NOCLOSE));
-      if (!PEM_write_bio_X509(bio_out.get(), x509.get())) {
+    if (!noout && !checkend && !modulus) {
```
WillChilds-Klein, line 275: we didn't check `modulus` previously here. why do it now?

andrewhop: I was trying to fix a failing test, turns out OpenSSL 1.0.2 and 1.1.1 have wildly different behavior of what gets written to the -out parameter vs what is always written to stdout. I think testing against 1.0.2 has very limited benefit and is hurting our testing against newer versions.

WillChilds-Klein (dismissed):

tool-openssl/x509.cc
```diff
@@ -22,13 +22,8 @@ static const argument_t kArguments[] = {
...
-bool WriteSignedCertificate(X509 *x509, const std::string &out_path) {
-  ScopedFILE out_file(fopen(out_path.c_str(), "wb"));
-  if (!out_file) {
-    fprintf(stderr, "Error: unable to open output file '%s'\n", out_path.c_str());
-    return false;
-  }
-  if (!PEM_write_X509(out_file.get(), x509)) {
+static bool WriteSignedCertificate(X509 *x509, bssl::UniquePtr<BIO> &output_bio, const std::string &out_path) {
+  if (!PEM_write_bio_X509(output_bio.get(), x509)) {
     fprintf(stderr, "Error: error writing certificate to '%s'\n", out_path.c_str());
```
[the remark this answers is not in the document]
andrewhop: Good catch, I didn't think about this for this change. Turns out it is the empty string which we pass in as the default value in `GetString(&out_path, "-out", "", parsed_args);` I'll update that so it makes a bit more sense.

samuel40791765 (approved):

tool-openssl/x509.cc
```diff
@@ -22,13 +22,8 @@ static const argument_t kArguments[] = {
...
 };
 
-bool WriteSignedCertificate(X509 *x509, const std::string &out_path) {
-  ScopedFILE out_file(fopen(out_path.c_str(), "wb"));
-  if (!out_file) {
-    fprintf(stderr, "Error: unable to open output file '%s'\n", out_path.c_str());
-    return false;
-  }
-  if (!PEM_write_X509(out_file.get(), x509)) {
+static bool WriteSignedCertificate(X509 *x509, bssl::UniquePtr<BIO> &output_bio, const std::string &out_path) {
```
samuel40791765, line 25: Can `output_bio` be `const`?

WillChilds-Klein (dismissed):

samuel40791765 (approved):

WillChilds-Klein (approved):

PR #1801, opened 2024-08-26
UBE backend

torben-hansen:
### Description of changes: 

Implements the uniqueness-breaking event detection backend with the new function `get_ube_generation_number()` being the entry point. This is basically an abstraction on top of the existing detection methods AWS-LC currently support. The implementation in this PR includes the UBE logic and excludes:
- Using the new function `get_ube_generation_number()` in the new randomness generation.
- Using the fork/snapsafe event detection methods `CRYPTO_get_fork_generation` and `CRYPTO_get_snapsafe_generation`.

These fill follow later.

The UBE subsystem is stateful maintaining the global state `ube_state` which contain the global ube generation number. This value can be used to protect volatile state.

```
                                          caller                         
                                            ▲                            
     ube_state                              │                            
┌──────────────────────────┐                │                            
│    generation_number─────│─────►get_ube_generation_number()            
│                          │                                             
│    chached_fork_gn◄──────│───── get_fork_generation_number_mocked()    
│                          │                                             
│    chached_snapsafe_gn◄──│──────get_snapsafe_generation_number_mocked()
│                          │                                             
└──────────────────────────┘                                                                                    
```

If any error is observed, we transition into an error state and any subsequent calls to `get_ube_generation_number()` will return `0` (the error code). Note that an error is not necessarily fatal it means that caller must take other actions to protect volatile state.

If we are not in an error state, we maintain event detection. `ube_state` stores cached generation numbers for each UBE detection method (`cached_fork_gn` and `cached_snapsafe_gn`). On each entry of `get_ube_generation_number()` we pull new generation numbers for each method in `ube_get_detection_generation_numbers()`:
- if any of these numbers are different from the cached numbers in `ube_state` we synchronise incrementing the ube generation number and return the incremented number to caller.
- otherwise, we simply return the cached number in `ube_state` to caller.

### Testing:

Currently uses mocked events to verify that the detection and incrementation logic works as intended.


crypto/ube/ube.c
```diff
@@ -0,0 +1,238 @@
...
+static int ube_is_detected(struct detection_gn *current_detection_gn) {
+
+  if (ube_state.cached_fork_gn != current_detection_gn->current_fork_gn ||
+      ube_state.cached_snapsafe_gn != current_detection_gn->current_snapsafe_gn) {
+    return 1;
+  }
+  return 0;
+}
+
+// stays invariant after a failure state has been reached.
```
andrewhop, line 172: This feels like an incomplete thoughts, what stays invariant?

crypto/ube/ube.c
```diff
@@ -0,0 +1,238 @@
...
+int get_ube_generation_number(uint64_t *current_generation_number) {
+
+  GUARD_PTR(current_generation_number);
+  *current_generation_number = 0;
+
+  CRYPTO_once(&ube_state_initialize_once, ube_state_initialize);
```
andrewhop, line 178: What is the overhead to this CRYPTO_once getting checked very often? Should this initialization go in CRYPTO_library_init?

crypto/ube/ube.c
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  struct detection_gn current_detection_gn = { 0 };
+
+  // First read generation numbers for each supported detection method. We do
+  // not mutate |ube_state|. So, a read lock is sufficient at this point. Each
+  // individual detection method will have their own concurrency controls if
+  // needed.
+
+  CRYPTO_STATIC_MUTEX_lock_read(&ube_lock);
```
andrewhop, line 195: Why lock here? Nothing in ube_get_detection_generation_numbers touches ube_state, I think we only need the read lock around ube_is_detected.

crypto/ube/ube_test.cc
```diff
@@ -0,0 +1,94 @@
...
+  uint64_t cached_generation_number = 0;
+  if (get_ube_generation_number(&generation_number) == 0) {
+    // In this case, UBE detection is disabled, so just return
+    // successfully. The short-circuit feature means we can't mock detection
+    // methods.
+    return;
+  }
+
+  // The fork generation number is initially 1. Add a larger offset to avoid
+  // any noise.
```
andrewhop, line 34: What noise is there in this nest?

crypto/ube/ube_test.cc
```diff
@@ -0,0 +1,94 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include  <gtest/gtest.h>
```
andrewhop, line 4: Looking at the [coverage report](https://app.codecov.io/gh/aws/aws-lc/pull/1801?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws) can you add tests for:
1. ube_failed and short circuit behavior
2. UBE and race condition with another thread already updating the value

crypto/ube/ube.c
```diff
@@ -0,0 +1,238 @@
...
+// ube_update_state updates the global state |ube_state| with the generation
+// numbers loaded into |current_detection_gn| and increments the UBE generation
+// number.
+static void ube_update_state(struct detection_gn *current_detection_gn) {
+
+  ube_state.generation_number += 1;
+
+  // Make sure we cache all new generation numbers. Otherwise, we might detect
+  // a fork UBE but, in fact, both a fork and snapsafe UBE occurred. Then next
+  // time we enter, a redundant reseed will be emitted.
```
andrewhop, line 127: Can you add a test for this: both fork and snapsafe but ensure the generation number only increments once.

crypto/ube/internal.h
```diff
@@ -0,0 +1,44 @@
...
+// on technology that is unique to a platform. Hence, support for UBE detection
+// also depends on the platform AWS-LC is executed on.
+//
+// The parameter |current_generation_number| must be synchronized by the caller.
+//
+// Returns 1 on success and 0 if not supported. 0 means that UBE detection is
+// not supported and any volatile state must randomize before usage.
+// In case of an error or if UBE detection is unavailable, all subsequent
+// entries will immediately return.
+OPENSSL_EXPORT int get_ube_generation_number(uint64_t *current_generation_number);
```
andrewhop, line 32: Should these be prefixed with something like AWSLC_ to ensure we don't accidentally collide with anything else in the process? Do we have a standard for new APIs we introduce that are private but still exported?

crypto/ube/ube.c
```diff
@@ -0,0 +1,238 @@
...
+  // that only one thread increments the UBE generation number, because the
+  // cached detection method generation numbers have been updated by the thread
+  // that had the first entry.
+
+  CRYPTO_STATIC_MUTEX_lock_write(&ube_lock);
+  if (ube_get_detection_generation_numbers(&current_detection_gn) != 1) {
+    ube_failed();
+    CRYPTO_STATIC_MUTEX_unlock_write(&ube_lock);
+    return 0;
+  }
```
andrewhop, line 223: Instead of having to lock/unlock here did you consider implementing the counters in `ube_state` using atomic counters? Just thinking about our faster new c11 atomics and all the cool places we can use them instead of pthreads locks.

crypto/ube/ube.c
```diff
@@ -0,0 +1,238 @@
...
+  // needed.
+
+  CRYPTO_STATIC_MUTEX_lock_read(&ube_lock);
+  if (ube_get_detection_generation_numbers(&current_detection_gn) != 1) {
+    ube_failed();
+    CRYPTO_STATIC_MUTEX_unlock_read(&ube_lock);
+    return 0;
+  }
+  if (ube_is_detected(&current_detection_gn) == 0) {
+    // No UBE detected, so just grab UBE generation number from the state.
```
andrewhop, line 202: Double checking my understanding on the happy path (no UBE), the work here is:
1. get a read lock
2. check each UBE sensor
3. see no change
4. return current ube_state.generation_number
5. release the read lock

And we will have to do this for every call to volatile data, e.g. RAND_bytes

crypto/ube/ube.c
```diff
@@ -0,0 +1,238 @@
...
+// inefficient for global volatile states if care is not taken. Because every
+// thread only have visibility on their own per-thread generation number, we
+// could trigger a new randomization of the volatile state per-thread. While a
+// single randomization of the global volatile state should be sufficient.
+struct ube_global_state {
+  uint64_t generation_number;
+  uint64_t cached_fork_gn;
+  uint64_t cached_snapsafe_gn;
+};
+static struct ube_global_state ube_state = { 0 };
```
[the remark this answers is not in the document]
torben-hansen: changed as suggested.

crypto/ube/ube.c
```diff
@@ -0,0 +1,238 @@
...
+
+// Convenience object that makes it easier to extend the number of detection
+// methods without having to modify function signatures.
+struct detection_gn {
+#define NUMBER_OF_DETECTION_GENERATION_NUMBERS 2
```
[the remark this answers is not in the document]
torben-hansen: I think this became clear.
To be close to what the value `2` represent.

crypto/ube/ube.c
torben-hansen: That makes sense. I did that.

crypto/ube/ube.c
torben-hansen: forgot to remove...

crypto/ube/ube.c
torben-hansen: Three reasons:
- not all users need to use the rand backend. So, to stay true to lazy, we can avoid it.
- after the first entry. The code of CRYPTO_once is highly negligible. 
- It's great to have the initialization code here for clarity.

In fact, there is a 4th reason. I ponder making a `RAND_bytes()` call in the library initialisation logic, if we don't use a static initialisation only build type. Haven't finished pondering that. But in any case, I think it makes sense to keep the init here.

crypto/ube/ube.c
torben-hansen: Hmm, yeah I think that would work.

torben-hansen: Obviously, this require `ube_get_detection_generation_numbers` to be thread-safe. But we explicitly do make each detection function thread-safe as long as any shared argument memory is serialised. But no shared argument memory in this case, so we are good.

crypto/ube/ube.c
torben-hansen: That's correct.

> And we will have to do this for every call to volatile data, e.g. RAND_bytes

Also correct. Luckily the detection methods we use are really really fast: each method includes some read locking (in happy path no contention), a pointer dereference, and a couple of equality checks. So, on the happy path this work is very negligible compared to the cost of e.g. RAND_bytes as a whole.

The unhappy path (we detected something). This is a bit more expensive and we have the mutex. But that is the cost of allowing forking and other concurrency.

crypto/ube/ube.c
torben-hansen: I'm not sure we can replace synchronising here with atomic operations only. Because we also synchronise to ensure that we only increment once per UBE. Is there another primitive that ensures that?

If we do not mutex this entire block, how can we ensure that all logic in `ube_is_detected` use the updated global state?

crypto/ube/ube_test.cc
torben-hansen: I plan way more extensive testing down the line, quite a bit more than we currently have.

The test you mention require more extensive facilities. I want to add those in separate PRs to avoid doing many things at the same time.

crypto/ube/ube_test.cc
torben-hansen: This is a left over thought... Nothing right now! But tests are run concurrently, and stuff under `ube_get_detection_generation_numbers` is global. So, we need to be a bit careful when writing these tests because direct modification of the global state will affect other tests that might expect different behaviour.

torben-hansen: I changed the wording. But I planned to tackle this in subsequent work.

crypto/ube/ube.c
andrewhop: I was thinking we could make the global counters atomics, but that wouldn't get us the property that they are only incremented once per UBE, two threads could increment it at the same time and that might cause additional seeding which would be inefficient.

crypto/ube/ube.c
andrewhop: What do you mean in the 4th reason? Would we need to add a call to RAND_bytes in the library initializer?

crypto/ube/ube.c
```diff
@@ -0,0 +1,238 @@
...
+    *gn = testing_snapsafe_generation_number;
+  }
+  return 1;
+}
+static uint64_t get_fork_generation_number_mocked(void) {
+  if (testing_fork_generation_number != 0) {
+    return testing_fork_generation_number;
+  }
+  return 1;
+}
```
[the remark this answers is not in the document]
andrewhop: Ahhh if this needs to fit existing APIS for now that's fine, but I would like to make the APIs match in the future so it's easier to read and reason about the flow that needs to check them.

crypto/ube/ube.c
```diff
@@ -0,0 +1,236 @@
...
+      ube_global_state.cached_snapsafe_gn != current_detection_gn->current_snapsafe_gn) {
+    return 1;
+  }
+  return 0;
+}
+
+int CRYPTO_get_ube_generation_number(uint64_t *current_generation_number) {
+
+  GUARD_PTR(current_generation_number);
+  *current_generation_number = 0;
```
skmcgrail, line 175: Should we avoid overriding the value of `*current_generation_number` until after we are actually sure of the value to overwrite it with? Otherwise this will always set it to zero first but could then subsequently fail either due to ube detection being unavailable, failing to get the generation number etc. Might be easier to reason about it in a debugger later if you know it only overrides the value on successful returns?

skmcgrail (approved):

andrewhop (approved):

PR #1803, opened 2024-08-26
Add CAST for SP 800-56Cr2 One-Step function

skmcgrail:
### Description of changes: 
Adds a cryptographic algorithm self test (CAST) for SP 800-56Cr2 OneStep function.

Per the FIPS 140-3 I.G. Guidance Item:
> if the module implements an approved KDA (SP 800-56C rev1 or rev2), the module shall perform at
least one CAST covering a SP 800-56Crev1 Section 4 one-step KDF (if implemented) and another
CAST covering a SP 800-56Crev1 Section 5 two-step KDF (if implemented), including at least one
auxiliary function for each. For example, if a module implements both a two-step KDF using either
the HMAC or AES-CMAC auxiliary function, and one-step KDF using either the hash or HMAC
auxiliary function, then at least two CASTs are required: one for the two-step KDF (with either
HMAC or AES-CMAC), and the other for the one-step KDF (with either the hash or HMAC). The
implementations of the auxiliary functions used in the KDA CASTs do not require separate CASTs.
In addition, the module shall self-test all underlying prerequisite algorithms used in the remaining SP
800-56C rev1 or rev2 schemes implemented in the module, if the same implementations of the
underlying algorithms are not already self-tested either on their own or as part of other higher-level
algorithm CASTs.

I have interpreted this item as requiring us only needing to test either `SSKDF_digest` or `SSKDF_hmac` but aren't required to test both.

### Testing:
The KAT used in this test case was sourced from `util/fipstools/acvp/acvptool/test/vectors/KDA-OneStep.bz2` and `util/fipstools/acvp/acvptool/test/expected/KDA-OneStep.bz2`.


crypto/fipsmodule/self_check/self_check.c
```diff
@@ -1202,6 +1202,49 @@ static int boringssl_self_test_fast(void) {
...
+      0x6b, 0x40, 0xcb, 0x34, 0xba, 0xe2, 0xd2, 0x6e, 0x9f, 0x45, 0x52, 0xb6,
+      0xb1, 0xa2, 0x6f, 0xa5};  // 256 bytes
+  uint8_t sskdf_digest_output[sizeof(kSSKDFDigestDerivedKey)];
+  if (!SSKDF_digest(&sskdf_digest_output[0], sizeof(kSSKDFDigestDerivedKey),
+                    EVP_sha256(), &kSSKDFDigestSharedSecret[0],
+                    sizeof(kSSKDFDigestSharedSecret), &kSSKDFDigestInfo[0],
+                    sizeof(kSSKDFDigestInfo)) ||
+      !check_test(kSSKDFDigestDerivedKey, sskdf_digest_output,
+                  sizeof(kSSKDFDigestDerivedKey), "SSKDF_digest KAT")) {
+    fprintf(stderr, "SSKDF_digest failed.\n");
```
WillChilds-Klein, line 1245: do we need to `goto err;` here?

WillChilds-Klein (approved):

nebeid (approved):

PR #1808, opened 2024-08-27
Fix GitHub/CodeBuild Purge Lambda

justsmth:
### Issues:
Resolves P149524833

### Description of changes: 
* Fix our CodeBuild Purge Lambda

### DIFF
--- a/tests/ci/lambda/src/bin/purge-stale-builds.rs
+++ b/tests/ci/lambda/src/bin/purge-stale-builds.rs
@@ -94,10 +94,12 @@ async fn handle(_event: LambdaEvent<Value>) -> Result<(), Error> {
                 .set_filters(Some(ec2_framework_filters))
                 .send()
                 .await;
-            if let Ok(output) = result {
-                Some(output)
-            } else {
-                None
+            match result {
+                Ok(output) => Some(output),
+                Err(err) => {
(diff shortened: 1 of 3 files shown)


tests/ci/lambda/src/bin/purge-stale-builds.rs
```diff
@@ -40,6 +42,8 @@ async fn handle(_event: LambdaEvent<Value>) -> Result<(), Error> {
 
     let project = std::env::var("CODEBUILD_PROJECT_NAME").unwrap();
 
+    let have_ec2_permissions: bool = project == "aws-lc-ci-ec2-test-framework";
```
WillChilds-Klein, line 45: if you don't have permissions, should you just return here? what processing can you do without permissions?

justsmth: The operations related to `PrunedGitHubBuilds` can still be performed and the metrics should still be generated. (Most projects will have a 0-metric for `TerminatedEC2Instances`.) The problem was that the Lambda jobs for most of the projects (all except "aws-lc-ci-ec2-test-framework") were failing due to lack of EC2 permissions.

WillChilds-Klein (approved):

PR #1818, opened 2024-08-29
Add ED25519 ACVP Testing

skmcgrail:


util/fipstools/acvp/acvptool/test/expected/EDDSA.bz2
nebeid, line 1: Both expected responses for SigVer and KeyVer are false in this file. Should we include 2 Passed:true tests?

skmcgrail: Let me pull the demo vectors again, I'm pretty sure there were truthy test cases but the test trim tool might have pruned them out since it's agnostic to the tests themselves.

skmcgrail: Discussed offline, but the trim vectors tool intent is to only leave one ACVP test per ACVP test group. So this is currently an expected outcome behavior that this trimming is exhaustive regardless of the perceived of the test. It would appear this level of ACVP testing in the repository is more to confirm the functionality of the tool, but may not be an exhaustive representation of the full suite of tests you would get from NIST in this format.

 Long-term it would be nice to have something similar to our crypfuzz framework where we stash/keep a much larger set of ACVP vectors and use that it do additional exhaustive testing of our algorithms. Maybe a future intern project?

For now I confirmed that the vectors I validated with the NIST demo server had the variants in question that you referenced and successfully passed.

torben-hansen (dismissed):

nebeid (dismissed):

nebeid (approved):

torben-hansen (approved):

PR #1828, opened 2024-09-05
Move KEM API and ML-KEM definitions to FIPS module

torben-hansen:
### Issues:

Related to CryptoAlg-2615

### Description of changes: 

This is the 1 of 3 PRs that re-organize the KEM code. The goal is to move the KEM API into the FIPS module and it's associated glue code.

In this PR, we first move the ML-KEM definitions under the FIPS module and the set of KEM API. There are two objects to keep track of:

`KEM` is the "context" object of a kem method storing e.g. size information, unique identifiers, and a reference to a method table (of type `KEM_METHOD`).

`KEM_METHOD` is the method table that stores references to implementations for each abstract operation.

Both `KEM` and `KEM_METHOD` has been defined outside the FIPS module. Move this under the module now. Previously, also the KEM method function tables were defined in their own source code file. Move these together with the definitions of the KEMs. They will later go to their own file.

Note, that `KEM` was defined with only `const` fields. I think this was my invention but it was not an optimal choice. It only allows static initialization. This is not good with the current static FIPS build, so remove the `const` qualifier from all non-pointer fields in `KEM`.

Finally, drop part of the table search method. KEMs are now directly referenced instead of searching through a table. Retain the table search for the legacy Kyber kems for now though to minimize code changes in each PR.

### Call-outs:

In the following PRs:
- move the legacy kyber KEMs into the Kyber folder and delete the out-of-module kem folder.
- move the EVP kem method tables to the fips module.


crypto/fipsmodule/kem/kem.c
```diff
@@ -0,0 +1,270 @@
...
+#include <openssl/base.h>
+
+#include "../../kem/internal.h"
+#include "../delocate.h"
+#include "../ml_kem/ml_kem.h"
+#include "internal.h"
+
+static const uint8_t kOIDMLKEM512[]  = {0xff, 0xff, 0xff, 0xff};
+static const uint8_t kOIDMLKEM768[]  = {0xff, 0xff, 0xff, 0xff};
+static const uint8_t kOIDMLKEM1024[] = {0xff, 0xff, 0xff, 0xff};
```
andrewhop, line 13: Are there real values for this yet?

andrewhop (approved):

samuel40791765 (approved):

crypto/fipsmodule/kem/kem.c
torben-hansen: placeholder values as before. I will change before code freeze.

PR #1831, opened 2024-09-06
Revert "Remove param_decode and param_encode EVP_PKEY hooks."

samuel40791765:
This reverts part of  commit 32fdc512ca6aed2473a63f8a826705a122d4ea0c.

### Issues:
Addresses `CryptoAlg-2546`

### Description of changes: 
commit 32fdc512ca6aed2473a63f8a826705a122d4ea0c removed the function pointers for `param_decode` and `param_encode` and removed `PEM_read_bio_Parameters` and `PEM_write_bio_Parameters` along with it (which was the only API consuming these hooks). This reverts part of the commit to add these back. 

### Call-outs:
1. This is in draft for the time being until we have proper support for `EVP_PKEY_DH` ASN1 serialization.
2. Do we even need function pointers if only 3 types of `EVP_PKEY`s consume these? Considering if we can just pull the logic into the API instead of adding a function pointer for each `EVP_PKEY`.

### Testing:
WIP


crypto/pem/pem_pkey.c
```diff
@@ -156,6 +157,66 @@ int PEM_write_bio_PrivateKey(BIO *bp, EVP_PKEY *x, const EVP_CIPHER *enc,
...
+             !EVP_PKEY_set_type(ret, EVP_PKEY_DH)) {
+    // TODO: EVP_PKEY_DH ASN1 support is missing. Work on this once we have it.
+    goto err;
+  }
+
+  if (!ret->ameth->param_decode || !ret->ameth->param_decode(ret, &p, len)) {
+    EVP_PKEY_free(ret);
+    OPENSSL_PUT_ERROR(PEM, ERR_R_ASN1_LIB);
+    goto err;
+  }
```
justsmth, line 188: This `EVP_PKEY_free(ret)` would cause a double-free, as `EVP_PKEY_free(ret)` is called again after `err:`.

crypto/pem/pem_pkey.c
```diff
@@ -156,6 +157,66 @@ int PEM_write_bio_PrivateKey(BIO *bp, EVP_PKEY *x, const EVP_CIPHER *enc,
...
+
+
+  if (pkey != NULL) {
+    if (*pkey != NULL) {
+      EVP_PKEY_free(*pkey);
+    }
```
justsmth, line 194: Is it possible that `*pkey` was simply not initialized by the caller, in which case we could be freeing (or down-ref'ing) some arbitrary address? Is it documented somewhere that the `*pkey` must be initialized to either NULL or to a valid `EVP_PKEY` pointer?

crypto/pem/pem_pkey.c
```diff
@@ -156,6 +157,66 @@ int PEM_write_bio_PrivateKey(BIO *bp, EVP_PKEY *x, const EVP_CIPHER *enc,
...
+  } else if (strcmp(nm, PEM_STRING_DSAPARAMS) == 0 ||
+             !EVP_PKEY_set_type(ret, EVP_PKEY_DSA)) {
+    // TODO: This should work, we only support serializing DSA?
+    goto err;
+  } else if (strcmp(nm, PEM_STRING_DHPARAMS) == 0 ||
+             !EVP_PKEY_set_type(ret, EVP_PKEY_DH)) {
+    // TODO: EVP_PKEY_DH ASN1 support is missing. Work on this once we have it.
+    goto err;
+  }
+
```
justsmth, line 183: If these are the only 3 supported types, should there be a final `else` block that bails out?
```
} else {
  goto err;
}
```

samuel40791765 (commented):
Not sure if the original comments still apply when this was in "Draft" and unfinished. The logic has changed considerably since then.
I think we can get away with not implementing the `EVP_PKEY_DH` ASN1 hooks with this method going forward, feel free to leave any comments again if the original ones still apply.

include/openssl/pem.h
```diff
@@ -473,6 +474,19 @@ OPENSSL_EXPORT int PEM_write_PKCS8PrivateKey(FILE *fp, const EVP_PKEY *x,
                                              int klen, pem_password_cb *cd,
                                              void *u);
 
+// PEM_read_bio_Parameters is a generic PEM deserialization function that
+// parses the public "parameters" in |bio| and returns a corresponding
+// |EVP_PKEY|. If |*pkey| is non-null, the original |*pkey| is freed and the
+// returned |EVP_PKEY| is also written to |*pkey|. This is only supported with
```
justsmth, line 480: I know it's implied from this statement, but it would be good to explicitly say that `*pkey` *must be initialized* (either to NULL or to an allocated value) prior to being passed in. (Passing in an uninitialized pointer would be UB.)

crypto/pem/pem_pkey.c
```diff
@@ -156,6 +157,88 @@ int PEM_write_bio_PrivateKey(BIO *bp, EVP_PKEY *x, const EVP_CIPHER *enc,
...
+err:
+  EVP_PKEY_free(ret);
+  OPENSSL_free(nm);
+  OPENSSL_free(data);
+  return NULL;
+}
+
+int PEM_write_bio_Parameters(BIO *bio, EVP_PKEY *pkey) {
+  if (bio == NULL || pkey == NULL || pkey->ameth == NULL) {
```
justsmth, line 216: Why the check for `pkey->ameth`?  Should this be a check for `pkey->type`?

include/openssl/pem.h
samuel40791765: Added clarification

crypto/pem/pem_pkey.c
```diff
@@ -156,6 +157,88 @@ int PEM_write_bio_PrivateKey(BIO *bp, EVP_PKEY *x, const EVP_CIPHER *enc,
   return PEM_write_bio_PKCS8PrivateKey(bp, x, enc, (char *)kstr, klen, cb, u);
 }
 
+EVP_PKEY *PEM_read_bio_Parameters(BIO *bio, EVP_PKEY **pkey) {
+  char *nm = NULL;
+  unsigned char *data = NULL;
+  long len;
+  if (!PEM_bytes_read_bio(&data, &len, &nm, PEM_STRING_PARAMETERS, bio, 0,
```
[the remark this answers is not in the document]
samuel40791765: Added documentation

skmcgrail (approved):

justsmth (approved):

PR #1836, opened 2024-09-06
Add PKCS7-internal BIO_f_cipher

WillChilds-Klein:
### Issues:
Addresses CryptoAlg-2494

### Description of changes: 

This change introduces a new filter BIO, `BIO_f_cipher` for use in PR 1816.

The cipher BIO sits in front of a backing BIO, encrypting incoming writes before writing to the backing BIO and decrypting data read from the backing BIO. This implementation is almost an exact copy of [OpenSSL's](https://github.com/openssl/openssl/blob/8e0d479b98357bb20ab1bd073cf75f7d42531553/crypto/evp/bio_enc.c#L59) with some functionality removed. We try to change as little of the underlying logic as possible, but rename variables and add comments for clarity.

### Call-outs:
- n/a

### Testing:
- new unit tests. test considerations and flow are documented in-line.

---


include/openssl/bio.h
```diff
@@ -910,6 +910,14 @@ OPENSSL_EXPORT int (*BIO_meth_get_puts(const BIO_METHOD *method)) (BIO *, const
...
 OPENSSL_EXPORT OPENSSL_DEPRECATED const BIO_METHOD *BIO_s_secmem(void);
 
+// BIO_f_cipher is used internally by the pkcs7 module. It is not recommended
+// for external use.
+OPENSSL_EXPORT OPENSSL_DEPRECATED const BIO_METHOD *BIO_f_cipher(void);
+
+// BIO_set_cipher is used internally for testing. It is not recommended for
+// external use.
+OPENSSL_EXPORT OPENSSL_DEPRECATED int BIO_set_cipher(BIO *b, const EVP_CIPHER *cipher,
+                   const unsigned char *key, const unsigned char *iv, int enc);
```
samuel40791765, line 920: We could actually move this to an `internal.h` header file and still append `OPENSSL_EXPORT` ahead of it.
That way it can still be used during testing, but not exposed for external use.

samuel40791765 (commented):
Just some preliminary comments, haven't gone through everything yet.

crypto/bio/cipher.c
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  int bytes_processed = 0;
+  int remaining = outl;
+  while (remaining > 0) {
+    if (ctx->buf_len > 0) {
+      int bytes_decrypted;
+      int to_decrypt = remaining > ctx->buf_len ? ctx->buf_len : remaining;
+      char *out_pos = out + bytes_processed;
+      if (!EVP_DecryptUpdate(ctx->cipher, (uint8_t *)out_pos, &bytes_decrypted,
```
justsmth, line 109: The condition on this call to `EVP_DecryptUpdate` needs to be more strict. From its documentation:
> The number of output bytes may be up to |in_len| plus the block length minus one and |out| must have sufficient space.

So it must be the case that: `remaining > (to_decrypt + ctx->block_len - 1)`

crypto/bio/cipher.c
```diff
@@ -0,0 +1,300 @@
...
+                             &ctx->buf[ctx->buf_off], to_decrypt)) {
+        BIO_copy_next_retry(b);
+        return bytes_processed;
+      };
+      ctx->buf_len -= bytes_decrypted;
+      ctx->buf_off += bytes_decrypted;
```
justsmth, line 115: The change here should be based on the value of `to_decrypt` b/c that is the number of bytes processed from this buffer.

crypto/bio/cipher.c
```diff
@@ -0,0 +1,300 @@
...
+
+typedef struct enc_struct {
+  uint8_t done;     // > 0 when finished
+  uint8_t flushed;  // >0 when buffered data has been flushed
+  uint8_t ok;       // decryption status
+  int buf_off;      // start idx of buffered data
+  int buf_len;      // length of buffered data
+  EVP_CIPHER_CTX *cipher;
+  uint8_t buf[ENC_BLOCK_SIZE];
+} BIO_ENC_CTX;
```
justsmth, line 31: The block length of the cipher is a value that will need to be considered on each operation. I think it should be in this struct and initialized along with `cipher`.

crypto/bio/cipher.c
```diff
@@ -0,0 +1,300 @@
...
+  ctx->buf_off = 0;
+  ctx->buf_len = 0;
+  BIO_set_data(bi, ctx);
+  BIO_set_init(bi, 1);
+
+  return 1;
+}
+
+static int enc_free(BIO *a) {
+  BIO_ENC_CTX *b;
```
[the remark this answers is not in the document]
WillChilds-Klein: yes.

crypto/bio/cipher.c
```diff
@@ -0,0 +1,300 @@
...
+    NULL,  // enc_callback_ctrl
+};
+
+const BIO_METHOD *BIO_f_cipher(void) { return &methods_enc; }
+
+static int enc_new(BIO *bi) {
+  BIO_ENC_CTX *ctx;
+
+  if ((ctx = OPENSSL_zalloc(sizeof(*ctx))) == NULL) {
+    return 0;
```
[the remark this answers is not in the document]
WillChilds-Klein: yep, thanks.

crypto/bio/cipher.c
WillChilds-Klein: to avoid having to handle this in all possible `EVP_CipherInit` cases, i'd prefer to get the block size with a function call on the cipher. i can do this outside of the loop so it's only called once per invocation of cipher BIO read/write routine.

crypto/bio/cipher.c
WillChilds-Klein: yes, good point.

crypto/bio/cipher.c
WillChilds-Klein: good point. i've added this guard, slightly re-written for clarity.

crypto/bio/cipher.c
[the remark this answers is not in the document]
samuel40791765: An error is already emitted if the underlying call to `OPENSSL_malloc` fails: https://github.com/aws/aws-lc/blob/8ed554c8a3758df6879193936bac4dd3fc3882d5/crypto/mem.c#L199-L203

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,317 @@
...
+};
+
+const BIO_METHOD *BIO_f_cipher(void) { return &methods_enc; }
+
+static int enc_new(BIO *b) {
+  BIO_ENC_CTX *ctx;
+  GUARD_PTR(b);
+
+  if ((ctx = OPENSSL_zalloc(sizeof(*ctx))) == NULL) {
+    OPENSSL_PUT_ERROR(BIO, ERR_R_MALLOC_FAILURE);
```
samuel40791765, line 50: Just in case this comment gets lost: https://github.com/aws/aws-lc/pull/1836#discussion_r1779286649

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,317 @@
...
+    return 0;
+  }
+
+  switch (cmd) {
+    case BIO_CTRL_RESET:
+      ctx->ok = 1;
+      ctx->flushed = 0;
+      if (!EVP_CipherInit_ex(ctx->cipher, NULL, NULL, NULL, NULL,
+                             EVP_CIPHER_CTX_encrypting(ctx->cipher)))
+        return 0;
```
samuel40791765, line 232: Nit:
```suggestion
      if (!EVP_CipherInit_ex(ctx->cipher, NULL, NULL, NULL, NULL,
                             EVP_CIPHER_CTX_encrypting(ctx->cipher))) {
        return 0;
     }
```

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,317 @@
...
+    case BIO_CTRL_PENDING:
+      // Calculate number of bytes left to process if we have anything buffered,
+      // else consult underlying BIO.
+      ret = ctx->buf_len - ctx->buf_off;
+      if (ret <= 0) {
+        ret = BIO_ctrl(next, cmd, num, ptr);
+      }
+      break;
+    case BIO_CTRL_FLUSH:
+      enc_flush(b, next, ctx, /*do_final*/ 1);
```
samuel40791765, line 252: Does the return value of `enc_flush` needs to be checked here?

crypto/bio/cipher.c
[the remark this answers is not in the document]
WillChilds-Klein: Assuming we don't hit [either of the two preceding conditionals](https://github.com/aws/aws-lc/blob/8ed554c8a3758df6879193936bac4dd3fc3882d5/crypto/mem.c#L166-L175), yes.

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: ooo good find! i have [a test case](https://github.com/aws/aws-lc/pull/1836/files#diff-4cf34798a14703c26416adc6fe0643a8a226c8a0a1f4b9169c5b70741c9af715R261) for this, but it must have failed on flushing the underlying BIO, not when calling `enc_flush`. fixed.

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,320 @@
...
+  const int cipher_block_size = EVP_CIPHER_CTX_block_size(ctx->cipher);
+  while (remaining > 0) {
+    if (ctx->buf_len > 0) {
+      int bytes_decrypted;
+      int to_decrypt = remaining > ctx->buf_len ? ctx->buf_len : remaining;
+      uint8_t *out_pos = ((uint8_t *)out) + bytes_processed;
+      // |EVP_DecryptUpdate| may write up to cipher_block_size-1 more bytes than
+      // requested, so return early if we cannot accommodate that with current
+      // |remaining| byte count.
+      if ((to_decrypt > (remaining - cipher_block_size + 1)) ||
```
justsmth, line 109: There's no guarantee that the caller will ever pass in an `outl` that's larger than `cipher_block_size + 1`.  For reads, I think `ctx-buf` should be a buffer of plaintext instead of the ciphertext.

samuel40791765 (approved):

samuel40791765 (dismissed):

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: I've rewritten the read routine to buffer in plaintext, decrypting immediately after reading from the underlying BIO.

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,327 @@
...
+    if (ctx->buf_len == 0) {
+      ctx->buf_off = 0;
+      int to_encrypt =
+          remaining > (int)sizeof(ctx->buf) ? (int)sizeof(ctx->buf) : remaining;
+      if (!EVP_EncryptUpdate(ctx->cipher, &ctx->buf[0], &ctx->buf_len,
+                             (uint8_t *)in, to_encrypt)) {
+        ret = bytes_processed;
+        goto out;
+      };
+      bytes_processed += ctx->buf_len;
```
justsmth, line 201: The change to `bytes_processed` and `remaining` should be based on `to_encrypt`, i.e., the number of bytes passed as input, not the number of bytes generated as output (which could potentially be more than `remaining`).

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,327 @@
...
+      return outcome;
+    }
+    ctx->buf_off += outcome;
+    ctx->buf_len -= outcome;
+  }
+  ctx->buf_off = 0;
+  ctx->buf_len = 0;
+  if (do_final && !ctx->flushed) {
+    ctx->flushed = 1;
+    ctx->ok = EVP_CipherFinal_ex(ctx->cipher, ctx->buf, &ctx->buf_len);
```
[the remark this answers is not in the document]
WillChilds-Klein: agreed.

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,327 @@
...
+      }
+      return bytes_output;
+    } else if (bytes_read <= 0) {
+      BIO_copy_next_retry(b);
+      return bytes_output;
+    }
+    // Decrypt ciphertext in place, update |ctx->buf_len| with num bytes
+    // decrypted. Treat decrypt error as EOF and set error.
+    ctx->ok =
+        EVP_DecryptUpdate(ctx->cipher, &ctx->buf[ctx->buf_off], &ctx->buf_len,
```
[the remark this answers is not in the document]
WillChilds-Klein: yes. this was a choice in notation. i don't feel strongly between as-is, `&ctx->buf[0]`, or `ctx->buf`. here and elsewhere relevant, changed to `ctx->buf` as that seems more conventional.

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,327 @@
...
+      break;
+    }
+    ctx->buf_len = 0;
+    ctx->buf_off = 0;
+    // |EVP_DecryptUpdate| may write up to cipher_block_size-1 more bytes than
+    // requested, so only read bytes we're sure we can decrypt in place.
+    int to_read = (int)sizeof(ctx->buf) - cipher_block_size + 1;
+    int bytes_read = BIO_read(next, &ctx->buf[ctx->buf_off], to_read);
+    if (bytes_read == 0 && BIO_eof(next)) {
+      ctx->eof = 1;
```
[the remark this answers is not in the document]
WillChilds-Klein: i decided to consolidate `eof` and `flushed` into a singular `done`. i'm not sure of a case where setting `done` here vs. post-copy would matter, but i agree that it makes more logical sense to set it after we've copied everything out. changed.

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  assert(ctx->buf_off == ctx->buf_len && ctx->buf_len == 0);
+  int bytes_processed = 0;
+  int remaining = inl;
+  while (remaining > 0) {
+    if (ctx->buf_len == 0) {
+      ctx->buf_off = 0;
+      int to_encrypt =
+          remaining > (int)sizeof(ctx->buf) ? (int)sizeof(ctx->buf) : remaining;
```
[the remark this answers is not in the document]
WillChilds-Klein: yes. fixed.

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: agreed. fixed and renamed `bytes_processed` for clarity.

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,314 @@
...
+      ctx->ok = EVP_EncryptUpdate(ctx->cipher, ctx->buf, &ctx->buf_len,
+                                  (uint8_t *)in, to_encrypt);
+      if (!ctx->ok) {
+        goto out;
+      };
+      bytes_consumed += to_encrypt;
+      remaining -= to_encrypt;
+    }
+    int bytes_written = BIO_write(next, &ctx->buf[ctx->buf_off], ctx->buf_len);
+    if (bytes_written <= 0) {
```
justsmth, line 194: For [`BIO_write`](https://github.com/aws/aws-lc/blob/0ea1cb1ac4b283b318be8e87bc02022c1215be68/include/openssl/bio.h#L138-L143), I think we should distinguish `bytes_written == 0` (no bytes written) from `bytes_written < 0` (an error). On an error, we might need to set `ctx->ok = 0` and `ctx->done = 1`?

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,314 @@
...
+                                bytes_read);
+  }
+  return bytes_output;
+}
+
+static int enc_flush(BIO *b, BIO *next, BIO_ENC_CTX *ctx) {
+  GUARD_PTR(b);
+  GUARD_PTR(next);
+  GUARD_PTR(ctx);
+  while (ctx->buf_len > 0) {
```
justsmth, line 147: I think this loop instead be conditioned on `ctx->buf_len > 0 || ctx->done == 0`.  It seems possible/likely that `ctx->buf_len == 0` holds when `BIO_flush` is called, in which case this loop never executes (`EVP_CipherFinal_ex` never gets called), and it returns 1 as-if it succeeded.
```
  while (ctx->buf_len > 0 || ctx->done == 0) {
    if (ctx->buf_len > 0) {
...
    } else if(ctx->done == 0) {
...
    }
  }
```

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,314 @@
...
+    int bytes_written = BIO_write(next, &ctx->buf[ctx->buf_off], ctx->buf_len);
+    if (bytes_written <= 0) {
+      BIO_copy_next_retry(b);
+      return bytes_written;
+    }
```
justsmth, line 152: In this case, I think the return value should be 0 b/c the buffer was not fully flushed. (If we return a positive value, the caller may assume `EVP_CipherFinal_ex` was called and all data was written.

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,314 @@
...
+      ctx->ok = EVP_CipherFinal_ex(ctx->cipher, ctx->buf, &ctx->buf_len);
+      // We have more bytes to output, go back to top of loop.
+      if (ctx->buf_len > 0) {
+        continue;
+      }
+      ctx->done = 1;  // If we don't need to output any more bytes, set done.
+      return bytes_output;
+    } else if (bytes_read <= 0) {
+      BIO_copy_next_retry(b);
+      return bytes_output;
```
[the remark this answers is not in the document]
WillChilds-Klein: not necessarily. 0 would indicate EOF (not an error) and < 0 might be a transient error (not a terminal one) for the underlying BIO, so we copy the retry state to let caller handle that accordingly.

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: for same reason as above, i think we should copy underlying BIO's retry state and return without setting `ctx->ok` or `ctx->done` -- there might still be a recoverable case.

but we should probably do a better job of surfacing the underlying return code (right now we just `return bytes_consumed`, which may be >= 0, but now < 0).

how about this for handling <= 0 return value from BIO_read/BIO_write? if `bytes_consumed` > 0, return that so caller knows how much to increment the input pointer on next call. else, return return value from BIO_read/BIO_write.

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: i agree, came to this realization as well while reworking enc_flush to work with block ciphers.

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: agreed.

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,310 @@
...
+      // EVP_DecryptFinal_ex may write up to one block to our buffer. If that
+      // happens, continue the loop to process the decrypted block as normal.
+      ctx->ok = EVP_DecryptFinal_ex(ctx->cipher, ctx->buf, &ctx->buf_len);
+      ctx->done = 1;  // If we can't read any more bytes, set done.
+    } else {
+      // |BIO_read| returned <= 0, but no EOF. Copy retry and return.
+      BIO_copy_next_retry(b);
+      break;
+    }
+    if (!ctx->ok) {
```
justsmth, line 134: NP: I think this check is only relevant after line 122. Should it be moved to there?

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,310 @@
...
+      ctx->ok = EVP_DecryptUpdate(ctx->cipher, ctx->buf, &ctx->buf_len,
+                                  read_buf, to_decrypt);
+    } else if (BIO_eof(next)) {
+      // EVP_DecryptFinal_ex may write up to one block to our buffer. If that
+      // happens, continue the loop to process the decrypted block as normal.
+      ctx->ok = EVP_DecryptFinal_ex(ctx->cipher, ctx->buf, &ctx->buf_len);
+      ctx->done = 1;  // If we can't read any more bytes, set done.
+    } else {
+      // |BIO_read| returned <= 0, but no EOF. Copy retry and return.
+      BIO_copy_next_retry(b);
```
justsmth, line 131: If we reach a no-retry here situation, should we also set `ctx->done = 1` and `ctx->ok = 0` to avoid reading from `next` again?

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,310 @@
...
+}
+
+static int enc_flush(BIO *b, BIO *next, BIO_ENC_CTX *ctx) {
+  GUARD_PTR(b);
+  GUARD_PTR(next);
+  GUARD_PTR(ctx);
+  while (ctx->buf_len > 0 || !ctx->done) {
+    int bytes_written = BIO_write(next, &ctx->buf[ctx->buf_off], ctx->buf_len);
+    if (ctx->buf_len > 0 && bytes_written <= 0) {
+      BIO_copy_next_retry(b);
```
justsmth, line 148: If we reach a no-retry here situation, should we also set ctx->done = 1 and ctx->ok = 0 to avoid writing to `next` again?

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,310 @@
...
+    }
+  }
+  return bytes_output;
+}
+
+static int enc_flush(BIO *b, BIO *next, BIO_ENC_CTX *ctx) {
+  GUARD_PTR(b);
+  GUARD_PTR(next);
+  GUARD_PTR(ctx);
+  while (ctx->buf_len > 0 || !ctx->done) {
```
justsmth, line 145: Should this also be conditioned on `ctx->ok`?
```
while (ctx->ok && (ctx->buf_len > 0 || !ctx->done)) {
```

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,310 @@
...
+    }
+  }
+  return ctx->ok;
+}
+
+static int enc_write(BIO *b, const char *in, int inl) {
+  GUARD_PTR(b);
+  GUARD_PTR(in);
+  BIO_ENC_CTX *ctx = BIO_get_data(b);
+  if (ctx == NULL || ctx->cipher == NULL || ctx->done || !ctx->ok) {
```
justsmth, line 166: NP: Should we also bail out early if `inl <= 0`?

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,314 @@
...
+  BIO *next = BIO_next(b);
+  if (next == NULL) {
+    return 0;
+  }
+
+  int bytes_consumed = 0;
+  int remaining = inl;
+  const int max_crypt_size =
+      (int)sizeof(ctx->buf) - EVP_CIPHER_CTX_block_size(ctx->cipher) + 1;
+  while (remaining > 0) {
```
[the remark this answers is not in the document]
justsmth: Should we check for these conditions at the start of the loop?

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,310 @@
...
+    // is internal, we can fail loudly if they're called. If this case is hit,
+    // it likely means you're making a change that will require implementing
+    // these.
+    case BIO_CTRL_DUP:
+    case BIO_CTRL_GET_CALLBACK:
+    case BIO_CTRL_SET_CALLBACK:
+    case BIO_C_DO_STATE_MACHINE:
+    case BIO_C_GET_CIPHER_CTX:
+      OPENSSL_PUT_ERROR(PKCS7, ERR_R_BIO_LIB);
+      return 0;
```
justsmth, line 264: NP: Do any of these cases potentially need to be propagated to `next` (as we do for default) instead of producing an error.

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,310 @@
...
+  // We only support a modern subset of available EVP_CIPHERs. Other ciphers
+  // (e.g. DES) and cipher modes (e.g. CBC, CCM) had issues with block alignment
+  // and padding during testing, so they're forbidden for now.
+  const EVP_CIPHER *kSupportedCiphers[] = {
+      EVP_aes_128_cbc(),       EVP_aes_128_ctr(), EVP_aes_128_ofb(),
+      EVP_aes_256_cbc(),       EVP_aes_256_ctr(), EVP_aes_256_ofb(),
+      EVP_chacha20_poly1305(),
+  };
+  int supported = 0;
+  for (size_t i = 0; i < sizeof(kSupportedCiphers) / sizeof(EVP_CIPHER *);
```
justsmth, line 291: NP: Compute the size of the array into its own constant.
```
  const size_t kSupportedCiphersCount =
      sizeof(kSupportedCiphers) / sizeof(EVP_CIPHER *);
  int supported = 0;
  for (size_t i = 0; i < kSupportedCiphersCount; i++) {
```

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,310 @@
...
+  if (!supported) {
+    OPENSSL_PUT_ERROR(PKCS7, ERR_R_BIO_LIB);
+    return 0;
+  }
+
+  BIO_set_init(b, 1);
+
+  if (!EVP_CipherInit_ex(ctx->cipher, c, NULL, key, iv, enc)) {
+    return 0;
+  }
```
justsmth, line 307: The call to `BIO_set_init` should be after we successfully initialized `ctx->cipher`.

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,310 @@
...
+      }
+      break;
+    case BIO_CTRL_WPENDING:
+    case BIO_CTRL_PENDING:
+      // Return number of bytes left to process if we have anything buffered,
+      // else consult underlying BIO.
+      ret = ctx->buf_len;
+      if (ret <= 0) {
+        ret = BIO_ctrl(next, cmd, num, ptr);
+      }
```
justsmth, line 240: Should the return value be the sum of the two values? 
```
...
    ret = ctx->buf_len;
    int next_pending = BIO_ctrl(next, cmd, num, ptr);
    // sanity check next_pending
    ret += next_pending;
    break;
```

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,310 @@
...
+#include <openssl/pkcs7.h>
+#include <stdio.h>
+#include "../../fipsmodule/cipher/internal.h"
+#include "../internal.h"
+
+static int enc_write(BIO *h, const char *buf, int num);
+static int enc_read(BIO *h, char *buf, int size);
+static long enc_ctrl(BIO *h, int cmd, long arg1, void *arg2);
+static int enc_new(BIO *h);
+static int enc_free(BIO *data);
```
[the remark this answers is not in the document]
WillChilds-Klein: sure.

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,310 @@
...
+  BIO_set_init(b, 0);
+
+  return 1;
+}
+
+static int enc_read(BIO *b, char *out, int outl) {
+  GUARD_PTR(b);
+  GUARD_PTR(out);
+  BIO_ENC_CTX *ctx = BIO_get_data(b);
+  if (ctx == NULL || ctx->cipher == NULL || !ctx->ok) {
```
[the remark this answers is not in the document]
WillChilds-Klein: yeah, and same with `inl` in `enc_write`

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,310 @@
...
+      continue;
+    }
+    ctx->buf_len = 0;
+    ctx->buf_off = 0;
+    // |EVP_DecryptUpdate| may write up to cipher_block_size-1 more bytes than
+    // requested, so only read bytes we're sure we can decrypt in place.
+    int to_read = (int)sizeof(ctx->buf) - cipher_block_size + 1;
+    int bytes_read = BIO_read(next, read_buf, to_read);
+    if (bytes_read > 0) {
+      int to_decrypt = bytes_read;
```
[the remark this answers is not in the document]
WillChilds-Klein: yep, this was vestigial. removed.

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,310 @@
...
+      OPENSSL_memcpy(out_pos, &ctx->buf[ctx->buf_off], to_copy);
+      // Update buffer info and counters with number of bytes processed from our
+      // buffer.
+      ctx->buf_len -= to_copy;
+      ctx->buf_off += to_copy;
+      bytes_output += to_copy;
+      remaining -= to_copy;
+      continue;
+    }
+    ctx->buf_len = 0;
```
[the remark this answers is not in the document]
WillChilds-Klein: i don't think we need to as the loop is conditioned on it, and we only update `ctx->done` immediately before the loop conditional check.

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: no, this also handles the result of `EVP_DecryptFinal_ex` on L127.

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: yes.

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: agreed.

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: OpenSSL docs excerpted below, not particularly clear on the question 

> BIO_pending(), BIO_ctrl_pending(), BIO_wpending() and BIO_ctrl_wpending() return the amount of pending data. BIO_pending() and BIO_wpending() return negative value or 0 on error.

OpenSSL [has same behavior as us](https://github.com/openssl/openssl/blob/f30d6ba455e06572250e75132045eedde5d1daf0/crypto/evp/bio_enc.c#L327-L329), so i think we should keep it as-is for compatibility.

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: i don't think so. i want to error out here so we are made aware of a need to implement these operations if something depending on us needs them.

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: for flush, yes. for read/write, we check right after setting it.

crypto/pkcs7/bio/cipher.c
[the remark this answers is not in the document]
WillChilds-Klein: i don't think we need to. we check `ctx->ok` immediately after it's set, and don't modify `ctx->done` in this function.

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: yes, that makes sense.

crypto/pkcs7/bio/cipher.c
WillChilds-Klein: yes, that makes sense.

justsmth: Were you planning to add a block like this here as well?
```
      if (bytes_written < 0 && !BIO_should_retry(next)) {
        ctx->done = 1;
        ctx->ok = 0;
      }
```

WillChilds-Klein: ah, yep. thanks, fixed.

justsmth (approved):

crypto/pkcs7/bio/cipher.c
justsmth: Yeah, the concern for me is not performance so much as clarity/readability of the code.

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,315 @@
...
+    }
+  }
+  return bytes_output;
+}
+
+static int enc_flush(BIO *b, BIO *next, BIO_ENC_CTX *ctx) {
+  GUARD_PTR(b);
+  GUARD_PTR(next);
+  GUARD_PTR(ctx);
+  while (ctx->ok > 0 && (ctx->buf_len > 0 || !ctx->done)) {
```
samuel40791765, line 127: Nit: Since we're using `ctx->ok` as a boolean, the logic may be clearer if we use it as one.
```suggestion
  while (ctx->ok && (ctx->buf_len > 0 || !ctx->done)) {
```

crypto/pkcs7/bio/cipher.c
```diff
@@ -0,0 +1,315 @@
...
+  return bytes_consumed;
+}
+
+static long enc_ctrl(BIO *b, int cmd, long num, void *ptr) {
+  GUARD_PTR(b);
+  long ret = 1;
+
+  BIO_ENC_CTX *ctx = BIO_get_data(b);
+  BIO *next = BIO_next(b);
+  if (ctx == NULL) {
```
samuel40791765, line 199: Nit:
```suggestion
  if (ctx == NULL || next == NULL) {
```

samuel40791765 (approved):

PR #1845, opened 2024-09-11
add support for PEM_write_bio_PrivateKey_traditional

samuel40791765:
### Issues:
Resolves `CryptoAlg-2546`

### Description of changes: 
The main difference between `PEM_write_bio_PrivateKey_traditional` and the modern `PEM_write_bio_PrivateKey` is that `PEM_write_bio_PrivateKey` writes a PKCS#8 encoding, while the traditional one writes with DER encoding.
OpenSSL's `PEM_write_bio_PrivateKey_traditional` calls into `i2d_PrivateKey` which calls into another additional `old_priv_encode` EVP_PKEY_ASN1_METHOD hook. We also need to make the call into `i2d_PrivateKey`, but luckily for us BoringSSL removed the hook and let the call point to the corresponding `i2d_RSA/DSA/ECPrivateKey` methods instead. We can built on this directly to reimplement `PEM_write_bio_PrivateKey_traditional`.
* https://github.com/aws/aws-lc/blob/51d9a8d3b3270fc506fb039a1af0c6f3033f37b6/crypto/x509/i2d_pr.c#L65-L79
* OpenSSL
    * Usage of ASN1 hook: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/asn1/i2d_pr.c#L17-L33
    * Check for ASN1 hook: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/pem/pem_pkey.c#L105-L120

`PEM_read_bio_PrivateKey` already has [logic](https://github.com/aws/aws-lc/blob/51d9a8d3b3270fc506fb039a1af0c6f3033f37b6/crypto/pem/pem_pkey.c#L132-L141) to parse the "traditional form" of these PEM files. I've also added tests to verify.

### Call-outs:
N/A

### Testing:
Tests for each type writable/readable. A single test for DH, which shouldn't be writable.

### DIFF
--- a/crypto/pem/pem_pkey.c
+++ b/crypto/pem/pem_pkey.c
@@ -157,7 +157,6 @@ int PEM_write_bio_PrivateKey(BIO *bp, EVP_PKEY *x, const EVP_CIPHER *enc,
   return PEM_write_bio_PKCS8PrivateKey(bp, x, enc, (char *)kstr, klen, cb, u);
 }
 
-<<<<<<< HEAD
 EVP_PKEY *PEM_read_bio_Parameters(BIO *bio, EVP_PKEY **pkey) {
   if (bio == NULL) {
     OPENSSL_PUT_ERROR(PEM, ERR_R_PASSED_NULL_PARAMETER);
@@ -269,6 +268,11 @@ int PEM_write_bio_PrivateKey_traditional(BIO *bp, EVP_PKEY *x,
                                          const EVP_CIPHER *enc,
                                          unsigned char *kstr, int klen,
                                          pem_password_cb *cb, void *u) {
+  if (bp == NULL || x == NULL || x->ameth == NULL ||
+      x->ameth->pem_str == NULL) {
+    OPENSSL_PUT_ERROR(PEM, ERR_R_PASSED_NULL_PARAMETER);
+    return 0;
+  }
   char pem_str[80];
   BIO_snprintf(pem_str, 80, "%s PRIVATE KEY", x->ameth->pem_str);
   return PEM_ASN1_write_bio(i2d_PrivateKey_void, pem_str, bp, x, enc, kstr,
--- a/crypto/pem/pem_pkey.c
+++ b/crypto/pem/pem_pkey.c
@@ -157,6 +157,7 @@ int PEM_write_bio_PrivateKey(BIO *bp, EVP_PKEY *x, const EVP_CIPHER *enc,
   return PEM_write_bio_PKCS8PrivateKey(bp, x, enc, (char *)kstr, klen, cb, u);
 }
 
+<<<<<<< HEAD
 EVP_PKEY *PEM_read_bio_Parameters(BIO *bio, EVP_PKEY **pkey) {
   if (bio == NULL) {
     OPENSSL_PUT_ERROR(PEM, ERR_R_PASSED_NULL_PARAMETER);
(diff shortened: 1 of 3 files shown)


include/openssl/pem.h
```diff
@@ -489,6 +489,13 @@ OPENSSL_EXPORT EC_GROUP *PEM_read_bio_ECPKParameters(BIO *bio,
 OPENSSL_EXPORT int PEM_write_bio_ECPKParameters(BIO *out,
                                                 const EC_GROUP *group);
 
+// PEM_write_bio_PrivateKey_traditional calls |PEM_ASN1_write_bio| to write
+// out |x|'s private key in the "traditional" ASN1 format. Use
+// |PEM_write_bio_PrivateKey| instead.
+OPENSSL_EXPORT int PEM_write_bio_PrivateKey_traditional(
```
WillChilds-Klein, line 495: should this be `OPENSSL_DEPRECATED`?

include/openssl/pem.h
```diff
@@ -489,6 +489,13 @@ OPENSSL_EXPORT EC_GROUP *PEM_read_bio_ECPKParameters(BIO *bio,
 OPENSSL_EXPORT int PEM_write_bio_ECPKParameters(BIO *out,
                                                 const EC_GROUP *group);
 
+// PEM_write_bio_PrivateKey_traditional calls |PEM_ASN1_write_bio| to write
+// out |x|'s private key in the "traditional" ASN1 format. Use
```
WillChilds-Klein, line 493: what is the "traditional" ASN.1 format? is it specified in a particular RFC or PKCS document? does it impose restrictions on key algorithm/type?

crypto/pem/pem_pkey.c
```diff
@@ -156,6 +157,20 @@ int PEM_write_bio_PrivateKey(BIO *bp, EVP_PKEY *x, const EVP_CIPHER *enc,
...
+static int i2d_PrivateKey_void(const void *key, uint8_t **out) {
+  return i2d_PrivateKey((const EVP_PKEY *)key, out);
+}
+
+int PEM_write_bio_PrivateKey_traditional(BIO *bp, EVP_PKEY *x,
+                                         const EVP_CIPHER *enc,
+                                         unsigned char *kstr, int klen,
+                                         pem_password_cb *cb, void *u) {
+  char pem_str[80];
+  BIO_snprintf(pem_str, 80, "%s PRIVATE KEY", x->ameth->pem_str);
```
WillChilds-Klein, line 169: does `x->ameth->pem_str` contain any of the actual key material? we should `OPENSSL_cleanse` `pem_str` before returning if so.

include/openssl/pem.h
samuel40791765: I was weighing it, but the API naming seemed like OpenSSL was planning to maintain this long term. It's not deprecated in 3.x, so I decided to not mark it.

include/openssl/pem.h
samuel40791765: Not related to anything particular, they just used "traditional" as an alias for the ASN.1 encoding they were initially doing within the PEM files. They changed the original to use PKCS8 instead and created this API for the ASN1 encoding.
* https://github.com/openssl/openssl/commit/05dba8151bd418cdc111d62102aaf9f4e7bd2f3f

crypto/pem/pem_pkey.c
samuel40791765: No, it just points to a string representing the key type (added in https://github.com/aws/aws-lc/commit/3662fc13a4c4bccadb40a55218d4e71a491f1e5b).
But that does remind me that I should do an assertion on `x->ameth->pem_str` and `x->ameth` here.

WillChilds-Klein (dismissed):

crypto/pem/pem_pkey.c
```diff
@@ -260,6 +260,25 @@ int PEM_write_bio_Parameters(BIO *bio, EVP_PKEY *pkey) {
...
+                                         const EVP_CIPHER *enc,
+                                         unsigned char *kstr, int klen,
+                                         pem_password_cb *cb, void *u) {
+  if (bp == NULL || x == NULL || x->ameth == NULL ||
+      x->ameth->pem_str == NULL) {
+    OPENSSL_PUT_ERROR(PEM, ERR_R_PASSED_NULL_PARAMETER);
+    return 0;
+  }
+  char pem_str[80];
+  BIO_snprintf(pem_str, 80, "%s PRIVATE KEY", x->ameth->pem_str);
```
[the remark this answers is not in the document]
samuel40791765: No particular reason, this was taken from OpenSSL. They probably just chose a size large enough to fit the string with "pem_str" and `PRIVATE KEY`.

smittals2 (approved):

WillChilds-Klein (approved):

PR #1849, opened 2024-09-11
Add MLKEM768 Hybrid Groups to libssl

alexw91:
### Issues:
None

### Description of changes: 
Add support for `X25519MLKEM768` and `SecP256r1MLKEM768` from https://datatracker.ietf.org/doc/html/draft-kwiatkowski-tls-ecdhe-mlkem

### Call-outs:
A bug in `BadHybridKeyShareAcceptTest` and `BadHybridKeyShareFinishTest` was fixed. The for loop to corrupt each half of the Hybrid key share did not correctly calculate the bounds index after the 1st iteration. This bug would only *partially* corrupt the buffer on later iterations, and not corrupt the buffer at all if the 2nd KeyShare was smaller than the first (which is true for X25519MLKEM768). 

### Testing:

All existing unit tests for `X25519Kyber768Draft00` and `SecP256r1Kyber768Draft00` were updated or copied, and equivalent tests added for `X25519MLKEM768` and `SecP256r1MLKEM768` . 

Interoperability with BoringSSL for `X25519MLKEM768` has been confirmed to work with AWS-LC as both client and server. BoringSSL does not implement `SecP256r1MLKEM768`. 

```
./boringssl/build/tool/bssl s_server -curves X25519MLKEM768:X25519Kyber768Draft00:X25519 -accept 44330 
./aws-lc/build/tool/bssl s_client -curves X25519MLKEM768 -connect 127.0.0.1:44330
```

```
./aws-lc/build/tool/bssl s_server -curves X25519MLKEM768:X25519Kyber768Draft00:X25519 -accept 44330
./boringssl/build/tool/bssl s_client -curves X25519MLKEM768 -connect 127.0.0.1:44330
```


ssl/ssl_key_share.cc
```diff
@@ -677,14 +684,20 @@ CONSTEXPR_ARRAY NamedGroup kNamedGroups[] = {
...
+    {NID_SecP256r1MLKEM768, SSL_GROUP_SECP256R1_MLKEM768, "SecP256r1MLKEM768", ""},
+    {NID_X25519MLKEM768, SSL_GROUP_X25519_MLKEM768, "X25519MLKEM768", ""},
 };
 
 CONSTEXPR_ARRAY uint16_t kPQGroups[] = {
     SSL_GROUP_KYBER512_R3,
     SSL_GROUP_KYBER768_R3,
     SSL_GROUP_KYBER1024_R3,
+    SSL_GROUP_MLKEM768,
+    SSL_GROUP_MLKEM1024,
```
torben-hansen, line 696: Not actually used anywhere. Indifferent to whether to keep or not.

ssl/ssl_key_share.cc
```diff
@@ -695,13 +708,30 @@ CONSTEXPR_ARRAY HybridGroup kHybridGroups[] = {
...
+    {
+      SSL_GROUP_SECP256R1,         // component_group_ids[0]
+      SSL_GROUP_MLKEM768,          // component_group_ids[1]
+    },
+  },
+
+  {
+    SSL_GROUP_X25519_MLKEM768,     // group_id
+    {
+        // Note: MLKEM768 is sent first due to FIPS requirements.
```
torben-hansen, line 730: Indentation

ssl/ssl_test.cc
```diff
@@ -686,6 +727,38 @@ static const HybridGroupTest kHybridGroupTests[] = {
...
+    },
+  },
+  {
+    NID_X25519MLKEM768,
+    SSL_GROUP_X25519_MLKEM768,
+    X25519_KEYSHARE_SIZE + MLKEM768_PUBLIC_KEY_BYTES,
+    X25519_KEYSHARE_SIZE + MLKEM768_CIPHERTEXT_BYTES,
+    X25519_SECRET_SIZE + MLKEM768_SHARED_SECRET_LEN,
+    {
+        // MLKEM768 is sent first for X25519MLKEM768 for FIPS compliance
```
torben-hansen, line 752: indentations.

ssl/ssl_key_share.cc
alexw91: I have a slight preference to keep `SSL_GROUP_MLKEM1024` IANA code point, but I'm not against removing it. If we're already pulling in the `SSL_GROUP_MLKEM768` code point from [draft-connolly-tls-mlkem-key-agreement](https://datatracker.ietf.org/doc/html/draft-connolly-tls-mlkem-key-agreement.html) I think we should also pull in the remaining code points in that draft for consistency; and if that code point is defined it should be marked as being a PQ code point.

ssl/ssl_test.cc
alexw91: Fixed

torben-hansen (dismissed):

WillChilds-Klein:
Hm, I'm getting 404'd on the [description's linked draft RFC](https://www.ietf.org/archive/id/draft-kwiatkowski-tls-ecdhe-mlkem.html).

include/openssl/ssl.h
```diff
@@ -2675,13 +2675,21 @@ OPENSSL_EXPORT int SSL_set1_groups_list(SSL *ssl, const char *groups);
...
+#define SSL_GROUP_X25519_MLKEM768    0x11EC
+
 // PQ and hybrid group IDs are not yet standardized. Current IDs are driven by
 // community consensus and are defined at
 // https://github.com/open-quantum-safe/oqs-provider/blob/main/oqs-template/oqs-kem-info.md
 #define SSL_GROUP_KYBER512_R3 0x023A
 #define SSL_GROUP_KYBER768_R3 0x023C
 #define SSL_GROUP_KYBER1024_R3 0x023D
 
+// https://datatracker.ietf.org/doc/html/draft-connolly-tls-mlkem-key-agreement.html
```
andrewhop, line 2689: Why pull these values from this IETF instead of the values defined in oqs-kem-info.md like Kyber? Is this IETF more widely accepted as authoritative?

crypto/obj/objects.txt
```diff
@@ -134,8 +134,11 @@ secg-ellipticCurve 39		: sect571r1
  : ffdhe8192
 
 # PQ Group OIDs from OQS
+# https://github.com/open-quantum-safe/oqs-provider/blob/main/ALGORITHMS.md#oids
```
andrewhop, line 137: From this page:
> Along the same lines as the code points, X.509 OIDs may be subject to change prior to final standardization

Are there more official values that are not in 1.3.9999 "reserved" section?

ssl/ssl_test.cc
```diff
@@ -686,6 +727,38 @@ static const HybridGroupTest kHybridGroupTests[] = {
...
+    },
+  },
+  {
+    NID_X25519MLKEM768,
+    SSL_GROUP_X25519_MLKEM768,
+    X25519_KEYSHARE_SIZE + MLKEM768_PUBLIC_KEY_BYTES,
+    X25519_KEYSHARE_SIZE + MLKEM768_CIPHERTEXT_BYTES,
+    X25519_SECRET_SIZE + MLKEM768_SHARED_SECRET_LEN,
+    {
+      // MLKEM768 is sent first for X25519MLKEM768 for FIPS compliance
```
andrewhop, line 752: Why is the order of the key shares important for FIPS?

ssl/ssl_test.cc
```diff
@@ -1014,6 +1089,318 @@ static const HybridHandshakeTest kHybridHandshakeTests[] = {
...
+    "SecP256r1MLKEM768:X25519MLKEM768:x25519:prime256v1",
+    TLS1_3_VERSION,
+    "X25519MLKEM768:prime256v1:x25519",
+    TLS1_3_VERSION,
+    SSL_GROUP_X25519_MLKEM768,
+    true,
+  },
+
+  // Like the previous case, but the client's prioritization of ECC and PQ
+  // is inverted.
```
andrewhop, line 1231: Interesting, so if the client supports PQ, but doesn't put it first the server won't do a HRR to get the PQ key share? Does this follow s2n's behavior?

andrewhop:
> Interoperability with BoringSSL for X25519MLKEM768 has been confirmed to work with AWS-LC as both client and server. BoringSSL does not implement SecP256r1MLKEM768.

Can you add this as an automated test to our CI to ensure we don't break in the future. Have you tested this with s2n-tls?

ssl/ssl_key_share.cc
```diff
@@ -695,13 +708,30 @@ CONSTEXPR_ARRAY HybridGroup kHybridGroups[] = {
...
   {
     SSL_GROUP_X25519_KYBER768_DRAFT00,     // group_id
     {
       SSL_GROUP_X25519,               // component_group_ids[0]
       SSL_GROUP_KYBER768_R3,          // component_group_ids[1]
     },
+  },
+
+  {
+    SSL_GROUP_SECP256R1_MLKEM768,     // group_id
```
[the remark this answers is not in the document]
alexw91: If no supported group preference list is configured for the SSL connection, the default supported group list is used here: 
https://github.com/aws/aws-lc/blob/main/ssl/extensions.cc#L309-L320

include/openssl/ssl.h
alexw91: They are the same values. The [OQS codepoint for ML-KEM768 is `0x0768`](https://github.com/open-quantum-safe/oqs-provider/blob/main/ALGORITHMS.md?plain=1#L44) and `draft-connolly-tls-mlkem-key-agreement` also proposes [the same `0x0768` codepoint for the same ML-KEM768 algorithm](https://datatracker.ietf.org/doc/html/draft-connolly-tls-mlkem-key-agreement.html).

Generally, the draft RFC's define are considered more authoritative since only a few of the codepoints intended for standardization for internet-wide compatibility are turned into draft RFC's. Whereas LibOQS codepoints are defined for internal/experimental use that other libraries can use (but no long-term guarantees are made).

crypto/obj/objects.txt
alexw91: Not that I'm aware of. There doesn't appear to be official OID's for `X25519MLKEM768` yet, but I believe an OID is needed in order to successfully use these supported groups from Python. I forget the exact reason why, but @WillChilds-Klein should know?

ssl/ssl_test.cc
alexw91:
```
Both groups enable the derivation of TLS session keys using FIPS-approved schemes.
NIST's special publication 800-56Cr2 SP56C approves the usage of HKDF with two
distinct shared secrets, with the condition that the first one is computed by a FIPS-
approved key-establishment scheme. FIPS also requires a certified implementation of 
the scheme, which will remain more ubiqutous for secp256r1 in the coming years. For 
this reason we put the ML-KEM-768 shared secret first in X25519MLKEM768, and the 
secp256r1 shared secret first in SecP256r1MLKEM768.
```
https://datatracker.ietf.org/doc/html/draft-kwiatkowski-tls-ecdhe-mlkem-02#name-negotiated-groups

ssl/ssl_test.cc
alexw91: This is different than s2n's behavior, but matches the existing long term Openssl/Boringssl/AWS-LC/LibreSSL behavior to use the client's preference order no matter what by default.

s2n's negotiation logic is more complex "PQ > 1-RTT > ServerPreference" is used, where PQ algorithms are preferred over non-PQ, then algorithms negotiable in 1-RTT are preferred over 2-RTT, and finally the server-side preference order is used.

alexw91:
> > Interoperability with BoringSSL for X25519MLKEM768 has been confirmed to work with AWS-LC as both client and server. BoringSSL does not implement SecP256r1MLKEM768.
> 
> Can you add this as an automated test to our CI to ensure we don't break in the future. Have you tested this with s2n-tls?

If this was a small addition to an existing test, I'd be happy to, but creating a brand new AWS-LC to BoringSSL CI integration test seems out of scope for this PR. s2n-tls does not yet have support for either of these algorithms.

alexw91:
> Hm, I'm getting 404'd on the [description's linked draft RFC](https://www.ietf.org/archive/id/draft-kwiatkowski-tls-ecdhe-mlkem.html).

Fixed: https://datatracker.ietf.org/doc/html/draft-kwiatkowski-tls-ecdhe-mlkem

andrewhop (approved):

torben-hansen (approved):

ssl/ssl_key_share.cc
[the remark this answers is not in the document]
WillChilds-Klein: should we add the hybrid MLKEM groups to this list and prefer them over regular ECDH? that would enable PQ TLS by default for consumers who don't configure a preference list (i'm not sure how common this is among our libssl consumers).

crypto/obj/objects.txt
WillChilds-Klein: the code path is a little convoluted, but if i recall correctly it was:

1. python [calls](https://github.com/WillChilds-Klein/cpython/blob/bd3d31f380cd451a4ab6da5fbfde463fed95b5b5/Modules/_ssl.c#L4386) `OBJ_sn2nid`
2. `OBJ_sn2nid` [calls](https://github.com/aws/aws-lc/blob/9d21f38b8ea04a70a753626708e3de21d8d0b733/crypto/obj/obj.c#L265) `lh_ASN1_OBJET_retrieve(global_added_by_short_name, ...)`
3. that array only gets populated with members that have OIDs

i might be recalling 3. incorrectly, but the functions called by python assumed that all valid "curves" (overloaded term here) had valid OIDs.

WillChilds-Klein (approved):

PR #1854, opened 2024-09-16
support building on illumos systems

iliana:
### Issues:
n/a

### Description of changes: 
The team that works on [awslabs/tough](https://github.com/awslabs/tough) wants to support using the aws-lc-rs Rust crate (https://github.com/awslabs/tough/issues/824). As one of several downstream consumers of tough outside of AWS, we were asked for feedback. Oxide uses tough as part of our product which runs on the illumos operating system.

When linking Rust code using aws-lc-rs, the linker cannot find several `aws_lc_*` symbols:

```
Undefined            first referenced
 symbol                  in file
aws_lc_0_21_1_bn_gather5            /home/iliana/git/omicron/target/debug/deps/libaws_lc_sys-ff0ab07b18206e1f.rlib(bcm.c.o)
aws_lc_0_21_1_rsaz_1024_red2norm_avx2 /home/iliana/git/omicron/target/debug/deps/libaws_lc_sys-ff0ab07b18206e1f.rlib(bcm.c.o)
aws_lc_0_21_1_aesni_gcm_encrypt     /home/iliana/git/omicron/target/debug/deps/libaws_lc_sys-ff0ab07b18206e1f.rlib(bcm.c.o)
aws_lc_0_21_1_chacha20_poly1305_seal /home/iliana/git/omicron/target/debug/deps/libaws_lc_sys-ff0ab07b18206e1f.rlib(e_chacha20poly1305.c.o)
aws_lc_0_21_1_chacha20_poly1305_open /home/iliana/git/omicron/target/debug/deps/libaws_lc_sys-ff0ab07b18206e1f.rlib(e_chacha20poly1305.c.o)
[ ... this goes on for some hundreds of lines ... ]
```

`uname -p` on illumos systems returns `i386` on 64-bit machines. This resulted in assembly code not being linked into the AWS-LC library due to architecture misdetection. To determine the native instruction set, `isainfo -n` is used instead.

Once these cryptic errors were out of the way, we got a more normal linking error indicating some missing libraries; symbols that are in the libc on other platforms are in separate libraries on illumos.

I'm sending this PR on behalf of my coworker, @jclulow, but I can help handle any requested changes.

### Call-outs:
n/a

### Testing:
Tested on an illumos system via aws-lc-rs by patching the crate in Cargo.toml with:

```
[patch.crates-io.aws-lc-sys]
git = "https://github.com/oxidecomputer/aws-lc-rs"
branch = "illumos/aws-lc-sys/v0.21.1"
```

which updates the AWS-LC submodule to this commit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,7 +61,7 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
   endif()
 endif()
 
-if(CMAKE_HOST_SYSTEM_NAME STREQUAL "SunOS")
+if(CMAKE_HOST_SYSTEM_NAME STREQUAL "SunOS" AND NOT CMAKE_CROSSCOMPILING)
   # Determine if the host is running an illumos distribution:
   execute_process(COMMAND /usr/bin/uname -o OUTPUT_VARIABLE UNAME_O
     OUTPUT_STRIP_TRAILING_WHITESPACE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,28 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
   endif()
 endif()
 
+if(CMAKE_SYSTEM_NAME STREQUAL "SunOS")
+  # Determine if the host is running an illumos distribution:
+  execute_process(COMMAND /usr/bin/uname -o OUTPUT_VARIABLE UNAME_O
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+  if (UNAME_O STREQUAL "illumos")
+    set(ILLUMOS 1)
+  endif()
+
+  if (ILLUMOS)
+    #
+    # illumos systems require linking libsocket and libnsl to get various
+    # networking routines sometimes found in libc on other platforms:
+    #
+    if(NOT BUILD_SHARED_LIBS)
+      set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -lsocket -lnsl")
+    else()
+      set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -lsocket -lnsl")
+    endif()
+  endif()
+endif()
+
 # Tests and libssl both require the CXX language to be enabled. If a consumer
 # chooses to disable building the tests and libssl, do not enable CXX
 if(BUILD_TESTING OR BUILD_LIBSSL)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -61,6 +61,28 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
   endif()
 endif()
 
+if(CMAKE_SYSTEM_NAME STREQUAL "SunOS")
+  # Determine if the host is running an illumos distribution:
+  execute_process(COMMAND /usr/bin/uname -o OUTPUT_VARIABLE UNAME_O
```
justsmth, line 66: A few nits:
* Is `uname` always located at `/usr/bin/uname` on a "SunOS" host?
* Re: cross-compiling -- the target system could be "SunOS", but the build host system might be something different. Should we condition this on `CMAKE_HOST_SYSTEM_NAME STREQUAL "SunOS"` also being true? 
* Re: cross-compiling -- is there a use-case for the build host being "illumos" but the target is a different "SunOS" variant?

I don't consider these concerns to be blocking.

CMakeLists.txt
```diff
@@ -789,9 +811,20 @@ if(OPENSSL_NO_SSE2_FOR_TESTING)
...
-# Matching below is based on lower-case.
-string(TOLOWER "${CMAKE_SYSTEM_PROCESSOR}" CMAKE_SYSTEM_PROCESSOR_LOWER)
+if(ILLUMOS)
+  #
+  # CMAKE_SYSTEM_PROCESSOR unfortunately comes from the output of "uname -p",
+  # which on illumos systems emits "i386".  Instead, use the value from
+  # "isainfo -n", which prints "the name of the native instruction set used by
+  # portable applications"; e.g., "amd64".
+  #
+  execute_process(COMMAND /usr/bin/isainfo -n OUTPUT_VARIABLE
```
justsmth, line 821: Note: Reference for `isainfo`: https://illumos.org/man/1/isainfo

justsmth (dismissed):

CMakeLists.txt
iliana: @jclulow is going to answer these questions (he has far more historical context than I) but I'm pushing a commit to address (2), since it doesn't make a lot of sense to try and detect illumos via uname if you're running on Linux. Good catch!

CMakeLists.txt
```diff
@@ -61,16 +61,16 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
   endif()
 endif()
 
-if(CMAKE_SYSTEM_NAME STREQUAL "SunOS")
+if(CMAKE_HOST_SYSTEM_NAME STREQUAL "SunOS")
```
justsmth, line 64: It's seems that the logic below only applies when the host and target systems are the same. CMake sets a [`CMAKE_CROSSCOMPILING`](https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html) whenever the build has had `CMAKE_SYSTEM_NAME` manually set.

This seems like something that should be checked here:
```
if(CMAKE_HOST_SYSTEM_NAME STREQUAL "SunOS" AND NOT CMAKE_CROSSCOMPILING)
```

justsmth (approved):

justsmth:
Sorry, this didn't make it into the [AWS-LC v1.35.0](https://github.com/aws/aws-lc/releases/tag/v1.35.0) release. Hopefully, we'll put up another release soon. (We typically align new releases of [aws-lc-sys](https://crates.io/crates/aws-lc-sys) with the latest release of AWS-LC.)

iliana:
We at least have Cargo's `[patch.crates-io]`! :)

CMakeLists.txt
jclulow: * Yes, `/bin` has been a link to `/usr/bin` on illumos (and our ancestral Solaris) for at least 20 years.  It's important to get the system-provided `uname` here because for reasons that I will never understand sometimes people will ship the GNU `uname` as well, in a directory that ends up in `PATH` first on their system, and the output of `guname -o` is ... unreliable at best!  (It has often said **Solaris** on illumos systems, which is pretty much the opposite of helpful)
* Yes, your point about cross-compiling is well-taken!  What we really want is to be a first class citizen in CMake at this point, I think, so that we can have the framework itself work these things out without executing additional commands on the build host or confusing cross compilation, etc.  I have filed CMake issue [26302 first class support for illumos systems](https://gitlab.kitware.com/cmake/cmake/-/issues/26302) to try and get the ball rolling there, but I expect it will take quite some time to get that to the point where we can just rely on it instead.

Thanks for looking at all this, we really appreciate it!

WillChilds-Klein (approved):

andrewhop (approved):

PR #1857, opened 2024-09-17
Update Service Indicator to handle custom crypto through *_METHOD structs

smittals2:
### Issues:
`CryptoAlg-2633`

### Description of changes: 
The service indicator "update" functions were modified to check for the existence of custom crypto in RSA_METHOD and EC_KEY_METHOD structs. If custom crypto is utilized for a given operation, the service indicator should not be updated. The checks are conservative and do not update the service indicator for a signing operation for EC/RSA keys if either custom sign or sign_sig/sign_raw functionality is provided - regardless of whether these may be invoked. 

### Testing:
Tested both EC_KEY_METHOD and RSA_METHOD with custom crypto and ensured the service indicator was not updating. 

### DIFF
--- a/crypto/fipsmodule/service_indicator/service_indicator.c
+++ b/crypto/fipsmodule/service_indicator/service_indicator.c
@@ -221,26 +221,38 @@ static int is_md_fips_approved_for_verifying(int md_type, int pkey_type) {
 // or |eckey_method| fields for a given |RSA| or |EC_KEY| respectively. For
 // |RSA| keys, custom verify and sign functionality is supported. For |EC_KEY|
 // keys, only custom sign functionality is supported.
-// Returns one if no custom crypto was invoked and zero otherwise.
-static int no_custom_meth_invoked(const EVP_PKEY_CTX *ctx) {
+// Returns one if custom crypto was invoked and zero otherwise.
+static int custom_meth_invoked(const EVP_PKEY_CTX *ctx) {
   const int pkey_type = EVP_PKEY_id(ctx->pkey);
-  if(pkey_type == EVP_PKEY_RSA || pkey_type == EVP_PKEY_RSA_PSS) {
-    const RSA_METHOD *meth = ctx->pkey->pkey.rsa->meth;
-    // Must be either |EVP_PKEY_OP_VERIFY| or |EVP_PKEY_OP_SIGN|
-    if(ctx->operation == EVP_PKEY_OP_VERIFY) {
-      return meth->verify_raw ? 0 : 1;
-    } else { // EVP_PKEY_OP_SIGN
-      // There are cases where custom |sign| functionality may be set but not
-      // |sign_raw|. However, this check is more conservative and fails if
-      // custom functionality is provided for either function pointer.
-      return (meth->sign || meth->sign_raw) ? 0 : 1;
-    }
-  } else if (pkey_type == EVP_PKEY_EC) {
-    const EC_KEY_METHOD *meth = ctx->pkey->pkey.ec->eckey_method;
-    return (meth->sign || meth->sign_sig) ? 0 : 1;
-  }
+  switch (pkey_type) {
+    case EVP_PKEY_RSA:
+    case EVP_PKEY_RSA_PSS: {
+      const RSA_METHOD *meth = ctx->pkey->pkey.rsa->meth;
+      // Must be either |EVP_PKEY_OP_VERIFY| or |EVP_PKEY_OP_SIGN|
+      switch (ctx->operation) {
+        case EVP_PKEY_OP_VERIFY:
+          return meth->verify_raw ? 1 : 0;
 
-  return 0;
+        case EVP_PKEY_OP_SIGN:
+          // There are cases where custom |sign| functionality may be set but
+          // not |sign_raw|. This check is more conservative and fails if
+          // custom functionality is provided for either function pointer.
+          return (meth->sign || meth->sign_raw) ? 1 : 0;
+
+        default:
+          return 0;  // custom crypto can't be invoked for unsupported ops
+      }
+    }
+
+    case EVP_PKEY_EC: {
+      const EC_KEY_METHOD *meth = ctx->pkey->pkey.ec->eckey_method;
+      return (meth->sign || meth->sign_sig) ? 1 : 0;
+    }
+
+    default:
+      // custom crypto can't be invoked for unsupported key types
+      return 0;
+  }
 }
 
 static void evp_md_ctx_verify_service_indicator(const EVP_MD_CTX *ctx,
@@ -300,7 +312,7 @@ static void evp_md_ctx_verify_service_indicator(const EVP_MD_CTX *ctx,
     // custom operations from |pkey.rsa->meth| were invoked.
     if (md_ok(md_type, pkey_type) &&
         ((rsa_1024_ok && n_bits == 1024) || (n_bits >= 2048 && n_bits % 2 == 0))
-        && no_custom_meth_invoked(pctx)) {
+        && !custom_meth_invoked(pctx)) {
       FIPS_service_indicator_update_state();
     }
   } else if (pkey_type == EVP_PKEY_EC) {
@@ -308,7 +320,7 @@ static void evp_md_ctx_verify_service_indicator(const EVP_MD_CTX *ctx,
     // if custom operations from |pkey.ec->eckey_method| were invoked.
     int curve_nid = EC_GROUP_get_curve_name(pkey->pkey.ec->group);
     if (md_ok(md_type, pkey_type) && is_ec_fips_approved(curve_nid) &&
-        no_custom_meth_invoked(pctx)) {
+        !custom_meth_invoked(pctx)) {
       FIPS_service_indicator_update_state();
     }
   }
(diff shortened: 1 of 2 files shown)


crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -217,6 +217,39 @@ static int is_md_fips_approved_for_verifying(int md_type, int pkey_type) {
...
+// custom_meth_invoked checks whether custom crypto was invoked in the |meth|
+// or |eckey_method| fields for a given |RSA| or |EC_KEY| respectively. For
+// |RSA| keys, custom verify and sign functionality is supported. For |EC_KEY|
+// keys, only custom sign functionality is supported.
+// Returns one if no custom crypto was invoked and zero otherwise.
+static int no_custom_meth_invoked(const EVP_PKEY_CTX *ctx) {
+  const int pkey_type = EVP_PKEY_id(ctx->pkey);
+  if(pkey_type == EVP_PKEY_RSA || pkey_type == EVP_PKEY_RSA_PSS) {
+    const RSA_METHOD *meth = ctx->pkey->pkey.rsa->meth;
+    if (meth == NULL) {
```
nebeid, line 229: `meth` is always non-NULL even if no custom methods are defined?

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -2876,6 +2936,120 @@ TEST_P(ECDSAServiceIndicatorTest, ManualECDSASignVerify) {
...
+  bssl::UniquePtr<EVP_PKEY> pkey(EVP_PKEY_new());
+  bssl::ScopedEVP_MD_CTX md_ctx;
+  ASSERT_TRUE(eckey);
+  ASSERT_TRUE(EC_KEY_set_group(eckey.get(), group));
+
+  EC_KEY_METHOD *meth = EC_KEY_METHOD_new(NULL);
+  EC_KEY_METHOD_set_sign(meth, ecdsa_sign, NULL, NULL);
+  EC_KEY_set_method(eckey.get(), meth);
+
+  // Generate a generic EC key.
```
nebeid, line 2988: what does `generic` mean here? It's based on the group, right?

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -2876,6 +2936,120 @@ TEST_P(ECDSAServiceIndicatorTest, ManualECDSASignVerify) {
...
+  // |EVP_DigestSignFinal| should not return an approval check because no crypto
+  // is being done when |nullptr| is given as the |out_sig| field.
+  size_t max_sig_len;
+  CALL_SERVICE_AND_CHECK_APPROVED(approved,
+                                  ASSERT_TRUE(EVP_DigestSignFinal(md_ctx.get(),
+                                                                  nullptr,
+                                                                  &max_sig_len)));
+  EXPECT_EQ(approved, AWSLC_NOT_APPROVED);
+  std::vector<uint8_t> signature(max_sig_len);
+  // The second call performs the actual operation.
```
nebeid, line 3016: I think the comments before or this one can be extended to mention we expect it not to be approved because of the custom signature.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -2876,6 +2936,120 @@ TEST_P(ECDSAServiceIndicatorTest, ManualECDSASignVerify) {
...
+  size_t sig_len = max_sig_len;
+  CALL_SERVICE_AND_CHECK_APPROVED(approved,
+                                  ASSERT_TRUE(EVP_DigestSignFinal(md_ctx.get(),
+                                                                  signature.data(),
+                                                                  &sig_len)));
+  ASSERT_STREQ(static_cast<const char*>(EC_KEY_get_ex_data(eckey.get(), 0)), "ecdsa_sign");
```
nebeid, line 3022: Why the custom EC sign is well defined and checked for the data it generates while this is not the case for RSA test? Is it more tedious for RSA?

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -2876,6 +2936,120 @@ TEST_P(ECDSAServiceIndicatorTest, ManualECDSASignVerify) {
...
+
+  ASSERT_LE(sig_len, signature.size());
+  EXPECT_EQ(approved, AWSLC_NOT_APPROVED);
+
+  // Test using the one-shot |EVP_DigestSign| function for approval.
```
nebeid, line 3027: Same observation about what we expect the indicator to be.

crypto/fipsmodule/service_indicator/service_indicator.c
smittals2: yep you are correct, added the check as force of habit

crypto/fipsmodule/service_indicator/service_indicator_test.cc
smittals2: Yes it is more complex to emulate the signing behavior for RSA keys. Even more so because we'd have to define both the sign and sign_raw functionalities (since RSA_PSS keys directly call sign_raw). It shouldn't matter either way, all we care about is the existence of custom function pointers (sort of hand waving what actually happens in those custom pointers however).

nebeid (dismissed):

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -217,6 +217,32 @@ static int is_md_fips_approved_for_verifying(int md_type, int pkey_type) {
   }
 }
 
+// custom_meth_invoked checks whether custom crypto was invoked in the |meth|
+// or |eckey_method| fields for a given |RSA| or |EC_KEY| respectively. For
+// |RSA| keys, custom verify and sign functionality is supported. For |EC_KEY|
+// keys, only custom sign functionality is supported.
+// Returns one if no custom crypto was invoked and zero otherwise.
+static int no_custom_meth_invoked(const EVP_PKEY_CTX *ctx) {
```
justsmth, line 225: * This name is a little confusing -- It's stated in the negative, which is harder to reason about.
* The default for both [`RSA_METHOD`](https://github.com/aws/aws-lc/blob/f89c9bec9aea680839ae10aef2d9ac627a9c08c3/crypto/fipsmodule/rsa/rsa_impl.c#L1299) and [`EC_KEY_METHOD`](https://github.com/aws/aws-lc/blob/f89c9bec9aea680839ae10aef2d9ac627a9c08c3/crypto/fipsmodule/ec/ec_key.c#L572) is an all-zero struct.  To better future proof this, can we just compare it to this "default" value, and if there have been any changes then we assume this operation was not FIPS compliant?

smittals2: Changed the name. Second point discussed offline.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -217,6 +217,44 @@ static int is_md_fips_approved_for_verifying(int md_type, int pkey_type) {
...
+
+        default:
+          return 0;  // custom crypto can't be invoked for unsupported ops
+      }
+    }
+
+    case EVP_PKEY_EC: {
+      const EC_KEY_METHOD *meth = ctx->pkey->pkey.ec->eckey_method;
+      return (meth->sign || meth->sign_sig) ? 1 : 0;
+    }
```
justsmth, line 250: This should only (potentially) return 1 for an `EVP_PKEY_EC` if the operation is `EVP_PKEY_OP_SIGN`:
```
    case EVP_PKEY_EC: {
      const EC_KEY_METHOD *meth = ctx->pkey->pkey.ec->eckey_method;
      // Only custom sign functionality is supported.
      if (ctx->operation == EVP_PKEY_OP_SIGN) {
        return (meth->sign || meth->sign_sig) ? 1 : 0;
      }
      return 0;
    }
```

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -217,6 +217,44 @@ static int is_md_fips_approved_for_verifying(int md_type, int pkey_type) {
...
+        case EVP_PKEY_OP_SIGN:
+          // There are cases where custom |sign| functionality may be set but
+          // not |sign_raw|. This check is more conservative and fails if
+          // custom functionality is provided for either function pointer.
+          return (meth->sign || meth->sign_raw) ? 1 : 0;
+
+        default:
+          return 0;  // custom crypto can't be invoked for unsupported ops
+      }
+    }
```
justsmth, line 245: NP: I like the `switch` statement at the top level, but `switch` statements can be confusing when nested: 
```
    case EVP_PKEY_RSA:
    case EVP_PKEY_RSA_PSS: {
      const RSA_METHOD *meth = ctx->pkey->pkey.rsa->meth;
      // Must be either |EVP_PKEY_OP_VERIFY| or |EVP_PKEY_OP_SIGN|
      if (ctx->operation == EVP_PKEY_OP_VERIFY) {
        return meth->verify_raw ? 1 : 0;
      }
      if(ctx->operation == EVP_PKEY_OP_SIGN) {
        // There are cases where custom |sign| functionality may be set but
        // not |sign_raw|. This check is more conservative and fails if
        // custom functionality is provided for either function pointer.
        return (meth->sign || meth->sign_raw) ? 1 : 0;
      }
      return 0;  // custom crypto can't be invoked for unsupported ops
    }
```

justsmth (approved):

nebeid (approved):

PR #1859, opened 2024-09-17
Add return checks on SHA3 functions in ML-KEM

manastasova:
### Issues:
Resolves #P155314914

### Description of changes: 
Add checks on the return values of `{SHA3,SHAKE}_{Init,Update,Final}`  and `SHAKE256` when consumed by ML-KEM. 

### Call-outs:
Coverity Defect is generated when `SHA3_Update` return code is not checked. Checks on the return values are added to all FIPS202 function calls.

### Testing:
`./crypto/crypto_test`


crypto/fipsmodule/ml_kem/ml_kem_ref/symmetric-shake.c
```diff
@@ -25,8 +25,12 @@ void kyber_shake128_absorb(KECCAK1600_CTX *ctx,
...
   extseed[KYBER_SYMBYTES+1] = y;
 
-  SHAKE_Init(ctx, SHAKE128_BLOCKSIZE);
-  SHA3_Update(ctx, extseed, sizeof(extseed));
+  if (SHAKE_Init(ctx, SHAKE128_BLOCKSIZE) == 0) {
+    return;
+  }
+  if (SHA3_Update(ctx, extseed, sizeof(extseed)) == 0) {
+    return;
+  }
```
andrewhop, line 33: While this is checking the return code now, is this safe to do? If SHAKE_Init fails can ML-KEM continue? Options I see:
1. Make SHAKE_Init void (if it can be void and never fails)
2. Make kyber_shake128_absorb non-void and bubble up the failure
3. If SHAKE_init fails abort

My concern is if kyber_shake128_absorb silently fails and continues AWS-LC will be in an inconsistent state and generate an invalid result but callers will think AWS-LC succeeded.

manastasova: From my understanding, `SHA3_Update` is the one creating the issue. 

1. We _could_ make it void, however, it's success/failure is based on the execution of `SHA3_Absorb`, which returns the number of bytes still to be processed. Called from `SHA3_Update`, `SHA3_Absorb` is called to process the leftover data in the buffer (if any from previous calls to `SHA3_Update`). If the data is enough to fill at least an entire block, `SHA3_Absorb` will be called first on a single block. In case the block is not completely processed, the return value of `SHA3_Absorb` is indicating an error. This case, indeed, should never happen (even if `Keccak` is not executed correctly, it is a void function, so the incorrect data will be stored independently, indicating the _success_ of `SHA3_Absorb`, thus, of `SHA3_Update`). So, we could make `SHA3_Update` a void function, however, we should not check the returns that are called from it either, and we cannot make `SHA3_Absorb` void.
2. In the ML-KEM there are no checks for the xof success. 
3. We could abort, however, would that not impact the flow when two parties are executing the kem?

From my understanding, the tests on SHA3, SHAKE are checking the correctness of SHA3/SHAKE functionality. Since we only use the SHA3/SHAKE tested functions, we _could_ assume they are executing correctly if SHA3/SHAKE tests pass.

dkostic (approved):

andrewhop (approved):

PR #1860, opened 2024-09-18
Create mutable EC_GROUP API for OpenSSL compatibility

samuel40791765:
### Issues:
Addresses `CryptoAlg-2547`

### Description of changes: 
Further investigation led us to discover that Ruby [depends](https://github.com/ruby/ruby/blob/ruby_3_1/ext/openssl/lib/openssl/pkey.rb#L296-L311) on mutable `EC_GROUP`s from OpenSSL. These were made immutable with the introduction of our default static `EC_GROUP` curves, which have been beneficial for us and our existing consumers. This was done by making the mutable portions of `EC_GROUP`s no-ops (`EC_GROUP_set_asn1_flag` and `EC_GROUP_set_point_conversion_form`), which OpenSSL couldn't do due to their more complex use case.
* https://github.com/aws/aws-lc/commit/51073ce05587f413d46f908407b47e884d6d325e
* https://github.com/aws/aws-lc/commit/e166375a20c06155e149c87d4f176f9770abc29a

Ruby (and possibly other dynamic languages) mutates the data being stored in `EC_GROUP` and depends on it for subsequent operations. Since our default groups are static and immutable, this causes issues when integrating with Ruby. We don't want to sacrifice the benefits of static groups and reversing the optional call to `EC_GROUP_free` is a dangerously silent memory leak. 
We've decided to introduce a special API for mutable `EC_GROUP`s for our default curves. This gives us a path forward for Ruby and keeps the benefits of static curves for existing consumers. Custom curves are already dynamically allocated so they don't need another special API, but the additional complexities for maintaining the underlying `generator` warrant another commit separate from this.

### Call-outs:
1. Custom curves aren't mutable just yet, but will be in another PR.
2. It would be nice to get additional verification that the bookkeeping in `EC_GROUP_dup` and `EC_GROUP_cmp` were done right.

### Testing:
Extended against tests using existing `EC_GROUP`s and conversion formats. These were focused on retrieving the `conv_form` from the `EC_GROUP` to use.

### DIFF
--- a/crypto/fipsmodule/ec/ec.c
+++ b/crypto/fipsmodule/ec/ec.c
@@ -144,6 +144,8 @@ DEFINE_METHOD_FUNCTION(EC_GROUP, EC_group_p224) {
   ec_group_set_a_minus3(out);
   out->has_order = 1;
   out->field_greater_than_order = 1;
+  out->conv_form = POINT_CONVERSION_UNCOMPRESSED;
+  out->mutable_ec_group = 0;
 }
 
 DEFINE_METHOD_FUNCTION(EC_GROUP, EC_group_p256) {
@@ -176,6 +178,8 @@ DEFINE_METHOD_FUNCTION(EC_GROUP, EC_group_p256) {
   ec_group_set_a_minus3(out);
   out->has_order = 1;
   out->field_greater_than_order = 1;
+  out->conv_form = POINT_CONVERSION_UNCOMPRESSED;
+  out->mutable_ec_group = 0;
 }
 
 DEFINE_METHOD_FUNCTION(EC_GROUP, EC_group_p384) {
@@ -206,6 +210,8 @@ DEFINE_METHOD_FUNCTION(EC_GROUP, EC_group_p384) {
   ec_group_set_a_minus3(out);
   out->has_order = 1;
   out->field_greater_than_order = 1;
+  out->conv_form = POINT_CONVERSION_UNCOMPRESSED;
+  out->mutable_ec_group = 0;
 }
 
 DEFINE_METHOD_FUNCTION(EC_GROUP, EC_group_p521) {
@@ -240,6 +246,8 @@ DEFINE_METHOD_FUNCTION(EC_GROUP, EC_group_p521) {
   ec_group_set_a_minus3(out);
   out->has_order = 1;
   out->field_greater_than_order = 1;
+  out->conv_form = POINT_CONVERSION_UNCOMPRESSED;
+  out->mutable_ec_group = 0;
 }
 
 DEFINE_METHOD_FUNCTION(EC_GROUP, EC_group_secp256k1) {
@@ -266,6 +274,8 @@ DEFINE_METHOD_FUNCTION(EC_GROUP, EC_group_secp256k1) {
   ec_group_set_a_zero(out);
   out->has_order = 1;
   out->field_greater_than_order = 1;
+  out->conv_form = POINT_CONVERSION_UNCOMPRESSED;
+  out->mutable_ec_group = 0;
 }
 
 EC_GROUP *EC_GROUP_new_curve_GFp(const BIGNUM *p, const BIGNUM *a,
@@ -300,6 +310,12 @@ EC_GROUP *EC_GROUP_new_curve_GFp(const BIGNUM *p, const BIGNUM *a,
     return NULL;
   }
   ret->references = 1;
+  // TODO: Treat custom curves as "immutable" for now. There's the possibility
+  // that we'll need dynamically allocated custom curve |EC_GROUP|s. But there
+  // are additional complexities around untangling the expectations already set
+  // for |EC_GROUP_new_curve_GFp| and |EC_GROUP_set_generator|.
+  // Note: See commit cb16f17b36d9ee9528a06d2e520374a4f177af4d.
+  ret->mutable_ec_group = 0;
   ret->meth = EC_GFp_mont_method();
   bn_mont_ctx_init(&ret->field);
   bn_mont_ctx_init(&ret->order);
@@ -391,12 +407,43 @@ EC_GROUP *EC_GROUP_new_by_curve_name(int nid) {
   }
 }
 
+EC_GROUP *EC_GROUP_new_by_curve_name_mutable(int nid) {
+  EC_GROUP *ret = NULL;
+  switch (nid) {
+    case NID_secp224r1:
+      ret = (EC_GROUP *)OPENSSL_memdup(EC_group_p224(), sizeof(EC_GROUP));
+      break;
+    case NID_X9_62_prime256v1:
+      ret = (EC_GROUP *)OPENSSL_memdup(EC_group_p256(), sizeof(EC_GROUP));
+      break;
+    case NID_secp384r1:
+      ret = (EC_GROUP *)OPENSSL_memdup(EC_group_p384(), sizeof(EC_GROUP));
+      break;
+    case NID_secp521r1:
+      ret = (EC_GROUP *)OPENSSL_memdup(EC_group_p521(), sizeof(EC_GROUP));
+      break;
+    case NID_secp256k1:
+      ret = (EC_GROUP *)OPENSSL_memdup(EC_group_secp256k1(), sizeof(EC_GROUP));
+      break;
+    default:
+      OPENSSL_PUT_ERROR(EC, EC_R_UNKNOWN_GROUP);
+      return NULL;
+  }
+  ret->mutable_ec_group = 1;
+  return ret;
+}
+
 void EC_GROUP_free(EC_GROUP *group) {
-  if (group == NULL ||
-      // Built-in curves are static.
-      group->curve_name != NID_undef ||
-      !CRYPTO_refcount_dec_and_test_zero(&group->references)) {
+  if (group == NULL) {
     return;
+  }
+
+  if (!group->mutable_ec_group) {
+    if (group->curve_name != NID_undef ||
+        !CRYPTO_refcount_dec_and_test_zero(&group->references)) {
+      // Built-in curves are static.
+      return;
+    }
   }
 
   bn_mont_ctx_cleanup(&group->order);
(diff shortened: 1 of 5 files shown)


crypto/fipsmodule/ec/ec.c
```diff
@@ -1095,6 +1180,15 @@ void EC_GROUP_set_point_conversion_form(EC_GROUP *group,
       form != POINT_CONVERSION_COMPRESSED) {
     abort();
   }
+  // Only dynamically allocated groups are mutable.
```
WillChilds-Klein, line 1183: how does this comment relate to the conversion form you're setting two lines below?

crypto/fipsmodule/ec/ec_key.c
```diff
@@ -280,6 +280,9 @@ point_conversion_form_t EC_KEY_get_conv_form(const EC_KEY *key) {
 
 void EC_KEY_set_conv_form(EC_KEY *key, point_conversion_form_t cform) {
   key->conv_form = cform;
+  if (key->group->mutable_ec_group) {
```
WillChilds-Klein, line 283: minor nit: null check on `key` and `key->group`

crypto/fipsmodule/ec/internal.h
```diff
@@ -650,6 +650,17 @@ struct ec_group_st {
   // otherwise.
   int field_greater_than_order;
 
+  // conv_form represents the encoding format of the elliptic curve point.
+  // The default is |POINT_CONVERSION_UNCOMPRESSED|. This is not changed unless
+  // the user allocates the |EC_GROUP| with |EC_GROUP_new_by_curve_name_mutable|
+  // and customizes |conv_form| with |EC_GROUP_set_point_conversion_form|.
```
WillChilds-Klein, line 656: do we support anything other than `POINT_CONVERSION_UNCOMPRESSED`?

crypto/fipsmodule/ec/ec.c
```diff
@@ -391,46 +407,115 @@ EC_GROUP *EC_GROUP_new_by_curve_name(int nid) {
...
+    return NULL;
   }
-  if (a->curve_name != NID_undef) {
-    // Built-in curves may be compared by curve name alone.
-    return 0;
+  ret->has_order = a->has_order;
+  ret->a_is_minus3 = a->a_is_minus3;
+  ret->curve_name = a->curve_name;
+  ret->conv_form = a->conv_form;
+  ret->field_greater_than_order = a->field_greater_than_order;
```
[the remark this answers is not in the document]
samuel40791765: After staring more closely into the `EC_GROUP` structure, it seems like only `generator->group`  is a pointer redirection (and it's directing to itself). I think we can do a shallow copy here, all the tests pass with this behavior.
* https://github.com/aws/aws-lc/blob/2835116571e8607861ba8dd0be2676f2c5cfaaab/crypto/fipsmodule/ec/internal.h#L621-L654

crypto/fipsmodule/ec/ec.c
```diff
@@ -391,46 +407,115 @@ EC_GROUP *EC_GROUP_new_by_curve_name(int nid) {
...
+  if (!BN_MONT_CTX_copy(&ret->field, &a->field) ||
+      !BN_MONT_CTX_copy(&ret->order, &a->order)) {
+    EC_GROUP_free(ret);
+    ret = NULL;
   }
+  return ret;
+}
 
+int EC_GROUP_cmp(const EC_GROUP *a, const EC_GROUP *b, BN_CTX *ignored) {
+  if (!a->mutable_ec_group) {
```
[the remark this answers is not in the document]
samuel40791765: Absolutely. Great catch, I forgot to update this while I was piecing things together.

crypto/fipsmodule/ec/ec.c
```diff
@@ -391,46 +407,103 @@ EC_GROUP *EC_GROUP_new_by_curve_name(int nid) {
...
+int EC_GROUP_cmp(const EC_GROUP *a, const EC_GROUP *b, BN_CTX *ignored) {
+  if (!a->mutable_ec_group && !b->mutable_ec_group) {
+    // Note this function returns 0 if equal and non-zero otherwise.
+    if (a == b) {
+      return 0;
+    }
+    if (a->curve_name != b->curve_name) {
+      return 1;
+    }
+    if (a->curve_name != NID_undef) {
```
WillChilds-Klein, line 502: nit: so, the idea here is that if both static/built-in curves have `curve_name == NID_undef`, we want to fall through to comparing individual parameters? this a bit confusing, you may want to either add a comment explaining the intent or re-write L449-504 to clarify.

crypto/fipsmodule/ec/ec_test.cc
```diff
@@ -1855,19 +1887,63 @@ TEST_P(ECCurveTest, EncodeInfinity) {
...
+static std::vector<CurveParam> AllCurves() {
   const size_t num_curves = EC_get_builtin_curves(nullptr, 0);
   std::vector<EC_builtin_curve> curves(num_curves);
   EC_get_builtin_curves(curves.data(), num_curves);
-  std::vector<int> nids;
-  for (const auto& curve : curves) {
-    nids.push_back(curve.nid);
+  std::vector<CurveParam> nids;
+  for (const auto &curve : curves) {
+    CurveParam curve_param{};
```
WillChilds-Klein, line 1929: GCC 4 [doesn't seem to like](https://us-west-2.codebuild.aws.amazon.com/project/eyJlbmNyeXB0ZWREYXRhIjoiWk5IUWJGRGxBcTJ2Mkp4WGF3dnBwYjc5V0ZZYSt5SVVGbkwvODkydTNTaVQ2V2FMN3hwa0tjSWNFemw2QWtCWW5welFWV3lpRFpKVitwejgvelhpRWh3NDNqcWhKalpPYW9hL2tLMDlJSDFPT1NkNyIsIml2UGFyYW1ldGVyU3BlYyI6Ilp4VXRNQXFGM1BZYVlaRkIiLCJtYXRlcmlhbFNldFNlcmlhbCI6MX0%3D/build/fd193587-2582-4ae9-93cc-765b786fd5e2) lack of initialization here.

crypto/fipsmodule/ec/ec.c
```diff
@@ -391,46 +407,103 @@ EC_GROUP *EC_GROUP_new_by_curve_name(int nid) {
...
-  if (group == NULL ||
-      // Built-in curves are static.
-      group->curve_name != NID_undef ||
-      !CRYPTO_refcount_dec_and_test_zero(&group->references)) {
+  if (group == NULL) {
     return;
   }
 
+  if (!group->mutable_ec_group) {
+    if (group->curve_name != NID_undef ||
```
WillChilds-Klein, line 445: are we sure that custom curves will always have `curve_name == NID_undef`? because we currently treat all custom curves as "immutable", if the caller sets some other NID on a custom curve, we'd decrement the refcount and return instead of free'ing. is that what we want?

crypto/fipsmodule/ec/ec.c
samuel40791765: I see what the confusion is. I think it'd make sense to move this comment higher and re-clarify the intent behind returning 0 here. As you were saying, the fall through to comparing custom curve contents is what we're trying to convey.

crypto/fipsmodule/ec/ec.c
samuel40791765: From what I've observed for our `EC_GROUP` implementations it does seem like this is the main differentiation between custom and named curves. We don't expose any methods to set an NID for custom curves, although that may change if we are ever forced to implement `EC_GROUP_set_curve_name`. 

It does make me feel a bit more at ease, when looking at [OpenSSL's logic ](https://github.com/openssl/openssl/blob/2f362e99a1178263c7102474f0190836166f416d/crypto/ec/ec_lib.c#L483-L490)for the function though.
```
void EC_GROUP_set_curve_name(EC_GROUP *group, int nid)
{
    group->curve_name = nid;
    group->asn1_flag =
        (nid != NID_undef)
        ? OPENSSL_EC_NAMED_CURVE
        : OPENSSL_EC_EXPLICIT_CURVE;
}
```
Only named curves are expected to have NIDs, explicit/custom curves follow the same logic of having `NID_undef` in OpenSSL as well.

crypto/fipsmodule/ec/ec.c
```diff
@@ -391,46 +407,105 @@ EC_GROUP *EC_GROUP_new_by_curve_name(int nid) {
...
+  if (!BN_MONT_CTX_copy(&ret->field, &a->field) ||
+      !BN_MONT_CTX_copy(&ret->order, &a->order)) {
+    EC_GROUP_free(ret);
+    ret = NULL;
   }
+  return ret;
+}
 
+int EC_GROUP_cmp(const EC_GROUP *a, const EC_GROUP *b, BN_CTX *ignored) {
+  if (!a->mutable_ec_group && !b->mutable_ec_group) {
```
dkostic, line 494: not sure I understand why you need all this new code, wouldn't the original code be sufficient?

crypto/fipsmodule/ec/ec_key.c
```diff
@@ -279,7 +279,14 @@ point_conversion_form_t EC_KEY_get_conv_form(const EC_KEY *key) {
 }
 
 void EC_KEY_set_conv_form(EC_KEY *key, point_conversion_form_t cform) {
+  if (key == NULL || key->group == NULL) {
+    OPENSSL_PUT_ERROR(EC, ERR_R_PASSED_NULL_PARAMETER);
+    return;
+  }
   key->conv_form = cform;
+  if (key->group->mutable_ec_group) {
```
dkostic, line 287: why do we set the conversion form in the parent EC_GROUP as well? Can that group be associated with other keys?

samuel40791765: Great question, I did this since OpenSSL was doing this in [their implementation](https://github.com/openssl/openssl/blob/12d14de641c299ec080edc521f7080acc44e366f/crypto/ec/ec_key.c#L916-L921) of `EC_KEY_set_conv_form`. The `EC_GROUP` associated with `EC_KEY` [is duplicated](https://github.com/aws/aws-lc/blob/ad93747dbb36d5c18c1e640e000e0d15d40ab513/crypto/fipsmodule/ec/ec_key.c#L208-L222) with `EC_GROUP_dup`, so I think the intention was that it wouldn't effect other keys.

I should move the check for `key->group` further below however.

crypto/fipsmodule/ec/ec.c
samuel40791765: That's true. I initially thought we should be comparing the entire curves structure for dynamically allocated named curves. But now that you mention it, it doesn't seem necessary, none of the parts of the comparison criteria are really changeable for named curves.
I'll remove this.

crypto/fipsmodule/ec/ec.c
```diff
@@ -391,46 +407,105 @@ EC_GROUP *EC_GROUP_new_by_curve_name(int nid) {
...
+    // Groups are logically immutable (but for |EC_GROUP_set_generator| which
+    // must be called early on), so we simply take a reference.
+    EC_GROUP *group = (EC_GROUP *)a;
+    CRYPTO_refcount_inc(&group->references);
+    return group;
   }
-  if (a->curve_name != b->curve_name) {
-    return 1;
+
+  // Directly duplicate the |EC_GROUP| if it was dynamically allocated. We do a
```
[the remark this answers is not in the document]
samuel40791765: We can and will :) The subsequent PR for mutable custom curve `EC_GROUP`s will clean up the reference count above and use the new code below. ([here](https://github.com/aws/aws-lc/pull/1881/files#diff-00e1807dece216ad0f398e399fdc7cae5e060f9da0e157a43cbb92bc9e322f05R450)).

I thought directly returning the pointer for static curves was worth keeping, so we didn't need to all the additional copying.

crypto/fipsmodule/ec/ec_key.c
dkostic: not sure I understand, more than one key can be associated with an `EC_GROUP`, right?

samuel40791765: Yes `EC_GROUP`s are used to generate multiple `EC_KEY`s based on the group parameters. However, each `EC_KEY` has it's own duplicated reference to `EC_GROUP` assigned to itself. `EC_KEY`'s pointer reference to `EC_GROUP` is now different from the original `EC_GROUP` pointer that was used. 
The contents of `EC_KEY->EC_GROUP` can be mutated without effecting the original pointer. The original `EC_GROUP` can be continued to be used to generate new `EC_KEY`s. 

It's a mess since we now have two structures saving the `conv_form`.. But now that we're doing this, I believe the intention was to allow `EC_KEY` and it's own `EC_GROUP` pointer's `conv_form` type to align. The original `EC_GROUP` pointer that was used to generate `EC_KEY` shouldn't be effected.

samuel40791765: Do you think adding a test for this would help? I'm open to doing so if it helps clarify the pointer boundaries :)

dkostic: oh ok, I got it now, looks good to me.

dkostic (approved):

crypto/fipsmodule/ec/ec_key.c
nebeid: Thanks, Samuel, for offering to add a test. I didn't review the PR, but I think adding a test that would explain to us (when we ask ourselves later) what is done here is a great idea, if not too much work. Could be in another PR if you prefer.

WillChilds-Klein (approved):

crypto/fipsmodule/ec/ec_key.c
samuel40791765: No problem, I have subsequent PR for implementing this behavior for custom curves. I'll add the test in that PR since the same concept applies: https://github.com/aws/aws-lc/pull/1881

PR #1868, opened 2024-09-20
ML-KEM encapsulation key modulus check

dkostic:
### Issues:
CryptoAlg-2620

### Description of changes: 
ML-KEM encapsulation key modulus check as specified in FIPS 203 Section 7.2.
https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf

### Call-outs:

### Testing:

### DIFF
--- a/crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
+++ b/crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
@@ -5,6 +5,7 @@
 #include "kem.h"
 #include "indcpa.h"
 #include "verify.h"
+#include "reduce.h"
 #include "symmetric.h"
 #include "openssl/rand.h"
 
@@ -108,6 +109,17 @@ static void byte_encode_12(uint8_t out[BYTE_ENCODE_12_OUT_SIZE],
   bits_to_bytes(out, BYTE_ENCODE_12_OUT_SIZE, bits, BYTE_ENCODE_12_NUM_BITS);
 }
 
+// Converts a centered representative |in| which is an integer in
+// {-(q-1)/2, ..., (q-1)/2}, to a positive representative in {0, ..., q-1}.
+// It implements in constant-time the following operation:
+//   return (in < 0) ? in + KYBER_Q : in;
+static int16_t centered_to_positive_representative(int16_t in) {
+  // mask = (in < 0) ? b11..11 : b00..00;
+  crypto_word_t mask = constant_time_is_zero_w(in >> 15);
+  int16_t in_fixed = in + KYBER_Q;
+  return constant_time_select_int(mask, in, in_fixed);
+}
+
 #define BYTE_DECODE_12_OUT_SIZE (256)
 #define BYTE_DECODE_12_IN_SIZE  (32 * 12)
 #define BYTE_DECODE_12_NUM_BITS (256 * 12)
@@ -123,7 +135,7 @@ static void byte_decode_12(int16_t out[BYTE_DECODE_12_OUT_SIZE],
     for (size_t j = 0; j < 12; j++) {
       val |= bits[i * 12 + j] << j;
     }
-    out[i] = val % KYBER_Q; // TODO(dkostic): make CT.
+    out[i] = centered_to_positive_representative(barrett_reduce(val));
   }
 }
 
@@ -140,6 +152,7 @@ static int encapsulation_key_modulus_check(ml_kem_params *params, const uint8_t 
     byte_decode_12(&ek_decoded[i * BYTE_DECODE_12_OUT_SIZE], &ek[i * BYTE_DECODE_12_IN_SIZE]);
     byte_encode_12(&ek_recoded[i * BYTE_ENCODE_12_OUT_SIZE], &ek_decoded[i * BYTE_ENCODE_12_IN_SIZE]);
   }
+
   return verify(ek_recoded, ek, params->k * BYTE_ENCODE_12_OUT_SIZE);
 }
 
--- a/crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
+++ b/crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
@@ -59,6 +59,90 @@ int crypto_kem_keypair(ml_kem_params *params,
   return 0;
 }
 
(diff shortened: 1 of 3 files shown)


crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -59,6 +59,90 @@ int crypto_kem_keypair(ml_kem_params *params,
...
+
+// FIPS 203. Section 7.2 Encapsulation key check.
+static int encapsulation_key_modulus_check(ml_kem_params *params, const uint8_t *ek) {
+
+  int16_t ek_decoded[ENCAPS_KEY_DECODED_MAX_SIZE];
+  uint8_t ek_recoded[ENCAPS_KEY_ENCODED_MAX_SIZE];
+
+  for (size_t i = 0; i < params->k; i++) {
+    byte_decode_12(&ek_decoded[i * BYTE_DECODE_12_OUT_SIZE], &ek[i * BYTE_DECODE_12_IN_SIZE]);
+    byte_encode_12(&ek_recoded[i * BYTE_ENCODE_12_OUT_SIZE], &ek_decoded[i * BYTE_ENCODE_12_IN_SIZE]);
```
WillChilds-Klein, line 141: suggest setting `#define BYTE_ENCODE_12_OUT_SIZE BYTE_DECODE_12_IN_SIZE` above. if those values get out of sync, these encode/decode operations would become misaligned.

nebeid: Similarly, the `DECODE_12` macros can be defined in terms of `ENCODE_12` macros.

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -59,6 +59,90 @@ int crypto_kem_keypair(ml_kem_params *params,
...
+static void byte_decode_12(int16_t out[BYTE_DECODE_12_OUT_SIZE],
+                           const uint8_t in[BYTE_DECODE_12_IN_SIZE]) {
+  uint8_t bits[BYTE_DECODE_12_NUM_BITS] = {0};
+  bytes_to_bits(bits, BYTE_DECODE_12_NUM_BITS, in, BYTE_DECODE_12_IN_SIZE);
+  for (size_t i = 0; i < BYTE_DECODE_12_OUT_SIZE; i++) {
+    int16_t val = 0;
+    for (size_t j = 0; j < 12; j++) {
+      val |= bits[i * 12 + j] << j;
+    }
+    out[i] = val % KYBER_Q; // TODO(dkostic): make CT.
```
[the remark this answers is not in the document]
dkostic: good point, I already pushed a commit that does it in constant-time but I will reconsider.

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
dkostic: I did that at first but it came out confusing (to me). The definitions as they are now are easy to map to the algorithm definitions in the FIPS spec so I think I'd keep it as is if you don't mind.

WillChilds-Klein: that's reasonable

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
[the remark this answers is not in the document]
dkostic: I think it's better to do it in constant-time because the same function can be used for decoding the secret key in decaps (as specified in FIPS 203). It's not used at the moment, but to avoid future mishaps where someone would accidentally start using the function with secret data I suggest we keep it constant-time. If any performance issues arise from this change we can reconsider the decision.

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -112,13 +196,17 @@ int crypto_kem_enc_derand(ml_kem_params *params,
...
-* Returns 0 (success)
+* Returns 0 (success), or 1 when the encapsulation key check fails.
 **************************************************/
 int crypto_kem_enc(ml_kem_params *params,
                    uint8_t *ct,
                    uint8_t *ss,
                    const uint8_t *pk)
 {
+  if (encapsulation_key_modulus_check(params, pk) != 0) {
+    return 1;
```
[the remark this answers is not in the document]
dkostic: I looked into this -- the wording in that comment is a bit weird. It says an "invalid public key" but what it means is a public key that doesn't match the corresponding secret key. So I updated the comment. There was another issue with the test code that I fixed as well. The code was "invalidating" the key by negating its first byte. This can make the key actually invalid in the sense of FIPS 203 and with the change in this PR such key would be rejected, while the intention of the test is not to test real invalid keys but valid public keys that are not matching the secret key and thus would fail the final handshake.

WillChilds-Klein (approved):

nebeid (approved):

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -59,6 +60,102 @@ int crypto_kem_keypair(ml_kem_params *params,
...
+
+#define ENCAPS_KEY_ENCODED_MAX_SIZE (BYTE_ENCODE_12_OUT_SIZE * KYBER_K_MAX)
+#define ENCAPS_KEY_DECODED_MAX_SIZE (BYTE_DECODE_12_OUT_SIZE * KYBER_K_MAX)
+
+// FIPS 203. Section 7.2 Encapsulation key check.
+static int encapsulation_key_modulus_check(ml_kem_params *params, const uint8_t *ek) {
+
+  int16_t ek_decoded[ENCAPS_KEY_DECODED_MAX_SIZE];
+  uint8_t ek_recoded[ENCAPS_KEY_ENCODED_MAX_SIZE];
+
```
jvdsn, line 150: Why is the size of ek not verified here?

jvdsn: Answered in https://github.com/aws/aws-lc/pull/1873/files#r1773741238

PR #1872, opened 2024-09-23
Default entropy source

torben-hansen:
### Description of changes: 

This PR implements a default entropy source object for the randomness generation. This is done by defining the function table that is already assumed by the randomness generation code in `new_rand.c`. The configuration is similar to the one already being used but simplified. It's now assumed that an entropy source can never not be initialised which seems reasonable...

Since the entropy source method table is a const object, `get_entropy_source()` is changed to return a reference to it. Related code is also changed to take this into account.

Finally, the entropy source object "cleanup" function is changed to not return anything. It's called at thread exit and if it fails there is no way to recover anyway.

### DIFF
--- a/crypto/fipsmodule/rand/entropy/entropy_sources.c
+++ b/crypto/fipsmodule/rand/entropy/entropy_sources.c
@@ -7,34 +7,52 @@
 #include <openssl/base.h>
 
 #include "internal.h"
+#include "../internal.h"
+#include "../../delocate.h"
 
-static int fake_void(void) {
+static int entropy_default_initialize(void) {
   return 1;
 }
 
-static int fake_rand(uint8_t a[CTR_DRBG_ENTROPY_LEN]) {
-  OPENSSL_cleanse(a, CTR_DRBG_ENTROPY_LEN);
+static void entropy_default_cleanup(void) {
+}
+
+static int entropy_default_get_seed(uint8_t seed[CTR_DRBG_ENTROPY_LEN]) {
+  CRYPTO_sysrand_for_seed(seed, CTR_DRBG_ENTROPY_LEN);
   return 1;
 }
 
-static int fake_rand_(uint8_t a[RAND_PRED_RESISTANCE_LEN]) {
-  OPENSSL_cleanse(a, RAND_PRED_RESISTANCE_LEN);
+static int entropy_default_get_prediction_resistance(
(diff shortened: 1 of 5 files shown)


crypto/fipsmodule/rand/entropy/entropy_sources.c
```diff
@@ -7,34 +7,52 @@
...
+  uint8_t pred_resistance[RAND_PRED_RESISTANCE_LEN]) {
+  if (have_fast_rdrand() != 1 ||
+      rdrand(pred_resistance, RAND_PRED_RESISTANCE_LEN) != 1) {
+    return 0;
+  }
+  return 1;
+}
 
-  GUARD_PTR(entropy_source);
+static int entropy_default_randomize(void) {
```
dkostic, line 34: just curious, what's this function meant to do?

dkostic (approved):

crypto/fipsmodule/rand/entropy/entropy_sources.c
torben-hansen: I model the "entropy source" as a stateful object and I (might) need a function that can request a randomization of the object.

I do that because I want to implement a tree-DRBG that has a live entropy source at it's root. But this means that to handle uniqueness-breaking events (e.g. a fork), we need a method to randomize the tree-DRBG. This function is intended to do that.

However, I might decide later that this function is actually redundant and the entropy source backend must handle this UBEs by its own.

crypto/fipsmodule/rand/entropy/entropy_sources.c
```diff
@@ -7,34 +7,52 @@
...
-
-  // In the future this function will lazily initialise a global entropy source.
+static int entropy_default_get_prediction_resistance(
+  uint8_t pred_resistance[RAND_PRED_RESISTANCE_LEN]) {
+  if (have_fast_rdrand() != 1 ||
+      rdrand(pred_resistance, RAND_PRED_RESISTANCE_LEN) != 1) {
+    return 0;
+  }
+  return 1;
+}
```
andrewhop, line 32: Am I missing something here, or does this do nothing?

andrewhop: I'm still a little confused, what is this supposed to do? It never returns the prediction_resistance array and it's a local variable.

andrewhop (approved):

dkostic (approved):

PR #1873, opened 2024-09-23
ML-KEM decapsulation key hash check

dkostic:
### Issues:
CryptoAlg-2620

### Description of changes: 
ML-KEM decapsulation key check as specified in Section 7.3 of FIPS 203: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf

### Call-outs:

### Testing:

### DIFF
--- a/crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
+++ b/crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
@@ -156,6 +156,22 @@ static int encapsulation_key_modulus_check(ml_kem_params *params, const uint8_t 
   return verify(ek_recoded, ek, params->k * BYTE_ENCODE_12_OUT_SIZE);
 }
 
+// FIPS 203. Section 7.3 Decapsulation key hash check
+// The spec defines the decapsulation key as following:
+//   dk <-- (dk_pke || ek || H(ek) || z).
+// This check takes |ek| out of |dk|, computes H(ek), and verifies that it is
+// the same as the H(ek) portion stored in |dk|.
+static int decapsulation_key_hash_check(ml_kem_params *params, const uint8_t *dk) {
+  uint8_t dk_pke_hash_computed[KYBER_SYMBYTES] = {0};
+
+  hash_h(dk_pke_hash_computed, &dk[params->indcpa_secret_key_bytes],
+                               params->indcpa_public_key_bytes);
+  const uint8_t *dk_pke_hash_expected = &dk[params->indcpa_secret_key_bytes +
+                                            params->indcpa_public_key_bytes];
+
+  return verify(dk_pke_hash_computed, dk_pke_hash_expected, KYBER_SYMBYTES);
+}
+
 /*************************************************
 * Name:        crypto_kem_enc_derand
 *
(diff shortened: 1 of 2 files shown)


crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -156,6 +156,22 @@ static int encapsulation_key_modulus_check(ml_kem_params *params, const uint8_t
   return verify(ek_recoded, ek, params->k * BYTE_ENCODE_12_OUT_SIZE);
 }
 
+// FIPS 203. Section 7.3 Decapsulation key hash check
+// The spec defines the decapsulation key as following:
+//   dk <-- (dk_pke || ek || H(ek) || z).
+// This check takes |ek| out of |dk|, computes H(ek), and verifies that it is
+// the same as the H(ek) portion stored in |dk|.
```
WillChilds-Klein, line 163: this sounds like input check 3. from the spec, but what about the "type checks" in 1. and 2.?

nebeid: I'm wondering particularly about check 1 on the ciphertext which is said to have to be performed for every decapsulation.

dkostic: yes, this is the Hash check (as specified in L159). The other two Decaps checks and also the first check for Encaps (Section 7.2) are all length checks on the input arrays. We can safely omit them here because those checks are done in higher level functions. The required lengths for different variants of ML-KEM are hard-coded here:
- https://github.com/aws/aws-lc/blob/2835116571e8607861ba8dd0be2676f2c5cfaaab/crypto/fipsmodule/kem/kem.c#L133
The ciphertext length for Decaps is checked here:
- https://github.com/aws/aws-lc/blob/2835116571e8607861ba8dd0be2676f2c5cfaaab/crypto/fipsmodule/evp/p_kem.c#L272
If a key is generated by aws-lc then it satisfies the length requirements. If a key is generated outside of aws-lc, it has to be imported into an `EVP_PKEY` object to be used within aws-lc. We provide only these three functions to do that: `EVP_PKEY_kem_new_raw_key`, `EVP_PKEY_kem_new_raw_secret_key`, `EVP_PKEY_kem_new_raw_public_key`. The lengths are checked for example here:
- https://github.com/aws/aws-lc/blob/2835116571e8607861ba8dd0be2676f2c5cfaaab/crypto/fipsmodule/evp/p_kem.c#L465

nebeid: Thanks, @dkostic. Maybe you can summarise somewhere that all checks for encapsulate and decapsulate are done as per fips 203 in various places because I think someone coming to this part of the code where it says as in Sec 7.2 or 7.3 will wonder where the other checks are.

WillChilds-Klein (approved):

nebeid (approved):

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
jvdsn: > I think someone coming to this part of the code where it says as in Sec 7.2 or 7.3 will wonder where the other checks are.

That's me, and I was indeed wondering that. I would appreciate it if the comments could be added.

PR #1874, opened 2024-09-24
ML-KEM encaps key modulus check optimization

dkostic:
### Issues:
CryptoAlg-2620

### Description of changes: 
PR #1868 implemented the encapsulation key modulus check
naively, as specified in FIPS 203 which introduced a significant
performance hit to encapsulation. In this change we optimize
encoding/decoding to/from bytes functions to reduce the performance
regression.

For example, encapsulation performance (ops/s) measured on M1 macbook:

```
| Variant | Before #1868 | After #1868 | After this change |
| 512     |    94738     |    68293    |       92494       |
| 768     |    58672     |    42083    |       55572       |
| 1024    |    38893     |    28536    |       36820       |
```

### Call-outs:

### Testing:

### DIFF
--- a/crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
+++ b/crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
@@ -60,54 +60,77 @@ int crypto_kem_keypair(ml_kem_params *params,
   return 0;
 }
 
+// REFERENCE IMPLEMENTATION OF SEVERAL FIPS 203 FUNCTIONS.
+// Further below we implement optimized versions of the functions
+// that are actually used. We commented out and kept the reference
+// code for posterity.
+//
 // FIPS 203. Algorithm 3 BitsToBytes
 // Converts a bit array (of a length that is a multiple of eight)
 // into an array of bytes.
-static void bits_to_bytes(uint8_t *bytes, size_t num_bytes,
-                          const uint8_t *bits, size_t num_bits) {
-  assert(num_bits == num_bytes * 8);
-
-  for (size_t i = 0; i < num_bytes; i++) {
-    uint8_t byte = 0;
-    for (size_t j = 0; j < 8; j++) {
-      byte |= (bits[i * 8 + j] << j);
-    }
-    bytes[i] = byte;
-  }
-}
-
+// static void bits_to_bytes(uint8_t *bytes, size_t num_bytes,
+//                           const uint8_t *bits, size_t num_bits) {
+//   assert(num_bits == num_bytes * 8);
+// 
+//   for (size_t i = 0; i < num_bytes; i++) {
+//     uint8_t byte = 0;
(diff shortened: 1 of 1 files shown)


crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -120,22 +143,58 @@ static int16_t centered_to_positive_representative(int16_t in) {
...
-#define BYTE_DECODE_12_NUM_BITS (256 * 12)
+#define BYTE_ENCODE_12_IN_SIZE  (256)
+#define BYTE_ENCODE_12_OUT_SIZE (32 * 12)
+#define BYTE_DECODE_12_OUT_SIZE (BYTE_ENCODE_12_IN_SIZE)
+#define BYTE_DECODE_12_IN_SIZE  (BYTE_ENCODE_12_OUT_SIZE)
+
+// FIPS 203. Algorithm 5 ByteEncode_12
+// Encodes an array of 256 12-bit integers into a byte array.
+// Intuition for the implementation:
+//   in:  |xxxxxxxxyyyy| |yyyyzzzzzzzz| ...
```
nebeid, line 154: I'm a bit confused how this is the same as the commented-out implementation: I thought that each 16-bit array member contained a 12-bit digit and the top 4 are discarded. 
The inner loops of the above implementations which looped over the bits was iterating up to 12, then control moved to the next element, unless I'm misunderstanding.

dkostic: > each 16-bit array member contained a 12-bit digit and the top 4 are discarded.

that's correct, and that's what we are doing here as well

nebeid: I see it now, thanks.

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -120,22 +143,58 @@ static int16_t centered_to_positive_representative(int16_t in) {
...
+  }
+}
 
-// FIPS 203. Algorithm 5 ByteDecode_12
+// FIPS 203. Algorithm 6 ByteDecode_12
 // Decodes a byte array into an array of 256 12-bit integers.
+// Intuition for the implementation:
+//   in:  |xxxxxxxx| |yyyyyyyy| |zzzzzzzz| ...
+//   out: |xxxxxxxxyyyy| |yyyyzzzzzzzz| ...
+// We divide the input in triples of elements (3 x 8 bits = 24 bits),
```
nebeid, line 177: Nit
```suggestion
// We divide the input in triplets of elements (3 x 8 bits = 24 bits),
```

WillChilds-Klein (approved):
Should we skip these checks when not built in FIPS mode?

nebeid (approved):

PR #1876, opened 2024-09-25
Fix flaky ssl BadKemKeyShare tests

dkostic:
### Issues:
N/A

### Description of changes: 
The way the test was generating a public key that is not consistent with a secret key is by xor-ing the first byte of the key with 1. Such key modifications can inadvertently make the key invalid and thus fail the test.

For example, before performing encapsulation ML-KEM decodes the public key bytes to an array of 12-bit coefficients and checks that all coefficients are in the range [0, 3328]. If the first two bytes of the key encode the coefficient 3328 then xor-ing the first byte with 1 will make the coefficient equal to 3329. The call to encapsulate will then fail because 3329 is an invalid coefficient.

### Call-outs:

### Testing:

### DIFF
--- a/ssl/ssl_test.cc
+++ b/ssl/ssl_test.cc
@@ -11966,30 +11966,29 @@ TEST_P(BadKemKeyShareAcceptTest, BadKemKeyShareAccept) {
   {
     bssl::UniquePtr<SSLKeyShare> server_key_share = bssl::SSLKeyShare::Create(t.group_id);
     bssl::UniquePtr<SSLKeyShare> client_key_share = bssl::SSLKeyShare::Create(t.group_id);
+    bssl::UniquePtr<SSLKeyShare> random_key_share = bssl::SSLKeyShare::Create(t.group_id);
     ASSERT_TRUE(server_key_share);
     ASSERT_TRUE(client_key_share);
+    ASSERT_TRUE(random_key_share);
     uint8_t server_alert = 0;
     uint8_t client_alert = 0;
     Array<uint8_t> server_secret;
     Array<uint8_t> client_secret;
     CBB server_out_public_key;
     CBB client_out_public_key;
+    CBB random_out_public_key;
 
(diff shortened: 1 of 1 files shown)


alexw91 (approved):

ssl/ssl_test.cc
```diff
@@ -11966,30 +11966,29 @@ TEST_P(BadKemKeyShareAcceptTest, BadKemKeyShareAccept) {
...
-    // Modify the public key making it incompatible with the secret key
-    uint8_t *modified_client_public_key_buf =
-      (uint8_t *)OPENSSL_malloc(t.offer_key_share_size);
-    ASSERT_TRUE(modified_client_public_key_buf);
-    const uint8_t *client_out_public_key_data = CBB_data(&client_out_public_key);
-    ASSERT_TRUE(client_out_public_key_data);
-    OPENSSL_memcpy(modified_client_public_key_buf, client_out_public_key_data,
-                   t.offer_key_share_size);
-    modified_client_public_key_buf[0] ^= 1;
+    // Generate a radnom public key that is incompatible with client's secret key
```
samuel40791765, line 11985:
```suggestion
    // Generate a random public key that is incompatible with client's secret key
```

samuel40791765 (approved):

ssl/ssl_test.cc
```diff
@@ -12116,16 +12119,13 @@ TEST_P(BadKemKeyShareFinishTest, BadKemKeyShareFinish) {
   {
     // The server's public key was already correctly generated previously in
     // a call to Accept(). Here we modify it.
```
nebeid, line 12121: It would be good to modify this comment too to say the public key is randomly generated as in the test above.

dkostic: done

nebeid (approved):

WillChilds-Klein (approved):

PR #1881, opened 2024-09-26
Add EC_GROUP mutablility to custom curves

samuel40791765:
### Issues:
Resolves `CryptoAlg-2547`

### Description of changes: 
Following up on https://github.com/aws/aws-lc/pull/1860 which introduces an API to return a mutable `EC_GROUP`, this commit applies the same concept for custom curves. Custom curves are already dynamically allocated, but were logically immutable due to the strict enforcement of calling `EC_GROUP_new_curve_GFp` -> `EC_GROUP_set_generator`.
The reference call in `EC_GROUP` that was used for custom curves is no longer needed. We actually duplicate the structure instead.

### Call-outs:
N/A

### Testing:
Test using a custom generator identical to P-256.

### DIFF
--- a/crypto/fipsmodule/ec/ec.c
+++ b/crypto/fipsmodule/ec/ec.c
@@ -309,13 +309,7 @@ EC_GROUP *EC_GROUP_new_curve_GFp(const BIGNUM *p, const BIGNUM *a,
   if (ret == NULL) {
     return NULL;
   }
-  ret->references = 1;
-  // TODO: Treat custom curves as "immutable" for now. There's the possibility
-  // that we'll need dynamically allocated custom curve |EC_GROUP|s. But there
-  // are additional complexities around untangling the expectations already set
-  // for |EC_GROUP_new_curve_GFp| and |EC_GROUP_set_generator|.
-  // Note: See commit cb16f17b36d9ee9528a06d2e520374a4f177af4d.
-  ret->mutable_ec_group = 0;
+  ret->mutable_ec_group = 1;
   ret->conv_form = POINT_CONVERSION_UNCOMPRESSED;
   ret->meth = EC_GFp_mont_method();
   bn_mont_ctx_init(&ret->field);
@@ -336,11 +330,10 @@ err:
 int EC_GROUP_set_generator(EC_GROUP *group, const EC_POINT *generator,
                            const BIGNUM *order, const BIGNUM *cofactor) {
   if (group->curve_name != NID_undef || group->has_order ||
-      generator->group != group) {
+      EC_GROUP_cmp(generator->group, group, NULL)) {
     // |EC_GROUP_set_generator| may only be used with |EC_GROUP|s returned by
     // |EC_GROUP_new_curve_GFp| and may only used once on each group.
-    // |generator| must have been created from |EC_GROUP_new_curve_GFp|, not a
-    // copy, so that |generator->group->generator| is set correctly.
+    // |generator| must be of the same |EC_GROUP| as |group|.
     OPENSSL_PUT_ERROR(EC, ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
     return 0;
   }
@@ -443,8 +436,7 @@ void EC_GROUP_free(EC_GROUP *group) {
   }
 
   if (!group->mutable_ec_group) {
-    if (group->curve_name != NID_undef ||
-        !CRYPTO_refcount_dec_and_test_zero(&group->references)) {
+    if (group->curve_name != NID_undef) {
       // Built-in curves are static.
       return;
     }
@@ -465,12 +457,6 @@ EC_GROUP *EC_GROUP_dup(const EC_GROUP *a) {
       // Built-in curves are static.
       return (EC_GROUP *)a;
     }
-
-    // Groups are logically immutable (but for |EC_GROUP_set_generator| which
-    // must be called early on), so we simply take a reference.
-    EC_GROUP *group = (EC_GROUP *)a;
-    CRYPTO_refcount_inc(&group->references);
-    return group;
   }
 
   // Directly duplicate the |EC_GROUP| if it was dynamically allocated. We do a
@@ -492,10 +478,6 @@ EC_GROUP *EC_GROUP_dup(const EC_GROUP *a) {
 }
 
 int EC_GROUP_cmp(const EC_GROUP *a, const EC_GROUP *b, BN_CTX *ignored) {
-  // Note this function returns 0 if equal and non-zero otherwise.
-  if (a == b) {
-    return 0;
-  }
   // Built-in static curves may be compared by curve name alone.
   if (a->curve_name != b->curve_name) {
     return 1;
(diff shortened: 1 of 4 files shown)


crypto/fipsmodule/ec/ec.c
```diff
@@ -443,8 +436,7 @@ void EC_GROUP_free(EC_GROUP *group) {
   }
 
   if (!group->mutable_ec_group) {
-    if (group->curve_name != NID_undef ||
-        !CRYPTO_refcount_dec_and_test_zero(&group->references)) {
+    if (group->curve_name != NID_undef) {
```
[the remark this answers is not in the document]
samuel40791765: I don't think it's mentioned anywhere, it's mainly implied from reading the code surrounding it. So yes, definitely worth documenting.

crypto/fipsmodule/ec/ec.c
```diff
@@ -506,17 +488,20 @@ int EC_GROUP_cmp(const EC_GROUP *a, const EC_GROUP *b, BN_CTX *ignored) {
...
-  return a->meth != b->meth ||  //
-         !a->has_order || !b->has_order ||
-         BN_cmp(&a->order.N, &b->order.N) != 0 ||
-         BN_cmp(&a->field.N, &b->field.N) != 0 ||
-         !ec_felem_equal(a, &a->a, &b->a) ||  //
-         !ec_felem_equal(a, &a->b, &b->b) ||
-         !ec_GFp_simple_points_equal(a, &a->generator.raw, &b->generator.raw);
+  // |a| and |b| are both custom curves. We compare the entire curve structure
+  // if both |a| and |b| are complete. If either are incomplete (due to legacy
+  // OpenSSL mistakes, custom curve construction is sadly done in two parts), we
```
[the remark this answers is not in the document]
samuel40791765: No problem, I remember this not being the most intuitive when I going through the code the first time.

crypto/fipsmodule/ec/ec.c
```diff
@@ -336,11 +330,10 @@ EC_GROUP *EC_GROUP_new_curve_GFp(const BIGNUM *p, const BIGNUM *a,
 int EC_GROUP_set_generator(EC_GROUP *group, const EC_POINT *generator,
                            const BIGNUM *order, const BIGNUM *cofactor) {
   if (group->curve_name != NID_undef || group->has_order ||
-      generator->group != group) {
+      EC_GROUP_cmp(generator->group, group, NULL)) {
```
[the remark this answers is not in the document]
samuel40791765: I don't think so, the first two checks are to verify that only a "custom curve" `EC_GROUP` is being configured and that it does not have a generator set yet. 
This specific check is to verify that the generator's `EC_GROUP` is of the same type/group as the one being configured.

crypto/fipsmodule/ec/ec.c
```diff
@@ -506,17 +488,20 @@ int EC_GROUP_cmp(const EC_GROUP *a, const EC_GROUP *b, BN_CTX *ignored) {
...
+  // |a| and |b| are both custom curves. We compare the entire curve structure
+  // if both |a| and |b| are complete. If either are incomplete (due to legacy
+  // OpenSSL mistakes, custom curve construction is sadly done in two parts), we
+  // only compare the parts that are available.
+  if (a->has_order && b->has_order) {
+    return a->meth != b->meth || BN_cmp(&a->order.N, &b->order.N) != 0 ||
+           BN_cmp(&a->field.N, &b->field.N) != 0 ||
+           !ec_felem_equal(a, &a->a, &b->a) ||
+           !ec_felem_equal(a, &a->b, &b->b) ||
+           !ec_GFp_simple_points_equal(a, &a->generator.raw, &b->generator.raw);
```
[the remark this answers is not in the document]
samuel40791765: Great question, I assumed so since we had already been doing this and the [criteria](https://github.com/aws/aws-lc/blob/bda01b40727126e8cc2ec1e375301700b0fa6ba7/crypto/fipsmodule/ec/simple.c#L224-L269) of `ec_GFp_simple_points_equal` was relatively strict as opposed to  `ec_felem_equal`. But I'm hesitant on jumping to my own conclusions on this, I'd feel more comfortable getting someone more familiar with elliptic curves to weigh in. 

cc @dkostic: Do we need an additional check for `ec_GFp_simple_points_equal(b, &a->generator.raw, &b->generator.raw)` at this point?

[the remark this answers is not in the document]
dkostic: we don't need the additional check at this point

crypto/fipsmodule/ec/ec.c
```diff
@@ -506,17 +488,21 @@ int EC_GROUP_cmp(const EC_GROUP *a, const EC_GROUP *b, BN_CTX *ignored) {
...
-         BN_cmp(&a->order.N, &b->order.N) != 0 ||
-         BN_cmp(&a->field.N, &b->field.N) != 0 ||
-         !ec_felem_equal(a, &a->a, &b->a) ||  //
-         !ec_felem_equal(a, &a->b, &b->b) ||
-         !ec_GFp_simple_points_equal(a, &a->generator.raw, &b->generator.raw);
+  // |a| and |b| are both custom curves. We compare the entire curve structure
+  // if both |a| and |b| are complete. If either are incomplete (due to legacy
+  // OpenSSL mistakes, custom curve construction is sadly done in two parts
+  // |EC_GROUP_new_curve_GFp| -> |EC_GROUP_set_generator|), we only compare
+  // the parts that are available.
```
dkostic, line 495: why this change to "only compare the parts that are available"?

crypto/fipsmodule/ec/ec.c
```diff
@@ -506,17 +488,21 @@ int EC_GROUP_cmp(const EC_GROUP *a, const EC_GROUP *b, BN_CTX *ignored) {
...
+  if (a->has_order && b->has_order) {
+    return a->meth != b->meth || BN_cmp(&a->order.N, &b->order.N) != 0 ||
+           BN_cmp(&a->field.N, &b->field.N) != 0 ||
+           !ec_felem_equal(a, &a->a, &b->a) ||
+           !ec_felem_equal(a, &a->b, &b->b) ||
+           !ec_GFp_simple_points_equal(a, &a->generator.raw, &b->generator.raw);
+  } else {
```
dkostic, line 502: in the `else` branch you can end up with one of the groups having a generator and the other one not. We should probably ask that either both groups have a generator (order) or both don't have it?

samuel40791765: This was intentional since I noticed that `ECTest.BIGNUMConvert` was failing. But after looking more closely, it's a test that I had wrote. The test is using an outdated generator to compare the point contents, I think we can just fix the test and add the additional check here.

crypto/fipsmodule/ec/ec.c
```diff
@@ -506,17 +492,22 @@ int EC_GROUP_cmp(const EC_GROUP *a, const EC_GROUP *b, BN_CTX *ignored) {
...
-  // same object, we consider them always unequal.
-  return a->meth != b->meth ||  //
-         !a->has_order || !b->has_order ||
-         BN_cmp(&a->order.N, &b->order.N) != 0 ||
-         BN_cmp(&a->field.N, &b->field.N) != 0 ||
-         !ec_felem_equal(a, &a->a, &b->a) ||  //
-         !ec_felem_equal(a, &a->b, &b->b) ||
-         !ec_GFp_simple_points_equal(a, &a->generator.raw, &b->generator.raw);
+  // |a| and |b| are both custom curves. We compare the entire curve structure
+  // if both |a| and |b| are complete. If either are incomplete (due to legacy
```
dkostic, line 496: I think this needs to be updated now
```
If either are incomplete ... we only compare the parts that are available.
```

crypto/fipsmodule/ec/ec.c
```diff
@@ -506,17 +492,22 @@ int EC_GROUP_cmp(const EC_GROUP *a, const EC_GROUP *b, BN_CTX *ignored) {
...
+  if (a->has_order && b->has_order) {
+    return a->meth != b->meth || BN_cmp(&a->order.N, &b->order.N) != 0 ||
+           BN_cmp(&a->field.N, &b->field.N) != 0 ||
+           !ec_felem_equal(a, &a->a, &b->a) ||
+           !ec_felem_equal(a, &a->b, &b->b) ||
+           !ec_GFp_simple_points_equal(a, &a->generator.raw, &b->generator.raw);
+  } else {
+    return a->meth != b->meth || a->has_order != b->has_order ||
+           BN_cmp(&a->field.N, &b->field.N) != 0 ||
+           !ec_felem_equal(a, &a->a, &b->a) || !ec_felem_equal(a, &a->b, &b->b);
```
dkostic, line 509: nit: all these checks are done above as well, it'd be cool to refactor the code to avoid duplication it possible

samuel40791765: Great point! will do 👍

crypto/fipsmodule/ec/ec.c
samuel40791765: Ah yes, will change to "if both are incomplete"

dkostic (approved):

WillChilds-Klein (approved):

PR #1886, opened 2024-09-27
Add PKCS7-internal BIO_f_md

WillChilds-Klein:
### Issues:
Addresses CryptoAlg-2494

### Description of changes: 

This change introduces a new filter BIO, `BIO_f_md` for use in PR 1816.

### Call-outs:
- n/a

### Testing:
- new unit tests

---

### DIFF
--- a/crypto/pkcs7/bio/md.c
+++ b/crypto/pkcs7/bio/md.c
@@ -9,21 +9,6 @@
 #include "../crypto/bio/internal.h"
 #include "../internal.h"
 
-// |md_write| and |md_read| both contribute to the digest, |md_gets| returns
-// digest contents
-static int md_write(BIO *h, char const *buf, int num);
-static int md_read(BIO *h, char *buf, int size);
-static int md_gets(BIO *h, char *str, int size);
-static long md_ctrl(BIO *h, int cmd, long arg1, void *arg2);
-static int md_new(BIO *h);
-static int md_free(BIO *data);
-
-
-static const BIO_METHOD methods_md = {
-    BIO_TYPE_MD, "message digest", md_write, md_read, /*puts*/ NULL,
-    md_gets,     md_ctrl,          md_new,   md_free, /*callback_ctrl*/ NULL,
-};
-const BIO_METHOD *BIO_f_md(void) { return &methods_md; }
 
 static int md_new(BIO *b) {
   GUARD_PTR(b);
@@ -65,7 +50,7 @@ static int md_read(BIO *b, char *out, int outl) {
   ret = BIO_read(next, out, outl);
   if (ret > 0) {
     if (EVP_DigestUpdate(ctx, (unsigned char *)out, ret) <= 0) {
-      return -1;
+      ret = -1;
     }
   }
 
@@ -90,9 +75,8 @@ static int md_write(BIO *b, const char *in, int inl) {
 
   ret = BIO_write(next, in, inl);
   if (ret > 0) {
-    if (!EVP_DigestUpdate(ctx, (const unsigned char *)in, ret)) {
-      BIO_clear_retry_flags(b);
-      return 0;
+    if (EVP_DigestUpdate(ctx, (const unsigned char *)in, ret) <= 0) {
+      ret = -1;
     }
   }
 
@@ -172,3 +156,18 @@ static int md_gets(BIO *b, char *buf, int size) {
 
   return ret;
 }
+
+static const BIO_METHOD methods_md = {
+    BIO_TYPE_MD,       // type
+    "message digest",  // name
+    md_write,          // bwrite
+    md_read,           // bread
+    NULL,              // bputs
+    md_gets,           // bgets
+    md_ctrl,           // ctrl
+    md_new,            // create
+    md_free,           // destroy
+    NULL,              // callback_ctrl
+};
+
+const BIO_METHOD *BIO_f_md(void) { return &methods_md; }
--- a/crypto/pkcs7/bio/md.c
+++ b/crypto/pkcs7/bio/md.c
@@ -9,7 +9,8 @@
 #include "../crypto/bio/internal.h"
 #include "../internal.h"
 
-// BIO_put and BIO_get both add to the digest, BIO_gets returns the digest
+// |md_write| and |md_read| both contribute to the digest, |md_gets| returns
+// digest contents
 static int md_write(BIO *h, char const *buf, int num);
 static int md_read(BIO *h, char *buf, int size);
 static int md_gets(BIO *h, char *str, int size);
--- a/crypto/pkcs7/bio/md.c
+++ b/crypto/pkcs7/bio/md.c
@@ -1,0 +1,177 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include <errno.h>
+#include <openssl/buffer.h>
+#include <openssl/evp.h>
+#include <openssl/pkcs7.h>
+#include "../../internal.h"
+#include "../crypto/bio/internal.h"
+#include "../internal.h"
+
+// BIO_put and BIO_get both add to the digest, BIO_gets returns the digest
+static int md_write(BIO *h, char const *buf, int num);
+static int md_read(BIO *h, char *buf, int size);
+static int md_gets(BIO *h, char *str, int size);
+static long md_ctrl(BIO *h, int cmd, long arg1, void *arg2);
+static int md_new(BIO *h);
+static int md_free(BIO *data);
+
+
+static const BIO_METHOD methods_md = {
+    BIO_TYPE_MD, "message digest", md_write, md_read, /*puts*/ NULL,
+    md_gets,     md_ctrl,          md_new,   md_free, /*callback_ctrl*/ NULL,
+};
+const BIO_METHOD *BIO_f_md(void) { return &methods_md; }
+
+static int md_new(BIO *b) {
+  GUARD_PTR(b);
+  EVP_MD_CTX *ctx;
+
+  ctx = EVP_MD_CTX_new();
+  if (ctx == NULL) {
+    return 0;
+  }
+
+  BIO_set_data(b, ctx);
+
+  return 1;
+}
+
+static int md_free(BIO *b) {
+  GUARD_PTR(b);
+  EVP_MD_CTX_free(BIO_get_data(b));
+  BIO_set_data(b, NULL);
+  BIO_set_init(b, 0);
+
+  return 1;
+}
+
+static int md_read(BIO *b, char *out, int outl) {
+  GUARD_PTR(b);
+  GUARD_PTR(out);
+  int ret = 0;
+  EVP_MD_CTX *ctx;
+  BIO *next;
+
+  ctx = BIO_get_data(b);
+  next = BIO_next(b);
+
+  if ((ctx == NULL) || (next == NULL) || outl <= 0) {
+    return 0;
+  }
+
+  ret = BIO_read(next, out, outl);
+  if (BIO_get_init(b)) {
+    if (ret > 0) {
+      if (EVP_DigestUpdate(ctx, (unsigned char *)out, ret) <= 0) {
+        return -1;
+      }
+    }
+  }
+  BIO_clear_retry_flags(b);
+  BIO_copy_next_retry(b);
+  return ret;
+}
+
+static int md_write(BIO *b, const char *in, int inl) {
+  GUARD_PTR(b);
+  GUARD_PTR(in);
+  int ret = 0;
+  EVP_MD_CTX *ctx;
+  BIO *next;
+
+  if (inl <= 0) {
+    return 0;
+  }
+
+  ctx = BIO_get_data(b);
+  next = BIO_next(b);
+  if ((ctx != NULL) && (next != NULL)) {
+    ret = BIO_write(next, in, inl);
+  }
+
+  if (BIO_get_init(b)) {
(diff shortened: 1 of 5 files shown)


crypto/pkcs7/bio/md.c
```diff
@@ -0,0 +1,177 @@
...
+
+#include <errno.h>
+#include <openssl/buffer.h>
+#include <openssl/evp.h>
+#include <openssl/pkcs7.h>
+#include "../../internal.h"
+#include "../crypto/bio/internal.h"
+#include "../internal.h"
+
+// BIO_put and BIO_get both add to the digest, BIO_gets returns the digest
```
samuel40791765, line 12: Was this a leaked note? It doesn't seem like `BIO_f_md` has a puts method or am I overlooking something

crypto/pkcs7/bio/md.c
```diff
@@ -0,0 +1,177 @@
...
+      } else {
+        ret = 0;
+      }
+      if (ret > 0) {
+        ret = BIO_ctrl(next, cmd, num, ptr);
+      }
+      break;
+    case BIO_C_GET_MD_CTX:
+      pctx = ptr;
+      *pctx = ctx;
```
samuel40791765, line 132: From cross-referencing the [OpenSSL implementation](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/evp/bio_md.c#L167C9-L167C21), does this need a call to `BIO_set_init`?

crypto/pkcs7/bio/md.c
WillChilds-Klein: hm, should be "md_write and md_read both contribute to the digest, md_gets returns digest contents"

crypto/pkcs7/bio/md.c
WillChilds-Klein: I removed that intentionally. i don't think it makes much sense for a getter to set initialization.

crypto/pkcs7/bio/md.c
```diff
@@ -0,0 +1,178 @@
...
+static int md_gets(BIO *h, char *str, int size);
+static long md_ctrl(BIO *h, int cmd, long arg1, void *arg2);
+static int md_new(BIO *h);
+static int md_free(BIO *data);
+
+
+static const BIO_METHOD methods_md = {
+    BIO_TYPE_MD, "message digest", md_write, md_read, /*puts*/ NULL,
+    md_gets,     md_ctrl,          md_new,   md_free, /*callback_ctrl*/ NULL,
+};
```
justsmth, line 25: NP: This would be easier to read if each field was on a separate line.

crypto/pkcs7/bio/md.c
```diff
@@ -0,0 +1,178 @@
...
+
+  ctx = BIO_get_data(b);
+  next = BIO_next(b);
+
+  if ((ctx == NULL) || (next == NULL) || outl <= 0) {
+    return 0;
+  }
+
+  ret = BIO_read(next, out, outl);
+  if (BIO_get_init(b)) {
```
justsmth, line 66: This seems to support the use-case of leaving this `BIO` uninitialized while still reading data from `next`. In this case, data is returned that has not been digested. Is this use-case tested as well.

crypto/pkcs7/bio/md.c
```diff
@@ -0,0 +1,178 @@
...
+    if (ret > 0) {
+      if (EVP_DigestUpdate(ctx, (unsigned char *)out, ret) <= 0) {
+        return -1;
+      }
+    }
+  }
+  BIO_clear_retry_flags(b);
+  BIO_copy_next_retry(b);
```
justsmth, line 74: If `b` was not initialized, should we still be calling `BIO_clear_retry_flags` and `BIO_copy_next_retry`?

crypto/pkcs7/bio/md.c
```diff
@@ -0,0 +1,178 @@
...
+
+  if ((ctx == NULL) || (next == NULL) || outl <= 0) {
+    return 0;
+  }
+
+  ret = BIO_read(next, out, outl);
+  if (BIO_get_init(b)) {
+    if (ret > 0) {
+      if (EVP_DigestUpdate(ctx, (unsigned char *)out, ret) <= 0) {
+        return -1;
```
justsmth, line 69: If this ever returns an error, should we somehow mark this `BIO` and being not OK for subsequent use? In this case, `next` has provided new data that can not be reprocessed on any subsequent call.

crypto/pkcs7/bio/md.c
justsmth: There's mention of why this is done in OpenSSL's documentation: https://docs.openssl.org/master/man3/BIO_f_md/#description

> Calling BIO_get_md_ctx() will return the context and initialize the BIO state. This allows applications to initialize the context externally if the standard calls such as BIO_set_md() are not sufficiently flexible.

I think the idea is that the caller can "get" the context, allowing it can perform whatever initialization it requires.  So the act of getting the context implies that the caller will ensure it's initialized.

crypto/pkcs7/bio/md.c
```diff
@@ -0,0 +1,178 @@
...
+  EVP_MD_CTX *ctx;
+  unsigned int ret;
+
+  ctx = BIO_get_data(b);
+
+  if (((size_t)size) < EVP_MD_CTX_size(ctx)) {
+    return 0;
+  }
+
+  if (EVP_DigestFinal_ex(ctx, (unsigned char *)buf, &ret) <= 0) {
```
justsmth, line 173: I suspect that when calling `EVP_DigestFinal_ex` we should also `BIO_set_init(b, 0)` so that we don't attempt to update it again with more data.

crypto/pkcs7/bio/md.c
```diff
@@ -0,0 +1,178 @@
...
+      if (ret > 0) {
+        ret = BIO_ctrl(next, cmd, num, ptr);
+      }
+      break;
+    case BIO_C_GET_MD_CTX:
+      pctx = ptr;
+      *pctx = ctx;
+      break;
+    case BIO_C_SET_MD:
+      md = ptr;
```
justsmth, line 136: Should we verify that `ptr != NULL` here? The [implementation of `EVP_DigestInit_ex`](https://github.com/aws/aws-lc/blob/0afd2a8ab9ae310264e5089380fa42835d351f80/crypto/fipsmodule/digest/digest.c#L229) doesn't appear to check for that.

crypto/pkcs7/bio/md.c
justsmth: I think the answer to this is "yes". It would be needed to support its use prior to (or without) initialization.

crypto/pkcs7/bio/md.c
WillChilds-Klein: I agree, but this was `clang-format`'s doing. I've been using it for formatting stuff in the `pkcs7/` module. If there's some set of flags/options that we prefer to use I'm happy to specify them.

crypto/pkcs7/bio/md.c
WillChilds-Klein: that could cause a segfault on L234 of `digest.c`, so yeah we should probably error out here

crypto/pkcs7/bio/md.c
WillChilds-Klein: OpenSSL [doesn't do this](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/evp/bio_md.c#L98), but i agree philosophically that we should fail loudly and not silently corrupt the digest.

`EVP_DigestUpdate` 's documentation [stipulates](https://github.com/aws/aws-lc/blob/cf969b57ae1fcef117bbbe32a8a43f53389f71a8/include/openssl/digest.h#L165) an interface contract of "It returns one". Looking at [the implementation](https://github.com/aws/aws-lc/blob/cf969b57ae1fcef117bbbe32a8a43f53389f71a8/crypto/fipsmodule/digest/digest.c#L272), the only case where it can break that contract is if an `update` function pointer isn't configured on `ctx`. I believe we can trust that to [always be populated](https://github.com/aws/aws-lc/blob/cf969b57ae1fcef117bbbe32a8a43f53389f71a8/crypto/fipsmodule/digest/digest.c#L234) as long as `ctx` has been initialized by `EVP_DigestInit_ex`.

So, in other words, it looks like `EVP_DigestUpdate` can only fail if `ctx` hasn't been initialized. This makes me consider... should we fail `b`'s reads/writes before `next`'s IO if `b` hasn't been initialized? this would address your concern here and make it harder to misuse the `BIO_f_md`.

crypto/pkcs7/bio/md.c
WillChilds-Klein: Ah, ok. That makes sense given that we otherwise rely on `BIO_set_md` to init the BIO state. will replace.

crypto/pkcs7/bio/md.c
WillChilds-Klein: On second thought, I don't think we should do this. The current behavior allows the caller to reset the BIO with `BIO_reset` and skip additional `BIO_set_md` calls. un-init'ing here would require the caller to call `BIO_set_md` after `BIO_reset`, hampering compatibility with OpenSSL.

crypto/pkcs7/bio/md.c
WillChilds-Klein: Per discussion [here](https://github.com/aws/aws-lc/pull/1886#discussion_r1799791769), we now require the BIO to be initialized before it can do any IO.

WillChilds-Klein: hm, at any rate, it looks like IO on uninitialized BIO's is prevented higher in the call stack [in bio.c](https://github.com/aws/aws-lc/blob/460a9dd71dbe87554526949c40ec516bebe092ac/crypto/bio/bio.c#L175).

crypto/pkcs7/bio/md.c
```diff
@@ -0,0 +1,174 @@
...
+  if ((ctx == NULL) || (next == NULL) || inl <= 0) {
+    return 0;
+  }
+
+  ret = BIO_write(next, in, inl);
+  if (ret > 0) {
+    if (!EVP_DigestUpdate(ctx, (const unsigned char *)in, ret)) {
+      BIO_clear_retry_flags(b);
+      return 0;
+    }
```
justsmth, line 96: NP: Since this is a non-recoverable error, shouldn't the return value be negative? The `BIO_write` documentation seems to say that.
> // returns the value from calling the |callback_ex|, otherwise |BIO_write|
> // returns the number of bytes written, or a negative number on error.

I see OpenSSL returns 0 for this, so compatibility might dictate we do the same. (I'm also bothered that the we're unable to inform the caller about the actual number of bytes written to `next` b/c we can't return a value indicating success. But that's just the nature of this API.)

justsmth (dismissed):

crypto/pkcs7/bio/md.c
```diff
@@ -0,0 +1,174 @@
...
+#include "../internal.h"
+
+// |md_write| and |md_read| both contribute to the digest, |md_gets| returns
+// digest contents
+static int md_write(BIO *h, char const *buf, int num);
+static int md_read(BIO *h, char *buf, int size);
+static int md_gets(BIO *h, char *str, int size);
+static long md_ctrl(BIO *h, int cmd, long arg1, void *arg2);
+static int md_new(BIO *h);
+static int md_free(BIO *data);
```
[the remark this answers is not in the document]
WillChilds-Klein: sure.

crypto/pkcs7/bio/md.c
WillChilds-Klein: As discussed elswhere, EVP_DigestUpdate can only fail if `ctx` is uninitialized here, but greed that we should return -1 and be consistent with `md_read` above.

justsmth (approved):

justsmth (approved):

crypto/pkcs7/internal.h
```diff
@@ -199,6 +199,16 @@ int pkcs7_add_signed_data(CBB *out,
...
+// BIO_f_md is used internally by the pkcs7 module. It is not recommended
+// for external use. The BIO must be initialized with |BIO_set_md| or
+// |BIO_get_md_ctx| before it can be used.
+OPENSSL_EXPORT const BIO_METHOD *BIO_f_md(void);
+
+// BIO_get_md_ctx writes a reference of |b|'s EVP_MD_CTX* to |*mdcp|
+#define BIO_get_md_ctx(b, mdcp) BIO_ctrl(b, BIO_C_GET_MD_CTX, 0, mdcp)
+
+// BIO_set_md set's |b|'s EVP_MD* to |md|
+#define BIO_set_md(b, md) BIO_ctrl(b, BIO_C_SET_MD, 0, md)
```
samuel40791765, line 211: Nit: Could we do a direct function call instead of macros here? Similar to what https://github.com/aws/aws-lc/blob/8d9809ec06c3a3a350f8978d583e1fc324ec0cbc/crypto/bio/bio.c#L503-L505 is doing.

Macros allow for [potential type mismatches](https://github.com/aws/aws-lc/blame/8d9809ec06c3a3a350f8978d583e1fc324ec0cbc/include/openssl/bio.h#L451-L452) when passing parameters.

samuel40791765 (approved):

crypto/pkcs7/internal.h
WillChilds-Klein: hm, that's a good point. i'll adjust this in my next PR.

justsmth (approved):

samuel40791765 (approved):

PR #1887, opened 2024-09-27
Improve build and fix X509 test failures for Ruby

samuel40791765:
### Issues:
Resolves `CryptoAlg-2701`

### Description of changes: 
1. `bignum_to_string` is called when trying to print out the X509 Extension value. Initially I thought the value wasn't being set correctly, but it turns out that we were printing the value in hex form, rather than decimal (which OpenSSL does and Ruby expects). AWS-LC prints the hex value if the value is more than 32 bits, while OpenSSL has a much more lax restriction at 128 bits. Tweaking this to align with OpenSSL gets past the test for `test_x509crl.rb`. Tweaking the value to align with OpenSSL gets past the test.

   * AWS-LC: https://github.com/aws/aws-lc/blob/8ed554c8a3758df6879193936bac4dd3fc3882d5/crypto/x509/v3_utl.c#L168-L175
   * OpenSSL: https://github.com/openssl/openssl/blob/master/crypto/x509/v3_utl.c#L130C14-L140

2. Great news is I don't think we need to do anything for the test failure in `test_x509req`. This was a testing mistake on Ruby's end, documented in this commit: https://github.com/ruby/ruby/commit/6b120135afe1529a8ed1532b9da6878f1f4b1fbf
Only version 1 is available for CSRs and Ruby attempts to set an invalid version in its tests. OpenSSL 3.3 disallows this behavior now and Ruby has removed the test in it's mainline branch. We can brush up the patch to account for this and skip the test with AWS-LC.

3. I also took the chance to add back some of the defines Ruby depends on as no-ops. The X509 defines aren't actually used in neither OpenSSL or AWS-LC as found by this commit: https://github.com/aws/aws-lc/commit/496838a41073951d1db831dac5156a0bc9f2cc88 

### Call-outs:
N/A

### Testing:
N/A


crypto/x509/x509_test.cc
```diff
@@ -3190,9 +3190,11 @@ TEST(X509Test, PrettyPrintIntegers) {
       "-42",
       "256",
       "-256",
+      "4886718345",
+      "-4886718345",
       // Large numbers are pretty-printed in hex to avoid taking quadratic time.
-      "0x0123456789",
-      "-0x0123456789",
+      "0x0123456789012345678901234567890123456789",
+      "-0x0123456789012345678901234567890123456789",
```
WillChilds-Klein, line 3197: minor nit: you should only need 32 hex digits to hit your threshold of 2^128

include/openssl/ocsp.h
```diff
@@ -58,6 +62,9 @@ extern "C" {
 // Certificates included within |bs| or |req| will be included in the
 // search for the signing certificate by default, unless |OCSP_NOINTERN| is set.
 #define OCSP_NOINTERN 0x2
+// OCSP_NOSIGS does nothing. In OpenSSL, this skips signature verification in
+// |OCSP_basic_verify| and |OCSP_request_verify|.
```
WillChilds-Klein, line 66: so we'll _always_ check the signatures, right?

WillChilds-Klein (dismissed):

crypto/x509/x509_test.cc
samuel40791765: True, I just wanted the numbers to align. Will change

include/openssl/ocsp.h
samuel40791765: Yes, the intention was to not do this until it's specifically asked for. There's not really a scenario where the signature shouldn't be verified.

skmcgrail (approved):

WillChilds-Klein (approved):

PR #1897, opened 2024-10-01
Integrate ube for rand engine

torben-hansen:
### Description of changes: 

This PR integrates protection of the thread-local state into the new randomness generation implementation. `rand_ensure_ctr_drbg_uniquness()` is the function that determines whether a randomization of the thread-local state is necessary. `rand_ensure_ctr_drbg_uniquness()` is called inline in the core randomness generation code path (in `RAND_bytes_core()` and invoked on every entry.

The only mechanism currently implemented that can force a randomization is the UBE mechanism implemented in https://github.com/aws/aws-lc/commit/bc7aeff2f5c12c62f17df71abdd0c7b55c3eb984. Note that if UBE is "unavailable" then a randomization is forced every time.

### Testing:

Some additional support code is implemented to mock the UBE detection and determine if correct behaviour occurred.

The failing tests are because the target branch is `randomness_generation` that haven't been rebased on `main` branch with fixes.

### DIFF
--- a/crypto/fipsmodule/rand/new_rand.c
+++ b/crypto/fipsmodule/rand/new_rand.c
@@ -27,7 +27,7 @@ struct rand_thread_local_state {
   // |drbg| since its initialization.
   uint64_t reseed_calls_since_initialization;
 
-  // generate_number caches the UBE generation number.
+  // generation_number caches the UBE generation number.
   uint64_t generation_number;
 
   // Entropy source. UBE volatile state.
@@ -71,17 +71,17 @@ static int rand_ensure_valid_state(struct rand_thread_local_state *state) {
   return 1;
 }
 
-// rand_ensure_ctr_drbg_uniquness computes whether |state| must be randomized to
-// ensure uniqueness.
+// rand_ensure_ctr_drbg_uniqueness computes whether |state| must be randomized
+// to ensure uniqueness.
 //
-// Note: If |rand_ensure_ctr_drbg_uniquness| returns 1 it does not necessarily
+// Note: If |rand_ensure_ctr_drbg_uniqueness| returns 1 it does not necessarily
 // imply that an UBE occurred. It can also mean that no UBE detection is
 // supported or that UBE detection failed. In these cases, |state| must also be
 // randomized to ensure uniqueness. Any special future cases can be handled in
 // this function. 
 //
 // Return 1 if |state| must be randomized. 0 otherwise.
-static int rand_ensure_ctr_drbg_uniquness(struct rand_thread_local_state *state) {
+static int rand_ensure_ctr_drbg_uniqueness(struct rand_thread_local_state *state) {
 
   uint64_t current_generation_number = 0;
   if (CRYPTO_get_ube_generation_number(&current_generation_number) != 1) {
@@ -358,7 +358,7 @@ int NR_PREFIX(RAND_pseudo_bytes)(uint8_t *out, size_t out_len) {
   return NR_PREFIX(RAND_bytes)(out, out_len);
 }
 
-// Returns the number of generate calls made using the thread-local state since
+// Returns the number of generate calls made on the thread-local state since
 // last seed/reseed. Returns 0 if thread-local state has not been initialized.
 uint64_t get_thread_generate_calls_since_seed(void) {
 
@@ -371,8 +371,8 @@ uint64_t get_thread_generate_calls_since_seed(void) {
   return state->generate_calls_since_seed;
 }
 
-// Returns the number of generate calls made using the thread-local state since
-// last seed/reseed. Returns 0 if thread-local state has not been initialized.
+// Returns the number of reseed calls made on the thread-local state since
+// initialization. Returns 0 if thread-local state has not been initialized.
 uint64_t get_thread_reseed_calls_since_initialization(void) {
 
   struct rand_thread_local_state *state =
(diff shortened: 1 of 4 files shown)


crypto/fipsmodule/rand/new_rand.c
```diff
@@ -50,10 +54,27 @@ static int rand_ensure_valid_state(void) {
...
-static int rand_ensure_ctr_drbg_uniquness(struct rand_thread_local_state *state,
-  size_t out_len) {
+// rand_ensure_ctr_drbg_uniquness computes whether |state| must be randomized to
+// ensure uniqueness.
+//
+// Note: If |rand_ensure_ctr_drbg_uniquness| returns 0 it does not necessarily
+// imply that an UBE occurred. It can also mean that no UBE detection is
+// supported or that UBE detection failed. In these cases, |state| must also be
+// randomized to ensure uniqueness. Any special future cases can be handled in
+// this function.
```
andrewhop, line 64: How can the caller determine if 0 means no UBE (do not randomize state), or if 0 means no UBE supported (must randomize state)? Would returning 1 be safer, so callers know if this returns 1 they must randomize regardless of reason.

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -50,10 +54,27 @@ static int rand_ensure_valid_state(void) {
...
+
+  uint64_t current_generation_number = 0;
+  if (CRYPTO_get_ube_generation_number(&current_generation_number) != 1) {
+    return 0;
+  }
+
+  if (current_generation_number != state->generation_number) {
+    state->generation_number = current_generation_number;
+    return 0;
+  }
```
andrewhop, line 77: Shouldn't this return 1? Isn't this saying the generation number is different because of an UBE? 
> Return 1 if |state| must be randomized. 0 otherwise.

crypto/fipsmodule/rand/new_rand.c
torben-hansen: My bad, it's a typo. The `0` in line 60 should be a 1...

crypto/fipsmodule/rand/new_rand.c
torben-hansen: Yes, indeed, I screwed this up and you reviewed before I could add some tests (:

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -22,6 +23,13 @@ struct rand_thread_local_state {
   // since it was last (re)seeded. Must be bounded by |kReseedInterval|.
   uint64_t generate_calls_since_seed;
 
+  // reseed_calls_since_initialization is the number of reseed calls made of
+  // |drbg| since its initialization.
+  uint64_t reseed_calls_since_initialization;
+
+  // generate_number caches the UBE generation number.
```
dkostic, line 30: nit:
```suggestion
  // generation_number caches the UBE generation number.
```

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -44,18 +52,48 @@ static void rand_thread_local_state_free(void *state_in) {
...
+// ensure uniqueness.
+//
+// Note: If |rand_ensure_ctr_drbg_uniquness| returns 1 it does not necessarily
+// imply that an UBE occurred. It can also mean that no UBE detection is
+// supported or that UBE detection failed. In these cases, |state| must also be
+// randomized to ensure uniqueness. Any special future cases can be handled in
+// this function. 
+//
+// Return 1 if |state| must be randomized. 0 otherwise.
+static int rand_ensure_ctr_drbg_uniquness(struct rand_thread_local_state *state) {
```
dkostic, line 84: nit:
```suggestion
static int rand_ensure_ctr_drbg_uniqueness(struct rand_thread_local_state *state) {
```

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -316,3 +357,29 @@ int NR_PREFIX(RAND_priv_bytes)(uint8_t *out, size_t out_len) {
...
+      CRYPTO_get_thread_local(OPENSSL_THREAD_LOCAL_PRIVATE_RAND);
+  if (state == NULL) {
+    return 0;
+  }
+
+  return state->generate_calls_since_seed;
+}
+
+// Returns the number of generate calls made using the thread-local state since
+// last seed/reseed. Returns 0 if thread-local state has not been initialized.
```
dkostic, line 375:
```suggestion
// initialization. Returns 0 if thread-local state has not been initialized.
```

dkostic (approved):

crypto/fipsmodule/rand/new_rand.c
```diff
@@ -252,7 +293,7 @@ static void RAND_bytes_core(
 
   OPENSSL_cleanse(pred_resistance, RAND_PRED_RESISTANCE_LEN);
 
-  if (rand_ensure_valid_state() != 1) {
+  if (rand_ensure_valid_state(state) != 1) {
```
andrewhop, line 296: NP: in a future PR can you add a comment here explaining why we need to abort. I get that it means a UBE occurred and if that happened after line 235 and before line 242 you could end up with duplicate state, we might just forget/think this looks weird in the future.

andrewhop (approved):

PR #1903, opened 2024-10-02
Remove special s2n-bignum symbol handling sauce from build

torben-hansen:
### Description of changes: 

Symbols from s2n-bignum needed special build shenanigans because they essentially are "third-party" (although still AWS owned) in the context of what code AWS-LC repository can touch. Therefore they didn't have the special "prefix" header files. This corrupts the prefix build in two ways that needed to be handled:
- references to s2n-bignum files are prefixed, but the definitions are not. Obviously, this doesn't work...
- when we test the prefix build, we must take care to not assume s2n-bignum symbols are prefixed...

Removing these two corruptions, means it's now a bit easier to import s2n-bignum code. A small step closer to maybe switching to a submodule.

Should also make the rust build easier. Hopefully...

### Call-outs:

Originally I went with a method that required s2n-bignum source code changes. Essentially, one can define a path to prefix symbols and inject into the pre-processor we run s2n-bignum files through. The benefit is that we don't have to reason about tooling. It's just simple macro gymnastics.

Andrew suggested `--include` flag as a way to not do s2n-bignum code changes. That seems a bit easier for now. We can revert to another more portable method if required. Having more AWS-LC special sauce in s2n-bignum is not the preferred for sure.

### Testing:

CI has several test suites for the prefix build.

### DIFF
--- a/third_party/s2n-bignum/include/_internal_s2n_bignum.h
+++ b/third_party/s2n-bignum/include/_internal_s2n_bignum.h
@@ -15,3 +15,7 @@
 #else
 #   define S2N_BN_SYM_PRIVACY_DIRECTIVE(name)  /* NO-OP: S2N_BN_SYM_PRIVACY_DIRECTIVE */
 #endif
+
+#ifdef AWSLC_PRIVATE_BUILD
+    #include <openssl/boringssl_prefix_symbols.h>
+#endif
--- a/third_party/s2n-bignum/include/_internal_s2n_bignum.h
+++ b/third_party/s2n-bignum/include/_internal_s2n_bignum.h
@@ -15,7 +15,3 @@
 #else
 #   define S2N_BN_SYM_PRIVACY_DIRECTIVE(name)  /* NO-OP: S2N_BN_SYM_PRIVACY_DIRECTIVE */
 #endif
-
-#ifdef AWSLC_PRIVATE_BUILD
-    #include <openssl/boringssl_prefix_symbols.h>
-#endif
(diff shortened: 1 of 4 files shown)


third_party/s2n-bignum/include/_internal_s2n_bignum.h
```diff
@@ -14,4 +14,8 @@
 #   endif
 #else
 #   define S2N_BN_SYM_PRIVACY_DIRECTIVE(name)  /* NO-OP: S2N_BN_SYM_PRIVACY_DIRECTIVE */
-#endif
\ No newline at end of file
+#endif
+
+#ifdef AWSLC_PRIVATE_BUILD
+    #include <openssl/boringssl_prefix_symbols.h>
+#endif
```
andrewhop, line 21: Can you always include this header file? It should either be filled with the prefixes, or be an empty file if it's not a prefix build. 

Or is this file supposed to come from s2n-bignum and be generic? If so can you solve it by including it in S2N_BIGNUM_PREFIX_INCLUDE?

>-include file
>    Process file as if "#include "file"" appeared as the first line of the primary source file. However, the first directory searched for file is the preprocessor's working directory instead of the directory containing the main source file. If not found there, it is searched for in the remainder of the "#include "..."" search chain as normal.

torben-hansen: You can't because s2n-bignum is not always build in the context of AWS-LC

andrewhop (approved):

skmcgrail (approved):

PR #1905, opened 2024-10-03
add support for EVP_PKEY_CTX callback functions

samuel40791765:
### Issues:
Resolves `CryptoAlg-1716`
Resolves `CryptoAlg-2698`

### Description of changes: 
We tried to no-op these functions, but it turns out Ruby depends on them pretty extensively as the interruption mechanism for threads. One of[ Ruby's tests](https://github.com/ruby/ruby/blame/ruby_3_1/test/openssl/test_pkey_rsa.rb#L73-L77) depends on `EVP_PKEY_CTX_get_app_data` to return an actual value from the callback function, but we return NULL as a no-op. Ruby seems to depend on the `EVP_PKEY` callback function and relevant application data to correctly handle interruptions. Based on the relevant commit messages, the expectation is that the operation is interrupted, but AWS-LC continues resuming the operation and returns a generated RSA key. It looks like we may have to consider implementing functionality for these callback functions. This issue also applies to a test failure in `test/openssl/test_pkey_dh.rb` and `test/openssl/test_pkey_dsa.rb`. We probably aren't going to support DSA, but this will need to be applied to DH somewhere down the line.

* Commits:
    * https://github.com/ruby/openssl/commit/88b90fb8568a9f104dfed1ace3b33b8586eafeed
    * https://github.com/ruby/ruby/commit/d3507e3ea697be7aab4e9344c379d6b277cf81cf
* Relevant code: https://github.com/ruby/ruby/blob/0209b9554dc86b95018b454d81ad48d1a9c6af27/ext/openssl/ossl_pkey.c#L282-L343 

### Call-outs:
N/A

### Testing:
new test that verifies this works with `EVP_PKEY_RSA`

### DIFF
--- a/crypto/evp_extra/evp_extra_test.cc
+++ b/crypto/evp_extra/evp_extra_test.cc
@@ -2936,3 +2936,62 @@ TEST_P(PerMLKEMTest, InputValidation) {
   ctx->pkey->pkey.kem_key->secret_key[GetParam().secret_key_len - 64] ^= 1;
   ASSERT_FALSE(EVP_PKEY_decapsulate(ctx.get(), ss_expected.data(), &ss_len, ct.data(), ct_len));
 }
+
+struct dummy_cb_app_data {
+  bool state;
+};
+
+// Dummy callback function used for testing.
+static int dummy_gen_cb(EVP_PKEY_CTX *ctx) {
+  // Get the application-specific data.
+  auto *app_data =
+      static_cast<dummy_cb_app_data *>(EVP_PKEY_CTX_get_app_data(ctx));
+  EXPECT_TRUE(app_data);
+
+  app_data->state = true;
+
+  return 1;  // Return success (1).
+}
+
+TEST(EVPExtraTest, Callbacks) {
+  bssl::UniquePtr<EVP_PKEY_CTX> ctx(EVP_PKEY_CTX_new_id(EVP_PKEY_RSA, nullptr));
+  ASSERT_TRUE(ctx);
+
+  // Check the initial values of |ctx->keygen_info|.
+  int keygen_info = EVP_PKEY_CTX_get_keygen_info(ctx.get(), -1);
+  ASSERT_EQ(keygen_info, EVP_PKEY_CTX_KEYGEN_INFO_COUNT);
+  for (int i = 0; i < keygen_info; i++) {
+    EXPECT_EQ(EVP_PKEY_CTX_get_keygen_info(ctx.get(), i), 0);
+  }
+
+  // Generating an RSA key would have triggered the callback.
+  EVP_PKEY *pkey = EVP_PKEY_new();
+  ASSERT_EQ(EVP_PKEY_keygen_init(ctx.get()), 1);
+  ASSERT_TRUE(EVP_PKEY_keygen(ctx.get(), &pkey));
+  ASSERT_TRUE(pkey);
+
+  // Verify that |ctx->keygen_info| has not been updated since a callback hasn't
+  // been set.
+  for (int i = 0; i < keygen_info; i++) {
+    EXPECT_EQ(EVP_PKEY_CTX_get_keygen_info(ctx.get(), i), 0);
+  }
+
+  // Now we set the application data and callback.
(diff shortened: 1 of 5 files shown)


crypto/fipsmodule/evp/evp_ctx.c
```diff
@@ -669,21 +669,37 @@ int EVP_PKEY_CTX_ctrl_str(EVP_PKEY_CTX *ctx, const char *name,
...
-  return NULL;
+  return ctx->app_data;
 }
 
 int EVP_PKEY_CTX_get_keygen_info(EVP_PKEY_CTX *ctx, int idx) {
-  // No-op
-  return 0;
+  if (idx == -1) {
+    return EVP_PKEY_CTX_KEYGEN_INFO_COUNT;
+  }
```
skmcgrail, line 700:
```suggestion
  GUARD_PTR(ctx);
  if (idx == -1) {
    return EVP_PKEY_CTX_KEYGEN_INFO_COUNT;
  }
```

crypto/fipsmodule/evp/evp_ctx.c
```diff
@@ -669,21 +669,38 @@ int EVP_PKEY_CTX_ctrl_str(EVP_PKEY_CTX *ctx, const char *name,
...
+  return ctx->app_data;
 }
 
 int EVP_PKEY_CTX_get_keygen_info(EVP_PKEY_CTX *ctx, int idx) {
-  // No-op
-  return 0;
+  GUARD_PTR(ctx);
+  if (idx == -1) {
+    return EVP_PKEY_CTX_KEYGEN_INFO_COUNT;
+  }
```
justsmth, line 701: Should this return 0 if `ctx->operation != EVP_PKEY_OP_KEYGEN`?

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -2936,3 +2936,62 @@ TEST_P(PerMLKEMTest, InputValidation) {
...
+  app_data->state = true;
+
+  return 1;  // Return success (1).
+}
+
+TEST(EVPExtraTest, Callbacks) {
+  bssl::UniquePtr<EVP_PKEY_CTX> ctx(EVP_PKEY_CTX_new_id(EVP_PKEY_RSA, nullptr));
+  ASSERT_TRUE(ctx);
+
+  // Check the initial values of |ctx->keygen_info|.
```
justsmth, line 2960: Should the keygen operation be initialized (`EVP_PKEY_keygen_init`) prior to this check?

crypto/fipsmodule/evp/evp_ctx.c
```diff
@@ -669,21 +669,38 @@ int EVP_PKEY_CTX_ctrl_str(EVP_PKEY_CTX *ctx, const char *name,
...
 void EVP_PKEY_CTX_set_app_data(EVP_PKEY_CTX *ctx, void *data) {
-  // No-op
+  ctx->app_data = data;
 }
 
 void *EVP_PKEY_CTX_get_app_data(EVP_PKEY_CTX *ctx) {
-  // No-op
-  return NULL;
+  return ctx->app_data;
 }
```
justsmth, line 695: These functions should also guard against `ctx == NULL`.

crypto/fipsmodule/evp/evp_ctx.c
samuel40791765: 1. Sure, I think that makes sense. I initially tried returning 0 if the key type was not RSA or DH, but that didn't work because there was the possibility where `EVP_PKEY_CTX` did not have a set type yet. This should be a good way to enforce something similar.
2. My interpretation of the array and how it's used in OpenSSL is that it's only used to map to the arguments available in `BN_GENCB`. Each operation that uses this only has 2 available arguments to configure, so I thought it was safe to lock this as `2`.
I did document some parts of this in `EVP_PKEY_CTX_get_keygen_info`, but forget to mention how I came to the conclusion for this variable. Will add!

skmcgrail (approved):

justsmth (approved):

PR #1911, opened 2024-10-09
Avoid allocating EVP_PKEY on size checks

geedo0:
### Issues:
PQCrypto-103 

### Description of changes: 
`EVP_PKEY_keygen_deterministic` allows callers to query the required
size for the `seed` parameter. This change avoids the unnecessary
allocation of memory for the `out_pkey` parameter to ensure that this
integer getter has no side effects.

### Call-outs:

### Testing:

Updated the unit tests.

### DIFF
--- a/include/openssl/experimental/kem_deterministic_api.h
+++ b/include/openssl/experimental/kem_deterministic_api.h
@@ -19,9 +19,9 @@ extern "C" {
 // performs deterministic keygen based on the value specified in |seed| of
 // length |seed_len| bytes.
 //
-// If |seed| is set to NULL it is assumed that the caller is doing a size check:
-// the function will write the size of the required seed in |seed_len| and return
-// successfully.
+// If |out_pkey| and |seed| are set to NULL it is assumed that the caller is
+// doing a size check and the function will write the size of the required seed
+// in |seed_len| and return successfully.
 //
 // EVP_PKEY_keygen_deterministic performs a deterministic key generation
 // operation using the values from |ctx|, and the given |seed|. If |*out_pkey|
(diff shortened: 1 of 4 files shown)


include/openssl/experimental/kem_deterministic_api.h
jakemas, line 1: A call out to the long-term plan for the `include/openssl/experimental/` directory. We originally created this location to home `experimental` APIs for ML-KEM, as we were unsure on the exact requirements around FIPS 204 on allowing access to de-randomized modes. As FIPS 204 is now finalized we can reason about exact language.

FIPS 204 Sec 3.3 Page 16 reads:

```
Controlled access to internal functions. The key-encapsulation mechanism ML-KEM makes use
of internal, “derandomized” functions ML-KEM.KeyGen_internal, ML-KEM.Encaps_internal, and
ML-KEM.Decaps_internal, specified in Section 6. The interfaces for these functions should not
be made available to applications other than for testing purposes. In particular, the sampling of
random values required for key generation (as specified in ML-KEM.KeyGen) and encapsulation
(as specified in ML-KEM.Encaps) shall be performed by the cryptographic module.
```

i.e., that internal/de-randomized APIs such as these *should* not be made available to applications other than for testing purposes. While I am not prescribing any particular change in this PR, I wanted to make the call out that we (1) may like to add documentation from FIPS204 regarding the use of these APIs, (2) we may like to change the long-term location of these APIs now that FIPS204 is finalized.

andrewhop (approved):

include/openssl/experimental/kem_deterministic_api.h
geedo0: I don't think we should mix these documentation concerns into this PR. However, there's certainly a discussion to be had on the many facets of this thread. A comment thread in this ancillary PR may not be the proper venue, but some questions that come to mind for me...
- Should we limit adoption of these experimental APIs so that we can prevent complications in any inevitable code motions?
- What exactly is meant by NIST's clause on testing? How will we choose to interpret that? In the strongSwan case, they are using it for *their* testing. Is that okay? Or does NIST mean for this to specifically refer to CAVP testing within the module itself?
- Is there any new thinking around where these APIs will land today now that we have FIPS 204?

WillChilds-Klein (approved):

PR #1912, opened 2024-10-09
Handle Windows not supporting static array dimension

torben-hansen:
### Issues:

CryptoAlg-2241

### Description of changes: 

https://github.com/awslabs/s2n-bignum/pull/98 handles Windows not supporting static array dimensions. We have our own header files with function declarations for s2n-bignum. Hence that work was never ported automatically.

Adopt same detection approach as in referenced PR. MSVC doesn't support it, neither does C++. So, try to detect those cases as well as if variable-length arrays aren't supported.

#### Previous solution
Support it now but use our own macro `OEPNSSL_WINDOWS` to determine whether we are executing a Windows build. Take a dependency on `target.h` because this is minimal. Also teach the s2n-bignum file pro-processing where to find `target.h` even though I don't think this is strictly necessary; non of the pre-processed files include the header file and everywhere the header file is included will have at least `base.h`. But do it anyway. 

### DIFF
--- a/third_party/s2n-bignum/include/s2n-bignum_aws-lc.h
+++ b/third_party/s2n-bignum/include/s2n-bignum_aws-lc.h
@@ -15,6 +15,14 @@
 #ifndef S2N_BIGNUM_AWS_LC_H
 #define S2N_BIGNUM_AWS_LC_H
 
+#include <openssl/target.h>
+
+#if defined(OPENSSL_WINDOWS)
+  #define S2N_BIGNUM_STATIC
+#else
+  #define S2N_BIGNUM_STATIC static
+#endif
+
 // ----------------------------------------------------------------------------
 // C prototypes for s2n-bignum functions used in AWS-LC
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


third_party/s2n-bignum/include/s2n-bignum_aws-lc.h
```diff
@@ -15,6 +15,14 @@
 #ifndef S2N_BIGNUM_AWS_LC_H
 #define S2N_BIGNUM_AWS_LC_H
 
+#include <openssl/target.h>
+
+#if defined(OPENSSL_WINDOWS)
```
andrewhop, line 20: Is the issue Windows or MSVC? Do other compilers on Windows support static?

torben-hansen: Okay, adopted the approach I implemented for https://github.com/awslabs/s2n-bignum/pull/98.

andrewhop (approved):

justsmth (approved):

PR #1913, opened 2024-10-09
build: fix pkgconfig files

theoparis:
### Description of changes: 

Currently the pkgconfig files are broken when aws-lc is built with Nix. 

See https://github.com/jtojnar/cmake-snips?tab=readme-ov-file#concatenating-paths-when-building-pkg-config-files, and the PR made to libfmt https://github.com/fmtlib/fmt/pull/1657

### Call-outs:

I'm not sure if bumping the minimum required CMake version is preferred. The alternative would be to add https://github.com/jtojnar/cmake-snips/blob/master/CMakeScripts/JoinPaths.cmake to the `cmake/` directory in the repository. However, most linux distributions have at least CMake 3.20, if not 3.25 so I don't see why this would be a problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1258,6 +1258,11 @@ set(VERSION 1.1.1)
 # the downstream integration may build with the system's OpenSSL version instead.
 # Consider adjusting the PKG_CONFIG_PATH environment to get around this.
 file(GLOB OPENSSL_PKGCONFIGS "pkgconfig/*.pc.in")
+
+include(cmake/JoinPaths.cmake)
+join_paths(libdir_for_pc_file "\${prefix}" "${CMAKE_INSTALL_LIBDIR}")
+join_paths(includedir_for_pc_file "\${prefix}" "${CMAKE_INSTALL_INCLUDEDIR}")
+
 foreach(in_file ${OPENSSL_PKGCONFIGS})
   file(RELATIVE_PATH in_file ${PROJECT_SOURCE_DIR} ${in_file})
   string(REPLACE ".in" "" pc_file ${in_file})
--- a/cmake/JoinPaths.cmake
+++ b/cmake/JoinPaths.cmake
@@ -1,0 +1,23 @@
+# This module provides function for joining paths
+# known from most languages
+#
+# SPDX-License-Identifier: (MIT OR CC0-1.0)
+# Copyright 2020 Jan Tojnar
+# https://github.com/jtojnar/cmake-snips
+#
+# Modelled after Python’s os.path.join
+# https://docs.python.org/3.7/library/os.path.html#os.path.join
+function(join_paths joined_path first_path_segment)
+  set(temp_path "${first_path_segment}")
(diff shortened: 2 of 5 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
```
justsmth, line 1: We support a few consumers with build environments that have older versions of CMake. (We have CI to ensure that [our build succeeds with CMake 3.2](https://github.com/aws/aws-lc/blob/main/.github/workflows/cmake.yml).)

Can this change be made in a way that would be supported by older versions of CMake?

cmake/JoinPaths.cmake
```diff
@@ -0,0 +1,13 @@
+function(join_paths joined_path first_path_segment)
```
justsmth, line 1: This file should have a copyright and a license statement (copied from the source if it's not original to you).

cmake/JoinPaths.cmake
```diff
@@ -0,0 +1,22 @@
+# This module provides function for joining paths
+# known from most languages
+#
+# SPDX-License-Identifier: (MIT OR CC0-1.0)
```
justsmth, line 4: A note for reference: Here's a link to the CC0-1.0 license: https://creativecommons.org/publicdomain/zero/1.0/

cmake/JoinPaths.cmake
```diff
@@ -0,0 +1,22 @@
...
+  foreach(current_segment IN LISTS ARGN)
+    if(NOT ("${current_segment}" STREQUAL ""))
+      if(IS_ABSOLUTE "${current_segment}")
+        set(temp_path "${current_segment}")
+      else()
+        set(temp_path "${temp_path}/${current_segment}")
+      endif()
+    endif()
+  endforeach()
+  set(${joined_path} "${temp_path}" PARENT_SCOPE)
```
justsmth, line 21: Since the path separators here are specific to unix-y environments, could we try to ["fix" the joined path](https://cmake.org/cmake/help/latest/command/file.html#to-native-path) prior to return? Perhaps, something like: 
```
file(TO_NATIVE_PATH "${joined_path}" joined_path)
```

justsmth:
Hello @theoparis -- thanks for the PR!

Is there an easy way to test this in our CI, so we can avoid a regression in the future?

cmake/JoinPaths.cmake
theoparis: I was also wondering how I could add Windows support - thanks!

theoparis:
Hmm, I'm not sure what the best way of testing this kind of change would be. I made a [Nix derivation here](https://github.com/tinted-software/nixpkgs/blob/theoparis/aws-lc/pkgs/by-name/aw/aws-lc/package.nix). Maybe it could be tested via a Nix github action?

skmcgrail (approved):

justsmth (approved):

PR #1915, opened 2024-10-11
Missing functionality + Adding Nmap to our CI

smittals2:
### Issues:
CryptoAlg-2552

### Description of changes: 
This PR adds Nmap tip of main to our CI. It includes: 
1. A patch to Nmap
2. Adding an include for <errno.h> to <err.h>, OpenSSL does this but we previously did not
3. Adding BIO_destroy_bio_pair (the functionality was already implemented, we are simply exposing that through a new func)
4. Creating a new alias for "DES" in the EVP_CIPHER API. The "default" mode used in OpenSSL 1.1.1 and 3.1 with DES is CBC [(link)](https://docs.openssl.org/1.1.1/man1/enc/#supported-ciphers:~:text=des%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20Alias%20for%20des%2Dcbc).

### Testing:
Locally built and ran nmap tests with AWS-LC and verified they pass. Conducted a dry run with the extra included file to ensure it doesn't break any consumers in live (refer to CryptoAlg-2552 for the dryrun link).

### DIFF
--- a/.github/workflows/integrations.yml
+++ b/.github/workflows/integrations.yml
@@ -98,6 +98,18 @@ jobs:
       - name: Run integration build
         run: |
           ./tests/ci/integration/run_socat_integration.sh
+  nmap:
+    if: github.repository_owner == 'aws'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install OS Dependencies
+        run: |
+          sudo apt-get update -o Acquire::Languages=none -o Acquire::Translation=none
+          sudo apt-get -y --no-install-recommends install cmake gcc ninja-build golang make libcairo2.dev gobject-introspection libgirepository1.0-dev
+      - uses: actions/checkout@v3
+      - name: Run nmap build
+        run: |
+          ./tests/ci/integration/run_nmap_integration.sh
   python-main:
     if: github.repository_owner == 'aws'
     runs-on: ubuntu-latest
--- a/.github/workflows/integrations.yml
+++ b/.github/workflows/integrations.yml
@@ -105,7 +105,7 @@ jobs:
       - name: Install OS Dependencies
         run: |
           sudo apt-get update -o Acquire::Languages=none -o Acquire::Translation=none
-          sudo apt-get -y --no-install-recommends install cmake gcc ninja-build golang make libcairo2.dev gir1.2-gtk-3.0 python3-gi gobject-introspection libgirepository1.0-dev
+          sudo apt-get -y --no-install-recommends install cmake gcc ninja-build golang make cairo libcario2 libcairo2-dev gir1.2-gtk-3.0 python3-gi gobject-introspection libgirepository1.0-dev
       - uses: actions/checkout@v3
       - name: Run nmap build
         run: |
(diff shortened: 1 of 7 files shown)


crypto/bio/pair.c
```diff
@@ -467,6 +467,11 @@ int BIO_new_bio_pair(BIO** bio1_p, size_t writebuf1_len,
   return 1;
 }
 
+int BIO_destroy_bio_pair(BIO *b) {
+  bio_destroy_pair(b);
+  return 1;
+}
```
andrewhop, line 473: Not handling a null BIO.

crypto/cipher_extra/cipher_extra.c
```diff
@@ -106,6 +106,7 @@ static const struct {
   const char* name;
 } kCipherAliases[] = {
```
andrewhop, line 107: What uses these aliases? Does adding DES here enable it to be used silently by default anywhere?

.github/workflows/integrations.yml
```diff
@@ -98,6 +98,18 @@ jobs:
...
         run: |
           ./tests/ci/integration/run_socat_integration.sh
+  nmap:
+    if: github.repository_owner == 'aws'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install OS Dependencies
+        run: |
+          sudo apt-get update -o Acquire::Languages=none -o Acquire::Translation=none
+          sudo apt-get -y --no-install-recommends install cmake gcc ninja-build golang make cairo libcario2 libcairo2-dev gir1.2-gtk-3.0 python3-gi gobject-introspection libgirepository1.0-dev
```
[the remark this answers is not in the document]
smittals2: Hmmm interesting I've been testing this on a ubuntu system as well. Since the packages aren't available in the github runner I can disable zenmap but I wonder if that leaves a compatibility gap? I guess it's just the UI though. (could try pip installing them as a part of the integration script since they aren't available through the system as well).

crypto/cipher_extra/cipher_extra.c
smittals2: 1. Adding this mapping does not "leak" anything new. We already expose des-cbc through the kCiphers array. This change adds another alias pointing "DES" to this mapping (cbc is the default mode for DES in OpenSSL). 
2. These aliases are checked in addition to the kCiphers array. This is due to differences in the string names applications try to fetch algorithms with. In general these string:EVP_CIPHER mappings are used by the EVP_get_cipherbyname function. This function is part of our public API, although deprecated, which means users can directly retrieve the EVP_des_cbc cipher object and then conduct operations with it. Internally I don't see any usages in our codebase other than testing.

.github/workflows/integrations.yml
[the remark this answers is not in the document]
WillChilds-Klein: i don't think we need to worry about the GUI until/if we learn of customer need.

WillChilds-Klein (approved):

andrewhop (approved):

PR #1916, opened 2024-10-11
Marshalling/Unmarshalling DH public keys

justsmth:
### Issues:
Addresses: CryptoAlg-2696

### Description of changes: 
* Support marshalling/unmarshalling of DH public keys.

### Testing:
Added test `DHTest.DHMarshalPubKey`

### DIFF
--- a/crypto/dh_extra/dh_test.cc
+++ b/crypto/dh_extra/dh_test.cc
@@ -1137,7 +1137,7 @@ TEST(DHTest, DHMarshalPubKey) {
     ASSERT_TRUE(BIO_flush(out_bio.get()));
     ASSERT_TRUE(BIO_mem_contents(out_bio.get(), &pubkey_pem, &pubkey_pem_len));
     ASSERT_GT(pubkey_pem_len, (size_t)0);
-    ASSERT_NE(pubkey_pem, nullptr);
+    ASSERT_TRUE(pubkey_pem);
     // We own the allocation after this
     pubkey_pem = (const uint8_t*)OPENSSL_memdup(pubkey_pem, pubkey_pem_len);
   }
--- a/crypto/dh_extra/dh_test.cc
+++ b/crypto/dh_extra/dh_test.cc
@@ -1051,3 +1051,110 @@ TEST(DHTest, DHCheckForStandardParams) {
   ASSERT_TRUE(DH_check(dh2.get(), &flags));
   EXPECT_EQ(flags, 0);
 }
+
+TEST(DHTest, DHMarshalPubKey) {
+  const char* dh512_pem =
+    "-----BEGIN DH PARAMETERS-----\n"
+    "MEYCQQDqvLe5oX3p+Dw8T7NWG7nlWVFK58Ev74xvxYH72DC4kqfPEFPvNnCpFoRB\n"
+    "RdxOz7DZ6JO/GxobSRyAAI766+GDAgEC\n"
+    "-----END DH PARAMETERS-----";
+  const uint64_t encoded_g = 2;
+  const char encoded_p_dec_str[] = "12294183602774786812319504504704470077603616440910559765086569005477513835495488680341310019770549315448633656928525381740662262980129138358936697450520963";
+
+  bssl::UniquePtr<EVP_PKEY> epkey_dh_params(nullptr);
+  {
+    const size_t pem_len = OPENSSL_strnlen(dh512_pem, 1024);
+    bssl::UniquePtr<BIO> in_bio(BIO_new_mem_buf(dh512_pem, pem_len));
+    ASSERT_TRUE(in_bio);
+
+    epkey_dh_params.reset(PEM_read_bio_Parameters(in_bio.get(), nullptr));
+    ASSERT_TRUE(epkey_dh_params);
+  }
+
+  // Sanity check the Param parsing
+  {
+    DH *dh_params = EVP_PKEY_get0_DH(epkey_dh_params.get());
+    const BIGNUM *p = DH_get0_p(dh_params);
+    const BIGNUM *g = DH_get0_g(dh_params);
+    uint64_t parsed_g = 0;
+    ASSERT_TRUE(BN_get_u64(g, &parsed_g));
+    ASSERT_EQ(parsed_g, encoded_g);
+    const char *parsed_p_dec_str = BN_bn2dec(p);
+    ASSERT_NE(parsed_p_dec_str, nullptr);
+    ASSERT_EQ(OPENSSL_strcasecmp(encoded_p_dec_str, parsed_p_dec_str), 0);
+    OPENSSL_free((void *)parsed_p_dec_str);
+  }
+
+  // Perform keygen operation
+  bssl::UniquePtr<EVP_PKEY> gen_dh(nullptr);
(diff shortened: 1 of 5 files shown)


crypto/evp_extra/p_dh_asn1.c
```diff
@@ -12,10 +12,71 @@
...
+  return 1;
+}
+
+static int dh_pub_decode(EVP_PKEY *out, CBS *params, CBS *key) {
+  // RFC 2786
+  BIGNUM *pubkey = NULL;
+  DH *dh = NULL;
+  if (out == NULL || params == NULL || CBS_len(params) == 0 || key == NULL ||
+      CBS_len(key) == 0) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_DECODE_ERROR);
```
WillChilds-Klein, line 46: minor: should this be an invalid params error?

crypto/evp_extra/p_dh_asn1.c
```diff
@@ -12,10 +12,71 @@
...
+  dh->pub_key = pubkey;
+
+  if (!EVP_PKEY_assign_DH(out, dh)) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_DECODE_ERROR);
+    goto err;
+  }
+  return 1;
+
+err:
+  DH_free(dh);
```
WillChilds-Klein, line 76: this only contains public materials, right?

why not use OPENSSL_free?

crypto/evp_extra/p_dh_asn1.c
```diff
@@ -12,10 +12,71 @@
...
+#include <openssl/x509.h>
 
-#include "internal.h"
-#include "../internal.h"
 #include "../fipsmodule/cpucap/internal.h"
+#include "../fipsmodule/dh/internal.h"
+#include "../internal.h"
+#include "internal.h"
+
+static int dh_pub_encode(CBB *out, const EVP_PKEY *key) {
```
WillChilds-Klein, line 22: should we null-check `key` before proceeding?

WillChilds-Klein (approved):
minor: should we cover any failure cases in parse testing?

crypto/evp_extra/p_dh_asn1.c
samuel40791765: I believe `DH_free` calls `OPENSSL_free` after freeing all it's other contents at the end
https://github.com/aws/aws-lc/blob/8ffe277c21915ca82dc78a3bdc6a92e10c284b92/crypto/fipsmodule/dh/dh.c#L99-L118

crypto/dh_extra/dh_test.cc
```diff
@@ -1051,3 +1051,110 @@ TEST(DHTest, DHCheckForStandardParams) {
...
+  const uint64_t encoded_g = 2;
+  const char encoded_p_dec_str[] = "12294183602774786812319504504704470077603616440910559765086569005477513835495488680341310019770549315448633656928525381740662262980129138358936697450520963";
+
+  bssl::UniquePtr<EVP_PKEY> epkey_dh_params(nullptr);
+  {
+    const size_t pem_len = OPENSSL_strnlen(dh512_pem, 1024);
+    bssl::UniquePtr<BIO> in_bio(BIO_new_mem_buf(dh512_pem, pem_len));
+    ASSERT_TRUE(in_bio);
+
+    epkey_dh_params.reset(PEM_read_bio_Parameters(in_bio.get(), nullptr));
```
samuel40791765, line 1070: Nit: We could just do
```suggestion
    bssl::UniquePtr<EVP_PKEY> epkey_dh_params(PEM_read_bio_Parameters(in_bio.get(), nullptr));
```

Another nit: Is there a particular reason for all these scope brackets in this test? They do seem a bit redundant.

crypto/dh_extra/dh_test.cc
```diff
@@ -1051,3 +1051,110 @@ TEST(DHTest, DHCheckForStandardParams) {
...
+  bssl::UniquePtr<EVP_PKEY> gen_dh(nullptr);
+  {
+    bssl::UniquePtr<EVP_PKEY_CTX> epkey_ctx(
+        EVP_PKEY_CTX_new(epkey_dh_params.get(), nullptr));
+    ASSERT_TRUE(epkey_ctx);
+
+    ASSERT_TRUE(EVP_PKEY_keygen_init(epkey_ctx.get()));
+    EVP_PKEY *gen_dh_raw = nullptr;
+    ASSERT_TRUE(EVP_PKEY_keygen(epkey_ctx.get(), &gen_dh_raw));
+    gen_dh.reset(gen_dh_raw);
```
samuel40791765, line 1098: Nit:
```suggestion
    bssl::UniquePtr<EVP_PKEY> gen_dh(gen_dh_raw);
```

crypto/evp_extra/p_dh_asn1.c
justsmth: That's a good point.  The caller [does not check for null](https://github.com/aws/aws-lc/blob/4916fe8641c5c220e6e6921defc026a8c564f855/crypto/evp_extra/evp_asn1.c#L144-L151) and neither does the [equivalent RSA function](https://github.com/aws/aws-lc/blob/4916fe8641c5c220e6e6921defc026a8c564f855/crypto/evp_extra/p_rsa_asn1.c#L69-L83). I think there should be a NULL check at the top-level.

crypto/dh_extra/dh_test.cc
```diff
@@ -1051,3 +1051,110 @@ TEST(DHTest, DHCheckForStandardParams) {
...
+  const uint8_t* pubkey_pem = NULL;
+  size_t pubkey_pem_len = 0;
+  {
+    bssl::UniquePtr<BIO> out_bio(BIO_new(BIO_s_mem()));
+    ASSERT_TRUE(out_bio);
+    ASSERT_TRUE(PEM_write_bio_PUBKEY(out_bio.get(), gen_dh.get()));
+    ASSERT_TRUE(BIO_flush(out_bio.get()));
+    ASSERT_TRUE(BIO_mem_contents(out_bio.get(), &pubkey_pem, &pubkey_pem_len));
+    ASSERT_GT(pubkey_pem_len, (size_t)0);
+    ASSERT_NE(pubkey_pem, nullptr);
```
samuel40791765, line 1140: Nit for consistency with our other assertions in this test. Also noticing inconsistencies in other places in the test as well
```suggestion
    ASSERT_TRUE(pubkey_pem);
```

crypto/dh_extra/dh_test.cc
justsmth: `epkey_dh_params` needs to be declared outside of the currently block/brackets so it's accessible by subsequent logic.

justsmth: > Another nit: Is there a particular reason for all these scope brackets in this test? They do seem a bit redundant.

I like to segment large test functions like this into smaller self-contained blocks. This reduces the "cognitive overhead" when reading the code, and allows for the use of simple variable names like `in_bio` and `out_bio`.

WillChilds-Klein (approved):

samuel40791765 (approved):

PR #1919, opened 2024-10-15
Globally clear thread-local states

torben-hansen:
### Description of changes: 

As per FIPS, something equivalent to sensitive memory must not be kept around in the clear. Often the sensitive stuff we produce is then handed off to a linked application. But we do have some memory we control, namely the thread-local CTR-DRBG states. It's not clear why CTR-DRBG states are actually sensitive for the purpose of zeroization; the primary security property of a DRBG is backtracking resistance, so someone would need to break the security of the scheme to take advantage of the state. Obviously, one can generate future outputs from a DRBG state, since it's deterministic. But that's not the purpose of zeroization. Regardless, zeroization is what we must do.

But when a process exists, all threads might not have been gracefully exited. To ensure this doesn't leave around sensitive memory, references to all thread-local states are kept in a doubly linked list. This creates a fair bit of issues that was mostly solved in the existing solution. I modified that a bit to take into account the new code-flow. The only major chance is that the CTR-DRBG generate call is now wrapped in a global lock under the while-loop iteration. I have tried to explain most in inline comments in the code. One can argue that when existing a process there is no need for this ands it is all just unreasonable to require, but that's the state of the world currently. And so, must be handled.

This feature was previously guarded by the FIPS build flag. But I removed that to remove one more source of non-FIPS/FIPS behaviour change...

### Testing:

Extended the current basic tests to being threaded - if threading is supported by the build.

### DIFF
--- a/crypto/fipsmodule/rand/new_rand_test.cc
+++ b/crypto/fipsmodule/rand/new_rand_test.cc
@@ -8,6 +8,8 @@
 #include "new_rand_internal.h"
 #include "../../ube/internal.h"
 
+#include <thread>
+
 // TODO
 // Remove when promoting to default
 #if !defined(BORINGSSL_PREFIX)
@@ -17,8 +19,7 @@
 
 #define MAX_REQUEST_SIZE (CTR_DRBG_MAX_GENERATE_LENGTH * 2 + 1)
 
-TEST(NewRand, Basic) {
-
+static void randBasicTests(bool *returnFlag) {
   uint8_t randomness[MAX_REQUEST_SIZE] = {0};
   uint8_t user_personalization_string[RAND_PRED_RESISTANCE_LEN] = {0};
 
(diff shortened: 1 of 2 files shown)


crypto/fipsmodule/rand/new_rand_test.cc
```diff
@@ -37,6 +38,26 @@ TEST(NewRand, Basic) {
     ASSERT_TRUE(RAND_bytes_with_additional_data(randomness, i, user_personalization_string));
     ASSERT_TRUE(RAND_bytes_with_user_prediction_resistance(randomness, i, user_personalization_string));    
   }
+
+  *returnFlag = true;
+}
+
+TEST(NewRand, Basic) {
```
andrewhop, line 45: This is a reasonable test for the happy case, can you add a test for the new cleanup code that ensures the locking works as expected. You might need to migrate this test from threads to futures which supports timeouts.

skmcgrail (dismissed):

andrewhop (dismissed):

skmcgrail (approved):

andrewhop (approved):

PR #1924, opened 2024-10-16
800-131Ar1: length of the key-derivation key shall be at least 112 bits.

skmcgrail:
### Description of changes: 
* KBKDF: Per SP 800-131Ar1, Section 8: "The length of the key-derivation key shall be at least 112 bits.” There must be a non-approved service indicator for KBKDF with KDK < 112 bits.

### DIFF
--- a/crypto/fipsmodule/service_indicator/service_indicator_test.cc
+++ b/crypto/fipsmodule/service_indicator/service_indicator_test.cc
@@ -4785,8 +4785,10 @@ INSTANTIATE_TEST_SUITE_P(All, KBKDFCtrHmacIndicatorTest,
 TEST_P(KBKDFCtrHmacIndicatorTest, KBKDF) {
   const KBKDFCtrHmacTestVector &vector = GetParam();
 
-  const uint8_t secret[20] = {'A', 'W', 'S', '-', 'L', 'C', ' ', 'K', 'B', 'K',
-                              'D', 'F', '-', 'C', 'T', 'R', ' ', 'K', 'E', 'Y'};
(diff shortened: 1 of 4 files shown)


crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -4789,7 +4789,9 @@ TEST_P(KBKDFCtrHmacIndicatorTest, KBKDF) {
                               'D', 'F', '-', 'C', 'T', 'R', ' ', 'K', 'E', 'Y'};
   const uint8_t info[16] = {'A', 'W', 'S', '-', 'L', 'C', ' ', 'K',
                             'B', 'K', 'D', 'F', '-', 'C', 'T', 'R'};
-  uint8_t output[16] = {0};
+  // 14 bytes (112 bits) is the minimum key-derivation key that would be
+  // approved if using an approved algorithm;
+  uint8_t output[14] = {0};
```
justsmth, line 4794: I'm a little confused. Both `sizeof(secret)` and `sizeof(output)` must be at least 14 for FIPS approval? `KBKDF_ctr_hmac_verify_service_indicator` only checks `secret_len`. Should it also (or instead) be checking `output_len`?

Would it be more appropriate for this PR to shorten the `secret` to length 14 for testing?

skmcgrail: Ugh yes, at first I had it as output check, then realized the constraint actually refers to the input secret length. Forgot to correct the test subsequently. Will fix here momentarily.

justsmth (approved):

nebeid (approved):

PR #1925, opened 2024-10-16
Also prune SSM documents from ec2-test-framework

samuel40791765:
### Issues:
Resolves `P149524833`

### Description of changes: 
There were some leaking SSM documents that weren't being pruned with the ec2 instances used. This adds the corresponding logic to remove them.

### Call-outs:
N/A

### Testing:
Tested on personal fork: https://github.com/samuel40791765/aws-lc/pull/35

### DIFF
--- a/tests/ci/lambda/src/bin/purge-stale-builds.rs
+++ b/tests/ci/lambda/src/bin/purge-stale-builds.rs
@@ -1,3 +1,4 @@
+use aws_sdk_ssm::operation::list_documents::ListDocumentsOutput;
 use aws_config::BehaviorVersion;
 use std::collections::HashMap;
 use std::time::{SystemTime, UNIX_EPOCH};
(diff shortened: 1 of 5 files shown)


tests/ci/lambda/src/bin/purge-stale-builds.rs
```diff
@@ -169,6 +182,66 @@ async fn handle(_event: LambdaEvent<Value>) -> Result<(), Error> {
...
+
+        let mut all_documents: Vec<String> = vec![];
+        let mut next_token: Option<String> = None;
+        loop {
+            // Make the request, setting the next_token if it exists
+            if let Some(ref ssm_client) = ssm_client_optional {
+                let result = ssm_client
+                    .list_documents()
+                    .set_filters(Some(ssm_client_filters.clone()))
+                    .set_next_token(next_token.clone())
```
skmcgrail, line 207: Note if you call [`into_paginator`](https://docs.rs/aws-sdk-ssm/latest/aws_sdk_ssm/operation/list_documents/builders/struct.ListDocumentsFluentBuilder.html#method.into_paginator) before calling send this handles the pagination token logic for you and allows you to use an iterator entirely. You can see similar code of how I do that with AWS CodeBuild here: https://github.com/aws/aws-lc/blob/d872db6de24223f0ceba782ccd2a157ea066e6c1/tests/ci/lambda/src/bin/purge-stale-builds.rs#L387-L411

Not a blocker per-se but would cleanup the code a bit.

samuel40791765: This actually helps a ton, I was getting stuck on this since the initial call to search through SSM wouldn't return the entire list and I had to do this workaround. It seemed a bit strange to me why the client had to handle all of this, will apply!

skmcgrail (approved):

justsmth (approved):

PR #1929, opened 2024-10-17
[fips-2024-09-27] Align X509 PARTIAL_CHAIN behavior with 1.1.1

samuel40791765:
Some consumers noticed that a behavior difference between AWS-LC and OpenSSL when the trust store contains certificates that are issued by other certificates that are also in the trust store. A common example of this would be the trust store containing both the intermediate and the root for the cert chain (`leaf -> intermediate -> root`). The default settings of AWS-LC and OpenSSL require `root` to be self signed for the chain to be verified.

Many TLS implementations set the `X509_V_FLAG_PARTIAL_CHAIN` flag however, which allows non self signed certificates to be trusted in the trust store. When `X509_V_FLAG_PARTIAL_CHAIN` is set, OpenSSL 1.1.1 will only verify leaf and intermediate, since intermediate is a trusted certificate. However, AWS-LC will continue building a certificate chain and include root within the chain of trust. This causes a behavioral difference with `X509_STORE_CTX_get0_chain`, where AWS-LC will return all 3 certificates (`leaf -> intermediate -> root`) and OpenSSL 1.1.1 will only return the first two (`leaf -> intermediate `).

Our upstream forked a bit before OpenSSL 1.0.2, so we don't have the new behavior. This described behavioral difference was introduced in openssl/openssl@d9b8b89 (along with many others), but the commit introduces too many backwards incompatible changes for us to take as a whole.
This subtle difference was due to [OpenSSL 1.1.1 continuously checking for trust while the chain's being established. The search for the next valid cert breaks early as soon as a valid chain has been built. Our current behavior builds the chain with all possible certs first and only breaks the loop if the final cert in the chain is self-signed. We can inherit this part of 1.1.1's new behavior to fix this issue.

### Call-outs:
I don't believe this really changes our X509 chain building or verification by much. We're only adding an additional check for trust while the chain is being established and the final chain still needs to go through the same building/verification process that exists in AWS-LC today.

### Testing:
Specific test for new behavior in `X509_V_FLAG_PARTIAL_CHAIN`

(cherry picked from commit 9fbfa706c9b7fece478605f57d4706b1d72c7b8e)

### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 

### Call-outs:

### Testing:

### DIFF
--- a/crypto/x509/x509_vfy.c
+++ b/crypto/x509/x509_vfy.c
@@ -362,6 +362,20 @@ int X509_verify_cert(X509_STORE_CTX *ctx) {
         ok = 0;
         goto end;
       }
+
+      // OpenSSL 1.1.1 continuously re-checks for trust and breaks the loop
+      // as soon as trust has been established. 1.0.2 builds the chain with all
+      // possible certs first and only checks for trust if the final cert in
+      // the chain is self-signed.
+      // This caused additional unanticipated certs to be in the established
+      // certificate chain, particularly when |X509_V_FLAG_PARTIAL_CHAIN| was
+      // set. We try checking continuously for trust here for better 1.1.1
+      // compatibility.
+      trust = check_trust(ctx);
+      if (trust == X509_TRUST_TRUSTED || trust == X509_TRUST_REJECTED) {
+        break;
+      }
+
       num++;
     }
 
--- a/crypto/x509/x509_test.cc
+++ b/crypto/x509/x509_test.cc
@@ -1701,6 +1701,52 @@ TEST(X509Test, TestVerify) {
   }
 }
 
+TEST(X509Test, PartialChain) {
+  bssl::UniquePtr<X509> root(CertFromPEM(kRootCAPEM));
+  bssl::UniquePtr<X509> intermediate(CertFromPEM(kIntermediatePEM));
+  bssl::UniquePtr<X509> leaf(CertFromPEM(kLeafPEM));
+  ASSERT_TRUE(root);
+  ASSERT_TRUE(intermediate);
+  ASSERT_TRUE(leaf);
+
+  // We're intentionally placing the intermediate cert in the trust store here.
+  // Many TLS implementations set |X509_V_FLAG_PARTIAL_CHAIN|, which allows
+  // non-self-signed certificates in the trust store to be trusted.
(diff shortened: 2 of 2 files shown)


skmcgrail (approved):

crypto/x509/x509_vfy.c
```diff
@@ -362,6 +362,20 @@ int X509_verify_cert(X509_STORE_CTX *ctx) {
...
+      // possible certs first and only checks for trust if the final cert in
+      // the chain is self-signed.
+      // This caused additional unanticipated certs to be in the established
+      // certificate chain, particularly when |X509_V_FLAG_PARTIAL_CHAIN| was
+      // set. We try checking continuously for trust here for better 1.1.1
+      // compatibility.
+      trust = check_trust(ctx);
+      if (trust == X509_TRUST_TRUSTED || trust == X509_TRUST_REJECTED) {
+        break;
+      }
```
justsmth, line 377: The tests also passes when this logic is conditioned on `X509_V_FLAG_PARTIAL_CHAIN`:
```
if (0 != (ctx->param->flags & X509_V_FLAG_PARTIAL_CHAIN)) {
  trust = check_trust(ctx);
  if (trust == X509_TRUST_TRUSTED || trust == X509_TRUST_REJECTED) {
    break;
  }
}
```
* Should there also be a test that confirms this behavioral change when `X509_V_FLAG_PARTIAL_CHAIN` has not been set? (Or do we only care about this behavior change when the flag is set?)

samuel40791765: * The real logic with `X509_V_FLAG_PARTIAL_CHAIN` is being done in `check_trust`, where we return for `X509_TRUST_TRUSTED` if it's set and we have a partial chain.
https://github.com/aws/aws-lc/blob/d4810f4f0530c9d9befbfe936c1b23b42faeb758/crypto/x509/x509_vfy.c#L770-L795

* I agree and we do have a test the checks that we have the intended behavioral change. The test above uses `for (bool partial_chain : {true, false}) {` to set the flag and then checks the found certificate chain length based on `partial_chain` with `    EXPECT_EQ(sk_X509_num(chain), partial_chain ? 2u : 3u);`.

justsmth (approved):

PR #1935, opened 2024-10-18
Let ECDSA_verify return -1 for ASN1 parsing fail

samuel40791765:
### Issues:
Addresses `CryptoAlg-2697`

### Description of changes: 
Ruby 3.1 [has a test which expects a fatal failure](https://github.com/ruby/ruby/blob/master/test/openssl/test_pkey_ec.rb#L143-L146) when parsing arbitrary ASN1 data during signature verification. I pinned this down to [Ruby's call](https://github.com/ruby/ruby/blob/7be9a333cabd97a17a2926b15f756f2ef9e57243/ext/openssl/ossl_pkey.c#L1394-L1408) to `EVP_PKEY_verify` which calls `ECDSA_verify` for EC operations. It turns out [OpenSSL](https://github.com/openssl/openssl/blob/master/crypto/ec/ecdsa_ossl.c#L419-L442) returns a -1 on any ASN1 parsing errors and only returns 0/1 to indicate signature verification failure/success. Ruby was dependent on the fatal error to determine whether to indicate an actual signature verification failure/success.
Tweaking the return code in `ECDSA_verify` in our code allows this test to pass.

### Call-outs:
Also tweaked the function documentation, this aligns with OpenSSL's documented behavior: https://github.com/openssl/openssl/blob/f4c467452694e1211395d17c2c027d99c35ee1e1/include/openssl/ec.h#L1455-L1467

### Testing:
N/A

### DIFF
--- a/crypto/fipsmodule/ecdsa/ecdsa.c
+++ b/crypto/fipsmodule/ecdsa/ecdsa.c
@@ -412,7 +412,7 @@ err:
 int ECDSA_verify(int type, const uint8_t *digest, size_t digest_len,
                  const uint8_t *sig, size_t sig_len, const EC_KEY *eckey) {
   ECDSA_SIG *s;
-  int ret = 0;
+  int ret = -1;
   uint8_t *der = NULL;
 
   // Decode the ECDSA signature.
--- a/include/openssl/ecdsa.h
+++ b/include/openssl/ecdsa.h
@@ -82,8 +82,8 @@ OPENSSL_EXPORT int ECDSA_sign(int type, const uint8_t *digest,
 
 // ECDSA_verify verifies that |sig_len| bytes from |sig| constitute a valid
 // signature by |key| of |digest|. (The |type| argument should be zero.) It
-// returns one on success or zero if the signature is invalid or an error
-// occurred.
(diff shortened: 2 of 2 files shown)


ctz (commented):
Please reconsider the reintroduction of badly-designed OpenSSL API practices. In this case, refer to [CVE-2013-2944](https://www.strongswan.org/blog/2013/04/30/strongswan-5.0.4-released-(cve-2013-2944).html) for an example. There is no semantic difference between an ASN1 encoding error, an invalid signature, or indeed any other error verifying a signature.

I also think this may be introducing the same vulnerability into `EVP_DigestVerify` and callers, which uses `EVP_PKEY_verify` in boolean context.

samuel40791765:
Hi, really appreciate the feedback! I agree that this was a poorly designed OpenSSL API and the CVE does give more context on how this could this be a dangerous change. This could also break any other callers checking for `!ECDSA_verify` that thought this was OK to do with AWS-LC. 

I hadn't given too much thought in tweaking this just yet, we've been drilling into these test failures and making tweaks for seemingly less harmful places. Given the potential push back on this, I think we can patch Ruby to get around this instead. I'll close this PR and leave it as a reference when we try to make a contribution upstream.

PR #1936, opened 2024-10-19
Add support for POINT_CONVERSION_HYBRID

samuel40791765:
### Issues:
Resolves `CryptoAlg-2697`

### Description of changes: 
This is the first time it has come up, but Ruby depends on the `POINT_CONVERSION_HYBRID` format in its tests and expects the underlying libcrypto to support it. 
* https://github.com/ruby/ruby/blob/ruby_3_1/test/openssl/test_pkey_ec.rb#L358-L366

There's not too much a difference between `POINT_CONVERSION_HYBRID` and `POINT_CONVERSION_UNCOMPRESSED` except that the first bit contains information about the quadratic equation y is at. We have the building blocks to support and can get this working with some tweaking around.

### Call-outs:
`POINT_CONVERSION_HYBRID` works very similar to `POINT_CONVERSION_UNCOMPRESSED` so we're reusing a lot of code. Most of this change is working with the first byte `POINT_CONVERSION_HYBRID` does differently and point the encoding to the right places for parsing.

### Testing:
* Testing parsing from and back with `POINT_CONVERSION_HYBRID`. What we're parsing to and from hybrid doesn't really matter here, they're already covered by other code paths. 

* Test vectors were generated from our test cases and ran against OpenSSL with this brief test program:

```
int main() {
  // Generate |ec_key|.
  EC_KEY *ec_key = EC_KEY_new();
  EC_GROUP *group = EC_GROUP_new_by_curve_name(NID_secp521r1);
  EC_KEY_set_group(ec_key, group);

  const uint8_t *inp = kECPublicKey_uncompressed;
  // Decoding an EC point.
  o2i_ECPublicKey(&ec_key, &inp, sizeof(kECPublicKey_uncompressed));
  // On successful exit of |o2i_ECPublicKey|, |*inp| is advanced by |len| bytes.
  if(kECPublicKey_uncompressed + sizeof(kECPublicKey_uncompressed) != inp) {
    return -1;
  }
  // Set |conv_form| of |ec_key|.
  EC_KEY_set_conv_form(ec_key, POINT_CONVERSION_HYBRID);
  uint8_t *p = nullptr;
  size_t len2 = i2o_ECPublicKey(ec_key, &p);
   for(uint32_t i=0; i < len2; i++) {
      if(i%12 == 0 && i != 0){
        printf("\n");
      }
      printf("0x%02x, ", p[i]);
    }
       printf("\n");
  return 0;
}
```

### DIFF
--- a/crypto/fipsmodule/ec/ec.c
+++ b/crypto/fipsmodule/ec/ec.c
@@ -1152,12 +1152,6 @@ int EC_METHOD_get_field_type(const EC_METHOD *meth) {
 
 void EC_GROUP_set_point_conversion_form(EC_GROUP *group,
                                         point_conversion_form_t form) {
-  /* NO-OP. However, abort if consumer attempts to set a representation that is
-   * not supported. */
-  if (form != POINT_CONVERSION_UNCOMPRESSED &&
-      form != POINT_CONVERSION_COMPRESSED) {
-    abort();
-  }
   // |conv_form| can only be set with OpenSSL compatible dynamically allocated
   // groups.
   if (group->mutable_ec_group) {
--- a/crypto/fipsmodule/ec/ec_test.cc
+++ b/crypto/fipsmodule/ec/ec_test.cc
@@ -211,6 +211,19 @@ static const uint8_t kP224PublicKey_compressed_0x02[] = {
   0xc5, 0x82, 0x78, 0x85
 };
 
+static const uint8_t kP224PublicKey_hybrid_0x02[] = {
+  /* uncompressed */
+  0x06,
+  /* x-coordinate */
+  0xd6, 0xf5, 0xf0, 0x6e, 0xf4, 0xc5, 0x56, 0x0a, 0xff, 0x8f, 0x49, 0x90,
+  0xef, 0xdb, 0xa5, 0x9a, 0xf8, 0xa8, 0xd3, 0x77, 0x0d, 0x80, 0x14, 0x6a,
+  0xc5, 0x82, 0x78, 0x85,
+  /* y-coordinate */
+  0xe0, 0x43, 0xae, 0x7b, 0xae, 0xa3, 0x77, 0x28, 0x60, 0x39, 0xc0, 0x7c,
+  0x04, 0x1b, 0x7a, 0x3b, 0x5d, 0x76, 0x96, 0xda, 0xdd, 0xa7, 0x05, 0x1a,
+  0xd6, 0x45, 0xa3, 0xea
+};
+
 static const uint8_t kP224PublicKey_uncompressed_0x03[] = {
   /* uncompressed */
   0x04,
@@ -230,6 +243,19 @@ static const uint8_t kP224PublicKey_compressed_0x03[] = {
   0xd6, 0xf5, 0xf0, 0x6e, 0xf4, 0xc5, 0x56, 0x0a, 0xff, 0x8f, 0x49, 0x90,
   0xef, 0xdb, 0xa5, 0x9a, 0xf8, 0xa8, 0xd3, 0x77, 0x0d, 0x80, 0x14, 0x6a,
   0xc5, 0x82, 0x78, 0x85
+};
+
+static const uint8_t kP224PublicKey_hybrid_0x03[] = {
+  /* uncompressed */
+  0x07,
+  /* x-coordinate */
+  0xd6, 0xf5, 0xf0, 0x6e, 0xf4, 0xc5, 0x56, 0x0a, 0xff, 0x8f, 0x49, 0x90,
+  0xef, 0xdb, 0xa5, 0x9a, 0xf8, 0xa8, 0xd3, 0x77, 0x0d, 0x80, 0x14, 0x6a,
+  0xc5, 0x82, 0x78, 0x85,
+  /* y-coordinate */
+  0x1f, 0xbc, 0x51, 0x84, 0x51, 0x5c, 0x88, 0xd7, 0x9f, 0xc6, 0x3f, 0x83,
+  0xfb, 0xe4, 0x85, 0xc3, 0xa2, 0x89, 0x69, 0x25, 0x22, 0x58, 0xfa, 0xe5,
+  0x29, 0xba, 0x5c, 0x17
 };
 
 static const uint8_t kP256PublicKey_uncompressed_0x02[] = {
@@ -253,6 +279,19 @@ static const uint8_t kP256PublicKey_compressed_0x02[] = {
   0x3f, 0x92, 0x72, 0x9b, 0x31, 0xc5, 0x5f, 0x7b
 };
 
+static const uint8_t kP256PublicKey_hybrid_0x02[] = {
+  /* uncompressed */
+  0x06,
+  /* x-coordinate */
+  0xe1, 0x5a, 0x44, 0x72, 0x91, 0xf0, 0x84, 0xfe, 0x88, 0x7a, 0x6c, 0x2c,
+  0x03, 0x22, 0x9a, 0xf3, 0x04, 0x8a, 0x5d, 0xfe, 0x84, 0x73, 0x70, 0xc9,
+  0x3f, 0x92, 0x72, 0x9b, 0x31, 0xc5, 0x5f, 0x7b,
+  /* y-coordinate */
+  0xc9, 0x53, 0x67, 0xc0, 0xd2, 0x90, 0x46, 0x86, 0x61, 0x8b, 0xf6, 0xf2,
+  0xd9, 0x0b, 0x7c, 0xcb, 0x31, 0xb0, 0xb4, 0x8c, 0x60, 0xc0, 0x28, 0x55,
+  0x6d, 0x1d, 0x3a, 0xbf, 0xdc, 0xd3, 0x1e, 0x42
+};
+
 static const uint8_t kP256PublicKey_uncompressed_0x03[] = {
   /* uncompressed */
   0x04,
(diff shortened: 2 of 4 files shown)


skmcgrail (dismissed):

crypto/fipsmodule/ec/ec.c
```diff
@@ -1152,12 +1152,6 @@ int EC_METHOD_get_field_type(const EC_METHOD *meth) {
 
 void EC_GROUP_set_point_conversion_form(EC_GROUP *group,
                                         point_conversion_form_t form) {
-  /* NO-OP. However, abort if consumer attempts to set a representation that is
-   * not supported. */
-  if (form != POINT_CONVERSION_UNCOMPRESSED &&
-      form != POINT_CONVERSION_COMPRESSED) {
```
dkostic, line 1158: why not just add `&& form != POINT_CONVERSION_HYBRID`?

crypto/fipsmodule/ec/ec_test.cc
```diff
@@ -211,6 +211,19 @@ static const uint8_t kP224PublicKey_compressed_0x02[] = {
   0xc5, 0x82, 0x78, 0x85
 };
 
+static const uint8_t kP224PublicKey_hybrid_0x02[] = {
```
dkostic, line 214: where did you get these values?

crypto/fipsmodule/ec/oct.c
```diff
@@ -110,6 +112,11 @@ size_t ec_point_to_bytes(const EC_GROUP *group, const EC_AFFINE *point,
     uint8_t y_buf[EC_MAX_BYTES];
     ec_felem_to_bytes(group, y_buf, &field_len, &point->Y);
     buf[0] = form + (y_buf[field_len - 1] & 1);
+    if (form == POINT_CONVERSION_HYBRID) {
+      // |POINT_CONVERSION_HYBRID| specifies y's solution of the quadratic
+      // equation, but also encodes the actual point along with it.
```
dkostic, line 117:
```suggestion
      // equation, but also encodes the y coordinate along with it.
```

crypto/fipsmodule/ec/oct.c
```diff
@@ -118,7 +125,10 @@ size_t ec_point_to_bytes(const EC_GROUP *group, const EC_AFFINE *point,
 int ec_point_from_uncompressed(const EC_GROUP *group, EC_AFFINE *out,
                                const uint8_t *in, size_t len) {
   const size_t field_len = BN_num_bytes(&group->field.N);
-  if (len != 1 + 2 * field_len || in[0] != POINT_CONVERSION_UNCOMPRESSED) {
+  // |POINT_CONVERSION_HYBRID| has the solution of y encoded in the first byte
+  // as well.
+  if (len != 1 + 2 * field_len || (in[0] != POINT_CONVERSION_UNCOMPRESSED &&
```
dkostic, line 130: why do we handle the hybrid case in function called `ec_point_from_uncompressed`?

crypto/fipsmodule/ec/oct.c
```diff
@@ -118,7 +125,10 @@ size_t ec_point_to_bytes(const EC_GROUP *group, const EC_AFFINE *point,
...
+                                   (in[0] & ~1u) != POINT_CONVERSION_HYBRID)) {
```
dkostic, line 131: please explain the magic `& ~1u`?

crypto/fipsmodule/ec/oct.c
```diff
@@ -142,7 +152,9 @@ static int ec_GFp_simple_oct2point(const EC_GROUP *group, EC_POINT *point,
   }
 
   point_conversion_form_t form = buf[0];
-  if (form == POINT_CONVERSION_UNCOMPRESSED) {
+  const int y_bit = form & 1;
+  form = form & ~1u;
```
dkostic, line 156: please explain the magic & ~1u?

crypto/fipsmodule/ec/ec_test.cc
samuel40791765: I ran the test vectors in the same file against [the brief test program](https://github.com/aws/aws-lc/pull/1936#issue-2598629333) in the PR description, when compiled against OpenSSL.

crypto/fipsmodule/ec/ec.c
samuel40791765: We're supporting all possible representations now, so I thought it made sense to just remove it since the comment above was no longer applicable. I could add it back if we really wanted to?

crypto/fipsmodule/ec/oct.c
samuel40791765: `~1u` takes an unsigned `1` and inverts all bits, which leaves behind `...1110`.
`form & ~1u` then performs an AND operation on `form` and `~1u`, which clears the far right bit of `form`. `form` is then adjusted to the nearest even value smaller than or equal to its original value..

crypto/fipsmodule/ec/oct.c
samuel40791765: When we ignore the first byte (which reveals the encoding format), uncompressed form is essentially identical to hybrid. It seemed like a bit redundant to write another `ec_point_from_hybrid` function with `uncompressed` parsing logic.
I could rename this function for a bit more clarity if needed? I was also considering adding a check in the hybrid decoding to see if `y` actually had the same solution that the first byte indicated, is that something we should do?

dkostic: I'd prefer having a separate `ec_point_from_hybrid` function, it's redundant but simpler.

> I was also considering adding a check in the hybrid decoding to see if y actually had the same solution that the first byte indicated, is that something we should do?

That's a great question. Since we are doing this for compatibility with OpenSSL I guess you should see if they are checking the point consistency?

crypto/fipsmodule/ec/oct.c
dkostic: So that code just sets the least significant bit of `form` to 0 which:
- does nothing if `form` is even,
- subtracts 1 if `form` is odd.

We do that because the hybrid form can be represented by either 6 or 7, but we internally encode it with `POINT_CONVERSION_HYBRID` which is equal to 6.

Anyway, consider adding a properly named helper function that does this for you and also shortly explains what I wrote above. For example:
```
int is_point_conversion_form_hybrid(form) { return POINT_CONVERSION_HYBRID == (form & ~1u); }

File: crypto/fipsmodule/ec/oct.c (hunk 4 above)
Comment (samuel40791765) [reply crypto/fipsmodule/ec/oct.c:130]: Sure, I did a quick look into what OpenSSL was doing. They have [a check](https://github.com/openssl/openssl/blob/8ff6edb9da6199b130bfb50bc27b2e58cc815932/crypto/ec/ecp_oct.c#L351-L352) for consistency that goes both ways. 
Our's will look slightly different since we save the date in `EC_FELEM`, but I'll figure something out here.

File: crypto/fipsmodule/ec/oct.c (hunk 6 above)
Comment (samuel40791765) [reply crypto/fipsmodule/ec/oct.c:167]: No problem will look into doing something for that to share the logic.

File: crypto/fipsmodule/ec/oct.c [hunk 7]
```diff
@@ -72,18 +72,29 @@
 
 #include "internal.h"
 
+static int is_point_conversion_form_hybrid(int form_bit) {
+  return POINT_CONVERSION_HYBRID == (form_bit & ~1u);
+}
+
+static int is_hybrid_bytes_consistent(const uint8_t *in, size_t field_len) {
+  // Check that the encoded solution in the first byte aligns with the computed
+  // point's.
```
dkostic, line 81:
```suggestion
  // point's, i.e., that they have the same parity.
```

crypto/fipsmodule/ec/oct.c
```diff
@@ -110,6 +121,14 @@ size_t ec_point_to_bytes(const EC_GROUP *group, const EC_AFFINE *point,
     uint8_t y_buf[EC_MAX_BYTES];
     ec_felem_to_bytes(group, y_buf, &field_len, &point->Y);
     buf[0] = form + (y_buf[field_len - 1] & 1);
+    if (form == POINT_CONVERSION_HYBRID) {
+      // |POINT_CONVERSION_HYBRID| specifies y's solution of the quadratic
+      // equation, but also encodes the y coordinate along with it.
+      OPENSSL_memcpy(buf + 1 + field_len, y_buf, field_len);
+      if (!is_hybrid_bytes_consistent(buf, field_len)) {
```
dkostic, line 128: there is no need for this check here because it can't fail. It can't fail because of how you set `buf[0]` in line 123: `buf[0] = form + (y_buf[field_len - 1] & 1);`

crypto/fipsmodule/ec/oct.c
```diff
@@ -133,6 +152,30 @@ int ec_point_from_uncompressed(const EC_GROUP *group, EC_AFFINE *out,
...
+  }
+
+  EC_FELEM x, y;
+  if (!ec_felem_from_bytes(group, &x, in + 1, field_len) ||
+      !ec_felem_from_bytes(group, &y, in + 1 + field_len, field_len) ||
+      !ec_point_set_affine_coordinates(group, out, &x, &y)) {
+    return 0;
+  }
+
+  if (!is_hybrid_bytes_consistent(in, field_len)) {
```
dkostic, line 172: you can move this to the first `if` in line 160.

skmcgrail (approved):

dkostic (approved):

PR #1940, opened 2024-10-22
Fix test memory leak

justsmth:
### Description of changes: 
* Call `EVP_PKEY_free` after each use.

### DIFF
--- a/crypto/evp_extra/evp_extra_test.cc
+++ b/crypto/evp_extra/evp_extra_test.cc
@@ -2999,6 +2999,7 @@ TEST(EVPExtraTest, KeygenCallbacks) {
   ASSERT_EQ(EVP_PKEY_keygen_init(ctx.get()), 1);
   ASSERT_TRUE(EVP_PKEY_keygen(ctx.get(), &pkey));
   ASSERT_TRUE(pkey);
+  EVP_PKEY_free(pkey);
 
   // Verify that |ctx->keygen_info| has not been updated since a callback hasn't
   // been set.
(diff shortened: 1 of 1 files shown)


torben-hansen (approved):

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -2999,6 +2999,7 @@ TEST(EVPExtraTest, KeygenCallbacks) {
   ASSERT_EQ(EVP_PKEY_keygen_init(ctx.get()), 1);
   ASSERT_TRUE(EVP_PKEY_keygen(ctx.get(), &pkey));
   ASSERT_TRUE(pkey);
+  EVP_PKEY_free(pkey);
```
andrewhop, line 3002: Could this test use a unique pointer up on line 2998 to avoid needing to manually handle this?
```
bssl::UniquePtr<EVP_PKEY> pkey(EVP_PKEY_new());
```

justsmth: * B/c `EVP_PKEY_keygen` takes a `EVP_PKEY**`, our tests typically declare a local "raw" pointer and pass its address in. We typically wrap it in a smart pointer after that.
* Our static analysis indicated there was a memory leak around `pkey` in this test function. Looking more closely at [`EVP_PKEY_assign`](https://github.com/aws/aws-lc/blob/fbeb5e8452898e4f8a527447a3fa18df6d59ab3a/crypto/fipsmodule/evp/evp.c#L425) and [`EVP_PKEY_set_type`](https://github.com/aws/aws-lc/blob/fbeb5e8452898e4f8a527447a3fa18df6d59ab3a/crypto/fipsmodule/evp/evp.c#L448), I think it's just a false positive. The subsequent call to `EVP_PKEY_keygen` on the same `pkey` takes care of freeing the necessary resources prior to its reinitialization.

PR #1946, opened 2024-10-24
HKDF, HKDF_expand, and PBKDF Truncated SHA2-512

skmcgrail:
### Description of changes: 
* ACVP testing for SHA2-512/224 and SHA2-512/256 for PBKDF and KDA: HKDF.
* Update service indicator for PBKDF, HKDF, and HKDF_expand (KBKDF feedback mode) to support SHA2-512/224 and SHA2-512/256 as approved algorithms.


justsmth (approved):

nebeid (approved):
- What is the source of the added test vectors in `service_inidicator_test.cc`, would be good to add the source to the comments if it's a standard or to the PR description.
- The sets (mentioned in L3374-L3378) of PBKDF test cases can be commented on because it's not clear which ones have short password/salt. The titles set 1, 2 or 3 can be put before each set in the `PBKDF2TestVector` struct.
- Most of the zip files got smaller especially KDA-HKDF.bz2, do we have the same coverage we'd like?

skmcgrail:
> * What is the source of the added test vectors in `service_inidicator_test.cc`, would be good to add the source to the comments if it's a standard or to the PR description.

There is a comment above each section. The HKDF ones from the RFC and the PBKDF ones come from a repository that contains python script for generating the vectors using OpenSSL via Python's cryptography pypi package.

>     * The sets (mentioned in L3374-L3378) of PBKDF test cases can be commented on because it's not clear which ones have short password/salt. The titles set 1, 2 or 3 can be put before each set in the `PBKDF2TestVector` struct.

There is a comment higher up describing the test cases, and that they are sourced from an external repository.

>     * Most of the zip files got smaller especially KDA-HKDF.bz2, do we have the same coverage we'd like?

Yes, either I forgot to trim the files last time, or they compress better this time. But they have what is needed to test the ACVP tool and module wrapper.

justsmth (approved):

justsmth (approved):

nebeid (approved):

PR #1951, opened 2024-10-26
Adding -verify and expanding -x509 options for our OpenSSL tool

smittals2:
### Issues:
`CryptoAlg-2679`

### Description of changes: 
- `-subject`, `-fingerprint`, `-inform`, `-enddate`, `-subject_hash`, `-subject_hash_old` options added to the x509 tool
- refactored the x509 tool to read from stdin
- A new `verify` tool with `-CAfile`
     - The `verify` tool takes a file as input or defaults to reading from stdin

### Callouts: 
The subject field differs in OpenSSL master and versions <= 3.2. A normalization function that removes whitespaces is used to compare output. 

### Testing:
Added option tests for new functionality and OpenSSL comparison tests for the new options. 


tool-openssl/verify.cc
```diff
@@ -0,0 +1,198 @@
...
+    goto end;
+  }
+
+  if (!CAfile.empty()) {
+    if (!X509_LOOKUP_load_file(lookup, CAfile.c_str(), X509_FILETYPE_PEM)) {
+      fprintf(stderr, "Error loading file %s\n", CAfile.c_str());
+      goto end;
+    }
+  } else {
+    X509_LOOKUP_load_file(lookup, NULL, X509_FILETYPE_DEFAULT);
```
justsmth, line 38: We should check the return value here.

tool-openssl/verify.cc
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  lookup = X509_STORE_add_lookup(store, X509_LOOKUP_hash_dir());
+  if (lookup == NULL) {
+    goto end;
+  }
+  X509_LOOKUP_add_dir(lookup, NULL, X509_FILETYPE_DEFAULT);
```
justsmth, line 45: Check return value.

tool-openssl/verify.cc
```diff
@@ -0,0 +1,198 @@
...
+  // Initialize certificate verification store
+  if (!store.get()) {
+    fprintf(stderr, "Error: Unable to setup certificate verification store.");
+    return false;
+  }
+  X509_STORE_set_verify_cb(store.get(), cb);
+
+  ERR_clear_error();
+
+  ret = 1;
```
justsmth, line 178: It would be better to assume failure (`ret = 0`), then only flip it to 1 on success.

tool-openssl/verify.cc
```diff
@@ -0,0 +1,198 @@
...
+  if (file) {
+    ScopedFILE cert_file(fopen(file, "rb"));
+    if (!cert_file) {
+      fprintf(stderr, "error %s: reading certificate failed\n", file);
+    }
+    x.reset(PEM_read_X509(cert_file.get(), nullptr, nullptr, nullptr));
+
+  } else {
+    bssl::UniquePtr<BIO> input(BIO_new_fp(stdin, BIO_CLOSE));
+    x.reset(PEM_read_bio_X509(input.get(), NULL, NULL, NULL));
```
justsmth, line 113: In one branch uses `nullptr`, the other uses `NULL`. I don't have a strong opinion on it, but I'd prefer `nullptr` to be used everywhere.

tool-openssl/verify.cc
smittals2: In the below code in the else block (lines 186-191):

The current logic, initializing `ret` to 1, is preferable. It allows us to:

1. Check all certificates without premature termination.
2. Efficiently track failures by setting `ret` to 0 on any single failure.
3. Avoid the complexity of tracking individual check results for multiple certs/files.

flipping the logic would require us to track every result and ensure they were all successful.

tool-openssl/x509.cc
```diff
@@ -250,6 +277,64 @@ bool X509Tool(const args_list_t &args) {
...
+      }
+
+      BIO_printf(output_bio.get(), "subject=");
+      X509_NAME_print_ex(output_bio.get(), subject_name, 0, XN_FLAG_ONELINE);
+      BIO_printf(output_bio.get(), "\n");
+    }
+
+    if (fingerprint) {
+      int j;
+      unsigned int n;
```
andrewhop, line 298: Please use more descriptive names, or move temp/index variables into a loop to make it clear.

tool-openssl/x509.cc
```diff
@@ -250,6 +277,64 @@ bool X509Tool(const args_list_t &args) {
...
+      unsigned char md[EVP_MAX_MD_SIZE];
+      const EVP_MD *digest = EVP_sha1();
+
+      if (!X509_digest(x509.get(), digest, md, &n)) {
+        fprintf(stderr, "Error: unable to obtain digest\n");
+        return false;
+      }
+      BIO_printf(output_bio.get(), "%s Fingerprint=",
+                 OBJ_nid2sn(EVP_MD_type(digest)));
+      for (j = 0; j < (int)n; j++) {
```
andrewhop, line 308:
```suggestion
      for (int j = 0; j < (int)n; j++) {
```

tool-openssl/x509_test.cc
```diff
@@ -13,6 +13,9 @@ X509* CreateAndSignX509Certificate() {
   bssl::UniquePtr<X509> x509(X509_new());
   if (!x509) return nullptr;
 
+  // Set version to X509v3
+  X509_set_version(x509.get(), 2);
```
andrewhop, line 17: Why is this change needed? Also please use the constant X509_VERSION_3

tool-openssl/x509_test.cc
```diff
@@ -113,6 +155,31 @@ TEST_F(X509Test, X509ToolModulusTest) {
   ASSERT_TRUE(result);
 }
 
+// Test -subject
```
andrewhop, line 158: What do we gain from these tests? What if `X509Tool` was a single line function that just returned true, would these tests notice?

tool-openssl/x509_test.cc
```diff
@@ -276,6 +368,19 @@ class X509ComparisonTest : public ::testing::Test {
   std::string openssl_output_str;
 };
 
+static std::string normalize_subject(std::string input) {
```
andrewhop, line 371: Can you describe what this is doing or provide some examples, I'm struggling to follow the logic below.

tool-openssl/x509_test.cc
```diff
@@ -325,6 +430,73 @@ TEST_F(X509ComparisonTest, X509ToolCompareModulusOpenSSL) {
   ASSERT_EQ(tool_output_str, openssl_output_str);
 }
 
+// Test against OpenSSL output "openssl x509 -in file -subject"
+TEST_F(X509ComparisonTest, X509ToolCompareSubjectOpenSSL) {
```
andrewhop, line 434: In a follow up PR can you please refactor all of these tests into a struct that takes the command to run and a cleanup function and then just loops over all of them.

tool-openssl/verify.cc
justsmth: I'm not sure why `ret` is needed at all.  The logic could just `return 0` in the places it currently has `ret = 0`. Is there a reason to continue checking after one of them fails?

justsmth: NP: A slightly less verbose (and clearer?) way to write this would be the following:

```
  ret = 1;

  if (args.size() == i) {
    ret &= check(store.get(), NULL);
  } else {
    // Certs provided as files
    for (; i < args.size(); i++) {
      ret &= check(store.get(), args[i].c_str());
    }
  }

  return ret == 1;
```

smittals2: agreed, made the change thank you!

tool-openssl/verify.cc
```diff
@@ -0,0 +1,202 @@
...
+    if (!X509_LOOKUP_load_file(lookup, NULL, X509_FILETYPE_DEFAULT)) {
+      goto end;
+    }
+  }
+
+  lookup = X509_STORE_add_lookup(store, X509_LOOKUP_hash_dir());
+  if (lookup == NULL) {
+    goto end;
+  }
+  if (!X509_LOOKUP_add_dir(lookup, NULL, X509_FILETYPE_DEFAULT)) {
```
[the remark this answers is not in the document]
smittals2: When the type passed into `X509_LOOKUP_add_dir` is `X509_FILETYPE_DEFAULT` the path is ignored and a default system path is used.

tool-openssl/verify.cc
```diff
@@ -0,0 +1,202 @@
...
+static const argument_t kArguments[] = {
+        { "-help", kBooleanArgument, "Display option summary" },
+        { "-CAfile", kOptionalArgument, "A file of trusted certificates. The "
+                "file should contain one or more certificates in PEM format." },
+        { "", kOptionalArgument, "" }
+};
+
+// setup_verification_store sets up an X509 certificate store for verification.
+// It configures the store with file and directory lookups. It loads the
+// specified CA file if provided and otherwise uses default locations.
```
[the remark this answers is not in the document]
smittals2: Yes I guess we don't need to add the default file lookup and can directly add the default dir.

tool-openssl/verify.cc
```diff
@@ -0,0 +1,202 @@
...
+               X509_STORE_CTX_get_error_depth(ctx),
+               X509_verify_cert_error_string(cert_error));
+
+    /*
+     * Pretend that some errors are ok, so they don't stop further
+     * processing of the certificate chain.  Setting ok = 1 does this.
+     * After X509_verify_cert() is done, we verify that there were
+     * no actual errors, even if the returned value was positive.
+     */
+    switch (cert_error) {
```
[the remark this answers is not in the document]
smittals2: There is a longer list of errors we support compared to what's included here ([link](https://github.com/aws/aws-lc/blob/f0c9b3aeacacf11d55ea928d1ed98c98b4b9d1e9/include/openssl/x509.h#L5582)). I chose to include the subset of those we support from what OpenSSL has configured in their callback to match their behavior ([link](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/apps/verify.c#L272))

tool-openssl/verify.cc
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  lookup = X509_STORE_add_lookup(store, X509_LOOKUP_file());
+  if (!lookup) {
+    goto end;
+  }
+
+  if (!CAfile.empty()) {
+    if (!X509_LOOKUP_load_file(lookup, CAfile.c_str(), X509_FILETYPE_PEM)) {
+      fprintf(stderr, "Error loading file %s\n", CAfile.c_str());
```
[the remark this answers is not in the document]
smittals2: It technically does, we just don't need it yet for the customer we want to support.

tool-openssl/verify.cc
```diff
@@ -0,0 +1,202 @@
...
+
+static int check(X509_STORE *ctx, const char *file) {
+  bssl::UniquePtr<X509> x;
+  int i = 0, ret = 0;
+  X509_STORE_CTX *store_ctx;
+
+  if (file) {
+    ScopedFILE cert_file(fopen(file, "rb"));
+    if (!cert_file) {
+      fprintf(stderr, "error %s: reading certificate failed\n", file);
```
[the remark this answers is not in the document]
smittals2: Yep missed that

tool-openssl/verify.cc
```diff
@@ -0,0 +1,202 @@
...
+bool VerifyTool(const args_list_t &args) {
+  std::string cafile;
+  bssl::UniquePtr<X509_STORE> store;
+  int ret;
+  size_t i = 0;
+
+  if (args.size() == 1 && args[0] == "-help") {
+    fprintf(stderr, "Usage: verify [options] cert.pem...\n"
+                    "Certificates must be in PEM format and specified in a file.\n"
+                    "We do not support reading from stdin yet. \n\n "
```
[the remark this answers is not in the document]
smittals2: yes this was outdated, forgot to change after adding support

tool-openssl/verify.cc
```diff
@@ -0,0 +1,202 @@
...
+                    "Valid options are:\n");
+    PrintUsage(kArguments);
+    return false;
+  } else if (args.size() > 1 && args[0] == "-CAfile") {
+    cafile = args[1];
+    i += 2;
```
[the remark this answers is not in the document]
smittals2: i is tracking the number of arguments. i.e. if -CAfile is provided, the option and the file name are two arguments. Later in the code, we check if i == args.size(), read from stdin (no cert file(s) were specified)

tool-openssl/verify.cc
```diff
@@ -0,0 +1,202 @@
...
+
+  if (args.size() == 1 && args[0] == "-help") {
+    fprintf(stderr, "Usage: verify [options] cert.pem...\n"
+                    "Certificates must be in PEM format and specified in a file.\n"
+                    "We do not support reading from stdin yet. \n\n "
+                    "Valid options are:\n");
+    PrintUsage(kArguments);
+    return false;
+  } else if (args.size() > 1 && args[0] == "-CAfile") {
+    cafile = args[1];
```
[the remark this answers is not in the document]
smittals2: That framework doesn't support arguments to the tool that aren't tied to a flag (i.e without a -option). parsed_args is the map created and iterated through to validate each argument. I can look into modifying it to support this use case but verify is the only tool so far that needs this functionality so using a counter seemed like the efficient approach.

tool-openssl/x509_test.cc
smittals2: These tests wouldn't, but the openssl comparison tests would. These tests are simply verifying that we support the advertised options and are able to parse them correctly. I guess the openssl comparison tests also verify this. I think we could possibly make this more efficient by having an array of options and checking them in one function. I can tackle this in a follow up PR along with the refactor for comparison tests suggested below.

tool-openssl/x509_test.cc
smittals2: Noted, changed to using the constant. X509_VERSION_3 adds support for extensions. We use these extensions to configure a self signed certificate.

tool-openssl/x509_test.cc
smittals2: We find "subject=" in the provided input. If that string is found, we look for the next newline in input. We extract the substring from "subject=" to the newline. Then we remove all whitespace characters from this string and replace it back into input. 

I've added some documentation for the function

tool-openssl/x509.cc
smittals2: Noted

tool-openssl/verify_test.cc
```diff
@@ -0,0 +1,157 @@
...
+
+
+// -------------------- Verify OpenSSL Comparison Tests --------------------------
+
+// Comparison tests cannot run without set up of environment variables:
+// AWSLC_TOOL_PATH and OPENSSL_TOOL_PATH.
+
+class VerifyComparisonTest : public ::testing::Test {
+protected:
+    void SetUp() override {
```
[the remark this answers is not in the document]
smittals2: Now the setup is smaller only setting up cert and ca. Are you thinking helper function that does this and then use that in both setups?

tool-openssl/verify.cc
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  return store.release();
+}
+
+static int cb(int ok, X509_STORE_CTX *ctx) {
+  int cert_error = X509_STORE_CTX_get_error(ctx);
+  X509 *current_cert = X509_STORE_CTX_get_current_cert(ctx);
+
+  if (!ok) {
```
justsmth, line 48: NP: These variables could be moved inside the `if (!ok) { ... }`.

justsmth (dismissed):

tool-openssl/verify.cc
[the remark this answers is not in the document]
andrewhop: I would like to see ParseKeyValueArguments (or a new thing) support these use cases. The other OpenSSL tools do funky manual argument parsing as well and I would like to clean them all up.

tool-openssl/verify.cc
```diff
@@ -0,0 +1,177 @@
...
+     * no actual errors, even if the returned value was positive.
+     */
+    switch (cert_error) {
+      case X509_V_ERR_NO_EXPLICIT_POLICY:
+        /* fall thru */
+      case X509_V_ERR_CERT_HAS_EXPIRED:
+        /* Continue even if the leaf is a self-signed cert */
+      case X509_V_ERR_DEPTH_ZERO_SELF_SIGNED_CERT:
+        /* Continue after extension errors too */
+      case X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY:
```
andrewhop, line 74: Is this intentionally different from OpenSSL? There tool does not special case `X509_V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY`

tool-openssl/verify_test.cc
```diff
@@ -0,0 +1,139 @@
...
+// Test against OpenSSL with -CAfile & self-signed cert fed through stdin
+// "cat cert.pem | openssl verify -CAfile cert.pem"
+TEST_F(VerifyComparisonTest, VerifyToolOpenSSLCAFileSelfSignedStdinComparison) {
+  std::string tool_command = "cat " + std::string(ca_path) + " | " + std::string(tool_executable_path) + " verify -CAfile " + ca_path + " &> " + out_path_tool;
+  std::string openssl_command = "cat " + std::string(ca_path) + " | " + std::string(openssl_executable_path) + " verify -CAfile " + ca_path + " &> " + out_path_openssl;
+
+  RunCommandsAndCompareOutput(tool_command, openssl_command, out_path_tool, out_path_openssl, tool_output_str, openssl_output_str);
+
+  ASSERT_EQ(tool_output_str, openssl_output_str);
+}
```
andrewhop, line 139: Can you add a test that verify verifies a public cert (like amazon.com), this will verify that we are correctly loading the default system trust store certs. Traps to avoid:
1. Don't create a time bomb
2. Not everywhere that runs our CI has network access

tool-openssl/x509_test.cc
```diff
@@ -367,3 +541,71 @@ TEST_F(X509ComparisonTest, X509ToolCompareDatesNooutOpenSSL) {
...
+
+  tool_command = std::string(tool_executable_path) + " x509 -in " + der_cert_path + " -inform DER -enddate -out " + out_path_tool;
+  openssl_command = std::string(openssl_executable_path) + " x509 -in " + der_cert_path + " -inform DER -enddate -out " + out_path_openssl;
+
+  RunCommandsAndCompareOutput(tool_command, openssl_command, out_path_tool, out_path_openssl, tool_output_str, openssl_output_str);
+
+  ASSERT_EQ(tool_output_str, openssl_output_str);
+}
+
+// Test against OpenSSL output "openssl x509 -in file -inform DER -enddate"
```
andrewhop, line 579: Can you add a test for using the wrong inform.

tool-openssl/verify.cc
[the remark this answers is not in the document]
smittals2: Noted, I'll work on adding this feature along with cleaning up testing code in the follow up

tool-openssl/verify.cc
smittals2: Ah good catch, I had added it in for figuring out why our output was different from OpenSSL in some cases. This was the error message that was problematic but this change did not do anything - I just forgot to remove it.

tool-openssl/x509_test.cc
smittals2: Do you mean you want to verify that the output for the error case of using a wrong inform is the same as openssl? 

I already added an erroneous usage test case for this option here: https://github.com/aws/aws-lc/pull/1951/files#diff-8b8002fb4d9429e7a3d9c3b9f2fddee7d50b81db27af7321ab566d0a1359b430R283

justsmth (approved):

andrewhop (approved):

PR #1952, opened 2024-10-26
Support Finished-based APIs for TLS 1.3

samuel40791765:
### Issues:
Addresses `CryptoAlg-2699`

### Description of changes: 
Another [Ruby 3.1 test](https://github.com/ruby/ruby/blame/ruby_3_1/test/openssl/test_ssl.rb#L484-L510) is depending on certain SSL APIs to retain information when using TLS1.3. This time it's the `SSL_get(_peer)_finished` APIs that are involved. OpenSSL treats the fields identically in 1.2 and 1.3 and uses the same API to update both state machines.
We already have our own individual `ssl_get/send_finished` for TLS 1.2 and `tls13_add/process_finished` for TLS 1.3. I've added the field updates within the corresponding TLS 1.3 functions so that these return the correct value within a 1.3 setting.

### Call-outs:
The field buffer for `previous_server_finished` and `previous_client_finished` has to grow past 12 since it's also used for 1.3 now. This breaks the original SSL Transfer assumption and we'll have to bump the version while adding corresponding logic to account for the updated size. I've regenerated the SSL Transfer bytes for the round trip tests and also added a test for v3 as well.

### Testing:
WIP


ssl/ssl_transfer_asn1.cc
```diff
@@ -485,11 +494,17 @@ static int SSL3_STATE_from_bytes(SSL *ssl, CBS *cbs, const SSL_CTX *ctx) {
       !CBS_get_asn1_uint64(&s3, &early_data_reason) ||
       early_data_reason > ssl_early_data_reason_max_value ||
       !CBS_get_asn1(&s3, &previous_client_finished, CBS_ASN1_OCTETSTRING) ||
-      CBS_len(&previous_client_finished) != PREV_FINISHED_MAX_SIZE ||
+      (is_pre_v3 &&
+       CBS_len(&previous_client_finished) != PRE_V3_PREV_FINISHED_MAX_SIZE) ||
+      (!is_pre_v3 &&
+       CBS_len(&previous_client_finished) != PREV_FINISHED_MAX_SIZE) ||
```
smittals2, line 500: Not sure if i'm understanding this correctly, but should `previous_client_finished` in the v3 case always be equal to `PREV_FINISHED_MAX_SIZE`? Should the check not be > MAX_SIZE? Depending on hash function used, the output length could be shorter than PREV_FINISHED_MAX_SIZE (which looks like is 64 to account for SHA-512)?

ssl/ssl_transfer_asn1.cc
```diff
@@ -485,11 +494,17 @@ static int SSL3_STATE_from_bytes(SSL *ssl, CBS *cbs, const SSL_CTX *ctx) {
...
       !CBS_get_asn1_uint64(&s3, &previous_client_finished_len) ||
       previous_client_finished_len > PREV_FINISHED_MAX_SIZE ||
       !CBS_get_asn1(&s3, &previous_server_finished, CBS_ASN1_OCTETSTRING) ||
-      CBS_len(&previous_server_finished) != PREV_FINISHED_MAX_SIZE ||
+      (is_pre_v3 &&
+       CBS_len(&previous_server_finished) != PRE_V3_PREV_FINISHED_MAX_SIZE) ||
+      (!is_pre_v3 &&
+       CBS_len(&previous_server_finished) != PREV_FINISHED_MAX_SIZE) ||
```
smittals2, line 507: same as above

ssl/ssl_transfer_asn1.cc
```diff
@@ -521,7 +536,7 @@ static int SSL3_STATE_from_bytes(SSL *ssl, CBS *cbs, const SSL_CTX *ctx) {
     return 0;
   }
 
-  bool is_v2 = (serde_version == SSL3_STATE_SERDE_VERSION_TWO);
+  bool is_v2 = (serde_version >= SSL3_STATE_SERDE_VERSION_TWO);
```
smittals2, line 539: Confirming is_v2 should be true if its v2 or v3?

smittals2 (approved):

smittals2 (approved):

skmcgrail (approved):

PR #1956, opened 2024-10-30
Account for cipher auth with multiple cert slots

samuel40791765:
### Issues:
Addresses `CryptoAlg-2699`

### Description of changes: 
Ruby has a dependency on the multiple certificate slot mechanisms that OpenSSL allows for. We've already done the work to support this, but [another Ruby 3.1 test](https://github.com/ruby/ruby/blame/ruby_3_1/test/openssl/test_ssl.rb#L137-L178) has exposed a gap in our support. We were only depending on the negotiated signature algorithms to retrieve the right certificate for the server to send back, but the cipher authentication scheme was [also checked in OpenSSL](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/ssl/t1_lib.c#L2576-L2593) as well. Ruby's tests happen to only depend on configuring the authentication scheme which brought this to light.
We happen to do only do this when checking private keys for TLS 1.0/1.1 , this also fixes it to check the cipher against the initial public key. This change also introduces all of the mentioned cipher authenticated check behavior  for TLS 1.2.

### Call-outs:
N/A

### Testing:
New unit test that allows all possible sigalgs, with the cipher authentication suite being the only restriction.

### DIFF
--- a/ssl/ssl_privkey.cc
+++ b/ssl/ssl_privkey.cc
@@ -386,7 +386,16 @@ static bool ssl_public_key_rsa_pss_check(EVP_PKEY *pubkey, uint16_t sigalg) {
 bool ssl_public_key_supports_signature_algorithm(SSL_HANDSHAKE *hs,
                                                  uint16_t sigalg) {
   SSL *const ssl = hs->ssl;
-  if (!pkey_supports_algorithm(ssl, hs->local_pubkey.get(), sigalg)) {
+  // We may have a public key that supports the signature algorithm, but the
+  // negotiated cipher needs to also allow it. This behavior is only done prior
+  // to TLS 1.2 servers in OpenSSL since TLS 1.3 does not have cipher-based
+  // authentication configuration.
+  const uint32_t auth_allowed =
+      !ssl->server ||
+      (hs->new_cipher->algorithm_auth &
+       (ssl_cipher_auth_mask_for_key(hs->local_pubkey.get()) | SSL_aGENERIC));
+  if (!auth_allowed ||
+      !pkey_supports_algorithm(ssl, hs->local_pubkey.get(), sigalg)) {
     return false;
   }
 
@@ -397,8 +406,7 @@ bool ssl_public_key_supports_signature_algorithm(SSL_HANDSHAKE *hs,
   return true;
 }
 
-UniquePtr<EVP_PKEY> ssl_cert_parse_leaf_pubkey(
-    STACK_OF(CRYPTO_BUFFER) *chain) {
+UniquePtr<EVP_PKEY> ssl_cert_parse_leaf_pubkey(STACK_OF(CRYPTO_BUFFER) *chain) {
   const CRYPTO_BUFFER *buf = sk_CRYPTO_BUFFER_value(chain, 0);
   if (buf == nullptr) {
     return nullptr;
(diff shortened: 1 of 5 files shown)


ssl/ssl_privkey.cc
```diff
@@ -386,7 +386,16 @@ static bool ssl_public_key_rsa_pss_check(EVP_PKEY *pubkey, uint16_t sigalg) {
 bool ssl_public_key_supports_signature_algorithm(SSL_HANDSHAKE *hs,
                                                  uint16_t sigalg) {
   SSL *const ssl = hs->ssl;
-  if (!pkey_supports_algorithm(ssl, hs->local_pubkey.get(), sigalg)) {
+  // We may have a public key that supports the signature algorithm, but the
+  // negotiated cipher needs to also allow it. This behavior is only done prior
+  // to TLS 1.2 servers in OpenSSL since TLS 1.3 does not have cipher-based
+  // authentication configuration.
```
skmcgrail, line 392: nit: Might be worth making a note that the TLS 1.3 variant is checked via the `SSL_aGENERIC` bit flag being set. Since all those ciphers specify this value. Assuming my understanding of this code flow is correct.

ssl/ssl_privkey.cc
```diff
@@ -457,20 +476,29 @@ bool ssl_cert_private_keys_supports_signature_algorithm(SSL_HANDSHAKE *hs,
...
+        ssl_cert_parse_leaf_pubkey(cert->cert_private_keys[i].chain.get());
+    if (private_key != nullptr && public_key != nullptr) {
+      // We may have a private key that supports the signature algorithm,
+      // but we need to verify that the negotiated cipher allows it. This
+      // behavior is only done prior to TLS 1.2 servers in OpenSSL since TLS
+      // 1.3 does not have cipher-based authentication configuration.
+      const uint32_t auth_allowed =
+          !ssl->server ||
+          hs->new_cipher->algorithm_auth &
+              (ssl_cipher_auth_mask_for_key(private_key) | SSL_aGENERIC);
```
skmcgrail, line 488: Might be worth making this a macro, or inline function so that the definitions are kept in-sync with the other function (`ssl_public_key_supports_signature_algorithm`).

ssl/ssl_privkey.cc
samuel40791765: Yes that was the intention behind using `SSL_aGENERIC`. No problem, definately worth making a note of.

skmcgrail (dismissed):

ssl/ssl_privkey.cc
```diff
@@ -383,10 +383,25 @@ static bool ssl_public_key_rsa_pss_check(EVP_PKEY *pubkey, uint16_t sigalg) {
   return true;
 }
 
+static bool tls12_pkey_supports_cipher_auth(SSL_HANDSHAKE *hs,
+                                            const EVP_PKEY *key) {
+  SSL *const ssl = hs->ssl;
+  // We may have a private key that supports the signature algorithm, but we
+  // need to verify that the negotiated cipher allows it. This behavior is only
+  // done prior to TLS 1.2 servers in OpenSSL since TLS 1.3 does not have
```
WillChilds-Klein, line 391:
```suggestion
  // done in OpenSSL servers with TLS version 1.2 and below since TLS 1.3 does not have
```

ssl/ssl_privkey.cc
```diff
@@ -383,10 +383,25 @@ static bool ssl_public_key_rsa_pss_check(EVP_PKEY *pubkey, uint16_t sigalg) {
...
+  // cipher-based authentication configuration. Since cipher-based
+  // authentication is already built into TLS 1.3, we use the |SSL_aGENERIC|
```
WillChilds-Klein, line 393: this is a little confusing... suggest something like "Since authentication is configured outside the ciphersuite in TLSv1.3, ..."

ssl/ssl_test.cc
```diff
@@ -6598,9 +6595,52 @@ TEST_P(MultipleCertificateSlotTest, AutomaticSelection) {
       {certificate_key_param().corresponding_sigalg}, SSL_PKEY_ED25519, true);
 }
 
+TEST_P(MultipleCertificateSlotTest, AutomaticSelectionCipherAuth) {
+  if ((version < TLS1_2_VERSION && slot_index == SSL_PKEY_ED25519) ||
+      version >= TLS1_3_VERSION) {
+    // ED25519 is not supported in versions prior to TLS1.2.
+    // TLS 1.3 not have cipher-based authentication configuration.
+    return;
```
WillChilds-Klein, line 6603: minor nit here, L6672, and L6645: `GTEST_SKIP()` instead of returning.

ssl/ssl_privkey.cc
samuel40791765: No problem at all, will change.

ssl/ssl_test.cc
samuel40791765: Ah yes forgot about that, I went ahead and did it for all tests we're already slightly optimizing.

WillChilds-Klein (approved):

skmcgrail (approved):

PR #1959, opened 2024-11-01
WIP: Adding the OpenSSL s_client tool

smittals2:
### Issues:
`CryptoAlg-2679`

### Description of changes: 
Adding the s_client tool to our openssl CLI executable. It mostly uses the same code from our existing bssl CLI s_client tool. There are various changes to functions to help distinguish which CLI variant is being called. The output of bssl's CLI for s_client is also very different from OpenSSL 1.1.1. There are a lot of utility printing functions added and changes in logic to suppress output when using the openssl CLI. 

### Call-outs:
The verification output still doesn't match OpenSSL 1.1.1 fully. There is also some missing data like peer CAs, peer tmp key, public key bits, and sig_algs. 

### Testing:

### DIFF
--- a/tool-openssl/s_client.cc
+++ b/tool-openssl/s_client.cc
@@ -1,0 +1,46 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include <openssl/base.h>
+#include <openssl/pem.h>
+#include "internal.h"
+#include "../tool/internal.h"
+
+static const argument_t kArguments[] = {
+        { "-help", kBooleanArgument, "Display option summary" },
+        { "-connect", kRequiredArgument,
+          "The hostname and port of the server to connect to, e.g. foo.com:443" },
+        // In OpenSSL, CAFile by itself may not neccessitate verification, verify if this
+        // is the case or not (-verify option may be needed). In AWS-LC it is and verify option not supported for bssl.
+        { "-CAfile", kOptionalArgument,
+          "A filename containing one or more PEM root certificates. Implies that "
+          "verification is required." },
+        { "-showcerts", kBooleanArgument,
+          "Displays the server certificate list as sent by the server: it only "
+          "consists of certificates the server has sent (in the order the server "
+          "has sent them). It is not a verified chain. " },
+        { "-verify", kOptionalArgument,
+          "The verify depth to use. This specifies the maximum length of the server "
+          "certificate chain and turns on server certificate verification. "
+          "Currently the verify operation continues after errors so all the problems "
+          "with a certificate chain can be seen. As a side effect the connection will "
+          "never fail due to a server certificate verify failure." },
+        { "", kOptionalArgument, "" },
+};
+
+bool SClientTool(const args_list_t &args) {
+  std::map<std::string, std::string> args_map;
+
+  if (!ParseKeyValueArguments(&args_map, args, kArguments)) {
+    PrintUsage(kArguments);
+    return false;
+  }
+
+  if(args_map.count("help")) {
+    fprintf(stderr, "Usage: s_client [options] [host:port]\n");
+    PrintUsage(kArguments);
+    return false;
+  }
+
+  return DoClient(args_map, 1);
+}
--- a/tool-openssl/s_client.cc
+++ b/tool-openssl/s_client.cc
@@ -9,22 +9,22 @@
 static const argument_t kArguments[] = {
         { "-help", kBooleanArgument, "Display option summary" },
         { "-connect", kRequiredArgument,
-          "The hostname and port of the server to connect to, e.g. foo.com:443" },
-        // In OpenSSL, CAFile by itself may not neccessitate verification, verify if this
-        // is the case or not (-verify option may be needed). In AWS-LC it is and verify option not supported for bssl.
+                "The hostname and port of the server to connect to, e.g. foo.com:443" },
         { "-CAfile", kOptionalArgument,
-          "A filename containing one or more PEM root certificates. Implies that "
-          "verification is required." },
+                "A file containing trusted certificates to use during server authentication "
+                "and to use when attempting to build the client certificate chain. " },
+        { "-CApath", kOptionalArgument,
+                "The directory to use for server certificate verification. " },
         { "-showcerts", kBooleanArgument,
-          "Displays the server certificate list as sent by the server: it only "
-          "consists of certificates the server has sent (in the order the server "
-          "has sent them). It is not a verified chain. " },
+                "Displays the server certificate list as sent by the server: it only "
+                "consists of certificates the server has sent (in the order the server "
+                "has sent them). It is not a verified chain. " },
         { "-verify", kOptionalArgument,
-          "The verify depth to use. This specifies the maximum length of the server "
-          "certificate chain and turns on server certificate verification. "
-          "Currently the verify operation continues after errors so all the problems "
-          "with a certificate chain can be seen. As a side effect the connection will "
-          "never fail due to a server certificate verify failure." },
+                "The verify depth to use. This specifies the maximum length of the server "
+                "certificate chain and turns on server certificate verification. "
+                "Currently the verify operation continues after errors so all the problems "
+                "with a certificate chain can be seen. As a side effect the connection will "
+                "never fail due to a server certificate verify failure." },
         { "", kOptionalArgument, "" },
 };
 
--- a/tool/client.cc
+++ b/tool/client.cc
@@ -245,17 +245,90 @@ static bool WaitForSession(SSL *ssl, int sock) {
   return true;
 }
 
+static void PrintOpenSSLConnectionInfo(SSL *ssl, int show_certs) {
+  STACK_OF(X509) *sk;
+  X509 *peer;
+  int i;
+//  EVP_PKEY *public_key;
+//
+  sk = SSL_get_peer_cert_chain(ssl);
+  if (sk != NULL) {
+    fprintf(stdout, "---\nCertificate chain\n");
+    for (i = 0; i < (int) sk_X509_num(sk); i++) {
+      fprintf(stdout, "%2d s:", i);
+      X509_NAME_print_ex_fp(stdout, X509_get_subject_name(sk_X509_value(sk, i)),
+                            0, XN_FLAG_ONELINE);
+      fprintf(stdout, "\n   i:");
+      X509_NAME_print_ex_fp(stdout, X509_get_issuer_name(sk_X509_value(sk, i)),
+                            0, XN_FLAG_ONELINE);
+      fprintf(stdout, "\n");
+//      public_key = X509_get_pubkey(sk_X509_value(sk, i));
+//      if (public_key != NULL) {
+//        fprintf(stdout, "   a:PKEY: %s, %d (bit); sigalg: %s\n",
+//                   OBJ_nid2sn(EVP_PKEY_get_base_id(public_key)),
+//                   EVP_PKEY_get_bits(public_key),
+//                   OBJ_nid2sn(X509_get_signature_nid(sk_X509_value(sk, i))));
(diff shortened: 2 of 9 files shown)


tool/transport_common.cc
```diff
@@ -137,7 +137,7 @@ static void PrintSocketError(const char *function) {
 // Connect sets |*out_sock| to be a socket connected to the destination given
 // in |hostname_and_port|, which should be of the form "www.example.com:123".
 // It returns true on success and false otherwise.
-bool Connect(int *out_sock, const std::string &hostname_and_port) {
+bool Connect(int *out_sock, const std::string &hostname_and_port, int quiet) {
```
samuel40791765, line 140: Can make this `bool` now that we're free from the shackles of C 
```suggestion
bool Connect(int *out_sock, const std::string &hostname_and_port, bool quiet) {
```

tool/internal.h
```diff
@@ -137,6 +137,11 @@ bool WriteToFile(const std::string &path, const uint8_t *in, size_t in_len);
 
 bool Ciphers(const std::vector<std::string> &args);
 bool Client(const std::vector<std::string> &args);
+// DoClient is a common function used to support the s_client option in both
+// bssl and openssl tools. It takes an additional parameter |tool| to indicate
+// which tool's s_client is being invoked. 1 indicates openssl and 0 indicates
+// bssl.
+bool DoClient(std::map<std::string, std::string> args_map, int tool);
```
samuel40791765, line 144: We can also make `tool` a boolean here. The `tool` naming also applies to both, so we should rename it to something like `is_openssl_s_client` , so it makes more sense when set to true or false.

tool/client.cc
```diff
@@ -245,19 +245,89 @@ static bool WaitForSession(SSL *ssl, int sock) {
...
+        (int)BIO_number_read(SSL_get_rbio(ssl)),
+        (int)BIO_number_written(SSL_get_wbio(ssl)));
+  print_verify_details(ssl);
+  X509_free(peer);
+}
+
 static bool DoConnection(SSL_CTX *ctx,
                          std::map<std::string, std::string> args_map,
-                         bool (*cb)(SSL *ssl, int sock)) {
+                         bool (*cb)(SSL *ssl, int sock), int tool) {
```
samuel40791765, line 315: Same boolean comment here, but I think we could also rename the parameter name to be more clearer. Probably could also consider using `is_openssl_s_client` here

tool/client.cc
```diff
@@ -378,18 +453,86 @@ static void InfoCallback(const SSL *ssl, int type, int value) {
...
 }
 
-bool Client(const std::vector<std::string> &args) {
-  if (!InitSocketLibrary()) {
-    return false;
+static int verify_cb(int ok, X509_STORE_CTX *ctx)
+{
+  X509 *err_cert;
+  int err, depth;
+  BIO* bio_err = BIO_new_fp(stdout, BIO_CLOSE);
```
samuel40791765, line 460: Should this point to `stdout` or `stderr`? Only asking because the `bio_err` naming seems to indicate `stderr`, ~~but no strong opinions on which should be used here.~~ Rest of the code seems to be using `stdout`

tool/client.cc
```diff
@@ -378,18 +453,86 @@ static void InfoCallback(const SSL *ssl, int type, int value) {
...
+
+  err_cert = X509_STORE_CTX_get_current_cert(ctx);
+  err = X509_STORE_CTX_get_error(ctx);
+  depth = X509_STORE_CTX_get_error_depth(ctx);
```
samuel40791765, line 464: Minor nit: we can create and directly assign the variables in the same line here

tool/client.cc
```diff
@@ -245,19 +245,89 @@ static bool WaitForSession(SSL *ssl, int sock) {
...
+
+static void PrintOpenSSLConnectionInfo(SSL *ssl, int show_certs) {
+  STACK_OF(X509) *sk;
+  X509 *peer;
+  int i;
+
+  sk = SSL_get_peer_cert_chain(ssl);
+  if (sk != NULL) {
+    fprintf(stdout, "---\nCertificate chain\n");
+    for (i = 0; i < (int) sk_X509_num(sk); i++) {
```
samuel40791765, line 267:
```suggestion
    for (size_t i = 0; i < sk_X509_num(sk); i++) {
```

tool/client.cc
```diff
@@ -245,19 +245,89 @@ static bool WaitForSession(SSL *ssl, int sock) {
...
+    fprintf(stdout, "Verification: OK\n");
+  } else {
+    const char *reason = X509_verify_cert_error_string(verify_err);
+    fprintf(stdout, "Verification error: %s\n", reason);
+  }
+}
+
+static void PrintOpenSSLConnectionInfo(SSL *ssl, int show_certs) {
+  STACK_OF(X509) *sk;
+  X509 *peer;
```
samuel40791765, line 261: Same boolean comment for `show_certs` and we can use a unique pointer to manage `peer`. It'd also be preferable to check the return values of all the non-void function calls below as well.

tool/client.cc
smittals2: you are correct, should be stderr - changed it :)

tool/internal.h
```diff
@@ -137,6 +137,11 @@ bool WriteToFile(const std::string &path, const uint8_t *in, size_t in_len);
 
 bool Ciphers(const std::vector<std::string> &args);
 bool Client(const std::vector<std::string> &args);
+// DoClient is a common function used to support the s_client option in both
+// bssl and openssl tools. It takes an additional parameter |tool| to indicate
+// which tool's s_client is being invoked. 1 indicates openssl and 0 indicates
+// bssl.
```
samuel40791765, line 143: May need to change the documentation around this to reflect the new naming/type

tool-openssl/s_client.cc
```diff
@@ -0,0 +1,46 @@
...
+    return false;
+  }
+
+  if(args_map.count("help")) {
+    fprintf(stderr, "Usage: s_client [options] [host:port]\n");
+    PrintUsage(kArguments);
+    return false;
+  }
+
+  return DoClient(args_map, 1);
```
samuel40791765, line 45: Would be best to change the other calls to `DoClient` as well
```suggestion
  return DoClient(args_map, true);
```

smittals2: Yep sorry verified all calls

tool/client.cc
```diff
@@ -540,24 +682,58 @@ bool Client(const std::vector<std::string> &args) {
...
+  std::string certPathFlag;
+  int verify = SSL_VERIFY_NONE;
   if (args_map.count("-root-certs") != 0) {
+    certPathFlag = "-root-certs";
+    verify = SSL_VERIFY_PEER;
+  }
+  // For the OpenSSL tool, simply specifying -CAfile does not imply verification
+  if (args_map.count("-CAfile") != 0) {
+    certPathFlag = "-CAfile";
+  }
```
justsmth, line 694: * Should we only allow one of the four options (`-root-certs`, `-CAfile`, `-root-cert-dir` or `-CApath`) to be specified in a single invocation, otherwise produce an error? 
  * It seems clear that at least certain combinations should not be allowed b/c they could lead to unexpected (or misleading) outcomes. 
  * For example, `-root-certs` and `-CAfile` would cause the `-root-certs` parameter to be ignored, but our `SSL_VERIFY_PEER` flag would still be set. And the client might interpret the outcome as applying to the `-root-certs` parameter when actually that parameter was ignored.
  * Even if OpenSSL allows certain "invalid" combinations, if the outcome would likely mislead our consumer, we should diverge from OpenSSL's behavior to improve the security.

tool/client.cc
```diff
@@ -540,24 +682,58 @@ bool Client(const std::vector<std::string> &args) {
...
+  // For the OpenSSL tool, simply specifying -CApath does not imply verification
+  if (args_map.count("-CApath") != 0) {
+    certPathFlag = "-CApath";
+  }
+
+  if (!certPathFlag.empty()) {
     if (!SSL_CTX_load_verify_locations(
-            ctx.get(), nullptr, args_map["-root-cert-dir"].c_str())) {
+            ctx.get(), nullptr, args_map[certPathFlag].c_str())) {
```
justsmth, line 716: NP: The logic would be more robust (i.e., more likely to catch future bugs) if it used [`at`](https://en.cppreference.com/w/cpp/container/map/at) instead of [`operator[]`](https://en.cppreference.com/w/cpp/container/map/operator_at) to retrieve values.

justsmth:
I know that there are no OpenSSL comparison tests yet, but shouldn't we at least have a few "sanity tests" to verify fundamental behaviors?

tool/client.cc
```diff
@@ -540,24 +682,58 @@ bool Client(const std::vector<std::string> &args) {
...
-    SSL_CTX_set_verify(ctx.get(), SSL_VERIFY_PEER, nullptr);
   }
 
+  certPathFlag = "";
   if (args_map.count("-root-cert-dir") != 0) {
+    certPathFlag = "-root-cert-dir";
+    verify = SSL_VERIFY_PEER;
+  }
+  // For the OpenSSL tool, simply specifying -CApath does not imply verification
+  if (args_map.count("-CApath") != 0) {
```
justsmth, line 710: `-CApath` and `-CAfile` are not listed in `kArguments` at the top of the file.  Is that intentional?

tool/client.cc
smittals2: Ah so -root-certs and -root-cert-dir are equivalents to -CAfile and -CApath dir, but they are only specified for the bssl tool. CAfile and CApath are new for our openssl tool (which is different from the bssl tool). In the startup process for the openssl s_client tool, we parse the provided arguments ([link](https://github.com/aws/aws-lc/pull/1959/files/ca4212844eb02c773d8433481242a4d65d1198c1#diff-44dd5728650be0db0986f6c19ace232b650cbe797d21d413c58198d766e9d4d5R34)) and would fail if -root-certs or -root-cert-dir are provided (since they aren't part of the valid argument map for the openssl tool ([link](https://github.com/aws/aws-lc/pull/1959/files/ca4212844eb02c773d8433481242a4d65d1198c1#diff-44dd5728650be0db0986f6c19ace232b650cbe797d21d413c58198d766e9d4d5R9))).

tool/client.cc
smittals2: Yes this is intentional. We are utilizing the same body of logic for both the bssl and openssl tool. CApath and CAfile are for the openssl tool, root-cert and root-cert-dir are for the bssl tool (which are specified in this file).

smittals2:
> I know that there are no OpenSSL comparison tests yet, but shouldn't we at least have a few "sanity tests" to verify fundamental behaviors?

The core logic utilized is from the previously written bssl tooling. I can add in some tests to ensure the advertised options are parsed as expected

tool/client.cc
smittals2: Hmmm my understanding is operator[] will create a new key during lookup if it doesn't exist while at will throw an exception. Since we obtain the args_map after parsing against a known list of keys, and check whether that key exists (args_map.count) before accessing the index, will this be beneficial?

justsmth: So, the problem is that it "creates a new key during lookup" -- creation of a new key is not a desired outcome and could result in a silent failure. The lookup for a non-existent key here should fail loudly. Explicit failures are always preferable to silent failures.

To clarify, I think the current code has the necessary checks to prevent unintended keys being created. My concern is more about future changes that might accidentally introduce a bug that results in invalid/misleading results for consumers due to the silent failure.

tool/client.cc
justsmth: Since such combinations of parameters should be impossible, it could be helpful to have some debug assertions in this function to verify this as a precondition.

justsmth (approved):

samuel40791765 (approved):

tool/client.cc
smittals2: noted, will address in follow up PR.

PR #1964, opened 2024-11-04
ML-KEM keygen Pairwise Consistency Test

dkostic:
### Issues:
N/A

### Description of changes: 
This change implements the ML-KEM Pairwise Consistency Test (PCT)
for key generation as required by FIPS 140-3 Implementation Guidance.

### Call-outs:

### Testing:

### DIFF
--- a/crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
+++ b/crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
@@ -1,14 +1,17 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <string.h>
-#include "params.h"
-#include "kem.h"
-#include "indcpa.h"
-#include "verify.h"
-#include "reduce.h"
-#include "symmetric.h"
+#include "./params.h"
+#include "./kem.h"
+#include "./indcpa.h"
+#include "./verify.h"
+#include "./reduce.h"
+#include "./symmetric.h"
+#include "../../../internal.h"
+
 #include "openssl/rand.h"
 
+#if defined(AWSLC_FIPS)
 // FIPS 203. Pair-wise Consistency Test (PCT) required per [FIPS 140-3 IG](https://csrc.nist.gov/csrc/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf):
 // The PCT consists of applying the encapsulation key to encapsulate a shared
 // secret leading to ciphertext, and then applying decapsulation key to
@@ -23,6 +26,7 @@ static int keygen_pct(ml_kem_params *params, const uint8_t *ek, const uint8_t *d
 
   return verify(ss_enc, ss_dec, KYBER_SSBYTES);
 }
+#endif
 
 /*************************************************
 * Name:        crypto_kem_keypair_derand
@@ -50,10 +54,12 @@ int crypto_kem_keypair_derand(ml_kem_params *params,
   /* Value z for pseudo-random output on reject */
   memcpy(sk+params->secret_key_bytes-KYBER_SYMBYTES, coins+KYBER_SYMBYTES, KYBER_SYMBYTES);
 
+#if defined(AWSLC_FIPS)
   // Abort in case of PCT failure.
   if (keygen_pct(params, pk, sk)) {
     BORINGSSL_FIPS_abort();
   }
+#endif
   return 0;
 }
 
@@ -77,6 +83,7 @@ int crypto_kem_keypair(ml_kem_params *params,
   uint8_t coins[2*KYBER_SYMBYTES];
   RAND_bytes(coins, 2*KYBER_SYMBYTES);
   int res = crypto_kem_keypair_derand(params, pk, sk, coins);
+  assert(res == 0);
 
   // FIPS 203. Section 3.3 Destruction of intermediate values.
   OPENSSL_cleanse(coins, sizeof(coins));
--- a/crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
+++ b/crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
@@ -9,6 +9,21 @@
 #include "symmetric.h"
 #include "openssl/rand.h"
 
+// FIPS 203. Pair-wise Consistency Test (PCT) required per [FIPS 140-3 IG](https://csrc.nist.gov/csrc/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf):
+// The PCT consists of applying the encapsulation key to encapsulate a shared
+// secret leading to ciphertext, and then applying decapsulation key to
+// retrieve the same shared secret. Returns 0 if the PCT passes, 1 otherwise.
+static int keygen_pct(ml_kem_params *params, const uint8_t *ek, const uint8_t *dk) {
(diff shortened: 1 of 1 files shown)


crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -34,6 +49,11 @@ int crypto_kem_keypair_derand(ml_kem_params *params,
   hash_h(sk+params->secret_key_bytes-2*KYBER_SYMBYTES, pk, params->public_key_bytes);
   /* Value z for pseudo-random output on reject */
   memcpy(sk+params->secret_key_bytes-KYBER_SYMBYTES, coins+KYBER_SYMBYTES, KYBER_SYMBYTES);
+
+  // Abort in case of PCT failure.
+  if (keygen_pct(params, pk, sk)) {
+    BORINGSSL_FIPS_abort();
```
justsmth, line 55: * The declaration of `BORINGSSL_FIPS_abort` is [several directories up](https://github.com/aws/aws-lc/blob/90d2a34ead8d51c3a97be33f1e6eff751707807c/crypto/internal.h#L1270):
```
#include "./../../../internal.h"
```
* `BORINGSSL_FIPS_abort` is only available in a FIPS build, but the logic should abort regardless.
```
  if (keygen_pct(params, pk, sk)) {
#if defined(BORINGSSL_FIPS)
    BORINGSSL_FIPS_abort();
#else
    abort();
#endif
  }
```
*

nebeid, line 55: For the non-FIPS tests to pass, the abort will need to be surrounded by `#if defined(AWSLC_FIPS)` as in https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/ec/ec_key.c#L549.

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -9,6 +9,21 @@
 #include "symmetric.h"
 #include "openssl/rand.h"
 
+// FIPS 203. Pair-wise Consistency Test (PCT) required per [FIPS 140-3 IG](https://csrc.nist.gov/csrc/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf):
+// The PCT consists of applying the encapsulation key to encapsulate a shared
+// secret leading to ciphertext, and then applying decapsulation key to
+// retrieve the same shared secret. Returns 0 if the PCT passes, 1 otherwise.
+static int keygen_pct(ml_kem_params *params, const uint8_t *ek, const uint8_t *dk) {
```
justsmth, line 16: NP: Would it be more clear to name  `ek` as `pk` (public key), and `dk` as `sk` (secret key)? These are the names/terms used by the caller below.

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -48,19 +68,20 @@ int crypto_kem_keypair_derand(ml_kem_params *params,
 *              - uint8_t *sk: pointer to output private key
 *                (an already allocated array of KYBER_SECRETKEYBYTES bytes)
 *
-* Returns 0 (success)
+* Returns 0 on success, aborts on failure.
 **************************************************/
 int crypto_kem_keypair(ml_kem_params *params,
```
nebeid, line 73: Is it possible to add a death test in `crypto/evp_extra/evp_extra_test.cc` as done in https://github.com/aws/aws-lc/blob/25709caeecb4f8206afc75bfd931599cf741b8dd/crypto/fipsmodule/ec/ec_test.cc#L1493

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -48,19 +68,20 @@ int crypto_kem_keypair_derand(ml_kem_params *params,
...
                        uint8_t *pk,
                        uint8_t *sk)
 {
   uint8_t coins[2*KYBER_SYMBYTES];
   RAND_bytes(coins, 2*KYBER_SYMBYTES);
-  crypto_kem_keypair_derand(params, pk, sk, coins);
+  int res = crypto_kem_keypair_derand(params, pk, sk, coins);
```
justsmth, line 79: NP: Add an assert that the return value was 0:
```
assert(res == 0);
```

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
justsmth: NP: It would be more clear if the header files from this source directory (at the top of the file) more clearly indicated that they're relative paths:
```
#include "./params.h"
#include "./kem.h"
#include "./indcpa.h"
#include "./verify.h"
#include "./reduce.h"
#include "./symmetric.h"
```

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
dkostic: we are moving towards using `ek` and `dk` to align with FIPS 203 naming

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
dkostic: not really because the PCT is part of keygen itself (and KEM params are hard-coded constants). In a future change I think I'll pull out the PCT as a separate function and add it as one of KEM methods and then call it in the high-level API. Maybe then we'll be able to implement a death test.

nebeid (approved):
Are there other checks that give assurance about the generated key in non-FIPS builds?

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -22,7 +41,7 @@
 *              - uint8_t *coins: pointer to input randomness
 *                (an already allocated array filled with 2*KYBER_SYMBYTES random bytes)
 **
-* Returns 0 (success)
+* Returns 0 on success, aborts on failure.
```
justsmth, line 44: NP: It only "aborts on failure" when in FIPS mode.

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -48,19 +74,21 @@ int crypto_kem_keypair_derand(ml_kem_params *params,
 *              - uint8_t *sk: pointer to output private key
 *                (an already allocated array of KYBER_SECRETKEYBYTES bytes)
 *
-* Returns 0 (success)
+* Returns 0 on success, aborts on failure.
```
justsmth, line 77: NP: It only "aborts on failure" when in FIPS mode.

crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c
```diff
@@ -34,6 +53,13 @@ int crypto_kem_keypair_derand(ml_kem_params *params,
   hash_h(sk+params->secret_key_bytes-2*KYBER_SYMBYTES, pk, params->public_key_bytes);
```
justsmth, line 53: NP: This [hash operation](https://github.com/aws/aws-lc/blob/6a6ead0effbf7e24f0b3320ee890fd0799e7e492/crypto/fipsmodule/sha/internal.h#L374) documents that it returns NULL on failure. Should we be checking that?

justsmth (approved):

PR #1966, opened 2024-11-05
Fix i2d behavior for i2d_SSL_SESSION

samuel40791765:
### Issues:
Addresses `CryptoAlg-2700`

### Description of changes: 
`i2d_SSL_SESSION` wasn't exactly following the correct behavior of the legacy `i2d` functions when `pp` is non-NULL, but `*pp` is NULL. This caused issues with functions expecting a newly allocated buffer when calling `i2d_SSL_SESSION` with the recommended documented behavior. See `i2d_SAMPLE` for more details.
https://github.com/aws/aws-lc/blob/48a40573eae3b09ef9b0c3d2d0180506fb214b0c/include/openssl/asn1.h#L254-L266

### Call-outs:
N/A

### Testing:
New test below usage of `i2d_SSL_SESSION` to verify the behavior.
 
### DIFF
--- a/ssl/ssl_asn1.cc
+++ b/ssl/ssl_asn1.cc
@@ -92,6 +92,7 @@
 #include <openssl/mem.h>
 #include <openssl/x509.h>
 
+#include "../crypto/bytestring/internal.h"
 #include "../crypto/internal.h"
 #include "internal.h"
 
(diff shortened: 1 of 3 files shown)


ssl/ssl_asn1.cc
```diff
@@ -841,20 +842,14 @@ int i2d_SSL_SESSION(SSL_SESSION *in, uint8_t **pp) {
...
-    OPENSSL_free(out);
-    OPENSSL_PUT_ERROR(SSL, ERR_R_OVERFLOW);
-    return -1;
-  }
-
-  if (pp) {
-    OPENSSL_memcpy(*pp, out, len);
-    *pp += len;
+  ScopedCBB cbb;
+  if (!CBB_init(cbb.get(), 256) || !CBB_add_bytes(cbb.get(), out, len)) {
```
WillChilds-Klein, line 848: do we need to advance `*pp`?

```
 // If |outp| and |*outp| are non-NULL, the function writes the result to 
 // |*outp|, which must have enough space available, and advances |*outp| just 
 // past the output. 
```

samuel40791765: Yes the very useful `CBB_finish_i2d` happens to handle all of the `i2d` shenanigans for us :) 

https://github.com/aws/aws-lc/blob/65d8c946b7ec57d258f35158d02d0f6f90c86eff/crypto/bytestring/internal.h#L60-L67

WillChilds-Klein: nice. looking back i see we assert that on L2040 of the test.

torben-hansen (approved):

WillChilds-Klein (approved):

PR #1989, opened 2024-11-13
Cleanup test File utilities

justsmth:
### Issues:
Addresses: P168988559

### Description of changes: 
* Only use move semantic for `TemporaryFile` and `ScopedFD`.
* Set umask prior to creating temporary file.

### DIFF
--- a/crypto/test/file_util.cc
+++ b/crypto/test/file_util.cc
@@ -122,7 +122,9 @@ bool TemporaryFile::Init(bssl::Span<const uint8_t> content) {
 #else
   std::string path = temp_dir + "bssl_tmp_file.XXXXXX";
   // TODO(davidben): Use |path.data()| when we require C++17.
+  mode_t prev_umask = umask(S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH);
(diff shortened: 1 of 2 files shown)


crypto/test/file_util.cc
```diff
@@ -122,7 +122,9 @@ bool TemporaryFile::Init(bssl::Span<const uint8_t> content) {
 #else
   std::string path = temp_dir + "bssl_tmp_file.XXXXXX";
   // TODO(davidben): Use |path.data()| when we require C++17.
+  mode_t prev_umask = umask(S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH);
```
smittals2, line 125: Where did this mask come from?

smittals2 (approved):

crypto/test/file_util.cc
justsmth: This particular just ensures that Read/Write permissions for "group" and "others" will not be granted when the file is created. (Typically, created files allow others to read.) https://man7.org/linux/man-pages/man2/umask.2.html

PR #1996, opened 2024-11-15
Implement PKCS7_encrypt and PKC7_decrypt

WillChilds-Klein:
### Issues:
Addresses CryptoAlg-2494

### Description of changes: 
This PR adds 2 new functions to encrypt/decrypt BIO contents into/out of "enveloped"-type PKCS7 objects. 

### Call-outs:

Like OpenSSL, this implementation of `PKCS7_decrypt` contains mitigations against the "Million Message Attack" (MMA) as prescribed in [RFC 3218](https://www.rfc-editor.org/rfc/rfc3218). A more detailed description is given in source comments.

### Testing:
- new unit tests. further coverage to be added in PR 1993 and beyond

---


WillChilds-Klein:
Interesting... `aws-lc-ci-linux-x86`'s `ubuntu2004_clang8x_x86_64` job (and only that job) [has failed](https://us-west-2.codebuild.aws.amazon.com/project/eyJlbmNyeXB0ZWREYXRhIjoiWk5IUWJGRGxBcTJ2Mkp4WGF3dnBwYjc5V0ZZYSt5SVVGbkwvODkydTNTaVQ2V2FMN3hwa0tjSWNFemw2QWtCWW5welFWV3lpRFpKVitwejgvelhpRWh3NDNqcWhKalpPYW9hL2tLMDlJSDFPT1NkNyIsIml2UGFyYW1ldGVyU3BlYyI6Ilp4VXRNQXFGM1BZYVlaRkIiLCJtYXRlcmlhbFNldFNlcmlhbCI6MX0%3D/build/c4036ecc-81f1-41a4-9b19-aeccab8cbe79) with the following error:

```
...
[ RUN      ] PKCS7Test.TestEnveloped
../crypto/pkcs7/pkcs7_test.cc:1813: Failure
Expected equality of these values:
  max_decrypt - 1
    Which is: 79
  decrypted_len
    Which is: 78
...
```

I suspect this is due to a different flavor of the same MMA defense edge case accounted for on L1812 of the test -- random occurrence of valid PKCS#7 ciphertext padding (note that this is about padding for symmetrically encrypted content, _not_ the asymmetric key encryption attacked by MMA). Originally, we accounted for one byte of randomly valid padding (i.e. `0x01`) occurring with probability 1 / 16 for AES (16 bytes is AES block size). Two bytes of randomly valid padding would be `0x02 0x02` occurring with probability 1 / 16^2 -- not common, but not rare. I'll think about how we can better account for this in our MMA countermeasure tests...

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1062,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+  //    to output gibberish "plaintext"
+  // 4. Always pay same allocation costs, regardless of |cek| decrypt result
+
+  // If |pcert| was specified, find the matching recipient info
+  if (pcert) {
+    for (size_t ii = 0; ii < sk_PKCS7_RECIP_INFO_num(rsk); ii++) {
+      ri = sk_PKCS7_RECIP_INFO_value(rsk, ii);
+      // No decryption operation here, so we can return early without divulging
+      // information that could be used in MMA.
+      if (!pkcs7_cmp_ri(ri, pcert)) {
```
justsmth, line 1258: There is ambiguity in the return value of `pkcs7_cmp_ri`.  When it returns 0, it can either mean that the two parameters are equal or that an error occurred while processing. (It might actually be better for it to return 1 on an error.) However, I think the logic here needs to distinguish these two cases -- when an error occurs, we would not want to subsequently call `pkcs7_decrypt_rinfo`.

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1062,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+
+  ret = 1;
+  *ek_out = ek;
+
+err:
+  EVP_PKEY_CTX_free(ctx);
+  return ret;
+}
+
+static int pkcs7_cmp_ri(PKCS7_RECIP_INFO *ri, X509 *pcert) {
```
samuel40791765, line 1191: We should probably document that this returns `0` if the comparison is equal

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1062,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+err:
+  OPENSSL_free(cek);
+  OPENSSL_free(dummy_key);
+  BIO_free_all(out);
+  BIO_free_all(cipher_bio);
+  BIO_free_all(data_bio);
+  return NULL;
+}
+
+PKCS7_RECIP_INFO *PKCS7_add_recipient(PKCS7 *p7, X509 *x509) {
```
samuel40791765, line 1358: We should null check `p7` and `x509`

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1062,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+
+  if ((bio = pkcs7_data_decode(p7, pkey, cert)) == NULL ||
+      !pkcs7_bio_copy_content(bio, data)) {
+    OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_DECRYPT_ERROR);
+    goto err;
+  }
+
+  // Check whether content decryption was successful
+  OPENSSL_BEGIN_ALLOW_DEPRECATED
+  if (!BIO_get_cipher_status(bio)) {
```
samuel40791765, line 1397: `BIO_get_cipher_status` could return `-2`, so it's probably best we make this check more specific

include/openssl/pkcs7.h
```diff
@@ -359,6 +359,31 @@ OPENSSL_EXPORT OPENSSL_DEPRECATED int PKCS7_set_digest(PKCS7 *p7, const EVP_MD *
 // |PKCS7_RECIP_INFO| or NULL if none are present.
 OPENSSL_EXPORT OPENSSL_DEPRECATED STACK_OF(PKCS7_RECIP_INFO) *PKCS7_get_recipient_info(PKCS7 *p7);
 
+// BIO_get_cipher_status returns 1 if the cipher is in a healthy state or 0
+// otherwise. Unhealthy state could indicate decryption failure or other
+// abnormalities. Data read from an unhealthy cipher should not be considered
+// authentic.
+OPENSSL_EXPORT OPENSSL_DEPRECATED int BIO_get_cipher_status(BIO *b);
```
samuel40791765, line 366: Does this need to be external, it seems like we can get away with defining this in an internal header

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1062,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+      if (tmp_cek && !cek) {
+        cek = tmp_cek;
+      }
+    }
+  }
+  // Clear any decryption errors to minimize behavioral difference under MMA
+  ERR_clear_error();
+
+  EVP_CIPHER_CTX *evp_ctx = NULL;
+  BIO_get_cipher_ctx(cipher_bio, &evp_ctx);
```
samuel40791765, line 1295: We should check the return value of `BIO_get_cipher_ctx`

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1062,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+  out = cipher_bio;
+
+  if (data_body && data_body->length > 0) {
+    data_bio = BIO_new_mem_buf(data_body->data, data_body->length);
+  } else {
+    data_bio = BIO_new(BIO_s_mem());
+    if (data_bio == NULL) {
+      goto err;
+    }
+    BIO_set_mem_eof_return(data_bio, 0);
```
samuel40791765, line 1341: We should check `BIO_set_mem_eof_return` as well

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1062,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+err:
+  EVP_PKEY_CTX_free(ctx);
+  return ret;
+}
+
+static int pkcs7_cmp_ri(PKCS7_RECIP_INFO *ri, X509 *pcert) {
+  GUARD_PTR(ri);
+  GUARD_PTR(pcert);
+  int ret =
+      X509_NAME_cmp(ri->issuer_and_serial->issuer, X509_get_issuer_name(pcert));
```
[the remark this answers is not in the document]
WillChilds-Klein: good point. i'll replace the `GUARD_PTR` usages with non-zero return value.

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1062,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+  }
+  size_t len;
+  if (!EVP_PKEY_decrypt(ctx, NULL, &len, ri->enc_key->data,
+                        ri->enc_key->length) ||
+      (ek = OPENSSL_malloc(len)) == NULL) {
+    OPENSSL_PUT_ERROR(EVP, ERR_R_EVP_LIB);
+    return 0;
+  }
+
+  // Do not update |ret| on decryption failure, simply null out |*pek|
```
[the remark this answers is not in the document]
WillChilds-Klein: updated.

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1062,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+    OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_WRONG_CONTENT_TYPE);
+    goto err;
+  }
+  if (!PKCS7_set_cipher(p7, cipher)) {
+    OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_ERROR_SETTING_CIPHER);
+    goto err;
+  }
+
+  for (size_t i = 0; i < sk_X509_num(certs); i++) {
+    x509 = sk_X509_value(certs, i);
```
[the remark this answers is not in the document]
WillChilds-Klein: Yes, I believe so. Of [all the `EVP_PKEY_METHODS`s](https://github.com/aws/aws-lc/blob/85f58da29e68906d094d378fe968f6b90807d05f/crypto/evp_extra/p_methods.c#L10-L32), RSA is the only one where encrypt/decrypt apply. KEMs are similar, but they have distinct encaps/decaps functions and don't define [encrypt/decrypt](https://github.com/aws/aws-lc/blob/85f58da29e68906d094d378fe968f6b90807d05f/crypto/fipsmodule/evp/p_kem.c#L316-L317). I'll add relevant checks and update documentation.

include/openssl/pkcs7.h
WillChilds-Klein: hm. good point. we use it in the tests, but we should still be able to move it to internal.h

crypto/pkcs7/pkcs7.c
WillChilds-Klein: `BIO_set_mem_eof_return` returns 1 in all cases today, but i suppose that could change in the future.

crypto/pkcs7/pkcs7.c
WillChilds-Klein: maybe for openssl, but I don't think it can return -2 for us. we're pretty careful in [the cipher BIO](https://github.com/aws/aws-lc/blob/main/crypto/pkcs7/bio/cipher.c) to keep `ctx->ok` either 0 or 1.

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1061,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+    goto err;
+  }
+
+  // RFC 3218 provides an overview of, and mitigations for, the "Million Message
+  // Attack" (MMA) on RSA encryption with PKCS-1 padding. Section 2.3 describes
+  // implementor countermeasures. We implement the following countermeasures, as
+  // does OpenSSL.
+  //
+  // 1. Do not branch on |cek| decryption failure when checking recip infos
+  // 2. Clear error state after |cek| decrypt is attempted
```
dkostic, line 1247: why would we even fill the error state when decrypting |cek|?

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1061,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+  } else {
+    // Attempt to decrypt every recipient info. Don't exit early as
+    // countermeasure for MMA.
+    for (size_t ii = 0; ii < sk_PKCS7_RECIP_INFO_num(rsk); ii++) {
+      ri = sk_PKCS7_RECIP_INFO_value(rsk, ii);
+      uint8_t *tmp_cek;
+      // |pkcs7_decrypt_rinfo| will only return false on critical failure, not
+      // on decryption failure. Check whether |tmp_cek| is present after the
+      // call to determine if decryption succeeded.
+      if (!pkcs7_decrypt_rinfo(&tmp_cek, ri, pkey)) {
```
dkostic, line 1282: if more than one call succeeds you will have memory leak, right? each success allocates new memory for `tmp_cek` but you only keep ownership of the first one.

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1061,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+        goto err;
+      }
+      // Set |cek| to the first successful decryption and keep going
```
dkostic, line 1285: just curious, OpenSSL seems not to be doing that, isn't that a behavioral difference between the two libraries?

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1061,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+    goto err;
+  }
+  // Always generate random bytes for the dummy key, regardless of |cek| decrypt
+  dummy_key = OPENSSL_malloc(len);
+  RAND_bytes(dummy_key, len);
+  // At this point, null |cek| indicates that no content encryption key was
+  // successfully decrypted. We don't want to return early due to MMA. So, swap
+  // in the dummy key and proceed. Content decryption result will be gibberish.
+  if (cek == NULL) {
+    cek = dummy_key;
```
dkostic, line 1322: this might result in a memory leak, right? you'll lose the pointer to the memory that `cek` pointed to previously and hence won't free it.

crypto/pkcs7/pkcs7.c
```diff
@@ -1086,3 +1061,347 @@ OPENSSL_END_ALLOW_DEPRECATED
...
+    OPENSSL_PUT_ERROR(EVP, ERR_R_EVP_LIB);
+    return 0;
+  }
+
+  int ok =
+      EVP_PKEY_decrypt(ctx, ek, &len, ri->enc_key->data, ri->enc_key->length);
+  // We only return 0 on critical failure, not decryption failure. However, we
+  // still need to set |ek| to NULL to signal decryption failure to callers so
+  // they can use random bytes as content encryption key for MMA defense.
+  if (!ok) {
```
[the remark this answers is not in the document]
WillChilds-Klein: My understanding is that MMA is (theoretically?) exploitable in 2 ways:

1. decrypt success or failure error code
2. timing differences between success and failure
From [section 2.2 of RFC 3218](https://www.rfc-editor.org/rfc/rfc3218.html#section-2.2):

```
   The MMA also requires that the attacker be able to distinguish cases
   where M' was PKCS-1 formatted from cases where it was not.  In the
   case of CMS the attacker will be sending CMS messages with C'
   replacing the wrapped CEK.  Thus, there are five possibilities:

   1. M' is improperly formatted.
   2. M' is properly formatted but the CEK is prima facie bogus (wrong
      length, etc.)
   3. M' is properly formatted and the CEK appears OK.  A signature or
      MAC is present so integrity checking fails.
   4. M' is properly formatted and no integrity check is applied.  In
      this case there is some possibility (approximately 1/32) that the
      CBC padding block will verify properly.  (The actual probability
      depends highly on the receiving implementation.  See "Note on
      Block Cipher Padding" below).  The message will appear OK at the
      CMS level but will be bogus at the application level.
   5. M' is properly formatted and the resulting CEK is correct.  This
      is extremely improbable but not impossible.

   The MMA requires the attacker to be able to distinguish case 1 from
   cases 2-4.  (He can always distinguish case 5, of course).  This
   might happen if the victim returned different errors for each case.
   The attacker might also be able to distinguish these cases based on
   timing -- decrypting the message and verifying the signature takes
   some time.  If the victim responds uniformly to all four errors then
   no attack is possible.
```

crypto/pkcs7/pkcs7.c
WillChilds-Klein: that's below our level of control. the EVP API will populate an error on RSA decrypt failure (i.e. invalid PKCS #1.5 padding exploited in the MMA attack).

crypto/pkcs7/pkcs7.c
WillChilds-Klein: yes, great catch.

crypto/pkcs7/pkcs7.c
WillChilds-Klein: no, in this case `cek` is NULL and doesn't point to anything. we then free the memory allocated on L1318 on L1332 below.

crypto/pkcs7/pkcs7.c
WillChilds-Klein: good [catch](https://github.com/openssl/openssl/blob/2de7e1d69851a363cadd9d6bdd95302b89a4383b/crypto/pkcs7/pk7_doit.c#L623). we should probably conform with OpenSSL.

crypto/pkcs7/pkcs7.c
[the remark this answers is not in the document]
dkostic: ok, so the timing difference that this `if` block creates can be used to distinguish case 5 from cases 1-4. But it can't help distinguish 1 from 2-4, right?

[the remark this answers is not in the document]
dkostic: sounds good, thank you!

dkostic (approved):

crypto/pkcs7/pkcs7.c
samuel40791765: Yeah I noticed that we were doing everything right in `cipher bio`'s code, it's just [this place](https://github.com/aws/aws-lc/blob/80f984ef78203bbdb61a2ef3fee55d8956ad7e8c/crypto/bio/bio.c#L361-L364) in `BIO_get_cipher_status` that leaves me wary.
It's unlikely, but if any user passes in a `bio` with weird function pointers, it may trigger a success here?

WillChilds-Klein: oh, yeah. i didn't notice that in `BIO_ctrl`. we should handle that correctly here. will update.

dkostic (approved):

crypto/pkcs7/internal.h
```diff
@@ -223,6 +223,12 @@ OPENSSL_EXPORT int BIO_set_cipher(BIO *b, const EVP_CIPHER *cipher,
                                   const unsigned char *key,
                                   const unsigned char *iv, int enc);
 
+// BIO_get_cipher_status returns 1 if the cipher is in a healthy state or 0
+// otherwise. Unhealthy state could indicate decryption failure or other
+// abnormalities. Data read from an unhealthy cipher should not be considered
+// authentic.
```
samuel40791765, line 229: Nit just in case we misuse this in the future, but we can include this in the next PR.
```suggestion
// BIO_get_cipher_status returns 1 if the cipher is in a healthy state or 0
// otherwise. A negative value could be returned if |b| is in an uninitialized
// state. Unhealthy state could indicate decryption failure or other
// abnormalities. Data read from an unhealthy cipher should not be considered
// authentic.
```

samuel40791765 (approved):

crypto/pkcs7/internal.h
WillChilds-Klein: thanks. will incorporate in next PR.

justsmth (approved):

PR #1999, opened 2024-11-18
Add internal APIs for ML-DSA

jakemas:
### Issues:
Resolves #CryptoAlg-2723

### Description of changes: 
This PR adds the internal functions from FIPS 204: Module-Lattice-Based Digital Signature Standard. We base this implementation on the upstream reference implementation of ML-DSA https://github.com/pq-crystals/dilithium/commit/444cdcc84eb36b66fe27b3a2529ee48f6d8150c2. However, the upstream commit only includes implementation of `ML-DSA.Sign_internal` and `ML-DSA.Verify_internal`, so we also include `ML-DSA.KeyGen_internal` to complete the implementation.

Changes:
* Implementation of external ML-DSA algorithm functions:
   * FIPS 204 Algorithm 1 `ML-DSA.KeyGen` as [crypto_sign_keypair](https://github.com/jakemas/aws-lc/blob/33dee338d7b0dd3e1254feb126987b08ceb92a39/crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c#L75-L95)
   * FIPS 204 Algorithm 2 `ML-DSA.Sign` as [crypto_sign_signature_internal](https://github.com/jakemas/aws-lc/blob/33dee338d7b0dd3e1254feb126987b08ceb92a39/crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c#L222-L270)
   * FIPS 204 Algorithm 3 `ML-DSA.Verify` as [crypto_sign_signature_internal](https://github.com/jakemas/aws-lc/blob/33dee338d7b0dd3e1254feb126987b08ceb92a39/crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c#L399-L437)
* Implementation of internal ML-DSA algorithm functions:
    * FIPS 204 Algorithm 6 `ML-DSA.KeyGen_internal` as [crypto_sign_keypair_internal](https://github.com/jakemas/aws-lc/blob/33dee338d7b0dd3e1254feb126987b08ceb92a39/crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c#L12-L30)
    * FIPS 204 Algorithm 7 `ML-DSA.Sign_internal` as [crypto_sign_signature_internal](https://github.com/jakemas/aws-lc/blob/33dee338d7b0dd3e1254feb126987b08ceb92a39/crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c#L97-L124)
    * FIPS 204 Algorithm 8 `ML-DSA.Verify_internal` as [crypto_sign_verify_internal](https://github.com/jakemas/aws-lc/blob/33dee338d7b0dd3e1254feb126987b08ceb92a39/crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c#L310-L335)
* Addition of internal functions to the `pqdsa` API:
    * ml_dsa_65_keypair_internal
    * ml_dsa_65_sign_internal
    * ml_dsa_65_verify_internal
* Addition of KATs from seeds (KATs from https://github.com/post-quantum-cryptography/KAT/tree/main)
    * New testing framework to run KATs from seeds (as in ACVP)
* Removal of `pq_custom_randombytes.{c/h}`

### Call-outs:
We can remove the testing mechanism for the KATs `pq_custom_randombytes` as we now support KATs that use the internal functions that provide randomness via an input seed.

### Testing:
The KATs have been migrated to use the internal functions.


crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -68,68 +73,86 @@ int crypto_sign_keypair(ml_dsa_params *params, uint8_t *pk, uint8_t *sk) {
...
+*                             array of CRYPTO_PUBLICKEYBYTES bytes)
+*              - uint8_t *sk: pointer to output private key (allocated
+*                             array of CRYPTO_SECRETKEYBYTES bytes)
+*
+* Returns 0 (success)
+**************************************************/
+int crypto_sign_keypair(ml_dsa_params *params, uint8_t *pk, uint8_t *sk) {
+  uint8_t seed[SEEDBYTES];
+  /* Get randomness for rho, rhoprime and key */
+  RAND_bytes(seed, SEEDBYTES);
```
andrewhop, line 92: RAND_bytes can never fail, but you should still check the return code so static analysis tools don't complain about not checking the return value.

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -68,68 +73,86 @@ int crypto_sign_keypair(ml_dsa_params *params, uint8_t *pk, uint8_t *sk) {
...
-    rnd[n] = 0;
-#endif
-  shake256(rhoprime, CRHBYTES, key, SEEDBYTES + RNDBYTES + CRHBYTES);
+  /* Compute rhoprime = CRH(key, rnd, mu) */
+  shake256_init(&state);
+  shake256_absorb(&state, key, SEEDBYTES);
+  shake256_absorb(&state, rnd, RNDBYTES);
+  shake256_absorb(&state, mu, CRHBYTES);
+  shake256_finalize(&state);
+  shake256_squeeze(rhoprime, CRHBYTES, &state);
```
andrewhop, line 155: How should I review these changes? They aren't immediately obvious like the keygen changes.

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -196,6 +219,56 @@ int crypto_sign_signature(ml_dsa_params *params,
...
+  for(i = 0; i < ctxlen; i++)
+    pre[2 + i] = ctx[i];
+
+#ifdef DILITHIUM_RANDOMIZED_SIGNING
+  if (!RAND_bytes(rnd, RNDBYTES))
+    return -1;
+#else
+  for(i=0;i<RNDBYTES;i++)
+    rnd[i] = 0;
+#endif
```
andrewhop, line 266: Why is this a compile time option and not two normal APIs for sign and sign_deterministic?

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -325,6 +396,46 @@ int crypto_sign_verify(ml_dsa_params *params,
...
+                       size_t mlen,
+                       const uint8_t *ctx,
+                       size_t ctxlen,
+                       const uint8_t *pk)
+{
+  size_t i;
+  uint8_t pre[257];
+
+  if(ctxlen > 255)
+    return -1;
```
andrewhop, line 429: Here and in other spots in this PR please update the code to always add the brackets even for single line statements. 
```suggestion
  if(ctxlen > 255) {
    return -1;
  }
```

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -19,7 +18,6 @@
 
 #include "../test/file_test.h"
 #include "../test/test_util.h"
-#include "../rand_extra/pq_custom_randombytes.h"
```
andrewhop, line 22: 🎉

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -373,65 +371,59 @@ TEST_P(PQDSAParameterTest, KAT) {
...
+
+    // Prepare m_prime = (0 || ctxlen || ctx) as in FIPS 204: Algorithm 2 line 10
+    size_t i;
+    uint8_t m_prime[257];
+    size_t m_prime_len = ctxstr.size() + 2;
+    m_prime[0] = 0;
+    m_prime[1] = ctxstr.size();
+    for(i = 0; i < ctxstr.size(); i++) {
+      m_prime[2 + i] = ctxstr[i];
+    }
```
andrewhop, line 413: Can this just use memcpy to be simpler and maybe faster?

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -196,6 +219,56 @@ int crypto_sign_signature(ml_dsa_params *params,
...
+  uint8_t rnd[RNDBYTES];
+
+  if(ctxlen > 255)
+    return -1;
+
+  /* Prepare pre = (0, ctxlen, ctx) */
+  pre[0] = 0;
+  pre[1] = ctxlen;
+  for(i = 0; i < ctxlen; i++)
+    pre[2 + i] = ctx[i];
```
andrewhop, line 258: memcpy here too? Either way needs {}

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: As mentioned in the PR description, we base this implementation on the upstream reference implementation of ML-DSA https://github.com/pq-crystals/dilithium/commit/444cdcc84eb36b66fe27b3a2529ee48f6d8150c2. However, the upstream commit only includes implementation of ML-DSA.Sign_internal and ML-DSA.Verify_internal, so we also include ML-DSA.KeyGen_internal to complete the implementation.

Thus, this PR is pulling changes from upstream (https://github.com/pq-crystals/dilithium/commit/444cdcc84eb36b66fe27b3a2529ee48f6d8150c2) for both ML-DSA.Sign_internal and ML-DSA.Verify_internal. To review, please verify that the implementation is consistent with upstream. Hope this helps!

crypto/dilithium/ml_dsa.h
```diff
@@ -15,22 +15,39 @@
...
 
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
 int ml_dsa_65_keypair(uint8_t *public_key,
-                      uint8_t *private_key);
+                      uint8_t *secret_key);
+
+int ml_dsa_65_keypair_internal(uint8_t *public_key,
```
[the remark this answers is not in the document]
jakemas: We used `deterministic` for ML-KEM as we wrote these functions before the final release of the standard came out, and we were working from the IPD and naming conventions from the PQC mailing list. Since the release of FIPS 203, 204 final, we now see these deterministic methods described as `internal`, so to make the implementation of the standard clearer I named them internal and added descriptors to the standard itself, e.g.:

```
/*************************************************
 * Name:        crypto_sign_keypair_internal
 *
 * Description: FIPS 204: Algorithm 6 ML-DSA.KeyGen_internal.
 *              Generates public and private key. Internal API.
 ...


```
To refer to:
Screenshot 2024-11-18 at 4 56 52 PM

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: I will add, that the upstream changes do make the implementation of the internal functions easier to directly check against the standard.

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -1,36 +1,41 @@
...
   const uint8_t *rho, *rhoprime, *key;
   polyvecl mat[DILITHIUM_K_MAX];
   polyvecl s1, s1hat;
   polyveck s2, t1, t0;
 
-  /* Get randomness for rho, rhoprime and key */
-  pq_custom_randombytes(seedbuf, SEEDBYTES);
+  OPENSSL_memcpy(seedbuf, seed, SEEDBYTES);
   seedbuf[SEEDBYTES+0] = params->k;
   seedbuf[SEEDBYTES+1] = params->l;
```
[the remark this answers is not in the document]
jakemas: This is performing line 1 of algorithm 6 FIPS 204:

Screenshot 2024-11-19 at 9 36 47 AM

The values of `k` and `l` are 1 byte values (defined in `params.h`) these are `(k, l) = (4, 4)` for ML-DSA-44, `(k, l) = (6, 5)` for ML-DSA-65, and `(k, l) = (8, 7)` for ML-DSA-87. (this is where the ML-DSA-XX names come from). 

In order to add domain separation between the three parameter sets, we include a parameter into the seed expansion hash. As `k,l` are well-defined, they fit in a single byte of `seedbuf`. Is there a better way to handle this?

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: Good point! For reference, I will call out that deterministic signing is a little different to the `determinism` in ML-KEM. For ML-KEM we used deterministic APIs for testing purposes, we do that here too for ML-DSA, however, deterministic signing modes is also a configurable option for ML-DSA. It's defined in FIPS 204 and can too be FIPS compliant. The non-deterministic mode, called "hedged" signing (as it hedges its bets by including a randomness source, as well as using the message itself to provide some entropy) is recommended for general purpose usage, and will be the configuration we primarily focus on.

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: added in [54ca109](https://github.com/aws/aws-lc/pull/1999/commits/54ca1098fc3bc9a0629a24931e041dc72561bda6)

crypto/dilithium/p_pqdsa_test.cc
jakemas: added in [54ca109](https://github.com/aws/aws-lc/pull/1999/commits/54ca1098fc3bc9a0629a24931e041dc72561bda6)

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: this was reference code, and I assume they did this to avoid a dependency on openssl. added in [54ca109](https://github.com/aws/aws-lc/pull/1999/commits/54ca1098fc3bc9a0629a24931e041dc72561bda6)

crypto/dilithium/pqcrystals_dilithium_ref_common/README.md
```diff
@@ -10,13 +10,12 @@ The code was refactored in [this PR](https://github.com/aws/aws-lc/pull/1910) by
...
 - `ntt.c`, `poly.c`, `reduce.c`, `reduce.h`: have been modified with a code refactor. The function `fqmul` has been added to bring mode code consistency with Kyber/ML-KEM. See https://github.com/aws/aws-lc/pull/1748 for more details on this change.
 - `reduce.c`: a small fix to documentation has been made on the bounds of `reduce32`.
 - `poly.c`: a small fix to documentation has been made on the bounds of `poly_reduce`.
 - `polyvec.c`: a small fix to documentation has been made on the bounds of `polyveck_reduce`.
 
 **Testing** 
 
-The KATs were obtained from https://github.com/pq-crystals/dilithium/tree/master/ref/nistkat.
-To compile the KAT programs on Linux or macOS, go to the `ref/` directory and run `make nistkat`. This will produce executables within `nistkat` which once executed will produce the KATs: `PQCsignKAT_Dilithium2.rsp`, `PQCsignKAT_Dilithium3.rsp`,`PQCsignKAT_Dilithium5.rsp`.
+The KATs were obtained from https://github.com/post-quantum-cryptography/KAT.
```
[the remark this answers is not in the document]
jakemas: added in [1bbbb24](https://github.com/aws/aws-lc/pull/1999/commits/1bbbb244c7d7aa3284394b288570256ef1a8144d)

crypto/dilithium/ml_dsa.h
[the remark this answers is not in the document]
andrewhop: Do we have an issue tracking the renaming of ML-KEM functions?

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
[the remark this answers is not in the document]
andrewhop: Oh, can we just change the size of k/l in params to be uint8_t?

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
andrewhop: Should sign/verify be called hedge to follow the standard?

crypto/dilithium/ml_dsa.h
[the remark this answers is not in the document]
jakemas: compiling little nits I come across here https://github.com/aws/aws-lc/issues/2002

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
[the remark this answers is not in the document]
jakemas: like this? [dc471ec](https://github.com/aws/aws-lc/pull/1999/commits/dc471ec45e5bb095f0db6b3f5c3373d2ee0cd200)

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: Firstly, no need for any changes to `verify`, as it works the same regardless if deterministic/hedged signatures were signed. As for should we change the name of `crypto_sign_signature` to `crypto_sign_signature_hedged`, I prefer not to.

The `hedged` term only gets spoken to deep within FIPS 204, and I worry from a customers perspective that this term may be confusing as an API name. I did add a comment to both `crypto_sign_signature` and `crypto_sign_signature_deterministic` that states if we are in hedged/deterministic mode (see https://github.com/jakemas/aws-lc/blob/c4a0fdb3ee8f189ce03f8ab6121f97f9e4b19ccc/crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c#L227 and https://github.com/jakemas/aws-lc/blob/c4a0fdb3ee8f189ce03f8ab6121f97f9e4b19ccc/crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c#L270.

I am open to suggestions if you disagree!

jakemas: Alternatively, we can remove the deterministic mode completely, and add it when it is required.

crypto/dilithium/ml_dsa.c
```diff
@@ -32,6 +32,14 @@ int ml_dsa_65_keypair(uint8_t *public_key   /* OUT */,
   return (crypto_sign_keypair(&params, public_key, private_key) == 0);
 }
 
+int ml_dsa_65_keypair_internal(uint8_t *public_key   /* OUT */,
```
dkostic, line 35: I'm not sure I understand why we need to define and expose the internal functions?

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -373,65 +371,56 @@ TEST_P(PQDSAParameterTest, KAT) {
...
-    ASSERT_TRUE(EVP_DigestSign(ctx.get(), signature.data(), &sig_len, msg.data(), mlen_int));
-    EXPECT_EQ(Bytes(sm), Bytes(signature.data(), sig_len));
-    ctx.Reset();
-
-    // Verify the signature
-    ASSERT_TRUE(EVP_DigestVerifyInit(ctx.get(), nullptr, nullptr, nullptr, pkey.get()));
-    ASSERT_TRUE(EVP_DigestVerify(ctx.get(), signature.data(), sig_len, msg.data(), mlen_int));
+    // Generate key pair from seed xi and assert that public and private keys
+    // are equal to expected values from KAT
+    ASSERT_TRUE(ml_dsa_65_keypair_internal(pub.data(),priv.data(),xi.data()));
```
dkostic, line 401: isn't this a generic test that should work for any PQDSA scheme? why are we hard-coding mldsa65 parameters and functions?

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.h
```diff
@@ -4,17 +4,33 @@
...
+                                 uint8_t *sk,
+                                 const uint8_t *seed);
+
 int crypto_sign_signature(ml_dsa_params *params,
                           uint8_t *sig, size_t *siglen,
                           const uint8_t *m, size_t mlen,
                           const uint8_t *ctx, size_t ctxlen,
                           const uint8_t *sk);
 
+int crypto_sign_signature_deterministic(ml_dsa_params *params,
```
dkostic, line 21: I don't see the deterministic variants used anywhere, why are we adding them?

crypto/dilithium/kat/mldsa65.txt
dkostic, line 1: why can't we keep these tests in addition to the new ones?

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
dkostic: I agree with removing the deterministic mode completely since it's not used.

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.h
jakemas: Removed in [aada875](https://github.com/aws/aws-lc/pull/1999/commits/aada875da98c6f173d11a001195b81f47cdcc451)

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: removed in [aada875](https://github.com/aws/aws-lc/pull/1999/commits/aada875da98c6f173d11a001195b81f47cdcc451)

crypto/dilithium/p_pqdsa_test.cc
jakemas: I have plans to modify when additional security levels were added. We can't use a generic EVP API that calls into specific algorithm functionality, as we don't have de-randomized EVP APIs for sign. So, I extract the name of the scheme from the `PQDSATestVector` struct and conditionally call the relevant APIs. See [aada875](https://github.com/aws/aws-lc/pull/1999/commits/aada875da98c6f173d11a001195b81f47cdcc451).

crypto/dilithium/kat/mldsa65.txt
jakemas: To keep these tests we'd need to keep `PQ_randbytes` etc. because the KATs from the upstream repo do not provide keygen/sign seeds. The KATs included from https://github.com/post-quantum-cryptography/KAT/tree/main/MLDSA provide assurance from an independent source that claims "Compliant with FIPS-204" from seeds, supporting the modes outlined in the standard.

crypto/dilithium/ml_dsa.c
jakemas: We need to expose them for ACVP testing, as part of the process of FIPS validation.

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -373,65 +371,61 @@ TEST_P(PQDSAParameterTest, KAT) {
...
+    // are equal to expected values from KAT
+    if (name == "MLDSA65") {
+      ASSERT_TRUE(ml_dsa_65_keypair_internal(pub.data(),priv.data(),xi.data()));
+    }
+    EXPECT_EQ(Bytes(pub), Bytes(pk));
+    EXPECT_EQ(Bytes(priv), Bytes(sk));
+
+    // Prepare m_prime = (0 || ctxlen || ctx) as in FIPS 204: Algorithm 2 line 10
+    uint8_t m_prime[257];
+    size_t m_prime_len = ctxstr.size() + 2;
```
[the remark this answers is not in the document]
jakemas: Summarizing an offline discussion:
`ctxstr` is an optional context string that is combined with the message to be signed. It's well defined in FIPS 204 as a byte string of 255 or fewer bytes. In fact, the signing function Algorithm 2: `ML-DSA.Sign` is responsible for ensuring that the string does not exceed this length, otherwise we return fail (see line 1 - 3 of Algorithm 2).

Line 10 of algorithm 2 describes `m_prime`, the structure we are creating; `M' = Bytestobits(0 || ctx_string_len || ctx ) || M`. You can see [from upstream](https://github.com/pq-crystals/dilithium/commit/444cdcc84eb36b66fe27b3a2529ee48f6d8150c2#diff-75ebb091b78aba35449cc4e4e3731978a8888d21ccfd47c15597518c4c556302R309) how we the external signing function pre-processes the input context string:

```
 uint8_t pre[257];
  /* Prepare pre = (0, ctxlen, ctx) */
  pre[0] = 0;
  pre[1] = ctxlen;
  memcpy(&pre[2], ctx, ctxlen);
  ```

We have to emulate this within `p_pqdsa_test.cc` as we call the internal function that expects this pre-processing to be performed.

File: crypto/dilithium/p_pqdsa_test.cc [hunk 15]
```diff
@@ -373,65 +371,61 @@ TEST_P(PQDSAParameterTest, KAT) {
...
-    // Verify the signature
-    ASSERT_TRUE(EVP_DigestVerifyInit(ctx.get(), nullptr, nullptr, nullptr, pkey.get()));
-    ASSERT_TRUE(EVP_DigestVerify(ctx.get(), signature.data(), sig_len, msg.data(), mlen_int));
+    // Generate key pair from seed xi and assert that public and private keys
+    // are equal to expected values from KAT
+    if (name == "MLDSA65") {
+      ASSERT_TRUE(ml_dsa_65_keypair_internal(pub.data(),priv.data(),xi.data()));
+    }
+    EXPECT_EQ(Bytes(pub), Bytes(pk));
+    EXPECT_EQ(Bytes(priv), Bytes(sk));
```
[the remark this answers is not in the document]
jakemas: Discussed offline! in future we will have an if-statement that performs keygen on MLDSA87

crypto/dilithium/p_pqdsa_test.cc
andrewhop: Also discussed passing in function pointers with the parametrized test instead of the string name, I'm okay with that coming later.

andrewhop (dismissed):

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -357,7 +355,7 @@ CMP_VEC_AND_PTR(vec, pkey->pkey.pqdsa_key->public_key, len)
 CMP_VEC_AND_PTR(vec, pkey->pkey.pqdsa_key->private_key, len)
 
 static const struct PQDSATestVector parameterSet[] = {
-  {"MLDSA65", NID_MLDSA65, 1952, 4032, 3309,  "dilithium/kat/mldsa65.txt", mldsa65kPublicKey, mldsa65kPublicKeySPKI, 1974},
+  {"MLDSA65", NID_MLDSA65, 1952, 4032, 3309,  "dilithium/kat/MLDSA_65_hedged_pure.txt", mldsa65kPublicKey, mldsa65kPublicKeySPKI, 1974},
```
dkostic, line 358: did we explain what does `hedged_pure` mean?

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -63,73 +68,92 @@ int crypto_sign_keypair(ml_dsa_params *params, uint8_t *pk, uint8_t *sk) {
...
+*              - uint8_t *pk: pointer to output public key (allocated
+*                             array of CRYPTO_PUBLICKEYBYTES bytes)
+*              - uint8_t *sk: pointer to output private key (allocated
+*                             array of CRYPTO_SECRETKEYBYTES bytes)
+*
+* Returns 0 (success) -1 on failure
+**************************************************/
+int crypto_sign_keypair(ml_dsa_params *params, uint8_t *pk, uint8_t *sk) {
+  uint8_t seed[SEEDBYTES];
+  /* Get randomness for rho, rhoprime and key */
```
dkostic, line 90:
```suggestion
```

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -63,73 +68,92 @@ int crypto_sign_keypair(ml_dsa_params *params, uint8_t *pk, uint8_t *sk) {
...
+                                   const uint8_t *pre,
+                                   size_t prelen,
+                                   const uint8_t *rnd,
+                                   const uint8_t *sk)
 {
   unsigned int n;
-  uint8_t seedbuf[2*SEEDBYTES + TRBYTES + RNDBYTES + 2*CRHBYTES];
-  uint8_t *rho, *tr, *key, *mu, *rhoprime, *rnd;
+  uint8_t seedbuf[2*SEEDBYTES + TRBYTES + 2*CRHBYTES];
+  uint8_t *rho, *tr, *key, *mu, *rhoprime;
```
dkostic, line 128: I'm not sure I understand why we are doing this super convoluted thing with a single buffer and pointers to parts of the buffer. Why aren't `rho, tr, ...` individual buffers?

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -63,73 +68,92 @@ int crypto_sign_keypair(ml_dsa_params *params, uint8_t *pk, uint8_t *sk) {
...
-  rnd = key + SEEDBYTES;
-  mu = rnd + RNDBYTES;
+  mu = key + SEEDBYTES;
   rhoprime = mu + CRHBYTES;
   unpack_sk(params, rho, tr, key, &t0, &s1, &s2, sk);
 
-  /* Compute mu = CRH(tr, 0, ctxlen, ctx, msg) */
-  mu[0] = 0;
-  mu[1] = ctxlen;
+  /* Compute mu = CRH(tr, pre, msg) */
```
dkostic, line 142: can you add references to lines in FIPS 204, Algorithm 7 please (for verify_internal and keygen_internal as well)? For example, I believe this comment refers to line 6 that says: 𝜇 ← H(BytesToBits(𝑡𝑟)||𝑀 ′, 64) ?

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -373,65 +371,61 @@ TEST_P(PQDSAParameterTest, KAT) {
...
-    // "sm" down to a constant DILITHIUM3_SIGNATURE_BYTES.
-
+    std::vector<uint8_t> pub(pk_len);
+    std::vector<uint8_t> priv(sk_len);
     std::vector<uint8_t> signature(sig_len);
-    sm.resize(sig_len);
 
-    // Convert string read from KAT to int
     size_t mlen_int = std::stoi(mlen);
+    sm.resize(sig_len);
```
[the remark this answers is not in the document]
jakemas: The KATs provide the signed message, which is the signature with the message appended at the end. We use the APIs for sign_signature and not sign_message, which return just the signature without the appended message on the end. So we truncate the signed message in the KAT to just the size of the signature.

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -373,65 +371,61 @@ TEST_P(PQDSAParameterTest, KAT) {
...
-    ASSERT_TRUE(EVP_DigestVerify(ctx.get(), signature.data(), sig_len, msg.data(), mlen_int));
+    // Generate key pair from seed xi and assert that public and private keys
+    // are equal to expected values from KAT
+    if (name == "MLDSA65") {
+      ASSERT_TRUE(ml_dsa_65_keypair_internal(pub.data(),priv.data(),xi.data()));
+    }
+    EXPECT_EQ(Bytes(pub), Bytes(pk));
+    EXPECT_EQ(Bytes(priv), Bytes(sk));
+
+    // Prepare m_prime = (0 || ctxlen || ctx) as in FIPS 204: Algorithm 2 line 10
```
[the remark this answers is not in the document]
jakemas: It's not ML-DSA specific, this same pre-processing is true of SLH-DSA: see https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.205.pdf Algorithm 22 `slh_sign` line 8 -- we do the same pre-processing for all PQ DSA.

crypto/dilithium/ml_dsa.h
```diff
@@ -15,22 +15,39 @@
...
-                   const uint8_t *ctx_string,
-                   size_t ctx_string_len);
+                   uint8_t *sig, size_t *sig_len,
+                   const uint8_t *message, size_t message_len,
+                   const uint8_t *ctx_string, size_t ctx_len);
+
+int ml_dsa_65_sign_internal(const uint8_t *private_key,
+                            uint8_t *sig, size_t *sig_len,
+                            const uint8_t *message, size_t message_len,
+                            const uint8_t *pre, size_t pre_len,
```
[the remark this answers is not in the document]
jakemas: This is intentional! `pre` and `ctx_string` are different things. The external functions take `ctx_string` and `ctx_string_len` as input, and internal functions take `pre` and `pre_len` as input. (This was added after FIPS 204, 205 IPD and before final standards).

`ctx` is an optional context message that is combined into the signature.
`pre` is the processed message to be signed, context message, and context message length.

The two names are intentional to make this distinction. This is consistant with the upstream implementation, see:

- internal naming example https://github.com/pq-crystals/dilithium/blob/444cdcc84eb36b66fe27b3a2529ee48f6d8150c2/ref/sign.c#L78-L79
- external naming example https://github.com/pq-crystals/dilithium/blob/444cdcc84eb36b66fe27b3a2529ee48f6d8150c2/ref/sign.c#L200-L201

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: removed in [061ebce](https://github.com/aws/aws-lc/pull/1999/commits/061ebceca90a012a0e8ba261f71c8724251a42ea)

crypto/dilithium/p_pqdsa_test.cc
jakemas: added explanation in readme KATs section in [061ebce](https://github.com/aws/aws-lc/pull/1999/commits/061ebceca90a012a0e8ba261f71c8724251a42ea)

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -373,65 +371,61 @@ TEST_P(PQDSAParameterTest, KAT) {
...
+    m_prime[1] = ctxstr.size();
+    OPENSSL_memcpy(m_prime + 2 , ctxstr.data(), ctxstr.size());
+
+    // Generate signature by signing |msg|, assert that signature is equal
+    // to expected value from KAT, then verify signature.
+    if (name == "MLDSA65") {
+      ASSERT_TRUE(ml_dsa_65_sign_internal(priv.data(),
+                                    signature.data(), &sig_len,
+                                    msg.data(), mlen_int,
+                                    m_prime,m_prime_len,
```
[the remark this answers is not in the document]
jakemas: fixed in [061ebce](https://github.com/aws/aws-lc/pull/1999/commits/061ebceca90a012a0e8ba261f71c8724251a42ea)

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -373,65 +371,61 @@ TEST_P(PQDSAParameterTest, KAT) {
...
-    ASSERT_TRUE(EVP_DigestSign(ctx.get(), signature.data(), &sig_len, msg.data(), mlen_int));
-    EXPECT_EQ(Bytes(sm), Bytes(signature.data(), sig_len));
-    ctx.Reset();
-
-    // Verify the signature
-    ASSERT_TRUE(EVP_DigestVerifyInit(ctx.get(), nullptr, nullptr, nullptr, pkey.get()));
-    ASSERT_TRUE(EVP_DigestVerify(ctx.get(), signature.data(), sig_len, msg.data(), mlen_int));
+    // Generate key pair from seed xi and assert that public and private keys
+    // are equal to expected values from KAT
+    if (name == "MLDSA65") {
```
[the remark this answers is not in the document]
jakemas: added a lookup/find in [061ebce](https://github.com/aws/aws-lc/pull/1999/commits/061ebceca90a012a0e8ba261f71c8724251a42ea)

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -373,65 +371,61 @@ TEST_P(PQDSAParameterTest, KAT) {
...
+      ASSERT_TRUE(ml_dsa_65_keypair_internal(pub.data(),priv.data(),xi.data()));
```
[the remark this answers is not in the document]
jakemas: fixed in [061ebce](https://github.com/aws/aws-lc/pull/1999/commits/061ebceca90a012a0e8ba261f71c8724251a42ea)

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -373,65 +371,61 @@ TEST_P(PQDSAParameterTest, KAT) {
...
+    }
+    EXPECT_EQ(Bytes(pub), Bytes(pk));
+    EXPECT_EQ(Bytes(priv), Bytes(sk));
+
+    // Prepare m_prime = (0 || ctxlen || ctx) as in FIPS 204: Algorithm 2 line 10
+    uint8_t m_prime[257];
+    size_t m_prime_len = ctxstr.size() + 2;
+    m_prime[0] = 0;
+    m_prime[1] = ctxstr.size();
+    OPENSSL_memcpy(m_prime + 2 , ctxstr.data(), ctxstr.size());
```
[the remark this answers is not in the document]
jakemas: added in [061ebce](https://github.com/aws/aws-lc/pull/1999/commits/061ebceca90a012a0e8ba261f71c8724251a42ea)

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: Yes, I'm staging a PR that adds a LOT of documentation and call out to each of the algorithms in FIPS 204 and their implementation here. 

However, I want to get the larger code integrations complete before I start touching every single file in the dilithium repo to add comments. This will prevent me from re-base/conflicts pain down the line.

I'll add specific call outs to the lines in sign/verify in that forthcoming PR.

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: This is upstream code (https://github.com/pq-crystals/dilithium/blob/444cdcc84eb36b66fe27b3a2529ee48f6d8150c2/ref/sign.c#L95-L96).

To aid review I'm trying to not touch upstream code too much, so we can compare against the reference implementation. I agree that a lot of the reference code is convoluted (probably because they are avoiding dependencies). However, I don't think we're at a stage of development in which we want to radically differ from the reference implementation. Unless necessary for security, or compliance to the standard.

If you have a cleaner proposal for this, I'm happy to take it -- I just don't want to be radically changing the upstream reference before you both get to review with no basis to understand these changes.

crypto/dilithium/p_pqdsa_test.cc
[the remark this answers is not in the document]
jakemas: For reference, I put the check outside the `FileTestGTest` so that we don't perform this check every iteration of the test vector loop. I tested functionality by renaming a test vector name.

crypto/dilithium/p_pqdsa_test.cc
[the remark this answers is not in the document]
jakemas: For reference, I put the check outside the FileTestGTest main loop so that we don't perform this check every iteration of the test vector loop. I tested functionality by renaming a test vector name.

crypto/dilithium/p_pqdsa_test.cc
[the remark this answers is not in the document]
dkostic: please add that as a comment to the code.

crypto/dilithium/p_pqdsa_test.cc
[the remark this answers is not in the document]
dkostic: ok, please add that as a comment.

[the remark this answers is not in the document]
jakemas: added in [7e78803](https://github.com/aws/aws-lc/pull/1999/commits/7e788036d16d820c8c5c212eee618e9ef843d565)

crypto/dilithium/p_pqdsa_test.cc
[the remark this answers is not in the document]
jakemas: added in [7e78803](https://github.com/aws/aws-lc/pull/1999/commits/7e788036d16d820c8c5c212eee618e9ef843d565)

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
dkostic: fair enough!

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
dkostic: sure, but the point of documentation is to make the review easier, once we merge the code it's not as useful :)

jakemas: Ahh I understand now, added these callouts to lines in the spec in [8c3d53e](https://github.com/aws/aws-lc/pull/1999/commits/8c3d53e8abe8cdd6783c64bc34e98352d44a83fd). I plan on adding a lot more documentation like this.

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -55,147 +60,214 @@ int crypto_sign_keypair(ml_dsa_params *params, uint8_t *pk, uint8_t *sk) {
...
-  mu = rnd + RNDBYTES;
+  mu = key + SEEDBYTES;
   rhoprime = mu + CRHBYTES;
+  /* FIPS 204: line 1 */
   unpack_sk(params, rho, tr, key, &t0, &s1, &s2, sk);
 
-  /* Compute mu = CRH(tr, 0, ctxlen, ctx, msg) */
-  mu[0] = 0;
-  mu[1] = ctxlen;
+  /* FIPS 204: line 6 Compute mu = CRH(tr, pre, msg) */
```
dkostic, line 143: Line 6 in the spec: 𝜇 ← H(BytesToBits(𝑡𝑟)||𝑀 ′, 64), where M' is an input argument to the function. Ideally we'd refactor the code to align with the spec, but we can leave that for later as you suggested. Can you please clarify in the comment this discrepancy between the implementation and the spec.

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -55,147 +60,214 @@ int crypto_sign_keypair(ml_dsa_params *params, uint8_t *pk, uint8_t *sk) {
...
   polyveck_invntt_tomont(params, &h);
   polyveck_sub(params, &w0, &w0, &h);
   polyveck_reduce(params, &w0);
-  if(polyveck_chknorm(params, &w0, params->gamma2 - params->beta))
+  if(polyveck_chknorm(params, &w0, params->gamma2 - params->beta)) {
     goto rej;
+  }
 
-  /* Compute hints for w1 */
+  /* FIPS 204: line 26 Compute hints for w1 */
```
dkostic, line 208:
```suggestion
  /* FIPS 204: line 25 */
```

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -55,147 +60,214 @@ int crypto_sign_keypair(ml_dsa_params *params, uint8_t *pk, uint8_t *sk) {
...
   polyveck_pointwise_poly_montgomery(params, &h, &cp, &t0);
   polyveck_invntt_tomont(params, &h);
   polyveck_reduce(params, &h);
-  if(polyveck_chknorm(params, &h, params->gamma2))
+  if(polyveck_chknorm(params, &h, params->gamma2)) {
     goto rej;
+  }
```
dkostic, line 215:
```suggestion
/* FIPS 204: line 26 Compute signer's hint */
```

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -270,28 +344,28 @@ int crypto_sign_verify(ml_dsa_params *params,
...
 
-  if(ctxlen > 255 || siglen != params->bytes)
+  if(siglen != params->bytes) {
     return -1;
-
+  }
+  /* FIPS 204: line 1 */
   unpack_pk(params, rho, &t1, pk);
-  if(unpack_sig(params, c, &z, &h, sig))
+  if(unpack_sig(params, c, &z, &h, sig)) {
```
dkostic, line 352:
```suggestion
  /* FIPS 204: line 2 */
  if(unpack_sig(params, c, &z, &h, sig)) {
```

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -270,28 +344,28 @@ int crypto_sign_verify(ml_dsa_params *params,
...
     return -1;
-  if(polyvecl_chknorm(params, &z, params->gamma1 - params->beta))
+  }
+  if(polyvecl_chknorm(params, &z, params->gamma1 - params->beta)) {
     return -1;
+  }
 
-  /* Compute CRH(H(rho, t1), msg) */
+  /* FIPS 204: line 7 Compute CRH(H(rho, t1), msg) */
```
dkostic, line 359: can you please clarify the discrepancy between the spec and the code. the spec says:

6: 𝑡𝑟 ← H(𝑝𝑘, 64)
7: 𝜇 ← (H(BytesToBits(𝑡𝑟)||𝑀 ′, 64))

jakemas: Great call out, added in [dfd1601](https://github.com/aws/aws-lc/pull/1999/commits/dfd1601d19493fedf2602ae2f3c1a2de52173055)

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: [dfd1601](https://github.com/aws/aws-lc/pull/1999/commits/dfd1601d19493fedf2602ae2f3c1a2de52173055)

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: [dfd1601](https://github.com/aws/aws-lc/pull/1999/commits/dfd1601d19493fedf2602ae2f3c1a2de52173055)

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: [dfd1601](https://github.com/aws/aws-lc/pull/1999/commits/dfd1601d19493fedf2602ae2f3c1a2de52173055)

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
jakemas: Great call out added in [dfd1601](https://github.com/aws/aws-lc/pull/1999/commits/dfd1601d19493fedf2602ae2f3c1a2de52173055)

dkostic (dismissed):

dkostic (approved):

andrewhop (approved):

PR #2003, opened 2024-11-19
[EC] Unify scalar_mul_base point for ec_nistp curves

dkostic:
### Issues:
CryptoAlg-2544

### Description of changes: 
Added unified scalar multiplication of the base point for curves
implemented in ec_nistp. This is a refactor of the algorithm in
p384.c and p521.c that makes it generic. The implementation 
in p384.c and p521.c is substituted with this new unified
implementation.

### Call-outs:
Point out areas that need special attention or support during the review process. Discuss architecture or design changes.

### Testing:
How is this change tested (unit tests, fuzz tests, etc.)? Are there any testing steps to be verified by the reviewer?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

PR #2004, opened 2024-11-20
[EC] Unify scalar_mul_public for ec_nistp curves

dkostic:
### Issues:
CryptoAlg-2545

### Description of changes: 
Added unified scalar_mul_public implemented in ec_nistp.
This is a refactor of the algorithm in p384.c and p521.c
that makes it generic. The implementations in p384.c, p521.c,
as well as in p256.c, are substituted with this new unified
implementation.

### Call-outs:
Builds on top of #2003 

### Testing:


crypto/fipsmodule/ec/wnaf.c
```diff
@@ -85,8 +85,17 @@
 //   http://link.springer.com/chapter/10.1007%2F3-540-45537-X_13
 //   http://www.bmoeller.de/pdf/TI-01-08.multiexp.pdf
 
-void ec_compute_wNAF(const EC_GROUP *group, int8_t *out,
-                     const EC_SCALAR *scalar, size_t bits, int w) {
+// Returns the value of the bit at |idx| place in the scalar.
+static int ec_scalar_bit_at_idx(const EC_SCALAR *s, size_t s_bits, size_t idx) {
```
nebeid, line 89: What is the need for this function? Isn't it similar to the function that was used https://github.com/aws/aws-lc/blob/85f58da29e68906d094d378fe968f6b90807d05f/crypto/fipsmodule/bn/shift.c#L261 or is it to inline it?

dkostic: `ec_compute_wNAF` calls `bn_is_bit_set_words` with `group->order.N.width` which is why it needs `group` as an input argument. Since `ec_nistp_scalar_mul_public` doesn't have access to the group and I wanted to be able to reuse `ec_compute_wNAF` there, I had to remove `group` from `ec_compute_wNAF`'s arguments. So I had to add a replacement for `bn_is_bit_set_words` that works with the bit size of the scalar instead of the number of words.

I hope this makes sense :)

nebeid: I guessed this was the reason, but I thought you could call `bn_is_bit_set_words` with the second argument being `bits` as in the current call (the first argument would be `s->words`, I believe), no?

dkostic: no no, the second argument of `bn_is_bit_set_words` is number of words, not bits.

nebeid: Discussed offline. @dkostic reverted back to using `bn_is_bit_set_words` after calculating the number of words to pass to it.

crypto/fipsmodule/ec/ec_nistp.c
```diff
@@ -636,3 +640,132 @@ void ec_nistp_scalar_mul_base(const ec_nistp_meth *ctx,
...
+      } else {
+        // Otherwise, add to the accumulator either the point at position idx
+        // in the table or its negation.
+        const ec_nistp_felem_limb *y_tmp ;
+        y_tmp = &g_table[idx * g_point_num_limbs + ctx->felem_num_limbs];
+        if (is_neg) {
+          ctx->felem_neg(ftmp, &g_table[idx * g_point_num_limbs + ctx->felem_num_limbs]);
+          y_tmp = ftmp;
+        }
+        ctx->point_add(x_out, y_out, z_out, x_out, y_out, z_out, 1,
```
nebeid, line 766: Is the mixed flag explained elsewhere (re removing the comments that were before this step.)

nebeid (approved):
Left a documentation question whether for this PR or an upcoming one.

crypto/fipsmodule/ec/ec_nistp.c
dkostic: Samuel asked the same so I'll add the comment back in my next PR that's going to deal with point addition anyway.

crypto/fipsmodule/ec/ec_nistp.c
```diff
@@ -636,3 +640,132 @@ void ec_nistp_scalar_mul_base(const ec_nistp_meth *ctx,
...
+  int8_t g_wnaf[SCALAR_MUL_MAX_SCALAR_BITS + 1] = {0};
+  ec_compute_wNAF(p_wnaf, p_scalar, ctx->felem_num_bits, SCALAR_MUL_WINDOW_SIZE);
+  ec_compute_wNAF(g_wnaf, g_scalar, ctx->felem_num_bits, SCALAR_MUL_WINDOW_SIZE);
+
+  // In the beginning res is set to point-at-infinity, so we set the flag.
+  int16_t res_is_inf = 1;
+  int16_t d, is_neg, idx;
+  ec_nistp_felem ftmp;
+
+  for (int i = ctx->felem_num_bits; i >= 0; i--) {
```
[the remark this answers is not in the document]
dkostic: yes, `i` has to be a signed integer for the loop to work.

crypto/fipsmodule/ec/ec_nistp.c
```diff
@@ -636,3 +640,132 @@ void ec_nistp_scalar_mul_base(const ec_nistp_meth *ctx,
...
+        res_is_inf = 0;
+      } else {
+        // Otherwise, add to the accumulator either the point at position idx
+        // in the table or its negation.
+        const ec_nistp_felem_limb *y_tmp;
+        y_tmp = &p_table[idx * p_point_num_limbs + ctx->felem_num_limbs];
+        if (is_neg) {
+          ctx->felem_neg(ftmp, y_tmp);
+          y_tmp = ftmp;
+        }
```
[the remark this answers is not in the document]
dkostic: because of line 731 (the else branch is basically moved to this line). I did this to avoid the unnecessary copying in the `!is_neg` case (instead of doing `p384_felem_copy`, we just take a pointer to the proper memory location)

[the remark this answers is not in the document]
samuel40791765: Ahhh that makes perfect sense, thanks.

samuel40791765 (approved):

PR #2005, opened 2024-11-20
Fix perl handling of paths w/ spaces

justsmth:
### Description of changes: 
Build was failing when the source path contained spaces. This is due to there not being quotes around the `$xlate` and `$output` at the start of the script.

### Call-outs:
* Although this change alters a few Perl scripts, no changes are made to the assembly files they produce.

### Testing:
Added a test that sets up the paths for both the source and build directories to contain spaces.

### DIFF
--- a/.github/workflows/misc-tests.yaml
+++ b/.github/workflows/misc-tests.yaml
@@ -83,7 +83,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, windows-latest, macos-13, macos-14-xlarge ]
+        os: [ ubuntu-latest, windows-latest, macos-14-xlarge ]
         fips:
           - "0"
           - "1"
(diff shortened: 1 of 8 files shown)


.github/workflows/misc-tests.yaml
```diff
@@ -77,3 +77,37 @@ jobs:
...
+  path-has-spaces:
+    if: github.repository_owner == 'aws'
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-latest, windows-latest, macos-13, macos-14-xlarge ]
+        fips:
+          - "0"
+          - "1"
```
andrewhop, line 89: Frugality: could we just update an existing CI dimension to add some spaces instead of add all new dimensions?

justsmth: Good point. I did consider this. 
* We currently do not have a test that spans across all of the operating systems.  So we'd need to alter multiple tests, each of which is already targeting a separate concern.
* If we mix this concern with another, how obvious (or not obvious) would the root cause be when a build or test is failing (in part or solely) due to spaces in the source/build/data paths, especially in the presence of other error messages? In many cases this cause might be obvious, but there also might be cases that the error messaging is noisy, missing or misleading. 
* This setup isolates the condition so that if other tests are passing, but not this one, then the cause is clear.

justsmth: I modified the test configuration so that one 3 new jobs are being added (instead of the 8 previously).

skmcgrail (approved):

skmcgrail (approved):

skmcgrail (approved):

andrewhop (approved):

PR #2008, opened 2024-11-22
strdup is not C99

justsmth:
### Issues:
Addresses:
* aws/aws-lc-rs#610

### Description of changes: 
`strdup` was only officially [added to libc in C23](https://en.cppreference.com/w/c/string/byte/strdup). Platforms we support may require the `__STDC_WANT_LIB_EXT2__` or `_POSIX_C_SOURCE` macros to be defined prior to the inclusion of the "string.h" header.

### Call-outs:
* Reference: https://en.cppreference.com/w/c/experimental/dynamic/strdup

### DIFF
--- a/crypto/err/err.c
+++ b/crypto/err/err.c
@@ -106,6 +106,16 @@
  * (eay@cryptsoft.com).  This product includes software written by Tim
  * Hudson (tjh@cryptsoft.com). */
 
+// `strdup` was only officially added in C23:
+// * https://en.cppreference.com/w/c/string/byte/strdup
+// The following macros are needed to ensure its definition
+// is provided by the headers.
+#ifdef __STDC_ALLOC_LIB__
+#define __STDC_WANT_LIB_EXT2__ 1
+#else
+#define _POSIX_C_SOURCE 200809L
+#endif
+
 // Ensure we can't call OPENSSL_malloc circularly.
 #define _BORINGSSL_PROHIBIT_OPENSSL_MALLOC
 #include <openssl/err.h>


crypto/err/err.c
```diff
@@ -106,6 +106,16 @@
  * (eay@cryptsoft.com).  This product includes software written by Tim
  * Hudson (tjh@cryptsoft.com). */
 
+// `strdup` was only officially added in C23:
```
torben-hansen, line 109: Suggest replacing the two occurrences with `OPENSSL_strdup()` instead.

```
$ grep -r " strdup" ./
.//crypto/err/err.c:    // Disable deprecated functions on msvc so it doesn't complain about strdup.
.//crypto/err/err.c:    dst->data = strdup(src->data);
.//crypto/err/err.c:  // Disable deprecated functions on msvc so it doesn't complain about strdup.
.//crypto/err/err.c:  char *copy = strdup(data);
.//include/openssl/mem.h:// OPENSSL_strdup has the same behaviour as strdup(3).
.//third_party/googletest/test/gtest_unittest.cc:    GTEST_DISABLE_MSC_DEPRECATED_PUSH_(/* getenv, strdup: deprecated */)
.//third_party/googletest/test/gtest_unittest.cc:      saved_tz_ = strdup(getenv("TZ"));
.//third_party/googletest/include/gtest/internal/gtest-string.h:  // This is different from strdup() in string.h, which allocates
.//third_party/googletest/include/gtest/internal/gtest-port.h:inline char* StrDup(const char* src) { return strdup(src); }
.//third_party/googletest/include/gtest/internal/gtest-port.h:inline char* StrDup(const char* src) { return strdup(src); }
.//third_party/googletest/include/gtest/internal/gtest-port.h:inline char* StrDup(const char* src) { return strdup(src); }
```

torben-hansen (approved):

skmcgrail (approved):

PR #2020, opened 2024-11-27
Only abort when RSA PWCT fail in FIPS

samuel40791765:
### Description of changes: 
We've been getting abort failures when building with FIPS mode against [Ruby's break tests](https://github.com/ruby/ruby/blob/d5de1a57893b16aff7bc3336b34fa2e9acefb3d2/test/openssl/test_pkey_rsa.rb#L80-L85). The issue happens to be related to the abort call we do when calls to `RSA_generate_key_ex` fail.
In the original commit where this was introduced (https://github.com/aws/aws-lc/commit/6bdd4c3403a17f85cf8cee0a245f220c5c0c28e5), it's mentioned that "It's required that the FIPS module aborts when PCT tests fail in `RSA_check_fips()`." Our current behavior fails regardless of a regular RSA failure or a PWCT failure, which causes regular RSA failures to unintentionally abort as well.
This changes aborting to only happen during failures in `RSA_check_fips`. Our existing death tests were also expecting failures during regular RSA failures rather than PWCT failures, so I've tweaked the tests to account for that. Ruby's RSA break test passes successfully with this change.

### Call-outs:
The new test is ran when the CFLAG `BORINGSSL_FIPS_BREAK_TESTS` is set, so I've updated our test script to run tests when building with this dimension. This leverages our existing break tests to test PWCT aborting behavior.

### Testing:
New test dimension

### DIFF
--- a/crypto/rsa_extra/rsa_test.cc
+++ b/crypto/rsa_extra/rsa_test.cc
@@ -695,8 +695,6 @@ TEST(RSATest, BadExponent) {
   ERR_clear_error();
 }
 
-#if !defined(AWSLC_FIPS)
-
 // Attempting to generate an excessively small key should fail.
 TEST(RSATest, GenerateSmallKey) {
   bssl::UniquePtr<RSA> rsa(RSA_new());
@@ -709,25 +707,6 @@ TEST(RSATest, GenerateSmallKey) {
   EXPECT_TRUE(
       ErrorEquals(ERR_get_error(), ERR_LIB_RSA, RSA_R_KEY_SIZE_TOO_SMALL));
 }
-
-#else
-// AWSLCAndroidTestRunner does not take tests that do |ASSERT_DEATH| very well.
-// GTEST issue: https://github.com/google/googletest/issues/1496.
-#if !defined(OPENSSL_ANDROID)
-
-// Attempting to generate an excessively small key should fail.
-// In the case of a FIPS build, expect abort() when |RSA_generate_key_ex| fails.
-TEST(RSADeathTest, GenerateSmallKeyAndDie) {
-  bssl::UniquePtr<RSA> rsa(RSA_new());
-  ASSERT_TRUE(rsa);
-  bssl::UniquePtr<BIGNUM> e(BN_new());
-  ASSERT_TRUE(e);
-  ASSERT_TRUE(BN_set_word(e.get(), RSA_F4));
-
-  ASSERT_DEATH_IF_SUPPORTED(RSA_generate_key_ex(rsa.get(), 255, e.get(), nullptr), "");
-}
-#endif
-#endif
 
 // Attempting to generate an funny RSA key length should round down.
 TEST(RSATest, RoundKeyLengths) {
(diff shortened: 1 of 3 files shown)


crypto/rsa_extra/rsa_test.cc
```diff
@@ -1214,70 +1193,33 @@ TEST(RSATest, KeygenFailOnce) {
...
+  return SetEnvironmentVariableA(name.c_str(), value.c_str()) != 0;
+#else
+  return setenv(name.c_str(), value.c_str(), 1) == 0;
+#endif
+}
 
-  ASSERT_DEATH_IF_SUPPORTED(RSA_generate_key_ex(rsa.get(), 2048, e.get(), &cb),"");
+// In the case of a FIPS build, expect abort() when |RSA_generate_key_fips|
+// fails.
+TEST(RSADeathTest, KeygenFailAndDie) {
```
smittals2, line 1211: What about this test would fail, why is this a valuable test? I think this is causing a bunch of CI failures too. 

Previously was the BN callback the reason the keygen would fail? What about the test setup indicates the PCT would fail if you are using supported key lengths and doing normal keygen?

samuel40791765: CI failures are on me, I think the environment variables aren’t being set properly in the CI environment. I'll change the env variable to be set in the script and skip the test if it's not defined.

> Why is this a valuable test?

FIPS has requirements for Pairwise Consistency Tests to be implemented in `RSA_check_fips`. If anything goes wrong in these specific consistency tests, the library should be aborted. This test was checking if the library was aborted correctly. Originally we were aborting for **any general** RSA key generation failure, this changes the behavior to abort only when the PCT tests fail.

> Previously was the BN callback the reason the keygen would fail? What about the test setup indicates the PCT would fail if you are using supported key lengths and doing normal keygen?

Yeah the callback was forcing a general failure in keygen. This was unrelated to PCT however.
In normal circumstances, the PCT can't fail unless something goes horribly wrong with the module. Testing that the module properly aborts is required for FIPS demonstrations, which is why we have existing logic to force RSA PCT failures.
https://github.com/aws/aws-lc/blob/59828538a790094113eacd5dd23d01be2885b36a/crypto/fipsmodule/rsa/rsa.c#L1453-L1455

The new test set up leverages the existing logic, so we properly test for PCT aborting failures.

samuel40791765:
```
+ export BORINGSSL_FIPS_BREAK_TEST=RSA_PWCT
+ BORINGSSL_FIPS_BREAK_TEST=RSA_PWCT
+ /codebuild/output/src2176229394/src/github.com/aws/aws-lc/test_build_dir/crypto/crypto_test --gtest_filter=RSADeathTest.KeygenFailAndDie
Note: Google Test filter = RSADeathTest.KeygenFailAndDie
[----------] 1 test from RSADeathTest
[ RUN      ] RSADeathTest.KeygenFailAndDie
[       OK ] RSADeathTest.KeygenFailAndDie (2080 ms)
[----------] 1 test from RSADeathTest (2080 ms total)

[==========] 1 test from 1 test suite ran. (2080 ms total)
[  PASSED  ] 1 test.
+ unset BORINGSSL_FIPS_BREAK_TEST
```

crypto/rsa_extra/rsa_test.cc
smittals2: Ah I see that makes sense. That last link is helpful - I was wondering how the failure was being forced. Thanks!

smittals2 (approved):

nebeid (approved):

PR #2021, opened 2024-11-29
Expose BN_set_flags as a no-op

samuel40791765:
### Issues:
Addresses `CryptoAlg-1477`

### Description of changes: 
We're currently leaning towards continuing to not support `BN_FLG_CONSTTIME`. It's a problematic flag that's poorly documented and could lead to constant time violations if we aren't careful.
We can expose `BN_set_flags` as a no-op however. Ruby has logic to get around the `BN_FLG_CONSTTIME` flag, but not around `BN_set/get_flags`. Exposing the set symbol would allow for a cleaner patch/build.

### Call-outs:
N/A

### Testing:
N/A

### DIFF
--- a/include/openssl/bn.h
+++ b/include/openssl/bn.h
@@ -1021,6 +1021,11 @@ OPENSSL_EXPORT int BN_mod_exp_mont_consttime_x2(BIGNUM *rr1, const BIGNUM *a1, c
                                                 const BIGNUM *m2, const BN_MONT_CTX *in_mont2,
                                                 BN_CTX *ctx);
 
+// BN_set_flags does nothing. See comments regarding |BN_FLG_CONSTTIME| being
+// intentionally omitted for more details.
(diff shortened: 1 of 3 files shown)


include/openssl/bn.h
```diff
@@ -1066,7 +1071,6 @@ OPENSSL_EXPORT unsigned BN_num_bits_word(BN_ULONG l);
 #define BN_FLG_STATIC_DATA 0x02
 
 #ifdef AWS_LC_INTERNAL_IGNORE_BN_SET_FLAGS
```
darylmartin100, line 1073: We should also delete this #ifdef as we can just add a -DBN_FLG_CONSTTIME=0x04 at build time OR properly #ifdef the upstream packages (OpenSSH comes to mind).

samuel40791765: Agree 100%

torben-hansen (approved):

WillChilds-Klein (approved):

PR #2029, opened 2024-12-03
Support Armv8.5 RNDR as prediction resistance

torben-hansen:
### Description of changes: 

Adds Arm as a hardware source for prediction resistance, if supported. This covers Linux+Armv8.5-A that impements `rndr`. For exmaple, this should be enabled on Graviton3/Graviton4+Amazon Linux.

Note, only multiples of 8 are supported. That's what we need now and I don't see that need changing.

### Call-outs:

Wrap the libcrypto inlined functions to workaround the Arm capability vector having internal linkage.

Apple doesn't implement `FEAT_RNG` on any of their M's. Might be able to implement for Windows. But I don't see a way to probe for the instruction here https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-isprocessorfeaturepresent.

Added `s3_3_c2_c4_0` as a known symbol to the delocator to avoid it from interpreting the string as an external symbol that needs a trampoline... It's not really a symbol... And this was the quickest way to move forward.

### Testing:

Tested that the function is executed successfully on a Graviton3 instance. This might also be executed through codebuild depending on which instances are spawned.

### DIFF
--- a/crypto/fipsmodule/CMakeLists.txt
+++ b/crypto/fipsmodule/CMakeLists.txt
@@ -100,6 +100,7 @@ if(ARCH STREQUAL "aarch64")
     md5-armv8.${ASM_EXT}
     p256-armv8-asm.${ASM_EXT}
     p256_beeu-armv8-asm.${ASM_EXT}
+    rndr-armv8.${ASM_EXT}
     sha1-armv8.${ASM_EXT}
     sha256-armv8.${ASM_EXT}
     sha512-armv8.${ASM_EXT}
@@ -149,6 +150,7 @@ if(PERL_EXECUTABLE)
   perlasm(p256-armv8-asm.${ASM_EXT} ec/asm/p256-armv8-asm.pl)
   perlasm(p256_beeu-armv8-asm.${ASM_EXT} ec/asm/p256_beeu-armv8-asm.pl)
   perlasm(rdrand-x86_64.${ASM_EXT} rand/asm/rdrand-x86_64.pl)
+  perlasm(rndr-armv8.${ASM_EXT} rand/asm/rndr-armv8.pl)
   perlasm(rsaz-avx2.${ASM_EXT} bn/asm/rsaz-avx2.pl)
   perlasm(rsaz-2k-avx512.${ASM_EXT} bn/asm/rsaz-2k-avx512.pl)
   perlasm(rsaz-3k-avx512.${ASM_EXT} bn/asm/rsaz-3k-avx512.pl)
--- a/crypto/fipsmodule/rand/asm/rndr-armv8.pl
+++ b/crypto/fipsmodule/rand/asm/rndr-armv8.pl
@@ -22,8 +22,6 @@ die "can't locate arm-xlate.pl";
 open OUT,"| \"$^X\" $xlate $flavour $output";
 *STDOUT=*OUT;
 
-my $rndr_reg = "s3_3_c2_c4_0";
-
 $code.=<<___;
 #include <openssl/arm_arch.h>
 
@@ -35,23 +33,25 @@ $code.=<<___;
 .type CRYPTO_rndr,%function
 .align 4
 CRYPTO_rndr:
-  mov x2, #0
+  cbz x1, .Lrndr_error        // out_len = 0 is not supported
+  mov x4, x1                  // out_len: requested number of bytes
+  mov x2, #0                  // Counts number of bytes generated
 
 .Lrndr_loop:
-  cbz x1, .Lrndr_done // out_len == 0?
+  mrs x3, s3_3_c2_c4_0        // rndr instruction
+  cbz x3, .Lrndr_error        // Check if RNDR failed
 
-  mrs x3, $rndr_reg
-  cbz x3, .Lrndr_done // Check of RNDR failed
-
-  cmp x1, #8 // Sets N if strictly less than 8 bytes left
+  cmp x1, #8                  // Sets N if strictly less than 8 bytes left
   blt .Lrndr_less_than_8_bytes
 
-  str x3, [x0], #8 // Copy 8 bytes to *out and increment pointer by 8
-  add x2, x2, #8 // Adds 8 to return value
+  str x3, [x0], #8            // Copy 8 bytes to *out and increment pointer by 8
+  add x2, x2, #8              // Adds 8 to counter
   sub x1, x1, #8
+  cbz x1, .Lrndr_done         // If multiple of 8 this will be 0 eventually
   b .Lrndr_loop
 
 .Lrndr_less_than_8_bytes:
+  // Copy remaining bytes one by one
   strb  w3, [x0]
   lsr x3, x3, #8
   add x2, x2, #1
@@ -60,7 +60,13 @@ CRYPTO_rndr:
   cbnz x1, .Lrndr_less_than_8_bytes
 
 .Lrndr_done:
-  mov x0, x2 // Return value
+  cmp x2, x4                  // Ensure correct number of bytes were generated
+  bne .Lrndr_error
+  mov x0, #1                 // Return value success
+  ret
+
+.Lrndr_error:
+  mov x0, #0                  // Return value error
   ret
 .size CRYPTO_rndr,.-CRYPTO_rndr
 ___
--- a/crypto/fipsmodule/rand/asm/rndr-armv8.pl
+++ b/crypto/fipsmodule/rand/asm/rndr-armv8.pl
@@ -26,7 +26,6 @@ my $rndr_reg = "s3_3_c2_c4_0";
 
 $code.=<<___;
 #include <openssl/arm_arch.h>
-#if __ARM_MAX_ARCH__ >= 8
 
 .arch armv8-a
 .text
@@ -64,7 +63,6 @@ CRYPTO_rndr:
   mov x0, x2 // Return value
   ret
 .size CRYPTO_rndr,.-CRYPTO_rndr
-#endif
 ___
 
 print $code;
--- a/crypto/fipsmodule/rand/asm/rndr-armv8.pl
+++ b/crypto/fipsmodule/rand/asm/rndr-armv8.pl
@@ -62,7 +62,7 @@ CRYPTO_rndr:
 .Lrndr_done:
   cmp x2, x4                  // Ensure correct number of bytes were generated
   bne .Lrndr_error
-  mov x0, #1                 // Return value success
+  mov x0, #1                  // Return value success
   ret
 
 .Lrndr_error:
--- a/crypto/fipsmodule/rand/asm/rndr-armv8.pl
+++ b/crypto/fipsmodule/rand/asm/rndr-armv8.pl
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0 OR ISC
 
 # RNDR from ARMv8.5-A.
-# System register encoding: s3_3_c2_c4_0
+# System register encoding: s3_3_c2_c4_0.
 # see https://developer.arm.com/documentation/ddi0601/2024-09/AArch64-Registers/RNDR--Random-Number
 
 # The first two arguments should always be the flavour and output file path.
@@ -38,11 +38,11 @@ CRYPTO_rndr:
   mov x2, #0                  // Counts number of bytes generated
 
 .Lrndr_loop:
-  mrs x3, s3_3_c2_c4_0        // rndr instruction
-  cbz x3, .Lrndr_error        // Check if RNDR failed
+  mrs x3, s3_3_c2_c4_0        // rndr instruction https://developer.arm.com/documentation/ddi0601/2024-09/Index-by-Encoding
+  cbz x3, .Lrndr_error        // Check if rndr failed
 
-  cmp x1, #8                  // Sets N if strictly less than 8 bytes left
-  blt .Lrndr_less_than_8_bytes
+  cmp x1, #8                  // If strictly less than 8, does not set condition flag C
+  blo .Lrndr_less_than_8_bytes
 
   str x3, [x0], #8            // Copy 8 bytes to *out and increment pointer by 8
   add x2, x2, #8              // Adds 8 to counter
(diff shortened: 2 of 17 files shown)


crypto/fipsmodule/rand/asm/rndr-armv8.pl
```diff
@@ -0,0 +1,75 @@
...
+.text
+
+# int CRYPTO_rndr(uint8_t *out, const size_t len)
+.globl CRYPTO_rndr
+.type CRYPTO_rndr,%function
+.align 4
+CRYPTO_rndr:
+  cbz x1, .Lrndr_error        // len = 0 is not supported
+  mov x4, x1                  // len: requested number of bytes
+  mov x2, #0                  // Counts number of bytes generated
```
dkostic, line 38: do you need this? I think you can update `x4` until it becomes 0?

crypto/fipsmodule/rand/asm/rndr-armv8.pl
```diff
@@ -0,0 +1,75 @@
...
+
+.Lrndr_loop:
+  mrs x3, s3_3_c2_c4_0        // rndr instruction
```
dkostic, line 41: can you please add a reference for this?

crypto/fipsmodule/rand/asm/rndr-armv8.pl
```diff
@@ -0,0 +1,75 @@
...
+  cbz x3, .Lrndr_error        // Check if RNDR failed
```
dkostic, line 42: I'm a bit surprised it can fail.

crypto/fipsmodule/rand/asm/rndr-armv8.pl
```diff
@@ -0,0 +1,75 @@
...
+
+  cmp x1, #8                  // Sets N if strictly less than 8 bytes left
```
dkostic, line 44: what's `N`?

crypto/fipsmodule/rand/asm/rndr-armv8.pl
```diff
@@ -0,0 +1,75 @@
...
+  blt .Lrndr_less_than_8_bytes
+
+  str x3, [x0], #8            // Copy 8 bytes to *out and increment pointer by 8
+  add x2, x2, #8              // Adds 8 to counter
+  sub x1, x1, #8
+  cbz x1, .Lrndr_done         // If multiple of 8 this will be 0 eventually
+  b .Lrndr_loop
+
+.Lrndr_less_than_8_bytes:
```
dkostic, line 53: I think it'd be easier if you handled this part in C. In other words, implement `CRYPTO_rndr` to be a single call to `rndr` that returns 8 bytes and then use this function in C code?

dkostic: actually, I see that the x86 implementation does something similar:
- https://github.com/aws/aws-lc/blob/905e9d037561bd54be19b1aff5a6c46b23f4ec35/crypto/fipsmodule/rand/asm/rdrand-x86_64.pl#L74
- https://github.com/aws/aws-lc/blob/905e9d037561bd54be19b1aff5a6c46b23f4ec35/crypto/fipsmodule/rand/rand.c#L304

can we do the same for Arm and also reuse the C code?

crypto/fipsmodule/rand/asm/rndr-armv8.pl
torben-hansen: Condition flags. Input are to be interpreted as unsigned though, so should be checking for `C` being set and use `blo`. The case of `C` not being set is the case where `x1 < 8` and we need to handle bit-by-bit.
I changed that.

crypto/fipsmodule/rand/asm/rndr-armv8.pl
torben-hansen: The `rdrand` is handled that way. I didn't do that on purpose to avoid having to do extra buffer allocation and copying.
Maybe easier, but the assembly code is not complicated.

torben-hansen: If you insist, I would rather just make it 8-byte multiple only support.

crypto/fipsmodule/rand/asm/rndr-armv8.pl
```diff
@@ -0,0 +1,75 @@
...
+my $output  = shift;
+
+$0 =~ m/(.*[\/\\])[^\/\\]+$/; $dir=$1;
+( $xlate="${dir}arm-xlate.pl" and -f $xlate ) or
+( $xlate="${dir}../../../perlasm/arm-xlate.pl" and -f $xlate) or
+die "can't locate arm-xlate.pl";
+
+open OUT,"| \"$^X\" $xlate $flavour $output";
+*STDOUT=*OUT;
+
```
dkostic, line 24: it'd be easier to follow the code if you assigned variable names to the registers you are using. For example:
```
my ($out, $len, $cnt, $res) = ("x0", "x1", "x2", "x3")
```

crypto/fipsmodule/rand/asm/rndr-armv8.pl
dkostic: I know but it seems superfluous to me, couldn't you just check `x1 == 0` in the end? that way you could get rid of both x2 and x4

crypto/fipsmodule/rand/asm/rndr-armv8.pl
dkostic: fair enough

crypto/fipsmodule/CMakeLists.txt
```diff
@@ -149,6 +150,7 @@ if(PERL_EXECUTABLE)
   perlasm(p256-armv8-asm.${ASM_EXT} ec/asm/p256-armv8-asm.pl)
   perlasm(p256_beeu-armv8-asm.${ASM_EXT} ec/asm/p256_beeu-armv8-asm.pl)
   perlasm(rdrand-x86_64.${ASM_EXT} rand/asm/rdrand-x86_64.pl)
+  perlasm(rndr-armv8.${ASM_EXT} rand/asm/rndr-armv8.pl)
```
darylmartin100, line 153: Why not use intrinsics?  We could use int __rndr (uint64_t *) as defined here (https://developer.arm.com/documentation/101028/0012/8--Data-processing-intrinsics)

dkostic: intrinsic support is compiler dependent, it's possible that it's not fully portable

dkostic (dismissed):

util/fipstools/delocate/delocate.go
```diff
@@ -2418,7 +2418,8 @@ func localEntryName(name string) string {
 func isSynthesized(symbol string, processor processorType) bool {
 	SymbolisSynthesized := strings.HasSuffix(symbol, "_bss_get") ||
 		symbol == "OPENSSL_ia32cap_get" ||
-		symbol == "BORINGSSL_bcm_text_hash"
+		symbol == "BORINGSSL_bcm_text_hash" ||
```
dkostic, line 2421: what I don't quite get is that I removed this line `symbol == "BORINGSSL_bcm_text_hash" ||` in https://github.com/aws/aws-lc/commit/c21a05c3ea120ea0f3666fe07ee92c1dd3f73dc6#diff-3a7ed42c3edacaf0a2c43110d7da1759e6386e9ba0734df5d078475b46e47105L2421 which is already merged in, so it's not present in main (https://github.com/aws/aws-lc/blob/850af98235851dca40acae72537a4d8f97265606/util/fipstools/delocate/delocate.go#L2428). And GH doesn't report a merge conflict but rather adds the line back.

crypto/fipsmodule/rand/asm/rndr-armv8.pl
```diff
@@ -0,0 +1,60 @@
...
+.arch armv8-a
+.text
+
+# int CRYPTO_rndr_multiple8(uint8_t *out, const size_t len)
+.globl CRYPTO_rndr_multiple8
+.type CRYPTO_rndr_multiple8,%function
+.align 4
+CRYPTO_rndr_multiple8:
+  cbz $len, .Lrndr_multiple8_error  // len = 0 is not supported
```
nebeid, line 38: you may want to xor $rndr with itself at the beginning to zero it.

crypto/fipsmodule/rand/asm/rndr-armv8.pl
```diff
@@ -0,0 +1,60 @@
...
+
+.Lrndr_multiple8_loop:
+  mrs $rndr64, s3_3_c2_c4_0             // rndr instruction https://developer.arm.com/documentation/ddi0601/2024-09/Index-by-Encoding
```
nebeid, line 41: Is this a better link https://developer.arm.com/documentation/ddi0601/2024-09/AArch64-Registers/RNDR--Random-Number? It's the same as the one you have above

crypto/fipsmodule/rand/asm/rndr-armv8.pl
```diff
@@ -0,0 +1,60 @@
...
+
+my ($out, $len, $rndr64) = ("x0", "x1", "x2");
+
+$code.=<<___;
+#include <openssl/arm_arch.h>
+
+.arch armv8-a
+.text
+
+# int CRYPTO_rndr_multiple8(uint8_t *out, const size_t len)
```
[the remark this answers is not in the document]
torben-hansen: That documentation is in the header file. Will try to keep that in one single location.

crypto/fipsmodule/rand/entropy/internal.h
```diff
@@ -40,6 +41,39 @@ OPENSSL_EXPORT void tree_jitter_free_thread_drbg(struct entropy_source_t *entrop
 OPENSSL_EXPORT int tree_jitter_get_seed(
   const struct entropy_source_t *entropy_source, uint8_t seed[CTR_DRBG_ENTROPY_LEN]);
 
+// rndr_multiple8 writes |len| number of bytes to |buf| generated using the
```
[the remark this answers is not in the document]
torben-hansen: I'm gonna move the rdrand stuff later on :)

crypto/fipsmodule/rand/urandom_test.cc
```diff
@@ -495,8 +495,8 @@ static std::vector<Event> TestFunctionPRNGModel(unsigned flags) {
   const size_t kAdditionalDataLength = 32;
   const size_t kPersonalizationStringLength = CTR_DRBG_ENTROPY_LEN;
   const size_t kPassiveEntropyWithWhitenFactor = PASSIVE_ENTROPY_LOAD_LENGTH;
-  const bool kHaveRdrand = have_rdrand();
-  const bool kHaveFastRdrand = have_fast_rdrand();
+  const bool kHaveRdrand = have_hw_rng_x86_64();
```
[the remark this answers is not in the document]
torben-hansen: yeah absolutely. But right now the model test targets the old implementation. I will port them to the new randomness generation implementation in another PR.

crypto/fipsmodule/rand/asm/rndr-armv8.pl
torben-hansen: Won't `mrs $rndr64, s3_3_c2_c4_0` fill the entire register?

crypto/fipsmodule/rand/asm/rndr-armv8.pl
torben-hansen: I added https://developer.arm.com/documentation/ddi0601/2024-09/Index-by-Encoding because Dusan in a previous comment wanted a link to the encoding. So, I linked the rndr instruction above, as you noted, instead.

crypto/fipsmodule/rand/entropy/entropy_source_tests.cc
```diff
@@ -0,0 +1,40 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include <gtest/gtest.h>
+
+#include "internal.h"
+
+#define MAX_MULTIPLE_FROM_RNG (16)
```
nebeid, line 8: What does the name mean? Is it number of iterations on calling the RNG?

crypto/fipsmodule/rand/entropy/entropy_source_tests.cc
```diff
@@ -0,0 +1,40 @@
...
+  if (have_hw_rng_aarch64_for_testing() != 1) {
+    GTEST_SKIP() << "Compiled for Arm64, but Aarch64 hw rng is not available in run-time";
+  }
+
+  // Extracting 0 bytes is never supported.
+  ASSERT_FALSE(rndr_multiple8(buf, 0));
+
+  // Multiples of 8 allowed.
+  for (size_t i = 8; i < MAX_MULTIPLE_FROM_RNG; i += 8) {
+    ASSERT_TRUE(rndr_multiple8(buf, i));
```
nebeid, line 32: Would you like to do a simple check that it's a new value every time to make sure it's not stuck? Or there are other tests in the code doing that whenever it's available?

crypto/fipsmodule/rand/asm/rndr-armv8.pl
nebeid: Yes, you're right.

crypto/fipsmodule/rand/entropy/entropy_source_tests.cc
torben-hansen: It's the bound `MAX_MULTIPLE_FROM_RNG * 8`.
Naming is hard.

crypto/fipsmodule/rand/entropy/entropy_source_tests.cc
torben-hansen: The next part of the new randomness generation implementation is to port over existing tests and also improve testing. I don't think there is a test that checks whether the entropy sources can be stuck on any values.
But I will add some basic tests to the existing test suite that covers the obvious "okay, this is quite broken"-case.

nebeid (approved):

dkostic (approved):

PR #2033, opened 2024-12-05
[EC] ec_nistp P-256 C scalar_mul_{base|public}

dkostic:
### Issues:
N/A

### Description of changes: 
Use `ec_nistp` implementation for P-256 C code.
Update `make_tables.go` to generate the base point
table in the same way as for P-384 and P-521.

### Call-outs:

### Testing:


crypto/fipsmodule/ec/ec_nistp.c
```diff
@@ -17,9 +17,9 @@
 // |----------------------------|
```
torben-hansen, line 17: Can remove this tracking table now?
Can be follow-up.

torben-hansen (approved):

crypto/fipsmodule/ec/ec_nistp.c
dkostic: will do, thanks!

PR #2038, opened 2024-12-06
Use SHA256 as default digest for OCSP signing

samuel40791765:
### Description of changes: 
OpenSSL apparently sets a default digest when consuming the `EVP_DigestSign*` APIs. We avoid this and emit [a relevant error](https://github.com/aws/aws-lc/blob/8226a0565dd81aab1c3eaf4e623301b7ce2c4423/crypto/fipsmodule/evp/digestsign.c#L144) when this happens. Ruby 3.2's double downed on this on their consumption of OCSP however: https://github.com/ruby/ruby/commit/4d6a29320dcd758c21944983da28b54037c200da We don't want to introduce a "default digest" to all calls to `EVP_DigestSign*`, so I've chosen to imitate OpenSSL's behavior in only the places necessary in this case (OCSP). This fixes the Ruby 3.2 OCSP tests we were failing before. 

### Call-outs:
N/A

### Testing:
Built upon existing OCSP tests


crypto/ocsp/ocsp_server.c
```diff
@@ -172,20 +172,37 @@ static int OCSP_basic_sign_ctx(OCSP_BASICRESP *resp, X509 *signer,
   return 1;
 }
 
+const EVP_MD *OCSP_get_default_digest(const EVP_MD *dgst, EVP_PKEY *signer) {
+  if (dgst != NULL) {
+    return dgst;
+  }
```
justsmth, line 178: NP: Why have this function take `dgst` as a parameter? It would seem more natural to have the callers conditionally invoke `OCSP_get_default_digest` only when `dgst` is NULL.

justsmth (approved):

WillChilds-Klein (approved):

crypto/ocsp/ocsp_server.c
samuel40791765: Agreed, that was the original way I had implemented it. But we would need double null checks (outlined below) in both functions without the abstraction into `OCSP_get_default_digest`. It didn't sit quite well with me since we were already trying to abstract out shared logic, which was why I moved it into the static function.
```
const EVP_MD *init_dgst = dgst;
if (init_dgst != NULL) {
    init_dgst = OCSP_get_default_digest(key);
}
if (init_dgst != NULL) {
  OPENSSL_PUT_ERROR(OCSP, EVP_R_NO_DEFAULT_DIGEST);
   goto err;
}
```

PR #2048, opened 2024-12-09
Bring in testing changes from upstream commit 5ee4e95

WillChilds-Klein:
### Issues:
Resolves i/CryptoAlg-2805

### Description of changes: 

Upstream commit [`5ee4e95`][1]'s source changes are almost identical to our
addition of `BIO_FP_TEXT` in [PR 1153][2]. The only difference in behavior is
that we will call `_setmode` w/ `_O_BINARY` on windows when no flags are
specified, where BoringSSL will simply no-op and leave the underlying file's
mode alone.

Upstream's commit has some nice tests, though, so we crib those and change a
single line of expectation to account for our difference in behavior.

[1]: https://github.com/google/boringssl/commit/5ee4e9512e9a99f97c4a3fad397034028b3457c2
[2]: https://github.com/aws/aws-lc/pull/1153
[3]: https://github.com/openssl/openssl/blob/dc10ffc2834e0d2f5ebc1c3e29bd97f1f43a0ead/crypto/bio/bss_file.c#L235-L237

### Open Question to Reviewers

In the interests of maximizing compatibility, I initially retained
our/[OpenSSL's behavior][3] of setting `_O_BINARY` mode on the file **when
`BIO_FP_TEXT` is not set**. It's possible that downstream consumers have worked
around this behavior and have come to expect it. BoringSSL's behavior better
honors the (likely) intent of the caller.

Should we conform to BoringSSL's behavior and (marginally) risk downstream
compatibility issues or stick with our/OpenSSL's current behavior? 

### Testing:
- CI

---

### DIFF
--- a/crypto/bio/bio_test.cc
+++ b/crypto/bio/bio_test.cc
@@ -50,6 +50,8 @@ using Socket = int;
 #define INVALID_SOCKET (-1)
 static int closesocket(int sock) { return close(sock); }
 static std::string LastSocketError() { return strerror(errno); }
+static const int kOpenReadOnlyBinary = O_RDONLY;
+static const int kOpenReadOnlyText = O_RDONLY;
 #else
 using Socket = SOCKET;
 static std::string LastSocketError() {
@@ -57,6 +59,8 @@ static std::string LastSocketError() {
   snprintf(buf, sizeof(buf), "%d", WSAGetLastError());
   return buf;
 }
+static const int kOpenReadOnlyBinary = _O_RDONLY | _O_BINARY;
+static const int kOpenReadOnlyText = O_RDONLY | _O_TEXT;
 #endif
 
 class OwnedSocket {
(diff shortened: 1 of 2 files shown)


crypto/bio/bio_test.cc
```diff
@@ -832,21 +836,86 @@ TEST(BIOTest, ExternalData) {
...
+  // |BIO_new_file| does not set |BIO_FP_TEXT|, so expect it to set the file's
+  // mode to binary in all cases. This odd behavior, but it's what OpenSSL does.
+  bio.reset(BIO_new_file(temp.path().c_str(), "rb"));
+  ASSERT_TRUE(bio);
+  expect_binary_mode(bio.get());
+  bio.reset(BIO_new_file(temp.path().c_str(), "r"));
+  ASSERT_TRUE(bio);
+  // NOTE: Our behavior here aligns with OpenSSL. BoringSSL would
+  // |expect_text_mode| below because they don't call |_setmode| unless
+  // |BIO_FP_TEXT| is set.
```
nebeid, line 879: If BoringSSL does not call |_setmode| unless |BIO_FP_TEXT| is set, why does it |expect_text_mode|? Is it because it's the default? I'm a bit confused about the logic difference since |BIO_FP_TEXT| is not set. Does Boring's `BIO_new_file` set BIO_FP_TEXT when the flag is "r"?

WillChilds-Klein: > Is it because it's the default?

i believe is the default file mode in Windows `fopen`, yes. please see [here](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170#generic-text-routine-mappings):

> If `t` or `b` isn't given in mode, the default translation mode is defined by the global variable [`_fmode`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/fmode?view=msvc-170).

and [here](https://learn.microsoft.com/en-us/cpp/c-runtime-library/fmode?view=msvc-170):

> The default setting of `_fmode` is `_O_TEXT` for text-mode translation.

> Does Boring's BIO_new_file set BIO_FP_TEXT when the flag is "r"?

[no](https://github.com/google/boringssl/blob/8e07cc4d884ccd2381938e309cd5c86a0c1b8cc8/crypto/bio/file.cc#L121).

`BIO_new_file` will [call `fopen`](https://github.com/google/boringssl/blob/8e07cc4d884ccd2381938e309cd5c86a0c1b8cc8/crypto/bio/file.cc#L108C10-L108C28) with whatever `mode` string is passed into `BIO_new_file`. so if the caller passes in `BIO_new_file("/some/file", "r")`, `"r"` will be interpreted natively by the underlying `fopen` implementation and the file will be opened in text mode. 

Boring will respect whatever mode the underlying file was opened with, whereas we/OpenSSL will always set it to `_O_BINARY` if `BIO_FP_TEXT` is not set in the call to `BIO_new_fp`.

This is the correctness/compatibility tradeoff described in the PR overview.

nebeid: Thank you, Will. I would suggest to clarify the comment as follows (or modify it if it's confusing)
```suggestion
  // NOTE: Our behavior here aligns with OpenSSL which is to |_setmode| the file
  // to binary. BoringSSL would |expect_text_mode| below because it respects 
  // default mode on Windows which is text and doesn't call |_setmode| (unless
  // |BIO_FP_TEXT| is set, which is not the case here).
```

nebeid (commented):
I'm leaning towards being compatible with OpenSSL as it's also not a change in our current behaviour, so it's less risky.

crypto/bio/bio_test.cc
WillChilds-Klein: yes, that's more clear. thank you.

nebeid (approved):

andrewhop (commented):
Initial impression everything looks good in the test. I'm waiting to hear back from any Windows customers before picking a side for default behavior.

andrewhop (approved):

PR #2051, opened 2024-12-10
Add fuzz testing for PKCS7_verify

justsmth:
### Description of changes: 
* Add fuzz testing around `PKCS7_verify`

### Call-outs:
* The only human-readable changes are in the first commit: https://github.com/aws/aws-lc/commit/bb68163dcf74012e9f91c2241c72daea84dcd540


crypto/pkcs7/pkcs7.c
```diff
@@ -1529,6 +1529,9 @@ static int pkcs7_signature_verify(BIO *in_bio, PKCS7 *p7, PKCS7_SIGNER_INFO *si,
   ret = 1;
 
 out:
+#if defined(BORINGSSL_UNSAFE_FUZZER_MODE)
```
WillChilds-Klein, line 1532: here and elsewhere, why does this always need to succeed when fuzzing? suggest adding a comment explaining.

fuzz/pkcs7_verify.cc
```diff
@@ -0,0 +1,96 @@
...
+  PKCS7_verify(pkcs7, certs, store, data_bio, nullptr, 0);
+
+end:
+  BIO_free(data_bio);
+  BIO_free(bio_cert);
+  X509_free(my_cert);
+  X509_STORE_free(store);
+  sk_X509_free(certs);
+  PKCS7_free(pkcs7);
+  return 0;
```
samuel40791765, line 94: Same comment regarding unique pointers from the other PR.

crypto/pkcs7/pkcs7.c
samuel40791765: Same question here, I was under the impression that functions didn't necessarily have to succeed during fuzz testing? We just have to be sure that we're handling all cases properly and that there isn't any undefined behavior.
`fuzz/read_pem.cc` might be a good example of us testing against `PEM_read_bio` without the extra knob.

justsmth: We want `pkcs7_signature_verify` to succeed here during fuzz testing so that the for loop in `PKCS7_verify` completes all of its iterations.  I'll add a comment relating to that.

fuzz/pkcs7_verify.cc
justsmth: Yeah. As mentioned in the other PR, the reason I'd prefer not to use smart pointers, and instead have explicit calls to the "xxx_free" functions, is to make it easier for us to debug/diagnose were a memory leak ever to surface. (Fuzz tests should always fail if there is a memory leak.)

WillChilds-Klein (dismissed):

fuzz/pkcs7_verify.cc
```diff
@@ -0,0 +1,104 @@
...
+    }
+    store = X509_STORE_new();
+    if (!store) {
+      abort();
+    }
+    if (!X509_STORE_add_cert(store, cert)) {
+      abort();
+    }
+    certs = sk_X509_new_null();
+    if (!sk_X509_unshift(certs, cert)) {
```
WillChilds-Klein, line 68: np

```suggestion
    if (!certs || !sk_X509_unshift(certs, cert)) {
```

WillChilds-Klein (dismissed):

samuel40791765 (dismissed):

samuel40791765 (approved):

PR #2062, opened 2024-12-17
Rehaul PQDSA Test Suite

jakemas:
### Issues:
Resolves #CryptoAlg-2800

### Description of changes: 

FIPS 204 states:

> Algorithm 3, implementing verification for ML-DSA, and Algorithm 5, implementing verification for HashMLDSA, specify the length of the signature sigma and the public key pk in terms of the parameters described in Table 1 of FIPS 204. If an implementation of ML-DSA can accept inputs for sigma or pk of any other length, it shall return false whenever the lengths of either of these inputs differ from their lengths specified in this standard.

This description outlines how these checks are performed within our implementation of PQDSA in AWS-LC.

#### Signature Length

When attempting to verify a signature, the user will call:
```
EVP_DigestVerifyInit(ctx.get(), nullptr, nullptr, nullptr, pkey.get());
EVP_DigestVerify(ctx.get(), signature.data(), sig_len, msg.data(), mlen_int);
```

This will call into the PKEY function method for ML-DSA verify `pkey_pqdsa_verify_signature`, defined within `p_pqdsa.c`. This function performs the check that the signature length provided matches the expected signature length defined for the particular algorithm:

```
  if (sig_len != pqdsa->signature_len ||
      !pqdsa->method->verify(key->public_key, sig, siglen, message, message_len, NULL, 0)) {
    OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_SIGNATURE);
    return 0;
  }
```

We are able to explicitly prove these failure modes within the PQDSA test suite `PQDSAParameterTest`. The `SIGOperations` test has been modified to include a check that signature verification fails when providing a signature of invalid length. For example:

```
  // Check that verification fails upon providing a signature of invalid length
  sig_len = GetParam().signature_len - 1;
  ASSERT_FALSE(EVP_DigestVerify(md_ctx_verify.get(), sig1.data(), sig_len, msg1.data(), msg1.size()));
  err = ERR_get_error();
  EXPECT_EQ(ERR_LIB_EVP, ERR_GET_LIB(err));
  EXPECT_EQ(EVP_R_INVALID_SIGNATURE, ERR_GET_REASON(err));
```

#### Public Key Length

As the length of the public key is not provided to the verify API, we are not able to validate the size in the same way as signature length. However, as the public key is provided via a PKEY structure, we can enforce the requirement that prevents public keys of the incorrect length by checking the size of the public key during it’s addition to the PKEY structure.

**How do ML-DSA keys enter EVP Verify?**

The EVP API for verify init is as follows:

`EVP_DigestVerifyInit(&md_ctx_verify, nullptr, nullptr, nullptr, key);`

Where the public key either enters explicitly as a PKEY argument `key` or through the `EVP_MD_CTX` argument `md_ctx_verify` as a `pctx`.

Thus, to show that verify will only accept public key inputs of the correct length, we must evidence that:
we cannot construct a PQDSA PKEY such that the public/private key is not of the size required for that parameter set.

We now verify for every method to construct an PQDSA PKEY that the size of the public/private key are exactly the size required for that parameter set. The methods to construct an PQDSA PKEY are as follows:

1. use `EVP_PKEY_pqdsa_new_raw_public_key`
2. use `EVP_PKEY_pqdsa_new_raw_private_key`
3. use `EVP_parse_public_key`
4. use `EVP_parse_private_key`
5. Create a valid `PQDSA` keypair, then set the parameters to a different parameter set using `EVP_PKEY_CTX_pqdsa_set_params`

We now discuss each method, outlining any testing that has been added to provide evidence that this cannot occur (where necessary).

(1) `EVP_PKEY_pqdsa_new_raw_public_key`

This function takes three inputs; `nid, in, len,` where:

* `nid` is the algorithm identifier
* `in` is the contents of the public key (raw)
* `len` is the length of `in`

We test failure modes for all three input parameters within the `RawFunctions` tests as part of the PQDSA test suite `PQDSAParameterTest`. In particular, we verify that the length of the input public key `len` must be exactly equal to the expected length of the associated public key for that parameter set. Explicitly, this is verified within the function `EVP_PKEY_pqdsa_new_raw_public_key` by the check:

[p_pqdsa.c lines 217-220](https://github.com/jakemas/aws-lc/blob/d681431d9478b104ac0acaa44cff6b252f0a8dc8/crypto/fipsmodule/evp/p_pqdsa.c#L217-L220)
```
  if (pqdsa->public_key_len != len) {
    OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_BUFFER_SIZE);
    goto err;
  }
```

(2) `EVP_PKEY_pqdsa_new_raw_private_key`

Similarly to above, this function takes three inputs; `nid, in, len,` where:

* `nid` is the algorithm identifier
* `in` is the contents of the private key (raw)
* `len` is the length of `in`

Failure modes are tested within the `RawFunctions` tests as part of the PQDSA test suite `PQDSAParameterTest`. In particular, we verify that the length of the input private key `len` must be exactly equal to the expected length of the associated private key for that parameter set. Explicitly, this is verified within the function `EVP_PKEY_pqdsa_new_raw_private_key` by the check:

[p_pqdsa.c lines 247-250](https://github.com/jakemas/aws-lc/blob/d681431d9478b104ac0acaa44cff6b252f0a8dc8/crypto/fipsmodule/evp/p_pqdsa.c#L247-L250)

```
 if (pqdsa->private_key_len != len) {
    OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_BUFFER_SIZE);
    goto err;
  }
```

(3) `EVP_parse_public_key`

`EVP_parse_public_key` will call the specific PQDSA ASN.1 decode function `pqdsa_pub_decode`, which in turn will call `PQDSA_KEY_set_raw_public_key`. This function performs a lookup of the expected public key size from the parameter set, and duplicates exactly that many bytes from the input buffer to the PKEY’s public key. Disallowing any other size value.

[pqdsa.c line 71](https://github.com/jakemas/aws-lc/blob/d681431d9478b104ac0acaa44cff6b252f0a8dc8/crypto/fipsmodule/pqdsa/pqdsa.c#L71)

`key->public_key = OPENSSL_memdup(in, key->pqdsa->public_key_len);`

(4) `EVP_parse_private_key`

`EVP_parse_private_key` will call the specific PQDSA ASN.1 decode function `pqdsa_priv_decode`, which in turn will call `PQDSA_KEY_set_raw_private_key`. This function performs a lookup of the expected private key size from the parameter set, and duplicates exactly that many bytes from the input buffer to the PKEY’s private key. Disallowing any other size value.

[pqdsa.c line 80](https://github.com/jakemas/aws-lc/blob/d681431d9478b104ac0acaa44cff6b252f0a8dc8/crypto/fipsmodule/pqdsa/pqdsa.c#L80)

`key->private_key = OPENSSL_memdup(in, key->pqdsa->private_key_len);`

(5) `EVP_PKEY_CTX_pqdsa_set_params`

Theoretically, we could create bad keypairs by misaligning a PKEY with the associated parameter set. The function that sets PQDSA parameter sets is `EVP_PKEY_CTX_pqdsa_set_params`, that takes the arguments:

* `ctx` the input `EVP_PKEY_CTX` that we want to set the parameter set on
* `nid`the algorithm identifier

Failure modes are tested within the `KeyGen` tests as part of the PQDSA test suite `PQDSAParameterTest`. In particular, we verify that:

* the function fails with error `ERR_R_PASSED_NULL_PARAMETER` when `ctx` is null.
* the function fails with error `ERR_R_PASSED_NULL_PARAMETER` when `ctx->data` is null.
* the function fails with error `EVP_R_INVALID_OPERATION` when `ctx->pkey` is not null.
* the function fails with error `EVP_R_UNSUPPORTED_ALGORITHM` when `nid` is not a PQDSA.
* calling `EVP_PKEY_pqdsa_set_params` on an initalized PKEY will null the public/private keys

### Call-outs:
As we prepare to remove the `enable_dilithium` flag, I am pre-empting missing test coverage.
The `SIGOperations` test has been improved to the following flow:

1. KeyGen: generate PQDSA key for that NID
2. Signature Initialization: initialize sign with `EVP_DigestSignInit`
3. Signature Allocation: call `EVP_DigestSign` with signature length as null to get the expected size of the signature
    1. Verify that these sizes are as expected
4. Signature Generation: call `EVP_DigestSign` with correct signature length
5. Verification Init: init verification by calling `EVP_VerifyInit`
6. Verification: call `EVP_DigestVerify` to verify signature
7. Verify the signed message fails upon a different message
8. verify the signed message fails upon a different public key
9. sign a second message
10. check signatures are not equal
11. check that verify fails when trying to use a different signature
12. check that signatures are not deterministic: call another signature on the same message 
13. size checks: invalid signature lengths


crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -1108,232 +1108,323 @@ TEST_P(PQDSAParameterTest, KAT) {
...
+  pk_len = 0;
+  sk_len = 0;
+  ASSERT_TRUE(EVP_PKEY_get_raw_public_key(pkey.get(), nullptr, &pk_len));
+  ASSERT_TRUE(EVP_PKEY_get_raw_private_key(pkey.get(), nullptr, &sk_len));
+  EXPECT_EQ(pk_len, GetParam().public_key_len);
+  EXPECT_EQ(sk_len, GetParam().private_key_len);
+
+  // ---- 4. Test creating PKEYs from raw data ----
+  bssl::UniquePtr<EVP_PKEY> public_pkey(EVP_PKEY_pqdsa_new_raw_public_key(nid, pkey->pkey.pqdsa_key->public_key, pk_len));
+  bssl::UniquePtr<EVP_PKEY> private_pkey(EVP_PKEY_pqdsa_new_raw_private_key(nid, pkey->pkey.pqdsa_key->private_key, sk_len));
```
nebeid, line 1273: Lines length

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -1108,232 +1108,323 @@ TEST_P(PQDSAParameterTest, KAT) {
...
+
+  // ---- 4. Test failure modes for EVP_PKEY_CTX_pqdsa_set_params. ----
+  // ctx is NULL.
+  ASSERT_FALSE(EVP_PKEY_CTX_pqdsa_set_params(nullptr, nid));
   uint32_t err = ERR_get_error();
   EXPECT_EQ(ERR_LIB_EVP, ERR_GET_LIB(err));
-  EXPECT_EQ(EVP_R_BUFFER_TOO_SMALL, ERR_GET_REASON(err));
-  OPENSSL_free(buf);
-  buf = nullptr;
+  EXPECT_EQ(ERR_R_PASSED_NULL_PARAMETER, ERR_GET_REASON(err));
```
nebeid, line 1160: If you want, these three lines that repeat can be made into a macro. I think all of them are ERR_LIB_EVP.

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -1343,42 +1434,97 @@ TEST_P(PQDSAParameterTest, SIGOperations) {
...
-  ASSERT_TRUE(EVP_DigestVerify(md_ctx.get(), signature1.data(), sig_len,
-                               msg1.data(), msg1.size()));
+  ASSERT_TRUE(EVP_DigestVerifyInit(md_ctx_verify.get(), nullptr, nullptr, nullptr, pkey.get()));
+  ASSERT_TRUE(EVP_DigestVerify(md_ctx_verify.get(), sig1.data(), sig_len, msg1.data(), msg1.size()));
 
-  // Verify the signed message fails upon a different message
-  ASSERT_FALSE(EVP_DigestVerify(md_ctx.get(), signature1.data(), sig_len,
-                                msg2.data(), msg2.size()));
+  // ---- 3. Test signature failure modes: incompatible messages/signatures ----
+  // Check that the verification of signature1 failes for a different message; message2
```
nebeid, line 1461:
```suggestion
  // Check that the verification of signature1 fails for a different message; message2
```

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -1343,42 +1434,97 @@ TEST_P(PQDSAParameterTest, SIGOperations) {
       0x4a, 0x41, 0x4b, 0x45, 0x20, 0x4d, 0x41, 0x53, 0x53, 0x49,
       0x4d, 0x4f, 0x20, 0x41, 0x57, 0x53, 0x32, 0x30, 0x32, 0x31, 0x2e};
 
-  ASSERT_TRUE(EVP_DigestSignInit(md_ctx.get(), nullptr, nullptr,
-                                 nullptr, pkey.get()));
-  ASSERT_TRUE(EVP_DigestSign(md_ctx.get(), signature1.data(), &sig_len,
-                             msg1.data(), msg1.size()));
+  // ---- 2. Test signature round trip (sign + verify) ----
+
+  // Initalize the signing context |md_ctx| with the |pkey| we generated
```
nebeid, line 1439: If you want, you can replace in the comments
message1 -> msg1
message2 -> msg2

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -1343,42 +1434,97 @@ TEST_P(PQDSAParameterTest, SIGOperations) {
...
+  uint32_t err = ERR_get_error();
+  EXPECT_EQ(ERR_LIB_EVP, ERR_GET_LIB(err));
+  EXPECT_EQ(EVP_R_INVALID_SIGNATURE, ERR_GET_REASON(err));
 
-  // Sign the different message
+  // reset the contexts between tests
+  md_ctx.Reset();
+  md_ctx_verify.Reset();
+
+  // PQDSA signature schemes can be in either randomized (every signature on a
```
nebeid, line 1471:
```suggestion
  // PQDSA signature schemes can be either in randomized (every signature on a
```

nebeid (commented):
Thank you for documenting these tests so thoroughly, Jake, and for the detailed PR description. Most comments are Nits. One return value is to be discussed.

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -1108,232 +1108,323 @@ TEST_P(PQDSAParameterTest, KAT) {
...
-  // Extract private key and check it is of the correct size
-  EXPECT_TRUE(EVP_PKEY_get_raw_private_key(pkey.get(), buf, &buf_size));
-  EXPECT_EQ((size_t)sk_len, buf_size);
+  // ctx->data is NULL
+  void *tmp = ctx.get()->data;
+  ctx.get()->data = nullptr;
+  ASSERT_FALSE(EVP_PKEY_CTX_pqdsa_set_params(ctx.get(), nid));
+  err = ERR_get_error();
+  EXPECT_EQ(ERR_LIB_EVP, ERR_GET_LIB(err));
+  EXPECT_EQ(ERR_R_PASSED_NULL_PARAMETER, ERR_GET_REASON(err));
```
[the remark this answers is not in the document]
jakemas: Hmm, we set `ctx->data` to NULL, then pass in `ctx`, expecting a failure because of a null value. This is consistant with how we use the error code in ML-KEM at https://github.com/jakemas/aws-lc/blob/ab8953b7478f72a2b80497cf54ce0b72ed719cfc/crypto/evp_extra/evp_extra_test.cc#L2229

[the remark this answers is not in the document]
nebeid: It's good to be consistent, I agree. We can later revisit it to see if we want to return another error message. I believe the data would be null if the context wasn't initialised or something.

crypto/dilithium/p_pqdsa_test.cc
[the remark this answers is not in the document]
jakemas: I actually disagree here, I was inspired by Dusan's extremely clean and readable tests for ML-KEM that order the code systematically, calling similar functionality together, for example see:
https://github.com/jakemas/aws-lc/blob/ab8953b7478f72a2b80497cf54ce0b72ed719cfc/crypto/evp_extra/evp_extra_test.cc#L2528-L2533

(Also, it's used just 7 lines down from where it's defined, which seems more readable personally to me)

crypto/dilithium/p_pqdsa_test.cc
jakemas: fixed in [f7d1dad](https://github.com/aws/aws-lc/pull/2062/commits/f7d1dadb64f54b7695d871a88d4dfacf42048637)

crypto/dilithium/p_pqdsa_test.cc
jakemas: fixed in [f7d1dad](https://github.com/aws/aws-lc/pull/2062/commits/f7d1dadb64f54b7695d871a88d4dfacf42048637)

crypto/dilithium/p_pqdsa_test.cc
jakemas: fixed in [f7d1dad](https://github.com/aws/aws-lc/pull/2062/commits/f7d1dadb64f54b7695d871a88d4dfacf42048637)

crypto/dilithium/p_pqdsa_test.cc
jakemas: fixed in [f7d1dad](https://github.com/aws/aws-lc/pull/2062/commits/f7d1dadb64f54b7695d871a88d4dfacf42048637)

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -1108,232 +1108,323 @@ TEST_P(PQDSAParameterTest, KAT) {
...
+
+  // Create a new PKEY from the generated raw key
+  bssl::UniquePtr<EVP_PKEY> pkey(raw);
+  EVP_PKEY_CTX_free(ctx);
+  return pkey;
 }
 
 TEST_P(PQDSAParameterTest, KeyCmp) {
-  // Generate two MLDSA keys are check that they are not equal.
+  // Generate two PQDSA keys are check that they are not equal.
```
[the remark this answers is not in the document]
jakemas: fixed in [f7d1dad](https://github.com/aws/aws-lc/pull/2062/commits/f7d1dadb64f54b7695d871a88d4dfacf42048637)

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -1108,232 +1108,323 @@ TEST_P(PQDSAParameterTest, KAT) {
...
+  ASSERT_TRUE(EVP_PKEY_pqdsa_set_params(pkey.get(), nid));
+  ASSERT_TRUE(pkey.get()->pkey.pqdsa_key->public_key == nullptr);
+  ASSERT_TRUE(pkey.get()->pkey.pqdsa_key->private_key == nullptr);
+}
+
+// Helper function that:
+// 1. Creates a |EVP_PKEY_CTX| object of type: EVP_PKEY_PQDSA.
+// 2. Sets the specific PQDSA parameters according to the |pqdsa_nid| provided.
+// 3. Generates a key pair.
+// 4. Creates an EVP PKEY object from the generated key (as a bssl::UniquePtr).
```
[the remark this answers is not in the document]
jakemas: fixed in [f7d1dad](https://github.com/aws/aws-lc/pull/2062/commits/f7d1dadb64f54b7695d871a88d4dfacf42048637)

crypto/dilithium/p_pqdsa_test.cc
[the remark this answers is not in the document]
jakemas: Shall we leave as is for now then? In my opinion the most useful such a generic error message can be is when it is consistant with other failures cases in the library.

crypto/dilithium/p_pqdsa_test.cc
jakemas: I'm not sure I understand exactly what you mean here. Open to suggestions.

nebeid: These 3 lines
```
 uint32_t err = ERR_get_error();
  EXPECT_EQ(ERR_LIB_EVP, ERR_GET_LIB(err));
  EXPECT_EQ(ERR_R_PASSED_NULL_PARAMETER, ERR_GET_REASON(err));
```
repeat throughout the tests and can be replaced by a single-line macro that takes the reason as a parameter (similar to the CMP macros).
It will condense the code a bit.

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -1182,11 +1187,6 @@ TEST_P(PQDSAParameterTest, KeyGen) {
   EXPECT_EQ(ERR_LIB_EVP, ERR_GET_LIB(err));
   EXPECT_EQ(EVP_R_UNSUPPORTED_ALGORITHM, ERR_GET_REASON(err));
   ctx.get()->pkey = (EVP_PKEY*) tmp;
-
-  // attempt to change NID after PKEY is set and verify that this null's the key
-  ASSERT_TRUE(EVP_PKEY_pqdsa_set_params(pkey.get(), nid));
-  ASSERT_TRUE(pkey.get()->pkey.pqdsa_key->public_key == nullptr);
-  ASSERT_TRUE(pkey.get()->pkey.pqdsa_key->private_key == nullptr);
```
jakemas: I removed this check because to run it, it requires exposing `EVP_PKEY_pqdsa_set_params`. Currently, this function isn't exposed, so changing the NID after the PKEY is set is not something that is possible -- by exposing the function to run the test, we would make such behaviour possible.

crypto/dilithium/p_pqdsa_test.cc
jakemas: Okay, see https://github.com/aws/aws-lc/pull/2062/commits/463aa0d675b49755a0f796ed8827483a52bd0bd2#diff-d043ba12b97b57a0d69c31db07b10cee8ca8c05cf41e1875981b254bc7cbe670L1186-L1189

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -1269,8 +1269,10 @@ TEST_P(PQDSAParameterTest, RawFunctions) {
...
 
   // ---- 4. Test creating PKEYs from raw data ----
-  bssl::UniquePtr<EVP_PKEY> public_pkey(EVP_PKEY_pqdsa_new_raw_public_key(nid, pkey->pkey.pqdsa_key->public_key, pk_len));
-  bssl::UniquePtr<EVP_PKEY> private_pkey(EVP_PKEY_pqdsa_new_raw_private_key(nid, pkey->pkey.pqdsa_key->private_key, sk_len));
+  bssl::UniquePtr<EVP_PKEY>public_pkey(
+    EVP_PKEY_pqdsa_new_raw_public_key(nid, pkey->pkey.pqdsa_key->public_key, pk_len));
+  bssl::UniquePtr<EVP_PKEY> private_pkey(
+    EVP_PKEY_pqdsa_new_raw_private_key(nid, pkey->pkey.pqdsa_key->private_key, sk_len));
 
   // check that public key is present and private key is not present
```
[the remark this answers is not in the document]
jakemas: added in https://github.com/aws/aws-lc/pull/2062/commits/463aa0d675b49755a0f796ed8827483a52bd0bd2#diff-d043ba12b97b57a0d69c31db07b10cee8ca8c05cf41e1875981b254bc7cbe670L1186-L1189

crypto/dilithium/p_pqdsa_test.cc
jakemas: [16f48d6](https://github.com/aws/aws-lc/pull/2062/commits/16f48d60912aa7e259f5fd809ba0f90abaa1c976) done

dkostic (approved):

crypto/dilithium/p_pqdsa_test.cc
```diff
@@ -986,6 +986,14 @@ CMP_VEC_AND_PTR(vec, pkey->pkey.pqdsa_key->public_key, len)
 #define CMP_VEC_AND_PKEY_SECRET(vec, pkey, len) \
 CMP_VEC_AND_PTR(vec, pkey->pkey.pqdsa_key->private_key, len)
 
+#define GET_ERR_AND_CHECK_REASON(reason)              \
+          {                                           \
+            uint32_t err = ERR_get_error();           \
```
jakemas, line 991: does this need to be scoped differently?

dkostic (approved):

nebeid (approved):

PR #2065, opened 2024-12-18
Alias OpenSSL SECLEVEL directives to ALL

WillChilds-Klein:
### Issues:
Addresses CryptoAlg-2792

### Description of changes: 

To increase compatibility with OpenSSL's notion of "security levels", this commit aliases all security levels ≤ 2 to AWS-LC's `ALL` alias. Described [here](https://docs.openssl.org/master/man3/SSL_CTX_set_security_level/#default-callback-behaviour), OpenSSL's security level 0 provides no minimum.

### Testing:

- CI with reduced CPython patch set

---

### DIFF
--- a/ssl/ssl_cipher.cc
+++ b/ssl/ssl_cipher.cc
@@ -1187,11 +1187,14 @@ static bool ssl_cipher_process_rulestr(const char *rule_str,
 
     // Ok, we have the rule, now apply it.
     if (rule == CIPHER_SPECIAL) {
-      if (buf_len != 8 || strncmp(buf, "STRENGTH", 8) != 0) {
+      if (buf_len == 8 && strncmp(buf, "STRENGTH", 8) == 0) {
+        if (!ssl_cipher_strength_sort(head_p, tail_p)) {
+          return false;
+        }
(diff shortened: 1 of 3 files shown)


samuel40791765:
Just to give some more context, I initially set our default security level to 3 under the same justifications. However, I had to change the value to 0 later on because we don't directly prohibit 512 bit RSA keys like OpenSSL states with Level 1.. 
This commit has more details: https://github.com/aws/aws-lc/commit/42cd981669896b9d5ae9b1d91417b87638d65b37

WillChilds-Klein:
Thanks for the context @samuel40791765.

> However, I had to change the value to 0 later on because we don't directly prohibit 512 bit RSA keys

Do you mean for parsing or generating?

I believe we enforce minimum of 2048 bits in FIPS mode, but looking around a bit, I do see a surprisingly low minimum key size threshold [of 256 bits](https://github.com/aws/aws-lc/blob/c0e927e9d0711f3830496f21140aa59803ca856b/crypto/fipsmodule/rsa/rsa_impl.c#L1014). Given this, I agree that we should only honor `SECLEVEL=0`.

PR #2070, opened 2024-12-20
Remove ENABLE_DILITHIUM flag

jakemas:
### Issues:

### Description of changes: 
- Added an experimental flag (based on the deprecated flag) to indicate the external APIs may still be subject to change as the standards space develops.
- Renamed `EVP_PKEY_pqdsa_new_raw_private_key` to `EVP_PKEY_pqdsa_new_raw_secret_key` to better match the equivalent API for KEMs (`EVP_PKEY_kem_new_raw_secret_key`). 

### Call-outs:

Removing the flag has little consequence -- other than it makes the APIs we expose in `include/openssl/evp.h` that much more "final". We should consider how much we like them before we commit to them. We made a point to refer to asymmetric keypairs as public and private keys, rather than public and secret keys. However, we haven't always been consistant with this, so there is a mix of both in the library. Users will find the consistency between `EVP_PKEY_pqdsa_new_raw_secret_key` and `EVP_PKEY_kem_new_raw_secret_key` more satisfying.

Once we are happy with the name, including the decision around `PQDSA`, then we can remove the experimental flags. Personally, I'd advocate for NISTDSA as all NIST signature schemes are matching these API conventions (sign, pre-hash-sign, context-sign -- see ed25519, ML-DSA, SLH-DSA, FN-DSA as examples). Most of this naming is internal, with the exception of the three external functions: `EVP_PKEY_CTX_pqdsa_set_params`, `EVP_PKEY_pqdsa_new_raw_secret_key`, `EVP_PKEY_pqdsa_new_raw_public_key`.

### Testing:
To celebrate the removal of this flag, enjoy this haiku:
```
feature flag removed,
post-quantum strength stands alone,
simpler code prevails.
```


jakemas:
Seems to be multiple definitions of `crypto_sign_keypair` and `crypto_sign_open` when integration testing against `libmariadb` who seem to use the same names for their `ed25519` implementation.

jakemas:
cleaning up and splitting PR. new PR is https://github.com/aws/aws-lc/pull/2081

PR #2077, opened 2024-12-23
Fixes varios issues with rebuilding CI Docker images

skmcgrail:
* Update Intel SDE from `9.21.1` to `9.44.0` as the older version is no longer available on at the mirror location.
* Update the AL2023 SSM Agent file to pull the changed path location for the manifest for GitHub.
* Update the CentOS 7 docker image's mirrors to use the vault location as the previous mirror is no longer serving the repository files.
* Since cryptofuzz is no longer available on GitHub from the original upstream author, update it to pull an archived file from a CloudFront distribution from our team AWS account. I have also put a copy of the code in AWS CodeCommit for a complete git history.
* Botan repository tip isn't building for us on AL2023 anymore (looks like C++ 20 related / issues with cryptofuzz maybe?), anyways pinned to the current cryptofuzz we've been using and botan version that was released around that time.
* Update CDK dependencies for good hygiene
* Turns out our Windows CDK image build was never truly automated....so fix that by automating the long-standing issue of not installing the Docker engine.


andrewhop (approved):

justsmth (approved):

tests/ci/cdk/cdk/windows_docker_image_build_stack.py
```diff
@@ -1,66 +1,124 @@
...
-                                vpc_subnets=ec2.SubnetSelection(subnet_type=ec2.SubnetType.PUBLIC),
-                                machine_image=machine_image)
+        block_device_volume = ec2.BlockDeviceVolume.ebs(
+            volume_size=200, delete_on_termination=True
+        )
+        block_device = ec2.BlockDevice(
+            device_name="/dev/sda1", volume=block_device_volume
+        )
+
+        setup_user_data = ec2.UserData.for_windows()
```
samuel40791765, line 101: Are there any gotchas or things to know about "user data"? This seems like something really useful I could steal for the ec2 test framework.

samuel40791765 (approved):

tests/ci/cdk/cdk/windows_docker_image_build_stack.py
skmcgrail: I believe the biggest gotcha is that the script are only ran on the first instance launch. So if you stop the instance, update the data, and start the instance again they won't automatically run. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/user-data.html has more details.

PR #1332, opened 2023-11-30
Attempt to get fips scope markers through adrp instead of adr to increase reach

torben-hansen:
### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 
Describe AWS-LC’s current behavior and how your code changes that behavior. If there are no issues this pr is resolving, explain why this change is necessary.

### Call-outs:
Point out areas that need special attention or support during the review process. Discuss architecture or design changes.

### Testing:
How is this change tested (unit tests, fuzz tests, etc.)? Are there any testing steps to be verified by the reviewer?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

PR #2082, opened 2024-12-26
Remove ENABLE_DILITHIUM flag

jakemas:
### Issues:

### Description of changes: 
- Added a description to APIs to indicate the external APIs may still be subject to change as the standards space develops. (We don't envision them changing, but they are extremely new).

### Call-outs:

Removing the flag has little consequence -- other than it makes the APIs we expose in `include/openssl/evp.h` that much more "final". We should consider how much we like them before we commit to them. We made a point to refer to asymmetric keypairs as public and private keys, rather than public and secret keys. However, we haven't always been consistant with this, so there is a mix of both in the library. Users will find the consistency between `EVP_PKEY_pqdsa_new_raw_secret_key` and `EVP_PKEY_kem_new_raw_secret_key` more satisfying.

However, after internal discussions, we'd also like to change the name of `EVP_PKEY_kem_new_raw_public_key` and `EVP_PKEY_kem_new_raw_secret_key` to `EVP_PKEY_kem_new_raw_encapsulation_key` and `EVP_PKEY_kem_new_raw_decapsulation_key`  to match the names within [FIPS 203](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf). Thus, we should stick with `secret`.

We did consider the use of an experimental flag (by using the OPENSSL_DEPRECATED alias), but, this seems a little over the top. Alternatively, we could place the EVP APIs in `include/openssl/experimental/`. The arguments of `EVP_PKEY_pqdsa_new_raw_secret_key`, `EVP_PKEY_pqdsa_new_raw_public_key`, and `EVP_PKEY_CTX_pqdsa_set_params` are stable and consistant with `KEM` and `EC` variants (see [EVP_PKEY_CTX_kem_set_params](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L914), [EVP_PKEY_CTX_set_dh_paramgen_prime_len](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L184), and [EVP_PKEY_CTX_set_ec_param_enc](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L1151), [EVP_PKEY_new_raw_private_key](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L278), [EVP_PKEY_new_raw_public_key](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L284), [EVP_PKEY_kem_new_raw_public_key](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L920)).
### Testing:
To celebrate the removal of this flag, enjoy this haiku:
```
feature flag removed,
post-quantum strength stands alone,
simpler code prevails.
```


crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -483,11 +483,11 @@ int crypto_sign_verify(ml_dsa_params *params,
   pre[0] = 0;
   pre[1] = ctxlen;
   OPENSSL_memcpy(pre + 2 , ctx, ctxlen);
-  return crypto_sign_verify_internal(params, sig, siglen, m, mlen, pre, 2 + ctxlen, pk);
+  return ml_dsa_verify_internal(params, sig, siglen, m, mlen, pre, 2 + ctxlen, pk);
 }
 
 /*************************************************
-* Name:        crypto_sign_open
+* Name:        ml_dsa_verify_message
```
jakemas, line 490: `sign_message` and `verify_message` variants are never used (we use sign/verify without the appended message on the end, so that signature sizes are constant)-- should we just remove them from the library?

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
```diff
@@ -300,7 +300,7 @@ int crypto_sign_signature(ml_dsa_params *params,
 }
 
 /*************************************************
-* Name:        crypto_sign
+* Name:        ml_dsa_sign_message
```
jakemas, line 303: `sign_message` and `verify_message` variants are never used (we use sign/verify without the appended message on the end, so that signature sizes are constant)-- should we just remove them from the library?

crypto/dilithium/pqcrystals_dilithium_ref_common/sign.c
WillChilds-Klein: I usually take the position that if

1. a function is never used and
2. it is not exposed as part of our public API

we should remove it.

WillChilds-Klein (commented):
Holding off on full review until  PR 2072, which this PR holds the contents of, has been merged.

> However, we haven't always been consistant with this, so there is a mix of both in the library. Users will find the consistency between EVP_PKEY_pqdsa_new_raw_secret_key and EVP_PKEY_kem_new_raw_secret_key more satisfying.

I associate "private key" with asymmetric keys and "secret key" with symmetric keys for HMAC, AES, etc. We have portions of the library that refer to asymmetric private keys as "secret keys"?

---

Why do we need new EVP APIs for ML-DSA? What inputs/outputs differ from other signature schemes supported by EVP_DigestSign?

jakemas:
> Holding off on full review until PR 2072, which this PR holds the contents of, has been merged.
> 
> > However, we haven't always been consistant with this, so there is a mix of both in the library. Users will find the consistency between EVP_PKEY_pqdsa_new_raw_secret_key and EVP_PKEY_kem_new_raw_secret_key more satisfying.
> 
> I associate "private key" with asymmetric keys and "secret key" with symmetric keys for HMAC, AES, etc. We have portions of the library that refer to asymmetric private keys as "secret keys"?

Yes, see `EVP_PKEY_kem_new_raw_secret_key`.
> 
> Why do we need new EVP APIs for ML-DSA? What inputs/outputs differ from other signature schemes supported by EVP_DigestSign?

We need them for the PKEY and ASN.1 functions, not for the `DigestSign` part. The APIs are:
```
// EVP_PKEY_CTX_pqdsa_set_params sets in |ctx| the parameters associated with
// the signature scheme defined by the given |nid|. It returns one on success
// and zero on error. This API is marked as EXPERIMENTAL (using the deprecated
// warning) to indicate that this API may change as the standards around the
// signature schemes finalize.
OPENSSL_EXPORT int EVP_PKEY_CTX_pqdsa_set_params(EVP_PKEY_CTX *ctx, int nid);

// EVP_PKEY_pqdsa_new_raw_public_key generates a new EVP_PKEY object of type
// EVP_PKEY_PQDSA, initializes the PQDSA key based on |nid| and populates the
// public key part of the PQDSA key with the contents of |in|. It returns the
// pointer to the allocated PKEY on sucess and NULL on error. This API is marked
// as EXPERIMENTAL to indicate that this API may change as the standards around
// the signature schemes finalize.
OPENSSL_EXPORT EVP_PKEY *EVP_PKEY_pqdsa_new_raw_public_key(int nid, const uint8_t *in, size_t len);

// EVP_PKEY_pqdsa_new_raw_private_key generates a new EVP_PKEY object of type
// EVP_PKEY_PQDSA, initializes the PQDSA key based on |nid| and populates the
// secret key part of the PQDSA key with the contents of |in|. It returns the
  // pointer to the allocated PKEY on sucess and NULL on error. This API is marked
// as EXPERIMENTAL to indicate that this API may change as the standards around
// the signature schemes finalize.
OPENSSL_EXPORT EVP_PKEY *EVP_PKEY_pqdsa_new_raw_private_key(int nid, const uint8_t *in, size_t len);
```

These are analog to the KEM, DH, and EC PKEY utility function APIs:

- [EVP_PKEY_CTX_kem_set_params](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L914)
- [EVP_PKEY_CTX_set_dh_paramgen_prime_len](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L184)
- [EVP_PKEY_CTX_set_ec_param_enc](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L1151)
- [EVP_PKEY_kem_new_raw_public_key](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L920)

We need these so we don't have to have distinct ASN.1 functions for every single ML-DSA parameter set, instead, we put all signature schemes (that support the NIST API framework, so all the PQ ones) into the `PQDSA` PKEY struct and use these utility functions to set a specific NID.

WillChilds-Klein:
Thanks for the explanation @jakemas, all makes sense. Are we using "pqdsa" instead of "mldsa" because we plan to re-use these functions for SLH-DSA?

jakemas:
> Thanks for the explanation @jakemas, all makes sense. The only (mild) concern I still have is whether the "pqdsa" naming might be too general here. Are we using "pqdsa" instead of "mldsa" because we plan to re-use these functions for SLH-DSA?

Yes, the idea is to support any signature scheme without having to add an ASN.1 functionality for each algorithm specific parameter set. This would include ML-DSA, HashML-DSA, ExtMuML-DSA, FN-DSA, HashFN-DSA, SLH-DSA, HashSLH-DSA, even ED25519 would fit in the API. 


> If so, with the current signature of `EVP_PKEY_CTX_pqdsa_set_params`, we'd need a distinct SLH-DSA NID for each of the 12 parameter sets listed in Section 11 of [FIPS 205](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.205). This doable, but a bit unwieldy.

I don't understand how this would be considered unwieldy, the PQDSA API supports this very simply. Demo:

```
// Generate a key-pair for the desired PQDSA algorithm.
int generate_key_pair(/* OUT */ EVP_PKEY **key,
                      /* IN  */ int pqdsa_nid) {

    EVP_PKEY_CTX *ctx = NULL;

    // Create the PQDSA contex.
    ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_PQDSA, NULL);
    if (!ctx ||
        !EVP_PKEY_CTX_pqdsa_set_params(ctx, pqdsa_nid) ||
        !EVP_PKEY_keygen_init(ctx) ||
        !EVP_PKEY_keygen(ctx, key)) {
        EVP_PKEY_free(*key);
        EVP_PKEY_CTX_free(ctx);
        return 0;
        }
    // Note: a single context can be used to generate many keys.
    EVP_PKEY_CTX_free(ctx);
    return 1;
}
```

You just set the particular algorithm parameter set using `EVP_PKEY_CTX_pqdsa_set_params(ctx, pqdsa_nid)`.

Your proposal for `EVP_PKEY_CTX_slhdsa_set_params` would work the exact same way, other than we'd now have more EVP APIs, one for ML-DSA and one for SLH-DSA, and then presumably more for each other algorithm? Why would we want that over a single algorithm setter function?

> It might be better to have a distinct `EVP_PKEY_CTX_slhdsa_set_params` that takes in relevant hash NIDs, in which case we'd probably want to rename `EVP_PKEY_CTX_pqdsa_set_params` to `EVP_PKEY_CTX_mldsa_set_params` to disambiguate. Alternatively, we could add the 2 additional hash NIDs as parameters to `EVP_PKEY_CTX_pqdsa_set_params` and instruct callers to specify `NID_undef` there for ML-DSA.

Why are you talking about hash NIDs here? Are you talking about pre-hash modes here? This has nothing to do with signing yet still, we're talking about EVP PKEY stuff. The pre-hash NIDs come in through the signing context.

> Lastly, what are the implications of XMSS for this API? Is XMSS considered part of NIST's "DSA" family of signature algorithms?

No plans to implement XMSS. No need to consider its support.

jakemas:
It looks like the same issue we saw in https://github.com/aws/aws-lc/pull/1332 is happening in this PR too. Perhaps the removal of the flag caused the offset to change and is now out of range?
```
bcm.c:7987: Error: pc-relative address offset out of range
bcm.c:7989: Error: pc-relative address offset out of range
bcm.c:7991: Error: pc-relative address offset out of range
```

WillChilds-Klein:
> I wanted to call it NISTDSA, because DSA is taken, and the API format is not PQ specific, it is NIST specific -- but I got push back in review on that name and have changed it to please the reviewers already.

If there's consensus has been reached previously regarding this naming convention, then please disregard my concern. I don't want to reconsider a decision that's already been made.

My goal here is to make sure we've discussed alternative interfaces because by removing this flag, we're committing to a public API that will require breaking changes to modify.

> I don't understand how this would be considered unwieldy

I wasn't specific enough here. Agreed that the current interface is easy for the caller to use. I meant that it would be "unwieldy" for us to add so many NIDs (one for each of 12 parameter sets) when the hash NID could be specified independently in `EVP_PKEY_CTX_pqdsa_set_params`. I didn't realize supporting a subset of parameter sets was on the table.

> Your proposal for EVP_PKEY_CTX_slhdsa_set_params would work the exact same way? We'd still have to select the particular parameter set from the OID. The down side would be that we'd now have more EVP APIs, one for ML-DSA and one for SLH-DSA, and then presumably more for each other algorithm? Why would we want that over a single algorithm setter function?

Not suggesting we move forward with this, but here's a sketch of the interface i was thinking:

```
int EVP_PKEY_CTX_mldsa_set_params(EVP_PKEY_CTX ctx, int mldsa_nid);
int EVP_PKEY_CTX_slhdsa_set_params(EVP_PKEY_CTX ctx, int hash_nid);
```

Caller would specify which ML-DSA security level they want via `mldsa_nid` or  SLH-DSA parameter set they want via `hash_nid`. But this is awkward (no dedicated SLH-DSA NIDs?) and inconsistent with the KEM API, with no discernible benefit other than reducing number of NIDs for us to provide -- a marginal benefit at best.

I agree that a single API with per-algorithm, per-parameter-set NIDs is preferable.

> I could see an argument for replacing all mentions of PQDSA with MLDSA and have it be the ML-DSA API, until we come to a point in the future in which we want to add another algorithm, then we can either rename it then to a generic name

I don't mean to suggest that you're advocating for this, but IMO we should avoid future renames at ~all costs. They're expensive in places where we have downstream telemetry (internal consumers), and very very difficult to do safely where we don't have downstream telemetry (external consumers).

So, we need to get this API right before removing the feature flag. I'm convinced that the current interface is the right one 👍

include/openssl/evp.h
```diff
@@ -944,24 +940,28 @@ OPENSSL_EXPORT int EVP_PKEY_kem_check_key(EVP_PKEY *key);
...
 // PQDSA specific functions.
 
-#ifdef ENABLE_DILITHIUM
 // EVP_PKEY_CTX_pqdsa_set_params sets in |ctx| the parameters associated with
 // the signature scheme defined by the given |nid|. It returns one on success
-// and zero on error.
+// and zero on error. This API is marked as EXPERIMENTAL (using the deprecated
+// warning) to indicate that this API may change as the standards around the
+// signature schemes finalize.
 OPENSSL_EXPORT int EVP_PKEY_CTX_pqdsa_set_params(EVP_PKEY_CTX *ctx, int nid);
```
WillChilds-Klein, line 948: per doc comment update:

```suggestion
OPENSSL_EXPORT OPENSSL_DEPRECATED int EVP_PKEY_CTX_pqdsa_set_params(EVP_PKEY_CTX *ctx, int nid);
```

include/openssl/evp.h
```diff
@@ -944,24 +940,28 @@ OPENSSL_EXPORT int EVP_PKEY_kem_check_key(EVP_PKEY *key);
...
 
 // EVP_PKEY_pqdsa_new_raw_public_key generates a new EVP_PKEY object of type
 // EVP_PKEY_PQDSA, initializes the PQDSA key based on |nid| and populates the
 // public key part of the PQDSA key with the contents of |in|. It returns the
-// pointer to the allocated PKEY on sucess and NULL on error.
+// pointer to the allocated PKEY on sucess and NULL on error. This API is marked
+// as EXPERIMENTAL to indicate that this API may change as the standards around
+// the signature schemes finalize.
 OPENSSL_EXPORT EVP_PKEY *EVP_PKEY_pqdsa_new_raw_public_key(int nid, const uint8_t *in, size_t len);
```
WillChilds-Klein, line 956: per doc comment update:

```suggestion
OPENSSL_EXPORT OPENSSL_DEPRECATED int EVP_PKEY_CTX_pqdsa_set_params(EVP_PKEY_CTX *ctx, int nid);
```

include/openssl/evp.h
```diff
@@ -944,24 +940,28 @@ OPENSSL_EXPORT int EVP_PKEY_kem_check_key(EVP_PKEY *key);
...
 
 // EVP_PKEY_pqdsa_new_raw_private_key generates a new EVP_PKEY object of type
 // EVP_PKEY_PQDSA, initializes the PQDSA key based on |nid| and populates the
 // secret key part of the PQDSA key with the contents of |in|. It returns the
-// pointer to the allocated PKEY on sucess and NULL on error.
+// pointer to the allocated PKEY on sucess and NULL on error. This API is marked
+// as EXPERIMENTAL to indicate that this API may change as the standards around
+// the signature schemes finalize.
 OPENSSL_EXPORT EVP_PKEY *EVP_PKEY_pqdsa_new_raw_private_key(int nid, const uint8_t *in, size_t len);
```
WillChilds-Klein, line 964: per doc comment update:

```suggestion
OPENSSL_EXPORT OPENSSL_DEPRECATED int EVP_PKEY_CTX_pqdsa_set_params(EVP_PKEY_CTX *ctx, int nid);
```

include/openssl/pem.h
```diff
@@ -88,12 +88,8 @@ extern "C" {
 #define PEM_STRING_PUBLIC "PUBLIC KEY"
 #define PEM_STRING_RSA "RSA PRIVATE KEY"
 #define PEM_STRING_RSA_PUBLIC "RSA PUBLIC KEY"
-
-#ifdef ENABLE_DILITHIUM
 #define PEM_STRING_DILITHIUM3 "DILITHIUM3 PRIVATE KEY"
 #define PEM_STRING_DILITHIUM3_PUBLIC "DILITHIUM3 PUBLIC KEY"
```
WillChilds-Klein, line 92: can we remove these? they're [not used anywhere](https://github.com/search?q=repo%3Aaws%2Faws-lc%20PEM_STRING_DILITHIUM3_PUBLIC&type=code).

jakemas: removed in [3c9fde8](https://github.com/aws/aws-lc/pull/2082/commits/3c9fde88fcf6ba11630b567249245199e33f3d71)

include/openssl/evp.h
```diff
@@ -944,24 +940,28 @@ OPENSSL_EXPORT int EVP_PKEY_kem_check_key(EVP_PKEY *key);
 
 // PQDSA specific functions.
 
-#ifdef ENABLE_DILITHIUM
 // EVP_PKEY_CTX_pqdsa_set_params sets in |ctx| the parameters associated with
 // the signature scheme defined by the given |nid|. It returns one on success
-// and zero on error.
+// and zero on error. This API is marked as EXPERIMENTAL (using the deprecated
+// warning) to indicate that this API may change as the standards around the
```
andrewhop, line 946: Functions not actually marked as experimental/deprecated.

andrewhop (commented):
The change looks good, but you'll need to fix the FIPS delaactor issue for ubuntu2204_gcc11x_aarch_fips.

include/openssl/evp.h
WillChilds-Klein: What's the right mechanism for this? `OPENSSL_DEPRECATED`? Or should we create a similar macro `OPENSSL_EXPERIMENTAL`?

jakemas: Andrew and I considered the use of an alias OPENSSL_EXPERIMENTAL that is set to indicate these are experimental, but without a plan in mind on when and how we plan to remove such a flag, I worry that we are just swapping one flag out for a form of another. Is there something from preventing us from reaching a stable API? The ML-KEM experimental functions don't use such a flag for example, and are placed in a different header file.

jakemas:
moved to https://github.com/aws/aws-lc/pull/2096

PR #2083, opened 2024-12-28
Allow TLS PSK without server certificate

WillChilds-Klein:
### Description of changes: 

While we currently support TLS [pre-shared keys (PSKs)][1] for TLS 1.2, our
python integration tests have shown that we don't currently support TLS 1.2
PSKs on the server side unless a certificate has been configured. PSKs can be
used for both secret establishment and authentication, so there likely exist
legitimate use-cases for establishing PSK connections without certificates.
CPython's integration tests [expect this][2] behavior.

This gap appears to be incidental, not intentional. Unrelated [OCSP work][3]
introduced a requirement that a valid certificate public key is loaded on the
server before we sort out handshake parameters. If that load fails, the
hanshake is aborted before PSK can be negotiated for secret establishment. To
fix this, we simply check whether the server has a PSK callback enabled, and
allow the certificate loading to fail if so. We also handle the potential
nullity of the public key.

### Call-outs:

This PR only applies to TLSv1.2 PSK. TLSv1.3 moved PSK negotiation earlier in
the handshake to client/server hello's, so does utilize this code path.
Currently, we only support TLSv1.3 PSK for session resumption, _not "pure" PSK
used for initial shared secret establishment. I have [a branch][4] where I'm
implementing this, but it's non-trivial.

We won't be able to delete the TLSv1.3 PSK python test patch until we implement
"pure" TLSv1.3 PSK.

### Testing:
- CI

[1]: https://datatracker.ietf.org/doc/html/rfc4279
[2]: https://github.com/python/cpython/blob/aeb9b65aa26444529e4adc7d6e5b0d3dd9889ec2/Lib/test/test_ssl.py#L4395
[3]: https://github.com/aws/aws-lc/pull/1120/files#diff-2743f6406343b4e3d671a71af35cf44a36ad77cef1b6cfbc02f25dc732915ee9R790
[4]: https://github.com/WillChilds-Klein/aws-lc/tree/tls13-pure-psk

### DIFF
--- a/ssl/handshake_server.cc
+++ b/ssl/handshake_server.cc
@@ -793,11 +793,13 @@ static enum ssl_hs_wait_t do_select_certificate(SSL_HANDSHAKE *hs) {
     }
   }
 
-  // Load |hs->local_pubkey| from the cert prematurely. The certificate could be
-  // subject to change once we negotiate signature algorithms later. If it
-  // changes to another leaf certificate the server and client has support for,
-  // we reload it.
-  if (!ssl_handshake_load_local_pubkey(hs)) {
+  // Load |hs->local_pubkey| from the cert (if present) prematurely. The
+  // certificate could be subject to change once we negotiate signature
+  // algorithms later. If it changes to another leaf certificate the server and
+  // client has support for, we reload it. The public key may only be absent if
+  // PSK is enabled on the server, as indicated by presense of a callback.
+  if (!ssl_handshake_load_local_pubkey(hs) &&
+      !(hs->local_pubkey == nullptr && hs->config->psk_server_callback)) {
     return ssl_hs_error;
   }
 
--- a/ssl/ssl_privkey.cc
+++ b/ssl/ssl_privkey.cc
@@ -385,6 +385,7 @@ static bool ssl_public_key_rsa_pss_check(EVP_PKEY *pubkey, uint16_t sigalg) {
 
 static bool tls12_pkey_supports_cipher_auth(SSL_HANDSHAKE *hs,
                                             const EVP_PKEY *key) {
+  GUARD_PTR(key);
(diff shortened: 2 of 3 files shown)


ssl/handshake_server.cc
```diff
@@ -793,11 +793,13 @@ static enum ssl_hs_wait_t do_select_certificate(SSL_HANDSHAKE *hs) {
...
-  // changes to another leaf certificate the server and client has support for,
-  // we reload it.
-  if (!ssl_handshake_load_local_pubkey(hs)) {
+  // Load |hs->local_pubkey| from the cert (if present) prematurely. The
+  // certificate could be subject to change once we negotiate signature
+  // algorithms later. If it changes to another leaf certificate the server and
+  // client has support for, we reload it. The public key may only be absent if
+  // PSK is enabled on the server, as indicated by presense of a callback.
+  if (!ssl_handshake_load_local_pubkey(hs) &&
+      !(hs->local_pubkey == nullptr && hs->config->psk_server_callback)) {
```
smittals2, line 802: Is the check for `hs->local_pubkey == nullptr` not redundant here?

WillChilds-Klein: Sort of. You're right that the call to `ssl_handshake_load_local_pubkey` [will check](https://github.com/aws/aws-lc/blob/697acc6616736ad07539fda1e0726cc043e1097a/ssl/ssl_cert.cc#L810) `hs->local_pubkey`'s nullity, but that's an implementation detail. I check it again here 1.) for clarity 2.) to preserve this logic if `ssl_handshake_load_local_pubkey`'s implementation is changed in the future.

justsmth (approved):

smittals2 (approved):

PR #2084, opened 2024-12-31
SHA3 and SHAKE - New API Design

manastasova:
### Issues:
Resolves #CryptoAlg-2810

### Description of changes: 
AWS-LC supports SHA3 and SHAKE algorithms though low level SHA3_init, SHA3_Update, SHA3_Final and SHAKE_init, SHAKE_Final APIs. Currently, there are two issues with the implementation and usage of SHA3 and SHAKE:

This PR addresses both of them as follows:
- Introduce a new API layers - Keccak1600, FIPS202, SHA3 and SHAKE. 
   - _Keccak1600_ layer implements KeccakF1600 Absorb and Squeeze functions; Keccak1600 layer does _not_ manage internal input/output buffers. 
   - _FIPS202_ layer implements Reset, Init, Update, and Finalize functionalities; FIPS202 layer manages the internal input/output buffers, allowing incremental requests (not necessarily multiple of block size) to Update (Absorb) and Squeeze for input/output processing. (Other functionalities, such as zero-ing of bitstate, block size checks, etc. are also handled by FIPS202 API layer)
   - _SHA3_ layer implements Init, Update, and Final functionalities; SHA3 layer only implements SHA3 algorithm, thus, offers a single-call SHA3_Final function. SHA3_Final will not update internal flags related to XOF functionalities.
   - _SHAKE_ layer implements Init, Update, and Finalize and Squeeze functionalities; SHAKE layer implements XOF SHAKE algorithm, therefore, offers a single-call SHAKE_Finalize function that finilized the absorb phase and initializes the squeeze phase. SHAKE_Final will process the first output value (based on the requested output length). Incremental XOF output generation is handled by SHAKE_Squeeze. SHAKE_Squeeze can be called multiple time. SHAKE_Finalize must be called before first call to SHAKE_Squeeze.
- SHA3 digest generation: SHA3_Init; SHA3_Update; SHA3_Final;
- SHAKE (incremental) output generation: SHAKE_Init; SHAKE_Update; SHAKE_Finalize; (SHAKE_Squeeze*);

### Call-outs:
Service indicator is updated: 
- Inside SHA3 and SHAKE single shot APIs (as previously in AWS-LC);
- Inside SHA3_Final (as previously in AWS-LC);
- _Not updated_ Inside SHAKE_Finalize (Incremental XOF output generation does not allow to define the _last_ SHAKE call) 

All other algorithms that use SHA3/SHAKE APIs are updated:
- ML-KEM (SHA3/SHAKE calls will be inlined, thus, additional XOF flag checks and branching will be avoided)
- ML-DSA (SHAKE_Squeeze (incremental XOF output functionality) inside ML-DSA is never invoked with the KAT test vectors and gtests)

### Testing:


crypto/fipsmodule/sha/sha3.c
```diff
@@ -109,62 +109,39 @@ uint8_t *SHAKE256(const uint8_t *data, const size_t in_len, uint8_t *out, size_t
...
   size_t block_size = ctx->block_size;
   size_t num, rem;
 
-  if (len == 0) {
-    return 1;
-  }
-
+  // Case |len| equals 0 is checked in SHA3/SHAKE higher level APIs
   // Process intermediate buffer.
   num = ctx->buf_load;
```
andrewhop, line 146: If FIPS202_Update isn't static we can't know for sure all callers of it will pass in valid arguments so it also needs to check that ctx is not null.

crypto/fipsmodule/sha/sha3.c
```diff
@@ -202,31 +179,133 @@ int SHA3_Update(KECCAK1600_CTX *ctx, const void *data, size_t len) {
...
+  if (ctx == NULL) {
+    return 0;
+  }
+
+  if (len == 0) {
+    return 1;
+  }
+
+  return FIPS202_Update(ctx, data, len);
+}
```
andrewhop, line 227: I know this was the plan but it looks a little weird that SHA3_update and SHAKE_Absorb are the exact same code, will they ever not be the same?

crypto/fipsmodule/sha/sha3.c
```diff
@@ -202,31 +179,133 @@ int SHA3_Update(KECCAK1600_CTX *ctx, const void *data, size_t len) {
...
+
+// SHA3_Final should be called once to process final digest value
+// |ctx->padded| flag does not need to be updated
```
andrewhop, line 230: This is a change from before, why doesn't it need to be updated now?

crypto/fipsmodule/sha/sha3.c
```diff
@@ -202,31 +179,133 @@ int SHA3_Update(KECCAK1600_CTX *ctx, const void *data, size_t len) {
...
+    if (FIPS202_Finalize(md, ctx) == 0) {
       return 0;
     }
   }
 
-  SHA3_Squeeze(ctx->A, md, ctx->md_size, block_size, ctx->padded);
+  Keccak1600_Squeeze(ctx->A, md, len, ctx->block_size, ctx->padded);
   ctx->padded = 1;
 
   FIPS_service_indicator_update_state();
```
andrewhop, line 309: I don't think we want to add the service indicator to SHAKE_Squeeze right now, because you've updated ML-DSA/ML-KEM to use SHAKE_squeeze this will mark it as approved. We will need to lock/unlock in ML-DSA/KEM and then have it update the service indicator if it is operating in an approved manor. 

Previously ML-DSA/KEM called SHA3_Update which doesn't update the indicator it was fine. In a follow up PR when you add the service indicator for SHAKE you'll need to add more tests to all users of SHAKE_Squeeze. I'm also a little surprised the existing service indicator tests for ML-KEM didn't catch this.

crypto/fipsmodule/sha/sha3.c
manastasova: The padded flag was introduced when the (block-wise) incremental XOF functionality was added. This flag _should_ be only needed for SHAKE, since SHA3_Final `should not` be called multiple times.
I actually want to change the flag to `state` and add it to all SHA3 and SHAKE functions. ctx->state will be {KECCAK1600_STATE_ABSORB (not_padded), KECCAK1600_STATE_SQUEEZE (padded), KECCAK1600_STATE_FINAL(no_more_squeezes_allowed)}, where SHA3 will move from KECCAK1600_STATE_ABSORB directly to KECCAK1600_STATE_FINAL, preventing from incremental calls to SHA3_Final through `EVP_DigestFinal`.
The new flag will also allow to define `EVP_DigestFinalXOF` as a signle-shot API (similar to ossl) with no incremental squeezes after it. Instead, `EVP_DigestSqueeze` should be only used for incremental XOF.

PR #2086, opened 2024-12-31
alignas(16) unsupported w/ GCC 7.2 for ARM32

justsmth:
### Issues:
* Addresses: CryptoAlg-2718

### Description of changes: 
* When targeting arm, GCC 7.2 claims to not support `alignas(16)` in C++ code.
```
...
.../arm-unknown-linux-musleabi/crypto/cipher_extra/aead_test.cc:835:54: error: requested alignment 16 is larger than 8 [-Werror=attributes]
   alignas(16) uint8_t key[EVP_AEAD_MAX_KEY_LENGTH + 1];
                                                      ^
/home/justsmth/workspace/CrossAWS-LC-FIPS/src/CrossAWS-LC-FIPS/build/private/buildroot/arm-unknown-linux-musleabi/crypto/cipher_extra/aead_test.cc:836:58: error: requested alignment 16 is larger than 8 [-Werror=attributes]
   alignas(16) uint8_t nonce[EVP_AEAD_MAX_NONCE_LENGTH + 1];
                                                          ^
/home/justsmth/workspace/CrossAWS-LC-FIPS/src/CrossAWS-LC-FIPS/build/private/buildroot/arm-unknown-linux-musleabi/crypto/cipher_extra/aead_test.cc:837:39: error: requested alignment 16 is larger than 8 [-Werror=attributes]
   alignas(16) uint8_t plaintext[32 + 1];
...
```

### Call-outs:
* We also use `alignas(16)` a few places in C code, but the compiler doesn't complain about those.
* Compilation of AWS-LC FIPS with this same compiler/target still fails with this change, but for an apparently unrelated reason.

### DIFF
--- a/crypto/cipher_extra/aead_test.cc
+++ b/crypto/cipher_extra/aead_test.cc
@@ -831,11 +831,17 @@ TEST_P(PerAEADTest, AliasedBuffers) {
   EXPECT_EQ(Bytes(kPlaintext), Bytes(in, out_len));
 }
 
+#if defined(__BIGGEST_ALIGNMENT__)
+#define UNALIGNED_TEST_ALIGNMENT __BIGGEST_ALIGNMENT__
+#else
+#define UNALIGNED_TEST_ALIGNMENT 8
+#endif // defined(__BIGGEST_ALIGNMENT__)
+
 TEST_P(PerAEADTest, UnalignedInput) {
-  alignas(16) uint8_t key[EVP_AEAD_MAX_KEY_LENGTH + 1];
-  alignas(16) uint8_t nonce[EVP_AEAD_MAX_NONCE_LENGTH + 1];
-  alignas(16) uint8_t plaintext[32 + 1];
-  alignas(16) uint8_t ad[32 + 1];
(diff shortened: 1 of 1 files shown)


crypto/cipher_extra/aead_test.cc
```diff
@@ -831,6 +831,7 @@ TEST_P(PerAEADTest, AliasedBuffers) {
   EXPECT_EQ(Bytes(kPlaintext), Bytes(in, out_len));
 }
 
+#if !defined(__BIGGEST_ALIGNMENT__) || __BIGGEST_ALIGNMENT__ >= 16
 TEST_P(PerAEADTest, UnalignedInput) {
   alignas(16) uint8_t key[EVP_AEAD_MAX_KEY_LENGTH + 1];
```
andrewhop, line 836: Could this test use alignas(8) and still cover the case of unaligned input because the test uses all the pointers + 1 to get unaligned like  `ciphertext + 1`?

justsmth: You're right. I changed the test to just use the `__BIGGEST_ALIGNMENT__` value if available.

andrewhop (approved):

torben-hansen (approved):

PR #2089, opened 2024-12-31
FreeBSD and OpenBSD fork detection

torben-hansen:
### Description of changes: 

Two things are happening in this PR:
- Refactor code in `fork_detect.c` to make is more readable and amendable
- Add new sufficient method for fork detection on FreeBSD and OpenBSD.

The new method uses `minherit`, tagging a memory page with a special value. The semantics are equivalent to the existing case with `madvise`. 

### Testing:

We still need some more testing to increase test coverage for fork detection. This is an upcoming task.

### DIFF
--- a/crypto/ube/fork_detect.c
+++ b/crypto/ube/fork_detect.c
@@ -14,6 +14,17 @@
 
 #include <openssl/target.h>
 
+// Methods to detect fork events aren't generally portable over our supported
+// platforms. Fork detection is therefore an opt-in. Capture the opt-in logic
+// below that categorizes a platform targets as either having
+//  1) fork detection support,
+//  2) forking doesn't exist, and
+//  3) no fork detection support.
+// For (1) we implement sufficient methods for detecting fork events. For (2),
+// since forking is not a semantic that exists for the platform, we do not need
+// to do anything. Except fake that fork detection is supported. For (3), we
+// can't do anything. In this case randomness generation falls back to
+// randomizing the state per-request.
 #if defined(OPENSSL_LINUX)
   #define AWSLC_FORK_DETECTION_SUPPORTED
   #if !defined(_GNU_SOURCE)
@@ -47,12 +58,12 @@ static struct CRYPTO_STATIC_MUTEX fork_detect_lock = CRYPTO_STATIC_MUTEX_INIT;
 // assume that it has exclusive access to it.
 static volatile char *fork_detect_addr = NULL;
 static uint64_t fork_generation = 0;
-static int ignore_madv_wipeonfork = 0;
-static int ignore_madv_inheritzero = 0;
+static int ignore_wipeonfork = 0;
+static int ignore_inheritzero = 0;
 
 static int ignore_all_fork_detection(void) {
-  if (ignore_madv_wipeonfork == 1 &&
-      ignore_madv_inheritzero == 1) {
+  if (ignore_wipeonfork == 1 &&
+      ignore_inheritzero == 1) {
     return 1;
   }
   return 0;
@@ -69,7 +80,7 @@ OPENSSL_STATIC_ASSERT(MADV_WIPEONFORK == 18, MADV_WIPEONFORK_is_not_18)
 #define MADV_WIPEONFORK 18
 #endif
 
-static int init_fork_detect_madv_wipeonfork(void *addr, long page_size) {
+static int init_fork_detect_wipeonfork(void *addr, long page_size) {
 
   GUARD_PTR(addr);
 
@@ -88,7 +99,7 @@ static int init_fork_detect_madv_wipeonfork(void *addr, long page_size) {
 
 #else // defined(OPENSSL_LINUX)
 
-static int init_fork_detect_madv_wipeonfork(void *addr, long page_size) {
+static int init_fork_detect_wipeonfork(void *addr, long page_size) {
   return 0;
 }
 
@@ -100,7 +111,7 @@ static int init_fork_detect_madv_wipeonfork(void *addr, long page_size) {
 #include <sys/mman.h>
 #include <unistd.h>
 
-static int init_fork_detect_madv_inheritzero(void *addr, long page_size) {
+static int init_fork_detect_inheritzero(void *addr, long page_size) {
 
   GUARD_PTR(addr);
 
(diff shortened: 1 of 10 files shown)


crypto/ube/fork_detect.c
```diff
@@ -12,29 +12,30 @@
...
-#include <stdlib.h>
+  #define AWSLC_FORK_DETECTION_SUPPORTED
+  #if !defined(_GNU_SOURCE)
+    #define _GNU_SOURCE  // needed for madvise() and MAP_ANONYMOUS.
+  #endif
+#elif defined(OPENSSL_FREEBSD) || defined(OPENSSL_OPENBSD)
+  /* FreeBSD requires POSIX compatibility off for its syscalls (enables __BSD_VISIBLE)
+   * Without the below line, <sys/mman.h> cannot be imported (it requires __BSD_VISIBLE) */
+  #undef _POSIX_C_SOURCE
+#elif defined(OPENSSL_WINDOWS) || defined(OPENSSL_TRUSTY)
```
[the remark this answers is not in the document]
justsmth: Likely not in scope, but NetBSD also supports `MAP_INHERIT_ZERO` since at least 7.0: https://man.netbsd.org/NetBSD-7.0/minherit.2

crypto/ube/fork_detect.c
```diff
@@ -166,4 +237,4 @@ uint64_t CRYPTO_get_fork_generation(void) { return 0xc0ffee; }
 // every RAND_bytes call.
 uint64_t CRYPTO_get_fork_generation(void) { return 0; }
```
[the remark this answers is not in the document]
torben-hansen: probably inherited from `base.h`. But I added an explicit dependency on it.

crypto/ube/fork_detect.c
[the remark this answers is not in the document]
torben-hansen: I don't believe we have any NetBSD in the CI. I would probably want that before trying to add something this non-portable.

> Where is AWSLC_FORK_DETECTION_SUPPORTED defined for these BSDs?

Missed it :)

crypto/ube/fork_detect.c
```diff
@@ -44,25 +62,29 @@ static struct CRYPTO_STATIC_MUTEX fork_detect_lock = CRYPTO_STATIC_MUTEX_INIT;
...
 
-  page_size = sysconf(_SC_PAGESIZE);
-  if (page_size <= 0) {
-    return 0;
-  }
+#if defined(MADV_WIPEONFORK)
+OPENSSL_STATIC_ASSERT(MADV_WIPEONFORK == 18, MADV_WIPEONFORK_is_not_18)
+#else
+#define MADV_WIPEONFORK 18
+#endif
```
justsmth, line 83: * If `MADV_WIPEONFORK` were not already defined in the headers provided by the system, why do we think it's safe to define the value here?
* Since it's a syscall parameter, the value of `MADV_WIPEONFORK` should be stable for a given target, but how do we know that the value is 18 across all Linux targets?
    * Possibly related: `MADV_WIPEONFORK` is not supported by linux kernels prior to v4.14. There are older kernels (like v4.4) that are still actively supported (till January 2027). 
* Should the macro guard for this section instead be something like:
```
#if defined(OPENSSL_LINUX) && defined(MADV_WIPEONFORK)
```

crypto/ube/fork_detect.c
```diff
@@ -71,29 +93,112 @@ static int init_fork_detect_madv_wipeonfork(void **addr_out) {
...
+
+#if defined(OPENSSL_FREEBSD) || defined(OPENSSL_OPENBSD)
+
+#include <sys/mman.h>
+#include <unistd.h>
+
+// Sometimes (for example, on FreeBSD) MAP_INHERIT_ZERO is called INHERIT_ZERO
+#if !defined(MAP_INHERIT_ZERO) && defined(INHERIT_ZERO)
+    #define MAP_INHERIT_ZERO INHERIT_ZERO
+#endif
```
justsmth, line 119: NP: Would it be safer to define our own macro with a value specific the target OS:
```
#if defined(OPENSSL_FREEBSD)
    #define AWS_LC_MAP_INHERIT_ZERO INHERIT_ZERO
#else // defined(OPENSSL_OPENBSD)
    #define AWS_LC_MAP_INHERIT_ZERO MAP_INHERIT_ZERO
#endif
```

crypto/ube/fork_detect.c
```diff
@@ -71,29 +93,112 @@ static int init_fork_detect_madv_wipeonfork(void **addr_out) {
...
-  if (ignore_madv_wipeonfork == 1) {
+  page_size = sysconf(_SC_PAGESIZE);
+  if (page_size <= 0) {
     return;
   }
 
-  if (init_fork_detect_madv_wipeonfork(&addr) != 1) {
+  addr = mmap(NULL, (size_t)page_size, PROT_READ | PROT_WRITE,
+                    MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+  if (addr == MAP_FAILED) {
```
justsmth, line 183: NP: You could check for `NULL` here, then not need the `GUARD_PTR` macro in the static functions above. (I don't think it's even possible for `mmap` to return NULL.)

crypto/ube/fork_detect.c
```diff
@@ -71,29 +93,112 @@ static int init_fork_detect_madv_wipeonfork(void **addr_out) {
...
+}
+
+#endif // defined(OPENSSL_FREEBSD) || defined(OPENSSL_OPENBSD)
+
+// We assume that a method in this function is sufficient to detect fork events.
+// Returns 1 if a method sufficient for fork detection successfully
+// initializes. Otherwise returns 0.
+static int init_fork_detect_methods_sufficient(void *addr, long page_size) {
+
+  if (ignore_wipeonfork != 1 &&
```
justsmth, line 145: NP: Should the `ignore_wipeonfork` and `ignore_inheritzero` value checks be inside their respective `init_fork_detect_XXX` functions?

tests/ci/run_benchmark_build_tests.sh
```diff
@@ -82,7 +82,7 @@ build_openssl_no_debug $openssl_3_2_branch
 build_openssl_no_debug $openssl_master_branch
 build_boringssl
 
-run_build -DASAN=1 -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_STANDARD=14 -DCMAKE_C_STANDARD=11 -DENABLE_DILITHIUM=ON -DBENCHMARK_LIBS="\
+run_build -DASAN=1 -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_STANDARD=17 -DCMAKE_C_STANDARD=11 -DENABLE_DILITHIUM=ON -DBENCHMARK_LIBS="\
```
justsmth, line 85: NP: Unrelated change? Why?

crypto/ube/fork_detect.c
torben-hansen: It's the existing behaviour and it has worked flawlessly.
It's just to increase support for the feature that we attempt to use a hard-coded value in case the define doesn't exist.
```
madvise(addr, (size_t)page_size, MADV_WIPEONFORK)
```
will fail if the feature is not supported.

crypto/ube/fork_detect.c
torben-hansen: The current form is a bit shorter. It has worked flawlessly in s2n-tls where I first used this pattern.

crypto/ube/fork_detect.c
torben-hansen: A matter of taste. Same functionality.

crypto/ube/fork_detect.c
torben-hansen: The guard is probably not necessary. But it was requested in a previous PR :)

tests/ci/run_benchmark_build_tests.sh
torben-hansen: Had to cherry-pick https://github.com/aws/aws-lc/pull/2089/commits/b549c4a92f66f2fe422b9b6ac93fcda7f1706d0b
Testing was borked without.

justsmth (approved):

smittals2 (approved):

PR #2092, opened 2025-01-02
Refactor TLS 1.3 cipher selection and fix SSL_get_ciphers

smittals2:
### Issues:
`CryptoAlg-2559`

### Description of changes: 
1. `SSL_CTX_new` now configures `ctx->tls13_cipher_suites` with default TLS 1.3 ciphers (did not previously). The default preference order  for TLS 1.3 ciphersuites differs based on whether AES Hardware acceleration is enabled in AWS-LC, we match this behavior in SSL_CTX_new. 
2. The `SSL_CONFIG` object holds it's own TLS 1.3 ciphersuites now
3. Server side and client side cipher selection logic is modified to account for the new field in `SSL_CONFIG`
4. Each time a cipher list is configured either on `SSL_CTX` or `SSL_CONFIG`, the main `cipher_list` is updated with values from its own `tls13_cipher_suites` via a new function `update_cipher_list`. This function also updates `in_group_flags` accordingly.  This means, TLS 1.3 ciphersuites are stored seperately while the main `cipher_list` var holds ALL ciphersuites in both objects
5. Outdated comments were updated

### Call-outs:
Note `SSL_CTX` generally has a longer lifetime than an `SSL` object. `SSL_CONFIG` objects live on `SSL` objects. Both `SSL_CTX` and `SSL_CONFIG` have their own `cipher_list` and `tls13_cipher_list` fields. We support `SSL_[CTX]_set_...` APIs to set these fields. 

In general, our TLS implementations pick cipher suites against the given connection constraints/parameters. Adding TLS 1.3 ciphersuites to `ctx->cipher_list` or `config->cipher_list` does not modify this behavior or introduce new problems. 

### Testing:
Added new testing to validate cipher selection behavior between SSL_CTX and SSL_CONFIG objects
Added change detection test for TLS 1.3 Client Hello
Updated old tests and to-do's regarding TLS 1.3
Testing to ensure `SSL_get_ciphers` returns ALL ciphersuites instead of TLS 1.2 and below. 


justsmth (approved):

include/openssl/ssl.h
```diff
@@ -1690,15 +1690,26 @@ OPENSSL_EXPORT size_t SSL_get_all_standard_cipher_names(const char **out,
...
+// configuration when AES hardware acceleration is enabled.
+#define TLS13_DEFAULT_CIPHER_LIST_AES_HW "TLS_AES_128_GCM_SHA256:" \
+                                          "TLS_AES_256_GCM_SHA384:" \
+                                          "TLS_CHACHA20_POLY1305_SHA256"
+
+// TLS13_DEFAULT_CIPHER_LIST_NO_AES_HW is the default TLS 1.3 cipher suite
+// configuration when no AES hardware acceleration is enabled.
+#define TLS13_DEFAULT_CIPHER_LIST_NO_AES_HW "TLS_CHACHA20_POLY1305_SHA256:" \
+                                            "TLS_AES_128_GCM_SHA256:" \
+                                            "TLS_AES_256_GCM_SHA384"
```
andrewhop, line 1711: Do these need to be in the public header? Do we expect customer to pass these back to us? Also how did you come up with these options?

ssl/ssl_lib.cc
```diff
@@ -586,7 +586,16 @@ SSL_CTX *SSL_CTX_new(const SSL_METHOD *method) {
...
+  const bool has_aes_hw = EVP_has_aes_hardware();
+  const char *cipher_rule;
+  if (has_aes_hw) {
+    cipher_rule = TLS13_DEFAULT_CIPHER_LIST_AES_HW;
+  } else {
+    cipher_rule = TLS13_DEFAULT_CIPHER_LIST_NO_AES_HW;
+  }
+
+  if (!SSL_CTX_set_ciphersuites(ret.get(), cipher_rule) ||
+      !SSL_CTX_set_strict_cipher_list(ret.get(), SSL_DEFAULT_CIPHER_LIST) ||
```
andrewhop, line 598: Is this changing our behavior? From the PR description:
> SSL_CTX_new now configures ctx->tls13_cipher_suites with default TLS 1.3 ciphers (did not previously).

What did we do before if you attempted to connect to a TLS 1.3 enabled endpoint?

ssl/ssl_cipher.cc
```diff
@@ -1234,7 +1234,88 @@ static bool is_known_default_alias_keyword_filter_rule(const char *rule,
   return false;
 }
 
-bool ssl_create_cipher_list(UniquePtr<SSLCipherPreferenceList> *out_cipher_list,
+// update_cipher_list updates |ctx->cipher_list| by:
+// 1. Removing any existing TLS 1.3 ciphersuites
+// 2. Adding configured ciphersuites from |ctx->tls13_cipher_list|
+// 3. Configuring a new |ctx->cipher_list->in_group_flags|
+// This function maintains the ordering of ciphersuites and places TLS 1.3
+// ciphersuites at the front of the list.
```
andrewhop, line 1242: Does the order of TLS 1.3 and the < 1.3 ciphersuites affect what gets negotiated?

ssl/ssl_cipher.cc
```diff
@@ -1234,7 +1234,88 @@ static bool is_known_default_alias_keyword_filter_rule(const char *rule,
...
+    tmp_cipher_list.reset(sk_SSL_CIPHER_dup(ctx->cipher_list->ciphers.get()));
+  } else {
+    tmp_cipher_list.reset(sk_SSL_CIPHER_new_null());
+  }
+
+  if (tmp_cipher_list == NULL) {
+    return 0;
+  }
+
+  // Delete any existing TLSv1.3 ciphersuites. These will be first in the list
```
andrewhop, line 1259: Is this enforced by AWS-LC or could a customer pass in a list with the wrong order?

ssl/ssl_cipher.cc
```diff
@@ -1234,7 +1234,88 @@ static bool is_known_default_alias_keyword_filter_rule(const char *rule,
   return false;
 }
 
-bool ssl_create_cipher_list(UniquePtr<SSLCipherPreferenceList> *out_cipher_list,
+// update_cipher_list updates |ctx->cipher_list| by:
```
andrewhop, line 1237: This comment explains what this function does, but why do we need to remove everything and then create a new list?

smittals2: we don't necessarily know whether ctx->cipher_list is getting updated or ctx->tls13_cipher_list. When ctx->tls13_cipher_list is updated, we need to remove old TLS 1.3 ciphersuites from ctx->cipher_list and add the new ones in. Note, we are only removing and adding TLS 1.3 ciphersuites here, not the entire list.

ssl/ssl_lib.cc
smittals2: In a TLS 1.3 connection they would still be written, but the logic for this was hardcoded in the tls 1.3 state machine. In the function ssl_write_client_cipher_list, TLS 1.3 ciphers are handled directly. This PR moves that logic earlier in the lifecycle to SSL_CTX initialization.

ssl/ssl_cipher.cc
smittals2: From my understanding, there is no conflict in between 1.2 and 1.3 ciphersuites (i.e. doesn't matter if 1.3 ciphersuites are placed in front of 1.2). In a given TLS version's ciphersuites, ordering does matter. This PR doesn't change that though. TLS 1.2 cipher ordering is left as is. And when TLS 1.3 ciphersuites are added to cipher_list, we make sure to maintain configured order.

ssl/ssl_cipher.cc
smittals2: SSL_CTX is an opaque struct. Consumers can only set ciphersuites via the relvant setter functions which require a rule str to be passed in. We enforce ordering and contents otherwise.

ssl/ssl_cipher.cc
```diff
@@ -1367,6 +1448,9 @@ bool ssl_create_cipher_list(UniquePtr<SSLCipherPreferenceList> *out_cipher_list,
     return false;
   }
 
+  // Update |ctx->cipher_list| with any ciphers in |ctx->tls13_cipher_list|
+  update_cipher_list(ctx);
```
justsmth, line 1452: This is ignoring the return value.

include/openssl/ssl.h
```diff
@@ -1714,7 +1713,8 @@ OPENSSL_EXPORT int SSL_CTX_set_strict_cipher_list(SSL_CTX *ctx,
 // behavior is strongly advised against and only meant for OpenSSL
 // compatibility.
 //
-// Note: this API only sets the TLSv1.2 and below ciphers.
+// Note: this API only sets the TLSv1.2 and below ciphers. If unchanged,
+// the default TLS 1.3 ciphersuites are pulled in as well.
```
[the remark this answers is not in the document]
smittals2: Just removed it from here. That internal detail is explained where the cipher_list var is defined. Consumers don't need to know that as a part of the contract for this function.

ssl/handshake_client.cc
```diff
@@ -259,49 +260,31 @@ static bool ssl_write_client_cipher_list(const SSL_HANDSHAKE *hs, CBB *out,
     return false;
   }
 
-  // Add TLS 1.3 ciphers.
-  if (hs->max_version >= TLS1_3_VERSION && ssl->ctx->tls13_cipher_list) {
-    // Use the configured TLSv1.3 ciphers list.
-    STACK_OF(SSL_CIPHER) *ciphers = ssl->ctx->tls13_cipher_list->ciphers.get();
+  // Add all ciphers unless TLS 1.3 only connection
+  if (hs->min_version < TLS1_3_VERSION && type != ssl_client_hello_inner) {
```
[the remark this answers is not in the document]
smittals2: nope, we enfore TLS 1.3 ciphersuites being at the front of the list at all times

ssl/handshake_client.cc
```diff
@@ -259,49 +260,31 @@ static bool ssl_write_client_cipher_list(const SSL_HANDSHAKE *hs, CBB *out,
...
+  } else if (hs->max_version >= TLS1_3_VERSION && ssl->ctx->tls13_cipher_list) {
+    // Only TLS 1.3 ciphers
+    STACK_OF(SSL_CIPHER) *ciphers = ssl->ctx->tls13_cipher_list->ciphers.get();
     bool any_enabled = false;
-    if (!collect_cipher_protocol_ids(SSL_get_ciphers(ssl), &child, mask_k,
+
+    if (!collect_cipher_protocol_ids(ciphers, &child, mask_k,
       mask_a, hs->max_version, hs->min_version, &any_enabled)) {
       return false;
     }
```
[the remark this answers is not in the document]
smittals2: Cannot combine the two blocks. Different cases for ECH

ssl/handshake_client.cc
```diff
@@ -259,49 +260,31 @@ static bool ssl_write_client_cipher_list(const SSL_HANDSHAKE *hs, CBB *out,
...
-  if (hs->max_version >= TLS1_3_VERSION && ssl->ctx->tls13_cipher_list) {
-    // Use the configured TLSv1.3 ciphers list.
-    STACK_OF(SSL_CIPHER) *ciphers = ssl->ctx->tls13_cipher_list->ciphers.get();
+  // Add all ciphers unless TLS 1.3 only connection
+  if (hs->min_version < TLS1_3_VERSION && type != ssl_client_hello_inner) {
     bool any_enabled = false;
-    if (!collect_cipher_protocol_ids(ciphers, &child, mask_k,
+    if (!collect_cipher_protocol_ids(SSL_get_ciphers(ssl), &child, mask_k,
       mask_a, hs->max_version, hs->min_version, &any_enabled)) {
       return false;
```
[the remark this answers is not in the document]
smittals2: Yes if min is less than 1.3, we want to add min to max (where max could also be anything). We pass this in to collect_cipher_protocol_ids. 

SSL_get_ciphers will return the SSL level (if configured) and otherwise SSL_CTX main cipher list which now holds all ciphersuites.

include/openssl/ssl.h
```diff
@@ -1690,10 +1690,9 @@ OPENSSL_EXPORT size_t SSL_get_all_standard_cipher_names(const char **out,
 // Once an equal-preference group is used, future directives must be
```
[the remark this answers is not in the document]
smittals2: The current documentation covers this case ("cipher list for |ssl|") but I'll make it more explicit. We just weren't living up to this contract initially.

ssl/ssl_cipher.cc
```diff
@@ -1234,6 +1234,86 @@ static bool is_known_default_alias_keyword_filter_rule(const char *rule,
...
+    }
+  }
+
+
+  if (!updated_in_group_flags.Init(num_added_tls13_ciphers + num_updated_tls12_ciphers)) {
+    return 0;
+  }
+  std::fill(updated_in_group_flags.begin(), updated_in_group_flags.end(), false);
+
+  // Copy in_group_flags from |ctx->tls13_cipher_list|
```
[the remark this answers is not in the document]
smittals2: https://github.com/aws/aws-lc/blob/81f138a0632151e24ecbf3e0b08c37d6f82f57ed/ssl/internal.h#L660

ssl/ssl_cipher.cc
```diff
@@ -1234,6 +1234,86 @@ static bool is_known_default_alias_keyword_filter_rule(const char *rule,
...
+  if (tls13_ciphers && tls13_ciphers->in_group_flags) {
+    const auto& tls13_flags = tls13_ciphers->in_group_flags;
+    // Ensure value of last element in |in_group_flags| is 0. The last cipher
+    // in a list must be the end of any group in that list.
+    if (tls13_flags[num_added_tls13_ciphers - 1] != 0) {
+      tls13_flags[num_added_tls13_ciphers - 1] = false;
+    }
```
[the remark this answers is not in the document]
smittals2: No, you can only group ciphersuites within the same version

ssl/handshake_client.cc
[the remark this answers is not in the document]
andrewhop: I agree you can't combine the two blocks into one shared block, but I'm still curious about the filtering, can this just use `ciphers` from up above which only includes 1.3 ciphers?

ssl/ssl_lib.cc
```diff
@@ -2178,17 +2187,25 @@ const char *SSL_get_cipher_list(const SSL *ssl, int n) {
...
                                                : EVP_has_aes_hardware();
-  return ssl_create_cipher_list(&ctx->cipher_list, has_aes_hw, str,
+  if (!ssl_create_cipher_list(&ctx->cipher_list, has_aes_hw, str,
                                 false /* not strict */,
-                                false /* don't configure TLSv1.3 ciphers */);
+                                false /* don't configure TLSv1.3 ciphers */)) {
+    return 0;
+  }
+
+  return update_cipher_list(ctx->cipher_list, ctx->cipher_list, ctx->tls13_cipher_list);
```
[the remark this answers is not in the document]
smittals2: The same ssl_create_cipher_list function is used to create tls13_cipher_lists and the main cipher_list. Therefore, I have chosen to do update's separately to properly distinguish what is being set.

ssl/handshake_client.cc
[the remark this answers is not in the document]
smittals2: Are you saying use the same SSL_get_ciphers call below as well? 

The reason we can't do this is if you are a ECH ClientHelloInner, and you support TLS versions less than 1.3, by RFC we still only send TLS 1.3 ciphersuites. Since collect_ciphers takes min max version from the handshake, using the above ciphers var would erroneously pull in TLS 1.2 and below ciphersuites. Therefore, we specifically pass in only the tls13_ciphersuites in the else block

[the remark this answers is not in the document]
andrewhop: In this block we only want to send 1.3 ciphers, but isn't that true before any filtering due to using `ciphers = ssl->ctx->tls13_cipher_list`? What is `collect_cipher_protocol_ids` filtering if we know `hs->max_version = 1.3`?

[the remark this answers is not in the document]
smittals2: Versions is one criteria, apparently we have masks to check for disabled algorithms also. That is filtered here too. Documentation for the collect_cipher_protocol_ids: 

> // collect_cipher_protocol_ids uses |cbb| to collect IANA-assigned numbers
// of |ciphers|. |mask_a|, |mask_k|, |max_version| and |min_version|
// are used to filter the |ciphers|. |any_enabled| will be true if not all
// ciphers are filtered out.
// It returns true when success. It returns false otherwise.

ssl/ssl_lib.cc
```diff
@@ -586,7 +586,16 @@ SSL_CTX *SSL_CTX_new(const SSL_METHOD *method) {
     return nullptr;
   }
 
-  if (!SSL_CTX_set_strict_cipher_list(ret.get(), SSL_DEFAULT_CIPHER_LIST) ||
+  const bool has_aes_hw = EVP_has_aes_hardware();
```
samuel40791765, line 589: I understand that we do the override detection in `SSL_CTX_set_ciphersuites`, but should we add it here as well?
It seems like that's [the general pattern](https://github.com/search?q=repo%3Aaws%2Faws-lc%20EVP_has_aes_hardware&type=code) of how `has_aes_hw` is written in the SSL layer. It would also make it more obvious what TLS13 ciphersuites are being set here at first glance (if the override is being set).

ssl/internal.h
```diff
@@ -3244,9 +3258,14 @@ struct SSL_CONFIG {
...
   X509_VERIFY_PARAM *param = nullptr;
 
-  // crypto
+  // cipher_list holds all available cipher suites for tls 1.3,
+  // and 1.2 and below
   UniquePtr<SSLCipherPreferenceList> cipher_list;
 
+  // tls13_cipher_list holds the default or configured tls1.3 and above
+  // cipher suites.
+  UniquePtr<SSLCipherPreferenceList> tls13_cipher_list;
```
samuel40791765, line 3267: Should we document that these take precedence over `SSL_CTX`'s version of the same variables?

ssl/tls13_server.cc
```diff
@@ -112,7 +112,13 @@ static int ssl_ext_supported_versions_add_serverhello(SSL_HANDSHAKE *hs,
 
 static const SSL_CIPHER *choose_tls13_cipher(const SSL *ssl) {
   STACK_OF(SSL_CIPHER) *tls13_ciphers = nullptr;
-  if (ssl->ctx->tls13_cipher_list &&
+  // First check config, otherwise fallback to ctx, and otherwise let client
+  // preference dictate cipher selection
```
samuel40791765, line 116:
```suggestion
  // First check config, otherwise fallback to ctx, then let the client
  // preference dictate cipher selection if all else fails.
```

Very minor nit: The if/else if statements are checking identical things on a `SSLCipherPreferenceList`, would abstracting the check into a static helper function help?

ssl/ssl_test.cc
```diff
@@ -2823,6 +2821,252 @@ TEST(SSLTest, FindingCipher) {
...
+  // STACK_OF(SSL_CIPHER) *client_ciphers = SSL_get_ciphers(client.get());
+  // STACK_OF(SSL_CIPHER) *server_ciphers = SSL_get_ciphers(server.get());
+  //
+  // ASSERT_TRUE(sk_SSL_CIPHER_find_awslc(client_ciphers, NULL, cipher1));
+  // ASSERT_TRUE(sk_SSL_CIPHER_find_awslc(client_ciphers, NULL, cipher2));
+  // ASSERT_FALSE(sk_SSL_CIPHER_find_awslc(client_ciphers, NULL, cipher3));
+  //
+  // ASSERT_TRUE(sk_SSL_CIPHER_find_awslc(server_ciphers, NULL, cipher1));
+  // ASSERT_FALSE(sk_SSL_CIPHER_find_awslc(server_ciphers, NULL, cipher2));
+  // ASSERT_FALSE(sk_SSL_CIPHER_find_awslc(server_ciphers, NULL, cipher3));
```
samuel40791765, line 3067: Were these tests meant to be enabled?

ssl/ssl_cipher.cc
```diff
@@ -1234,6 +1234,84 @@ static bool is_known_default_alias_keyword_filter_rule(const char *rule,
   return false;
 }
 
+int update_cipher_list(UniquePtr<SSLCipherPreferenceList> &dst,
+                       UniquePtr<SSLCipherPreferenceList> &ciphers,
+                       UniquePtr<SSLCipherPreferenceList> &tls13_ciphers) {
+  bssl::UniquePtr<STACK_OF(SSL_CIPHER)> tmp_cipher_list;
+  int num_removed_tls13_ciphers = 0, num_added_tls13_ciphers = 0;
```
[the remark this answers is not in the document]
smittals2: fair point, changed it where it makes sense :)

ssl/internal.h
smittals2: That's generally the paradigm between SSL_CTX and SSL objects but I'll specify it here for clarity.

ssl/tls13_server.cc
smittals2: We'd only be abstracting away the null and size check - not sure if that's a valuable "helper". Updated the documentation.

ssl/ssl_test.cc
smittals2: whoops good catch, I think I commented them out trying to debug something - back in now!

samuel40791765 (approved):

ssl/ssl_test.cc
```diff
@@ -4720,39 +4964,58 @@ TEST(SSLTest, ClientHello) {
...
+      0x0a, 0x00, 0x08, 0x00, 0x06, 0x00, 0x1d, 0x00, 0x17, 0x00, 0x18, 0x00,
+      0x0b, 0x00, 0x02, 0x01, 0x00, 0x00, 0x23, 0x00, 0x00, 0x00, 0x0d, 0x00,
+      0x14, 0x00, 0x12, 0x04, 0x03, 0x08, 0x04, 0x04, 0x01, 0x05, 0x03, 0x08,
+      0x05, 0x05, 0x01, 0x08, 0x06, 0x06, 0x01, 0x02, 0x01, 0x00, 0x33, 0x00,
+      0x26, 0x00, 0x24, 0x00, 0x1d, 0x00, 0x20, 0x00, 0x00, 0x00, 0x00, 0x00,
+      0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+      0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+      0x00, 0x00, 0x00, 0x00, 0x2d, 0x00, 0x02, 0x01, 0x01, 0x00, 0x2b, 0x00,
+      0x09, 0x08, 0x03, 0x04, 0x03, 0x03, 0x03, 0x02, 0x03, 0x01}}
   };
```
andrewhop, line 5019: Did you verify this test passes before you made this change? In theory this change detector could be one commit late if this PR introduced some kind of change.

ssl/ssl_test.cc
```diff
@@ -6412,17 +6689,17 @@ TEST(SSLTest, EmptyCipherList) {
   // Initially, the cipher list is not empty.
   EXPECT_NE(0u, sk_SSL_CIPHER_num(SSL_CTX_get_ciphers(ctx.get())));
 
+  // Configuring the empty cipher list with |SSL_CTX_set_ciphersuites| works
+  EXPECT_TRUE(SSL_CTX_set_ciphersuites(ctx.get(), ""));
+  EXPECT_EQ(0u, sk_SSL_CIPHER_num(ctx->tls13_cipher_list->ciphers.get()));
+  EXPECT_NE(0u, sk_SSL_CIPHER_num(SSL_CTX_get_ciphers(ctx.get())));
+
```
andrewhop, line 6696: Why did you change when this gets checked?

ssl/tls13_server.cc
```diff
@@ -112,7 +112,13 @@ static int ssl_ext_supported_versions_add_serverhello(SSL_HANDSHAKE *hs,
 
 static const SSL_CIPHER *choose_tls13_cipher(const SSL *ssl) {
   STACK_OF(SSL_CIPHER) *tls13_ciphers = nullptr;
-  if (ssl->ctx->tls13_cipher_list &&
+  // First check config, otherwise fallback to ctx, and then let client
+  // preference dictate cipher selection if all else fails.
```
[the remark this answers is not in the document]
smittals2: The fallback to client preference happens in the ssl_choose_tls13_cipher function. 

Now that we initialize the CTX by default with TLS 1.3 ciphersuites, if falling back to the CTX does not provide any tls 1.3 ciphersuites, then that means the user has explicitly configured it this way. So you are correct, we should not fallback to client preferences anymore.

ssl/ssl_test.cc
smittals2: Because every operation to a setter updates the main cipher_list. If I set the main cipher_list to "" first, it would still be getting updated with any remaining values in the tls 1.3 ciphersuites. Therefore, we first set the TLS 1.3 ciphersuites to empty. That way when we set the main cipher_list to empty, it remains empty and is not updated by any populated TLS 1.3 ciphersuites.

ssl/ssl_test.cc
smittals2: Yep verified this passes on upstream/mainline

ssl/ssl_test.cc
```diff
@@ -2823,6 +2821,252 @@ TEST(SSLTest, FindingCipher) {
...
+  ASSERT_TRUE(SSL_CTX_set_max_proto_version(client_ctx.get(), TLS1_2_VERSION));
+  ASSERT_TRUE(SSL_CTX_set_min_proto_version(server_ctx.get(), TLS1_2_VERSION));
+  ASSERT_TRUE(SSL_CTX_set_max_proto_version(server_ctx.get(), TLS1_3_VERSION));
+
+  bssl::UniquePtr<SSL> client, server;
+  // Have to ensure config is not shed per current implementation of SSL_get_ciphers.
+  ASSERT_TRUE(ConnectClientAndServer(&client, &server, client_ctx.get(), server_ctx.get(),
+    ClientConfig(), false));
+
+  ASSERT_TRUE(CompleteHandshakes(client.get(), server.get()));
```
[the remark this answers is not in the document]
smittals2: Discussed offline

ssl/ssl_test.cc
```diff
@@ -2823,6 +2821,252 @@ TEST(SSLTest, FindingCipher) {
...
+  bssl::UniquePtr<SSL> client2, server2;
+  ASSERT_TRUE(CreateClientAndServer(&client2, &server2, client_ctx.get(), server_ctx.get()));
+
+  // Modify TLS 1.3 ciphersuites for server2 SSL object, but not client
+  ASSERT_TRUE(SSL_set_ciphersuites(server2.get(), "TLS_AES_256_GCM_SHA384"));
+
+  // Handshake should fail as server SSL config and client CTX objects have no
+  // shared TLS 1.3 cipher.
+  ASSERT_FALSE(CompleteHandshakes(client2.get(), server2.get()));
+  ASSERT_EQ(ERR_GET_REASON(ERR_get_error()), SSL_R_NO_SHARED_CIPHER);
```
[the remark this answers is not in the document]
smittals2: So client2 and server2 are created with contexts that support TLS_AES_128_GCM_SHA256. Then I restrict the SSL server object to only use TLS_AES_256_GCM_SHA384. SSL configurations take preference over SSL_CTX preferences. 

Therefore, server2's SSL TLS_AES_256_GCM_SHA384 is unable to negotiate with client2's SSL_CTX TLS_AES_128_GCM_SHA256.

ssl/ssl_test.cc
```diff
@@ -2823,6 +2821,252 @@ TEST(SSLTest, FindingCipher) {
...
+
+  bssl::UniquePtr<SSL> client2, server2;
+  ASSERT_TRUE(CreateClientAndServer(&client2, &server2, client_ctx.get(), server_ctx.get()));
+
+  // Modify TLS 1.2 ciphersuites for server2 SSL object, but not client
+  ASSERT_TRUE(SSL_set_cipher_list(server2.get(), "TLS_RSA_WITH_AES_256_GCM_SHA384"));
+
+  // Handshake should fail as server SSL config and client CTX objects have no
+  // shared TLS 1.2 cipher.
+  ASSERT_FALSE(CompleteHandshakes(client2.get(), server2.get()));
```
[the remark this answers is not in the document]
smittals2: Similar logic as above. Client2 and server2 are created from CTX objects that support TLS_RSA_WITH_AES_256_CBC_SHA. Then I restrict server2's SSL object to only use TLS_RSA_WITH_AES_256_GCM_SHA384. Therefore, there is a mismatch and failed handshake is expected.

andrewhop (approved):

PR #2096, opened 2025-01-06
Removal of Enable Dilithium Flag

jakemas:
### Issues:

### Description of changes: 
- Added a description to APIs to indicate the external APIs may still be subject to change as the standards space develops. (We don't envision them changing, but they are extremely new).

### Call-outs:

Removing the flag has little consequence -- other than it makes the APIs we expose in `include/openssl/evp.h` that much more "final". We should consider how much we like them before we commit to them. We made a point to refer to asymmetric keypairs as public and private keys, rather than public and secret keys -- but there is inconsistency with `EVP_PKEY_kem_new_raw_secret_key`. After internal discussions, we'd also like to change the name of `EVP_PKEY_kem_new_raw_public_key` and `EVP_PKEY_kem_new_raw_secret_key` to `EVP_PKEY_kem_new_raw_encapsulation_key` and `EVP_PKEY_kem_new_raw_decapsulation_key`  to match the names within [FIPS 203](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf). Thus, we should stick with `private`.

We did consider the use of an experimental flag (by using the OPENSSL_DEPRECATED alias), but, this seems a little over the top. Alternatively, we could place the EVP APIs in `include/openssl/experimental/`. The arguments of `EVP_PKEY_pqdsa_new_raw_secret_key`, `EVP_PKEY_pqdsa_new_raw_public_key`, and `EVP_PKEY_CTX_pqdsa_set_params` are stable and consistant with `KEM` and `EC` variants (see [EVP_PKEY_CTX_kem_set_params](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L914), [EVP_PKEY_CTX_set_dh_paramgen_prime_len](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L184), and [EVP_PKEY_CTX_set_ec_param_enc](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L1151), [EVP_PKEY_new_raw_private_key](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L278), [EVP_PKEY_new_raw_public_key](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L284), [EVP_PKEY_kem_new_raw_public_key](https://github.com/jakemas/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/include/openssl/evp.h#L920)).


util/fipstools/delocate/delocate.peg.go
```diff
@@ -1,6 +1,6 @@
 package main
 
-// Code generated by /home/ubuntu/go/bin/peg delocate.peg DO NOT EDIT.
+// Code generated by /Users/jakemas/go/bin/peg /Users/jakemas/Documents/PQ/ML-DSA/jan6/aws-lc/util/fipstools/delocate/delocate.peg DO NOT EDIT.
```
WillChilds-Klein, line 3: not major, but may want to revert this line.

jakemas: Thanks! [49921f0](https://github.com/aws/aws-lc/pull/2096/commits/49921f0753147eda549a4d0e3e7ce71264bf2ad9). Making a note to point out in the [documentation ](https://github.com/aws/aws-lc/blob/73f0497dcc91a24cdf5a29b756c5fd545eeb44cc/util/fipstools/delocate/delocate.peg#L19-L24) that you need to change this line after!

jakemas:
PAUSE on this PR. Removing flag in https://github.com/aws/aws-lc/pull/2105

PR #2098, opened 2025-01-07
SHA3 and SHAKE - New API Design

manastasova:
### Issues:
Resolves #CryptoAlg-2810

### Description of changes: 
AWS-LC supports SHA3 and SHAKE algorithms though low level SHA3_Init, SHA3_Update, SHA3_Final and SHAKE_init, SHAKE_Final APIs. Currently, there are two issues with the implementation and usage of SHA3 and SHAKE:

This PR addresses both of them as follows:
- Introduce new API layers - FIPS202, SHA3 and SHAKE. 
   - _Keccak1600_ layer (https://github.com/aws/aws-lc/pull/2097) implements KeccakF1600 Absorb and Squeeze functions; Keccak1600 layer does _not_ manage internal input/output buffers. 
   - _FIPS202_ layer implements Reset, Init, Update, and Finalize functionalities; FIPS202 layer manages the internal input/output buffers, allowing incremental requests (not necessarily multiple of block size) to Update (Absorb) and Squeeze for input/output processing. (Other functionalities, such as zero-ing of bitstate, block size checks, etc. are also handled by FIPS202 API layer). 
     -  _FIPS202_ layer implements all common behavior between SHA3 and SHAKE algorithms.
     - _FIPS202_ layer checks/updates the |ctx->state| flag when handling a common behavior between SHA3 and SHAKE algorithms. |ctx->state|  is updated in the higher level SHA3_ SHAKE_ API layer when the behavior of both algorithms diverges (SHAKE _can_ allow incremental squeezes).
   - _SHA3_ layer implements Init, Update, and Final functionalities; SHA3 layer only implements SHA3 algorithm, thus, offers a single-call SHA3_Final function. SHA3_Final will update internal |ctx->state| flag to prevent any sequential calls.
    - _SHAKE_ layer implements XOF SHAKE algorithm, therefore, offers Init, Absorb, Squeeze, and Final functionalities; 
      - _SHAKE_ layer implements Init, and Absorb, Squeeze with incremental call support for absorb (byte-wise) and squeeze (block-wise).
      - _SHAKE_ layer implements a single-call SHAKE_Final function that generates an arbitrary length output and finalizes SHAKE. Incremental XOF output generation is handled by |SHAKE_Squeeze|. |SHAKE_Squeeze| can be called multiple times. SHAKE_Final should be called only once.

- KECCAK600_CTX struct updates:
   - Remove |padded| field
   - Introduce |state| field
      - |state| can be |KECCAK1600_STATE_ABSORB|, |KECCAK1600_STATE_SQUEEZE|, |KECCAK1600_STATE_FINAL|
      - |KECCAK1600_STATE_ABSORB| - allows incremental absorbs until the state is changed
      - |KECCAK1600_STATE_SQUEEZE| - allows incremental squeezes for |SHAKE_Squeeze|
      - |KECCAK1600_STATE_Final| - prevents from incremental squeezes via |SHAKE_Final| and prevents from consecutive calls to |SHA3_Final| (Final functions are single-shot functions).

SHA3 vs SHAKE algorithms (APIs usage):
>- SHA3 digest generation: SHA3_Init; SHA3_Update; SHA3_Final;
>- SHAKE (single-shot-output) output generation: SHAKE_Init; SHAKE_Absorb; SHAKE_Final;  
>- SHAKE (incremental) output generation: SHAKE_Init; SHAKE_Absorb; SHAKE_Squeeze<sup>+</sup>;

### Call-outs:
Service indicator is updated: 
- Inside SHA3 and SHAKE single shot APIs (as previously in AWS-LC);
- Inside SHA3_Final (as previously in AWS-LC);
- Inside SHAKE_Final (Single-Shot XOF Final output generation as previously in AWS-LC);
- Inside SHAKE_Squeeze (Streaming XOF Squeezes output generation updates the service indicator after each extendable output update);

All other algorithms that use SHA3/SHAKE APIs are updated:
- ML-KEM (SHA3/SHAKE calls will be inlined later)
- ML-DSA (SHAKE_Squeeze (incremental XOF output functionality) inside ML-DSA is never invoked with the KAT test vectors and gtests)
### Testing:


crypto/fipsmodule/sha/internal.h
```diff
@@ -396,32 +404,39 @@ OPENSSL_EXPORT uint8_t *SHAKE128(const uint8_t *data, const size_t in_len,
...
+// SHA3_Update check |ctx| pointer and |len| value and calls FIPS202_Update.
+int SHA3_Update(KECCAK1600_CTX *ctx, const void *data,
+                               size_t len);
 
-// SHA3_Reset zeros the bitstate and the amount of processed input.
-void SHA3_Reset(KECCAK1600_CTX *ctx);
+// SHA3_Final pads the last data block and processes it through Keccak1600_Absorb.
+// It writes |md_digest| bytes of finalized digest to |md|, returns 1 on
+// success and 0 on failure.
+OPENSSL_EXPORT int SHA3_Final(uint8_t *md, KECCAK1600_CTX *ctx);
```
darylmartin100, line 417: You're removing a parameter from the public API.  Will there be any impact for downstream users?

justsmth: I don't think this function is currently public. To verify, I checked our Rust bindings to see whether it was listed. I don't see it here: https://docs.rs/aws-lc-sys/latest/aws_lc_sys/#functions

manastasova: From my understanding, only the functions defined in the include/openssl are consider external APIs.
These APIs should be internal, however, they are exported so that they can be used in our tests. However, I am not sure if there may be some customers that use these internal APIs as well.

crypto/fipsmodule/sha/sha3.c
```diff
@@ -236,7 +257,62 @@ int SHAKE_Init(KECCAK1600_CTX *ctx, size_t block_size) {
...
 }
 
+int SHAKE_Absorb(KECCAK1600_CTX *ctx, const void *data, size_t len) {
+  if (ctx == NULL) {
+    return 0;
+  }
+
+  if (len == 0) {
+    return 1;
+  }
```
justsmth, line 267: Verify that `data != NULL`

crypto/fipsmodule/sha/sha3.c
```diff
@@ -113,7 +113,7 @@ uint8_t *SHAKE256(const uint8_t *data, const size_t in_len, uint8_t *out, size_t
 static void FIPS202_Reset(KECCAK1600_CTX *ctx) {
   memset(ctx->A, 0, sizeof(ctx->A));
```
justsmth, line 114: Should this instead use `OPENSSL_memset`? (Here and other places...)

crypto/fipsmodule/sha/sha3.c
```diff
@@ -169,9 +153,9 @@ int SHA3_Update(KECCAK1600_CTX *ctx, const void *data, size_t len) {
       return 1;
     }
 
-     // There is enough data to fill or overflow the intermediate
-     // buffer. So we append |rem| bytes and process the block,
-     // leaving the rest for later processing.
+    // There is enough data to fill or overflow the intermediate
+    // buffer. So we append |rem| bytes and process the block,
+    // leaving the rest for later processing.
     memcpy(ctx->buf + num, data_ptr_copy, rem);
```
justsmth, line 159: Should this instead use `OPENSSL_memcpy`? (Here and other places...)

crypto/fipsmodule/sha/sha3.c
```diff
@@ -196,36 +180,73 @@ int SHA3_Update(KECCAK1600_CTX *ctx, const void *data, size_t len) {
...
 
-    if (Keccak1600_Absorb(ctx->A, ctx->buf, block_size, block_size) != 0) {
-      return 0;
-    }
+// SHA3_Final should be called once to process final digest value
+// |ctx->state| flag does not need to be updated
+int SHA3_Final(uint8_t *md, KECCAK1600_CTX *ctx) {
+  if (ctx->md_size == 0) {
+    return 1;
   }
```
justsmth, line 237: Verify that `md != NULL` and `ctx != NULL`.

crypto/fipsmodule/sha/sha3.c
```diff
@@ -236,7 +257,62 @@ int SHAKE_Init(KECCAK1600_CTX *ctx, size_t block_size) {
...
+    return 1;
+  }
+
+  if (ctx->state == KECCAK1600_STATE_FINAL) {
+    return 0;
+  }
+
+  if (ctx->state == KECCAK1600_STATE_ABSORB) {
+    // Skip FIPS202_Finalize if the input has been padded and the last block has been processed
+    if (FIPS202_Finalize(md, ctx) == 0) {
```
justsmth, line 308: This comment feels like it's in the wrong place.

crypto/fipsmodule/sha/sha3.c
```diff
@@ -196,36 +180,73 @@ int SHA3_Update(KECCAK1600_CTX *ctx, const void *data, size_t len) {
...
+
+  if (Keccak1600_Absorb(ctx->A, ctx->buf, block_size, block_size) != 0) {
+    return 0;
+  }
+
+  return 1;
+}
+
+// SHA3 APIs implement SHA3 functionalities on top of FIPS202 API layer
+int SHA3_Init(KECCAK1600_CTX *ctx, size_t bit_len) {
```
nebeid, line 209: Can the logic in SHA3_Init and SHAKE_Init be reversed in terms of returning 0 inside the if statement? I believe they're the only ones in this file that return 0 at the end.

crypto/fipsmodule/sha/sha3.c
```diff
@@ -196,36 +180,73 @@ int SHA3_Update(KECCAK1600_CTX *ctx, const void *data, size_t len) {
...
-    ctx->buf[block_size - 1] |= 0x80;
+  return FIPS202_Update(ctx, data, len);
+}
 
-    if (Keccak1600_Absorb(ctx->A, ctx->buf, block_size, block_size) != 0) {
-      return 0;
-    }
+// SHA3_Final should be called once to process final digest value
+// |ctx->state| flag does not need to be updated
+int SHA3_Final(uint8_t *md, KECCAK1600_CTX *ctx) {
```
nebeid, line 234: ~Should there be a check that it was called in the right state?~ It's done in `FIPS202_Finalize()`

crypto/fipsmodule/sha/sha3.c
```diff
@@ -196,36 +180,73 @@ int SHA3_Update(KECCAK1600_CTX *ctx, const void *data, size_t len) {
...
-    ctx->buf[num] = ctx->pad;
-    ctx->buf[block_size - 1] |= 0x80;
+  return FIPS202_Update(ctx, data, len);
+}
 
-    if (Keccak1600_Absorb(ctx->A, ctx->buf, block_size, block_size) != 0) {
-      return 0;
-    }
+// SHA3_Final should be called once to process final digest value
+// |ctx->state| flag does not need to be updated
```
nebeid, line 233: The state is updated within the function, though.

crypto/fipsmodule/sha/sha3.c
```diff
@@ -236,7 +257,62 @@ int SHAKE_Init(KECCAK1600_CTX *ctx, size_t block_size) {
...
+  }
+
+  if (len == 0) {
+    return 1;
+  }
+
+  return FIPS202_Update(ctx, data, len);
+}
+
+// SHAKE_Final is a single-shot API and can be called once to finalize absorb and squeeze phases
```
nebeid, line 272:
```suggestion
// SHAKE_Final is to be called once to finalize absorb and squeeze phases
```
I think "single-shot" here is not the right term because it's used with non-streaming APIs such as `SHAKE_256()`

crypto/fipsmodule/sha/sha3.c
```diff
@@ -236,7 +257,62 @@ int SHAKE_Init(KECCAK1600_CTX *ctx, size_t block_size) {
...
+  Keccak1600_Squeeze(ctx->A, md, ctx->md_size, ctx->block_size, ctx->state);
+  ctx->state = KECCAK1600_STATE_FINAL;
+
+  FIPS_service_indicator_update_state();
+
+  return 1;
+}
+
+// SHAKE_Squeeze can be called multiple times
+// SHAKE_Squeeze should be called for incremental XOF output
```
nebeid, line 294:
```suggestion
// SHAKE_Squeeze can be called multiple time for incremental XOF output
```

crypto/fipsmodule/sha/sha3.c
```diff
@@ -236,7 +257,62 @@ int SHAKE_Init(KECCAK1600_CTX *ctx, size_t block_size) {
...
+  if (ctx->md_size == 0) {
+    return 1;
+  }
+
+  if (ctx->state == KECCAK1600_STATE_FINAL) {
+    return 0;
+  }
+
+  if (ctx->state == KECCAK1600_STATE_ABSORB) {
+    // Skip FIPS202_Finalize if the input has been padded and the last block has been processed
```
nebeid, line 307: Maybe move this comment one line up because it's talking about skipping this if statement (if I understood correctly).

crypto/fipsmodule/sha/sha3.c
```diff
@@ -236,7 +257,62 @@ int SHAKE_Init(KECCAK1600_CTX *ctx, size_t block_size) {
...
+  ctx->state = KECCAK1600_STATE_FINAL;
+
+  FIPS_service_indicator_update_state();
+
+  return 1;
+}
+
+// SHAKE_Squeeze can be called multiple times
+// SHAKE_Squeeze should be called for incremental XOF output
+int SHAKE_Squeeze(uint8_t *md, KECCAK1600_CTX *ctx, size_t len) {
```
nebeid, line 295: Do we have unit tests for SHAKE_Squeeze or we're relying on ml_dsa and ml_kem?

crypto/fipsmodule/sha/internal.h
```diff
@@ -71,6 +71,14 @@ extern "C" {
 // SHAKE128 has the maximum block size among the SHA3/SHAKE algorithms.
 #define SHA3_MAX_BLOCKSIZE SHAKE128_BLOCKSIZE
 
+// Define state flag values for Keccak-based functions
+#define KECCAK1600_STATE_ABSORB     0 
+#define KECCAK1600_STATE_SQUEEZE    1  
+// KECCAK1600_STATE_FINAL restricts the incremental calls to SHAKE_Final .
+// KECCAK1600_STATE_FINAL can be called once. SHAKE_Squeeze cannot be called after SHAKE_Final.
+// SHAKE_Squeeze should be called for streaming XOF output.
+#define KECCAK1600_STATE_FINAL      2
```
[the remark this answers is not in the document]
manastasova: I believe this should be covered by the returns of the functions. I don't think an additional value for |state| is needed since the purpose is to restrict the number/sequence of: _absorb_ (only before padding is performed), _squeeze_ (incremental squeezes are performed with no limitation), _final_ (one call to final is allowed) function. Once a _squeeze_ is called, no further _absorb_ should be allowed, once _final_ is called, no further _absorb_, _squeeze_, or _final_ is allowed.

crypto/fipsmodule/sha/sha3.c
justsmth: I agree that the checks here should not be needed in this "layer" if we're confident the parameters are checked on all logic paths going into these functions.  What I noticed was that there was some inconsistency between the functions, where some were checking certain parameters and others were not, and it wasn't clear to me why they weren't all performing the same checks.

justsmth (approved):

crypto/fipsmodule/sha/internal.h
```diff
@@ -396,32 +404,43 @@ OPENSSL_EXPORT uint8_t *SHAKE128(const uint8_t *data, const size_t in_len,
...
+// SHA3_Final pads the last data block and absorbs it through |FIPS202_Finalize|.
+// It processes the data through |Keccak1600_Squeeze| and returns 1 on success 
+// and 0 on failure.
 int SHA3_Final(uint8_t *md, KECCAK1600_CTX *ctx);
 
-// Keccak1600_Absorb processes the largest multiple of |r| out of |len| bytes and
-// returns the remaining number of bytes.
+// SHAKE_Init initialises |ctx| fields through |FIPS202_Init| and 
+// returns 1 on success and 0 on failure.
+int SHAKE_Init(KECCAK1600_CTX *ctx, size_t block_size);
```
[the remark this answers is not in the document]
nebeid: I think it's fine to leave the bitlength in SHA3_Init because there isn't an equivalent function for SHA2 that we need to align with (e.g. using byte length). If we used bitlength with SHAKE_Init, it would identify 128 or 256. It's minor anyway and if it makes more sense to you this way, that's good.

nebeid (approved):

PR #2099, opened 2025-01-07
Use getentropy for OpenBSD, FreeBSD, and MacOS

torben-hansen:
### Description of changes: 

Use `getentropy` for some operating systems. Namely, Apple MacOS, OpenBSD, and FreeBSD. `getentropy` is closer to the noise sources.

See also q/3t08AyoE0m4q.

### Testing:

Even though we didn't use `getentropy` prior to this PR, we had a unit test. Just re-use that.

### DIFF
--- a/crypto/fipsmodule/rand/urandom.c
+++ b/crypto/fipsmodule/rand/urandom.c
@@ -69,20 +69,6 @@
 #include <CommonCrypto/CommonRandom.h>
 #endif
 
-#if defined(OPENSSL_FREEBSD)
-#define URANDOM_BLOCKS_FOR_ENTROPY
-#include <sys/param.h>
-#if __FreeBSD_version >= 1200000
-// getrandom is supported in FreeBSD 12 and up.
-#define FREEBSD_GETRANDOM
-#include <sys/random.h>
-#endif
-#endif
-
-#if defined(OPENSSL_OPENBSD)
-#include <stdlib.h>
-#endif
-
 #include <openssl/thread.h>
 #include <openssl/mem.h>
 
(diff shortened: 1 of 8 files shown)


crypto/fipsmodule/rand/urandom.c
```diff
@@ -399,9 +399,26 @@ static int fill_with_entropy(uint8_t *out, size_t len, int block, int seed) {
   }
 #endif
 
+#if defined(OPENSSL_MACOS)
+  // POSIX 2024 says unistd, but man page for macos says the former
+  #include <sys/random.h>
+  // To get system randomness on MacOS and iOS we use |CCRandomGenerateBytes|
+  // rather than |getentropy| and /dev/urandom.
```
justsmth, line 406: Update comment here and for OPENSSL_IOS.

crypto/fipsmodule/rand/urandom.c
```diff
@@ -399,9 +399,26 @@ static int fill_with_entropy(uint8_t *out, size_t len, int block, int seed) {
...
+  // TODO at most can do 256 bytes
+  if (getentropy(out, len) == 0) {
+    return 1;
+  } else {
```
justsmth, line 410: This is inside the function `fill_with_entropy`, which is used for [both `CRYPTO_sysrand` and `CRYPTO_sysrand_for_seed`](https://github.com/aws/aws-lc/blob/d726d0616bc30029bf6e7bada9f0a2ba3a7e3b00/crypto/fipsmodule/rand/urandom.c#L475-L487).  

`CCRandomGenerateBytes`/`arc4random_buf` provide the output of a PRNG and are better suited for `CRYPTO_sysrand`, and `getentropy` is intended for seeding aPRNG and is better suited for `CRYPTO_sysrand_for_seed`.

On MacOS/OpenBSD, I think we should continue to use `CCRandomGenerateBytes`/`arc4random_buf` when `seed == 0` and only use `getentropy` when `seed == 1`.  How important is the distinction of `CRYPTO_sysrand` and `CRYPTO_sysrand_for_seed`?

include/openssl/target.h
```diff
@@ -82,6 +82,17 @@
 #define OPENSSL_APPLE
 #endif
 
+#if defined(__APPLE__)
```
justsmth, line 85: NP: There's another `#if defined(__APPLE__)` 4 lines above this.  These two could be merged.

crypto/fipsmodule/rand/urandom.c
torben-hansen: this is refactored now.

crypto/rand_extra/getentropy.c
```diff
@@ -0,0 +1,48 @@
...
+#if !defined(GETENTROPY_MAX)
+// Per POSIX 2024
+// https://pubs.opengroup.org/onlinepubs/9799919799/functions/getentropy.html
+#define GETENTROPY_MAX 256
+#endif
+
+void CRYPTO_sysrand(uint8_t *out, size_t requested) {
+  // getentropy max request size is GETENTROPY_MAX.
+  while (requested > 0) {
+    size_t request_chunk = (requested > GETENTROPY_MAX) ? GETENTROPY_MAX : requested;
```
skmcgrail, line 34: There isn't really a way to communicate though that the amount of bytes we wrote to `out` was less then requested? Seems potentially problematic, should it be `size_t *requested` so that we can be given how much bytes to read and set the value back to how many was actually read?

skmcgrail (approved):

justsmth (approved):

PR #2101, opened 2025-01-08
Only SHA3/SHAKE Init Updates via FIPS202 API layer

manastasova:
### Issues:
Resolves (some of) #CryptoAlg-2810

### Description of changes: 
AWS-LC supports SHA3 and SHAKE algorithms though low level SHA3_Init, SHA3_Update, SHA3_Final and SHAKE_init, SHAKE_Final APIs. Currently, there are two issues with the implementation and usage of SHA3 and SHAKE:

This PR introduces the APIs that need changes in the function parameter list. Particularly, SHA3_Init used a parameter |pad_char|, which was used for differentiating SHA3_Init from SHAKE_Init functions.

This PR introduces only the changes in the Init functions (FIPS202_Init, FIPS202_Reset, SHA3_Init, and SHAKE_Init):
- Introduce (partially) new API layers - FIPS202, SHA3 and SHAKE. 
   - _Keccak1600_ layer (https://github.com/aws/aws-lc/pull/2097) implements KeccakF1600 Absorb and Squeeze functions; Keccak1600 layer does _not_ manage internal input/output buffers. 
   - _FIPS202_ layer implements (so far) Reset and Init functions. (Update and Finalize functionalities are to be introduced in following PRs); FIPS202 layer manages the internal input/output buffers, allowing incremental requests (not necessarily multiple of block size) to Update (Absorb) and Squeeze for input/output processing. (Other functionalities, such as zero-ing of bitstate, block size checks, etc. are also handled by FIPS202 API layer). 
     -  _FIPS202_ layer implements (so far) Reset and Init common behavior between SHA3 and SHAKE algorithms.
     - _FIPS202_ layer checks/updates the |ctx->padded| flag when handling a common behavior between SHA3 and SHAKE algorithms.
   - _SHA3_ layer implements (so far) Init functionality (Update, and Final are to be introduced in following PRs);
    - _SHAKE_ layer implements XOF SHAKE algorithm, therefore, offers (so far) Init functionality (Absorb, Squeeze, and Final functionalities are to be introduced in following PRs); 

### Call-outs:
All SHA3_Init calls are updated:
- digests.c
- sha3_test.cc
- 
### Testing:


crypto/fipsmodule/sha/sha3.c
```diff
@@ -109,49 +109,43 @@ uint8_t *SHAKE256(const uint8_t *data, const size_t in_len, uint8_t *out, size_t
...
 }
 
-int SHAKE_Init(KECCAK1600_CTX *ctx, size_t block_size) {
-  // The SHAKE block size depends on the security level of the algorithm only
-  // It is independent of the output size
-  ctx->block_size = block_size;
-  return SHA3_Init(ctx, SHAKE_PAD_CHAR, 0);
+// FIPS202 APIs manage internal input/output buffer on top of Keccak1600 API layer
+static void FIPS202_Reset(KECCAK1600_CTX *ctx) {
+  memset(ctx->A, 0, sizeof(ctx->A));
```
andrewhop, line 114: Should this also zero out `ctx->buf`?
```suggestion
  OPENSSL_memset(ctx->A, 0, sizeof(ctx->A));
```

crypto/fipsmodule/sha/sha3.c
```diff
@@ -109,49 +109,43 @@ uint8_t *SHAKE256(const uint8_t *data, const size_t in_len, uint8_t *out, size_t
...
 }
 
-
-int SHAKE_Final(uint8_t *md, KECCAK1600_CTX *ctx, size_t len) {
-  ctx->md_size = len;
-  return SHA3_Final(md, ctx);
+static int FIPS202_Init(KECCAK1600_CTX *ctx, uint8_t pad, size_t block_size, size_t bit_len) {
+  if (pad != SHA3_PAD_CHAR && 
+      pad != SHAKE_PAD_CHAR) { 
+    return 0;
```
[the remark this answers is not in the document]
manastasova: The values for the padding are the same, however, they are 'interleaved' by 0's in case the message is shorter than the block. E.g. the pad for sha3 is 0x06 (added after the last byte of the input); the rest of the block size is set to zero, where the last byte of the block is set to 0x80. When the last byte (+1 byte) of the message (message of length block size - 1 bytes) and the last byte of the block coincide, the last byte becomes 0x06 ^ 0x80, which will be 0x86. When messages are shorter, more zero's are added in the middle of these values. However, the implementation covers all the cases using only the 0x06 and 0x1f pad, where all the rest is 0's and last block byte is 0x80.

crypto/fipsmodule/sha/sha3.c
manastasova: The `buf` is only accessible through the `buf_load` value. So, even if it is not zero'ed, it should not be an issue. Once that data is placed into the buffer, the `buf_load` is assigned the number of bytes that are 'useful' and accessible. When `buf_load` is 0, there is no useful data in the buffer.
However, I can zero it.

andrewhop: No need to zero.

andrewhop (approved):

crypto/fipsmodule/sha/sha3.c
```diff
@@ -109,49 +109,43 @@ uint8_t *SHAKE256(const uint8_t *data, const size_t in_len, uint8_t *out, size_t
...
+  }
+      
+  if (block_size <= sizeof(ctx->buf)) {
```
nebeid, line 125: Nit: my preference is to check for the erroneous condition and return 0 within the if statement, then continue with the regular processing afterwards as in (most of, if not all) the rest of the code base. 
Update: I see that similar processing below is following the same pattern, so you can ignore this comment.

crypto/fipsmodule/sha/sha3.c
```diff
@@ -109,49 +109,43 @@ uint8_t *SHAKE256(const uint8_t *data, const size_t in_len, uint8_t *out, size_t
...
+      ctx->md_size = bit_len / 8;
+      ctx->pad = pad;
+      return 1;
+    }
+    return 0;
 }
 
+// SHA3 APIs implement SHA3 functionalities on top of FIPS202 API layer
 void SHA3_Reset(KECCAK1600_CTX *ctx) {
   memset(ctx->A, 0, sizeof(ctx->A));
```
nebeid, line 137: Why not call `FIPS202_Reset()`?

crypto/fipsmodule/sha/sha3.c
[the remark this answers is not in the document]
nebeid: When the message length is -1 modulo the block size, i.e. one byte is left, and is filled with 0x86, how is it differentiated from a similar message that's one byte longer and has this value, 0x86, as the last byte? should another full block be added for padding?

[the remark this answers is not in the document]
manastasova: Yes, in this case another full block is absorbed; consisting of padding `0x06 ... 0x00 ... 0x80` with `block_size-2` bytes of 0x00's.

crypto/fipsmodule/sha/sha3.c
manastasova: I change it in the next PR: https://github.com/aws/aws-lc/blob/f48fb780e11386c7ddce36ce53300c1eeb05e43f/crypto/fipsmodule/sha/sha3.c#L113

nebeid (approved):

PR #2103, opened 2025-01-10
SCRUTINICE fixes

smittals2:
### Description of changes: 
- Removal of some dead code
- Casting work arounds

### Call-outs:

### Testing:


crypto/bn_extra/convert.c
```diff
@@ -448,7 +448,7 @@ BIGNUM *BN_mpi2bn(const uint8_t *in, size_t len, BIGNUM *out) {
   }
   out->neg = ((*in) & 0x80) != 0;
   if (out->neg) {
-    BN_clear_bit(out, BN_num_bits(out) - 1);
+    BN_clear_bit(out, (int)BN_num_bits(out) - 1);
```
justsmth, line 451: NP: Should we verify that `(BN_num_bits(out) - 1) <= INT_MAX`?

crypto/fipsmodule/modes/cbc.c
```diff
@@ -85,12 +85,10 @@ void CRYPTO_cbc128_encrypt(const uint8_t *in, uint8_t *out, size_t len,
     }
     (*block)(out, out, key);
     iv = out;
+    // This will always be true as a result of the first while loop
```
justsmth, line 88: Then we should instead change the `while (len)` above to an `if (len > 0)` and then there's no reason to `break`.

crypto/bio/bio.c
```diff
@@ -85,11 +85,15 @@ static int call_bio_callback_with_processed(BIO *bio, const int oper,
...
-    if (ret > 0) {
-      // BIO will only read int |len| bytes so this is a safe cast
-      ret = (int)processed;
+    long callback_ret = bio->callback_ex(bio, oper, buf, len, 0, 0L, ret, &processed);
+    if (callback_ret <= INT_MAX) {
+      ret = (int)callback_ret;
+      if (ret > 0) {
+        // BIO will only read int |len| bytes so this is a safe cast
+        ret = (int)processed;
+      }
```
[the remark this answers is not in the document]
smittals2: The function calling the bio callback is passed in a `ret` value. Per the documentation, generally the callback should return this ret value untouched (unless for whatever reason the consumer wants to modify it). A negative value is considered erroneous, so I think it should be okay to set to -1 if in the unlikely case the callback returns an "out of bounds" value. 

Documentation link: https://docs.openssl.org/3.4/man3/BIO_set_callback/

crypto/bn_extra/convert.c
```diff
@@ -448,7 +448,11 @@ BIGNUM *BN_mpi2bn(const uint8_t *in, size_t len, BIGNUM *out) {
   }
   out->neg = ((*in) & 0x80) != 0;
   if (out->neg) {
-    BN_clear_bit(out, BN_num_bits(out) - 1);
+    unsigned num_bits = BN_num_bits(out);
+    if (num_bits >= INT_MAX) {
+      return NULL;
+    }
```
justsmth, line 454: Before returning NULL, this should free `out` if it's been allocated:
```
if (out_is_alloced) {
  BN_free(out);
}
```

crypto/fipsmodule/modes/cbc.c
```diff
@@ -76,7 +76,7 @@ void CRYPTO_cbc128_encrypt(const uint8_t *in, uint8_t *out, size_t len,
     out += 16;
   }
 
-  while (len) {
```
skmcgrail, line 79: What's the background on making this change?

skmcgrail (approved):

PR #2110, opened 2025-01-10
Remove jent_read_entropy_safe usage from AWS-LC (main)

smittals2:
### Issues:
`P186365975`

### Description of changes: 
These changes remove any usage of jent_read_entropy_safe function from AWS-LC. This function has an implementation vulnerability described [here](https://github.com/smuellerDD/jitterentropy-library/blob/c62c88645f27716f78e1267a182cff40e96081c2/SECURITY.md#2024-12-30).

jent_read_entropy_safe is now unused in our codebase, however, it is still present in the Jitter code. These changes are outside the FIPS boundary and do not impact the FIPS integrity hash.

### Call-outs:
Eventually, we should pull in the newest Jitter release with the security fix. 

### Testing:


andrewhop (approved):

crypto/fipsmodule/rand/cpu_jitter_test.cc
```diff
@@ -42,10 +42,6 @@ TEST(CPUJitterEntropyTest, Basic) {
   EXPECT_EQ(jent_read_entropy(jitter_ec.instance,
                               (char*) data0, data_len), data_len);
 
-  // Draw some entropy with the "safe" API to check if it works.
-  EXPECT_EQ(jent_read_entropy_safe(&jitter_ec.instance,
-                                   (char*) data1, data_len), data_len);
-
```
andrewhop, line 48: I think the spirit of this test was a very simple sanity test that jitter was returning unique output based on line 49. Instead of removing this just use the same `jent_read_entropy` as above.

andrewhop (changes requested):

andrewhop (dismissed):

skmcgrail (approved):

andrewhop (approved):

skmcgrail (approved):

PR #2111, opened 2025-01-13
Align guard macros for OPENSSL_cpuid_setup

justsmth:
### Issues:
Addresses aws/aws-lc-rs#655

### Description of changes: 
* The criteria for invoking `OPENSSL_cpuid_setup` should align with the criteria for defining `OPENSSL_cpuid_setup`.

### Call-outs:
Consumers building `aws-lc-fips-sys` for `s390x` were seeing the following failure:
```
.../boringssl_prefix_symbols.h:20:26: error: implicit declaration of function 'aws_lc_fips_0_13_1_OPENSSL_cpuid_setup'; did you mean 'aws_lc_fips_0_13_1_OPENSSL_sk_set'? [-Werror=implicit-function-declaration]
```

### Testing:
I verified this fix works for the `aws-lc-fips-sys` build for `s390x-unknown-linux-gnu`.

### DIFF
--- a/crypto/fipsmodule/bcm.c
+++ b/crypto/fipsmodule/bcm.c
@@ -258,7 +258,7 @@ static void BORINGSSL_bcm_power_on_self_test(void) {
 // TODO: remove !defined(OPENSSL_PPC64BE) from the check below when starting to support
 // PPC64BE that has VCRYPTO capability. In that case, add `|| defined(OPENSSL_PPC64BE)`
 // to `#if defined(OPENSSL_PPC64LE)` wherever it occurs.
-#if !defined(OPENSSL_NO_ASM) && !defined(OPENSSL_PPC32BE) && !defined(OPENSSL_PPC64BE)
+#if defined(HAS_OPENSSL_CPUID_SETUP) && !defined(OPENSSL_NO_ASM)
   OPENSSL_cpuid_setup();
 #endif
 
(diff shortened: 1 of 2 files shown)


crypto/fipsmodule/bcm.c
```diff
@@ -258,7 +258,7 @@ static void BORINGSSL_bcm_power_on_self_test(void) {
 // TODO: remove !defined(OPENSSL_PPC64BE) from the check below when starting to support
 // PPC64BE that has VCRYPTO capability. In that case, add `|| defined(OPENSSL_PPC64BE)`
 // to `#if defined(OPENSSL_PPC64LE)` wherever it occurs.
-#if !defined(OPENSSL_NO_ASM) && !defined(OPENSSL_PPC32BE) && !defined(OPENSSL_PPC64BE)
+#if defined(HAS_OPENSSL_CPUID_SETUP) && !defined(OPENSSL_NO_ASM)
```
smittals2, line 261: If we're aligning the invocation of OPENSSL_cpuid_setup with it's declaration conditions, why an additional OPENSSL_NO_ASM condition here?

smittals2 (approved):

crypto/fipsmodule/bcm.c
justsmth: If `OPENSSL_NO_ASM` has been enabled then there's no need to call `OPENSSL_cpuid_setup`. The function's purpose is to identify available CPU features at runtime so we execute the assembly implementation best optimized for the CPU. But with `OPENSSL_NO_ASM`, there would be no assembly implementations available to execute.

PR #2112, opened 2025-01-13
Ensure service indicator is incremented only once, update RSA and ED25519 to ensure the state is locked

andrewhop:
### Issues:
Resolves P186477736

### Description of changes: 
Currently the service indicator checks that `before != after` and multiple approved APIs might call each other. If a lock is missed a lower approved algorithm will increment the count which incorrectly marks the higher level API as approved. This is happening in three spots:
1. Approved API's self tests run and increment the service indicator on first use
2. In Ed25519 sign/verify the calls to SHA were incrementing the indicator
3. In the Ec25519 and RSA keygen the PCT sign/verify was incrementing the count

This change updates the service indicator to enforce `before + 1 == after` with a debug assert.

### Call-outs
This doesn't change the external behavior of the service indicator, what algorithms are approved, or what APIs are approved. The service indicator tests are unchanged. This change just ensures what we expect to be modifying the indicator is in the thing doing the update.

### Testing:
The existing service indicator tests cover all approved APIs, and the new requirement that `before + 1 = after` ensures only one thing per call increments the count.

I took out a lock and verified it failed as expected:
```
[ RUN      ] ServiceIndicatorTest.ED25519KeyGen
Assertion failed: (before + 1 == after), function TestBody, file service_indicator_test.cc, line 5184.
OPENSSL_armcap=0x3D crypto/crypto_test [shard 8/10]
crypto/crypto_test failed to complete: signal: abort trap
```

### DIFF
--- a/crypto/curve25519_extra/curve25519_extra.c
+++ b/crypto/curve25519_extra/curve25519_extra.c
@@ -80,9 +80,9 @@ int ED25519ph_sign_digest(uint8_t out_sig[ED25519_SIGNATURE_LEN],
                           const uint8_t *context, size_t context_len) {
   FIPS_service_indicator_lock_state();
   boringssl_ensure_hasheddsa_self_test();
-  FIPS_service_indicator_unlock_state();
   int res = ED25519ph_sign_digest_no_self_test(out_sig, digest, private_key,
                                                context, context_len);
+  FIPS_service_indicator_unlock_state();
   if (res) {
     FIPS_service_indicator_update_state();
   }
--- a/crypto/fipsmodule/curve25519/curve25519.c
+++ b/crypto/fipsmodule/curve25519/curve25519.c
@@ -126,6 +126,7 @@ static void ed25519_keypair_pct(uint8_t public_key[ED25519_PUBLIC_KEY_LEN],
 
 void ED25519_keypair(uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
   uint8_t out_private_key[ED25519_PRIVATE_KEY_LEN]) {
+  FIPS_service_indicator_lock_state();
   boringssl_ensure_eddsa_self_test();
   SET_DIT_AUTO_RESET;
 
@@ -141,6 +142,7 @@ void ED25519_keypair(uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
 
   ed25519_keypair_pct(out_public_key, out_private_key);
 
+  FIPS_service_indicator_unlock_state();
   FIPS_service_indicator_update_state();
 }
 
--- a/crypto/fipsmodule/curve25519/curve25519.c
+++ b/crypto/fipsmodule/curve25519/curve25519.c
@@ -126,6 +126,8 @@ static void ed25519_keypair_pct(uint8_t public_key[ED25519_PUBLIC_KEY_LEN],
 
 void ED25519_keypair(uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
   uint8_t out_private_key[ED25519_PRIVATE_KEY_LEN]) {
+  // We have to avoid the self tests and digest function in ed25519_keypair_pct
+  // from updating the service indicator.
   FIPS_service_indicator_lock_state();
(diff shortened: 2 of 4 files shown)


smittals2 (approved):

samuel40791765 (approved):

crypto/fipsmodule/curve25519/curve25519.c
```diff
@@ -117,6 +117,7 @@ static void ed25519_keypair_pct(uint8_t public_key[ED25519_PUBLIC_KEY_LEN],
 
 void ED25519_keypair(uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
   uint8_t out_private_key[ED25519_PRIVATE_KEY_LEN]) {
+  FIPS_service_indicator_lock_state();
```
nebeid, line 120: Because of the distance between the lock and unlock statements, it's helpful here to add a comment about which service(s) are prevented from updating the service indicator.

include/openssl/service_indicator.h
```diff
@@ -50,7 +50,7 @@ enum FIPSStatus {
     int before = FIPS_service_indicator_before_call();              \
     func;                                                           \
     int after = FIPS_service_indicator_after_call();                \
-    if (before != after) {                                          \
+    if (before + 1 == after) {                                      \
```
samuel40791765, line 53: `FIPS_service_indicator_before_call` and `FIPS_service_indicator_after_call` both return an unsigned int. This was on purpose because we didn’t want the counter to overflow to a negative number if we used int. We also wanted to make sure we behaved correctly when the unsigned `int` counter wrapped around to `0`, so we decided not to do `before + 1 == after` and go ahead with `before != after` at the time.
If we’re still concerned about this edge case, I don’t think we should change the original behavior. But I do agree we should add better testing around the counter to make sure that we’re only incrementing by `1`.  We should also document the original intention… Sorry, I had almost forgotten why we did this at the time.

samuel40791765 (changes requested):

crypto/fipsmodule/curve25519/curve25519.c
```diff
@@ -132,14 +133,21 @@ void ED25519_keypair(uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
...
 
+  FIPS_service_indicator_unlock_state();
   FIPS_service_indicator_update_state();
 }
 
 int ED25519_sign(uint8_t out_sig[ED25519_SIGNATURE_LEN],
                  const uint8_t *message, size_t message_len,
                  const uint8_t private_key[ED25519_PRIVATE_KEY_LEN]) {
+  FIPS_service_indicator_lock_state();
   boringssl_ensure_eddsa_self_test();
```
skmcgrail, line 144: Would it just be easier for the self-test functions to just always do an lock/unlock (in the FIPS cases)? Would simplify things.

skmcgrail: Or optionally pull the indicator update up out of the self_test functions (which you did already here) which eliminates the issue. I can fix these in my other Ed25519 if you want since there is overlap here between what we pick?

skmcgrail:
You'll want to rebase since Ed25519ph got merged in, but I think we are aligned / should be fixed now in the functions you touched.

crypto/curve25519_extra/curve25519_extra.c
```diff
@@ -80,9 +80,9 @@ int ED25519ph_sign_digest(uint8_t out_sig[ED25519_SIGNATURE_LEN],
                           const uint8_t *context, size_t context_len) {
   FIPS_service_indicator_lock_state();
   boringssl_ensure_hasheddsa_self_test();
-  FIPS_service_indicator_unlock_state();
   int res = ED25519ph_sign_digest_no_self_test(out_sig, digest, private_key,
                                                context, context_len);
+  FIPS_service_indicator_unlock_state();
```
skmcgrail, line 85: Here I thought I got it right in all of the places :)

skmcgrail (approved):

samuel40791765 (approved):

PR #2113, opened 2025-01-13
ExternalMu mode for pre-hash ML-DSA

jakemas:
### Issues:
Resolves #CryptoAlg-2829

### Description of changes: 
This PR adds a new signing and verification mode for ML-DSA known as ExternalMu. This signing/verification mode allows for pre-hashing of large messages that are to be signed with ML-DSA.

This PR focuses mainly on connecting `EVP_PKEY_sign` and `EVP_PKEY_verify` to PQDSA ASN.1 functionality, to allow users access to these APIs for pre-hash modes of ML-DSA.

I have added a generic function `pkey_pqdsa_sign_generic` that bifurcates between calls from the `EVP_PKEY_METHOD` to call `pqdsa->method->pqdsa_verify_message` when `pkey_pqdsa_verify_message` is called and `pqdsa->method->pqdsa_verify` when `pkey_pqdsa_verify` is called.

These two new `PQDSA_METHOD` functions `pqdsa_sign` and `pqdsa_verify` are as follows:
```
  int (*pqdsa_sign)(const uint8_t *private_key,
                          uint8_t *sig,
                          size_t *sig_len,
                          const uint8_t *message,
                          size_t message_len);

  int (*pqdsa_verify)(const uint8_t *public_key,
                      const uint8_t *sig,
                      size_t sig_len,
                      const uint8_t *message,
                      size_t message_len);
```
These allow for the input message to represent the pre-hashed message (hence why `ctx_string` is removed, as this should already be in the pre-hash).

We can then define the `PQDSA_METHOD` as:
```
static const PQDSA_METHOD sig_ml_dsa_44_method = {
  ml_dsa_44_keypair,
  ml_dsa_44_sign,
  ml_dsa_extmu_44_sign,
  ml_dsa_44_verify,
  ml_dsa_extmu_44_verify
};
```

where `ml_dsa_extmu_44_sign` and `ml_dsa_extmu_44_verify` connect to external mu variants of ML-DSA.

### Call-outs:
It is important to note that external-mu ML-DSA is considered the same algorithm as FIPS 204 ML-DSA. As such, external-mu ML-DSA shares the same OIDs as ML-DSA. NIST’s AVCP lab have [confirmed](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/HXhRTjshe5E/m/mkkFd7zcAwAJ) that they will allow alternate interfaces for FIPS 204: Algorithm 7 (ML-DSA.Sign) and FIPS 204: Algorithm 7 (ML-DSA.Verify). Specifically, they will accept a variant of ML-DSA.Sign/ML-DSA.Sign_internal that directly accepts the pre-message string, and a variant of ML-DSA.Verify/ML-DSA.Verify_internal that directly accepts the pre-message string.

These are implemented as `ml_dsa_extmu_sign`, `ml_dsa_sign_internal` and `ml_dsa_verify_internal`.

### Testing:
Added new test suite:
`--gtest_filter=All/PerMLDSA*`

~~This includes a new KAT framework. As official extmu KATs are not yet available, we exercise the APIs through the existing KATs for ML-DSA. Three new KAT files `MLDSA_EXTMU_{44/65/87}_hedged_pure.txt` have been created. These files are exactly the same KATs as `MLDSA_{44/65/87}_hedged_pure.txt` with the exception that an additional line has been added to each KAT containing `mu` the result of computing line 6 of Algorithm 7 ML-DSA.Sign_internal and line 7 of Algorithm 8 ML-DSA.Verify_internal.~~

~~To prove that `mu` is equal to `H(tr | M')` we construct `mu` from `tr` and `M'` and compare with the value in the KAT (this hashing is timely, and slows down testing).~~

~~As the current ML-DSA KATs provide the following parameters: `count, xi, rng, seed, pk, sk, msg, mlen, sm, smlen, ctx` we are able to construct `mu` manually by constructing `mu = shake_256(shake_256(pk), 0 || ctx || ctx_size || msg`. That was done once to generate `mu` values to insert into the existing KAT files. I then compared this value with the actual value produced within `sign.c` by printing out the constructed `mu` as an intermediate value.~~

~~I reconstruct the construction of `mu` in the KAT from the same provided parameters, and compare it against the known result within the new KAT for external mu. This way, when alternative test vectors come out they will be a straight swap.~~

ACVP NIST KATs came out during this PR (how timely) 
- https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-keyGen-FIPS204
- https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-sigGen-FIPS204
- https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-sigVer-FIPS204

We run the ACVP test vectors obtained from https://github.com/usnistgov/ACVP-Server within the three functions `PerMLDSATest.ACVPKeyGen`, `PerMLDSATest.ACVPSigGen` and `PerMLDSATest.ACVPSigVer`. These correspond to the tests found at [ML-DSA-keyGen-FIPS204](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-keyGen-FIPS204), [ML-DSA-sigGen-FIPS204](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-sigGen-FIPS204), and [ML-DSA-sigVer-FIPS204](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-sigVer-FIPS204).

To test ML-DSA pure mode:`external_mu = False`, `deterministic  = false`, we use `tgId = 19, 21, 23` of sigGen and `tgId = 7, 9, 11` of sigVer.
To test ML-DSA ExternalMu mode:`external_mu = true`, `deterministic  = false`, we use `tgId = 20, 22, 24` of sigGen and `tgId = 8, 10, 12` of sigVer.

We also provide functional testing hat:
- Generates ML-DSA key for each parameter set
- Pre-hashes a message `M` by generating `mu = SHAKE256(SHAKE256(pk) || 0 || |ctx| || ctx || M)`
- Signs `mu` with `EVP_PKEY_sign`
- Verifies `mu` with `EVP_PKEY_verify`
- Bonus: verifies raw message `M` using `EVP_DigestVerify` to show compatibility between signing modes.


jakemas:
Nevine's previous comments:
- https://github.com/jakemas/aws-lc/commit/1f2f577177b4ee7059d1fc400a5271fc97f606f3#r151195281
- https://github.com/jakemas/aws-lc/commit/1f2f577177b4ee7059d1fc400a5271fc97f606f3#r151195352
- https://github.com/jakemas/aws-lc/commit/1f2f577177b4ee7059d1fc400a5271fc97f606f3#r151195574
- https://github.com/jakemas/aws-lc/commit/1f2f577177b4ee7059d1fc400a5271fc97f606f3#r151200087
- https://github.com/jakemas/aws-lc/commit/1f2f577177b4ee7059d1fc400a5271fc97f606f3#r151200565
- https://github.com/jakemas/aws-lc/commit/1f2f577177b4ee7059d1fc400a5271fc97f606f3#r151200814
- https://github.com/jakemas/aws-lc/commit/1f2f577177b4ee7059d1fc400a5271fc97f606f3#r151200856

jakemas:
I made codecov happier by adding in some negative testing around signature size (this is actually required by [FIPS 204 3.6.2](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf) Public-Key and Signature Length Checks) in [bd4e629](https://github.com/aws/aws-lc/pull/2113/commits/bd4e6290b44109343309c3d0b5a2afc88abcfb24).

As for the rest:
- the lines in [crypto/ml_dsa/ml_dsa_ref/sign.c](https://app.codecov.io/gh/aws/aws-lc/pull/2113?src=pr&el=tree&filepath=crypto%2Fml_dsa%2Fml_dsa_ref%2Fsign.c&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws#diff-Y3J5cHRvL21sX2RzYS9tbF9kc2FfcmVmL3NpZ24uYw==) are untestable
- the four new lines in [crypto/evp_extra/p_pqdsa.c](https://app.codecov.io/gh/aws/aws-lc/pull/2113?src=pr&el=tree&filepath=crypto%2Fevp_extra%2Fp_pqdsa.c&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws#diff-Y3J5cHRvL2V2cF9leHRyYS9wX3BxZHNhLmM=) are from 2 sign and 2 verify functions that now bifurcate around the pure vs pre-hash modes.  They are hit when there is no defined method for `!pqdsa->method->pqdsa_sign_message` or `!pqdsa->method->pqdsa_sign`. `pqdsa` methods are well-defined and I have manually verified that these function methods exist 😉

crypto/evp_extra/p_pqdsa.c
```diff
@@ -56,9 +56,9 @@ static int pkey_pqdsa_keygen(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey) {
   return 1;
 }
 
-static int pkey_pqdsa_sign_message(EVP_PKEY_CTX *ctx, uint8_t *sig,
-                                     size_t *sig_len, const uint8_t *message,
-                                     size_t message_len) {
+static int pkey_pqdsa_sign_generic(EVP_PKEY_CTX *ctx, uint8_t *sig,
+                                   size_t *sig_len, const uint8_t *message,
+                                   size_t message_len, int prehash) {
```
nebeid, line 61: To make it clear that this is not HashML-DSA, I suggest replacing `prehash` with `external_mu`.

crypto/evp_extra/p_pqdsa.c
```diff
@@ -96,16 +96,43 @@ static int pkey_pqdsa_sign_message(EVP_PKEY_CTX *ctx, uint8_t *sig,
     return 0;
   }
 
-  if (!pqdsa->method->pqdsa_sign(key->private_key, sig, sig_len, message, message_len, NULL, 0)) {
-    OPENSSL_PUT_ERROR(EVP, ERR_R_INTERNAL_ERROR);
-    return 0;
+  // |prehash| is a flag we use to indicate that the message to be signed has already
```
nebeid, line 99: Similarly to the comment above, I suggest explicitly saying something along the line
"|external_mu| indicates that the input is mu which is the result of hashing the public key, the context string and the message."

crypto/ml_dsa/ml_dsa.h
```diff
@@ -42,22 +42,41 @@ OPENSSL_EXPORT int ml_dsa_44_sign(const uint8_t *private_key,
                                   const uint8_t *message, size_t message_len,
                                   const uint8_t *ctx_string, size_t ctx_len);
```
nebeid, line 43: I noticed that in the implementation, `ctx_len` is called `ctx_string_len`. I think changing the former to the latter is good.

crypto/pqdsa/internal.h
```diff
@@ -15,22 +15,33 @@ typedef struct {
...
   int (*pqdsa_sign)(const uint8_t *private_key,
-                    uint8_t *sig,
-                    size_t *sig_len,
-                    const uint8_t *message,
-                    size_t message_len,
-                    const uint8_t *ctx_string,
-                    size_t ctx_string_len);
+                          uint8_t *sig,
+                          size_t *sig_len,
+                          const uint8_t *message,
```
nebeid, line 29: Would you like to refer to it as `digest` or `mu` here and all the way down until the internal/generic functions are used?

crypto/fipsmodule/evp/digestsign.c
```diff
@@ -78,8 +79,19 @@ DEFINE_LOCAL_DATA(struct evp_md_pctx_ops, EVP_MD_pctx_ops) {
 }
 
 static int uses_prehash(EVP_MD_CTX *ctx, enum evp_sign_verify_t op) {
-  return (op == evp_sign) ? (ctx->pctx->pmeth->sign != NULL)
-                          : (ctx->pctx->pmeth->verify != NULL);
+  // Pre-hash modes in ML-DSA differ from other signing algorithms, so we
```
nebeid, line 82:
```suggestion
  // Pre-hash mode of ML-DSA that uses an external mu calculation differs from other signing algorithms, so we
```

nebeid (approved):

crypto/evp_extra/p_pqdsa.c
jakemas: Not a good idea. This is the `PQDSA` logic for ALL PQDSA types. ExternalMu isn't defined for any PQDSA scheme other than ML-DSA. For example, why would SLH-DSA have a flag for ExternalMu? 

I have changed all the ML-DSA specific uses of a pre-hash flag to use `external_mu` but the ones that remain called `prehash` are intentional so that we can use the flag for the purposes of indicating any pre-hash method for any PQDSA signature scheme.

crypto/evp_extra/p_pqdsa.c
jakemas: Again, this is `PQDSA` logic, that would be true for all signature schemes in the PQDSA API, not just ML-DSA, so I don't want to add any comments that are only specific to how ML-DSA works.

jakemas: I could add a comment that says, `For ML-DSA |prehash| indicates that the input is mu which is the result of hashing the public key, the context string and the message.` or something to that effect?

crypto/fipsmodule/evp/digestsign.c
jakemas: added in [ba1856b](https://github.com/aws/aws-lc/pull/2113/commits/ba1856b920504afa5a4fb37ddf5ff80f70093b34)

crypto/evp_extra/p_pqdsa.c
jakemas: Added comment to that effect in [ba1856b](https://github.com/aws/aws-lc/pull/2113/commits/ba1856b920504afa5a4fb37ddf5ff80f70093b34)

crypto/pqdsa/internal.h
jakemas: I think I got them all the way down. I couldn't change the one inside `pkey_pqdsa_verify_generic` as it handles both `message` and `digest` options

crypto/fipsmodule/evp/digestsign.c
```diff
@@ -78,8 +79,19 @@ DEFINE_LOCAL_DATA(struct evp_md_pctx_ops, EVP_MD_pctx_ops) {
 }
 
 static int uses_prehash(EVP_MD_CTX *ctx, enum evp_sign_verify_t op) {
-  return (op == evp_sign) ? (ctx->pctx->pmeth->sign != NULL)
-                          : (ctx->pctx->pmeth->verify != NULL);
+  // Pre-hash modes of ML-DSA that uses an external mu calculation differs from
+  // other signing algorithms, so we specifically check for NIDs of type NID_MLDSAXX.
```
WillChilds-Klein, line 83: > Pre-hash modes of ML-DSA that uses an external mu calculation differs from other signing algorithms

in a way that isn't meaningful to the intent of this function? ML-DSA now has a pre-hash, so it's not clear why it's an exception to this function's semantic

include/openssl/evp.h
```diff
@@ -367,8 +369,9 @@ OPENSSL_EXPORT int EVP_DigestSign(EVP_MD_CTX *ctx, uint8_t *out_sig,
 // signing options.
 //
 // For single-shot signing algorithms which do not use a pre-hash, such as
-// Ed25519 and Dilithium, |type| should be NULL. The |EVP_MD_CTX| itself is
-// unused but is present so the API is uniform. See |EVP_DigestVerify|.
+// Ed25519, or when using ML-DSA in non pre-hash mode, |type| should be NULL.
```
WillChilds-Klein, line 372: do we eventually plan to support the other digest algorithms allowed by FIPS 204? if so, when would be a good time to do that?

right now, we're basically committing to `type == NULL` resulting in a default of SHAKE-256. changing that behavior in the future (i.e. after this release) would be backwards-incompatible.

I'm not saying this is necessarily a bad thing, I just want to make sure we're consciously making that decision and committing to it.

On the other hand, if we require that SHAKE-256 be specified explicitly by the caller here, we're not committing to any such defaults, and can still defer other digest types to later releases.

WillChilds-Klein (commented):
```
This includes a new KAT framework. As official extmu KATs are not yet available, we exercise the APIs through the existing KATs for ML-DSA. Three new KAT files MLDSA_EXTMU_{44/65/87}_hedged_pure.txt have been created. These files are exactly the same KATs as MLDSA_{44/65/87}_hedged_pure.txt with the exception that an additional line has been added to each KAT containing mu the result of computing line 6 of Algorithm 7 ML-DSA.Sign_internal and line 7 of Algorithm 8 ML-DSA.Verify_internal.
```

How was mu calculated here?

Also, we intend to replace these with NIST KATs once available, right?

crypto/evp_extra/p_pqdsa.c
andrewhop: I agree this shouldn't be external_mu, but it isn't clear if prehash means the input has been pre-hashed or if AWS-LC should hash the input before signing it, can this follow the method names and use sign_message or sign_digest 0/1?

crypto/fipsmodule/evp/digestsign.c
nebeid: The external mu calculation is precluded using `uses_prehash` from calculating the digest in `EVP_Digest<Sign|Verify>` because:
- it's pure ML-DSA with the digest calculation extracted out and left to the caller to perform.
- it's not a regular hash call on the message only as in the cases of ECDSA and RSA; it includes concatenating the public key with other bytes then the message.
- When FIPS 204 HashML-DSA is supported, it will have its own NIDs and will not be precluded from `uses_prehash`, because calculating the digest of the message first is a step in it.

include/openssl/evp.h
nebeid: We can enforce SHAKE-256 when we implement HashML-DSA as this will be used by EVP_DigestSign and will not be precluded in `uses_prehash` as per the comment https://github.com/aws/aws-lc/pull/2113#discussion_r1916992591, while now Pure and External Mu ML-DSA (which is still pure) is precluded from `uses_prehash`.
I think that making that contract for pure Ed25519 and ML-DSA with MD being NULL is beneficial to say that these are special modes that don't adhere to the typical algorithms in which `uses_prehash` is true.

jakemas:
> How was mu calculated here?

The plan is to re-use the KATs for ML-DSA to provide KATs for externalmu mode. As the current ML-DSA KATs provide the following parameters: `count, xi, rng, seed, pk, sk, msg, mlen, sm, smlen, ctx` we are able to construct `mu` manually by constructing `mu = shake_256(shake_256(pk), 0 || ctx || ctx_size || msg`. That was done once to generate `mu` values to insert into the existing KAT files. I then compared this value with the actual value produced within `sign.c` by printing out the constructed `mu` as an intermediate value.

I reconstruct the construction of `mu` in the KAT from the same provided parameters, and compare it against the known result within the new KAT for external mu. This way, when alternative test vectors come out they will be a straight swap.

crypto/evp_extra/p_pqdsa.c
jakemas: Thanks Andrew, have I understood you correctly with this update: [318f0c1](https://github.com/aws/aws-lc/pull/2113/commits/318f0c1c0835cbcd3c30643cb25bdc567c566926)

crypto/ml_dsa/ml_dsa_ref/README.md
```diff
@@ -19,7 +19,9 @@ that initialize a given structure with values corresponding to a parameter set.
 - Documentation has been added to `ntt.c`, `packing.c`, `poly.c`, `polyvec.c`, and `rounding.c` that outlines the algorithm specification (including algorithm number) in FIPS 204.
 - `poly.c` and `sign.c` have been modified to cleanse intermediate data as soon as it is no longer needed as defined in FIPS 204 Section 3.6.3.
 - Intermediate values are cleansed within `crypto_sign_keypair_internal`, `crypto_sign_keypair`, `crypto_sign_signature_internal`, `crypto_sign_verify_internal`, `crypto_sign_verify`, `poly_uniform_eta`, `poly_uniform_gamma1`, and `poly_challenge` as per FIPS 204 Section 3.6.3.
+- `sign.c` has been modified to provide support for ML-DSA in ExternalMu mode. This is an alternative implementation of ML-DSA sign and verify that accepts `mu` as input, rather than the raw message. As `mu` can be constructed (and thus hashed) in another cryptographic module.
```
nebeid, line 22: l.10, should the directory name be changed (or mention both the current name and the used-to-be name)?

jakemas: [195779c](https://github.com/aws/aws-lc/pull/2113/commits/195779c43648ed8e9bcbbb8354b02e4babf7b401) fixed -- great spot.

nebeid (dismissed):

nebeid (dismissed):

nebeid (dismissed):

crypto/ml_dsa/ml_dsa_ref/sign.c
```diff
@@ -291,14 +299,47 @@ int ml_dsa_sign(ml_dsa_params *params,
...
+                      uint8_t *sig,
+                      size_t *siglen,
+                      const uint8_t *mu,
+                      size_t mulen,
+                      const uint8_t *sk)
+{
+  uint8_t rnd[ML_DSA_RNDBYTES];
+
+  if (!RAND_bytes(rnd, ML_DSA_RNDBYTES)) {
+    return -1;
```
WillChilds-Klein, line 334: `EVP_PKEY_sign`'s documentation says:

```
// It returns one on success or zero on error. (Note: this differs from
// OpenSSL, which can also return negative values to indicate an error. )
```

Is that no longer true? here and elsewhere in this file where we return -1 (e.g. L292, L300, perhaps others), should we return 0 instead?

this might cause issues for callers who are doing something like `if (!EVP_DigestSign(...)) { /* handle failure case */ }` to fail improperly.

ok, fine. i'm sort of splitting hairs here, as `RAND_bytes` [can never fail](https://github.com/aws/aws-lc/blob/232b7b3fad9cda6a79c2698b31e4d2af3bb8315e/crypto/fipsmodule/rand/rand.c#L579), but is that true for other places in this file where we return -1?

perhaps this is best addressed in a follow-up, as it concerns code not currently touched by this PR.

WillChilds-Klein (approved):

crypto/ml_dsa/ml_dsa_ref/sign.c
jakemas: Were any of the functions in this file not return `0` then the calling function within `crypto/ml_dsa/ml_dsa.c` would pick up and process the error correctly. See https://github.com/aws/aws-lc/blob/49580d671650e390a85c14aff199f316adbf4ea9/crypto/ml_dsa/ml_dsa.c#L138 as an example. This is true for keygen, sign, verify -- all errors were processed at the layer above the algorithm code.

nebeid (approved):

crypto/ml_dsa/ml_dsa_ref/sign.c
jakemas: for extmu
`EVP_PKEY_sign` -> `pkey_pqdsa_sign` -> `pkey_pqdsa_sign_generic` -> `ml_dsa_extmu_44_sign`  -> `ml_dsa_extmu_sign`

for pure
`EVP_DigestSign` -> `pkey_pqdsa_sign_message` -> `pkey_pqdsa_sign_generic` -> `ml_dsa_44_sign` -> `ml_dsa_extmu_sign`

WillChilds-Klein (approved):

PR #2126, opened 2025-01-17
Prepare Docker image for CI integration jobs

nhatnghiho:
### Description of changes: 
Install packages needed for running CI integration jobs successfully.

### DIFF
--- a/tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
+++ b/tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
@@ -25,6 +25,11 @@ RUN set -ex && \
     apt-get -y --no-install-recommends install \
     software-properties-common \
     automake \
+    autoconf \
+    autoconf-archive \
+    autogen \
+    binutils-dev \
+    build-essential \
     check \
     cmake \
     curl \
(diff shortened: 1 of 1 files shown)


tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
```diff
@@ -25,47 +25,107 @@ RUN set -ex && \
...
     make \
     ninja-build \
     patch \
     perl \
-    libunwind-dev \
     pkg-config \
     git \
-    gnutls-dev \
     ca-certificates \
+    sudo \
```
andrewhop, line 43: This looks a little weird, can you install sudo into an image that doesn't have it and get it to work?

andrewhop (commented):
Curious how big is this image now? I'm thinking about how long it will take to pull from ECR and if we want this all in the base or a specific integration test image.

nhatnghiho:
> Curious how big is this image now? I'm thinking about how long it will take to pull from ECR and if we want this all in the base or a specific integration test image.

The image right now is around 1.05GB, which is almost a 50% size increase from previously (726MB). I like the idea of creating a separate test image, since a lot of the new packages are quite arbitrary

tests/ci/docker_images/linux-x86/ubuntu-22.04_base/Dockerfile
nhatnghiho: Oops it's an oversight from my side. This is not needed.

nhatnghiho: Re-added `sudo` since it turns out `run_openvpn_integration.sh` needs it

skmcgrail (dismissed):

skmcgrail (approved):

andrewhop (approved):

PR #2128, opened 2025-01-18
Add support for PKCS12_set_mac

samuel40791765:
### Issues:
Addresses `CryptoAlg-2821`

### Description of changes: 
Ruby has added a binding for an additional function called `PKCS12_set_mac` in it's master branch: https://github.com/ruby/ruby/commit/c79b4354074742ca1cbbb25a4f04bbffeb58407d This was discovered prior to the PR for Ruby's CI integration with the master branch being merged: https://github.com/aws/aws-lc/pull/2071

OpenSSL's implementation of the function directly sets a designated mac field within the `PKCS12` structure. Our `PKCS12` structure is folded into a string of bytes along with its length and there aren't any available fields for us to directly set. This means that AWS-LC's implementation of `PKCS12_set_mac` has to parse the proper contents from `PKCS12` and rerun the key and mac generation with the new parameters provided. 

### Call-outs:
1. The parsing logic is similar to `PKCS12_get_key_and_certs`, but the logic can't be shared. This is because we need to maintain pointers to certain parts of the CBS parsers in `PKCS12_set_mac`, so that we can properly rewrite the contents later.
2. I've abstracted the key and mac generation in `PKCS12_create` and `PKCS12_set_mac` to a single function called `pkcs12_gen_and_write_mac`.
3. OpenSSL 3.0 uses SHA-256 as the default hash, yet OpenSSL 1.1.1 uses SHA1. TODO: Make a decision

### Testing:
New Tests

### DIFF
--- a/crypto/pkcs8/pkcs12_test.cc
+++ b/crypto/pkcs8/pkcs12_test.cc
@@ -551,7 +551,7 @@ static bssl::UniquePtr<X509> MakeTestCert(EVP_PKEY *key) {
   return x509;
 }
 
-static bool PKCS12CreateVector(bssl::UniquePtr<PKCS12> *p12, EVP_PKEY *pkey,
+static bool PKCS12CreateVector(bssl::UniquePtr<PKCS12> &p12, EVP_PKEY *pkey,
                                const std::vector<X509 *> &certs) {
   bssl::UniquePtr<STACK_OF(X509)> chain(sk_X509_new_null());
   if (!chain) {
@@ -564,21 +564,21 @@ static bool PKCS12CreateVector(bssl::UniquePtr<PKCS12> *p12, EVP_PKEY *pkey,
     }
   }
 
-  p12->reset(PKCS12_create(kPassword, nullptr /* name */, pkey,
+  p12.reset(PKCS12_create(kPassword, nullptr /* name */, pkey,
                            nullptr /* cert */, chain.get(), 0, 0, 0, 0, 0));
-  if (!p12->get()) {
+  if (!p12) {
     return false;
   }
   return true;
 }
 
-static void ExpectPKCS12Parse(PKCS12 *p12, EVP_PKEY *expect_key,
-                              X509 *expect_cert,
+static void ExpectPKCS12Parse(bssl::UniquePtr<PKCS12> &p12,
+                              EVP_PKEY *expect_key, X509 *expect_cert,
                               const std::vector<X509 *> &expect_ca_certs) {
   EVP_PKEY *key = nullptr;
   X509 *cert = nullptr;
   STACK_OF(X509) *ca_certs = nullptr;
-  ASSERT_TRUE(PKCS12_parse(p12, kPassword, &key, &cert, &ca_certs));
+  ASSERT_TRUE(PKCS12_parse(p12.get(), kPassword, &key, &cert, &ca_certs));
 
   bssl::UniquePtr<EVP_PKEY> delete_key(key);
   bssl::UniquePtr<X509> delete_cert(cert);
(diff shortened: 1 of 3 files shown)


crypto/pkcs8/pkcs12_test.cc
```diff
@@ -533,7 +551,7 @@ static bssl::UniquePtr<X509> MakeTestCert(EVP_PKEY *key) {
   return x509;
 }
 
-static bool PKCS12CreateVector(std::vector<uint8_t> *out, EVP_PKEY *pkey,
+static bool PKCS12CreateVector(bssl::UniquePtr<PKCS12> *p12, EVP_PKEY *pkey,
```
skmcgrail, line 554:
```suggestion
static bool PKCS12CreateVector(bssl::UniquePtr<PKCS12> & p12, EVP_PKEY *pkey,
```

You can just pass this by C++ reference right? Then you don't have to take the address of the UniquePtr every time you pass it in.

crypto/pkcs8/pkcs12_test.cc
```diff
@@ -618,34 +622,37 @@ TEST(PKCS12Test, Order) {
...
   ASSERT_TRUE(PKCS12CreateVector(&p12, key1.get(),
                                  {cert1.get(), cert2.get(), cert3.get()}));
-  ExpectPKCS12Parse(p12, key1.get(), cert1.get(), {cert2.get(), cert3.get()});
+  ExpectPKCS12Parse(p12.get(), key1.get(), cert1.get(),
+                    {cert2.get(), cert3.get()});
 
   ASSERT_TRUE(PKCS12CreateVector(&p12, key1.get(),
                                  {cert3.get(), cert1.get(), cert2.get()}));
-  ExpectPKCS12Parse(p12, key1.get(), cert1.get(), {cert3.get(), cert2.get()});
+  ExpectPKCS12Parse(p12.get(), key1.get(), cert1.get(),
```
skmcgrail, line 633: I guess similar to the `PKCS12CreateVector` taking the smart pointer, why have the `ExpectPKCS12Parse` not take a reference to the smart pointer and then unwrap it versus passing unwrapped to the function?

skmcgrail (approved):
Minor style feedback on tests, non-blocking.

crypto/pkcs8/pkcs12_test.cc
samuel40791765: Agree with both comments, this was mostly me trying to hack something together while allowing the original code to use the same function.

skmcgrail (approved):

crypto/pkcs8/pkcs8_x509.c
```diff
@@ -1353,3 +1369,98 @@ void PKCS12_free(PKCS12 *p12) {
...
+  if (salt == NULL) {
+    if (!RAND_bytes(mac_salt, salt_len)) {
+      goto out;
+    }
+  } else {
+    OPENSSL_memcpy(mac_salt, salt, salt_len);
+  }
+  // Match OpenSSL in using SHA-1 as the default hash function.
+  if (md == NULL) {
+    md = EVP_sha1();
```
WillChilds-Klein, line 1396: does ruby require/expect that the default here is SHA1? i feel kind of uneasy using SHA1 for HMAC due to its [poor collision resistance](https://shattered.io/)...

samuel40791765: Great question, OpenSSL versions later than 3.x actually implement SHA-256 as the default. 
* https://github.com/openssl/openssl/blob/b1775ba4c9eff94df2a6395e2430749f7a7b259d/crypto/pkcs12/p12_mutl.c#L354

I originally had plans to follow the same default as OpenSSL 3.x, but that would introduce a behavioral change/misalignment in AWS-LC since we already implement [`SHA1` as the default](https://github.com/aws/aws-lc/blob/cc9c9f04c7b7d53bb0018e8c91185d26c9ed269c/crypto/pkcs8/pkcs8_x509.c#L1300-L1302) in `PKCS12_create`. OpenSSL's `PKCS12_create` and `PKCS12_set_mac` use the same default hash function and these functions should generally align.

I'm open to changing our behavior to align with OpenSSL 3.x, but wasn't sure if we wanted to do that now or when we ever update our [`OPENSSL_VERSION_NUMBER`](https://github.com/aws/aws-lc/blob/cc9c9f04c7b7d53bb0018e8c91185d26c9ed269c/include/openssl/opensslv.h#L26) to 3.x (if we ever decide to do so).
This may introduce a behavioral change, but I'm open to updating it if we're keen on doing so? If we do decide to change it, I would prefer a separate follow up PR since it is a larger behavioral change than just `PKCS12_set_mac`.

samuel40791765: Sorry I thought I had added a TODO comment around the default hash usage to be more transparent... Not sure where that change went, but I can add it back and create a ticket to track if we aren't ready to update the default to SHA256 right now.

WillChilds-Klein: > If we do decide to change it, I would prefer a separate follow up PR since it is a larger behavioral change than just `PKCS12_set_mac`

+1

crypto/pkcs8/pkcs8_x509.c
```diff
@@ -1353,3 +1371,99 @@ void PKCS12_free(PKCS12 *p12) {
...
+  int ret = 0;
+
+  if (mac_iterations == 0) {
+    mac_iterations = 1;
+  }
+  if (salt_len == 0) {
+    salt_len = PKCS5_SALT_LEN;
+  }
+  // Generate |mac_salt| if |salt| is NULL and copy if NULL.
+  uint8_t *mac_salt = OPENSSL_malloc(salt_len);
```
WillChilds-Klein, line 1388: need to check for allocation failure here

crypto/pkcs8/pkcs8_x509.c
```diff
@@ -1353,3 +1371,99 @@ void PKCS12_free(PKCS12 *p12) {
...
+  }
+  // There's no use case for |storage| anymore, so we free early.
+  OPENSSL_free(storage);
+
+  if (!CBS_get_asn1(&in, &pfx, CBS_ASN1_SEQUENCE) || CBS_len(&in) != 0 ||
+      !CBS_get_asn1_uint64(&pfx, &version)) {
+    OPENSSL_PUT_ERROR(PKCS8, PKCS8_R_BAD_PKCS12_DATA);
+    goto out;
+  }
+  if (version < 3) {
```
WillChilds-Klein, line 1419: do we support versions > 3 here?

crypto/pkcs8/pkcs8_x509.c
```diff
@@ -1353,3 +1371,99 @@ void PKCS12_free(PKCS12 *p12) {
...
+
+  if (!CBS_get_asn1(&pfx, &authsafe, CBS_ASN1_SEQUENCE)) {
+    OPENSSL_PUT_ERROR(PKCS8, PKCS8_R_BAD_PKCS12_DATA);
+    goto out;
+  }
+  // Save contents of |authsafe| to write back before the CBS is advanced.
+  const uint8_t *orig_authsafe = CBS_data(&authsafe);
+  size_t orig_authsafe_len = CBS_len(&authsafe);
+
+  // Parse for |authsafes| which is the data that we should be running HMAC on.
```
WillChilds-Klein, line 1432: what is "authsafe"? i don't see it mentioned anywhere in the overview

crypto/pkcs8/pkcs8_x509.c
```diff
@@ -1353,3 +1371,99 @@ void PKCS12_free(PKCS12 *p12) {
...
+  if (!CBS_get_asn1(&authsafe, &content_type, CBS_ASN1_OBJECT) ||
+      !CBS_get_asn1(&authsafe, &wrapped_authsafes,
+                    CBS_ASN1_CONTEXT_SPECIFIC | CBS_ASN1_CONSTRUCTED | 0) ||
```
WillChilds-Klein, line 1435: why `| 0`?

samuel40791765: Great question, 0 is the explicit tag that corresponds to this field. We use `CBS_ASN1_CONSTRUCTED` to OR into a tag to look up the constructed bit.

Other usages of `CBS_ASN1_CONSTRUCTED` follow a similar pattern: https://github.com/search?q=repo%3Aaws/aws-lc%20CBS_ASN1_CONSTRUCTED&type=code

WillChilds-Klein: right, but `f(x) = x | 0 = x` is just the identity function. if it's a form of "documentation", that makes sense.

WillChilds-Klein (approved):

skmcgrail (approved):

PR #2134, opened 2025-01-23
Add support for PKCS7_set/get_detached

samuel40791765:
### Description of changes: 
We’re still missing two PKCS7 symbols: https://github.com/aws/aws-lc/actions/runs/12898691659/job/35966230534#step:4:4012 This also applies to past releases.

* `PKCS7_get_detached` is basically an alias to `PKCS7_is_detached`, so I've done that accordingly.
* `PKCS7_set_detached` is used to "detach" data contents from the internal `PKCS7` structure. [The OpenSSL code](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/pkcs7/pk7_lib.c#L26-L48) uses a detached field in the PKCS7 structure that’s missing in AWS-LC. This field isn't necessarily needed though, we can just detect the functionality accordingly.

### Call-outs:
N/A

### Testing:
New tests

### DIFF
--- a/crypto/pkcs7/pkcs7.c
+++ b/crypto/pkcs7/pkcs7.c
@@ -843,6 +843,32 @@ int PKCS7_is_detached(PKCS7 *p7) {
   return 0;
 }
 
+int PKCS7_set_detached(PKCS7 *p7, int detach) {
+  GUARD_PTR(p7);
+  if (detach != 0 && detach != 1) {
+    return 0;
+  }
+
+  if (PKCS7_type_is_signed(p7)) {
+    if (p7->d.sign == NULL) {
+      OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_NO_CONTENT);
+      return 0;
+    }
+    if (detach && PKCS7_type_is_data(p7->d.sign->contents)) {
+      ASN1_OCTET_STRING_free(p7->d.sign->contents->d.data);
+      p7->d.sign->contents->d.data = NULL;
+    }
+    return detach;
+  } else {
+    OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_OPERATION_NOT_SUPPORTED_ON_THIS_TYPE);
+    return 0;
+  }
+}
+
+int PKCS7_get_detached(PKCS7 *p7) {
+  return PKCS7_is_detached(p7);
+}
+
 
 static BIO *pkcs7_find_digest(EVP_MD_CTX **pmd, BIO *bio, int nid) {
   GUARD_PTR(pmd);
--- a/crypto/pkcs7/pkcs7.c
+++ b/crypto/pkcs7/pkcs7.c
@@ -846,6 +846,7 @@ int PKCS7_is_detached(PKCS7 *p7) {
 int PKCS7_set_detached(PKCS7 *p7, int detach) {
   GUARD_PTR(p7);
   if (detach != 0 && detach != 1) {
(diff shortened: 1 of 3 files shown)


crypto/pkcs7/pkcs7.c
```diff
@@ -843,6 +843,32 @@ int PKCS7_is_detached(PKCS7 *p7) {
...
+  if (PKCS7_type_is_signed(p7)) {
+    if (p7->d.sign == NULL) {
+      OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_NO_CONTENT);
+      return 0;
+    }
+    if (detach && PKCS7_type_is_data(p7->d.sign->contents)) {
+      ASN1_OCTET_STRING_free(p7->d.sign->contents->d.data);
+      p7->d.sign->contents->d.data = NULL;
+    }
+    return detach;
```
justsmth, line 861: The semantic of the `detach` parameter is unexpected. The call always "fails" (returns 0) when `detach = 0`.  

Could we detect `detach == 0` and return 0 at the top? (The difference would be that calls to `OPENSSL_PUT_ERROR` in certain situations would not be made.)
```
if (detach != 1) {
    return 0;
}
```

If the calls to `OPENSSL_PUT_ERROR` are important, we should have test coverage to confirm they are made.

crypto/pkcs7/pkcs7_test.cc
```diff
@@ -2043,3 +2043,25 @@ TEST(PKCS7Test, PKCS7PrintNoop) {
...
+  // Access the |p7|'s internal contents to verify that |PKCS7_set_detached|
+  // has the right behavior.
+  EXPECT_TRUE(p7.get()->d.sign->contents->d.data);
+  EXPECT_FALSE(PKCS7_set_detached(p7.get(), 0));
+  EXPECT_TRUE(p7.get()->d.sign->contents->d.data);
+  EXPECT_FALSE(PKCS7_set_detached(p7.get(), 2));
+  EXPECT_TRUE(p7.get()->d.sign->contents->d.data);
+  // data is "detached" when |PKCS7_set_detached| is set with 1.
+  EXPECT_TRUE(PKCS7_set_detached(p7.get(), 1));
+  EXPECT_FALSE(p7.get()->d.sign->contents->d.data);
```
justsmth, line 2066: None of the tests check for a failure condition when the parameter `detach = 1`.

crypto/pkcs7/pkcs7.c
samuel40791765: The `detach` parameter is expected to be consumed as a boolean. Ruby has a good example of the usage here: https://github.com/ruby/ruby/blob/dd863714bf377b044645ea12b4db48920d49694e/ext/openssl/ossl_pkcs7.c#L497-L500

I completely agree with the weird semantic of the `detach` parameter, especially with `0`. OpenSSL's implementation allows for other numbers to be used, but seeing that it doesn't align with how the API was meant to be used, I decided to restrict the values to only `0` and `1`.
Since the parameter value was meant to take a `boolean` integer, I do think `0` should be allowed as a value though. I'll add coverage for the unsupported operations.

crypto/pkcs7/pkcs7_test.cc
samuel40791765: Will add

justsmth (approved):

WillChilds-Klein (approved):

PR #2137, opened 2025-01-24
Optimize x86/aarch64 MD5 implementation

olivergillespie:
### Description of changes: 

(Equivalent to https://github.com/openssl/openssl/commit/ebe34f9a62630b45a825bc07a2e9cf52731e836e)

As suggested in https://github.com/animetosho/md5-optimisation?tab=readme-ov-file#dependency-shortcut-in-g-function, we can delay the dependency on 'x' by recognizing that ((x & z) | (y & ~z)) is equivalent to ((x & z) + (y + ~z)) in this scenario, and we can perform those additions independently, leaving our dependency on x to the final addition. This speeds it up around 5% on both platforms.

### Call-outs:
I applied the x86 patch manually, it's trivial, and the aarch64 patch applied cleanly after fixing the file name/path.

### Testing:
`ninja-build run_tests` on x86. `bssl speed -filter MD5` on x86 shows the expected speedup:

```
Before
Did 7169250 MD5 (16 bytes) operations in 1000024us (7169077.9 ops/sec): 114.7 MB/s
Did 1826500 MD5 (256 bytes) operations in 1000119us (1826282.7 ops/sec): 467.5 MB/s
Did 433000 MD5 (1350 bytes) operations in 1001295us (432440.0 ops/sec): 583.8 MB/s
Did 76000 MD5 (8192 bytes) operations in 1005792us (75562.3 ops/sec): 619.0 MB/s
Did 39000 MD5 (16384 bytes) operations in 1025281us (38038.4 ops/sec): 623.2 MB/s


After
Did 7457250 MD5 (16 bytes) operations in 1000004us (7457220.2 ops/sec): 119.3 MB/s
Did 1919000 MD5 (256 bytes) operations in 1000057us (1918890.6 ops/sec): 491.2 MB/s
Did 456000 MD5 (1350 bytes) operations in 1001377us (455373.0 ops/sec): 614.8 MB/s
Did 80000 MD5 (8192 bytes) operations in 1004118us (79671.9 ops/sec): 652.7 MB/s
Did 41000 MD5 (16384 bytes) operations in 1022055us (40115.3 ops/sec): 657.2 MB/s 
```

### DIFF
--- a/crypto/fipsmodule/md5/asm/md5-armv8.pl
+++ b/crypto/fipsmodule/md5/asm/md5-armv8.pl
@@ -216,165 +216,165 @@ md5_blocks_loop:
         add w9, w9, w13               // Add constant 0x49b40821
         add w9, w9, w6                // Add aux function result
         ror w9, w9, #10               // Rotate left s=22 bits
-        bic x6, x8, x17               // Aux function round 2 G(x,y,z)=((x&z)|(~z&y))
+        bic x6, x8, x17               // Aux function round 2 (~z & y)
         add w9, w8, w9                // Add X parameter round 1 B=FF(B, C, D, A, 0x49b40821, s=22, M[15])
-        and x13, x9, x17              // Aux function round 2 G(x,y,z)=((x&z)|(~z&y))
-        orr x6, x6, x13               // End aux function round 2 G(x,y,z)=((x&z)|(~z&y))
         movz x13, #0x2562             // Load lower half of constant 0xf61e2562
         movk x13, #0xf61e, lsl #16    // Load upper half of constant 0xf61e2562
         add w4, w4, w20               // Add dest value
         add w4, w4, w13               // Add constant 0xf61e2562
-        add w4, w4, w6                // Add aux function result
+        and x13, x9, x17              // Aux function round 2 (x & z)
+        add w4, w4, w6                // Add (~z & y)
+        add w4, w4, w13               // Add (x & z)
         ror w4, w4, #27               // Rotate left s=5 bits
-        bic x6, x9, x8                // Aux function round 2 G(x,y,z)=((x&z)|(~z&y))
+        bic x6, x9, x8                // Aux function round 2 (~z & y)
         add w4, w9, w4                // Add X parameter round 2 A=GG(A, B, C, D, 0xf61e2562, s=5, M[1])
-        and x13, x4, x8               // Aux function round 2 G(x,y,z)=((x&z)|(~z&y))
-        orr x6, x6, x13               // End aux function round 2 G(x,y,z)=((x&z)|(~z&y))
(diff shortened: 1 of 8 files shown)


justsmth:
@AWSjswinney

crypto/fipsmodule/md5/asm/md5-x86_64.pl
```diff
@@ -52,9 +51,9 @@ sub round2_step
 	and	$x,		%r12d		/* x & z */
 	and	$y,		%r11d		/* y & (not z) */
 	mov	$k_next*4(%rsi),%r10d		/* (NEXT STEP) X[$k_next] */
-	or	%r11d,		%r12d		/* (y & (not z)) | (x & z) */
+	add	%r11d,		$dst		/* dst += (y & (not z)) */
```
nebeid, line 54: Again here, why not do this add secondly or swap lines 51 and 52?
Also for the subsequent `mov`s, are you relying on that the processor will delay them in out-of-order processing manner?

crypto/fipsmodule/md5/asm/md5-armv8.pl
```diff
@@ -216,165 +216,165 @@
...
+        bic x6, x8, x17               // Aux function round 2 (~z & y)
         add w9, w8, w9                // Add X parameter round 1 B=FF(B, C, D, A, 0x49b40821, s=22, M[15])
-        and x13, x9, x17              // Aux function round 2 G(x,y,z)=((x&z)|(~z&y))
-        orr x6, x6, x13               // End aux function round 2 G(x,y,z)=((x&z)|(~z&y))
         movz x13, #0x2562             // Load lower half of constant 0xf61e2562
         movk x13, #0xf61e, lsl #16    // Load upper half of constant 0xf61e2562
         add w4, w4, w20               // Add dest value
         add w4, w4, w13               // Add constant 0xf61e2562
-        add w4, w4, w6                // Add aux function result
+        and x13, x9, x17              // Aux function round 2 (x & z)
```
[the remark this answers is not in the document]
olivergillespie: If I recall, there was no special reason to move it closer, it just felt more natural to group these related operations together. The only reason for speedup is shortening the dependency path for 'x' (here, 'x13'), which is the longest path. Before, we had three operations (and, or, add). Now we have two (and, add).

Yes, I think this relies on effective out-of-order execution.

crypto/fipsmodule/md5/asm/md5-x86_64.pl
olivergillespie: If I understand your question correctly, yes it's similar to the above - we rely on the processor to order the independent instructions efficiently. I don't think you'd see any performance change from manual reordering.

nebeid (approved):

torben-hansen (approved):

PR #2142, opened 2025-01-28
Support for ML-DSA public key generation from private key

jakemas:
### Issues:
Resolves #CryptoAlg-2868

### Description of changes: 

It is often useful when serializing asymmetric key pairs to populate both the public and private elements, given only the private element. For this to be possible, an algorithm utility function is often provided to derive key material. ML-DSA does not support this in the reference implementation.

#### Background ML-DSA keypairs 

An ML-DSA private key is constructed of the following elements: (ref https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf)
```
sk = (
      rho, // public random seed (32-bytes)
      tr,  // public key hash (64-bytes)
      key, // private random seed (32-bytes) (utilized during sign)
      t0,  // polynomial vector: encodes the least significant bits of public-key polynomial t, facilitating certain computational efficiencies.
      s1,  // secret polynomial vectors. These vectors contain polynomials with coefficients in a specified range, 
      s2.  // serving as the secret components in the lattice-based structure of ML-DSA.
)
```

An ML-DSA public key is constructed of the following elements:
```
pk = (
      rho, // public random seed (32-bytes)
      t1.  // compressed representation of the public key polynomial 
)
```

- The vector t is decomposed into two parts:
- `t1`: Represents the higher-order bits of `t`.
- `t0`: Represents the lower-order bits of `t`.

One can see that to reconstruct the public key from the private key, one must:
1. Extract all elements from `sk`, using the existing function in `/ml_dsa_ref/packing.c`: `ml_dsa_unpack_sk`
    1. This will provide `sk = (rho, tr, key, t0, s1, s2)`.
2. Reconstruct `A` using `rho` with the existing function in `/ml_dsa_ref/polyvec.c`: `ml_dsa_polyvec_matrix_expand`
3. Reconstruct `t` from `t = A*s1 + s2`
4. Drop `d` lower bits from `t` to get `t1`
5. Pack `rho`, `t1` into public key.
6. Verify `pk` matches expected value, by comparing SHAKE256(pk) + `tr` (unpacked from secret key).

This has been implemented in `ml_dsa_pack_pk_from_sk` -- not tied to the name, just using what I've seen so far in common nomenclature.

As the values of `d` differ for each parameter set of ML-DSA, we must create packing functions for each parameter size. As such, `ml_dsa_44_pack_pk_from_sk``, `ml_dsa_65_pack_pk_from_sk``, and `ml_dsa_87_pack_pk_from_sk`` have been added to `ml_dsa.h` to serve as utility functions in higher level EVP APIs.

### Call-outs:

The scope of this PR is only the algorithm level, using these functions for useful tasks such as populating the public key automatically on private key import -- will be added in subsequent PRs.

### Testing:
A new test has been added to `PQDSAParameterTest`, namely, `KeyConsistencyTest` that will assert that packing the key is successful, and that the key produced matches the original public key.

### DIFF
--- a/crypto/ml_dsa/ml_dsa_ref/packing.c
+++ b/crypto/ml_dsa/ml_dsa_ref/packing.c
@@ -2,6 +2,63 @@
 #include "packing.h"
 #include "polyvec.h"
 #include "poly.h"
+#include "../../fipsmodule/sha/internal.h"
+
+/*************************************************
+* Name:        ml_dsa_pack_key
+*
+* Description: Takes a private key and constructs the corresponding public key.
+*              The hash of the contructed public key is then compared with
+*              the value of tr unpacked from the provided private key.
+*
+* Arguments:   - ml_dsa_params: parameter struct
+*              - uint8_t pk: pointer to output byte array
+*              - uint8_t sk: pointer to byte array containing bit-packed sk
+*
+* Returns 0 (when SHAKE256 hash of constructed pk matches tr)
+**************************************************/
+int ml_dsa_pack_key(ml_dsa_params *params,
+                     uint8_t *pk,
+                     const uint8_t *sk)
+{
+  uint8_t rho[ML_DSA_SEEDBYTES];
+  uint8_t tr[ML_DSA_TRBYTES];
+  uint8_t tr_validate[ML_DSA_TRBYTES];
+  uint8_t key[ML_DSA_SEEDBYTES];
+  polyvecl mat[ML_DSA_K_MAX];
+  polyvecl s1;
+  polyveck s2, t1, t0;
+
+  //unpack sk
+  ml_dsa_unpack_sk(params, rho, tr, key, &t0, &s1, &s2, sk);
+
+  // generate matrix A
+  ml_dsa_polyvec_matrix_expand(params, mat, rho);
+
+  // convert s1 into ntt representation
+  ml_dsa_polyvecl_ntt(params, &s1);
+
+  // construct  t1 = A * s1
+  ml_dsa_polyvec_matrix_pointwise_montgomery(params, &t1, mat, &s1);
+
+  // reduce t1 modulo field
+  ml_dsa_polyveck_reduce(params, &t1);
+
+  // take t1 out of ntt representation
+  ml_dsa_polyveck_invntt_tomont(params, &t1);
+
+  // construct t1 = A * s1 + s2
+  ml_dsa_polyveck_add(params, &t1, &t1, &s2);
+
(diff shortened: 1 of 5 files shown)


justsmth (approved):
Thanks for getting this change up so quickly!

nebeid (commented):
May I suggest renaming to `pack_pk_from_sk`? I understand they're internal functions but they're exported and so renaming them later may be more difficult.

Nit: comments in https://github.com/aws/aws-lc/blob/4fbcd100235278e81320704c4f65c42eeaa5410f/crypto/ml_dsa/ml_dsa_ref/packing.c#L182-L187 should not have `const`.

jakemas:
> May I suggest renaming to `pack_pk_from_sk`? I understand they're internal functions but they're exported and so renaming them later may be more difficult.

Thank you, changed to `ml_dsa_pack_pk_from_sk` in [bf01ead](https://github.com/aws/aws-lc/pull/2142/commits/bf01ead882989d5a5b27c72d15cfbbe16a85a172).

> 
> Nit: comments in
> 
> https://github.com/aws/aws-lc/blob/4fbcd100235278e81320704c4f65c42eeaa5410f/crypto/ml_dsa/ml_dsa_ref/packing.c#L182-L187
> 
> should not have `const`.

Got them, also fixed in
[bf01ead](https://github.com/aws/aws-lc/pull/2142/commits/bf01ead882989d5a5b27c72d15cfbbe16a85a172)

justsmth (approved):

nebeid (approved):

crypto/ml_dsa/ml_dsa_ref/packing.c
```diff
@@ -2,6 +2,63 @@
...
+  // construct t1 = A * s1 + s2
+  ml_dsa_polyveck_add(params, &t1, &t1, &s2);
+
+  // cxtract t1 and write public key
+  ml_dsa_polyveck_caddq(params, &t1);
+  ml_dsa_polyveck_power2round(params, &t1, &t0, &t1);
+  ml_dsa_pack_pk(params, pk, rho, &t1);
+
+  // we hash pk to reproduce tr, check it with unpacked value to verify
+  SHAKE256(pk, params->public_key_bytes, tr_validate, ML_DSA_TRBYTES);
```
vdukhovni, line 59: Another sanity check could be to verify that the just recomputed $`t_0`$ matches the original $`t_0`$ from the private key. If it doesn't, the original private key is invalid or the computation malfunctioned.

PR #2144, opened 2025-01-28
Add EVP API Support for ED25519ph

skmcgrail:
### Description of changes: 
Add support for HashEdDSA (Ed25519ph) to EVP APIs with support for setting the associated signature context. Opting not to expose Ed25519ctx explicitly at this time since there hasn't been an ask for it. Worse case scenario a user desiring it will have access to the low-level APIs where it is available.


crypto/evp_extra/p_ed25519ph.c
```diff
@@ -0,0 +1,175 @@
...
+    OPENSSL_PUT_ERROR(EVP, EVP_R_NOT_A_PRIVATE_KEY);
+    return 0;
+  }
+
+  if (sig == NULL) {
+    *siglen = 64;
+    return 1;
+  }
+
+  if (*siglen < 64) {
```
torben-hansen, line 84: could you one of the constants instead of 64 magic number. just a suggestion.

crypto/evp_extra/p_ed25519_asn1.c
```diff
@@ -257,12 +257,40 @@ static int ed25519_size(const EVP_PKEY *pkey) { return 64; }
...
+    NULL /* param_copy */,
+    NULL /* param_cmp */,
+    ed25519_free,
+};
+
+const EVP_PKEY_ASN1_METHOD ed25519ph_asn1_meth = {
+    EVP_PKEY_ED25519PH,
+    {0x2b, 0x65, 0x70},
+    3,
+    pem_str,
```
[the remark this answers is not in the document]
skmcgrail: No, Ed25519ph is not a type of key, its a signing method. That and there is no official OID either that's been adopted. I don't want to arbitrarily introduce a key type / format for a signing method that nobody else supports.

torben-hansen (dismissed):

skmcgrail:
Reworked the per out meeting to remove the keygen and ASN.1 serialization/deserialization support for ED25519ph. Should simplify the review some now.

crypto/evp_extra/p_ed25519_asn1.c
```diff
@@ -281,3 +281,28 @@ const EVP_PKEY_ASN1_METHOD ed25519_asn1_meth = {
     NULL /* param_cmp */,
     ed25519_free,
 };
+
+const EVP_PKEY_ASN1_METHOD ed25519ph_asn1_meth = {
+    EVP_PKEY_ED25519PH,
+    {}, /* oid */
```
torben-hansen, line 287: Windows compilers doesn't seem to like this notation.

torben-hansen (approved):

crypto/evp_extra/evp_test.cc
```diff
@@ -1529,3 +1529,280 @@ TEST(EVPTest, PKEY_asn1_find_str) {
...
+    ASSERT_TRUE(rk);
+    pubkey.reset(rk);
+    ASSERT_EQ(EVP_PKEY_ED25519PH, EVP_PKEY_id(pubkey.get()));
+
+    bssl::ScopedCBB temp;
+    ASSERT_TRUE(CBB_init(temp.get(), 0));
+    ASSERT_FALSE(EVP_marshal_public_key(temp.get(), pubkey.get()));
+  }
+
+  // prehash signature w/ context gen and verify
```
nebeid, line 1606: Does prehash signature with and without context work with the single-shot `EVP_DigestSign/Verify`?

crypto/evp_extra/p_ed25519_asn1.c
```diff
@@ -281,3 +281,28 @@ const EVP_PKEY_ASN1_METHOD ed25519_asn1_meth = {
     NULL /* param_cmp */,
     ed25519_free,
 };
+
+const EVP_PKEY_ASN1_METHOD ed25519ph_asn1_meth = {
```
nebeid, line 285: Is this a place holder for differentiating changes that may come to ed25519ph or there is something we need from this method in this PR that is different from that of ed25519?

skmcgrail: This is needed even though we don't support the ASN.1 PKCS#8 or SubjectPublicKeyInfo format serialization/deserialization with Ed25519ph. This method table is required though for the functionality of setting the raw public key or private keys on an EVP structure. Thus is why it is here.

crypto/evp_extra/evp_test.cc
skmcgrail: Yes it would work through that interface, those functions delegate down to the same `EVP_DigestSignUpdate` and `EVP_DigestSignFinal` tested here. So not sure if we need to test this specific entrypoint?

nebeid: You're right. No need then.

nebeid (approved):

nebeid (approved):

torben-hansen (approved):

PR #2145, opened 2025-01-28
Support keypair calculation for PQDSA PKEY

jakemas:
### Issues:
Resolves #CryptoAlg-2868

### Description of changes: 
Following from the first part of this PR to add low level support for ML-DSA public key generation from private keys. This PR uses the new `pqdsa` method `pqdsa->pqdsa_pack_pk_from_sk`. We supply `ml_dsa_44_pack_pk_from_sk`, `ml_dsa_65_pack_pk_from_sk`, and `ml_dsa_87_pack_pk_from_sk` from `ml_dsa.h` to provide this functionality for ML-DSA.

As we use `EVP_parse_private_key` to import `PQDSA` keys into AWS-LC, we need to modify the `PQDSA` asn.1 private key decoding function to additionally populate the corresponding public key.

The `pqdsa_priv_decode` function in `p_pqdsa_asn1.c` has been modified to attempt to generate a public key from the provided secret key, should the method be defined for that `pqdsa` structure. As ML-DSA is the only current `PQDSA`, the library currently does implement methods for all `PQDSA` PKEY types.

We populate the methods as `NULL` and will modify the G.test that expects the public key to be generated. We can make this change simply if/when we add to the `PQDSA` capabilites with an algorithm that doesn't support `pk` from `sk`.

### Testing:
To test functionality, I have adapted the `PQDSAParameterTest` G.test `MarshalParse` to now verify that the public key is populated after calling `EVP_parse_private_key`. We then verify that the public key calculated is equal to the original public key.


crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -149,7 +149,40 @@ static int pqdsa_priv_decode(EVP_PKEY *out, CBS *params, CBS *key, CBS *pubkey)
...
   }
-  return PQDSA_KEY_set_raw_private_key(out->pkey.pqdsa_key,CBS_data(key));
+
+  // Set the private key
+  if (!PQDSA_KEY_set_raw_private_key(out->pkey.pqdsa_key, CBS_data(key))) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_DECODE_ERROR);
+    return 0;
+  }
+
+  // Calculate public key from private key if method exists
```
andrewhop, line 159: How expensive is this calculation? What customer workflow needs the public key given the private key? Could this be lazily initialized?

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -149,7 +149,40 @@ static int pqdsa_priv_decode(EVP_PKEY *out, CBS *params, CBS *key, CBS *pubkey)
...
+  if (out->pkey.pqdsa_key->pqdsa->method->pqdsa_pack_pk_from_sk != NULL) {
```
justsmth, line 160: Shouldn't we always return failure (`0`) if we're unable to initialize the public key? I assume the `pqdsa_pack_pk_from_sk` function ptr should always be non-NULL here (line 360) so this condition would never fail?

Upon seeing a successful return code, a caller assumes the EVP_PKEY is initialized and might subsequently SEGFAULT on a call to (e.g.) `EVP_PKEY_cmp`.

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -149,7 +149,40 @@ static int pqdsa_priv_decode(EVP_PKEY *out, CBS *params, CBS *key, CBS *pubkey)
...
+    // attempt to construct the public key from priv
+    if (!out->pkey.pqdsa_key->pqdsa->method->pqdsa_pack_pk_from_sk(
+            public_key, CBS_data(key))) {
+      OPENSSL_free(public_key);
+      OPENSSL_PUT_ERROR(EVP, EVP_R_DECODE_ERROR);
+      return 0;
+    }
+    // set the public key
+    int ret = PQDSA_KEY_set_raw_public_key(out->pkey.pqdsa_key, public_key);
+    OPENSSL_free(public_key);
```
andrewhop, line 177: Since public_key was allocated internally can the PKEY just take ownership of it without having to copy it?

crypto/evp_extra/p_pqdsa_asn1.c
justsmth: Other types of EVP_PKEY (which have public/private components) ensure that the public key is populated whenever a private key is deserialized. For example, see [EC](https://github.com/aws/aws-lc/blob/80d9c1e036d3cbc81b8bacd8eb83dece45c28039/crypto/ec_extra/ec_asn1.c#L138) and [ED25519](https://github.com/aws/aws-lc/blob/3662fc13a4c4bccadb40a55218d4e71a491f1e5b/crypto/evp_extra/p_ed25519_asn1.c#L52).

Other EVP_PKEY_xxx methods (such as [EVP_PKEY_cmp](https://github.com/aws/aws-lc/blob/454d6576bf360ce88cd38b440fae8dc7e3cbfc16/crypto/fipsmodule/evp/evp.c#L151)) assume that the public key is populated for any EVP_PKEY that has functions available for public keys. (In fact `EVP_PKEY_cmp` can currently SEGFAULT for EVP_PKEY_PQDSA due to this.)

There might be other places for this computation to take place, but it needs to be done when deserializing a private key.

justsmth (approved):

crypto/evp_extra/p_pqdsa_asn1.c
jakemas: Ah I see. I was trying to prevent against a failure if the ptr is non-null, as you say. Removed this check in [cd8da04](https://github.com/aws/aws-lc/pull/2145/commits/cd8da04f33b1baacbe7857532ea260b3c1669bc3) and cleaned up the surrounding logic to only return `SUCESS` when both priv and pub are populated.

crypto/evp_extra/p_pqdsa_asn1.c
jakemas: I don't see this as being possible, as the size of the `public_key` is dynamic, and defined by `out->pkey.pqdsa_key->pqdsa->public_key_len`. Am I missing something?

justsmth (dismissed):

crypto/evp_extra/p_pqdsa_asn1.c
andrewhop: Yeah, you allocate the key on L162, looking at `PQDSA_KEY_set_raw_public_key` it just calls `OPENSSL_memdup` which allocates a new buffer and memcpy the data into that new buffer, then this function frees `public_key`. Why not store `public_key` in `out->pkey.pqdsa_key->public_key`?

skmcgrail (approved):

crypto/evp_extra/p_pqdsa_asn1.c
andrewhop: Ahhhh good point, looking at ED25519 a little more should we add raw support?
1. priv_decode -> ed25519_priv_decode decodes the private key and calls ed25519_set_priv_raw
2. set_priv_raw -> ed25519_set_priv_raw calculates the public key from the private key, and compares it to the supplied public key if provided

Here in ML-DSA:
1. priv_decode -> pqdsa_priv_decode decodes the private key and calculates the public key and sets it
2. set_prive_raw -> not implemented for pqdsa

jakemas: Some more background: `set_priv_raw` is not implemented for ML-DSA or ML-KEM. In ED25519 things are a little different, as there are no "parameter sets" for ED25519, it's just one algorithm. In ML-DSA/ML-KEM we have multiple NIDs to define algorithm versions. 

As such, we implement `EVP_PKEY_pqdsa_new_raw_private_key` at https://github.com/aws/aws-lc/blob/main/include/openssl/evp.h#L960-L970 just like for KEMs. I suppose your suggestion is then to remove `EVP_PKEY_pqdsa_new_raw_private_key` / `EVP_PKEY_pqdsa_new_raw_public_key` and support via `EVP_PKEY_new_raw_private_key`/ `EVP_PKEY_new_raw_public_key`?

crypto/evp_extra/p_pqdsa_asn1.c
jakemas: got it! [54224d2](https://github.com/aws/aws-lc/pull/2145/commits/54224d23a3af931471603b35e308d46f3e9e0349)

justsmth (approved):

crypto/evp_extra/p_pqdsa_asn1.c
jakemas: Yeah I considered that too, it works, but it does present an issue should another algorithm be added that has the same length of raw keys --  which may sound unlikely, but would present issues with HashML-DSA that has a different OID. I do have some ideas -- but likely out of scope for this PR.

andrewhop (approved):

PR #2146, opened 2025-01-29
Update benchmark to skip chunk sizes that doesn't work with the algorithm

andrewhop:
### Description of changes: 
AES XTS requires at least 16 bytes of input, all other algorithms support 1 or more bytes. This change also updates the remaining algorithm benchmarks to support any sized chunks.

### Testing:
Updated run_benchmark_build_tests.sh to test every libcrypto benchmark with 1 to 20,000 bytes.

### DIFF
--- a/tests/ci/run_benchmark_build_tests.sh
+++ b/tests/ci/run_benchmark_build_tests.sh
@@ -71,7 +71,7 @@ function build_boringssl {
 # has changes in speed.cc that could affect the comparison of the FIPS to non-FIPS, or if new
 # algorithms have been added to speed.cc
 build_aws_lc_fips
-"${BUILD_ROOT}/tool/bssl" speed -timeout_ms 10
+"${BUILD_ROOT}/tool/bssl" speed -timeout_ms 10 -chunks 1,2,16,256,20000
 
 build_aws_lc_branch fips-2021-10-20
 build_aws_lc_branch fips-2022-11-02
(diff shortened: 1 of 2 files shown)


tests/ci/run_benchmark_build_tests.sh
```diff
@@ -94,15 +94,15 @@ open32:${install_dir}/openssl-${openssl_3_2_branch};\
...
+LD_LIBRARY_PATH="${install_dir}/aws-lc-fips-2022-11-02/lib" "${BUILD_ROOT}/tool/aws-lc-fips-2022" -timeout_ms 10 -chunks 1,2,16,256,20000
+LD_LIBRARY_PATH="${install_dir}/aws-lc-fips-2024-09-27/lib" "${BUILD_ROOT}/tool/aws-lc-fips-2022" -timeout_ms 10 -chunks 1,2,16,256,20000
+LD_LIBRARY_PATH="${install_dir}/openssl-${openssl_1_0_2_branch}/lib" "${BUILD_ROOT}/tool/open102" -timeout_ms 10 -chunks 1,2,16,256,20000
+LD_LIBRARY_PATH="${install_dir}/openssl-${openssl_1_1_1_branch}/lib" "${BUILD_ROOT}/tool/open111" -timeout_ms 10 -chunks 1,2,16,256,20000
+LD_LIBRARY_PATH="${install_dir}/openssl-${openssl_3_1_branch}/lib64" "${BUILD_ROOT}/tool/open31" -timeout_ms 10 -chunks 1,2,16,256,20000
+LD_LIBRARY_PATH="${install_dir}/openssl-${openssl_3_2_branch}/lib64" "${BUILD_ROOT}/tool/open32" -timeout_ms 10 -chunks 1,2,16,256,20000
+LD_LIBRARY_PATH="${install_dir}/openssl-${openssl_master_branch}/lib64" "${BUILD_ROOT}/tool/openmaster" -timeout_ms 10 -chunks 1,2,16,256,20000
+LD_LIBRARY_PATH="${install_dir}/boringssl" "${BUILD_ROOT}/tool/boringssl" -timeout_ms 10 -chunks 1,2,16,256,20000
```
justsmth, line 105: NP: Could these chunk sizes just be set to the default so that we don't have to specify them for even invocation?

justsmth (approved):

tests/ci/run_benchmark_build_tests.sh
andrewhop: I don't want to change the actual default chunk size, I picked these values to test interesting inputs that previously caused the benchmark to fail.

torben-hansen (approved):

PR #2147, opened 2025-01-29
Flip to new randomness generation backend and remove the existing

torben-hansen:
### Description of changes: 

This PR removes the existing randomness generation and replaces it with the new one. It does a number of things to do that:
* removes "unsafe buffering" mode. This is not a thing in the new implementation and was never used.
* move `rdrand` implementation to the entropy source implementation file
* removes the ability to specify the "entropy source" at compile time. The flag is retained for build backwards compatibility. There is only a single (validable) entropy source left, and that is jitter entropy. (other sources are used in the randomness generation workflow, but not validatable).
* passive entropy is removed completely.
* the PRNG model test is disabled until I fix it.
* `RAND_bytes_with_user_prediction_resistance()` is introduced as a public API. It replaces the internal `RAND_bytes_with_additional_data()`.
* correct sandbox test to account for `getrandom` usage.
* disable state validation in `rand_bytes_core` when snapsafe testing is active. Otherwise, we have threading issues.

### Call-outs:

Future PRs will expand tests and re-enable the PRNG model test.

### Testing:

Existing testing now uses the new randomness generation backend.


crypto/fipsmodule/rand/rand.c
```diff
@@ -1,598 +1,542 @@
-/* Copyright (c) 2014, Google Inc.
```
torben-hansen, line 1: The majorly confusing diff is because I copied all source from `new_rand.c` to `rand.c`. The new implementation now lives in `rand.c` (this file).

skmcgrail (approved):

smittals2 (approved):

smittals2 (approved):

crypto/fipsmodule/rand/entropy/entropy_sources.c
```diff
@@ -85,3 +86,47 @@ int rndr_multiple8(uint8_t *buf, const size_t len) {
...
+#if defined(OPENSSL_X86_64) && !defined(OPENSSL_NO_ASM)
+
+// rdrand maximum retries as suggested by:
+// Intel® Digital Random Number Generator (DRNG) Software Implementation Guide
+// Revision 2.1
+// https://software.intel.com/content/www/us/en/develop/articles/intel-digital-random-number-generator-drng-software-implementation-guide.html
+#define RDRAND_MAX_RETRIES 10
+
+OPENSSL_STATIC_ASSERT(RDRAND_MAX_RETRIES > 0, rdrand_max_retries_must_be_positive)
+#define CALL_RDRAND_WITH_RETRY(rdrand_func, fail_ret_value)       \
```
justsmth, line 99: NP: I don't see any need for `fail_ret_value`. It's always 0.

util/all_tests.json
```diff
@@ -87,7 +87,7 @@
   },
   {
     "comment": "Run RAND test suite without fork detection",
-    "cmd": ["crypto/crypto_test", "--fork_unsafe_buffering", "--gtest_filter=RandTest.*:-RandTest.Fork"],
+    "cmd": ["crypto/crypto_test", "BORINGSSL_IGNORE_FORK_DETECTION=1", "--gtest_filter=RandTest.*:-RandTest.Fork"],
```
justsmth, line 90: Since `BORINGSSL_IGNORE_FORK_DETECTION` is an environment variable, shouldn't it be listed on an `"env"` key?

crypto/fipsmodule/rand/entropy/entropy_sources.c
```diff
@@ -85,3 +86,48 @@ int rndr_multiple8(uint8_t *buf, const size_t len) {
...
+    OPENSSL_memcpy(buf + len_multiple8, rand_buf, remainder);
+  }
+
+  return 1;
+}
+
+#else
+
+int rdrand(uint8_t *buf, const size_t len) {
+  return 0;
```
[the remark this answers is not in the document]
justsmth: Why not both `abort()` and have (an unreachable) return 0?

justsmth (approved):

PR #2148, opened 2025-01-29
CAST and PCT for ML-DSA

jakemas:
### Issues:
Resolves #CryptoAlg-2886

Related to https://github.com/aws/aws-lc/pull/1846 and https://github.com/aws/aws-lc/pull/1969

### Description of changes: 
As part of the Implementation Guidance for FIPS 140-3 and the Cryptographic Module Validation Program validation there is requirement  that states:

### Pairwise Consistency Tests (PCT)

Per the I.G guidance a PCT shall be conducted for every generated public and private key pair for the applicable approved algorithm.

> a pairwise consistency test (PCT) shall be conducted for every generated public and private key pair for the applicable approved algorithm. For approved algorithms from FIPS 186-5, SP 800-56Arev3 or SP 800-56Brev2, or FIPS 204,  if at the time a PCT on a key pair is performed it is known whether the keys will be used in a key agreement scheme, digital signature algorithm or to perform a key transport, then the PCT shall be performed consistent with the intended use of the keys (i.e., TE10.35.02 for signatures)

As such, we implement `VE10.35.02` of ISO/IEC 24759:2017. A simple test that generates an ML-DSA keypair, signs a message, and verifies the signature.

```
static int ml_dsa_keypair_pct(ml_dsa_params *params,
                              uint8_t *pk,
                              uint8_t *sk) {
  uint8_t signature[MLDSA87_SIGNATURE_BYTES];
  ml_dsa_sign(params, signature, &params->bytes, NULL, 0, NULL, 0, sk);
  return ml_dsa_verify(params, signature, params->bytes, NULL, 0, NULL, 0, pk) == 0;
}
```
In which we:
- assign the maximum signature size of bytes (`params->bytes` has the exact size but we can't dynamically allocate and we want to spent as little time in this function as possible -- allocate [max KEM size here](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/ml_kem/ml_kem_ref/kem.c#L20)
- sign `NULL` (quick and works)
- return verify result

This is called within the internal key generation function if defined `AWSLC_FIPS`
```
#if defined(AWSLC_FIPS)
  // Abort in case of PCT failure.
  if (!ml_dsa_keypair_pct(params, pk, sk)) {
    BORINGSSL_FIPS_abort();
  }
#endif
  return 0;
}

```

> _"This effectively triples the runtime of key generation."_
-Dusan Kostic https://github.com/aws/aws-lc/pull/1969

### Cryptographic Algorithm Self Test CAST)

Per the I.G guidance a CAST using a KAT is required for ML-DSA key generation

>if the module implements ML-DSA key generation, the module shall have an ML-DSA key generation
CAST. The ML-DSA key generation takes no inputs and outputs a public key (pk) and private key
(sk). The CAST shall use the ML-DSA key generation algorithm (i.e., Algorithm 1 in FIPS 204),
and for a KAT, using a fixed/predetermined random seed (ξ) value, to compare the resulting outputs
to the pre-computed values of pk and sk.

This is implemented by the new addition to the `self_check.c` test suite with test `boringssl_self_test_ml_dsa` which performs the following:

```
  if (!ml_dsa_44_keypair_internal_no_self_test(public_key, private_key, kMLDSAKeyGenSeed) ||
      !check_test(kMLDSAKeyGenPublicKey, public_key, sizeof(public_key), "ML-DSA keyGen public") ||
      !check_test(kMLDSAKeyGenPrivateKey, private_key, sizeof(private_key), "ML-DSA keyGen private")) {
    fprintf(stderr, "ML-DSA-44 KeyGen failed.\n");
    goto err;
  }
```
- Key generation using the new name for the core internal key generation from seed function (our wrapper to FIPS 204: Algorithm 6 ML-DSA.KeyGen_internal)  `ml_dsa_44_keypair_internal_no_self_test` with provided seed `kMLDSAKeyGenSeed` which is from the NIST ACVP KAT [Keygen Seed `tgId = 1` `tcId = 1`](https://github.com/usnistgov/ACVP-Server/blob/master/gen-val/json-files/ML-DSA-keyGen-FIPS204/prompt.json#L15)
- A `check_test` on the output `public_key` from that function, with public key matching that of the expected output at [public key `tgId = 1` `tcId = 1`](https://github.com/usnistgov/ACVP-Server/blob/master/gen-val/json-files/ML-DSA-keyGen-FIPS204/expectedResults.json#L13)
- A `check_test` on the output `private_key` from that function, with private key matching that of the expected output at [private key `tgId = 1` `tcId = 1`](https://github.com/usnistgov/ACVP-Server/blob/master/gen-val/json-files/ML-DSA-keyGen-FIPS204/expectedResults.json#L14)

Per the I.G guidance a CAST using a KAT is required for ML-DSA signature generation

>if the module implements digital signature generation, the module shall have an ML-DSA digital
signature generation CAST for either the “hedged” or “deterministic” algorithm (per FIPS 204
Section 5.2 and Algorithm 2) . ML-DSA signature generation algorithm takes as input a private key
(sk) and a message (M), and it outputs a signature (σ). The CAST shall use the ML-DSA signature
generation algorithm (i.e., Algorithm 2 in FIPS 204), and for a KAT, using fixed/predetermined sk
and M values, to compare the resulting outputs to the pre-computed value (σ). This requires the
randomization parameter rnd be fixed for the “hedged” algorithms (e.g., all zeros).

```
  if (!ml_dsa_44_sign_internal_no_self_test(private_key, signature, &sig_len, kMLDSASignPlaintext,
                                            mlen_int, NULL, 0, kMLDSASigGenSeed) ||
      !check_test(kMLDSASignSignature, signature, sizeof(signature), "ML-DSA SigGen signature")) {
    fprintf(stderr, "ML-DSA-44 sign failed.\n");
    goto err;
  }

```
- Signature generation using the newly named implementation of FIPS 204: Algorithm 7 ML-DSA.Sign_internal `ml_dsa_44_sign_internal_no_self_test`. The I.G states

```
  // ISO/IEC 19790:2012 Section 7.10.3.3 15. page 84 Requires the randomization
  // parameter rnd be fixed for the “hedged” algorithms (e.g., all zeros).
``` 
We implement "hedged" in aws-lc, rather than deterministic signatures (hedged short for "hedging your bets" and using both an RNG and hashing input from the message).
- The per message signing seed is `kMLDSASigGenSeed` is all zeros.
- The plaintext is 32 bytes of all zeros.
- The signature produced is checked against the precomputed signature `kMLDSASignSignature` for that message and private key. We are not able to use the same  NIST ACVP vector for this, as NIST do not provide vectors for the same public private key pairs between keygen and sigGen. We could load an additional public/private key pair from a known NIST CAVP vector, then we would have a known signature, but would have to store 2 public/private key pairs in the file (which feels obtuse). FIPS 140-3 doesn't state that the KAT has to be from CAVP in the implementation guidance, but if this isn't critical to performance, I can happily add a seed/pub/sig vector from the NIST ACVP KATs.

Per the I.G guidance a CAST using a KAT is required for ML-DSA signature verification

>• if the module implements digital signature verification, the module shall have an ML-DSA digital
signature verification CAST. The ML-DSA signature verification algorithm takes as input a public
key (pk), a message (M), and a signature (σ), and outputs a Boolean value (i.e., a value that is true if
the signature is valid with respect to the message and public key, and false if the signature is invalid).
The CAST shall use the ML-DSA signature verification algorithm (i.e., Algorithm 3 in FIPS 204),
and for a KAT, using a fixed/predetermined pk, M, and σ values, to verify the signature (i.e., returns
true).

```
  if (!ml_dsa_44_verify_internal_no_self_test(public_key, kMLDSASignSignature, sig_len, kMLDSASignPlaintext,
                                              mlen_int, NULL, 0)) {
    fprintf(stderr, "ML-DSA-44 verify failed.\n");
    goto err;
    }
```
- verify is simple and just makes sure the provided signature `kMLDSASignSignature` successfully verifies the provided plaintext `kMLDSASignPlaintext` and `public_key`.

---
### Test Configuration

For the above tests, only one ML-DSA parameter set variant needs to be tested:

>Note26: The above CASTs shall be performed on at least one of the following parameter-sets for MLDSA that are implemented in the approved mode: ML-DSA-44, ML-DSA-65, ML-DSA-87.

Also note that only one variant of pre-hash (ExternalMu-ML-DSA) and pure mode needs to be tested:
>Note 25: If both ML-DSA signature verification and pre-hash ML-DSA signature verification
(Section 5.4 and Algorithm 5 in FIPS 204) are implemented and they both share the same underlying
cryptographic implementations, a CAST on either one of them is sufficient to meet the CAST
requirements for both. Note33 of this IG would apply to ensure all the underlying approved
algorithms used within both algorithms (e.g., the hash or XOF function used to compute the message
digest) are covered by a CAST.

### Call-outs:
For reviewers:
- I noticed in the ML-KEM PCT we use the external encap/decap functions that generate randomness then call internals, it would be quicker to use the empty string as a seed and call the internal functions -- but is this compliant with the spec? ISO/IEC don't make the distinction.
- I tested the CASTs fail gracefuly by flipping some bits in `kMLDSA*`, and also by putting in the seed from [the IETF standard](https://datatracker.ietf.org/doc/draft-ietf-lamps-dilithium-certificates/) and verifying the key pair

### Testing:
The ML-DSA self-tests have been added to `--gtest_filter=SelfTests.`


crypto/ml_dsa/ml_dsa.h
```diff
@@ -36,6 +36,10 @@ OPENSSL_EXPORT int ml_dsa_44_keypair(uint8_t *public_key,
 OPENSSL_EXPORT int ml_dsa_44_pack_pk_from_sk(uint8_t *public_key,
                                              const uint8_t *private_key);
 
+OPENSSL_EXPORT int ml_dsa_44_keypair_internal_no_self_test(uint8_t *public_key,
```
andrewhop, line 39: Do these new functions need to be exported? I don't see anything outside of libcrypto calling them.

crypto/ml_dsa/ml_dsa.c
```diff
@@ -271,25 +327,33 @@ int ml_dsa_extmu_65_verify_internal(const uint8_t *public_key /* IN */,
...
+}
 
+int ml_dsa_87_keypair_internal_no_self_test(uint8_t *public_key   /* OUT */,
+                                            uint8_t *private_key  /* OUT */,
+                                            const uint8_t *seed   /* IN */) {
   ml_dsa_params params;
   ml_dsa_87_params_init(&params);
-  return ml_dsa_pack_pk_from_sk(&params, public_key, private_key) == 0;
+  return ml_dsa_keypair_internal(&params, public_key, private_key, seed) == 0;
 }
```
andrewhop, line 349: Do we need to do this? We only need the _no_self_test functions to avoid a infinite loop of the self test calling itself. Here your self test only tests ML-DSA44 so we should only need it for that size. The other 65/87 functions can call `boringssl_ensure_ml_dsa_self_test()` and it will run the 44 no self test version without hitting a loop.

andrewhop (commented):
This change looks like a CAST for ML-DSA which lazily runs once on first use and on demand calling the self test like we did for ML-KEM https://github.com/aws/aws-lc/pull/1846. But I don't see it performing a PCT for every keypair it generates like https://github.com/aws/aws-lc/pull/1969. I think we need both, just the PR description looks a little wrong.

crypto/ml_dsa/ml_dsa.c
jakemas: You are right! They aren't needed, I got rid of them. [6a296df](https://github.com/aws/aws-lc/pull/2148/commits/6a296df58e950ee71d01cea75f8815723ac9abe5)

jakemas:
Note to reviewers: I'm just checking my read on the spec here, because it says PCTs are required for keygen, but nothing about calling the PCT when using sign/verify. Thus, I may not need those sign/verify `no_self_test` functions, and can use the standard internal ones.

The strange thing is other signature schemes do perform self tests on sign/verify, I see it in ed25519 and RSA in the library... is it a difference in the standard, or my interpretation of?

crypto/ml_dsa/ml_dsa.h
jakemas: removed

WillChilds-Klein:
What's the performance impact of this change for key generation?

skmcgrail (dismissed):

crypto/ml_dsa/ml_dsa_ref/sign.c
```diff
@@ -7,6 +7,29 @@
...
+ * Returns 1 if the signature was successfully verified, 0 if it cannot.
+ *
+ * Note: FIPS 204 requires that public/private key pairs are to be used only for
+ * the calculation and/of verification of digital signatures.
+**************************************************/
+static int ml_dsa_keypair_pct(ml_dsa_params *params,
+                              uint8_t *pk,
+                              uint8_t *sk) {
+  uint8_t signature[MLDSA87_SIGNATURE_BYTES];
+  ml_dsa_sign(params, signature, &params->bytes, NULL, 0, NULL, 0, sk);
```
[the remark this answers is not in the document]
jakemas: Yes it does, the reason is ML-DSA actually signs `mu = (tr || 0 || ctx || ctx_len || M)` so even if `M` and `ctx` are `NULL` we still sign `mu = tr = SHAKE(pk)`. 


These get absorbed within `ml_dsa_sign_internal`, where any 0 length update is omitted, i.e., just `tr` is hashed into the message representative:
```
    SHAKE_Init(&state, SHAKE256_BLOCKSIZE);
    SHA3_Update(&state, tr, ML_DSA_TRBYTES);
    SHA3_Update(&state, pre, prelen);
    SHA3_Update(&state, m, mlen);
    SHAKE_Final(mu, &state, ML_DSA_CRHBYTES);
```

I wanted to make the PCT as simple as possible.

nebeid (approved):

nebeid (approved):

skmcgrail (approved):

PR #2155, opened 2025-02-01
SHAKE Incremental Byte Squeezes && EVP_ Tests

manastasova:
### Issues:
Resolves #CryptoAlg-2835 && 
Resolves #CryptoAlg-2836

### Description of changes: 
AWS-LC supports SHA3 and SHAKE algorithms. SHAKE, as defined in FIPS202, is an extendible output function, where the output data could be generated by multiple calls to "squeeze" function. Currently, there are two main features (internal and external APIs support) that are not supported by AWS-LC:
- (Internal APIs) Currently, AWS-LC supports incremental XOF output (internal only via SHAKE_Squeeze) generation when the requested data is a multiple of the |block_size|. I.e., ceil(requested_output_bytes/block_size) full blocks will be generated and directly outputted without internal buffering. Thus, it is the user's (other algorithm) responsibility to manage output data when the request is not a multiple of the |block_size|.
- Currently, AWS-LC does not support external (EVP_Digest) APIs for incremental squeezes.

This PR add both features for incremental squeezes in arbitrary length output requests (up to a byte): 
- Internal APIs (SHAKE_) changes:
   - Add support for incremental |SHAKE_Squeeze| with arbitrary size (up to a byte) output requests via internal buffer.
   - Add KECCAK600_CTX struct |state| field updates and checks yo ensure functions are called only in the corresponding bitstate, e.g., |KECCAK1600_STATE_SQUEEZE| allows further bitstate squeezes, |KECCAK1600_STATE_FINAL| does not allow further squeezes, etc.
- External APIs (EVP_Digest) changes:
   - Add external API support for incremental |SHAKE_Squeeze| with arbitrary size (up to a byte) output requests via |EVP_DigestSqueeze|.
   - Restrict the number of calls to |EVP_DigestFinalXOF| to only one. For incremental squeeze functionality |EVP_DigestSqueeze| should be used.
   - Restrict the use of |EVP_DigestFinal| to hash algorithms only. For XOF algorithms |EVP_DigestFinalXOF| and |EVP_DigestSqueeze| should be used.

This PR adds more tests for EVP_Digest XOF functionality (all test are running through the entire NIST Test Vector list):
- External APIs (EVP_Digest) additional tests:
   - Test Final - Assert fail when |EVP_DigestFinal| is called for XOF algorithms
   - Test Absorb - Assert success when |EVP_DigestUpdate| is called byte-by-byte
   - Test Squeeze - Assert success when |EVP_DigestSqueeze| is called byte-by-byte
   - Test Squeeze - Assert success when |EVP_DigestSqueeze| is called in set byte increments
   - Test Squeeze with random Input - Assert success when |EVP_DigestSqueeze| is called on a random message
   - Test Squeeze with random Input - Assert success when |EVP_DigestSqueeze| is called on a random message in set byte increments
   - Test Final XOF without Update - Assert fail when |EVP_DigestFinalXOF| is called as a streaming API

### Call-outs:
Service indicator is updated:
- Inside SHAKE_Squeeze (Streaming XOF Squeezes output generation **does not update** the service indicator after each extendable output update);

### Testing:


crypto/fipsmodule/sha/sha3.c
```diff
@@ -332,16 +332,51 @@ int SHAKE_Squeeze(uint8_t *md, KECCAK1600_CTX *ctx, size_t len) {
...
-  Keccak1600_Squeeze(ctx->A, md, len, ctx->block_size, ctx->state);
-  ctx->state = KECCAK1600_STATE_SQUEEZE;
+  // Process all full size output requested blocks
+  block_bytes = ctx->block_size * (len / ctx->block_size);
+  if (len > ctx->block_size) {
+    Keccak1600_Squeeze(ctx->A, md, block_bytes, ctx->block_size, ctx->state);
+    md += block_bytes;
+    len -= block_bytes;
+    ctx->state = KECCAK1600_STATE_SQUEEZE;
+  }
```
justsmth, line 367: `block_bytes` could be declared and initialized inside the `if` block. It's not used elsewhere in this function.

crypto/fipsmodule/sha/sha3.c
```diff
@@ -332,16 +332,51 @@ int SHAKE_Squeeze(uint8_t *md, KECCAK1600_CTX *ctx, size_t len) {
...
+    // call to SHAKE_Squeeze (incremental byte-wise SHAKE_Squeeze)
+    Keccak1600_Squeeze(ctx->A, ctx->buf, ctx->block_size, ctx->block_size, ctx->state);
+    OPENSSL_memcpy(md, ctx->buf, len);
+    ctx->buf_load = ctx->block_size - len; // how much there is still in buffer to be consumed    
+    md += len;
+    ctx->state = KECCAK1600_STATE_SQUEEZE;
+  }
 
   //FIPS_service_indicator_update_state();
   return 1;
```
justsmth, line 382: NP: No need to update local variable `md` immediately prior to the return.

crypto/fipsmodule/digest/internal.h
```diff
@@ -79,7 +79,11 @@ struct env_md_st {
   void (*init)(EVP_MD_CTX *ctx);
 
   // update hashes |len| bytes of |data| into the state in |ctx->md_data|.
-  void (*update)(EVP_MD_CTX *ctx, const void *data, size_t count);
+  // Digest update functions always return 1. update calls after |final| are 
+  // restricted via |ctx| check (|final| cleanses the |ctx|).
+  // Digest XOF update functions return 1 on success and 0 on failure, 
+  // returned from |SHAKE_Absorb|, to restrict update calls after |squeezeXOF|.
```
justsmth, line 85: NP: Reword to clarify. I found this sentence hard to parse.

crypto/fipsmodule/sha/sha3_test.cc
```diff
@@ -67,6 +107,195 @@ class SHA3TestVector {
...
+    ASSERT_TRUE(EVP_DigestInit_ex(ctx.get(), algorithm, NULL));
+    ASSERT_TRUE(EVP_DigestSqueeze(ctx.get(), digest.get(), digest_length/2));
+    ASSERT_FALSE(EVP_DigestFinalXOF(ctx.get(), digest.get() + digest_length/2,
+                                                      digest_length/2));
+
+    // Test Update after Squeeze
+    // Assert fail when |EVP_DigestUpdate| is called after |EVP_DigestSqueeze|
+    ASSERT_TRUE(EVP_DigestInit_ex(ctx.get(), algorithm, NULL));
+    ASSERT_TRUE(EVP_DigestSqueeze(ctx.get(), digest.get(), digest_length));
+    ASSERT_FALSE(EVP_DigestUpdate(ctx.get(), msg_.data(), msg_.size()));
```
justsmth, line 166: Does this sufficiently exhaust the ways that the API could be used incorrectly? Could a consumer try calling `EVP_DigestSqueeze` on an MD5?

crypto/fipsmodule/sha/sha3.c
manastasova: Thanks.

crypto/fipsmodule/sha/sha3.c
manastasova: Thank you

crypto/fipsmodule/sha/sha3_test.cc
manastasova: `EVP_DigestSqueeze` could be only called (successfully) with `EVP_MD_flags(ctx->digest) & EVP_MD_FLAG_XOF`. Other cases, such as `EVP_DigestSqueeze` after `EVP_DigestFinalXOF` and `EVP_Digest` will be restricted by `if (ctx->digest == NULL)`; `EVP_DigestSqueeze` can be called immediately after `EVP_DigestInit` or `EVP_DigestUpdate`.

include/openssl/digest.h
```diff
@@ -294,6 +294,12 @@ OPENSSL_EXPORT void EVP_MD_CTX_destroy(EVP_MD_CTX *ctx);
 OPENSSL_EXPORT int EVP_DigestFinalXOF(EVP_MD_CTX *ctx, uint8_t *out,
                                       size_t len);
 
+// EVP_DigestSqueeze provides byte-wise streaming XOF output generation for 
+// XOF digests, writing |len| bytes of extended output to |out|. It can be 
+// called multiple times with arbitrary length |len| output requests
+OPENSSL_EXPORT int EVP_DigestSqueeze(EVP_MD_CTX *ctx, uint8_t *out,
+                                      size_t len);
```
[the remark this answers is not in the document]
manastasova: Thanks.

crypto/fipsmodule/sha/sha3_test.cc
justsmth: I see in [the test coverage](https://app.codecov.io/gh/aws/aws-lc/pull/2155?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws) that there were a few lines missing coverage from `digest.c` around this check in both `EVP_DigestFinalXOF` and `EVP_DigestSqueeze`:
```
   if ((EVP_MD_flags(ctx->digest) & EVP_MD_FLAG_XOF) == 0) {
      OPENSSL_PUT_ERROR(DIGEST, ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
      return 0;
    }
```

I was thinking that extending coverage to these lines could be added fairly easily in this test?

crypto/fipsmodule/digest/internal.h
```diff
@@ -79,7 +79,11 @@ struct env_md_st {
   void (*init)(EVP_MD_CTX *ctx);
 
   // update hashes |len| bytes of |data| into the state in |ctx->md_data|.
-  void (*update)(EVP_MD_CTX *ctx, const void *data, size_t count);
+  // Digest update functions always return 1. update calls after |final| are 
+  // restricted via |ctx| check (|final| cleanses the |ctx|). Digest XOF update
+  // function propagates the return value from |SHAKE_Absorb|, that is 1 on
+  // success and 0 on failure, to restrict update calls after |squeezeXOF|.
+  int (*update)(EVP_MD_CTX *ctx, const void *data, size_t count);
```
justsmth, line 86: NP: I think mentioning `SHAKE_Absorb` in the description bothers me a little. Perhaps it can be slightly more generic:
```
// Returns 1 on success and 0 on failure. Failures can only occur on a 
// digest XOF update if called after |squeezeXOF|.
```

justsmth (approved):

crypto/fipsmodule/sha/sha3_test.cc
manastasova: True. I added extra tests in the |`NISTTestVectors`|, testing the XOF specific functions when a non XOF digest is expected.

crypto/fipsmodule/sha/sha3.c
```diff
@@ -174,8 +174,8 @@ static int FIPS202_Update(KECCAK1600_CTX *ctx, const void *data, size_t len) {
 
   if (rem != 0) {
     OPENSSL_memcpy(ctx->buf, data_ptr_copy + len - rem, rem);
-    ctx->buf_load = rem;
   }
+  ctx->buf_load = rem;
```
andrewhop, line 178: Is this a fix for a particular test?

crypto/fipsmodule/sha/sha3.c
```diff
@@ -312,18 +312,18 @@ int SHAKE_Final(uint8_t *md, KECCAK1600_CTX *ctx, size_t len) {
...
 
   FIPS_service_indicator_update_state();
-
   return 1;
 }
 
 // SHAKE_Squeeze can be called multiple time for incremental XOF output
 int SHAKE_Squeeze(uint8_t *md, KECCAK1600_CTX *ctx, size_t len) {
+  size_t block_bytes;
+  ctx->md_size = len;
```
andrewhop, line 321: Why did you move this? This is before the `ctx ==NULL` check now, can you add some more tests with invalid input to make sure we catch all of these issues.

crypto/fipsmodule/sha/sha3_test.cc
```diff
@@ -67,6 +115,195 @@ class SHA3TestVector {
...
+              Bytes(digest_.data(), digest_length));
+
+    // Test Final
+    // Assert fail when |EVP_DigestFinal| is called for XOF algorithms
+    OPENSSL_memset(digest.get(), 0, digest_length);
+    ASSERT_TRUE(EVP_DigestInit(ctx.get(), algorithm));
+    ASSERT_TRUE(EVP_DigestUpdate(ctx.get(), msg_.data(), msg_.size()));
+
+    ASSERT_FALSE(EVP_DigestFinal(ctx.get(), digest.get(), &digest_length));
+    ASSERT_FALSE(EVP_DigestFinalXOF(ctx.get(), digest.get(), digest_length));
```
andrewhop, line 152: Interesting, even though the `EVP_DigestFinal` call failed it has corrupted the ctx somehow such that `EVP_DigestFinalXOF` also fails now?

crypto/fipsmodule/sha/sha3_test.cc
```diff
@@ -67,6 +115,195 @@ class SHA3TestVector {
...
+      ASSERT_TRUE(EVP_DigestSqueeze(ctx.get(), digest.get() + i, 1));
+    }
+
+    EXPECT_EQ(Bytes(digest.get(), digest_length),
+              Bytes(digest_.data(), digest_length));
+
+    // Test Squeeze
+    // Assert success when |EVP_DigestSqueeze| is called in set byte increments
+    for (cur_test = 0, sqd_bytes = 0; cur_test < (int) (sizeof(stride_tests)/sizeof(stride_tests[0])); cur_test++, sqd_bytes = 0) {
+    to_sq_bytes = stride_tests[cur_test].startsz;
```
andrewhop, line 205: Indentation is off here.

crypto/fipsmodule/sha/sha3_test.cc
```diff
@@ -67,6 +115,195 @@ class SHA3TestVector {
...
+
+    for (size_t i = 0; i < digest_length; i++) {
+      ASSERT_TRUE(EVP_DigestSqueeze(ctx.get(), digest.get() + i, 1));
+    }
+
+    EXPECT_EQ(Bytes(digest.get(), digest_length),
+              Bytes(digest_.data(), digest_length));
+
+    // Test Squeeze
+    // Assert success when |EVP_DigestSqueeze| is called in set byte increments
```
andrewhop, line 203: Instead of hard-coding these tests in stride_tests can we just test all absorb and squeeze sizes from 1 to msg/digest length? That would also cover the above byte by byte squeeze/absord tests.

crypto/fipsmodule/sha/sha3_test.cc
```diff
@@ -67,6 +115,195 @@ class SHA3TestVector {
...
+    std::ifstream urandom("/dev/urandom", std::ios::binary);
+      if (!urandom) {
+          return;
+    }
+
+    std::vector<unsigned char> random_bytes(num_bytes);
+    urandom.read(reinterpret_cast<char*>(random_bytes.data()), num_bytes);
+    if (!urandom) {
+        return;
+    }
```
andrewhop, line 239: Can this just use RAND_bytes?

crypto/fipsmodule/sha/sha3_test.cc
```diff
@@ -67,6 +115,195 @@ class SHA3TestVector {
...
+          ASSERT_TRUE(EVP_DigestSqueeze(ctx.get(), digest.get() + sqd_bytes, to_sq_bytes));
+          sqd_bytes += to_sq_bytes;
+          to_sq_bytes = stride_tests[cur_test].incsz;
+      }
+     EXPECT_EQ(Bytes(digest.get(), digest_length),
+          Bytes(digest_.data(), digest_length));
+    }
+
+    // Test Squeeze with random Input
+    // Assert success when |EVP_DigestSqueeze| is called on a random message
```
andrewhop, line 223: Is there something in particular in SHAKE that is potentially dependent on the input message this is testing for?

include/openssl/digest.h
```diff
@@ -339,7 +346,7 @@ struct env_md_ctx_st {
   // |digest->update|. |digest->update| operates against |md_data| above, but
   // |HMAC_CTX| maintains its own data state in |HMAC_CTX->md_ctx|.
   // |HMAC_Update| also has an additional state transition to handle.
-  void (*update)(EVP_MD_CTX *ctx, const void *data, size_t count);
+  int (*update)(EVP_MD_CTX *ctx, const void *data, size_t count);
```
andrewhop, line 349: Oh `env_md_ctx_st` is in our public header, do we support users defining there own implementations and this might break them?

include/openssl/digest.h
```diff
@@ -339,7 +346,7 @@ struct env_md_ctx_st {
...
 
   // pctx is an opaque (at this layer) pointer to additional context that
   // EVP_PKEY functions may store in this object.
```
andrewhop, line 352: The comment for flags below is out of date

crypto/fipsmodule/sha/sha3_test.cc
manastasova: Yes. Perfect!

include/openssl/digest.h
manastasova: I don't think the user can implement the internal update function. The update is usually assigned to the EVP_MD md->update, in the case of digest functions https://github.com/aws/aws-lc/blob/82a16b62fa5370686ecd543c787cac442039afec/crypto/fipsmodule/digest/digest.c#L234 and https://github.com/aws/aws-lc/blob/82a16b62fa5370686ecd543c787cac442039afec/crypto/fipsmodule/digest/digest.c#L276
In the case of HMAC it is assigned to the hmac_update; - https://github.com/aws/aws-lc/blob/82a16b62fa5370686ecd543c787cac442039afec/crypto/fipsmodule/evp/digestsign.c#L142

include/openssl/digest.h
manastasova: I don't think the flags in `env_md_ctx_st` are updated; the flags in `env_md_st` support `EVP_MD_FLAG_XOF` flag.

justsmth (dismissed):

crypto/fipsmodule/sha/sha3_test.cc
andrewhop: Yeah I was thinking testing squeeze 1 byte at a time, 2 bytes at a time, 3, 4, 5, ... up to message length.

include/openssl/digest.h
andrewhop: Ohhhhhh

crypto/fipsmodule/sha/sha3_test.cc
manastasova: I can add a new test doing that (I am not sure if it adds value since there is already a byte-by-byte test). However, using set increments (such as _almost_ block_size, block_size/2, etc.) tests edge-cases and I think we should keep it.

andrewhop (approved):

justsmth (approved):

PR #2162, opened 2025-02-04
Add support to define a callback for FIPS test failures instead of aborting the process

andrewhop:
### Description of changes: 
This change adds a new build option `AWSLC_FIPS_FAILURE_CALLBACK` that then requires applications that use AWS-LC to define the `AWS_LC_fips_failure_callback(const char* message)` function. If that function is not defined AWS-LC fails to load. 

When the callback is defined and a FIPS failure (CAST or PWCT) happens AWS-LC calls that function instead. It is up to the calling application to ensure no further cryptographic operations occur. This is not an approved way to use AWS-LC in FIPS mode, for the approved way to build and use AWS-LC see the latest FIPS security policy. 

### Testing
This change adds a new gtest application that checks the callback function is called with the expected message and is used with break-kat.go and the runtime PCT environment variables to ensure the callback is passed the expected message.


crypto/fipsmodule/bcm.c
```diff
@@ -386,7 +376,13 @@ int BORINGSSL_integrity_test(void) {
 }
 #endif  // OPENSSL_ASAN
 
-void BORINGSSL_FIPS_abort(void) {
+WEAK_SYMBOL_FUNC(void, AWS_LC_fips_failure_callback, (const char* message))
+#include <unistd.h>
+void AWS_LC_FIPS_failure(const char* message) {
```
darylmartin100, line 381: Please correct me if I'm wrong, but this will break our next FIPS validation.  If we allow someone outside the boundary to register a callback our lab will not be able to prove that we stop processing events when a PCT or self-tests fails. When we detect a failure we need to abort immediately as defined in ISO 19790 section 7.3.3, b), 7.3.3 c), and 7.5. This needs to be wrapped in a #ifdef so we don't expose it when we are building as a 140-3 level 1 module.

andrewhop: This is no different than the current behavior: an application using AWS-LC can install a custom signal handler to perform any logging/cleanup before the abort terminates the process. However, with this callback the application gets a clearer signal if the abort is a FIPS anomaly or a different SIGABRT.

tests/ci/run_fips_callback_tests.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash -ex
```
skmcgrail, line 1:
```suggestion
#!/usr/bin/env bash
set -ex
```

Don't hard-code the shell paths.

tests/ci/run_fips_callback_tests.sh
```diff
@@ -0,0 +1,57 @@
...
+done
+
+# Use the original test executable that doesn't have a broken self test
+# Tell our test what test is expected to fail
+export FIPS_CALLBACK_TEST_EXPECTED_FAILURE="RSA_PWCT"
+# Tell bcm which test to break
+export BORINGSSL_FIPS_BREAK_TEST="RSA_PWCT"
+output=$(2>&1 $original_test --gtest_filter=FIPSCallback.RSARuntimeTest || true)
+if  [[ $output =~ "AWS-LC FIPS callback passed" && ! $output =~ "PASSED" && ! $output =~ "FAILED" ]]; then
+    echo "✅ Found callback success"
```
skmcgrail, line 41: Do these render correctly in code build or buildfarm?

crypto/fipsmodule/bcm.c
andrewhop: Updated to a callback that doesn't abort if set and put it all behind a compilation option.

tests/ci/run_fips_callback_tests.sh
andrewhop: They rendered nicely in codebuild, but now that the test doesn't abort I can rely on the default gtest result printing.

crypto/fipsmodule/bcm.c
```diff
@@ -251,6 +251,10 @@ static void BORINGSSL_maybe_set_module_text_permissions(int _permission) {}
 
 #endif  // !ASAN
 
+#if defined(AWSLC_FIPS_FAILURE_CALLBACK)
```
torben-hansen, line 254: Does the reallyyyy need a compile-time flag?

crypto/fipsmodule/bcm.c
```diff
@@ -398,7 +420,7 @@ void AWS_LC_FIPS_failure(const char* message) {
     exit(1);
   }
 }
-
+#endif
 #else  // BORINGSSL_FIPS
 void AWS_LC_FIPS_failure(const char* message) {
```
torben-hansen, line 425: Could be combined with the second function definition.

crypto/fips_callback_test.cc
```diff
@@ -0,0 +1,162 @@
...
+        FAIL() << "Failed to find expected message for FIPS_CALLBACK_TEST_POWER_ON_TEST_FAILURE=" << broken_kat;
+      }
+  } else {
+    FAIL() << "AWS_LC_fips_failure_callback called when no KAT was expected to be broken";
+  }
+  callback_call_count++;
+
+}
+
+  TEST(FIPSCallback, PowerOnSelfTests) {
```
torben-hansen, line 81: Random indentation...

crypto/fipsmodule/bcm.c
andrewhop: With this iteration of this design I think so. My fear is an application using AWS-LC for FIPS might not notice if something else in their dependencies defined the callback and then let the application continue using AWS-LC in a degraded state silently. And now this version doesn't abort so it's possible to continue using AWS-LC in a degraded state.

andrewhop:
I realized I could simplify the testing and put it all in crypto_test. I had to update some of the self_check values to not conflict, e.g. the self_check was using an all zero key and we had another test that had an all zero key so break-kat.go didn't know which one was the real one.

crypto/internal.h
```diff
@@ -1272,12 +1272,15 @@ static inline uint64_t CRYPTO_subc_u64(uint64_t x, uint64_t y, uint64_t borrow,
 // AWS_LC_FIPS_failure is called when a FIPS power-on or continuous test
 // fails. If the library is built in FIPS mode it prevents any further
 // cryptographic operations by the current process.
+#if defined(AWSLC_FIPS_FAILURE_CALLBACK)
+void AWS_LC_FIPS_failure(const char* message);
```
torben-hansen, line 1276: Need to have external linkage?

crypto/fipsmodule/self_check/self_check.c
```diff
@@ -2670,11 +2671,9 @@ static int boringssl_self_test_fast(void) {
     goto err;
   }
 
-  // PBKDF2 KAT - password/salt data from RFC 6070, derived key generated by
-  // Python's cryptography module
   static const uint8_t kPBKDF2Password[] = {
-    'p', 'a', 's', 's', 'w', 'o', 'r', 'd', 'P', 'A', 'S', 'S', 'W', 'O', 'R',
-    'D', 'p', 'a', 's', 's', 'w', 'o', 'r', 'd'
+    'A', 'W', 'S', '-', 'L', 'C', 'F', 'I', 'P', 'S', 'p', 'a', 's', 's', 'w',
+    'o', 'r', 'd'
```
WillChilds-Klein, line 2676: why do we prefer this over the RFC's values?

crypto/internal.h
```diff
@@ -1420,6 +1423,17 @@ OPENSSL_EXPORT int OPENSSL_vasprintf_internal(char **str, const char *format,
 #define GUARD_PTR(ptr) __AWS_LC_ENSURE((ptr) != NULL, OPENSSL_PUT_ERROR(CRYPTO, ERR_R_PASSED_NULL_PARAMETER); \
                                        return AWS_LC_ERROR)
 
+
+// Windows doesn't really support weak symbols as of May 2019, and Clang on
+// Windows will emit strong symbols instead. See
+// https://bugs.llvm.org/show_bug.cgi?id=37598
```
WillChilds-Klein, line 1429: should we disable this new functionality on windows entirely?

andrewhop: Yeah, that also fixes all the windows compilation issues.

torben-hansen (approved):

skmcgrail (approved):

WillChilds-Klein (approved):

PR #2163, opened 2025-02-04
Add SSL_CTX_use_cert_and_key  

smittals2:
### Issues:
`CryptoAlg-2863`

### Description of changes: 
Adds the function SSL_CTX_use_cert_and_key which sets the leaf, chain, and private key all in one operation. This function is also needed by librdkafka. 

### Call-outs:

### Testing:
Tests using this API to set the relevant fields and ensure TLS connection succeeds. 


ssl/ssl_cert.cc
```diff
@@ -302,7 +304,7 @@ static int cert_set_chain_and_key(
   }
 
   for (size_t i = 0; i < num_certs; i++) {
-    if (!PushToStack(certs_sk.get(), UpRef(certs[i]))) {
+    if (!PushToStack(certs_sk.get(), UpRef(sk_CRYPTO_BUFFER_value(certs->get(), i)))) {
       return 0;
     }
   }
```
justsmth, line 310: We could use `sk_CRYPTO_BUFFER_deep_copy` here:
```
  UniquePtr<STACK_OF(CRYPTO_BUFFER)> certs_sk(sk_CRYPTO_BUFFER_deep_copy(
      certs->get(), buffer_up_ref, CRYPTO_BUFFER_free));
```

The copy function `buffer_up_ref` is defined above.  It's also needed [here](https://github.com/aws/aws-lc/blob/c9d48a636986c82cd122a0d6cd586191aa9afb2c/ssl/ssl_session.cc#L217-L222). Perhaps the copy function should be defined somewhere they could share?

ssl/ssl_cert.cc
```diff
@@ -311,12 +313,19 @@ static int cert_set_chain_and_key(
...
-  // Update certificate slot index once all checks have passed.
   // Certificate slot validity already checked in |check_leaf_cert_and_privkey|.
   int idx = ssl_get_certificate_slot_index(privkey);
-  cert->cert_private_keys[idx].privatekey = UpRef(privkey);
+  CERT_PKEY *cert_pkey = &cert->cert_private_keys[idx];
+  if (!override && (cert_pkey->privatekey != NULL ||
+      cert_pkey->x509_leaf != NULL ||
+      cert_pkey->chain != NULL)) {
+    return 0;
+  }
```
justsmth, line 323: * Couldn't this check be at the start of the function (line ~280) and avoid the potentially unnecessary work/allocations above?
* Even better the check could be at the top of `SSL_CTX_use_cert_and_key` (the only caller that doesn't override).
```
  int idx = ssl_get_certificate_slot_index(privatekey);
  if (idx < 0) {
    OPENSSL_PUT_ERROR(SSL, SSL_R_UNKNOWN_CERTIFICATE_TYPE);
    return 0;
  }
  CERT_PKEY *cert_pkey = &ctx->cert->cert_private_keys[idx];
  if (!override && (cert_pkey->privatekey != NULL ||
      cert_pkey->x509_leaf != NULL ||
      cert_pkey->chain != NULL)) {
    return 0;
  }
```

smittals2: Great point, I moved it to the top of SSL_CTX_use_cert_and_key. This way we don't have to change the internal func's signature either.

ssl/ssl_cert.cc
```diff
@@ -952,23 +950,115 @@ BSSL_NAMESPACE_END
...
+  if (idx < 0) {
+    OPENSSL_PUT_ERROR(SSL, SSL_R_UNKNOWN_CERTIFICATE_TYPE);
+    return 0;
+  }
+  CERT_PKEY *cert_pkey = &cert->cert_private_keys[idx];
+  if (!override && (cert_pkey->privatekey != NULL ||
+      cert_pkey->x509_leaf != NULL ||
+      cert_pkey->chain != NULL)) {
+    return 0;
+  }
```
justsmth, line 1010: Shouldn't we also set a `SSL_R_NOT_REPLACING_CERTIFICATE` flag as OpenSSL does: https://github.com/openssl/openssl/blob/5cba3629098a55a559a9bb177c095fc77b1e8b88/ssl/ssl_rsa.c#L1044

smittals2: Ah good catch, I'll create a new error code for this

include/openssl/ssl.h
```diff
@@ -1239,6 +1239,22 @@ OPENSSL_EXPORT int SSL_set_signing_algorithm_prefs(SSL *ssl,
                                                    const uint16_t *prefs,
                                                    size_t num_prefs);
 
+// SSL_CTX_use_cert_and_key sets |x509|, |privatekey|, and |chain| on |ctx|.
+// The pkey argument must be the private key of the certificate |x509|.
```
justsmth, line 1243: NP: You refer to the "privatekey" parameter as "pkey".

include/openssl/ssl.h
```diff
@@ -1239,6 +1239,22 @@ OPENSSL_EXPORT int SSL_set_signing_algorithm_prefs(SSL *ssl,
...
+// If the override argument is 0, then |x509|, |privatekey|, and |chain| are
+// set only if all were not previously set. If override is non-0, then the
+// certificate, private key and chain certs are always set. |privatekey| and
+// |x509| are not copied or duplicated, their reference count is incremented.
```
justsmth, line 1247: NP: Plural - "... their reference counts are incremented."

include/openssl/ssl.h
```diff
@@ -1239,6 +1239,22 @@ OPENSSL_EXPORT int SSL_set_signing_algorithm_prefs(SSL *ssl,
...
+// In OpenSSL, a shallow copy of |chain| is stored with a reference count
+// increment for all X509 objects in the chain. In AWS-LC,
+// we represent X509 chains as a CRYPTO_BUFFER stack. Therefore, we create a
+// an internal copy and leave the |chain| parameter untouched. This means,
+// changes to |chain| after this function is called will not update in |ctx|.
```
justsmth, line 1252: NP: Does this differ from [OpenSSL behavior](https://docs.openssl.org/master/man3/SSL_CTX_use_certificate/)? (Based on the [upref](https://github.com/openssl/openssl/blob/5cba3629098a55a559a9bb177c095fc77b1e8b88/ssl/ssl_rsa.c#L1048-L1054), I think it does?) If so, we should explicitly say that it differs from OpenSSL.

justsmth (dismissed):

include/openssl/ssl.h
smittals2: Yea OpenSSL stores a reference directly as a X509 object, but we convert to CryptoBuffer to store internally. Will make the distinction more explicit in the comments.

justsmth (approved):

justsmth (dismissed):

ssl/ssl_cert.cc
```diff
@@ -274,7 +274,7 @@ static enum leaf_cert_and_privkey_result_t check_leaf_cert_and_privkey(
 }
 
 static int cert_set_chain_and_key(
-    CERT *cert, CRYPTO_BUFFER *const *certs, size_t num_certs,
+    CERT *cert, UniquePtr<STACK_OF(CRYPTO_BUFFER)> *certs, size_t num_certs,
```
skmcgrail, line 277:
```suggestion
    CERT *cert, UniquePtr<STACK_OF(CRYPTO_BUFFER)> &certs, size_t num_certs,
```

Since it's C++ pass by reference rather then passing the address. You'll have to fix up the other spots where this is called then.

skmcgrail: Also assumes this works with the usage of `std::move`?

smittals2: Made the change, based on some google searching std::move should work on references to UniqPtrs.

skmcgrail (approved):

justsmth (approved):

PR #2165, opened 2025-02-04
Wrap pointers to s2n-bignum functions

nebeid:
### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 
Wrap pointers to s2n-bignum functions so that the delocator does not think they're local_target while they have :got: references.

### Testing:
On Graviton3, FIPS static build with gcc-11 which used to fail with
```
Building ASM object crypto/fipsmodule/CMakeFiles/bcm_hashunset.dir/bcm-delocated.S.o
FAILED: crypto/fipsmodule/CMakeFiles/bcm_hashunset.dir/bcm-delocated.S.o 
/usr/bin/gcc-11 -DBORINGSSL_FIPS -DBORINGSSL_HAVE_LIBUNWIND -DBORINGSSL_IMPLEMENTATION -DFIPS_ENTROPY_SOURCE_PASSIVE -Isymbol_prefix_include -I../include -Wno-newline-eof -Wa,--noexecstack -O3 -DNDEBUG -fPIC -MD -MT crypto/fipsmodule/CMakeFiles/bcm_hashunset.dir/bcm-delocated.S.o -MF crypto/fipsmodule/CMakeFiles/bcm_hashunset.dir/bcm-delocated.S.o.d -o crypto/fipsmodule/CMakeFiles/bcm_hashunset.dir/bcm-delocated.S.o -c crypto/fipsmodule/bcm-delocated.S
bcm.c: Assembler messages:
bcm.c:7811: Error: pc-relative address offset out of range
bcm.c:7813: Error: pc-relative address offset out of range
bcm.c:7815: Error: pc-relative address offset out of range
```

### DIFF
--- a/crypto/fipsmodule/ec/p521.c
+++ b/crypto/fipsmodule/ec/p521.c
@@ -126,6 +126,27 @@ static const p521_limb_t p521_felem_p[P521_NLIMBS] = {
 
 #endif // EC_NISTP_USE_S2N_BIGNUM
 
+// The wrapper functions are needed for FIPS static build.
+// Otherwise, initializing ec_nistp_meth with pointers to s2n-bignum
+// functions directly generates :got: references that are also thought
+// to be local_target by the delocator.
+static inline void p521_felem_add_wrapper(ec_nistp_felem_limb *c,
+                                          const ec_nistp_felem_limb *a,
+                                          const ec_nistp_felem_limb *b) {
+  p521_felem_add(c, a, b);
+}
+
+static inline void p521_felem_sub_wrapper(ec_nistp_felem_limb *c,
+                                          const ec_nistp_felem_limb *a,
+                                          const ec_nistp_felem_limb *b) {
+  p521_felem_sub(c, a, b);
+}
+
+static inline void p521_felem_neg_wrapper(ec_nistp_felem_limb *c,
+                                          const ec_nistp_felem_limb *a) {
+  p521_felem_opp(c, a);
+}
+
 static p521_limb_t p521_felem_nz(const p521_limb_t in1[P521_NLIMBS]) {
   p521_limb_t is_not_zero = 0;
   for (int i = 0; i < P521_NLIMBS; i++) {
@@ -289,11 +310,11 @@ static void p521_point_add(p521_felem x3, p521_felem y3, p521_felem z3,
 DEFINE_METHOD_FUNCTION(ec_nistp_meth, p521_methods) {
     out->felem_num_limbs = P521_NLIMBS;
     out->felem_num_bits = 521;
-    out->felem_add = bignum_add_p521;
-    out->felem_sub = bignum_sub_p521;
+    out->felem_add = p521_felem_add_wrapper;
+    out->felem_sub = p521_felem_sub_wrapper;
     out->felem_mul = bignum_mul_p521_selector;
     out->felem_sqr = bignum_sqr_p521_selector;
-    out->felem_neg = bignum_neg_p521;
+    out->felem_neg = p521_felem_neg_wrapper;
     out->felem_nz  = p521_felem_nz;
     out->felem_one = p521_felem_one;
     out->point_dbl = p521_point_double;


jakemas:
Confirmed CI is now passing with this change applied on https://github.com/aws/aws-lc/pull/2148

jakemas (approved):

jakemas:
Created https://github.com/aws/aws-lc/pull/2166 to test this fix.

jakemas:
Confirmed that FIPS static builds are all passing on https://github.com/aws/aws-lc/pull/2166

jakemas:
Waiting on this to be merged to unblock: https://github.com/aws/aws-lc/pull/2148

crypto/fipsmodule/ec/p384.c
```diff
@@ -83,6 +83,26 @@ static p384_limb_t p384_felem_nz(const p384_limb_t in1[P384_NLIMBS]) {
 
 #endif // EC_NISTP_USE_S2N_BIGNUM
 
+// The wrapper functions are needed for FIPS static build.
+// Otherwise, initializing ec_nistp_meth with pointers to s2n-bignum
+// functions directly generates :got: references that are also thought
+// to be local_target by the delocator.
```
andrewhop, line 89: What is the difference between defining this wrapper function and how s2n-bignum defines bignum_add_p384? Is my mental model of C functions are equivalent to functions defined in the assembly files wrong?

skmcgrail (approved):

justsmth:
This is probably expected, but I did try building this branch for x86-64 using gcc-14 and it's still failing as before:
```
error while processing "\t.section\t.data.rel.ro.local,\"aw\"\n" on line 226756: ".data section found in module"
```

crypto/fipsmodule/ec/p384.c
nebeid: It's a valid question. This change stemmed from the observation that those 3 functions that were wrapped, when they were being assigned to the `p521_methods` struct in [crypto/fipsmodule/ec/p521.c#L292-L296](https://github.com/aws/aws-lc/blob/51ae4b1b9eb6e44e7d783e2841d81cdfe490f3a1/crypto/fipsmodule/ec/p521.c#L292-L296), they were creating the following (commented) assembly instructions that were replaced by delocate.go by the ones right underneath. 
```
.Lp521_methods_init_local_target:
p521_methods_init:
...
// WAS adrp x10, :got:bignum_add_p521
adr x10, .Lbignum_add_p521_local_target
// WAS adrp x9, :got:bignum_sub_p521
adr x9, .Lbignum_sub_p521_local_target
// WAS adrp x6, :got:bignum_neg_p521
adr x6, .Lbignum_neg_p521_local_target
```
In the case of gcc compilations (not all versions, it depends which ones get lucky), the target address of the `adr` instructions was further than the admissible 1MB.
The observation that other s2n-bignum functions assigned to the same method such as `bignum_sqr_p521_alt` was accessed via `bignum_sqr_p521_selector` and wasn't causing the same problem, rather
```
// WAS adrp x8, bignum_mul_p521_selector
adr x8, .Lbignum_mul_p521_selector_local_target
```
and not too far from it
```
.align 2
.p2align 4,,11
.type bignum_sqr_p521_selector, %function
.Lbignum_sqr_p521_selector_local_target:
bignum_sqr_p521_selector:
.LFB423:

.cfi_startproc
// WAS adrp x2, :got:OPENSSL_armcap_P
    sub sp, sp, 128
    stp x0, x30, [sp, #-16]!
    bl .LOPENSSL_armcap_P_addr
    mov x2, x0
    ldp x0, x30, [sp], #16
    add sp, sp, 128
// WAS ldr x2, [x2, #:got_lo12:OPENSSL_armcap_P]
    ldr w2, [x2]
    tst w2, 28672
    beq .L453
// WAS b bignum_sqr_p521_alt
    b .Lbignum_sqr_p521_alt_local_target
```
 the branch instructions on the last line doesn’t suffer from the same range limit as the adr instruction.
 
 That's why I thought that wrapping the s2n-bignum function which may be far with a "nearby" static function may result in the same pattern as with `bignum_sqr_p521_selector`

torben-hansen (approved):

PR #2173, opened 2025-02-06
Add new CAST tests to break-kat.go

andrewhop:
### Description of changes: 
This change updates all the KATs in self_check.c to work with break-kat.go. 
1. Mark all the KAT data as `static` which moves the value into the read only data section
2. Update the Ed* KATs to use different messages for the signing and verification test so the test can distinguish between the two breaking
3. Update break-kat.go to break both copies of the ML-KEM public key
4. Add code to test_fips to trigger the lazy self tests
5. Update test-break-kat.sh to save the output to a variable so it shows up in the logs
6. Update the test names to be consistent 

### Testing:
The existing run_fips_test.sh runs test-break-kat.sh which iterates over all the supported KAT values and ensure the error output references the KAT


util/fipstools/test_fips.c
```diff
@@ -394,6 +396,46 @@ int main(int argc, char **argv) {
...
+  /* Ed25519 */
+  printf("About to Ed25519 sign ");
+  hexdump(kPlaintextSHA256, sizeof(kPlaintextSHA256));
+  uint8_t ed_public_key[ED25519_PUBLIC_KEY_LEN];
+  uint8_t ed_private_key[ED25519_PRIVATE_KEY_LEN];
+  ED25519_keypair(ed_public_key, ed_private_key);
+  uint8_t ed_signature[ED25519_SIGNATURE_LEN];
+  if (!ED25519_sign(ed_signature,kPlaintextSHA256, sizeof(kPlaintextSHA256), ed_private_key) ||
+    !ED25519_verify(kPlaintextSHA256, sizeof(kPlaintextSHA256), ed_signature, ed_public_key)) {
+    printf("ED25519 Sign/Verify PWCT failed.\n");
```
nebeid, line 408: Is it only that the PWCT failed? Wouldn't the module have aborted if it was the case?

andrewhop: Good point, test_fips.c only needs to use the functions to trigger the the lazy self tests. I was following the ECDSA pattern up above L385-394 but yes in theory this failure should not happen.

util/fipstools/break-kat.go
```diff
@@ -67,17 +80,26 @@ func main() {
...
-	i := bytes.Index(binaryContents, testInputValue)
-	if i < 0 {
-		fmt.Fprintln(os.Stderr, "Expected test input value was not found in binary.")
-		os.Exit(3)
+	// ML-KEM decap keys include a copy of the public encaps key
+	// FIPS 203 section 6.1
+	// dk = (dkPKE‖ek‖H(ek)‖𝑧)
+	// where dkPKE is the secret key, ek is the public key, and z is the hash of the public key
+	expected_copies := 1
+	if test == "ML-KEM-keyGen-encaps" {
```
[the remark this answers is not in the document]
andrewhop: I'll update the comment, breaking the keygen-encaps test is setting the expected kengen-ecaps key to all zero in the test. The issue is the decaps key includes another copy of the encaps key inside it and this script was rightfully complaining the expected encaps key was still found in the binary.

Another idea I just had was we could use two different seeds to generate two different ML-KEM keys and store the encaps key of seed1, and the decaps key of seed2.

nebeid (dismissed):

util/fipstools/test_fips.c
```diff
@@ -394,6 +396,46 @@ int main(int argc, char **argv) {
...
+  };
+  if (!ED25519ph_sign(ed_signature, kPlaintextSHA256, sizeof(kPlaintextSHA256), ed_private_key, ed25519_ph_context, sizeof(ed25519_ph_context)) ||
+    !ED25519ph_verify(kPlaintextSHA256, sizeof(kPlaintextSHA256), ed_signature, ed_public_key, ed25519_ph_context, sizeof(ed25519_ph_context))) {
+    printf("ED25519ph Sign/Verify PWCT failed.\n");
+    goto err;
+  }
+  printf("got signature ");
+  hexdump(ed_signature, sizeof(ed_signature));
+
+  /* ML-KEM */
```
torben-hansen, line 430: check return values?

andrewhop: Updated.

torben-hansen (approved):

nebeid (approved):

PR #2179, opened 2025-02-07
Make install_shared_and_static test more robust

smittals2:
### Issues:
`P198393553`

### Description of changes: 
Making the test more robust to prevent flaky behavior: 
1. Add more logging
2. Use sync to force file state after delete operations
3. use --fresh with CMAKE to prevent using any cache 
4. Turn off parallel CMAKE build

### Call-outs:

### Testing:

### DIFF
--- a/tests/ci/run_install_shared_and_static.sh
+++ b/tests/ci/run_install_shared_and_static.sh
@@ -44,7 +44,7 @@ function install_aws_lc() {
     rm -rf "${BUILD_DIR:?}"
     sync
 
-    ${CMAKE_COMMAND} -H${AWS_LC_DIR} -B${BUILD_DIR} -GNinja -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR} -DBUILD_TESTING=OFF -D${BUILD_SHARED_LIBS} --fresh
+    ${CMAKE_COMMAND} --fresh -H${AWS_LC_DIR} -B${BUILD_DIR} -GNinja -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR} -DBUILD_TESTING=OFF -D${BUILD_SHARED_LIBS}
     ${CMAKE_COMMAND} --build ${BUILD_DIR} --target install
 }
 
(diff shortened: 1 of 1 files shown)


tests/ci/run_install_shared_and_static.sh
```diff
@@ -6,7 +6,7 @@ set -exo pipefail
 
 source tests/ci/common_posix_setup.sh
 
-export CMAKE_BUILD_PARALLEL_LEVEL=${NUM_CPU_THREADS}
+export CMAKE_BUILD_PARALLEL_LEVEL=1
```
samuel40791765, line 9: Great find, were most of the random CI failures due to race conditions with the CMake building or the invalid file states? Seems like the latter's more of the cause of our issues, but I could be wrong.

samuel40791765 (approved):

tests/ci/run_install_shared_and_static.sh
```diff
@@ -92,10 +96,16 @@ build_myapp() {
     local AWS_LC_INSTALL_DIR=$2 # which install dir should be used
     local EXPECT_USE_LIB_TYPE=$3 # (".so" or ".a") which types of libssl and libcrypto are expected to be used
 
+    echo "Build Parameters:"
```
andrewhop, line 99: We run our ci with [set -x](https://man7.org/linux/man-pages/man1/set.1p.html) and this is already output:
```
+ build_myapp BUILD_SHARED_LIBS=ON install-both .so
+ local BUILD_SHARED_LIBS=BUILD_SHARED_LIBS=ON
+ local AWS_LC_INSTALL_DIR=install-both
+ local EXPECT_USE_LIB_TYPE=.so
```

tests/ci/run_install_shared_and_static.sh
```diff
@@ -92,10 +96,16 @@ build_myapp() {
...
+    echo "EXPECT_USE_LIB_TYPE: ${EXPECT_USE_LIB_TYPE}"
+
     local BUILD_DIR=${SCRATCH_DIR}/build
     rm -rf "${BUILD_DIR:?}"
+    sync
 
-    cmake -H${MYAPP_SRC_DIR} -B${BUILD_DIR} -GNinja -D${BUILD_SHARED_LIBS} -DCMAKE_PREFIX_PATH=${SCRATCH_DIR}/${AWS_LC_INSTALL_DIR}
+    cmake --fresh -H${MYAPP_SRC_DIR} -B${BUILD_DIR} -GNinja -D${BUILD_SHARED_LIBS} -DCMAKE_PREFIX_PATH=${SCRATCH_DIR}/${AWS_LC_INSTALL_DIR}
     cmake --build ${BUILD_DIR}
     ldd ${BUILD_DIR}/myapp
```
andrewhop, line 110: Looking at the the log at this point we know myapp has the right libraries linked. Can you update line 122 to capture the output of ldd running the second time, then grep that output for LIB_NAME. Then we can see the output of the ldd and see if that is somehow changing between line 110 and 122.

tests/ci/run_install_shared_and_static.sh
```diff
@@ -41,7 +42,9 @@ function install_aws_lc() {
 
     local BUILD_DIR=${SCRATCH_DIR}/build
     rm -rf "${BUILD_DIR:?}"
-    ${CMAKE_COMMAND} -H${AWS_LC_DIR} -B${BUILD_DIR} -GNinja -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR} -DBUILD_TESTING=OFF -D${BUILD_SHARED_LIBS}
+    sync
+
+    ${CMAKE_COMMAND} --fresh -H${AWS_LC_DIR} -B${BUILD_DIR} -GNinja -DCMAKE_INSTALL_PREFIX=${INSTALL_DIR} -DBUILD_TESTING=OFF -D${BUILD_SHARED_LIBS}
```
andrewhop, line 47: Does `--fresh` behave any differently than removing the entire build folder on line 44?

smittals2: It's about having an explicit instruction to cmake to prevent any old cached stuff from being used. Removing the build folder likely takes care of this, but it doesn't hurt to be explicit here. Especially to safeguard against any future changes that might change this expectation

tests/ci/run_install_shared_and_static.sh
smittals2: Unsure what the CI failures were because of but my hypothesis is the main culprit is this parallelism. I wasn't able to reproduce the issues. 

The other file state stuff and using --fresh is to be more explicit about our expectations of state during the execution.

tests/ci/run_install_shared_and_static.sh
smittals2: Updated, let me know if that works!

tests/ci/run_install_shared_and_static.sh
smittals2: Updated, let me know if that works!

tests/ci/run_install_shared_and_static.sh
samuel40791765: Yeah I think `--fresh` and the file state issues you've discovered seem more like the culprit. CMake parallelism's generally relatively harmless (ninja uses multiple threads by default and `make -j(threads)` is often used to speed things up).
But nothing to get too fixated over here, this build test is fairly small and fast already.

samuel40791765 (approved):

PR #2180, opened 2025-02-07
SCRUTINICE fixes

smittals2:
### Issues:
`P197895641`, `P197898490`

### Description of changes: 
Scrutinice alerts for a null check and potential memory leak 

### Call-outs:

### Testing:

### DIFF
--- a/crypto/fipsmodule/pqdsa/pqdsa.c
+++ b/crypto/fipsmodule/pqdsa/pqdsa.c
@@ -92,14 +92,12 @@ int PQDSA_KEY_set_raw_keypair_from_seed(PQDSA_KEY *key, CBS *in) {
   //allocate buffers to store key pair
   uint8_t *public_key = OPENSSL_malloc(key->pqdsa->public_key_len);
   if (public_key == NULL) {
-    OPENSSL_PUT_ERROR(CRYPTO, ERR_R_MALLOC_FAILURE);
     return 0;
   }
 
   uint8_t *private_key = OPENSSL_malloc(key->pqdsa->private_key_len);
   if (private_key == NULL) {
     OPENSSL_free(public_key);
-    OPENSSL_PUT_ERROR(CRYPTO, ERR_R_MALLOC_FAILURE);
     return 0;
   }
 
--- a/crypto/fipsmodule/pqdsa/pqdsa.c
+++ b/crypto/fipsmodule/pqdsa/pqdsa.c
@@ -91,10 +91,14 @@ int PQDSA_KEY_set_raw_keypair_from_seed(PQDSA_KEY *key, CBS *in) {
 
   //allocate buffers to store key pair
   uint8_t *public_key = OPENSSL_malloc(key->pqdsa->public_key_len);
-  uint8_t *private_key = OPENSSL_malloc(key->pqdsa->private_key_len);
+  if (public_key == NULL) {
+    OPENSSL_PUT_ERROR(CRYPTO, ERR_R_MALLOC_FAILURE);
+    return 0;
+  }
 
-  // check buffers are allocated
-  if (public_key == NULL || private_key == NULL) {
+  uint8_t *private_key = OPENSSL_malloc(key->pqdsa->private_key_len);
+  if (private_key == NULL) {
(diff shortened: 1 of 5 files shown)


crypto/fipsmodule/pqdsa/pqdsa.c
```diff
@@ -91,10 +91,14 @@ int PQDSA_KEY_set_raw_keypair_from_seed(PQDSA_KEY *key, CBS *in) {
 
   //allocate buffers to store key pair
   uint8_t *public_key = OPENSSL_malloc(key->pqdsa->public_key_len);
-  uint8_t *private_key = OPENSSL_malloc(key->pqdsa->private_key_len);
+  if (public_key == NULL) {
+    OPENSSL_PUT_ERROR(CRYPTO, ERR_R_MALLOC_FAILURE);
```
andrewhop, line 95: OPENSSL_malloc already puts this error into the error queue https://github.com/aws/aws-lc/blob/main/crypto/mem.c#L189-L203

crypto/ml_dsa/ml_dsa_ref/sign.c
```diff
@@ -269,6 +269,7 @@ int ml_dsa_sign_internal(ml_dsa_params *params,
     goto rej;
   }
 
+  GUARD_PTR(sig);
```
andrewhop, line 272: Is this the best spot to put this check? We pass `sig` to ml_dsa_polyveck_pack_w1 on line 229, what happens if it's null there?

Also you can't use our safety macros in the ML-DSA/ML-KEM reference code. The macro uses 0 for error and 1 for success (which is what all the AWS-LC code uses). This reference code use 0 to mark success and !0 for error. We convert between the calling pattern in the function that wraps the reference code here https://github.com/aws/aws-lc/blob/main/crypto/ml_dsa/ml_dsa.c#L105-L106.

crypto/fipsmodule/pqdsa/pqdsa.c
smittals2: Ah I just copied the check that was already here, will get rid of it

crypto/ml_dsa/ml_dsa_ref/sign.c
smittals2: Fair enough, moved it up a level

justsmth (approved):

justsmth (approved):

crypto/ml_dsa/ml_dsa_ref/sign.c
```diff
@@ -26,8 +26,12 @@ static int ml_dsa_keypair_pct(ml_dsa_params *params,
                               uint8_t *pk,
                               uint8_t *sk) {
   uint8_t signature[MLDSA87_SIGNATURE_BYTES];
-  ml_dsa_sign(params, signature, &params->bytes, NULL, 0, NULL, 0, sk);
-  return ml_dsa_verify(params, signature, params->bytes, NULL, 0, NULL, 0, pk) == 0;
+  uint8_t empty_msg[1] = {0};
+  int ret = ml_dsa_sign(params, signature, &params->bytes, empty_msg, 0, NULL, 0, sk);
```
andrewhop, line 30: Is this intentional? empty_message is actually 1 uint8_t long, but you're passing in 0 for the length

crypto/fipsmodule/pqdsa/pqdsa.c
andrewhop: Ohhh, can you fix all of those while you're modifying the this?

crypto/ml_dsa/ml_dsa_ref/sign.c
smittals2: Yea previously this code was passing in NULL which never gets checked. I tried doing a 0 length array, but you cant initialize an array of 0 length on c89/90 I believe. So then I did a 1 length arr. But still passing in length 0 so it's considered an empty message. This is intentional. 

Per Jake, what the message is doesn't matter for the PCT.

andrewhop (approved):

justsmth (approved):

PR #2182, opened 2025-02-08
Update BORINGSSL_FIPS_abort to AWS_LC_FIPS_failure which takes a message.y

andrewhop:
### Description of changes: 
In preparation for the callback change in https://github.com/aws/aws-lc/pull/2162/ this change updates all the place that could abort to pass a message to a new abort function that explains why it's about to abort. 

### Testing:
The existing test-break-kat.go will ensure all the KATs still fail as expected and the messages they print match.

### DIFF
--- a/crypto/fipsmodule/bcm.c
+++ b/crypto/fipsmodule/bcm.c
@@ -184,6 +184,12 @@ extern const uint8_t BORINGSSL_bcm_rodata_start[];
 extern const uint8_t BORINGSSL_bcm_rodata_end[];
 #endif
 
+#define STRING_POINTER_LENGTH 18
+#define MAX_FUNCTION_NAME 32
+#define ASSERT_WITHIN_MSG "FIPS module doesn't span expected symbol (%s). Expected %p <= %p < %p\n"
+#define MAX_WITHIN_MSG_LEN sizeof(ASSERT_WITHIN_MSG) + (3 * STRING_POINTER_LENGTH) + MAX_FUNCTION_NAME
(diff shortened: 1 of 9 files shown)


crypto/fipsmodule/bcm.c
```diff
@@ -184,6 +184,12 @@ extern const uint8_t BORINGSSL_bcm_rodata_start[];
 extern const uint8_t BORINGSSL_bcm_rodata_end[];
 #endif
 
+#define STRING_POINTER_LENGTH 18
+#define MAX_FUNCTION_NAME 32
+#define ASSERT_WITHIN_MSG "FIPS module doesn't span expected symbol (%s). Expected %p <= %p < %p\n"
+#define MAX_WITHIN_MSG_LEN sizeof(ASSERT_WITHIN_MSG) + (3 * STRING_POINTER_LENGTH) + MAX_FUNCTION_NAME
+#define ASSERT_OUTSIDE_MSG "FIPS module spans unexpected symbol (%s), expected %p < %p || %p > %p\n"
+#define MAX_OUTSIDE_MSG_LEN sizeof(ASSERT_OUTSIDE_MSG) + (4 * STRING_POINTER_LENGTH) + MAX_FUNCTION_NAME
```
justsmth, line 192: NP: I don't think so much precision is needed for the `MAX_xxx_MSG_LEN` values. Those arrays are only on the stack in the case of a failure during library initialization -- as long as the array is not so large as to cause compilation to fail, it should be fine. A "large" number like `4096` would be way more than sufficient.

justsmth (approved):

skmcgrail (approved):

PR #2184, opened 2025-02-08
Update aws-lc-nginx.patch for nginx 1.27.4

robvanoostenrijk:
nginx v1.27.4 fails to compile with the existing aws-lc-nginx.patch file.

This updates aws-lc-nginx.patch to allow compiling nginx v1.27.4

Tested on both Linux (Alpine) and FreeBSD.


justsmth (approved):

smittals2:
Thank you for this change. We already had a PR open to fix the nginx patch and just merged it in. Therefore I am closing this PR. Thank you!

PR #2188, opened 2025-02-11
Update ABI Diff Action to work correctly on push events

skmcgrail:

### DIFF
--- a/.github/workflows/abidiff.yml
+++ b/.github/workflows/abidiff.yml
@@ -19,12 +19,12 @@ jobs:
       - uses: actions/checkout@v3
         with:
           repository: aws/aws-lc
-          ref: ${{ github.ref }}
+          ref: ${{ github.event_name == 'push' && github.event.ref || github.event.pull_request.head.sha }}
           path: ${{ github.workspace }}/next
       - uses: actions/checkout@v3
         with:
           repository: aws/aws-lc
-          ref: ${{ github.base.ref }}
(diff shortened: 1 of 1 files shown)


smittals2 (approved):
Looks fine but in the context of AWS-LC, what is considered a push event? 

We usually merge stuff in through PR's right (which I guess the second part of these statements covers?)

.github/workflows/abidiff.yml
```diff
@@ -19,12 +19,12 @@ jobs:
       - uses: actions/checkout@v3
         with:
           repository: aws/aws-lc
-          ref: ${{ github.ref }}
+          ref: ${{ github.event_name == 'push' && github.event.ref || github.event.pull_request.head.sha }}
```
justsmth, line 22: NP: Are parenthesis needed here and below?
```
ref: ${{ (github.event_name == 'push' && github.event.ref) || github.event.pull_request.head.sha }}
```

justsmth (approved):

.github/workflows/abidiff.yml
skmcgrail: GitHub doesn't seem to outright state precedence order for evaluation, but it appears to be left-to-right. See the example here that is basically the same as the line I have: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example

PR #2192, opened 2025-02-12
Add runtime options to break the pairwise consistency test for Ed, ML-KEM, and ML-DSA

andrewhop:
### Description of changes: 
To easily demonstrate the pairwise consistency tests function this change adds support to the existing BORINGSSL_FIPS_BREAK_TEST environment variable.

### Testing:
This adds a new test file for all the runtime tests and runs them from the run_fips_test.sh script on the break-able build.

### DIFF
--- a/util/fipstools/test-runtime-pwct.sh
+++ b/util/fipstools/test-runtime-pwct.sh
@@ -1,0 +1,48 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+# This script attempts to break each of the key generation pair wise consistency tests and checks that doing so
+# seems to work and at least mentions the correct KAT in the output.
+
+set -x
+set -e
+
+TEST_FIPS_BIN="test_build_dir/util/fipstools/test_fips"
+
+if [ ! -f $TEST_FIPS_BIN ]; then
+  echo "$TEST_FIPS_BIN is missing. Run this script from the top level of a"
+  echo "BoringSSL checkout and ensure that ./build-fips-break-test-binaries.sh"
+  echo "has been run first."
+  exit 1
+fi
+
+check_test_output() {
+  local test_name="$1"
+  local output="$2"
+  case "$test_name" in
+    "ECDSA_PWCT")  expected="EC keygen checks failed" ;;
+    "RSA_PWCT")    expected="RSA keygen checks failed" ;;
+    "MLKEM_PWCT")  expected="ML-KEM keygen PCT failed" ;;
+    "MLDSA_PWCT")  expected="ML-DSA keygen PCT failed" ;;
+    "EDDSA_PWCT")  expected="Ed25519 keygen PCT failed" ;;
+    *)             echo "Unknown test: $test_name"; return 1 ;;
+  esac
+
+  if ! echo "$output" | grep -q "$expected"; then
+    echo "Failure for ${test_name} did not contain expected message: '${expected}'"
+    echo "Actual output was: '${output}'"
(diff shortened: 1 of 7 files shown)


crypto/ml_dsa/ml_dsa_ref/sign.c
jakemas, line 1: LGTM!

jakemas (approved):

smittals2 (approved):

util/fipstools/test-runtime-pwct.sh
```diff
@@ -0,0 +1,48 @@
...
+# seems to work and at least mentions the correct KAT in the output.
+
+set -x
+set -e
+
+TEST_FIPS_BIN="test_build_dir/util/fipstools/test_fips"
+
+if [ ! -f $TEST_FIPS_BIN ]; then
+  echo "$TEST_FIPS_BIN is missing. Run this script from the top level of a"
+  echo "BoringSSL checkout and ensure that ./build-fips-break-test-binaries.sh"
```
torben-hansen, line 14:
```suggestion
  echo "AWS-LC checkout and ensure that ./build-fips-break-test-binaries.sh"
```
?

torben-hansen (dismissed):

util/fipstools/test_fips.c
```diff
@@ -36,6 +36,7 @@
 
 #include "../../crypto/fipsmodule/evp/internal.h"
 #include "../../crypto/fipsmodule/kem/internal.h"
+#include "../../crypto/fipsmodule/pqdsa/internal.h"
```
skmcgrail, line 39: Broader question, what is the purpose of `test_fips.c`. My recollection is that it is an abomination of a file with a very large main function, and I'm pretty sure doesn't include support for everything that is FIPS approved. Honestly it feels a bit duplicative to what should be in `crypto_test`?

skmcgrail: For example I'm pretty sure this doesn't have ed25519ph tested :)

andrewhop: You are correct, test_fips is a big file that turns into an executable with some of the algorithms. I think this file was used by BoringSSL to demonstrate certain things to their lab. We now use it in our CI to run/verify the break tests work as expected. In this case I needed to add algorithms to it to trigger the lazy self tests.

The break-tests could use crypto_test, but it's a little hard to write tests for the FIPS failure abort cases.

andrewhop: This doesn't need ed25519ph because this change is only adding support to break the pairwise consistency tests and all the Ed stuff uses the same keygen function. However I did need to add the the pre-hash changes to test breaking the self test and that's already checked in https://github.com/aws/aws-lc/blob/main/util/fipstools/test_fips.c#L416-L430.

torben-hansen (approved):

skmcgrail (approved):

hanno-becker:
This breaks #2176 and it is not immediately obvious how to salvage it, as it embeds runtime breakage logic into the ML-KEM reference implementation.

@andrewhop Would you consider a separate approach whereby every 'breakage-request' is through a dedicated environment variable? That would be easier to accommodate: I could introduce a matching `MLK_BREAK_PCT`, and set it in the compatibility layer between BoringSSL<->mlkem-native if and only if the respective environment variable is set. It would also remove the need for the runtime logic in `boringssl_fips_break_test`.

hanno-becker:
For the time being, I opened https://github.com/pq-code-package/mlkem-native/pull/788/ to basically mirror AWS-LC's runtime behaviour. I'd still prefer not having to do this, though, and understand if/why the runtime logic is needed here.

PR #2193, opened 2025-02-13
Fix C++98 compatibility in our header files

samuel40791765:
### Issues:
Addresses `CryptoAlg-2693`

### Description of changes: 
We encountered this when trying to compile Xtrabackup on gcc versions lower than 10. Apparently scoped enums are a C++11 feature, which the compiler can't understand without additional linker flags. 

```
/home/ubuntu/workplace/percona-xtrabackup/aws-lc/test_build_dir/include/openssl/ssl.h:174:30: warning: scoped enums only available with ‘-std=c++11’ or ‘-std=gnu++11’
  174 | #define BORINGSSL_ENUM_INT : int
      |                              ^~~
...
      |                              ^~~
/home/ubuntu/workplace/percona-xtrabackup/aws-lc/test_build_dir/include/openssl/ssl.h:4783:31: note: in expansion of macro ‘BORINGSSL_ENUM_INT’
 4783 | enum ssl_select_cert_result_t BORINGSSL_ENUM_INT {
      |                               ^~~~~~~~~~~~~~~~~~
make[2]: *** [libmysql/CMakeFiles/clientlib.dir/build.make:149: libmysql/CMakeFiles/clientlib.dir/__/sql-common/client_authentication.cc.o] Error 1
make[2]: Leaving directory '/home/ubuntu/workplace/percona-xtrabackup/build'
```
We also encountered this in https://github.com/aws/aws-lc/commit/42a8dbacd473d2ad95f9d0d5e47ba95a4e92f4c3, which prompted us to move it to `ssl.h`. Upon closer examination, I don't think the C++11 usage is a necessity especially with the pure C fallback that was included. Predeclaring enums reduces compilation dependencies and build times, but our build seems to work fine without it. Removing this would let us encounter less build issues with projects that aren't expecting C++ in our header files.

### Call-outs:
N/A

### Testing:
CI

### DIFF
--- a/tests/coding_guidelines/c99_cplusplus98_test.sh
+++ b/tests/coding_guidelines/c99_cplusplus98_test.sh
@@ -24,8 +24,7 @@ INCLUDE_FILES=`ls $INCLUDE_DIR/openssl/*.h | grep -v $INCLUDE_DIR/openssl/arm_ar
 # some non-ISO practices, but not all — only those for which ISO C requires a
 # diagnostic, and some others for which diagnostics have been added."
 # https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
-
-${CC} -std=c99 -c -I${INCLUDE_DIR} -include ${INCLUDE_FILES} -Wpedantic -fsyntax-only -Werror
+${CC} -std=c99 -c -I${INCLUDE_DIR} $(echo ${INCLUDE_FILES} | sed 's/[^ ]* */-include &/g') -Wpedantic -fsyntax-only -Werror ./tests/compiler_features_tests/builtin_swap_check.c
 
 # AWS C SDKs conforms to C99. They set `C_STANDARD 99` which will set the
 # flag `-std=gnu99`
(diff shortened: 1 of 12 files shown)


justsmth:
Pre-push check is failing:
```
[Container] 2025/02/13 01:52:19.380053 Running command (cd util && go run ./doc.go)
Failed to generate output: while parsing include/openssl/ssl.h: comment for "SSL_CTX_get0_chain" doesn't seem to match line include/openssl/ssl.h:1274

exit status 1
```

justsmth (approved):

andrewhop (commented):
Can you update c99_gcc_test.sh to also check for C++ header compatibility (and maybe rename the file). Also double check that it's running, I'm not sure anything defines AWSLC_C99_TEST right now....

justsmth (approved):

tests/coding_guidelines/c99_cplusplus98_test.sh
```diff
@@ -24,8 +24,7 @@ INCLUDE_FILES=`ls $INCLUDE_DIR/openssl/*.h | grep -v $INCLUDE_DIR/openssl/arm_ar
 # some non-ISO practices, but not all — only those for which ISO C requires a
 # diagnostic, and some others for which diagnostics have been added."
 # https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
-
-${CC} -std=c99 -c -I${INCLUDE_DIR} -include ${INCLUDE_FILES} -Wpedantic -fsyntax-only -Werror
+${CC} -std=c99 -c -I${INCLUDE_DIR} $(echo ${INCLUDE_FILES} | sed 's/[^ ]* */-include &/g') -Wpedantic -fsyntax-only -Werror ./tests/compiler_features_tests/builtin_swap_check.c
```
andrewhop, line 27: What is this sed removing from the list? And why add the builtin_swap_check to all of these targets?

andrewhop (approved):

tests/coding_guidelines/c99_cplusplus98_test.sh
samuel40791765: sed's mainly in charge of prepending an `-include` to each file in `$INCLUDE_FILES`. What we had previously didn't actually work :(. This also needs a basic "c" file to compile, so I chose `builtin_swap_check` since that was a good existing example that works with older compilers

PR #2194, opened 2025-02-13
Add support to export ML-DSA key-pairs in seed format

jakemas:
### Issues:
Resolves #CryptoAlg-2918 
#ACCP-130

### Description of changes: 
Support the ability to export ML-DSA key seeds. We modify the core algorithm implementation to store the `seed` used during key generation. This will allow the key pair to be reconstructed at a later stage from just the seed.

This is performed  within `ml_dsa_keypair`, which has been modified to accept an addition argument `seed` that is a pointer to **output** array of `ML_DSA_SEEDBYTES` bytes.

```
int ml_dsa_keypair(ml_dsa_params *params, uint8_t *pk, uint8_t *sk, uint8_t *seed) {
- uint8_t seed[ML_DSA_SEEDBYTES];
  if (!RAND_bytes(seed, ML_DSA_SEEDBYTES)) {
    return -1;
  }
  ml_dsa_keypair_internal(params, pk, sk, seed);
- OPENSSL_cleanse(seed, sizeof(seed));
  return 0;
}
```

These changes bubble up to the `ml_dsa.c` definitions of keygen, that are now modified to support the provided buffer to store the seed:
```
int ml_dsa_44_keypair(uint8_t *public_key   /* OUT */,
                      uint8_t *private_key  /* OUT */,
+                     uint8_t *seed         /* OUT */) {
  ml_dsa_params params;
  ml_dsa_44_params_init(&params);
  return (ml_dsa_keypair(&params, public_key, private_key, seed) == 0);
}
```

We store the seed in the `PQDSA_KEY` struct during `pkey_pqdsa_keygen`:
```
struct pqdsa_key_st {
  const PQDSA *pqdsa;
  uint8_t *public_key;
  uint8_t *private_key;
+ uint8_t *seed;
};
```

### API Changes

This PR modifies the ASN.1 encoding function for `PQDSA` keys. The function `pqdsa_priv_encode` now encodes the associated `pqdsa->key->seed`. As such the EVP API `EVP_marshal_private_key` will export the private seed. This has been noted in documentation.

### Performance Impact

Converting to seed-based storage for both public and private keys yields the following improvements:
```
ML-DSA-44: 99.17% (121x) reduction (3,840 bytes saved per key-pair).
ML-DSA-65: 99.46% (187x) reduction (5,952 bytes saved per key-pair).
ML-DSA-87: 99.57% (234x) reduction (7,456 bytes saved per key-pair).
```
Converting to seed-based storage for private keys yields the following improvements:
```
ML-DSA-44: 98.75% (80x) reduction (2,528 bytes saved per private key).
ML-DSA-65: 99.21% (126x) reduction (5,952 bytes saved per private key).
ML-DSA-87: 99.35% (153x) reduction (7,456 bytes saved per private key).
```
The proposed seed-based approach achieves an average storage reduction of 99.4% across all ML-DSA variants.

### Call-outs:

- FIPS Compliance: I'm glad you're asking, yes this is compliant with FIPS, NIST have published PQC FAQs specifically to address this exact implementation: https://csrc.nist.gov/Projects/post-quantum-cryptography/faqs#Rdc7.

- [3692f72](https://github.com/aws/aws-lc/pull/2194/commits/3692f7247aa7d846e753d4e3395f02617ad84bd5) implemented `PQDSA_KEY_get_priv_raw_seed` if it is needed in future.

- Modified ASN.1 method signatures to include `oid `parameter for better format handling

### Testing:
Added a failure mode test to `ParsePrivateKey` for the case that a key does not have an associated `seed`.


crypto/evp_extra/p_pqdsa_test.cc
```diff
@@ -1577,6 +1596,107 @@ TEST_P(PQDSAParameterTest, MarshalParse) {
             Bytes(pkey->pkey.pqdsa_key->public_key, GetParam().public_key_len));
 }
 
+// Helper function that:
+// 1. Creates a memory BIO
+// 2. Writes the provided DER data to BIO in PEM format
+// 3. Determines resulting PEM size and allocates buffer
+// 4. Reads PEM data into output buffer
+// 5. Returns the PEM string and length via out parameters
+static bool DER_to_PEM(const uint8_t* der, size_t der_len,
```
jakemas, line 1605: If this is excessive, I don't mind removing it. I just wanted to prove equivalence to some "KAT" and the ones from https://datatracker.ietf.org/doc/draft-ietf-lamps-dilithium-certificates/ serve as exactly that -- except the standard doesn't provide DER/RAW representations. Since I'm building the entire KAT from just the seed, it seems a shame not to reach a full known answer. Plus, this tests out ASN.1 parsing of seed format, rather than full format which is done as part of the X.509 parsing tests (I think it's X.509).

crypto/fipsmodule/pqdsa/pqdsa.c
```diff
@@ -109,10 +118,13 @@ int PQDSA_KEY_set_raw_keypair_from_seed(PQDSA_KEY *key, CBS *in) {
...
   }
 
+  // copy the seed data
+  OPENSSL_memcpy(seed, CBS_data(in), key->pqdsa->keygen_seed_len);
+
   // set the public and private key
   key->public_key = public_key;
   key->private_key = private_key;
-
+  key->seed = seed;
```
jakemas, line 127: Andrew I know you're going to be reading this and thinking that I should just let `key` take ownership of the memory and assign directly like `key->CBS_data(in)` but `CBS_data(in)` is a `const` so it fails. We don't want `seed` to be `const` in `int ml_dsa_keypair(ml_dsa_params *params, uint8_t *pk, uint8_t *sk, uint8_t *seed);` because `seed` is input as an empty buffer of allocated `ML_DSA_SEEDBYTES` bytes, and is output as the seed.

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -18,6 +18,40 @@ static void pqdsa_free(EVP_PKEY *pkey) {
   pkey->pkey.pqdsa_key = NULL;
 }
 
+static int pqdsa_get_priv_raw_seed(PQDSA_KEY *key, const PQDSA *pqdsa,
+                                  uint8_t *out,size_t *out_len) {
```
skmcgrail, line 22:
```suggestion
static int pqdsa_get_priv_raw_seed(PQDSA_KEY *key, const PQDSA *pqdsa,
                                  uint8_t *out,0 size_t *out_len) {
GUARD_PTR(key);
GUARD_PTR(out_len);
```

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -18,6 +18,40 @@ static void pqdsa_free(EVP_PKEY *pkey) {
...
+  if (key->seed == NULL) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_NO_KEY_SET);
+    return 0;
+  }
+
+  if (*out_len != key->pqdsa->keygen_seed_len) {
```
skmcgrail, line 28:
```suggestion
  if (*out_len < key->pqdsa->keygen_seed_len) {
```
A buffer should be allowed to be bigger I'd imagine? The point is that we set `*out_len` the actual value of what we wrote. Though I do realize this i gated later, I don't know if it needs to be gated here per-se.

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -18,6 +18,40 @@ static void pqdsa_free(EVP_PKEY *pkey) {
...
+    return 0;
+  }
+
+  OPENSSL_memcpy(out, key->seed, pqdsa->keygen_seed_len);
+  *out_len = pqdsa->keygen_seed_len;
+  return 1;
+}
+
+static int pqdsa_get_priv_raw_key(PQDSA_KEY *key, const PQDSA *pqdsa,
+                                  uint8_t *out,size_t *out_len) {
```
skmcgrail, line 39:
```suggestion
static int pqdsa_get_priv_raw_key(PQDSA_KEY *key, const PQDSA *pqdsa,
                                  uint8_t *out, size_t *out_len) {
GUARD_PTR(key);
GUARD_PTR(out_len);
```

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -18,6 +18,40 @@ static void pqdsa_free(EVP_PKEY *pkey) {
...
+  }
+
+  OPENSSL_memcpy(out, key->private_key, pqdsa->private_key_len);
+  *out_len = pqdsa->private_key_len;
+  return 1;
+}
+
 static int pqdsa_get_priv_raw(const EVP_PKEY *pkey, uint8_t *out,
                                    size_t *out_len) {
   if (pkey->pkey.pqdsa_key == NULL) {
```
skmcgrail, line 57: We should also check that `pkey` and `out_len` are not NULL here as well.

skmcgrail (commented):
Have we just considered exposing access to the opaque `PQDSA_KEY` type stored in the EVP and then having more proper access functions like get_private_key_seed etc using that type?  I'm not a big fan of the "raw" key meaning on EVP because it differs based on the key types, and now can very on the size dimension which seems really goofy.

crypto/evp_extra/p_pqdsa_asn1.c
jakemas: Thank you, added in [c307952](https://github.com/aws/aws-lc/pull/2194/commits/c30795274cf00bed09803b8f5fe6114112fb1422)

crypto/evp_extra/p_pqdsa_asn1.c
jakemas: Thank you, added in [c307952](https://github.com/aws/aws-lc/pull/2194/commits/c30795274cf00bed09803b8f5fe6114112fb1422)

jakemas:
> Have we just considered exposing access to the opaque `PQDSA_KEY` type stored in the EVP and then having more proper access functions like get_private_key_seed etc using that type? I'm not a big fan of the "raw" key meaning on EVP because it differs based on the key types, and now can very on the size dimension which seems really goofy.

Agreed, thank you for the feedback. I've updated the implementation ([c307952](https://github.com/aws/aws-lc/pull/2194/commits/c30795274cf00bed09803b8f5fe6114112fb1422)) so that `EVP_PKEY_get_raw_private_key` is unchanged. Instead, I created `PQDSA_KEY_get_priv_raw_seed` in `pqdsa.c` so that we can access the seed through this function. I haven't exported or exposed that new function yet, as I am under the impression that exporting these seeds as DER encodings using ` EVP_marshal_private_key_v2` is sufficient for now. It would be simple to add an external API from here, but we might like to land on a few decisions around that part first.

crypto/evp_extra/p_pqdsa_asn1.c
jakemas: added to `pqdsa_get_priv_raw` in [c307952](https://github.com/aws/aws-lc/pull/2194/commits/c30795274cf00bed09803b8f5fe6114112fb1422)

crypto/evp_extra/p_pqdsa_asn1.c
jakemas: added to `PQDSA_KEY_get_priv_raw_seed` in [c307952](https://github.com/aws/aws-lc/pull/2194/commits/c30795274cf00bed09803b8f5fe6114112fb1422)

crypto/fipsmodule/pqdsa/pqdsa.c
```diff
@@ -101,6 +104,12 @@ int PQDSA_KEY_set_raw_keypair_from_seed(PQDSA_KEY *key, CBS *in) {
     return 0;
   }
 
+  uint8_t *seed = OPENSSL_malloc(key->pqdsa->keygen_seed_len);
+  if (seed == NULL) {
+    OPENSSL_free(seed);
```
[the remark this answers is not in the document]
WillChilds-Klein: oh, i see. i think you'd need to free public_key and private_key here, rather than `seed`.

crypto/fipsmodule/pqdsa/pqdsa.c
```diff
@@ -109,10 +118,13 @@ int PQDSA_KEY_set_raw_keypair_from_seed(PQDSA_KEY *key, CBS *in) {
     return 0;
```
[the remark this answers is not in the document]
jakemas: Yes, and pub/priv, added in https://github.com/aws/aws-lc/pull/2194/commits/09bd679c18869c07445797f5c8d7a1b1583ced59

crypto/fipsmodule/pqdsa/pqdsa.c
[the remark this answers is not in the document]
jakemas: got it! https://github.com/aws/aws-lc/pull/2194/commits/09bd679c18869c07445797f5c8d7a1b1583ced59

WillChilds-Klein (approved):

WillChilds-Klein:
@jakemas -- what happens if we parse a private key in "expanded" form, then try to extract the seed? It seems to me this should fail.

jakemas:
> @jakemas -- what happens if we parse a private key in "expanded" form, then try to extract the seed? It seems to me this should fail.

Correct, asserted that with a failure mode test in [6d6127f](https://github.com/aws/aws-lc/pull/2194/commits/6d6127f97ee0ad6085a0c34f7b60094be30397ac)

WillChilds-Klein (approved):

jakemas:
> f178391

Thank you, added here [f178391](https://github.com/aws/aws-lc/pull/2194/commits/f178391c3ac9ee69ca01d4ff679d81aeb77b78e0)

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -129,6 +132,30 @@ static int pqdsa_pub_encode(CBB *out, const EVP_PKEY *pkey) {
...
+  PQDSA_KEY *key = pkey->pkey.pqdsa_key;
+  const PQDSA *pqdsa = key->pqdsa;
+  if (key->seed == NULL) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_ENCODE_ERROR);
+    return 0;
+  }
+  // See https://datatracker.ietf.org/doc/draft-ietf-lamps-dilithium-certificates/ section 6.
+  CBB pkcs8, algorithm, oid, seed;
+  if (!CBB_add_asn1(out, &pkcs8, CBS_ASN1_SEQUENCE) ||
+      !CBB_add_asn1_uint64(&pkcs8, 0 /* version */) ||
```
skmcgrail, line 145:
```suggestion
      !CBB_add_asn1_uint64(&pkcs8, PKCS8_VERSION_ONE) ||
```

skmcgrail: Also out of general curiosity I know the draft standard uses PKCS#8 v2 encoding so that it can include the public key. Is there a reason we aren't supporting serialization of v2 as well? (I can see why we would want to support the v1 encoding as well). I guess why not both?

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -235,7 +262,7 @@ const EVP_PKEY_ASN1_METHOD pqdsa_asn1_meth = {
   pqdsa_pub_cmp,
   pqdsa_priv_decode,
   pqdsa_priv_encode,
-  NULL /*priv_encode_v2*/,
+  pqdsa_priv_encode_seed,
```
skmcgrail, line 265: But this is PKCS#8 v1 encoding you are doing here, so this shouldn't be set as the v2 encoding function?

crypto/evp_extra/p_pqdsa_asn1.c
jakemas: No reason other than PR scope and the original request --  I wanted to add try to keep this PR to a single feature.

jakemas:
Ahh https://github.com/aws/aws-lc/pull/2194/commits/7afa89adc363207aa966e127307279dacb1b837b wont work because the IETF standard uses `VERSION` 0

crypto/fipsmodule/pqdsa/pqdsa.c
```diff
@@ -101,18 +104,31 @@ int PQDSA_KEY_set_raw_keypair_from_seed(PQDSA_KEY *key, CBS *in) {
...
+    OPENSSL_free(public_key);
+    OPENSSL_free(private_key);
+    OPENSSL_free(seed);
     OPENSSL_PUT_ERROR(EVP, EVP_R_DECODE_ERROR);
     return 0;
   }
 
+  // copy the seed data
+  OPENSSL_memcpy(seed, CBS_data(in), key->pqdsa->keygen_seed_len);
+
```
andrewhop, line 127: You shouldn't use memcpy with a CBS, that skips all the nice length safety checks it does internally. We should avoid accessing the underlying buffer using CBS_data unless we actually need the pointer to pass to another function. This should use CBS_copy_bytes.

jakemas: Great call out! Thank you, added in [b482db7](https://github.com/aws/aws-lc/pull/2194/commits/b482db7e346eec1314656871069dfc80fc7c18e7)

crypto/evp_extra/p_pqdsa_asn1.c
jakemas: Support full v2 encoding added in [9e86f22](https://github.com/aws/aws-lc/pull/2194/commits/9e86f22d1411ac6f5016f12cfb479616ee4e094a)

jakemas:
Okay, after offline conversation we decided to align with the draft spec RFC and encode private ML-DSA keys as 32 byte seeds. As such, we leave `EVP_marshal_private_key_v2` un-implemented for PQDSA, and modify the implementation of `EVP_marshal_private_key` to encode the 32-byte private key seed.

I have removed the utility function `PQDSA_KEY_get_priv_raw_seed` in [3692f72](https://github.com/aws/aws-lc/pull/2194/commits/3692f7247aa7d846e753d4e3395f02617ad84bd5) but call it out here in case of future need.

WillChilds-Klein (dismissed):

skmcgrail (dismissed):

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -156,31 +156,33 @@ static int pqdsa_priv_decode(EVP_PKEY *out, CBS *params, CBS *key, CBS *pubkey)
...
-      // PQDSA_KEY_set_raw_keypair_from_seed sets the appropriate error.
+    if (CBS_len(&seed) != out->pkey.pqdsa_key->pqdsa->keygen_seed_len) {
+      OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_BUFFER_SIZE);
       return 0;
     }
+
+    return PQDSA_KEY_set_raw_keypair_from_seed(out->pkey.pqdsa_key, &seed);
+  }
+
+  else {
```
WillChilds-Klein, line 182: minor style nit: collapse these

WillChilds-Klein (approved):

crypto/evp_extra/p_pqdsa_asn1.c
jakemas: fixed in [6164a3d](https://github.com/aws/aws-lc/pull/2194/commits/6164a3d7dc96b266df70c5822d0a675828d123fe)

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -153,49 +156,68 @@ static int pqdsa_priv_decode(EVP_PKEY *out, CBS *params, CBS *key, CBS *pubkey)
...
+      !CBB_add_asn1_uint64(&pkcs8, PKCS8_VERSION_ONE /* version */) ||
       !CBB_add_asn1(&pkcs8, &algorithm, CBS_ASN1_SEQUENCE) ||
       !CBB_add_asn1(&algorithm, &oid, CBS_ASN1_OBJECT) ||
       !CBB_add_bytes(&oid, pqdsa->oid, pqdsa->oid_len) ||
       !CBB_add_asn1(&pkcs8, &private_key, CBS_ASN1_OCTETSTRING) ||
-      !CBB_add_bytes(&private_key, key->private_key, pqdsa->private_key_len) ||
+      !CBB_add_asn1(&private_key, &seed_choice, CBS_ASN1_CONTEXT_SPECIFIC | 0) ||
+      !CBB_add_bytes(&seed_choice, key->seed, pqdsa->keygen_seed_len) ||
+      //!CBB_add_asn1(&private_key, &seed_choice, CBS_ASN1_OCTETSTRING) ||
+      //!CBB_add_bytes(&seed_choice, key->private_key, pqdsa->private_key_len) ||
```
WillChilds-Klein, line 220: intentional?

```suggestion
```

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -153,49 +156,68 @@ static int pqdsa_priv_decode(EVP_PKEY *out, CBS *params, CBS *key, CBS *pubkey)
...
+      return 0;
+    }
+
+    if (CBS_len(&expanded_key) != out->pkey.pqdsa_key->pqdsa->private_key_len) {
+      OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_BUFFER_SIZE);
+      return 0;
+    }
+
+    return PQDSA_KEY_set_raw_private_key(out->pkey.pqdsa_key, &expanded_key);
+  }else {
```
WillChilds-Klein, line 196:
```suggestion
  } else {
```

WillChilds-Klein (dismissed):

crypto/evp_extra/p_pqdsa_asn1.c
jakemas: nope, removed in [7796da1](https://github.com/aws/aws-lc/pull/2194/commits/7796da14ae8007c4e3c5da79773ab722a04bac56)

crypto/evp_extra/p_pqdsa_asn1.c
jakemas: [7796da1](https://github.com/aws/aws-lc/pull/2194/commits/7796da14ae8007c4e3c5da79773ab722a04bac56)

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -158,31 +158,20 @@ static int pqdsa_priv_decode(EVP_PKEY *out, CBS *params, CBS *key, CBS *pubkey)
...
-
-    if (CBS_len(&seed) != out->pkey.pqdsa_key->pqdsa->keygen_seed_len) {
-      OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_BUFFER_SIZE);
-      return 0;
-    }
-
-    return PQDSA_KEY_set_raw_keypair_from_seed(out->pkey.pqdsa_key, &seed);
-  } else if (CBS_peek_asn1_tag(key, CBS_ASN1_OCTETSTRING)) {
-    // Case 2: expandedKey OCTET STRING
+  if (CBS_len(key) == out->pkey.pqdsa_key->pqdsa->keygen_seed_len) {
```
skmcgrail, line 168: This should still need to peek to see that it is a `CBS_ASN1_OCTETSTRING` type and then specifically get that asn1 type.

jakemas: [7e68294](https://github.com/aws/aws-lc/pull/2194/commits/7e68294a918491fb59e8d8a48635298e504f934e)

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -153,49 +156,66 @@ static int pqdsa_priv_decode(EVP_PKEY *out, CBS *params, CBS *key, CBS *pubkey)
...
     }
+
+    return PQDSA_KEY_set_raw_keypair_from_seed(out->pkey.pqdsa_key, &seed);
+  } else if (CBS_peek_asn1_tag(key, CBS_ASN1_CONTEXT_SPECIFIC | 0)) {
+    // Case 2: expandedKey [0] OCTET STRING
+    CBS expanded_key;
+    if (!CBS_get_asn1(key, &expanded_key, CBS_ASN1_CONTEXT_SPECIFIC | 0)) {
+      OPENSSL_PUT_ERROR(EVP, EVP_R_DECODE_ERROR);
+      return 0;
+    }
```
WillChilds-Klein, line 188:
```suggestion
    CBS choice_0, expanded_key;
    if (!CBS_get_asn1(key, &choice_0, CBS_ASN1_CONTEXT_SPECIFIC | 0) ||
        !CBS_get_asn1(&choice_0, &expanded_key, CBS_ASN1_OCTETSTRING)) {
      OPENSSL_PUT_ERROR(EVP, EVP_R_DECODE_ERROR);
      return 0;
    }
```

WillChilds-Klein: see prior comment

crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -153,49 +156,66 @@ static int pqdsa_priv_decode(EVP_PKEY *out, CBS *params, CBS *key, CBS *pubkey)
...
+
```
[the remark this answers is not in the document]
jakemas: Chatted offline to discuss! Looked into X6901 Section 8.14 Encoding of a value of a prefixed type. Our reading of 8.14.4 If implicit tagging was used in the definition of the type, then: (b) the contents octets shall be the same as the contents octets of the base encoding.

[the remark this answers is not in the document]
WillChilds-Klein: Ok, looks like the context-specific tag `[0]` wholly replaces the `OCTET STRING` tag, and _does not wrap it_. So, please disregard this comment.

Shout out to @skmcgrail for tracking down this bit from "Recommendation ITU-T X.690 International Standard 8825-1";

> 8.14.2	The encoding of a tagged value shall be derived from the complete encoding of the corresponding data value of the type appearing in the "TaggedType" notation (called the base encoding) as specified in 8.14.3 and 8.14.4.
> ...
> 8.14.4	If implicit tagging was used in the definition of the type, then:
> a)	the encoding shall be constructed if the base encoding is constructed, and shall be primitive otherwise; and
> b)	the contents octets shall be the same as the contents octets of the base encoding.

WillChilds-Klein (approved):

PR #2195, opened 2025-02-13
ML-KEM: Move FIPS-abort upon PCT failure to top-level ML-KEM API

hanno-becker:

Discussed with: @andrewhop 

-----

If AWSLC_FIPS is set, ML-KEM includes a PCT check after key generation, in compliance with the FIPS-203 IG
https://csrc.nist.gov/csrc/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf If the PCT fails, AWS_LC_FIPS_failure is called to abort operation.

This PR hoists the call to AWS_LC_FIPS_failure one level up, into the toplevel ML-KEM API of crypto/fipsmodule/ml_kem/ml_kem.h: When the PCT fails, the lower-level key generation API that invokes the PCT returns -1, which is caught by the top-level ML-KEM API and converted into a call to AWS_LC_FIPS_failure. There is no other failure condition during key generation.

When the PCT fails in the randomized key generation API, the coins are still cleansed before propagating the PCT error. It could be considered whether the PK and SK should be cleansed in this case, too, but this commit does not change behvaiour in this regard.

The change facilitates replacing the ML-KEM reference implementation, specifically mlkem-native (see #2176), since only the return-value behaviour needs to be maintained, but an alternative implementation need not be aware of the abort macro AWS_LC_FIPS_failure.

### DIFF
--- a/crypto/fipsmodule/ml_kem/ml_kem.c
+++ b/crypto/fipsmodule/ml_kem/ml_kem.c
@@ -34,16 +34,32 @@ int ml_kem_512_keypair_deterministic_no_self_test(uint8_t *public_key  /* OUT */
                                                   uint8_t *secret_key  /* OUT */,
                                                   const uint8_t *seed  /* IN */) {
   ml_kem_params params;
+  int res;
   ml_kem_512_params_init(&params);
-  return ml_kem_keypair_derand_ref(&params, public_key, secret_key, seed);
+  res = ml_kem_keypair_derand_ref(&params, public_key, secret_key, seed);
+#if defined(AWSLC_FIPS)
+  /* PCT failure is the only failure condition for key generation. */
+  if (res != 0) {
+      AWS_LC_FIPS_failure("ML-KEM keygen PCT failed");
+  }
+#endif
+  return res;
 }
 
 int ml_kem_512_keypair(uint8_t *public_key /* OUT */,
                            uint8_t *secret_key /* OUT */) {
   boringssl_ensure_ml_kem_self_test();
+  int res;
   ml_kem_params params;
   ml_kem_512_params_init(&params);
-  return ml_kem_keypair_ref(&params, public_key, secret_key);
+  res = ml_kem_keypair_ref(&params, public_key, secret_key);
+#if defined(AWSLC_FIPS)
+  /* PCT failure is the only failure condition for key generation. */
+  if (res != 0) {
+      AWS_LC_FIPS_failure("ML-KEM keygen PCT failed");
+  }
+#endif
+  return res;
 }
 
 int ml_kem_512_encapsulate_deterministic(uint8_t *ciphertext       /* OUT */,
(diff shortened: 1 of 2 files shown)


andrewhop (dismissed):

jakemas (approved):
Looks good to me

crypto/fipsmodule/ml_kem/ml_kem.c
```diff
@@ -34,16 +34,36 @@ int ml_kem_512_keypair_deterministic_no_self_test(uint8_t *public_key  /* OUT */
                                                   uint8_t *secret_key  /* OUT */,
                                                   const uint8_t *seed  /* IN */) {
   ml_kem_params params;
+  int res;
   ml_kem_512_params_init(&params);
-  return ml_kem_keypair_derand_ref(&params, public_key, secret_key, seed);
+  res = ml_kem_keypair_derand_ref(&params, public_key, secret_key, seed);
+#if defined(AWSLC_FIPS)
```
torben-hansen, line 40: Can this be implemented in `AWS_LC_FIPS_failure`?

Maybe can avoid littering the code-base with `if defined(AWS_FIPS)`.

crypto/fipsmodule/ml_kem/ml_kem.c
```diff
@@ -34,16 +34,36 @@ int ml_kem_512_keypair_deterministic_no_self_test(uint8_t *public_key  /* OUT */
...
+  /* PCT failure is the only failure condition for key generation. */
+  if (res != 0) {
+      AWS_LC_FIPS_failure("ML-KEM keygen PCT failed");
+  }
+#else
+  assert(res == 0);
```
torben-hansen, line 46: Why capture here? All other code-paths normally throw it up the stack.

crypto/fipsmodule/ml_kem/ml_kem.c
hanno-becker: I stuck with the precedent here and prefer not to change it -- but it's ultimately your call.

crypto/fipsmodule/ml_kem/ml_kem.c
hanno-becker: This keeps the previous behavior from https://github.com/aws/aws-lc/pull/2195/files#diff-af6c334a9ca9f10f18987e623c41f74351fd662f0e01019d54a0716cdb710904L86. I'm OK either way.

torben-hansen: IMO remove

hanno-becker: Ack, will remove.

hanno-becker: Done.

hanno-becker:
@andrewhop @jakemas I rebased and addressed @torben-hansen's request to remove the `assert(res == 0)`. Can you please re-review?

crypto/fipsmodule/ml_kem/ml_kem.c
```diff
@@ -34,16 +34,32 @@ int ml_kem_512_keypair_deterministic_no_self_test(uint8_t *public_key  /* OUT */
                                                   uint8_t *secret_key  /* OUT */,
                                                   const uint8_t *seed  /* IN */) {
   ml_kem_params params;
+  int res;
   ml_kem_512_params_init(&params);
-  return ml_kem_keypair_derand_ref(&params, public_key, secret_key, seed);
+  res = ml_kem_keypair_derand_ref(&params, public_key, secret_key, seed);
+#if defined(AWSLC_FIPS)
+  /* PCT failure is the only failure condition for key generation. */
```
torben-hansen, line 41: can fixup comment style in a subsequent PR. We use `//`, generally.

torben-hansen (approved):

skmcgrail (approved):

crypto/fipsmodule/ml_kem/ml_kem.c
hanno-becker: Ack

PR #2196, opened 2025-02-13
Improve tool-openssl compatability for x509 and verify subcommands

skmcgrail:
### Description of changes: 
* OpenSSL supports passing lower-case `pem` or `der` when using the `-inform` argument to the `x509` sub-command.
* Add support for the `-untrusted` flag to the `verify` sub-command which allows for passing in a set of one or more chain certificates which may be used in the certificate chain building and validation.

### DIFF
--- a/tool-openssl/verify.cc
+++ b/tool-openssl/verify.cc
@@ -17,7 +17,9 @@ static const argument_t kArguments[] = {
     {"-untrusted", kOptionalArgument,
      "A file of untrusted certificates to be used for chain building. The "
      "file should contain one or more certificates in PEM format."},
-    { "", kOptionalArgument, "" }};
+    {"-x509_strict", kBooleanArgument,
+     "This argument is a no-op. AWS-LC is always strict."},
+    {"", kOptionalArgument, ""}};
 
 // setup_verification_store sets up an X509 certificate store for verification.
 // It configures the store with file and directory lookups. It loads the
--- a/tool-openssl/verify.cc
+++ b/tool-openssl/verify.cc
@@ -105,14 +105,24 @@ static int check(X509_STORE *ctx, const char* chainfile, const char *certfile) {
       return 0;
     }
     bssl::UniquePtr<BIO> chain_bio(BIO_new_fp(chain_file.get(), BIO_NOCLOSE));
+    size_t count = 0;
     while(1) {
       bssl::UniquePtr<X509> chain_cert(PEM_read_bio_X509(chain_bio.get(), NULL, NULL, NULL));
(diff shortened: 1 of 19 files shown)


torben-hansen (dismissed):

tool-openssl/verify.cc
```diff
@@ -88,14 +93,33 @@ static int cb(int ok, X509_STORE_CTX *ctx) {
...
+  if (chainfile) {
+    ScopedFILE chain_file(fopen(chainfile, "rb"));
+    if (!chain_file) {
+      fprintf(stderr, "error %s: reading certificate failed\n", certfile);
+      return 0;
+    }
+    bssl::UniquePtr<BIO> chain_bio(BIO_new_fp(chain_file.get(), BIO_NOCLOSE));
+    while(1) {
+      bssl::UniquePtr<X509> chain_cert(PEM_read_bio_X509(chain_bio.get(), NULL, NULL, NULL));
+      if(chain_cert.get() == nullptr) {
```
smittals2, line 110: I see that this acts as our condition to exit out of the loop. But could there be other reasons the read operation fails (eg. invalid cert)? In that case is it okay to just break and continue on instead of returning 0 for error?

tool-openssl/x509.cc
```diff
@@ -6,6 +6,8 @@
 #include <openssl/rsa.h>
 #include "internal.h"
 #include <ctime>
+#include <algorithm>
```
smittals2, line 9: what is the algorithm include needed for?

smittals2:
Can we add a test for the new -untrusted option? 

I remember there were some issues with output being differently formatted between OpenSSL and AWS-LC (I don't remember if this was X509 or verify though). If that is why this test was not added, ignore the comment.

tool-openssl/x509.cc
```diff
@@ -151,8 +163,7 @@ bool X509Tool(const args_list_t &args) {
 
   // Check -inform has a valid value
   if(!inform.empty()) {
-    if (inform != "DER" && inform != "PEM") {
-      fprintf(stderr, "Error: '-inform' option must specify a valid encoding DER|PEM\n");
```
torben-hansen, line 155: Can still keep this

torben-hansen (dismissed):

tool-openssl/x509.cc
skmcgrail: Leftover include from when I was originally going to use `std::transform`. Will remove.

torben-hansen (approved):

smittals2 (approved):
Intentional to

PR #2197, opened 2025-02-15
Remove access() call from Snapsafe detection

smittals2:
### Description of changes: 
This change uses stat() instead of access() to check for the existence of the Sysgenid device. This is more accurate to ascertain whether the device exists on a machine. access() could fail for various other reasons such as missing permissions. 

### DIFF
--- a/crypto/fipsmodule/rand/snapsafe_detect.c
+++ b/crypto/fipsmodule/rand/snapsafe_detect.c
@@ -9,6 +9,7 @@
 #include <fcntl.h>
 #include <stdlib.h>
 #include <sys/mman.h>
+#include <sys/stat.h>
 #include <unistd.h>
 #include "../delocate.h"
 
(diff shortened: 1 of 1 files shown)


justsmth (dismissed):

torben-hansen (dismissed):

justsmth (approved):

crypto/fipsmodule/rand/snapsafe_detect.c
```diff
@@ -21,28 +22,18 @@ DEFINE_STATIC_ONCE(aws_snapsafe_init)
...
+  struct stat buff;
+  if (stat(CRYPTO_get_sysgenid_path(), &buff) != 0) {
     return;
   }
-  *snapsafety_state_bss_get() = SNAPSAFETY_STATE_FAILED_INITIALISE;
 
+  *snapsafety_state_bss_get() = SNAPSAFETY_STATE_FAILED_INITIALISE;
   int fd_sgc = open(CRYPTO_get_sysgenid_path(), O_RDONLY);
-  if (fd_sgc == -1) {
+  if (fd_sgc != 0) {
```
justsmth, line 36: On success `open` returns the file descriptor, which is non-negative. I think the condition should be:
```
if (fd_sgc <= 0) {
```

smittals2: I think 0 is technically a non-negative integer, will make it < 0.

justsmth (approved):

PR #2198, opened 2025-02-17
Modify SSL to inherit ciphersuites from SSL_CTX at initialization

smittals2:
### Description of changes: 
This is a follow up to https://github.com/aws/aws-lc/commit/154f9986bc18f8f0e6126157a3abc84230456f77. 
These changes initialize the SSL object with configured ciphersuites from the parent SSL_CTX object at initialization. Now both SSL and SSL_CTX objects will have their own copy of ciphersuites. 

We inherited the behavior of defaulting to SSL_CTX ciphersuites when none are set on SSL from BoringSSL. OpenSSL does things differently and copies the ciphersuites to SSL at init. This behavioral difference is aligned in this PR. 

### Testing:
A test to ensure changes to SSL_CTX don't impact SSL configurations and we correctly inherit ciphersuites. 

### DIFF
--- a/ssl/ssl_lib.cc
+++ b/ssl/ssl_lib.cc
@@ -2172,8 +2172,7 @@ STACK_OF(SSL_CIPHER) *SSL_get_ciphers(const SSL *ssl) {
     return ssl->config->cipher_list->ciphers.get();
   }
 
-  return ssl->ctx && ssl->ctx->cipher_list ?
-         ssl->ctx->cipher_list->ciphers.get() : NULL;
+  return ssl->ctx->cipher_list->ciphers.get();
 }
 
 const char *SSL_get_cipher_list(const SSL *ssl, int n) {
--- a/ssl/ssl_lib.cc
+++ b/ssl/ssl_lib.cc
@@ -693,6 +693,15 @@ SSL *SSL_new(SSL_CTX *ctx) {
           ctx->alpn_client_proto_list) ||
       !ssl->config->verify_sigalgs.CopyFrom(ctx->verify_sigalgs)) {
     return nullptr;
+      }
+
+  if (ctx->cipher_list) {
+    ssl->config->cipher_list = MakeUnique<SSLCipherPreferenceList>();
+    ssl->config->cipher_list->Init(*ctx->cipher_list.get());
+  }
+  if (ctx->tls13_cipher_list) {
+    ssl->config->tls13_cipher_list = MakeUnique<SSLCipherPreferenceList>();
+    ssl->config->tls13_cipher_list->Init(*ctx->tls13_cipher_list.get());
   }
 
   if (ctx->psk_identity_hint) {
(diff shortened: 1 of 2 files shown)


ssl/ssl_test.cc
```diff
@@ -2821,6 +2821,49 @@ TEST(SSLTest, FindingCipher) {
...
+  ASSERT_TRUE(CreateClientAndServer(&client, &server, client_ctx.get(), server_ctx.get()));
+
+  // Modify CTX ciphersuites
+  ASSERT_TRUE(SSL_CTX_set_ciphersuites(client_ctx.get(), "TLS_AES_256_GCM_SHA384"));
+  ASSERT_TRUE(SSL_CTX_set_ciphersuites(server_ctx.get(), "TLS_AES_256_GCM_SHA384"));
+  ASSERT_TRUE(SSL_CTX_set_cipher_list(client_ctx.get(), "TLS_RSA_WITH_AES_256_CBC_SHA"));
+  ASSERT_TRUE(SSL_CTX_set_cipher_list(server_ctx.get(), "TLS_RSA_WITH_AES_256_CBC_SHA"));
+
+  // Ensure SSL object inherits initial CTX ciphersuites
+  STACK_OF(SSL_CIPHER) *client_ciphers = SSL_get_ciphers(client.get());
```
torben-hansen, line 2852: could check has length 2.

torben-hansen (dismissed):

torben-hansen (approved):

ssl/ssl_lib.cc
```diff
@@ -2158,11 +2167,9 @@ STACK_OF(SSL_CIPHER) *SSL_get_ciphers(const SSL *ssl) {
   if (ssl == NULL) {
     return NULL;
   }
-  if (ssl->config && ssl->config->cipher_list) {
-    return ssl->config->cipher_list->ciphers.get();
-  }
 
-  return ssl->ctx->cipher_list->ciphers.get();
+  return ssl->config && ssl->config->cipher_list ?
+         ssl->config->cipher_list->ciphers.get() : NULL;
```
andrewhop, line 2172: Is this even possible now? Also what happens in libssl if a ctx and ssl config have no ciphersuites configured?

ssl/ssl_lib.cc
```diff
@@ -695,6 +695,15 @@ SSL *SSL_new(SSL_CTX *ctx) {
...
   }
 
+  if (ctx->cipher_list) {
+    ssl->config->cipher_list = MakeUnique<SSLCipherPreferenceList>();
+    ssl->config->cipher_list->Init(*ctx->cipher_list.get());
+  }
+  if (ctx->tls13_cipher_list) {
+    ssl->config->tls13_cipher_list = MakeUnique<SSLCipherPreferenceList>();
+    ssl->config->tls13_cipher_list->Init(*ctx->tls13_cipher_list.get());
+  }
```
andrewhop, line 705: Double checking: `ctx->cipher_list` will have been constructed in a valid way that ensures cipher_list includes all the 1.3 ciphers as well right?

smittals2: yep, ctx->cipher_list will always have all configured ciphersuites.

ssl/ssl_lib.cc
smittals2: if a user explicitly sets a client or server SSL/CTX object to hold no ciphersuites (empty), you will get a NO_SHARED_CIPHER error during a connection. 

The config object may be shed after a connection if not explicitly kept. So still good practice to check if ssl->config exists.

torben-hansen (dismissed):

ssl/ssl_lib.cc
andrewhop: Can you add a test for that after whatever triggers the ssl config to shed it's copy of the ciphers.

andrewhop (approved):

PR #2200, opened 2025-02-18
Refactor AWS_LC_FIPS_failure to always exist

andrewhop:
### Description of changes: 
This change does two things:
1. Update bcm.c to always build a AWS_LC_FIPS_failure function that either prints the message and aborts or just prints the message
2. Update the remaining spots in self_check.c to always use that function and not print directly to stderr

### Call-outs:
This does not change any of the expected output from AWS-LC.

### Testing:
The existing tests that check the FIPS failure messages ensure there is no change in behavior with this change

### DIFF
--- a/crypto/fipsmodule/bcm.c
+++ b/crypto/fipsmodule/bcm.c
@@ -270,8 +270,7 @@ static void BORINGSSL_bcm_power_on_self_test(void) {
 
 #if defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
   if (jent_entropy_init()) {
-    fprintf(stderr, "CPU Jitter entropy RNG initialization failed.\n");
-    AWS_LC_FIPS_failure("CPU Jitter failed to initilize");
+    AWS_LC_FIPS_failure("CPU Jitter entropy RNG initialization failed");
   }
 #endif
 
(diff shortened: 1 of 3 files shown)


crypto/fipsmodule/bcm.c
```diff
@@ -270,8 +270,7 @@ static void BORINGSSL_bcm_power_on_self_test(void) {
 
 #if defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
   if (jent_entropy_init()) {
-    fprintf(stderr, "CPU Jitter entropy RNG initialization failed.\n");
-    AWS_LC_FIPS_failure("CPU Jitter failed to initilize");
+    AWS_LC_FIPS_failure("CPU Jitter entropy RNG initialization failed");
```
skmcgrail, line 273: > This does not change any of the expected output from AWS-LC.
Doesn't this change the output? Or was this an error from the earlier refactor?

andrewhop: Correct to both:
1. The message printing out the error to stderr "CPU Jitter entropy RNG initialization failed" has been there since we added jitter 
2. In https://github.com/aws/aws-lc/commit/5553a20c813aa0b98ff7f26af1822633eccbca20#diff-eb881f93cebfbb804479d55b9d097763dbb1d6e85da9d3d9bfbbf2d9cb1db04d I passed this new slightly different message to AWS_LC_FIPS_failure

skmcgrail (approved):

torben-hansen (approved):

PR #2201, opened 2025-02-18
Update pairwise consistency test failures to support gracefully continiung

andrewhop:
### Description of changes: 
Currently AWS_LC_FIPS_failure always aborts the process and it is impossible to continue after calling that function. In a future change this will be optional and AWS-LC will not abort and the overall API call should return a failure. This change updates the current pairwise consistency test spots to theoretically return a failure.

### Call-outs:
This PR does not change the behavior, if there is a PWCT failure the process still aborts.

### Testing:
With the current AWS_LC_FIPS_failure implementation it is impossible to test this change.

### DIFF
--- a/crypto/fipsmodule/ec/ec_key.c
+++ b/crypto/fipsmodule/ec/ec_key.c
@@ -550,9 +550,8 @@ int EC_KEY_generate_key_fips(EC_KEY *eckey) {
 
 #if defined(AWSLC_FIPS)
   AWS_LC_FIPS_failure("EC keygen checks failed");
-#else
-  return 0;
 #endif
+  return 0;
 }
 
 int EC_KEY_get_ex_new_index(long argl, void *argp, CRYPTO_EX_unused *unused,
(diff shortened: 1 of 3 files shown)


crypto/fipsmodule/ec/ec_key.c
```diff
@@ -550,9 +550,8 @@ int EC_KEY_generate_key_fips(EC_KEY *eckey) {
 
 #if defined(AWSLC_FIPS)
   AWS_LC_FIPS_failure("EC keygen checks failed");
-#else
-  return 0;
 #endif
+  return 0;
```
skmcgrail, line 554: Does this affect any unreachable-code analysis? As the `return 0;` could be seen as unreachable due to `AWS_LC_FIPS_failure` trigger an abort?

andrewhop: In theory yes, but nothing seems to be catching the issue. Looks like GCC removed support for the check a while ago https://gcc.gnu.org/bugzilla/show_bug.cgi?id=82100#c1, Clang has an option but we don't have it turned on https://clang.llvm.org/docs/DiagnosticsReference.html#wunreachable-code. 

I enabled that option for clang and it correctly flags the `exit(1)` is unreachable after the abort. But it doesn't catch this new unreachable code. Reading the bug reports it looks like detecting unreachable code is a hard problem and very dependent on compiler versions and optimizations.

Debug build:
```
bcm.c:398:5: error: code will never be executed [-Werror,-Wunreachable-code]
  398 |     exit(1);
      |     ^~~~
```

Release build didn't find anything else with Clang.

andrewhop:
https://github.com/aws/aws-lc/pull/2195 took care of the ML-KEM change, the lower function returns the error from the PCT and in the event that the AWS_LC_FIPS_failure returns it returns the failure result to the caller.

skmcgrail (approved):

PR #2208, opened 2025-02-19
Use AWSLC_SOURCE_DIR and AWSLC_BINARY_DIR

justsmth:
### Description of changes: 
* Use slightly less ambiguous [AWSLC_SOURCE_DIR](https://cmake.org/cmake/help/v3.0/variable/PROJECT-NAME_SOURCE_DIR.html) and [AWSLC_BINARY_DIR](https://cmake.org/cmake/help/v3.0/variable/PROJECT-NAME_BINARY_DIR.html) instead of [PROJECT_SOURCE_DIR](https://cmake.org/cmake/help/v3.0/variable/PROJECT_SOURCE_DIR.html#variable:PROJECT_SOURCE_DIR) and [PROJECT_BINARY_DIR](https://cmake.org/cmake/help/v3.0/variable/PROJECT_BINARY_DIR.html).

### Call-outs:
* No change to logic. The change is a simple text substitution.


andrewhop (approved):
Was this causing an issue? Is it worth putting some pre-push check that ensures there are no PROJECT variables in our CMakeList?

justsmth:
> Was this causing an issue? Is it worth putting some pre-push check that ensures there are no PROJECT variables in our CMakeList?

I don't think it was causing any actual problems for us. I noticed it while debugging a build issue and I got concerned about which "project" was in context. The use of `PROJECT_SOURCE_DIR` is a potential source of ambiguity when working with third-party libraries (like jitterentropy).

skmcgrail (approved):

PR #2230, opened 2025-02-28
Rfc5280 4.2.1.9

skmcgrail:
### Description of changes: 
Improves validation of certificates by also applying additional checks for certificates retrieved from the trust store. This aligns AWS-LC with OpenSSL 1.1.1 and later behavior.

### Testing:
Add a root certificate that does not have the expected basicConstraints extension with the cA bit set to true. This certificate would be rejected by OpenSSL, but is currently allowed to be trusted by AWS-LC due to it being in the trust store.


crypto/x509/x509_vfy.c
```diff
@@ -568,8 +568,9 @@ static int check_chain_extensions(X509_STORE_CTX *ctx) {
   int ok = 0, plen = 0;
   int purpose = ctx->param->purpose;
 
-  // Check all untrusted certificates
-  for (int i = 0; i < ctx->last_untrusted; i++) {
+  int num = (int)sk_X509_num(ctx->chain);
+
+  for (int i = 0; i < num; i++) {
```
samuel40791765, line 573: Just to verify, this is essentially making our certificate checks slightly stricter but aligning with 1.1.1?

samuel40791765 (approved):

crypto/x509/x509_vfy.c
skmcgrail: Correct you can see the equivalent line here: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/x509/x509_vfy.c#L485 where `num` is set to `sk_X509_num(ctx->chain)` on line 462.

justsmth (approved):

PR #2231, opened 2025-02-28
Add IbmTpm to our CI

smittals2:
### Issues:
`CryptoAlg-2857`

### Description of changes: 
Add IbmTpm to our CI. No tests found in the library.

### Call-outs:

### Testing:


tests/ci/integration/ibmtpm_patch/ibmtpm-mainline-awslc.patch
```diff
@@ -0,0 +1,122 @@
...
++++ b/src/BnToOsslMath.c
+@@ -86,6 +86,12 @@
+ #ifdef MATH_LIB_OSSL
+ #  include "BnToOsslMath_fp.h"
+
++#ifdef OPENSSL_IS_AWSLC
++    #define BN_FIELD_SIZE(a) ((a)->width)
++#else
++    #define BN_FIELD_SIZE(a) ((a)->top)
++#endif
```
andrewhop, line 13: We should not add a dependency to the internals of one of our structs. Please add a proper getter method for this.

tests/ci/integration/ibmtpm_patch/ibmtpm-mainline-awslc.patch
```diff
@@ -0,0 +1,122 @@
...
+
+ //*** OsslToTpmBn()
+@@ -105,10 +111,10 @@ BOOL OsslToTpmBn(bigNum bn, BIGNUM* osslBn)
+ 	{
+ 	    int i;
+ 	    //
+-	    GOTO_ERROR_UNLESS((unsigned)osslBn->top <= BnGetAllocated(bn));
+-	    for(i = 0; i < osslBn->top; i++)
++	    GOTO_ERROR_UNLESS((unsigned)BN_FIELD_SIZE(osslBn) <= BnGetAllocated(bn));
++	    for(i = 0; i < BN_FIELD_SIZE(osslBn); i++)
```
andrewhop, line 25: Oh I see they were already using the internals from OpenSSL. Instead of continuing this pattern can this use a common function between OpenSSL and AWS-LC? What is this code doing?

tests/ci/integration/ibmtpm_patch/ibmtpm-mainline-awslc.patch
```diff
@@ -0,0 +1,122 @@
...
+ {
+diff --git a/src/TpmProfile_Common.h b/src/TpmProfile_Common.h
+index 13dfc0d..3c87c3e 100644
+--- a/src/TpmProfile_Common.h
++++ b/src/TpmProfile_Common.h
+@@ -65,6 +65,8 @@
+ // Requires basic YES/NO defines are already set (by TpmBuildSwitches.h)
+ // Less frequently changed items are in other TpmProfile Headers.
+
++#include <openssl/base.h>
```
andrewhop, line 107: This is an AWS-LC specific header, if you just need access to OPENSSL_NO_CAMELLIA this should use opensslconf.h which is common with OpenSSL.

tests/ci/integration/run_ibmtpm_integration.sh
```diff
@@ -0,0 +1,71 @@
...
+rm -rf "${SCRATCH_FOLDER:?}"/*
+cd ${SCRATCH_FOLDER}
+
+function ibmtpm_build() {
+  # Check which AWS-LC library folder name we must use.
+  if [ -d ${AWS_LC_INSTALL_FOLDER}/lib64 ]; then
+    AWS_LC_LIBRARY_FOLDER="lib64"
+  else
+    AWS_LC_LIBRARY_FOLDER="lib"
+  fi
```
andrewhop, line 37: You can also force CMake to be consistent with CMAKE_INSTALL_LIBDIR.

tests/ci/integration/run_ibmtpm_integration.sh
```diff
@@ -0,0 +1,71 @@
...
+cd ${IBMTPM_SRC_FOLDER}
+mkdir -p ${AWS_LC_BUILD_FOLDER} ${AWS_LC_INSTALL_FOLDER}
+ls
+
+aws_lc_build "$SRC_ROOT" "$AWS_LC_BUILD_FOLDER" "$AWS_LC_INSTALL_FOLDER" -DBUILD_TESTING=OFF -DBUILD_TOOL=OFF -DCMAKE_BUILD_TYPE=Debug -DBUILD_SHARED_LIBS=1
+
+# Build ibmtpm from source.
+pushd ${IBMTPM_SRC_FOLDER}
+ibmtpm_patch_build
+ibmtpm_build
```
andrewhop, line 70: Are there any easy tests we can run?

smittals2: They have no tests in the IbmTpm repo

tests/ci/integration/ibmtpm_patch/ibmtpm-mainline-awslc.patch
smittals2: i don't think we define getters for the fields of a bignum struct. This seems to be how it's accessed in other libraries too: https://github.com/lwakefield/libgpucrypto/blob/master/rsa_cuda.cu#L196C1-L197C1

tests/ci/integration/run_ibmtpm_integration.sh
justsmth: 🔥

justsmth: It looks like there are some tests here: https://github.com/kgoldman/ibmswtpm2/blob/master/src/AlgorithmTests.c

I'm not sure how they should be invoked.

justsmth (dismissed):

tests/ci/integration/run_ibmtpm_integration.sh
smittals2: For tracking discussed offline: 
We need the tss application, a client (provided via tpm2-tools), and other kernel config to ensure TPM support is enabled in order to run the self tests on this application. Currently not running the tests.

tests/ci/integration/ibmtpm_patch/ibmtpm-mainline-awslc.patch
andrewhop: Two things:
1. Just because that's how other people are doing it doesn't make it right
2. Is OpenSSL top the same as AWS-LC width?
OpenSSL:
```
    BN_ULONG *d;                /*
                                 * Pointer to an array of 'BN_BITS2' bit
                                 * chunks. These chunks are organised in
                                 * a least significant chunk first order.
                                 */
    int top;                    /* Index of last used d +1. */
 ```

AWS-LC's width is different:
```
  // d is a pointer to an array of |width| |BN_BITS2|-bit chunks in
  // little-endian order. This stores the absolute value of the number.
  BN_ULONG *d;
  // width is the number of elements of |d| which are valid. This value is not
  // necessarily minimal; the most-significant words of |d| may be zero.
  // |width| determines a potentially loose upper-bound on the absolute value
  // of the |BIGNUM|.
  //
  // Functions taking |BIGNUM| inputs must compute the same answer for all
  // possible widths. |bn_minimal_width|, |bn_set_minimal_width|, and other
  // helpers may be used to recover the minimal width, provided it is not
  // secret. If it is secret, use a different algorithm. Functions may output
  // minimal or non-minimal |BIGNUM|s depending on secrecy requirements, but
  // those which cause widths to unboundedly grow beyond the minimal value
  // should be documented such.
  //
  // Note this is different from historical |BIGNUM| semantics.
  int width;
```

OpenSSL d could point to 100 BN_BITS, 50 are used, top would be 51.
AWS-LC d could point to 100 BN_ULONG, 50 are used, width could be 50 up to 100 padded with zero.

Can this code handle this difference in behavior?

andrewhop: Can this code use BN_num_bytes?

smittals2: Good point, I believe the equivalent per this documentation and code is bn_minimal_width. This will give us the index after the last non-zero chunk - which is what top is holding.

justsmth (approved):

andrewhop (approved):

PR #2244, opened 2025-03-05
Add break-kat for ml-dsa

jakemas:
### Issues:
Resolves #PQCrypto-130

### Description of changes: 

- Added ML-DSA to the break-kat tool.
- The related PWCT test https://github.com/aws/aws-lc/pull/2192 had to be modified such that:
- - it now calls the `_internal` variants of `sign/verify`. This is so that per-message randomness `rnd` can be controlled. We need this signature to be deterministic, so we can set the seed ourselves.
- - it now signs the same message as expected in `break-kat`. Originally, the PCT message was the all zero string, however, `break-kat.go` works by inserting an all zero string in place of the message to corrupt it, thus, the message must be non-zero to start with. Without setting the per message randomness, this signature will alter each time.

Related break-kat.go change for Ed25519/ML-KEM https://github.com/aws/aws-lc/pull/2173

### Call-outs:
As the break kat test modifies input to switch it all out for zeros, I had to change the `kMLDSASignPlaintext` in `self_check.c` to no longer be all zeros. As such the expected signature also has changed.

### Testing:
Running break-Kat.


jakemas:
https://github.com/aws/aws-lc/pull/2253

PR #2245, opened 2025-03-05
Add public wrapper to internal bn_minimal_width function

smittals2:
### Description of changes: 
This PR adds a public wrapper for an existing internal function bn_minimal_width. OpenSSL stores the minimal number of words needed in a field called top within the bignum struct. AWS-LC does not do this, instead we have a field called width which is a loose upper bound on this value. The new function, BN_get_minimal_width will provide us with the value of "top" for a given Bignum. 

### DIFF
--- a/include/openssl/bn.h
+++ b/include/openssl/bn.h
@@ -222,6 +222,10 @@ OPENSSL_EXPORT unsigned BN_num_bits(const BIGNUM *bn);
 // will always fit in |int|.
 OPENSSL_EXPORT unsigned BN_num_bytes(const BIGNUM *bn);
 
+// BN_get_minimal_width returns the minimal number of words needed to represent
+// |bn|.
+OPENSSL_EXPORT int BN_get_minimal_width(const BIGNUM *bn);
+
 // BN_zero sets |bn| to zero.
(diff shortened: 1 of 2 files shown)


torben-hansen (approved):

include/openssl/bn.h
```diff
@@ -222,6 +222,10 @@ OPENSSL_EXPORT unsigned BN_num_bits(const BIGNUM *bn);
 // will always fit in |int|.
 OPENSSL_EXPORT unsigned BN_num_bytes(const BIGNUM *bn);
 
+// BN_get_minimal_width returns the minimal number of words needed to represent
+// |bn|.
+OPENSSL_EXPORT int BN_get_minimal_width(const BIGNUM *bn);
```
andrewhop, line 227: Is there an equivalent OpenSSL function that we can rename ours to so customers get a consistent experience?

smittals2: Nothing in the public API, closest thing they have in their internal functions is a function called bn_check_top but that validates the value of top, doesn't return it. 

I named this according to the internal function name. If it makes more sense, happy to change this to BN_get_words or BN_get_top or something else

andrewhop (approved):

PR #2246, opened 2025-03-05
Make BIO_get_mem_data a function again

skmcgrail:
### Description of changes: 
Can't track it down the historical reason why, but we have `BIO_get_mem_data` defined as a preprocessor macro, but BoringSSL does not. This creates an annoying compatibility issue for rust-openssl, and has forced us to implement the function properly in our aws-lc-sys crates in Rust. This redefines the function to have a concrete definition.

As a side-note, the macro was technically wrong as it was casting contents to a `char *` when it was actually `char **`, but this wasn't problematic per-se since it was being passed into a `void *` argument. If you look at the ctrl function definition you can see that it is indeed treated as a `char **` pointer.

### DIFF
--- a/crypto/bio/bio_mem.c
+++ b/crypto/bio/bio_mem.c
@@ -296,6 +296,10 @@ int BIO_mem_contents(const BIO *bio, const uint8_t **out_contents,
   return 1;
 }
 
+long BIO_get_mem_data(BIO *bio, char **contents) {
+  return BIO_ctrl(bio, BIO_CTRL_INFO, 0, contents);
+}
+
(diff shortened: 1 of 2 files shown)


justsmth (approved):

samuel40791765 (approved):
Original Context: I was forced to change this to a macro because mySQL was using the function improperly. I submitted [a change](https://github.com/mysql/mysql-server/commit/285924c81d43615f84c754cca7612d9ee77e6c7d) to upstream mysql to fix this. They've accepted it and it's in versions after mysql-8.3.0: https://bugs.mysql.com/bug.php?id=112846
We've agreed to use the referenced commit as a patch for any possible consumers of MySQL versions earlier than 8.3.

PR #2247, opened 2025-03-06
Add x4 batched SHAKE128 and SHAKE256 APIs

manastasova:
**_mlkem-native_** uses batched SHAKE to compute different XOF digests _simultaneously_ I. e., every call to batched SHAKE function performs four independent algorithm execution (x4). 

This PR introduces the required new shake SHAKE x4 APIs. Internally, they call 4 independent SHAKE functions (either the entire algorithm or a part of it, such as init, absorb, squeeze).

### Issues:
Resolves #CryptoAlg-2959

### Description of changes: 
 - SHAKE128_Init_x4 
 - SHAKE128_Absorb_once_x4
 - SHAKE128_Squeezeblocks_x4
 - SHAKE256_x4

### Call-outs:
Currently, the APIs are named as mlkem-specific functions since there is no other use-case for them (absorbing once, squeezing entire blocks). This naming will allow to implement them optimally later on, skipping some unnecessary checks. However, since they are internally calling the existing aws-lc shake functions, they inherit all functionalities supported by the aws-lc shake functions. (incremental absorbs, incremental byte-wise squeezes).

The aws-lc integration of the x4 batched SHAKE APIs will allow to comply with the current mkem-native design. These functions will not allow a SIMD assembly keccak implementation to be added to aws-lc without further changes. However, they set the batched APIs and could be inlined in mlkem-native code.

### Testing:
New SHAKETest_x4 test: 
            ```./crypto/crypto_test --gtest_filter="SHAKETest_x4.RandomMessages"``` 


hanno-becker (changes requested):
Thank you @manastasova!

Could we please document the failure conditions for those functions? The ML-KEM call-site does not expect errors from the SHAKE API and hence needs to argue why none of the failure conditions may occurr.

(This is also an issue with the pre-existing x1 API for which no documentation of failure conditions is given, but the description may be easier for the non-incremental case).

Also, do we need squeeze to accept an arbitrary length, or can it be restricted to multiples of the block size? The current API forces the internal maintenance of an offset pointer, which would not be necessary if only full blocks were squeezed.

crypto/fipsmodule/sha/sha3.c
```diff
@@ -377,3 +377,48 @@ int SHAKE_Squeeze(uint8_t *md, KECCAK1600_CTX *ctx, size_t len) {
   //FIPS_service_indicator_update_state();
   return 1;
 }
+
+int SHAKE128_Init_x4(KECCAK1600_CTX_x4 *ctx) {
+
+  int ok = (SHAKE_Init(&(*ctx)[0], SHAKE128_BLOCKSIZE) &&
```
hanno-becker, line 383: If it is intentional that later calls are skipped if earlier calls fail, this may be worth documenting. Same for the othe functions.

crypto/fipsmodule/sha/sha3.c
```diff
@@ -377,3 +377,48 @@ int SHAKE_Squeeze(uint8_t *md, KECCAK1600_CTX *ctx, size_t len) {
   //FIPS_service_indicator_update_state();
   return 1;
 }
+
+int SHAKE128_Init_x4(KECCAK1600_CTX_x4 *ctx) {
```
hanno-becker, line 381: Can this function actually fail? If so, when?

crypto/fipsmodule/sha/sha3.c
```diff
@@ -377,3 +377,48 @@ int SHAKE_Squeeze(uint8_t *md, KECCAK1600_CTX *ctx, size_t len) {
...
+
+  int ok = (SHAKE_Init(&(*ctx)[0], SHAKE128_BLOCKSIZE) &&
+            SHAKE_Init(&(*ctx)[1], SHAKE128_BLOCKSIZE) &&
+            SHAKE_Init(&(*ctx)[2], SHAKE128_BLOCKSIZE) &&
+            SHAKE_Init(&(*ctx)[3], SHAKE128_BLOCKSIZE));
+
+  return ok;
+}
+
+int SHAKE128_Absorb_once_x4(KECCAK1600_CTX_x4 *ctx, const void *data0, const void *data1,
```
hanno-becker, line 391: Under what condition would this function fail?

crypto/fipsmodule/sha/sha3.c
```diff
@@ -377,3 +377,48 @@ int SHAKE_Squeeze(uint8_t *md, KECCAK1600_CTX *ctx, size_t len) {
...
+
+  int ok = (SHAKE_Absorb(&(*ctx)[0], data0, len) &&
+            SHAKE_Absorb(&(*ctx)[1], data1, len) &&
+            SHAKE_Absorb(&(*ctx)[2], data2, len) &&
+            SHAKE_Absorb(&(*ctx)[3], data3, len));
+
+  return ok;
+}
+
+int SHAKE128_Squeezeblocks_x4(uint8_t *md0, uint8_t *md1, uint8_t *md2, uint8_t *md3,
```
hanno-becker, line 402: Do we need the generality of an arbitrary `len`, or can we stick to a multiple of the block size for now? I seem to remember that ML-KEM only squeezes a block a time.

hanno-becker: If the generality of an arbitary length is needed, the function should not be called `Squeezeblocks`.

crypto/fipsmodule/sha/sha3.c
manastasova: I added a comment https://github.com/aws/aws-lc/blob/4fca7ae7122f5f6d0c15625f90a3211e54f76772/crypto/fipsmodule/sha/internal.h#L464.

crypto/fipsmodule/sha/sha3.c
manastasova: Commented here https://github.com/aws/aws-lc/blob/4fca7ae7122f5f6d0c15625f90a3211e54f76772/crypto/fipsmodule/sha/internal.h#L473

crypto/fipsmodule/sha/sha3.c
manastasova: SHAKE128_Squeezeblocks_x4 aims at defining the new API needed by mlkem-native. Thus, it is defined as mlkem-specific function (only allows squeezes of entire blocks), while _currently_ internally calling existing SHAKE sequential APIs (to split changes into incremental updates).
That is my understanding based on our previous discussions.
The same reasoning is applied to the SHAKE128_Absorb_once_x4 mlkem-specific API which inherits the generic functionality from the underlying existing SHAKE functions.

hanno-becker: My point is that the current implementation does not squeeze a number of blocks, but `len` is used as the number of bytes?

hanno-becker: I think you want something like (but 4-fold) https://github.com/aws/aws-lc/blob/09ad19eb3290157c666953a9d455efd0fed3fced/crypto/fipsmodule/ml_kem/fips202_glue.h#L36-L41

crypto/fipsmodule/sha/sha3_test.cc
```diff
@@ -496,6 +501,62 @@ TEST(KeccakInternalTest, SqueezeOutputBufferOverflow) {
...
+      ASSERT_TRUE(SHAKE128(random_in[j], RAND_BYTES_x4, digest[j],
+                                              RAND_OUT_BYTES_BLCKS * SHAKE128_BLOCKSIZE));
+    }
+
+    // Compute one batched x4 SHAKE128
+    ASSERT_TRUE(SHAKE128_Init_x4(&ctx));
+    ASSERT_TRUE(SHAKE128_Absorb_once_x4(&ctx, random_in[0], random_in[1], random_in[2], random_in[3],
+                                                                                          RAND_BYTES_x4));
+    ASSERT_TRUE(SHAKE128_Squeezeblocks_x4(digest_x4[0], digest_x4[1], digest_x4[2], digest_x4[3],
+                                                        &ctx, RAND_OUT_BYTES_BLCKS * SHAKE128_BLOCKSIZE));
```
hanno-becker, line 530: `Squeezeblocks` should take the number of blocks, not the number of bytes, and the multiplication with `SHAKE128_BLOCKSIZE` to get the number of bytes should be in the implementation.

crypto/fipsmodule/sha/sha3.c
hanno-becker: If the function always succeeds, can we drop the return code?

crypto/fipsmodule/sha/sha3.c
hanno-becker: If the function always succeeds, can we drop the return code?

crypto/fipsmodule/sha/sha3.c
manastasova: Thank you! I just updated the code.

crypto/fipsmodule/sha/sha3_test.cc
manastasova: Thank you!

crypto/fipsmodule/sha/sha3.c
manastasova: I think it is better to keep it. It is, indeed, a function designed specifically for mlkem, nevertheless, nothing prevents from calling it in different use case scenario (if some). Wouldn't it be better to document the _skipped_ return code check in mlkem-native, and argument it with the input values?

crypto/fipsmodule/sha/sha3.c
manastasova: Same as above: I think it is better to keep it. It is, indeed, a function designed specifically for mlkem, nevertheless, nothing prevents from calling it in different use case scenario (if some). Wouldn't it be better to document the _skipped_ return code check in mlkem-native, and argument it with the input values?

crypto/fipsmodule/sha/sha3.c
```diff
@@ -110,12 +110,16 @@ uint8_t *SHAKE256(const uint8_t *data, const size_t in_len, uint8_t *out, size_t
 }
 
 // FIPS202 APIs manage internal input/output buffer on top of Keccak1600 API layer
+// FIPS202_Reset zero's |ctx| fields.
```
justsmth, line 113: NP: It seems that the "FIPS202 APIs manage ..." comment might be better placed inside a multi-line `/* ... */` comment as it relates to several functions below.

justsmth (dismissed):

crypto/fipsmodule/sha/sha3.c
manastasova: Thanks! Just updated.

crypto/fipsmodule/sha/sha3.c
```diff
@@ -364,16 +381,64 @@ int SHAKE_Squeeze(uint8_t *md, KECCAK1600_CTX *ctx, size_t len) {
...
+
+/*
+ * SHAKE batched (x4) APIs implement SHAKE functionalities in batches of four on top of SHAKE API layer
+ */
+int SHAKE128_Init_x4(KECCAK1600_CTX_x4 *ctx) {
+
+  int ok = (SHAKE_Init(&(*ctx)[0], SHAKE128_BLOCKSIZE) &&
+            SHAKE_Init(&(*ctx)[1], SHAKE128_BLOCKSIZE) &&
+            SHAKE_Init(&(*ctx)[2], SHAKE128_BLOCKSIZE) &&
+            SHAKE_Init(&(*ctx)[3], SHAKE128_BLOCKSIZE));
```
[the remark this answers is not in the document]
manastasova: Since `ctx` is the pointer to (not actually the) array of structs:
`typedef KECCAK1600_CTX KECCAK1600_CTX_x4[4];`
...
`KECCAK1600_CTX_x4 ctx;`
`ASSERT_TRUE(SHAKE128_Init_x4(**&ctx**));`
We need to dereference to get back to the array of stucts, then to access some index, e.g., struct [3]; later we need the address of that struct for the next function.
`int SHAKE_Init(KECCAK1600_CTX *ctx, size_t block_size);`

crypto/fipsmodule/sha/sha3_test.cc
```diff
@@ -496,6 +501,62 @@ TEST(KeccakInternalTest, SqueezeOutputBufferOverflow) {
...
+    // Compute one batched x4 SHAKE128
+    ASSERT_TRUE(SHAKE128_Init_x4(&ctx));
+    ASSERT_TRUE(SHAKE128_Absorb_once_x4(&ctx, random_in[0], random_in[1], random_in[2], random_in[3],
+                                                                                          RAND_BYTES_x4));
+    ASSERT_TRUE(SHAKE128_Squeezeblocks_x4(digest_x4[0], digest_x4[1], digest_x4[2], digest_x4[3],
+                                                        &ctx, RAND_OUT_BLCKS));
+
+    for (int j = 0; j < BATCHED_x4; j++) {
+      EXPECT_EQ(EncodeHex(bssl::MakeConstSpan(digest_x4[j], RAND_OUT_BLCKS * SHAKE128_BLOCKSIZE)),
+                EncodeHex(bssl::MakeConstSpan(digest[j], RAND_OUT_BLCKS * SHAKE128_BLOCKSIZE)));
```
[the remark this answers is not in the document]
manastasova: Thank you!

crypto/fipsmodule/sha/sha3.c
hanno-becker: > I think it is better to keep it. It is, indeed, a function designed specifically for mlkem, nevertheless, nothing prevents from calling it in different use case scenario (if some). 

Yes, but my understanding is that this function will never fail, regardless of who calls it?

crypto/fipsmodule/sha/sha3.c
hanno-becker: Yes, but this is not a question about the caller/use case?

crypto/fipsmodule/sha/internal.h
```diff
@@ -406,39 +409,91 @@ OPENSSL_EXPORT uint8_t *SHAKE128(const uint8_t *data, const size_t in_len,
 OPENSSL_EXPORT uint8_t *SHAKE256(const uint8_t *data, const size_t in_len,
                                  uint8_t *out, size_t out_len);
 
-// SHA3_Init initialises |ctx| fields through |FIPS202_Init| and 
-// returns 1 on success and 0 on failure.
+// SHA3_Init initialises |ctx| fields through |FIPS202_Init| and
+// returns 1 on success and 0 on failure. SHA3_Init fails if
+// |ctx| is nullptr or |bitlen| is inproper for SHA3 (any security
+// level) block size length. Otherwise, it propagates the |FIPS202_Init|
```
hanno-becker, line 415: As a caller of this function, the details of which functions are called underneath are not important -- but the caller needs to understand failure conditions. Here, one could recursively lookup the documentation of `FIPS202_Update` to understand its failure conditions, but it's a bit cumbersome. Is there a way one can express success/failure conditions at this level without resorting to implementation details?

Similar questions below.

crypto/fipsmodule/sha/internal.h
```diff
@@ -406,39 +409,91 @@ OPENSSL_EXPORT uint8_t *SHAKE128(const uint8_t *data, const size_t in_len,
...
+// It fails on the first |SHAKE_Init| function fail, however, if called with
+// valid inputs, this function never fails.
+OPENSSL_EXPORT int SHAKE128_Init_x4(KECCAK1600_CTX_x4 *ctx);
+
+// SHAKE128_Absorb_once_x4 is a batched API that operates on four independent
+// Keccak bitstates. It absorbs all four inputs |data0|, |data1|, |data2|, |data3|
+// of equal length of |len| bytes through four consecutive calls to |SHAKE_Absorb|
+// and returns 1 on success and 0 on failure. SHAKE128_Absorb_once_x4 succeeds when
+// all four |SHAKE_Absorb| functions succeed. It fails on the first |SHAKE_Absorb|
+// function fail, however, if called with valid inputs, this function never fails.
```
hanno-becker, line 472: It is unclear what 'valid' means here. You want the input to come fresh out of `SHAKE128_Init_x4`, or?

hanno-becker (changes requested):
Thank you @manastasova for the further work!

I think there is still need for improvement in the documentation: We should not re-state implementation details of which lower level functions are called (this will change), but provide a high-level contract to the caller indicating when to expect success/failure. This may not always be possible, but in this instance, I think it may.

Concretely, could one say something like: SHAKE context must go through the flow (a) Init, (b) Absorb [many times], (c) Squeeze [many times], (d) Final. If this call-discipline is maintained and the pointers passed to the functions are valid and of respective sizes, there will be no error. 

Similarly, for SHAKEx4, it would be: If you go through the flow (a) Init, (b) Absorb_once [once], (c) Squeezeblocks [many times], (d) Final -- and pointers are always valid, then there will be no error.

Can you clarify if the above is true? If so, I think this would be a good high-level documentation that callers such as ML-KEM/ML-DSA can reason against when ignoring return values, without being concerned with which lower level functions are being called, and what their failure conditions may be.

crypto/fipsmodule/sha/internal.h
manastasova: I updated the description https://github.com/aws/aws-lc/blob/026f32ed4a328356aca90712bd47ac459310fa97/crypto/fipsmodule/sha/internal.h#L422 following the suggested documentation format. Thank you.

crypto/fipsmodule/sha/internal.h
manastasova: 'Valid input' -> 'valid input pointer'. 
However, I updated this following the suggested documentation format. Thank you.

crypto/fipsmodule/sha/internal.h
```diff
@@ -406,39 +409,98 @@ OPENSSL_EXPORT uint8_t *SHAKE128(const uint8_t *data, const size_t in_len,
...
+// SHAKE128_Squeezeblocks_x4 is a batched API that operates on four independent Keccak
+// bitstates. It squeezes |blks| number of blocks for all four XOF digests through
+// four consecutive calls to |SHAKE_Squeeze| and returns 1 on success and 0 on failure.
+// SHAKE128_Squeezeblocks_x4 succeeds when all four |SHAKE_Squeeze| functions succeed.
+// It fails on the first |SHAKE_Squeeze| function fail.
+// As part of MLKEM PQ algorithm: SHAKE128_Squeezeblocks_x4 always returns 1 since it
+// is called with a valid |ctx|, |md0|, |md1|, |md2|, |md3| and with |blks| number of
+// blocks.
+OPENSSL_EXPORT int SHAKE128_Squeezeblocks_x4(uint8_t *md0, uint8_t *md1, uint8_t *md2, uint8_t *md3,
+                                  KECCAK1600_CTX_x4 *ctx, size_t blks);
```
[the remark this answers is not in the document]
andrewhop: Huh, that's fine.

crypto/fipsmodule/sha/sha3.c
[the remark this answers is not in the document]
andrewhop: Oh, that's subtle

andrewhop (approved):

crypto/fipsmodule/sha/internal.h
```diff
@@ -372,6 +375,14 @@ OPENSSL_EXPORT int SHA512_256_get_state(
     SHA512_CTX *ctx, uint8_t out_h[SHA512_256_CHAINING_LENGTH],
     uint64_t *out_n);
 
+/*
+ * SHA3/SHAKE single-shot APIs implement SHA3 functionalities on top
+ * of SHA3/SHAKE API layer
+ *
+ * SHA3/SHAKE single-shot functions never fail when the later call-discipline is
+ * adhered to: (a) the pointers passed to the functions are valid, is satisfied.
```
hanno-becker, line 383: The sentence structure seems off here

crypto/fipsmodule/sha/internal.h
```diff
@@ -405,40 +416,128 @@ OPENSSL_EXPORT uint8_t *SHAKE128(const uint8_t *data, const size_t in_len,
...
+ *
+ * SHAKE128_x4_ functions never fail when the later call-discipline is adhered to:
+ * (a) the context execution flow is followed (b) the pointers passed to the
+ * functions are valid (c) any additional per-function parameter value conditions,
+ * detailed above each SHAKE128_x4_ function signature, is satisfied.
+ */
+
+// SHAKE128_Init_x4 is a batched API that operates on four independent
+// Keccak bitstates. It initialises all four |ctx| fields through four
+// consecutive calls to |SHAKE_Init| and returns 1 on success and 0 on failure.
```
hanno-becker, line 500: Can we remove the mention of the implementation details here and elsewhere? Once we start implementing fast batched x4, we won't merely fall back to x1 anymore.

crypto/fipsmodule/sha/internal.h
manastasova: Thanks.

crypto/fipsmodule/sha/internal.h
manastasova: I tried to follow the general pattern for function description in the file/library. I was thinking once we start implementing it, we could update these comments.
I just fixed it. Thank you.

justsmth (approved):

andrewhop (approved):

PR #2256, opened 2025-03-07
Update EVP_PKEY ED keygen to use an internal function that can return the result of the PWCT

andrewhop:
### Description of changes: 
`ED25519_keypair` has always been void, this used to be okay because it was infallible. Now it can fail due to the PWCT failing. Usually if that happened the process would abort which is the default (and only approved) behavior of `AWS_LC_FIPS_failure`. However, if AWS-LC is built with the callback and that callback doesn't abort it is possible for `EVP_PKEY_keygen` to continue. Changes:
 * `EVP_PKEY_keygen` now detects this failure and return the error to the caller
 * `ED25519_keypair` doesn't increment the service indicator if the PWCT fails

### Testing:
I updated the fips_callback_test to use the EVP API for ED and check the return value

### DIFF
--- a/crypto/fipsmodule/curve25519/curve25519.c
+++ b/crypto/fipsmodule/curve25519/curve25519.c
@@ -32,6 +32,12 @@
 #include "../cpucap/internal.h"
 #include "internal.h"
 
+#if defined(NDEBUG)
+#define CHECK(x) (void) (x)
+#else
+#define CHECK(x) assert(x)
+#endif
+
 const uint8_t RFC8032_DOM2_PREFIX[DOM2_PREFIX_SIZE] = {
     'S', 'i', 'g', 'E', 'd', '2', '5', '5', '1', '9', ' ',
     'n', 'o', ' ', 'E', 'd', '2', '5', '5', '1', '9', ' ',
@@ -166,7 +172,7 @@ void ED25519_keypair(uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
   // The existing public function is void, ED25519_keypair_internal can only
   // fail if the PWCT fails and we're in a callback build where AWS_LC_FIPS_failure
   // doesn't abort on FIPS failure.
-  assert(ED25519_keypair_internal(out_public_key, out_private_key));
+  CHECK(ED25519_keypair_internal(out_public_key, out_private_key));
 }
 
 int ED25519_sign(uint8_t out_sig[ED25519_SIGNATURE_LEN],
--- a/crypto/fipsmodule/curve25519/curve25519.c
+++ b/crypto/fipsmodule/curve25519/curve25519.c
@@ -112,7 +112,7 @@ void ED25519_keypair_from_seed(uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
     ED25519_PUBLIC_KEY_LEN);
 }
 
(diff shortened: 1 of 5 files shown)


tests/ci/run_fips_tests.sh
```diff
@@ -35,7 +36,7 @@ if static_linux_supported || static_openbsd_supported; then
   run_build -DFIPS=1 \
     -DCMAKE_C_FLAGS="-DBORINGSSL_FIPS_BREAK_TESTS -DAWSLC_FIPS_FAILURE_CALLBACK" \
     -DCMAKE_CXX_FLAGS="-DAWSLC_FIPS_FAILURE_CALLBACK"
-  ./test_build_dir/crypto/crypto_test
+  shard_gtest ./test_build_dir/crypto/crypto_test
```
WillChilds-Klein, line 39: why is this needed?

crypto/fipsmodule/curve25519/curve25519.c
```diff
@@ -149,10 +151,21 @@ void ED25519_keypair(uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
...
+  }
+  return result;
+}
+
+void ED25519_keypair(uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
+  uint8_t out_private_key[ED25519_PRIVATE_KEY_LEN]) {
+  // The existing public function is void, ED25519_keypair_internal can only
+  // fail if the PWCT fails and we're in a callback build where AWS_LC_FIPS_failure
+  // doesn't abort on FIPS failure.
+  assert(ED25519_keypair_internal(out_public_key, out_private_key));
```
WillChilds-Klein, line 168: does AWS-LC's build enable `assert`?

tests/ci/run_fips_tests.sh
andrewhop: I just noticed it was slow and Torben reminded it exists. It could go in it's own review.

crypto/fipsmodule/curve25519/curve25519.c
andrewhop: For debug builds yeah, we use asserts all over the place.

tests/ci/run_fips_tests.sh
WillChilds-Klein: no, seems fine.

crypto/fipsmodule/curve25519/curve25519.c
```diff
@@ -149,10 +158,21 @@ void ED25519_keypair(uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
   ED25519_keypair_from_seed(out_public_key, out_private_key, seed);
   OPENSSL_cleanse(seed, ED25519_SEED_LEN);
 
-  ed25519_keypair_pct(out_public_key, out_private_key);
+  int result = ed25519_keypair_pct(out_public_key, out_private_key);
 
   FIPS_service_indicator_unlock_state();
-  FIPS_service_indicator_update_state();
+  if (result) {
+    FIPS_service_indicator_update_state();
```
WillChilds-Klein, line 165: > `ED25519_keypair` doesn't increment the service indicator if the PWCT fails

why not?

WillChilds-Klein (approved):

torben-hansen (approved):

PR #2258, opened 2025-03-08
Add the rehash utility to the openssl CLI tool

smittals2:
### Description of changes: 
Adding the rehash utility to the OpenSSL CLI tool. This design makes a number of different design choices from OpenSSL. Namely: 
1) Multiple directory paths are not supported, therefore duplicate filenames are not supported
2) Like OpenSSL, any existing symlinks matching the hash regex are removed from the given directory. We do not provide an option to preserve these.
3) We return a binary true or false to indicate status. OpenSSL counts and returns the number of errors. 
4) Simplified the general processing logic and hash table structures. 
5) Removed windows filesystem support, we only need this tool for unix based systems. 
6) We don't support the legacy hash format (MD5)

A wrapper bash script was created to call the openssl rehash functionality. This script is made executable and placed in the install folder via cmake. We also copy a test for this bash script to the build dir. 

### Testing:
Unit tests for the two tier hashtable and integration tests with symlink generation. Added another test to verify the functionality of the produced c_rehash script. 


crypto/test/test_util.h
```diff
@@ -116,6 +116,10 @@ size_t createTempFILEpath(char buffer[PATH_MAX]);
 FILE* createRawTempFILE();
 TempFILE createTempFILE();
 
+#ifndef OPENSSL_WINDOWS
+size_t createTempDIRpath(char buffer[PATH_MAX]);
+#endif
```
justsmth, line 121: Is there a reason we can't support Windows? I know it's a bit messy to create a temp directory on Windows, but I think we should support it as well.

tool-openssl/c_rehash.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+# Get absolute path to script directory
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
+
+# Check for openssl binary in same directory
+if [ -f "${SCRIPT_DIR}/openssl" ]; then
```
justsmth, line 9: This should check that the file is executable.

```
if [ -x "${SCRIPT_DIR}/openssl" ]; then
```

tool-openssl/c_rehash_test.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+# Get the directory where the script is located
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Run c_rehash with -help
+output=$("${SCRIPT_DIR}/c_rehash" -help 2>&1)
```
justsmth, line 9: Shouldn't this be `${SCRIPT_DIR}/c_rehash.sh`?

tool-openssl/c_rehash.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+# Get absolute path to script directory
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
```
justsmth, line 6: Unless there's reason not to, these bash scripts should start by ensuring that any failed commands also cause the script to fail:
```
set -ex
set -o pipefail
```

skmcgrail: Can combine them into one line as well `set -exo pipefail`

tool-openssl/rehash.cc
```diff
@@ -0,0 +1,420 @@
...
+#include <limits.h>
+#include <errno.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <openssl/pem.h>
+#include <openssl/x509.h>
+
+#include "../crypto/internal.h"
+
+#define MAX_COLLISIONS 256
```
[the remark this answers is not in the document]
skmcgrail: Or Maybe `MAX_BUCKET_ENTRIES`?

crypto/test/test_util.h
smittals2: I didn't add windows support since I needed this func for the rehash utility and we aren't supporting windows there. I can look into making this platform agnostic/have support for windows.

tool-openssl/c_rehash_test.sh
smittals2: File is renamed when placed into build and install dirs: 
https://github.com/aws/aws-lc/pull/2258/files#diff-770a4bd1d030385193d7470f7eef5fead79d759b05f9bd3a9da5a3801fd196b7R49

tool-openssl/rehash_test.cc
```diff
@@ -0,0 +1,257 @@
...
+  ASSERT_FALSE(result);
+  // errno should not be set
+  ASSERT_EQ(errno, 0);
+}
+
+// We cannot force the order in which rehash processes files from the given
+// directory. Therefore, we cannot deterministically know the symlink # suffix
+// for a given filename. We only check that the correct symlinks exist, with
+// the correct hash value, and the correct suffix ("r" vs "") for CRL vs CERT.
+// We do not verify the number suffix.
```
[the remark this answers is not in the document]
smittals2: Yes, OpenSSL also just reads the files from the dir. Which order those files get processed in is not enforced (they enforce stuff like process existing symlinks first, but nothing for the order of crl or pem files). 

The reason I left this comment here is previously I was trying to write the test to assert a certain file would have a certain suffix (in the case of collisions). However, I had to rewrite the test to make more general assertions instead of mapping a file to a specific suffix.

tool-openssl/rehash.cc
```diff
@@ -0,0 +1,420 @@
...
+
+  // If not found, create a new bucket at |hash_idx|
+  if (bucket == NULL) {
+    bucket = (BUCKET *)OPENSSL_zalloc(sizeof(*bucket));
+    if (bucket == NULL) {
+      fprintf(stderr, "ERROR: Failed to allocate new bucket\n");
+      return;
+    }
+    // Store any previous bucket at the same index but with a
+    // different |type| + |hash| combination or NULL
```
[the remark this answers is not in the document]
smittals2: Fair enough, I guess I already specify we're checking for existing buckets with a type + hash combo before this, so if not found, we can just say inserting to list and the rest is implied.

tool-openssl/rehash.cc
```diff
@@ -0,0 +1,420 @@
...
+
+  if (extra_args.empty()) { // No directory path provided on command line
+    const char* ssl_cert_dir = getenv("SSL_CERT_DIR");
+    if (ssl_cert_dir != nullptr) {
+      directory_path = ssl_cert_dir;
+    } else {
+      directory_path = X509_get_default_cert_dir();
+    }
+  } else {
+    directory_path = extra_args[0];
```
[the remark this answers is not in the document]
smittals2: OpenSSL does support passing in multiple directory paths, and it will then sequentially process them. There was a lot of logic around it to support duplicate filenames across dirs which increased the hashtable/processing complexities. It was a conscious decision to support one dir at a time for now. Our use cases don't require multiple dir and we gracefully fail if more than one dir is provided: https://github.com/aws/aws-lc/pull/2258/files#diff-e590531f75e21ff646dc20b9f15dc108fdd89d3c07689179a4e83968644f620cR341

tool-openssl/c_rehash.sh
justsmth: Right, the question is whether we want our custom error message to be displayed in this case. I suppose it's highly unlikely that the file would exist but not be executable.

tool-openssl/rehash.cc
```diff
@@ -0,0 +1,420 @@
...
+    if (bucket == NULL) {
+      fprintf(stderr, "ERROR: Failed to allocate new bucket\n");
+      return;
+    }
+    // Insert new bucket as head of linked-list
+    bucket->next = hash_table[hash_idx];
+    bucket->type = type;
+    bucket->hash = hash;
+    hash_table[hash_idx] = bucket;
+  }
```
justsmth, line 76: Why not use [`std::map`](https://en.cppreference.com/w/cpp/container/map) and have most this logic provided for you?

tool-openssl/rehash.cc
```diff
@@ -0,0 +1,420 @@
...
+
+  bssl::UniquePtr<STACK_OF(X509_INFO)> x509_info_stack(
+    PEM_X509_INFO_read_bio(bio, nullptr, nullptr, nullptr));
+  BIO_free(bio);
+
+  // Ensure there is only one cert/CRL in the file, this is not an error
+  if (!x509_info_stack || sk_X509_INFO_num(x509_info_stack.get()) != 1) {
+    return;
+  }
```
justsmth, line 145: Should we treat `x509_info_stack == nullptr` the same way as `k_X509_INFO_num(x509_info_stack.get()) != 1`?

tool-openssl/rehash.cc
```diff
@@ -0,0 +1,420 @@
...
+    type = TYPE_CRL;
+    if (!X509_CRL_digest(x509_info->crl, EVP_sha1(), digest, &digest_len)) {
+      status_flag = false;
+      return;
+    }
+    x509_name = X509_CRL_get_issuer(x509_info->crl);
+  } else {
+    status_flag = false;
+    return;
+  }
```
justsmth, line 174: No messages output regarding these errors?

tool-openssl/rehash.cc
```diff
@@ -0,0 +1,420 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include "internal.h"
+#include "../tool/internal.h"
+
+#if !defined(OPENSSL_WINDOWS)
```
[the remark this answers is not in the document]
smittals2: I don't have a great understanding of every caveat needed for windows. C++17 onwards you have platform agnostic utilities but to support C++11 we would have to use Windows specific functions. 
For example we use dirent to process entries in a directory, would need windows specific code here. Would need windows specific code for removing existing symlinks, and then creating new ones. 

So pretty much every processing step of this script would need ifdefs for windows (which can always be added later). I just chose not to invest in that since not needed (and this doesn't break anyone since it's a new utility we're adding).

tool-openssl/rehash.cc
smittals2: I started with OpenSSL design and refactored/cut down on stuff we don't need/or support. I didn't know when starting out if any type of ordering is enforced, or whether their custom type+hash index calculations were a intentional design choice on OpenSSL's end. To maintain most of their processing logic, I did very targeted refactors at least while coding this. 

In hindsight, we could use a map also. Currently we support an index -> list of buckets. bucket -> list of hash entries. This can be done with nested maps which would be a bit messy. Or we could get rid of the list of buckets layer and instead store each type,hash tuple as a key pointing to a list of hash entries - this eliminates the need for bucket lists.

I have unit and integration tests for this hashtable structure to ensure these operations are valid. Moving to map is feasible, but a bit of a design refactor. This is just a utility tool and not "operational" code but happy to do it if you think it's worth it.

tool-openssl/rehash.cc
```diff
@@ -0,0 +1,420 @@
...
+      status_flag = false;
+      return;
+    }
+
+    regex_t regex;
+    int ret = regcomp(&regex, "^[0-9a-f]{8}\\.([r])?[0-9]+$", REG_EXTENDED | REG_NOSUB);
+    if (ret) {
+      fprintf(stderr, "Could not compile regex\n");
+      return;
+    }
```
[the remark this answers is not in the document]
smittals2: - We would only produce symlinks matching this regex, so technically we should only remove/touch symlinks we have created. But at the same time, if we have duplicate symlinks in the same dir where capitalization is the only difference, downstream applications may be confused/not distinguish between this. Updating the regex to match uppercase also. 
- Fair, compiling the regex shouldn't be failing.

tool-openssl/rehash.cc
```diff
@@ -0,0 +1,420 @@
...
+  }
+
+  // Create new entry
+  if (bucket->num_entries >= MAX_BUCKET_ENTRIES) {
+    fprintf(stderr, "Error: hash table overflow for %s\n", filename);
+    status_flag = false;
+    return;
+  }
```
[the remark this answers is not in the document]
smittals2: OpenSSL doesn't error out in this case. If there is a overflow, that just means we can't hash that cert - we should still continue other certs in the dir. OpenSSL collects the number of errors and returns it at the end. 

Relevant check in OpenSSL code: 
https://github.com/openssl/openssl/blob/2d50cb660cdf1802a9c3e4b5fc5366a85e2158b4/apps/rehash.c#L163

Accumulating errors: https://github.com/openssl/openssl/blob/2d50cb660cdf1802a9c3e4b5fc5366a85e2158b4/apps/rehash.c#L305

tool-openssl/rehash.cc
smittals2: Logic wise yes, neither are errors and should continue. I've decoupled the checks and added a warning print statement for  sk_X509_INFO_num(x509_info_stack.get()) != 1)

tool-openssl/rehash.cc
```diff
@@ -406,12 +404,14 @@ bool RehashTool(const args_list_t &args) {
   regex_t regex;
   int ret = regcomp(&regex, "^[0-9a-fA-F]{8}\\.([r])?[0-9]+$", REG_EXTENDED | REG_NOSUB);
   if (ret) {
+    regfree(&regex);
```
justsmth, line 407: NP:
* `regex` should be initialized to NULL in case the call to `regcomp` fails and it doesn't modify `regex`.
* Is it ok to call `regfree` with a NULL value? Documentation doesn't say: https://linux.die.net/man/3/regfree

justsmth (approved):

justsmth (approved):

skmcgrail (approved):

PR #2259, opened 2025-03-09
Avoid double seed for non-UBE support runtimes

torben-hansen:
### Description of changes: 

In the non-UBE support runtime, a fresh lazy init will spawn a seed of the CTR-DRBG state. But then in `rand_bytes_core()`, there will be an immediate spawn of a reseed of the CTR-DRBG state to ensure uniqueness for non-UBE support.

Fix this by telling `rand_bytes_core` whether the CTR-DRBG state is fresh or not.

This requires a few changes in the tests to account for a reseed that might not happen. Now I'm there, refactor a bit.

### DIFF
--- a/crypto/fipsmodule/rand/new_rand_test.cc
+++ b/crypto/fipsmodule/rand/new_rand_test.cc
@@ -61,9 +61,6 @@ static void randBasicTests(bool *returnFlag) {
 }
 
 TEST_F(newRandTest, Basic) {
-  bool returnFlag = false;
-  randBasicTests(&returnFlag);
-  ASSERT_TRUE(returnFlag);
   ASSERT_TRUE(threadTest(number_of_threads, randBasicTests));
 }
 
@@ -116,9 +113,6 @@ TEST_F(newRandTest, ReseedIntervalWhenUbeIsSupported) {
   if (!UbeIsSupported()) {
     GTEST_SKIP() << "UBE detection is not supported";
   }
-  bool returnFlag = false;
-  randReseedIntervalUbeIsSupportedTests(&returnFlag);
-  ASSERT_TRUE(returnFlag);
   ASSERT_TRUE(threadTest(number_of_threads, randReseedIntervalUbeIsSupportedTests));
 }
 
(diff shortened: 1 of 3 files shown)


crypto/fipsmodule/rand/rand.c
```diff
@@ -489,18 +495,24 @@ static void rand_bytes_private(uint8_t *out, size_t out_len,
 
     rand_state_initialize(state);
     thread_local_list_add_node(state);
+
+    ctr_drbg_state_is_fresh = 0;
```
andrewhop, line 499: Should this be a 1?
```suggestion
    ctr_drbg_state_is_fresh = 1;
```

torben-hansen: sigh

skmcgrail (approved):

crypto/fipsmodule/rand/rand.c
```diff
@@ -489,18 +496,24 @@ static void rand_bytes_private(uint8_t *out, size_t out_len,
...
   FIPS_service_indicator_update_state();
 }
 
 int RAND_bytes_with_user_prediction_resistance(uint8_t *out, size_t out_len,
   const uint8_t user_pred_resistance[RAND_PRED_RESISTANCE_LEN]) {
-  
+
+  if (user_pred_resistance == NULL) {
+    abort();
+  }
```
andrewhop, line 515: Why not return 0 for error? I also checked and rand_bytes_core uses GUARD_PTR_ABORT so this should at least be consistent. But I'm still interested why all the aborts on invalid inputs.

crypto/fipsmodule/rand/new_rand_test.cc
```diff
@@ -18,39 +18,59 @@
...
+static const size_t global_request_len = 64;
+static const size_t number_of_threads = 8;
+
+static void test_all_exported_functions(size_t request_len, uint8_t *out_buf,
+  uint8_t user_pred_res[RAND_PRED_RESISTANCE_LEN]) {
+  ASSERT_TRUE(RAND_bytes(out_buf, request_len));
+  ASSERT_TRUE(RAND_priv_bytes(out_buf, request_len));
+  ASSERT_TRUE(RAND_pseudo_bytes(out_buf, request_len));
+  ASSERT_TRUE(RAND_bytes_with_user_prediction_resistance(out_buf, request_len, user_pred_res));
+}
```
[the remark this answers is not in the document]
torben-hansen: Not anything basic. It's a simple check to verify they succeed for different request lengths.

crypto/fipsmodule/rand/rand.c
torben-hansen: I was debating that. Could return 0. But then I changed to abort because wanted to stay consistent with the fact that all failure modes result in an abort.

torben-hansen: I think I lean towards following the current "abort" pattern.

PR #2263, opened 2025-03-11
ML-KEM memory safety

m271828:
WIP

### DIFF
--- a/crypto/fipsmodule/ml_kem/ml_kem.c
+++ b/crypto/fipsmodule/ml_kem/ml_kem.c
@@ -24,18 +24,25 @@
 // platform support.
 
 int ml_kem_512_keypair_deterministic(uint8_t *public_key  /* OUT */,
+                                         size_t *public_len /* IN_OUT */,
                                          uint8_t *secret_key  /* OUT */,
+                                         size_t *secret_len /* IN_OUT */,
                                          const uint8_t *seed  /* IN */) {
   boringssl_ensure_ml_kem_self_test();
-  return ml_kem_512_keypair_deterministic_no_self_test(public_key, secret_key, seed);
+  return ml_kem_512_keypair_deterministic_no_self_test(public_key, public_len, secret_key, secret_len, seed);
 }
 
 int ml_kem_512_keypair_deterministic_no_self_test(uint8_t *public_key  /* OUT */,
+                                                  size_t *public_len /* IN_OUT */,
                                                   uint8_t *secret_key  /* OUT */,
+                                                  size_t *secret_len /* IN_OUT */,
                                                   const uint8_t *seed  /* IN */) {
   ml_kem_params params;
   int res;
   ml_kem_512_params_init(&params);
+  if (*public_len < params.public_key_bytes || *secret_len < params.secret_key_bytes) {
+    return 0;
+  }
   res = ml_kem_keypair_derand_ref(&params, public_key, secret_key, seed);
 #if defined(AWSLC_FIPS)
   /* PCT failure is the only failure condition for key generation. */
@@ -43,15 +50,24 @@ int ml_kem_512_keypair_deterministic_no_self_test(uint8_t *public_key  /* OUT */
       AWS_LC_FIPS_failure("ML-KEM keygen PCT failed");
   }
 #endif
+  if (res == 0) {
+    *public_len = params.public_key_bytes;
+    *secret_len = params.secret_key_bytes;
+  }
   return res;
 }
 
 int ml_kem_512_keypair(uint8_t *public_key /* OUT */,
-                           uint8_t *secret_key /* OUT */) {
+                           size_t *public_len /* IN_OUT */,
+                           uint8_t *secret_key /* OUT */,
+                           size_t *secret_len /* IN_OUT */) {
   boringssl_ensure_ml_kem_self_test();
   int res;
   ml_kem_params params;
   ml_kem_512_params_init(&params);
+  if (*public_len < params.public_key_bytes || *secret_len < params.secret_key_bytes) {
+    return 0;
+  }
   res = ml_kem_keypair_ref(&params, public_key, secret_key);
 #if defined(AWSLC_FIPS)
   /* PCT failure is the only failure condition for key generation. */
@@ -59,6 +75,10 @@ int ml_kem_512_keypair(uint8_t *public_key /* OUT */,
       AWS_LC_FIPS_failure("ML-KEM keygen PCT failed");
   }
 #endif
+  if (res == 0) {
+    *public_len = params.public_key_bytes;
+    *secret_len = params.secret_key_bytes;
+  }
   return res;
 }
 
@@ -106,12 +126,17 @@ int ml_kem_512_decapsulate_no_self_test(uint8_t *shared_secret    /* OUT */,
 
 
 int ml_kem_768_keypair_deterministic(uint8_t *public_key  /* OUT */,
+                                         size_t *public_len /* IN_OUT */,
                                          uint8_t *secret_key  /* OUT */,
+                                         size_t *secret_len /* IN_OUT */,
                                          const uint8_t *seed  /* IN */) {
   boringssl_ensure_ml_kem_self_test();
   ml_kem_params params;
   int res;
   ml_kem_768_params_init(&params);
+  if (*public_len < params.public_key_bytes || *secret_len < params.secret_key_bytes) {
+    return 0;
+  }
   res = ml_kem_keypair_derand_ref(&params, public_key, secret_key, seed);
 #if defined(AWSLC_FIPS)
   /* PCT failure is the only failure condition for key generation. */
@@ -119,15 +144,24 @@ int ml_kem_768_keypair_deterministic(uint8_t *public_key  /* OUT */,
       AWS_LC_FIPS_failure("ML-KEM keygen PCT failed");
   }
 #endif
+  if (res == 0) {
+    *public_len = params.public_key_bytes;
+    *secret_len = params.secret_key_bytes;
+  }
   return res;
 }
 
 int ml_kem_768_keypair(uint8_t *public_key /* OUT */,
-                           uint8_t *secret_key /* OUT */) {
+                           size_t *public_len /* IN_OUT */,
+                           uint8_t *secret_key /* OUT */,
+                           size_t *secret_len /* IN_OUT */) {
   boringssl_ensure_ml_kem_self_test();
   ml_kem_params params;
   int res;
   ml_kem_768_params_init(&params);
+  if (*public_len < params.public_key_bytes || *secret_len < params.secret_key_bytes) {
+    return 0;
+  }
   res = ml_kem_keypair_ref(&params, public_key, secret_key);
 #if defined(AWSLC_FIPS)
   /* PCT failure is the only failure condition for key generation. */
@@ -135,6 +169,10 @@ int ml_kem_768_keypair(uint8_t *public_key /* OUT */,
       AWS_LC_FIPS_failure("ML-KEM keygen PCT failed");
   }
 #endif
+  if (res == 0) {
+    *public_len = params.public_key_bytes;
+    *secret_len = params.secret_key_bytes;
+  }
   return res;
 }
 
@@ -167,12 +205,17 @@ int ml_kem_768_decapsulate(uint8_t *shared_secret    /* OUT */,
 }
 
 int ml_kem_1024_keypair_deterministic(uint8_t *public_key  /* OUT */,
+                                          size_t *public_len /* IN_OUT */,
                                           uint8_t *secret_key  /* OUT */,
+                                          size_t *secret_len /* IN_OUT */,
                                           const uint8_t *seed  /* IN */) {
   boringssl_ensure_ml_kem_self_test();
   ml_kem_params params;
   int res;
   ml_kem_1024_params_init(&params);
+  if (*public_len < params.public_key_bytes || *secret_len < params.secret_key_bytes) {
+    return 0;
+  }
   res = ml_kem_keypair_derand_ref(&params, public_key, secret_key, seed);
 #if defined(AWSLC_FIPS)
   /* PCT failure is the only failure condition for key generation. */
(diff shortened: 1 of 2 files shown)


crypto/fipsmodule/kem/internal.h
```diff
@@ -53,7 +62,9 @@ typedef struct {
 struct kem_key_st {
   const KEM *kem;
```
torben-hansen, line 63: `KEM` object encodes the length of parameters for a particular KEM parameter set. Maybe those could be used?

crypto/fipsmodule/evp/p_kem.c
```diff
@@ -67,9 +67,11 @@ static int pkey_kem_keygen_deterministic(EVP_PKEY_CTX *ctx,
   }
 
   KEM_KEY *key = KEM_KEY_new();
+  size_t public_len = kem->public_key_len;
```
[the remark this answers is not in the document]
torben-hansen: Oops I read `kem` as `key`...

crypto/fipsmodule/ml_kem/ml_kem.c
```diff
@@ -16,213 +16,369 @@
...
 
 int ml_kem_512_keypair_deterministic_no_self_test(uint8_t *public_key  /* OUT */,
+                                                  size_t *public_len /* IN_OUT */,
                                                   uint8_t *secret_key  /* OUT */,
+                                                  size_t *secret_len /* IN_OUT */,
                                                   const uint8_t *seed  /* IN */) {
   ml_kem_params params;
   int res;
   ml_kem_512_params_init(&params);
+  if (*public_len < params.public_key_bytes || *secret_len < params.secret_key_bytes) {
```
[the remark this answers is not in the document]
torben-hansen: Ignore this...

crypto/fipsmodule/self_check/self_check.c
```diff
@@ -1049,10 +1053,29 @@ static int boringssl_self_test_ml_kem(void) {
...
+          keygen_encaps, &encaps_len, keygen_decaps, &decaps_len, kKeyGenDKSeed) ||
       !check_test(kKeyGenDK, keygen_decaps, sizeof(keygen_decaps),
-                  "ML-KEM-keyGen-decaps")) {
+                  "ML-KEM-keyGen-decaps") ||
+      encaps_len > MLKEM512_PUBLIC_KEY_BYTES ||
+      decaps_len > MLKEM512_SECRET_KEY_BYTES) {
+    goto err;
+  }
+
+  size_t wrong_encaps_len = 1;
```
torben-hansen, line 1067: Should have a dedicated test file for these type of tests. not in self check.

crypto/fipsmodule/kem/internal.h
```diff
@@ -53,7 +62,9 @@ typedef struct {
...
   uint8_t *public_key;
+  size_t *public_key_len;
```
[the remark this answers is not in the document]
m271828: Good catch, I didn't end up using them.

crypto/fipsmodule/kem/internal.h
```diff
@@ -14,22 +14,31 @@ extern "C" {
...
 typedef struct {
   int (*keygen_deterministic)(uint8_t *ctx,
+                              size_t *ctx_len,
                               uint8_t *pkey,
+                              size_t *pkey_len,
                               const uint8_t *seed);
 
   int (*keygen)(uint8_t *public_key,
-                uint8_t *secret_key);
+                size_t *public_key_len,
```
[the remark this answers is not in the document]
m271828: They're an in/out value. If the length is longer than what's used, the written length is put in the parameter. That's also why it's not `const`.

crypto/fipsmodule/ml_kem/ml_kem.h
```diff
@@ -28,76 +28,117 @@
...
+#endif
+
+OPENSSL_EXPORT int ml_kem_512_keypair_deterministic(uint8_t *public_key /* OUT */,
+                                         size_t *public_len /* IN_OUT */,
                                          uint8_t *secret_key /* OUT */,
+                                         size_t *secret_len /* IN_OUT */,
                                          const uint8_t *seed /* IN */);
 
-int ml_kem_512_keypair_deterministic_no_self_test(uint8_t *public_key  /* OUT */,
+OPENSSL_EXPORT int ml_kem_512_keypair_deterministic_no_self_test(uint8_t *public_key  /* OUT */,
```
torben-hansen, line 41: `foo_no_self_test` functions don't need to be exported.

crypto/fipsmodule/ml_kem/ml_kem.c
```diff
@@ -16,213 +16,369 @@
...
+  }
   res = ml_kem_keypair_derand_ref(&params, public_key, secret_key, seed);
 #if defined(AWSLC_FIPS)
   /* PCT failure is the only failure condition for key generation. */
   if (res != 0) {
       AWS_LC_FIPS_failure("ML-KEM keygen PCT failed");
   }
 #endif
+  if (res == 0) {
+    *public_len = params.public_key_bytes;
```
torben-hansen, line 53: I don't believe these are needed. What you are effectively verifying is that the values in the parameter set `params` is equal to the parameters from `kem` at the original call site in functions from `p_kem.c`.

Could keep the length checks above, but I'm not sure I see what sending the value back up the call-stack provides.

If one wanted to micro-optimize, could even do the parameter-equal test once, instead of doing it on every call path. it wouldn't save much though.

torben-hansen: (not actually sure why we have two objects, that stores the same kyber parameter values)

m271828: Keygen is the only path where the memory allocation is controlled inside the EVP layer. The others are controlled outside of it. The ticket @andrewhop wrote had the lengths being in/out parameters in case the array length was more than what's needed as a way to signal back the usable length.

The length checks are only being done when the pointer is being written to, so we do want to check it each time these functions are called.

We could make them input only parameters but we'd need to reconcile between review comments and the ticket. Since there's a similar ticket for the PQ team where the intention is for them to write this the same way, I'm inclined to leave it as is or make sure we update the ticket for PQ so it's consistent.

crypto/fipsmodule/ml_kem/ml_kem.h
m271828: This needed to be exported so it could be reached by gtest. The code isn't fully exercisable by EVP and when I asked in Slack if there was an alternative to exporting the symbols, the answer I got was no, just do an export in an internal header.

hanno-becker:
FWIW, this is compatible with #2176.

crypto/fipsmodule/ml_kem/ml_kem.h
torben-hansen: Yeah, the other exports are good. I meant that e.g. `ml_kem_512_keypair_deterministic_no_self_test()` and in general functions with suffix `_no_self_test()` does not seem to be used in the testing code. Maybe I missed them?

m271828: I'm following. Yes, I didn't write tests for these because the versions that do a self test run their check and then call the `*_no_self_test()` versions. It seemed redundant to write a second test so you're correct, the exports can be removed on these functions.

crypto/fipsmodule/ml_kem/ml_kem.c
```diff
@@ -16,213 +16,318 @@
...
 #include "./ml_kem_ref/symmetric-shake.c"
 #include "./ml_kem_ref/verify.c"
-#include "../../internal.h"
+
+// Ensure buffer is long enough and zero any extra memory
+static int check_and_initialize_buffer(uint8_t *buffer, const size_t len, const size_t expeted_len) {
+  if (buffer == NULL || len < expeted_len) {
+    return 0;
+  }
+  for (size_t i = expeted_len; i < len; i++) {
```
hanno-becker, line 25: What's the motivation for this change? It feels unnecessary to me.

m271828: It was to zero out the portion of the buffer that wouldn't be written to but other algorithms don't seem to do that so I removed it.

m271828:
Updated to resolve merge conflicts from https://github.com/aws/aws-lc/pull/2176 (change in how parameters are being used, new functions that were added have length checks too).

crypto/fipsmodule/ml_kem/mlkem/kem.h
```diff
@@ -10,6 +10,10 @@
 #include "common.h"
 #include "sys.h"
 
+#if defined(__cplusplus)
```
torben-hansen, line 13: Changes in this file should be done upstream in mlkem-native. It would get erased on next import otherwise.

crypto/fipsmodule/ml_kem/ml_kem.c
```diff
@@ -28,169 +28,375 @@
...
+int ml_kem_1024_decapsulate_no_self_test(uint8_t *shared_secret    /* OUT */,
+                                         size_t *shared_secret_len /* IN_OUT */,
+                                         const uint8_t *ciphertext /* IN */,
+                                         const uint8_t *secret_key /* IN */) {
+  OutputBuffer ss = {shared_secret, shared_secret_len, MLKEM1024_SHARED_SECRET_LEN};
+  return ml_kem_common_decapsulate(mlkem1024_dec, ss, ciphertext, secret_key);
+}
+
+
+int check_buffer(const OutputBuffer data) {
```
torben-hansen, line 332: `check_buffer()` and `set_written_len_on_success()` can both be static functions. Could also consider moving them up to the start of the file - then they won't need a declaration either.

crypto/fipsmodule/ml_kem/ml_kem.c
```diff
@@ -28,169 +28,375 @@
...
 #include "./ml_kem.h"
 
-int ml_kem_512_keypair_deterministic(uint8_t *public_key  /* OUT */,
-                                     uint8_t *secret_key  /* OUT */,
-                                     const uint8_t *seed  /* IN */) {
+typedef struct {
+  uint8_t *buffer;
+  size_t *length;
+  const size_t expected_length;
+} OutputBuffer;
```
torben-hansen, line 35: snake case is typically used for type names.

crypto/fipsmodule/ml_kem/ml_kem.c
m271828: Done

crypto/fipsmodule/ml_kem/ml_kem.c
m271828: Done

crypto/fipsmodule/ml_kem/mlkem/kem.h
m271828: Ended up not being needed after all changes were made for the removal of the parameter types.

torben-hansen (approved):

crypto/fipsmodule/ml_kem/ml_kem_test.cc
```diff
@@ -0,0 +1,1628 @@
...
+
+TEST_P(MLKEMKeypairLengthTest, AllShortLengthDeterministic) {
+  MLKEMKeypairTestVector params = GetParam();
+  size_t encaps_len = params.public_len/2;
+  encaps = new uint8_t[encaps_len];
+  size_t decaps_len = params.secret_len/2;
+  decaps = new uint8_t[decaps_len];
+  ASSERT_NE(0, params.keypair_deterministic(encaps, &encaps_len, decaps, &decaps_len, params.seed));
+  ASSERT_EQ(params.public_len/2, encaps_len);
+  ASSERT_EQ(params.secret_len/2, decaps_len);
```
andrewhop, line 112: Double checking: we're saying the low level keypair_determinisitic does not change output lengths if they're too small? Can we make the length a const pointer or just pass in a regular int?

crypto/fipsmodule/ml_kem/ml_kem_test.cc
```diff
@@ -0,0 +1,1628 @@
...
+  size_t encaps_len = params.public_len*2;
+  encaps = new uint8_t[encaps_len];
+  size_t decaps_len = params.secret_len*2;
+  decaps = new uint8_t[decaps_len];
+  ASSERT_EQ(0, params.keypair(encaps, &encaps_len, decaps, &decaps_len));
+  ASSERT_EQ(params.public_len, encaps_len);
+  ASSERT_EQ(params.secret_len, decaps_len);
+}
+
+static const uint8_t encaps512Key[MLKEM512_PUBLIC_KEY_BYTES] = {
```
andrewhop, line 192: Just wondering does this test depend on this particular key or could it pull an existing KAT or generate one? It's a small announce because it makes reading the file harder (self_check.c is out of control but requires the KAT be baked into the library).

crypto/kyber/kem_kyber.c
```diff
@@ -12,30 +12,39 @@
 // and fails (0). Define wrappers to handle that.
 
 static int kyber512r3_keygen_deterministic(uint8_t *public_key,
+                                           size_t *public_len,
```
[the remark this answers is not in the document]
m271828: This uses the same API as ML-KEM. I didn't make changes here per @darylmartin100 as this code is being replaced.

crypto/fipsmodule/ml_kem/ml_kem_test.cc
m271828: This behavior comes from the public API and can't be changed. I discussed this in one of the resolved comments on the PR previously. The EVP layer for keygen (which is more than just ML-KEM) has the following behavior:
- Takes in a pointer to a memory block and a pointer to the length of the memory block
- If the allocated length is too short, it returns everything unchanged
- If the length is sufficient, the generated key is written to memory and the length parameter is changed to how many bytes were written (as allocated length may be longer than needed)
It's a bit of a misuse of the length parameter in my mind, but we don't want to change our public APIs.

crypto/fipsmodule/ml_kem/ml_kem_test.cc
m271828: It doesn't depend on this particular keys, just that the key be valid. When testing, I've always used a hard coded known good key in cases like this to avoid false failures (this is testing the length checks only, we don't want a test failing because of an invalid key, that should cause a different test to fail).

torben-hansen (approved):

andrewhop (approved):

PR #2264, opened 2025-03-12
Add support for verifying PKCS7 signed attributes

samuel40791765:
### Issues:
Resolves `CryptoAlg-2946`

### Description of changes: 
I discovered this while trying to fix our PKCS7 implementation to use indefinite encoding. The PKCS7 file that Ruby's PKCS7 tests uses signed attributes, but `PKCS7_verify` bails out whenever it encounters files that have signed attributes. There are still other issues with the Ruby PKCS7 test that we'll have to fix (indefinite length ASN1), but I believe we should fix the missing support for verifying signed attributes first.
AWS-LC turns on `PKCS7_NOATTR` by default in `PKCS7_sign`, so our existing `PKCS7_verify` implementation can do a successful sign/verify round trip against itself. However, OpenSSL does not turn on `PKCS7_NOATTR` by default and signed attributes are added automatically to the PKCS7 file if no flags are set. This means that the current state of AWS-LC's `PKCS7_verify` would fail against files generated by the default of OpenSSL's `PKCS7_sign`. This PR adds support for verifying PKCS7 signed attributes to fix the misalignment.

### Call-outs:
N/A

### Testing:
1. Test file from Ruby. The intention is to build upon the test to fix the ASN1 indefinite length issues later on.
2. Test file generated by OpenSSL. Our `PKCS7_sign` does not support signed attributes, so we can only test against generated files by OpenSSL.


crypto/pkcs7/pkcs7_test.cc
```diff
@@ -2068,3 +2071,153 @@ TEST(PKCS7Test, SetDetached) {
   EXPECT_TRUE(PKCS7_set_detached(p7.get(), 1));
   EXPECT_FALSE(p7.get()->d.sign->contents->d.data);
 }
+
+TEST(PKCS7Test, PKCS7SignedAttributes) {
```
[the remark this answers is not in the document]
samuel40791765: Great point that we could consider doing. I believe we'll be redoing some of our PKCS7 ASN1 parsing soon to fix the indefinite ber issues we've been having.
We can add these two files along with any new test files that come up to the corpus in a PR once both are completed.

crypto/pkcs7/pkcs7.c
```diff
@@ -1548,26 +1572,64 @@ static int pkcs7_signature_verify(BIO *in_bio, PKCS7 *p7, PKCS7_SIGNER_INFO *si,
...
+      OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_UNABLE_TO_FIND_MESSAGE_DIGEST);
+      goto out;
+    }
+    if (message_digest->length != (int)md_len ||
+        OPENSSL_memcmp(message_digest->data, md_data, md_len) != 0) {
+      OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_DIGEST_FAILURE);
+      goto out;
+    }
+
+    if (!EVP_VerifyInit_ex(mdc_tmp, EVP_get_digestbynid(md_type), NULL)) {
```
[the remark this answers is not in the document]
samuel40791765: It seems a bit redundant, considering we do a search and match against `md_type` earlier in the function. But will add.

crypto/pkcs7/pkcs7.c
```diff
@@ -1528,6 +1544,14 @@ static int pkcs7_signature_verify(BIO *in_bio, PKCS7 *p7, PKCS7_SIGNER_INFO *si,
...
 
+  // Create a new |EVP_MD_CTX| to be consumed, so that the original |EVP_MD_CTX|
+  // is still in a usable state.
+  EVP_MD_CTX *mdc_tmp = EVP_MD_CTX_new();
+  if (mdc_tmp == NULL) {
+    OPENSSL_PUT_ERROR(PKCS7, ERR_R_MALLOC_FAILURE);
+    goto out;
+  }
+
   const int md_type = OBJ_obj2nid(si->digest_alg->algorithm);
```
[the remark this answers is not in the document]
samuel40791765: Added

crypto/pkcs7/pkcs7.c
```diff
@@ -461,6 +457,27 @@ ASN1_TYPE *PKCS7_get_signed_attribute(const PKCS7_SIGNER_INFO *si, int nid) {
...
+  if (si == NULL) {
+    OPENSSL_PUT_ERROR(PKCS7, ERR_R_PASSED_NULL_PARAMETER);
+    return NULL;
+  }
+  return get_attribute(si->auth_attr, nid);
+}
+
+static ASN1_OCTET_STRING *PKCS7_digest_from_attributes(
+    STACK_OF(X509_ATTRIBUTE) *sk) {
+  if (sk == NULL) {
```
WillChilds-Klein, line 470: nit: should we use `GUARD_PTR`?

crypto/pkcs7/pkcs7.c
```diff
@@ -1548,26 +1575,65 @@ static int pkcs7_signature_verify(BIO *in_bio, PKCS7 *p7, PKCS7_SIGNER_INFO *si,
     bio = BIO_next(bio);
   }
 
-  // We don't currently support signed attributes. See |PKCS7_NOATTR|.
-  if (si->auth_attr && sk_X509_ATTRIBUTE_num(si->auth_attr) != 0) {
-    OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_INVALID_SIGNED_DATA_TYPE);
+  // mdc is the digest ctx that we want, unless there are attributes, in
+  // which case the digest is the signed attributes.
+  if (!EVP_MD_CTX_copy_ex(mdc_tmp, mdc)) {
     goto out;
```
WillChilds-Klein, line 1581: `mdc` is initialized to NULL on L1559, never populated, then used as the `in` parameter in the call to `EVP_MD_CTX_copy_ex`. so what are we copying into `mdc_tmp` here? and why allocate to `mdc_tmp` if it just gets overwritten here before it's used?

crypto/pkcs7/pkcs7.c
samuel40791765: It works and we've contemplated it before in similar instances, but there were concerns with `GUARD_PTR` returning a `0` and returning `0` as `NULL` being un-guaranteed behavior within C
* https://c-faq.com/null/runtime0.html

crypto/pkcs7/pkcs7.c
samuel40791765: `mdc` is initially `NULL`, until we find the correct `EVP_MD_CTX` to work with from `bio`. That is then retrieved with `BIO_get_md_ctx` on L1568, which directly returns a pointer to the underlying `EVP_MD_CTX`.

I believe the copying's done because upstream/we wanted to use `EVP_MD_CTX` for further verification operations, but we still wished to preserve the original state of the retrieved context. This allows for multiple operations of `PKCS7_verify`, without consuming the context. Perhaps we should add a test for that as well.

samuel40791765: Test added!

crypto/pkcs7/pkcs7.c
```diff
@@ -866,9 +883,7 @@ int PKCS7_set_detached(PKCS7 *p7, int detach) {
   }
 }
 
-int PKCS7_get_detached(PKCS7 *p7) {
-  return PKCS7_is_detached(p7);
-}
+int PKCS7_get_detached(PKCS7 *p7) { return PKCS7_is_detached(p7); }
```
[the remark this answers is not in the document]
samuel40791765: Yes it does that whenever the implementation can be fit within 78 characters.

justsmth (approved):

WillChilds-Klein (approved):

PR #2266, opened 2025-03-12
Make AWS_LC_fips_failure_callback optional in builds with AWSLC_FIPS_FAILURE_CALLBACK

andrewhop:
### Description of changes: 
This makes the FIPS callback optional when AWS-LC is built with AWSLC_FIPS_FAILURE_CALLBACK which turns it on. Previously if AWS-LC was built with AWSLC_FIPS_FAILURE_CALLBACK then any application that might link with AWS-LC would need to define the callback or AWS-LC would abort. Turns out that's hard to ensure, this change goes back to the old behavior:
* Regular FIPS build of AWS-LC doesn't require a callback and always aborts
* FIPS build with AWSLC_FIPS_FAILURE_CALLBACK allows a callback to be defined and AWS-LC will call that and not abort
* FIPS build with AWSLC_FIPS_FAILURE_CALLBACK and no callback defined fallback to the first bullet point behavior

### Call-outs:
I had to remove `EXPECT_TRUE(EVP_PKEY_keygen(ed_ctx.get(), &ed_key));` because that keygen call should only happen after the check for the environment variable and now we're testing both paths.

### Testing:
I created a new test target that partially replicates our internal FIPS callback build.

### DIFF
--- a/crypto/fipsmodule/bcm.c
+++ b/crypto/fipsmodule/bcm.c
@@ -270,13 +270,6 @@ static void BORINGSSL_bcm_power_on_self_test(void) __attribute__ ((constructor))
 #endif
 
 static void BORINGSSL_bcm_power_on_self_test(void) {
-#if defined(AWSLC_FIPS_FAILURE_CALLBACK)
-  if (AWS_LC_fips_failure_callback == NULL) {
-    fprintf(stderr, "AWS_LC_fips_failure_callback not defined but AWS-LC built with AWSLC_FIPS_FAILURE_CALLBACK\n");
-    fflush(stderr);
-    abort();
-  }
-#endif
 // TODO: remove !defined(OPENSSL_PPC64BE) from the check below when starting to support
 // PPC64BE that has VCRYPTO capability. In that case, add `|| defined(OPENSSL_PPC64BE)`
 // to `#if defined(OPENSSL_PPC64LE)` wherever it occurs.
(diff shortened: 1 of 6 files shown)


WillChilds-Klein (approved):

crypto/fipsmodule/bcm.c
```diff
@@ -408,21 +401,18 @@ int BORINGSSL_integrity_test(void) {
 
 void AWS_LC_FIPS_failure(const char* message) {
 #if defined(AWSLC_FIPS_FAILURE_CALLBACK)
-  if (AWS_LC_fips_failure_callback == NULL) {
-    fprintf(stderr, "AWS_LC_fips_failure_callback not defined but AWS-LC built with AWSLC_FIPS_FAILURE_CALLBACK. FIPS failure:\n%s", message);
-    fflush(stderr);
-    abort();
-  } else {
+  if (AWS_LC_fips_failure_callback != NULL) {
```
justsmth, line 404: We should update the [documentation here](https://github.com/aws/aws-lc/blob/dc3b44be65ac29fc1999ab2aad2d06c64f7525b6/crypto/internal.h#L1273) to indicate that a failure (even in FIPS mode) might be ignored by the callback. I.e., the failure doesn't necessarily "prevent any further cryptographic operations by the current process".

tests/ci/run_fips_callback_tests.sh
nebeid, line 1: It's not clear to me which path in the test is for the case when the build flag AWSLC_FIPS_FAILURE_CALLBACK is set but the callback is undefined.

crypto/fipsmodule/bcm.c
andrewhop: I thought about it, I was leaving it implied because to build AWS-LC in FIPS mode you can't specify the flag to turn the callback on. But I can add something explaining this behavior.

tests/ci/run_fips_callback_tests.sh
andrewhop: The callback is only defined in crypto_test, so ssl_test, test_fips, and bssl don't have the callback defined. This is why I was only running crypto_test before: the other tests/executables didn't have the callback defined and if you tried to call them they would abort because the callback was previosuly required.

nebeid (approved):

justsmth (approved):

PR #2273, opened 2025-03-15
Add support for more SSL BIO functions

samuel40791765:
### Issues:
Resolves `CryptoAlg-2926`

### Description of changes: 
Xtrabackup happens to take a dependency on some of OpenSSL's BIO_ssl methods. These are essentially helper BIOs that maintain an `SSL` within them. We have the necessary functionality available, this is just wrapping these BIOs around them.

### Call-outs:
N/A

### Testing:
Ideally we would test against the "connect" BIO created within `BIO_new_ssl_connect` with `BIO_do_connect`, but this isn't quite easy since we do not have any BIO methods that set up sockets on the server end (`BIO_s_accept`). We have other mechanisms of doing so in our `bssl` tool and `ocsp_integration_tests`, but pulling the functionality over just to test these `BIO`s seemed a bit overkill for my liking. I've given my reasoning in the test comments, if we ever do support `BIO_s_accept` we can look to update the test.

### DIFF
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -5112,6 +5112,34 @@ OPENSSL_EXPORT int SSL_CTX_sess_timeouts(const SSL_CTX *ctx);
 OPENSSL_EXPORT int SSL_CTX_sess_cache_full(const SSL_CTX *ctx);
 
 
+// SSL BIO methods
+
+// BIO_f_ssl returns a |BIO_METHOD| that can wrap an |SSL*| in a |BIO*|. Note
+// that this has quite different behaviour from the version in OpenSSL (notably
+// that it doesn't try to auto renegotiate).
+//
+// IMPORTANT: if you are not curl, don't use this.
+OPENSSL_EXPORT const BIO_METHOD *BIO_f_ssl(void);
+
+// BIO_set_ssl sets |ssl| as the underlying connection for |bio|, which must
+// have been created using |BIO_f_ssl|. If |take_owership| is true, |bio| will
+// call |SSL_free| on |ssl| when closed. It returns one on success or something
+// other than one on error.
+OPENSSL_EXPORT long BIO_set_ssl(BIO *bio, SSL *ssl, int take_owership);
+
+// BIO_get_ssl assigns the internal |SSL| of |bio| to |*ssl|. |*ssl| should
+// not be freed. It returns one on success or something other than one on error.
+OPENSSL_EXPORT long BIO_get_ssl(BIO *bio, SSL **ssl);
+
+// BIO_new_ssl_connect uses |ctx| to return a newly allocated BIO chain with
+// |BIO_new_ssl|, followed by a connect BIO.
+OPENSSL_EXPORT BIO *BIO_new_ssl_connect(SSL_CTX *ctx);
+
+// BIO_new_ssl returns a newly allocated SSL BIO created with |ctx|. A client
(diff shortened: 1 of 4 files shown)


include/openssl/ssl.h
```diff
@@ -5112,6 +5112,34 @@ OPENSSL_EXPORT int SSL_CTX_sess_timeouts(const SSL_CTX *ctx);
...
+// not be freed. It returns one on success or something other than one on error.
+OPENSSL_EXPORT long BIO_get_ssl(BIO *bio, SSL **ssl);
+
+// BIO_new_ssl_connect uses |ctx| to return a newly allocated BIO chain with
+// |BIO_new_ssl|, followed by a connect BIO.
+OPENSSL_EXPORT BIO *BIO_new_ssl_connect(SSL_CTX *ctx);
+
+// BIO_new_ssl returns a newly allocated SSL BIO created with |ctx|. A client
+// SSL is created if |client| is non-zero, and a server is created if otherwise.
+OPENSSL_EXPORT BIO *BIO_new_ssl(SSL_CTX *ctx, int client);
```
skmcgrail, line 5140: Since this uses `BIO_f_ssl` under the hood should it have the same/similar warning as the documentation above for `BIO_f_ssl`?

include/openssl/ssl.h
```diff
@@ -5112,6 +5112,34 @@ OPENSSL_EXPORT int SSL_CTX_sess_timeouts(const SSL_CTX *ctx);
...
 
 
+// SSL BIO methods
+
+// BIO_f_ssl returns a |BIO_METHOD| that can wrap an |SSL*| in a |BIO*|. Note
+// that this has quite different behaviour from the version in OpenSSL (notably
+// that it doesn't try to auto renegotiate).
+//
+// IMPORTANT: if you are not curl, don't use this.
+OPENSSL_EXPORT const BIO_METHOD *BIO_f_ssl(void);
```
smittals2, line 5122: Why is this warning here? As we do these shim projects, what is the impact to projects using this function (since we won't know , it'll just compile)?

samuel40791765: Good point, `BIO_f_ssl` seems to be a simplified version of the original, according to this commit: https://github.com/aws/aws-lc/commit/f5b30cc28c685b7744c3752a2a78cb920eae7ced

I think we should remove the comment, but document the discrepancies. I'll update this to do so.

include/openssl/ssl.h
samuel40791765: Yeah we should. I'll document the discrepancies like I mentioned above.

smittals2 (approved):

ssl/bio_ssl.cc
```diff
@@ -190,3 +198,45 @@ const BIO_METHOD *BIO_f_ssl(void) { return &ssl_method; }
...
+  bssl::UniquePtr<BIO> ret(BIO_new(BIO_f_ssl()));
+  SSL *ssl = SSL_new(ctx);
+
+  if (!ret || !ssl) {
+    return nullptr;
+  }
+  if (client) {
+    SSL_set_connect_state(ssl);
+  }
+  else {
```
skmcgrail, line 232:
```suggestion
  } else {
```

skmcgrail (dismissed):

skmcgrail (approved):

smittals2 (approved):

PR #2276, opened 2025-03-17
Import Jitter Entropy v3.6.3 using new method and switch code to using it

torben-hansen:
### Description of changes: 

This PR does three things: 1) Starts using the new Jitter Entropy import method 2) Imports Jitter Entropy v3.6.3 3) switch to using the new imported code.

Code changes at the AWS-LC level are based on https://github.com/aws/aws-lc/pull/2220.

Jitter Entropy imported source code is now in under `/third-party/jitterentropy/jitterentropy-library/`. Meta data on the import is in `/third_party/jitterentropy/META.yml`.

### DIFF
--- a/crypto/fipsmodule/rand/cpu_jitter_test.cc
+++ b/crypto/fipsmodule/rand/cpu_jitter_test.cc
@@ -59,7 +59,7 @@ TEST(CPUJitterEntropyTest, Basic) {
   EXPECT_EQ(jent_read_entropy(jitter_ec.instance,
                               (char*) data1, data_len), data_len);
 
-  // Verify that the Jitter Entropy library version is v3.6.2.
-  unsigned int jitter_version = 3060200;
+  // Verify that the Jitter Entropy library version is as expected.
+  unsigned int jitter_version = 3060300;
   EXPECT_EQ(jitter_version, jent_version());
 }


third_party/jitterentropy/CMakeLists.txt
```diff
@@ -34,7 +34,13 @@ else()
     endif()
 endif()
 
+if(BORINGSSL_PREFIX)
+  set(JITTER_ENETROPY_PREFIX_INCLUDE "--include=${PROJECT_BINARY_DIR}/symbol_prefix_include/openssl/boringssl_prefix_symbols.h")
```
justsmth, line 38: I believe this should be `boringssl_prefix_symbols_asm.h`. You might also need to ensure that `${PROJECT_BINARY_DIR}/symbol_prefix_include/` is one of the include search paths.

third_party/jitterentropy/META.yml
```diff
@@ -0,0 +1,5 @@
+name: jitterentropy-library
```
justsmth, line 1: Could we just make jitterentropy a git submodule?

crypto/fipsmodule/rand/cpu_jitter_test.cc
```diff
@@ -60,6 +60,6 @@ TEST(CPUJitterEntropyTest, Basic) {
                               (char*) data1, data_len), data_len);
 
   // Verify that the Jitter library version is v3.4.0.
-  unsigned int jitter_version = 3040000;
+  unsigned int jitter_version = 3060200;
```
justsmth, line 63: Update the comment as well.

third_party/jitterentropy/CMakeLists.txt
justsmth: Nevermind -- `boringssl_prefix_symbols_asm.h` is only needed for processing assembly files.

third_party/jitterentropy/META.yml
torben-hansen: yes, we could. But we do not use it anywhere else. It would be a strictly bigger behaviour change. Probably something we should do a cross the whole library at the same time.

crypto/fipsmodule/rand/cpu_jitter_test.cc
torben-hansen: ahh right. done.

justsmth (dismissed):

smittals2 (approved):

justsmth (approved):

PR #2279, opened 2025-03-19
Add more build options to match callback build

andrewhop:
### Resolves
CryptoAlg-2437

### Description of changes: 
Mirror more of the build options to make sure we catch any build issues early in GitHub. Adding the release with debug issue raised an issue with break-kat.go failing with: `Test input value was still found after erasing it. Second copy?`. This happens because the function which contains the test vectors are inlined into two functions which results in two copies of the data. This only appears to happen with GCC 7.3.1. This change fixes that by marking all of the tests as no inline. 

### DIFF
--- a/tests/ci/run_fips_callback_tests.sh
+++ b/tests/ci/run_fips_callback_tests.sh
@@ -71,7 +71,7 @@ function run_all_break_tests() {
 
 echo "Testing AWS-LC static breakable build with custom callback and Jitter enabled"
 build_and_test -DCMAKE_BUILD_TYPE=Release \
-  -DCMAKE_BUILD_SHARED_LIBS=OFF \
+  -DBUILD_SHARED_LIBS=OFF \
   -DFIPS=1 \
   -DCMAKE_INSTALL_LIBDIR=lib \
   -DCMAKE_INSTALL_INCLUDEDIR=include \
@@ -87,7 +87,7 @@ run_all_break_tests
 
 echo "Testing AWS-LC static build with custom callback and Jitter enabled"
 build_and_test -DCMAKE_BUILD_TYPE=Release \
-  -DCMAKE_BUILD_SHARED_LIBS=OFF \
+  -DBUILD_SHARED_LIBS=OFF \
   -DFIPS=1 \
   -DCMAKE_INSTALL_LIBDIR=lib \
   -DCMAKE_INSTALL_INCLUDEDIR=include \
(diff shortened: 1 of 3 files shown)


tests/ci/run_fips_callback_tests.sh
```diff
@@ -70,19 +70,31 @@ function run_all_break_tests() {
 }
 
 echo "Testing AWS-LC static breakable build with custom callback and Jitter enabled"
-build_and_test -DFIPS=1 \
-  -DCMAKE_C_FLAGS="-DBORINGSSL_FIPS_BREAK_TESTS -DAWSLC_FIPS_FAILURE_CALLBACK" \
+build_and_test -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_BUILD_SHARED_LIBS=OFF \
+  -DFIPS=1 \
```
justsmth, line 75: NP: This test can only be run on Linux (i.e., targets for which we have a FIPS static build). 

Do we need to test targets for which we produce a FIPS shared build?

tests/ci/run_fips_callback_tests.sh
```diff
@@ -70,19 +70,31 @@ function run_all_break_tests() {
...
 echo "Testing AWS-LC static build with custom callback and Jitter enabled"
-build_and_test -DFIPS=1 \
-  -DCMAKE_C_FLAGS="-DAWSLC_FIPS_FAILURE_CALLBACK" \
+build_and_test -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_BUILD_SHARED_LIBS=OFF \
+  -DFIPS=1 \
+  -DCMAKE_INSTALL_LIBDIR=lib \
+  -DCMAKE_INSTALL_INCLUDEDIR=include \
+  -DCMAKE_INSTALL_BINDIR=bin \
+  -DCMAKE_C_FLAGS="-ggdb -DAWSLC_FIPS_FAILURE_CALLBACK" \
```
justsmth, line 95: Since you're enabling debug should the `CMAKE_BUILD_TYPE` instead be "RelWithDebInfo"?

tests/ci/run_fips_callback_tests.sh
WillChilds-Klein: > Do we need to test targets for which we produce a FIPS shared build?

ACCP always links AWS-LC statically, so not as far as ACCP is concerned.

tests/ci/run_fips_callback_tests.sh
andrewhop: I was following our internal pattern where regardless of the build type we pass in `-ggdb`

skmcgrail (approved):

PR #2284, opened 2025-03-20
Add req to OpenSSL CLI tool

smittals2:
### Issues:
`CryptoAlg-2992`

### Description of changes: 
Add the req tool. Only some options are supported. We don't support the default config file interface that OpenSSL does, default values are hardcoded where appropriate. 

This PR also modified the createTempDirPath utility function. This func previously had a race condition causing intermittent CI failures. 

### Testing: 
Unit test for parsing subject function. OpenSSL comparison tests for CSR and Cert generation with cross-checking of attributes. 


crypto/test/test_util.cc
```diff
@@ -166,27 +166,29 @@ size_t createTempFILEpath(char buffer[PATH_MAX]) {
...
     return 0;
   }
 
-  // Generate a unique name using Windows API
-  if (0 == GetTempFileNameA(pathname, "awslctestdir", 0, tempdir)) {
+  // Create the path string, use process id and time for uniqueness
+  int written = snprintf(buffer, PATH_MAX, "%s\\awslctest_%lu_%llu",
+                        temp_path,
+                        GetCurrentProcessId(),
+                        (unsigned long long)time(NULL));
```
justsmth, line 180: The granularity of `time` is [seconds since the Epoch](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/time-time32-time64?view=msvc-170), so isn't this is likely to conflict between threads in the same process?

Could we just call `RAND_bytes`?

tool-openssl/req.cc
```diff
@@ -0,0 +1,666 @@
...
+  if (*subject_name_ptr++ != '/') {
+    fprintf(stderr, "name is expected to be in the format "
+                    "/type0=value0/type1=value1/type2=... where characters may "
+                    "be escaped by \\. This name is not in that format: '%s'\n",
+                   --subject_name_ptr);
+    return nullptr;
+  }
+
+  // Create new X509_NAME
+  X509_NAME* name = X509_NAME_new();
```
justsmth, line 110: Using `bssl::UniquePtr` would simplify management of these pointers.

tool-openssl/req.cc
```diff
@@ -0,0 +1,666 @@
...
+
+    // Skip '='
+    subject_name_ptr++;
+
+    // Parse value
+    while (*subject_name_ptr && *subject_name_ptr != '/') {
+      if (*subject_name_ptr == '\\' && *(subject_name_ptr + 1)) {
+        // Handle escaped character
+        subject_name_ptr++;
+        value += *subject_name_ptr++;
```
justsmth, line 142: Isn't this just adding the escaped character as if it was not escaped? That is "\nhello" would parse as "nhello" not "hello"? What's expected?

tool-openssl/req.cc
```diff
@@ -0,0 +1,666 @@
...
+      } else {
+        fprintf(stdout, "%s: ", attr.short_desc);
+      }
+      fflush(stdout);
+
+      // Get input with fgets
+      if (fgets(buffer, sizeof(buffer), stdin) == NULL) {
+        fprintf(stderr, "Error reading input\n");
+        return NULL;
+      }
```
justsmth, line 297: If someone were to paste multi-line text into this, might there be more than one line feed in the data? Should we instead start at the beginning and stop at the first line feed if present?

tool-openssl/req.cc
```diff
@@ -0,0 +1,666 @@
...
+// RSA key with 2048 bit length is used by default is |keyspec| is not valid.
+static EVP_PKEY *generate_key(const char *keyspec) {
+    EVP_PKEY_CTX *ctx = NULL;
+    EVP_PKEY *pkey = NULL;
+    long keylen = DEFAULT_KEY_LENGTH;
+    int pkey_type = EVP_PKEY_RSA;
+
+    // Parse keyspec
+    if (OPENSSL_strncasecmp(keyspec, "rsa:", 4) == 0) {
+      keylen = atol(keyspec + 4);
```
justsmth, line 67: `strtol` is more robust.

tool-openssl/req.cc
```diff
@@ -0,0 +1,666 @@
...
+    } else {
+      fprintf(stderr, "Unknown key specification: %s, using RSA key with 2048 bit length\n", keyspec);
+    }
+
+
+    // Validate key length
+    if (keylen < MIN_KEY_LENGTH) {
+        fprintf(stderr, "Key length too short (minimum %d bits)\n", MIN_KEY_LENGTH);
+        return NULL;
+    }
```
justsmth, line 79: Should there be a MAX_KEY_LEN.  (You might be waiting all day for `keylen = 16000`.)

tool-openssl/req.cc
```diff
@@ -0,0 +1,666 @@
...
+      fflush(stdout);
+
+      // Get input with fgets
+      if (fgets(buffer, sizeof(buffer), stdin) == NULL) {
+        fprintf(stderr, "Error reading input\n");
+        return NULL;
+      }
+
+      // Remove newline character if present
+      size_t len = strlen(buffer);
```
justsmth, line 300: Here and other places, use `strnlen`.

```
size_t len = strnlen(sizeof(buffer), buffer);
```

tool-openssl/req.cc
```diff
@@ -0,0 +1,666 @@
...
+      return 0;
+    }
+
+    // Remove trailing newline
+    int verify_len = strlen(verify_buf);
+    if (verify_len > 0 && verify_buf[verify_len-1] == '\n')
+      verify_buf[--verify_len] = '\0';
+
+    // Check if passwords match
+    if (strcmp(buf, verify_buf) != 0) {
```
justsmth, line 414: Unless one of the parameters is a literal, instead use `strncmp` to limit how far into the buffer you read:
```
if (strcmp(buf, verify_buf, BUF_SIZE) != 0) {
```

tool-openssl/req.cc
```diff
@@ -0,0 +1,666 @@
...
+  if (!X509_REQ_set_pubkey(req, pkey)) {
+    return 0;
+  }
+
+  return 1;
+}
+
+static int req_password_callback(char *buf, int size, int rwflag, void *userdata) {
+  const char *prompt = "Enter PEM pass phrase:";
+  char verify_buf[1024];
```
justsmth, line 379: Since it's used for a few buffers, I think there should be a `BUF_SIZE` constant.

crypto/test/test_util.cc
smittals2: You are correct, this could still be hitting race conditions - changed to Rand_bytes

tool-openssl/req.cc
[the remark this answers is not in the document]
smittals2: made it a uniqptr

tool-openssl/req.cc
```diff
@@ -0,0 +1,666 @@
...
+      if (value && value[0]) {
+        X509_ATTRIBUTE* x509_attr = X509_ATTRIBUTE_create_by_NID(nullptr,
+                                                               attr.nid,
+                                                               MBSTRING_ASC,
+                                                               reinterpret_cast<const unsigned char*>(value),
+                                                               -1);
+        if (!x509_attr) {
+          fprintf(stderr, "Error creating attribute %s\n", attr.field_name);
+          return NULL;
+        }
```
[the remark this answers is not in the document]
smittals2: made it a uniqptr

tool-openssl/req.cc
smittals2: openssl doesn't seem to restrict this in their tool, I can add a check to at least make sure nothing beyond 16000 is picked.

tool-openssl/req.cc
smittals2: So if you wanted to escape \n, you would say "\\\nhello", then we would skip over the first two "\\" and provide "\nhello" as the name. 

You can see an example of what escaping looks like here: https://github.com/aws/aws-lc/pull/2284/files#diff-d581491b3de79e020c39ed0c9e09ff4c8babd8a473966bf39103f1736418c5abR375

tool-openssl/req.cc
smittals2: I believe fgets only reads up until the first new line by default. 
https://www.ibm.com/docs/en/i/7.4?topic=functions-fgets-read-string#:~:text=The%20fgets(),string%20is%20empty.


A user shouldn't be giving multiline data for any of these fields. Should they do so, stdin buffer would become misaligned and the next fgets would read in wrong data.

crypto/test/test_util.cc
```diff
@@ -166,27 +167,33 @@ size_t createTempFILEpath(char buffer[PATH_MAX]) {
...
+  if (!RAND_bytes(random_bytes, sizeof(random_bytes))) {
     return 0;
   }
 
-  // Delete the file that GetTempFileNameA created
-  DeleteFileA(tempdir);
+  int written = snprintf(buffer, PATH_MAX, "%s\\awslctest_%02x%02x%02x%02x%02x%02x%02x%02x",
+                        temp_path,
+                        random_bytes[0], random_bytes[1], random_bytes[2], random_bytes[3],
+                        random_bytes[4], random_bytes[5], random_bytes[6], random_bytes[7]);
```
justsmth, line 185: NP: It would be cleaner to declare `random_bytes` as a `uint64_t`:
```
  uint64_t random_bytes[8]; // 64 bits of randomness is sufficient
...
  if (!RAND_bytes(&random_bytes, sizeof(random_bytes))) {
    return 0;
  }

  int written = snprintf(buffer, PATH_MAX, "%s\\awslctest_%" , temp_path, random_bytes);
```

tool-openssl/req.cc
justsmth: NP: I was assuming the input was simply "\nhello". 

This condition drops the first escape `\` character but always keeps whatever the following character is regardless... unless it's at the end of the string, in which case it keeps the escape character.

Is that the intended behavior?

tool-openssl/req.cc
```diff
@@ -0,0 +1,675 @@
...
+    // Add entry to the name
+    if (!X509_NAME_add_entry_by_NID(name.get(), nid, DEFAULT_CHAR_TYPE,
+                                   (unsigned char*)value.c_str(),
+                                   -1, -1, 0)) {
+      OPENSSL_PUT_ERROR(X509, ERR_R_X509_LIB);
+      return nullptr;
+    }
+  }
+
+  return name.release();
```
justsmth, line 188: You could return the smart pointer and use it throughout.

tool-openssl/req.cc
```diff
@@ -0,0 +1,675 @@
...
+      buffer[len - 1] = '\0';
+      len--;
+    }
+
+    // Process input
+    const char* value = nullptr;
+    if (strncmp(buffer, ".", sizeof(buffer)) == 0) {
+      // Empty entry requested
+      value = "";
+    } else if (len == 0 && field.default_value) {
```
justsmth, line 265: Is it possible to leave every field empty? Do we need to do any input validation to avoid creating invalid requests?

tool-openssl/req.cc
```diff
@@ -0,0 +1,675 @@
...
+  // Extra attributes for CSR
+  const struct {
+    const char* field_name;
+    const char* short_desc;
+    const char* default_value;
+    int nid;
+  } extra_attributes[] = {
+    {"challengePassword", "A challenge password", "", NID_pkcs9_challengePassword},
+    {"unstructuredName", "An optional company name", "", NID_pkcs9_unstructuredName}
+  };
```
justsmth, line 217: NP: Since this struct is used twice, we could give it a name and declare it outside the function.

tool-openssl/req.cc
```diff
@@ -0,0 +1,675 @@
...
+      } else if (len == 0 && attr.default_value) {
+        // Use default value
+        value = attr.default_value;
+      } else if (len > 0) {
+        // Use provided input
+        value = buffer;
+      } else {
+        // Empty input and no default - use empty string
+        value = "";
+      }
```
justsmth, line 329: Except for the variable name (`attr` versus `field`), this logic is identical to what's above. We should split this out into its own function.

tool-openssl/req.cc
```diff
@@ -0,0 +1,675 @@
...
+#include <openssl/evp.h>
+#include <openssl/err.h>
+#include <openssl/bio.h>
+#include <openssl/rsa.h>
+#include <string.h>
+#include <stdio.h>
+
+#define DEFAULT_KEY_LENGTH 2048
+#define MIN_KEY_LENGTH 512
+#define MAX_KEY_LENGTH 16000
```
justsmth, line 17: NP: I feel like ` 16384` (2**14) would be a better "round" number.

tool-openssl/req.cc
smittals2: Yes I believe this is intended. 

If the input string contains a new line character (ASCII 10), then the parsing logic would not see a backslash and this condition would not be triggered. 

But if you pass in \nhello as a string literal, then you must escape the provided backslash with another backslash to indicate "\n" is intended not just "n". Then this code would correctly parse it. 

Also since this is for fields in the subject, users should technically not be passing in new line symbols.

tool-openssl/req.cc
smittals2: Ah good call out, OpenSSL did this validation in a deeper layer of their req tooling (they handle reading from conf/user in a different part of the application). 

We should be verifying there is at least one subject field added should the user ask for every field to be empty. I have added this additional check.

justsmth (approved):

tool-openssl/req.cc
```diff
@@ -0,0 +1,637 @@
...
+
+  // Add extensions from config to the certificate
+  bool result = X509V3_EXT_add_nconf(conf.get(), &ctx, "v3_ca", cert) != 0;
+
+  return result;
+}
+
+// Generate a random serial number for a certificate
+static bool generate_serial(X509 *cert) {
+  // Create a new BIGNUM to hold the random value
```
samuel40791765, line 446: The comments here feel a bit redundant. The great function name and what BN is doing below are quite self explanatory.

tool-openssl/req.cc
```diff
@@ -0,0 +1,637 @@
...
+  }
+
+  // Generate random number with 159 bits (same as OpenSSL's SERIAL_RAND_BITS)
+  constexpr int SERIAL_RAND_BITS = 159;
+  if (!BN_rand(bn.get(), SERIAL_RAND_BITS, BN_RAND_TOP_ANY, BN_RAND_BOTTOM_ANY)) {
+    fprintf(stderr, "Failed to generate random serial number\n");
+    return false;
+  }
+
+  // Get a pointer to the certificate's ASN1_INTEGER serial number field
```
samuel40791765, line 460: Also a bit redundant.

tool-openssl/req.cc
```diff
@@ -0,0 +1,637 @@
...
+    fprintf(stderr, "Verifying - %s", prompt);
+    fflush(stderr);
+
+    if (fgets(verify_buf, sizeof(verify_buf), stdin) == NULL) {
+      fprintf(stderr, "Error reading verification password\n");
+      return 0;
+    }
+
+    // Remove trailing newline
+    int verify_len = strnlen(verify_buf, sizeof(verify_buf));
```
samuel40791765, line 387:
```suggestion
    int verify_len = OPENSSL_strnlen(verify_buf, sizeof(verify_buf));
```

tool-openssl/req.cc
```diff
@@ -0,0 +1,637 @@
...
+  fflush(stderr);
+
+  // Get password
+  if (fgets(buf, size, stdin) == NULL) {
+    fprintf(stderr, "Error reading password\n");
+    return 0;
+  }
+
+  // Remove trailing newline
+  len = strnlen(buf, sizeof(buf));
```
samuel40791765, line 371:
```suggestion
  int len = OPENSSL_strnlen(buf, sizeof(buf));
```

tool-openssl/req.cc
```diff
@@ -0,0 +1,637 @@
...
+  if (len > 0 && buf[len-1] == '\n')
+    buf[--len] = '\0';
```
samuel40791765, line 373: Nit:
```suggestion
  if (len > 0 && buf[len-1] == '\n') {
    buf[--len] = '\0';
  }
```

tool-openssl/req.cc
```diff
@@ -0,0 +1,637 @@
...
+  bssl::UniquePtr<EVP_PKEY> pkey(generate_key(keyspec.c_str()));
+  if (!pkey) {
+    fprintf(stderr, "Error: Failed to generate private key.\n");
+    return false;
+  }
+
+  // Generate and write private key
+  EVP_CIPHER *cipher = NULL;
+  if (!nodes) {
+	  cipher = (EVP_CIPHER*)EVP_des_ede3_cbc();
```
samuel40791765, line 524: Can we just directly use `const EVP_CIPHER` here to avoid the extra cast? `PEM_write_bio_PrivateKey` also seem to take a `const EVP_CIPHER` below.

tool-openssl/req.cc
```diff
@@ -0,0 +1,637 @@
...
+
+    // Set expiration to be 'days' days from now
+    if (!X509_gmtime_adj(X509_getm_notBefore(cert.get()), 0)) {
+      fprintf(stderr, "Failed to set notBefore field\n");
+      return false;
+    }
+    if (!X509_time_adj_ex(X509_getm_notAfter(cert.get()), days, 0, NULL)) {
+      fprintf(stderr, "Failed to set notAfter field\n");
+      return false;
+    }
```
samuel40791765, line 589: Just curious, what's the reason for using `X509_time_adj` first, then `X509_time_adj_ex` the second time?

tool-openssl/req.cc
```diff
@@ -0,0 +1,637 @@
...
+  fflush(stdout);
+
+  // Get input with fgets
+  if (fgets(buffer, buffer_size, stdin) == NULL) {
+    fprintf(stderr, "Error reading input\n");
+    return NULL;
+  }
+
+  // Remove newline character if present
+  size_t len = strnlen(buffer, buffer_size);
```
samuel40791765, line 211:
```suggestion
  size_t len = OPENSSL_strnlen(buffer, buffer_size);
```

tool-openssl/req.cc
```diff
@@ -0,0 +1,637 @@
...
+      type += *subject_name_ptr++;
+    }
+
+    if (!*subject_name_ptr) {
+      fprintf(stderr, "Hit end of string before finding the equals.\n");
+      return nullptr;
+    }
+
+    // Skip '='
+    subject_name_ptr++;
```
samuel40791765, line 144: Nice comment. Do we need to check if "=" is actually there if there's a formatting requirement for passed in `subject_string`s? If OpenSSL is also doing this, I think this is fine as is though.

samuel40791765 (commented):
Some of the spacing's a bit off in the new files. Would recommend just running `clang-format` to fix that.

tool-openssl/req_test.cc
```diff
@@ -0,0 +1,438 @@
...
+      std::string result = "";
+      if (len > 0) {
+        result.assign(reinterpret_cast<char*>(tmp), len);
+      }
+      OPENSSL_free(tmp);
+      return result;
+    }
+};
+
+void PrintTo(const SubjectNameTestCase& test_case, std::ostream* os) {
```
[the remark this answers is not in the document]
smittals2: It's a GTest framework thing, when a test fails, GTest will call the PrintTo function. Easier debugging and I was getting a weird valgrind issue with my parameterized test (likely due to std:vector<std:string>) and this fixed it.

tool-openssl/req_test.cc
```diff
@@ -0,0 +1,438 @@
...
+        expect_success(expect_success_),
+        expected_entry_count(expected_entry_count_),
+        expected_values(expected_values_) {}
+};
+
+void PrintTo(const SubjectNameTestCase& test_case, std::ostream* os);
+class SubjectNameTest : public testing::TestWithParam<SubjectNameTestCase> {
+  protected:
+    static std::string GetEntryValue(X509_NAME* name, int index) {
+      unsigned char* tmp;
```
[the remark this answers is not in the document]
smittals2: I believe UniquePtr<char> doesn't work directly with ASN1_STRING_to_UTF8() because the function requires a pointer-to-pointer (unsigned char**) where it will store a newly allocated memory address, not a buffer we pre-allocate.

tool-openssl/req.cc
smittals2: The while loop above this (where we parse type) iterates until we hit a '=', so we know at this line we are skipping over the '='.

tool-openssl/req.cc
smittals2: X509_gmtime_adj with 0 seconds for notBefore → Sets current time as start time
X509_time_adj_ex with days days for notAfter → Sets expiration days in the future

X509_time_adj_ex essentially allows us to specify time in days instead of having to calculate seconds with the non _ex variant. Looking at the comments, I see that X509_gmtime_adj actually calls X509_time_adj_ex but just with a 0 days offset.

crypto/dsa/internal.h
```diff
@@ -40,8 +40,6 @@ struct dsa_st {
   CRYPTO_EX_DATA ex_data;
 };
 
-#define OPENSSL_DSA_MAX_MODULUS_BITS 10000
```
samuel40791765, line 43: These changes seem to have been leaked over from https://github.com/aws/aws-lc/commit/0c9733744447c7acb2eff0f1631f2cec720e0079

smittals2: ah yes sorry, I was addressing a comment from the PR for that commit. I merged it in without the change so I could get it in for the release. 

I'll update the commit message here to reflect this change as well. 

Comment on PR: https://github.com/aws/aws-lc/pull/2293#discussion_r2017773098

tool-openssl/req.cc
```diff
@@ -0,0 +1,673 @@
...
+  // Process each subject field
+  for (const auto &field : subject_fields) {
+    const char *value = prompt_field(field, buffer, sizeof(buffer));
+
+    if (value == NULL) {
+      return NULL;
+    }
+
+    // Only add non-empty values
+    if (value && value[0]) {
```
justsmth, line 307: NP: The second `if` condition and be simplified:
```
if (value[0]) {
...
```
... and (arguably?) clarified:
```
if (0 < OPENSSL_strnlen(value, BUF_SIZE)) {
...
```

tool-openssl/req.cc
```diff
@@ -0,0 +1,673 @@
...
+      {"commonName", "Common Name (e.g. server FQDN or YOUR name)", "",
+       NID_commonName},
+      {"emailAddress", "Email Address", "", NID_pkcs9_emailAddress}};
+
+  // Extra attributes for CSR
+  const ReqField extra_attributes[] = {
+      {"challengePassword", "A challenge password", "",
+       NID_pkcs9_challengePassword},
+      {"unstructuredName", "An optional company name", "",
+       NID_pkcs9_unstructuredName}};
```
justsmth, line 273: NP: The function would flow better if these constants were moved to be right before the function.

tool-openssl/req.cc
```diff
@@ -0,0 +1,673 @@
...
+          "incorporated\n");
+  fprintf(stdout, "into your certificate request.\n");
+  fprintf(stdout,
+          "What you are about to enter is what is called a Distinguished Name "
+          "or a DN.\n");
+  fprintf(stdout,
+          "There are quite a few fields but you can leave some blank\n");
+  fprintf(stdout, "For some fields there will be a default value,\n");
+  fprintf(stdout, "If you enter '.', the field will be left blank.\n");
+  fprintf(stdout, "\n");
```
justsmth, line 294: NP: To reduce noise in this function, could these instructions be moved up to `make_certificate_request`, or possibly to its own function?

tool-openssl/req.cc
```diff
@@ -0,0 +1,673 @@
...
+    // Process each extra attribute
+    for (const auto &attr : extra_attributes) {
+      const char *value = prompt_field(attr, buffer, sizeof(buffer));
+
+      if (value == NULL) {
+        return NULL;
+      }
+
+      // Only add non-empty attributes
+      if (value && value[0]) {
```
justsmth, line 335: NP: See comment above.

justsmth (approved):

include/openssl/dsa.h
```diff
@@ -187,8 +189,8 @@ OPENSSL_EXPORT DSA *DSAparams_dup(const DSA *dsa);
 // Key generation.
 
 // DSA_generate_key generates a public/private key pair in |dsa|, which must
-// already have parameters setup. It returns one on success and zero on
-// error.
+// already have parameters setup. Only supports generating upto |OPENSSL_DSA_MAX_MODULUS_BITS|
```
samuel40791765, line 192:
```suggestion
// already have parameters setup. Only supports generating up to |OPENSSL_DSA_MAX_MODULUS_BITS|
```

samuel40791765 (approved):

justsmth (approved):

samuel40791765 (approved):

PR #2285, opened 2025-03-20
Implement legacy callback with BIO_set_callback

kingstjo:
### Issues:
Addresses #CRYPTOALG-3030

### Description of changes: 
Adds legacy callback to BIO meeting previous application compatibility.
Router function implemented to prefer enhanced callback but provide legacy callback if needed and available.

### Call-outs:
We now check if callback is NULL after checking if we have a callback.
Rather than calling enhanced callback directly we check with router function get_callback

### Testing:
Updated parameters for BIOPairTest to include boolean pairs enabling current enhanced callback testing to test for legacy callback.
Changed multiple tests from BIOTest to BIOPairTest for parameterized testing.
Will be adding test for get_callback in future PR

### DIFF
--- a/crypto/bio/bio.c
+++ b/crypto/bio/bio.c
@@ -248,6 +248,7 @@ void BIO_free_all(BIO *bio) {
 
 int BIO_read(BIO *bio, void *buf, int len) {
   int ret = 0;
+  size_t processed = 0;
   if (bio == NULL || bio->method == NULL || bio->method->bread == NULL) {
     OPENSSL_PUT_ERROR(BIO, BIO_R_UNSUPPORTED_METHOD);
     return -2;
@@ -256,7 +257,6 @@ int BIO_read(BIO *bio, void *buf, int len) {
     return 0;
   }
 
-  if (HAS_CALLBACK(bio)) {
     BIO_callback_fn_ex cb = get_callback(bio);
     if (cb != NULL) {
       long callback_ret = cb(bio, BIO_CB_READ, buf, len, 0, 0L, 1L, NULL);
@@ -267,7 +267,7 @@ int BIO_read(BIO *bio, void *buf, int len) {
         return INT_MIN;
       }
     }
-  }
+
   if (!bio->init) {
     OPENSSL_PUT_ERROR(BIO, BIO_R_UNINITIALIZED);
     return -2;
@@ -275,10 +275,23 @@ int BIO_read(BIO *bio, void *buf, int len) {
   ret = bio->method->bread(bio, buf, len);
   if (ret > 0) {
     bio->num_read += ret;
+    processed = ret;
+    ret = 1;
   }
 
-  ret = call_bio_callback_with_processed(bio, BIO_CB_READ | BIO_CB_RETURN, buf,
-                                         len, ret);
+  if (cb != NULL) {
+    ret = cb(bio, BIO_CB_READ | BIO_CB_RETURN, buf, len, 0, 0L,
+      ret, &processed);
+  }
+
+  if (ret <= INT_MAX && ret >= INT_MIN) {
+    if (ret > 0) {
+      // BIO will only read int |len| bytes so this is a safe cast
+      ret = (int)processed;
+    }
+  } else {
+    ret = -1;
+  }
 
   return ret;
 }
--- a/crypto/bio/bio.c
+++ b/crypto/bio/bio.c
@@ -68,7 +68,24 @@
 
 #include "../internal.h"
 
-#define HAS_CALLBACK(b) ((b)->callback_ex != NULL)
+#define HAS_CALLBACK(b) ((b)->callback_ex != NULL || (b)->callback != NULL)
+
+static long callback_fn_wrap_ex(BIO *bio, int oper, const char *argp,
+                              size_t len, int argi, long argl, int ret,
+                              size_t *processed) {
+  return bio->callback(bio, oper, argp, argi, argl, ret);
+}
+
+static BIO_callback_fn_ex get_callback(BIO *bio) {
+  if (bio->callback_ex != NULL) {
+    return bio->callback_ex;
+  }
+  else if (bio->callback != NULL) {
+    // Wrap old-style callback in extended format
+    return callback_fn_wrap_ex;
+  }
+  return NULL;  // Explicit return, though this case should never happen if called after HAS_CALLBACK
+}
 
 // Helper function to create a placeholder |processed| that the callback can
 // modify and return to the caller. Used only in callbacks that pass in
@@ -76,6 +93,11 @@
 static int call_bio_callback_with_processed(BIO *bio, const int oper,
                                         const void *buf, int len, int ret) {
   if (HAS_CALLBACK(bio)) {
+    // Preserves flag behavior for old-style callback to match callback_ex
+    if (bio->callback_ex == NULL && bio->callback != NULL) {
+      long callback_ret = bio->callback(bio, oper, buf, 0, 0L, ret);
+      return callback_ret;
+    }
     size_t processed = 0;
     // The original BIO return value can be an error value (less than 0) or
     // the number of bytes read/written
@@ -112,6 +134,7 @@ BIO *BIO_new(const BIO_METHOD *method) {
   ret->shutdown = 1;
   ret->references = 1;
   ret->callback_ex = NULL;
+  ret->callback = NULL;
   CRYPTO_new_ex_data(&ret->ex_data);
 
   if (method->create != NULL && !method->create(ret)) {
@@ -136,7 +159,8 @@ int BIO_free(BIO *bio) {
       bio->method->destroy(bio);
     }
     if (HAS_CALLBACK(bio)) {
-      long ret = bio->callback_ex(bio, BIO_CB_FREE, NULL, 0, 0, 0L, 1L, NULL);
+      BIO_callback_fn_ex cb = get_callback(bio);
+      long ret = cb(bio, BIO_CB_FREE, NULL, 0, 0, 0L, 1L, NULL);
       if (ret <= 0) {
         if (ret >= INT_MIN) {
           return (int)ret;
@@ -175,7 +199,8 @@ int BIO_read(BIO *bio, void *buf, int len) {
   }
 
   if (HAS_CALLBACK(bio)) {
-    long callback_ret = bio->callback_ex(bio, BIO_CB_READ, buf, len, 0, 0L, 1L, NULL);
+    BIO_callback_fn_ex cb = get_callback(bio);
+    long callback_ret = cb(bio, BIO_CB_READ, buf, len, 0, 0L, 1L, NULL);
     if (callback_ret <= 0) {
       if (callback_ret >= INT_MIN) {
         return (int)callback_ret;
@@ -229,7 +254,8 @@ int BIO_gets(BIO *bio, char *buf, int len) {
   }
 
   if (HAS_CALLBACK(bio)) {
-    long callback_ret = bio->callback_ex(bio, BIO_CB_GETS, buf, len, 0, 0L, 1L, NULL);
+    BIO_callback_fn_ex cb = get_callback(bio);
+    long callback_ret = cb(bio, BIO_CB_GETS, buf, len, 0, 0L, 1L, NULL);
     if (callback_ret <= 0) {
       if (callback_ret >= INT_MIN) {
         return (int)callback_ret;
@@ -261,7 +287,8 @@ int BIO_write(BIO *bio, const void *in, int inl) {
   }
 
   if (HAS_CALLBACK(bio)) {
-    long callback_ret = bio->callback_ex(bio, BIO_CB_WRITE, in, inl, 0, 0L, 1L, NULL);
+    BIO_callback_fn_ex cb = get_callback(bio);
+    long callback_ret = cb(bio, BIO_CB_WRITE, in, inl, 0, 0L, 1L, NULL);
     if (callback_ret <= 0) {
       if (callback_ret >= INT_MIN) {
         return (int)callback_ret;
@@ -334,7 +361,8 @@ int BIO_puts(BIO *bio, const char *in) {
     return -2;
   }
   if(HAS_CALLBACK(bio)) {
-    long callback_ret = bio->callback_ex(bio, BIO_CB_PUTS, in, 0, 0, 0L, 1L, NULL);
+    BIO_callback_fn_ex cb = get_callback(bio);
+    long callback_ret = cb(bio, BIO_CB_PUTS, in, 0, 0, 0L, 1L, NULL);
     if (callback_ret <= 0) {
       if (callback_ret >= INT_MIN) {
         return (int)callback_ret;
@@ -383,7 +411,8 @@ long BIO_ctrl(BIO *bio, int cmd, long larg, void *parg) {
   }
   long ret = 0;
   if (HAS_CALLBACK(bio)) {
-    ret = bio->callback_ex(bio, BIO_CB_CTRL, parg, 0, cmd, larg, 1L, NULL);
+    BIO_callback_fn_ex cb = get_callback(bio);
+    ret = cb(bio, BIO_CB_CTRL, parg, 0, cmd, larg, 1L, NULL);
     if (ret <= 0) {
       return ret;
     }
@@ -391,7 +420,8 @@ long BIO_ctrl(BIO *bio, int cmd, long larg, void *parg) {
 
   ret = bio->method->ctrl(bio, cmd, larg, parg);
   if (HAS_CALLBACK(bio)) {
-    ret = bio->callback_ex(bio, BIO_CB_CTRL | BIO_CB_RETURN, parg, 0, cmd, larg,
+    BIO_callback_fn_ex cb = get_callback(bio);
+    ret = cb(bio, BIO_CB_CTRL | BIO_CB_RETURN, parg, 0, cmd, larg,
                          ret, NULL);
   }
   return ret;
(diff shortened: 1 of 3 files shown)


crypto/bio/bio.c
```diff
@@ -136,14 +182,18 @@ int BIO_free(BIO *bio) {
...
-          return (int)ret;
+      BIO_callback_fn_ex cb = get_callback(bio);
+      if (cb != NULL) {
+        long ret = cb(bio, BIO_CB_FREE, NULL, 0, 0, 0L, 1L, NULL);
+        if (ret <= 0) {
+          if (ret >= INT_MIN) {
+            return (int)ret;
+          }
+          return INT_MIN;
         }
```
justsmth, line 193: To avoid needing to do the return value "sanity checks" everywhere we use a callback, why not have wrappers for both `bio->callback_ex` and `bio->callback` and have the sanity check embedded in the wrapper? 

(I see that `BIO_ctrl` returns a `long`, but do we ever expect the return value of the callback to be outside the range of `0..INT_MAX`?)

crypto/bio/bio.c
```diff
@@ -68,14 +68,59 @@
...
 // modify and return to the caller. Used only in callbacks that pass in
 // |processed|.
 static int call_bio_callback_with_processed(BIO *bio, const int oper,
                                         const void *buf, int len, int ret) {
   if (HAS_CALLBACK(bio)) {
+    // Preserves flag behavior for old-style callback to match callback_ex
+    if (bio->callback_ex == NULL && bio->callback != NULL) {
+      long callback_ret = bio->callback(bio, oper, buf, 0, 0L, ret);
+      return callback_ret;
+    }
```
justsmth, line 123: For consistency with the others, why not do `cb = get_callback(bio)` then use `cb` below?

kingstjo: This is only works for legacy callback and would cause issues if  `cb = get_callback(bio)` returned cb was callback_ex.
`get_callback(bio)` prefers callback_ex and although unlikely I wanted to ensure we would not make this call with enhanced callback.

crypto/bio/bio.c
kingstjo: For this I changed lines 185-187.

As lines 188-192 were brought within the context of `if (cb != NULL) {` it is marked as changed with my commit.
Thank you for the feedback on having sanity checks within the wrappers, I will look into this behavior further.

crypto/bio/bio.c
```diff
@@ -68,14 +68,59 @@
...
+//
+//   |len| and |processed| parameters from the extended interface are ignored since
+//   they are not supported by the legacy callback interface.
+//
+//   Returns -1 on NULL |BIO| or callback, otherwise returns the result of the legacy
+//   callback.
+static long callback_fn_wrap_ex(BIO *bio, int oper, const char *argp,
+                              size_t len, int argi, long argl, int ret,
+                              size_t *processed) {
+  if (bio == NULL || bio->callback == NULL) {
```
smittals2, line 85: comments for this function say "function should only be called when |callback_ex| is not available". Is that something we should check for here (maybe as a part of the if statement)? Does OpenSSL do this?

crypto/bio/bio.c
smittals2: If the contract is you can only have one defined, then cb = get_callback() will appropriately return the correct callback. 
If for whatever reason a user defines both, we want to fallback on the "enhanced" variant - which happens through this current code anyway.

crypto/bio/bio.c
```diff
@@ -68,14 +68,59 @@
...
+//   |callback_fn_wrap_ex| adapts the legacy callback interface |BIO_callback_fn| to the
+//   extended callback interface |BIO_callback_fn_ex|. This function should only be
+//   called when |callback_ex| is not available and the legacy callback is set.
+//
+//   |len| and |processed| parameters from the extended interface are ignored since
+//   they are not supported by the legacy callback interface.
+//
+//   Returns -1 on NULL |BIO| or callback, otherwise returns the result of the legacy
+//   callback.
+static long callback_fn_wrap_ex(BIO *bio, int oper, const char *argp,
```
smittals2, line 82: Looks like OpenSSL still modifies processed, len even when the legacy callback is called based on it's return value (they just don't pass in those params but the wrapper func has additional processing).

Can look at `BIO_debug_callback` as an example of what processing inside a legacy callback might look like and what the return values look like. 

They have a static func `bio_call_callback` which does the additional processing around the legacy callback. We should likely be doing something like that here. 

For both things above, make sure you're looking at OpenSSL 1.1.1 behavior.

include/openssl/bio.h
```diff
@@ -1002,6 +1010,20 @@ struct bio_st {
...
   // |BIO_CB_CTRL|+|BIO_CB_RETURN|, and |BIO_CB_FREE|.
   BIO_callback_fn_ex callback_ex;
+
+  // Legacy callback function that handles the same events as |callback_ex| but without
+  // length and processed parameters.
+  // When both callbacks are set, |callback_ex| will be used.
+  // Handles |BIO_read|, |BIO_write|, |BIO_free|, |BIO_gets|, |BIO_puts|,
+  // and |BIO_ctrl| operations.
+  // Callbacks are only called with for the following events: |BIO_CB_READ|,
+  // |BIO_CB_READ|+|BIO_CB_RETURN|, |BIO_CB_WRITE|,
```
smittals2, line 1020: This will also likely have to update once you figure out the correct legacy callback behavior around processed/len in accordance with these flags.

crypto/bio/bio.c
justsmth: > ... I wanted to ensure we would not make this call with enhanced callback.

What's the expected behavior when both callbacks have been set?  Is the expected behavior different here than in the other places?

crypto/bio/bio.c
```diff
@@ -68,14 +68,59 @@
...
+static BIO_callback_fn_ex get_callback(BIO *bio) {
+  if (bio->callback_ex != NULL) {
+    return bio->callback_ex;
+  }
+  if (bio->callback != NULL) {
+    // Wrap old-style callback in extended format
+    return callback_fn_wrap_ex;
+  }
+  return NULL;  // Explicit return, though this case should never happen if called after HAS_CALLBACK
+}
```
justsmth, line 111: Since this should never happen, we should add an `assert` so that if it occurs during testing, the tests will fail:
```
assert(false);
```

crypto/bio/bio.c
```diff
@@ -333,13 +392,16 @@ int BIO_puts(BIO *bio, const char *in) {
...
     return -2;
   }
-  if(HAS_CALLBACK(bio)) {
-    long callback_ret = bio->callback_ex(bio, BIO_CB_PUTS, in, 0, 0, 0L, 1L, NULL);
-    if (callback_ret <= 0) {
-      if (callback_ret >= INT_MIN) {
-        return (int)callback_ret;
+  if (HAS_CALLBACK(bio)) {
+    BIO_callback_fn_ex cb = get_callback(bio);
+    if (cb != NULL) {
```
justsmth, line 397: Actually, I don't see much value in the `HAS_CALLBACK` macro b/c you're still having to subsequently check `if (cb != NULL) {`.

Why not just eliminate the macro call and just have:
```
  BIO_callback_fn_ex cb = get_callback(bio);
  if (cb != NULL) {
    ...
  }
```

crypto/bio/bio.c
kingstjo: > Maybe im misunderstanding. If the contract is you can only have one defined, then cb = get_callback() will appropriately return the correct callback. If for whatever reason a user defines both, we want to fallback on the "enhanced" variant - which happens through this current code anyway. So get_callback() should suffice?

> > ... I wanted to ensure we would not make this call with enhanced callback.
> 
> What's the expected behavior when both callbacks have been set? Is the expected behavior different here than in the other places?

The expected behavior is as mentioned with get_callback().
- If both callback and callback_ex are set callback_ex is used.

Additionally, this may tie into other comment regarding wrappers for both callbacks which could handle all sanity checks, and len or processed information where needed.

crypto/bio/bio.c
justsmth: It might be better to just not use the macro and just use the return value of `get_callback` to determine what to do.

crypto/bio/bio.c
```diff
@@ -227,26 +320,44 @@ int BIO_gets(BIO *bio, char *buf, int len) {
...
+  }
+
+  if (ret <= INT_MAX && ret >= INT_MIN) {
+    if (ret > 0) {
+      // BIO will only read int |len| bytes so this is a safe cast
+      ret = (int)processed;
+    }
+  } else {
+    ret = -1;
   }
```
justsmth, line 359: * Why is is safe to assume that `0 < processed < INT_MAX`? 
  * `processed` is a `size_t`, which can hold values greater than `INT_MAX`.
  * We passed a reference `&processed` to the callback, so it could've been changed by the callback.
* Here (and at the ends of `BIO_read`, `BIO_write`, `BIO_puts`, etc.) the logic seems a bit stilted. I think it would be clearer as:
```
  if (ret > INT_MAX || ret < INT_MIN) {
    return -1;
  }
  if (ret > 0) {
    // Validation checks on `processed` then return?
    return (int) processed;
  }
  return ret;
```

crypto/bio/bio.c
```diff
@@ -382,18 +524,24 @@ long BIO_ctrl(BIO *bio, int cmd, long larg, void *parg) {
...
-    ret = bio->callback_ex(bio, BIO_CB_CTRL | BIO_CB_RETURN, parg, 0, cmd, larg,
-                         ret, NULL);
-  }
+
+    cb = get_callback(bio);
+    if (cb != NULL) {
+      ret = cb(bio, BIO_CB_CTRL | BIO_CB_RETURN, parg, 0, cmd, larg,
+                           ret, NULL);
+    }
+
```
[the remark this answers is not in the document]
kingstjo: This follows the previous path of using HAS_CALLBACK and calling bio->callback_ex again.

Between these two `get_callback` calls we pass bio to `ret = bio->method->ctrl(bio, cmd, larg, parg);` potentially changing state of bio.

crypto/bio/bio.c
kingstjo: You are correct, this is not a safe assumption.

This was taken from our previous implementation of `call_bio_callback_with_processedcall_bio_callback_with_processed`.
It is now updated to properly verify `processed` will fit before the cast.
Thank you for this call out, and providing more clear checking within INT bounds!

crypto/bio/bio.c
kingstjo: We are now handling these checks within helper function `handle_callback_return`.

crypto/bio/bio.c
```diff
@@ -68,34 +68,110 @@
...
+    } else if (oper == BIO_CB_WRITE || oper == BIO_CB_PUTS) {
+      bio->num_write += ret;
+    }
+    processed = ret;
+    ret = 1;
+  }
+
+  BIO_callback_fn_ex cb = get_callback(bio);
+  if (cb != NULL) {
+    ret = cb(bio, oper | BIO_CB_RETURN, buf, len, 0, 0L, ret, &processed);
```
justsmth, line 161: * `ret` is of type `int, but `cb` returns a `long`.
* I think this should be a separate variable and the range check pulled inside the block:
```
  if (cb != NULL) {
    long callback_ret = cb(bio, oper | BIO_CB_RETURN, buf, len, 0, 0L, ret, &processed);
    if (callback_ret > INT_MAX || callback_ret < INT_MIN) {
      return -1;
    }
    ret = (int)callback_ret;
  }
```

crypto/bio/bio.c
```diff
@@ -68,34 +68,115 @@
...
+// The extended interface parameters |len| and |processed| are mapped to the legacy
+// interface parameters |argi| and |bio_ret| respectively.
+//
+// Returns -1 on NULL |BIO| or callback, otherwise returns the result of the legacy
+// callback.
+static long callback_fn_wrap_ex(BIO *bio, int oper, const char *argp,
+                              size_t len, int argi, long argl, int bio_ret,
+                              size_t *processed) {
+  assert(bio != NULL);
+  assert(bio->callback != NULL);
```
smittals2, line 84: Should we also assert callback_ex should be NULL here?

crypto/bio/bio.c
```diff
@@ -68,34 +68,115 @@
...
-      ret = (int)callback_ret;
-      if (ret > 0) {
-        // BIO will only read int |len| bytes so this is a safe cast
-        ret = (int)processed;
-      }
+
+    argi = (int)len;
+  }
+
+  if (bio_ret && (oper & BIO_CB_RETURN) && bareoper != BIO_CB_CTRL) {
```
smittals2, line 99: Should this be bio_ret or bio_ret > 0? Could bio_ret be a negative value? What should happen then?

crypto/bio/bio.c
```diff
@@ -68,34 +68,115 @@
...
+      return -1;
+    }
+
+    bio_ret = *processed;
+  }
+
+
+  long ret = bio->callback(bio, oper, argp, argi, argl, bio_ret);
+
+  if (ret >= 0 && (oper & BIO_CB_RETURN) && bareoper != BIO_CB_CTRL) {
```
smittals2, line 110: is this check supposed to be ret >= 0 or ret > 0? Is 0 a valid return value from callback?

crypto/bio/bio.c
kingstjo: The only way we can use `callback_fn_wrap_ex` is `if (bio->callback_ex != NULL)` is false.

Additionally callback_ex is never called within this scope and could not lead to a null pointer error.

crypto/bio/bio.c
```diff
@@ -68,34 +68,115 @@
...
+    *processed = (size_t)ret;
+    ret = 1;
+  }
+
+  return ret;
+}
+
+//   |get_callback| returns the appropriate callback function for a given |BIO|, preferring
```
smittals2, line 118: nit: spacing between text and // for comments.

crypto/bio/bio.c
```diff
@@ -68,34 +68,115 @@
...
+//   When only the legacy callback is available, it is wrapped in the extended format
+//   via |callback_fn_wrap_ex| to provide a consistent interface. The extended callback
+//   provides additional parameters for length and bytes processed tracking.
+//
+//   Returns the |callback_ex| function if available, a wrapped legacy callback if only
+//   |callback| is set, or NULL if no callbacks are set.
+static BIO_callback_fn_ex get_callback(BIO *bio) {
+  if (bio == NULL) {
+    return NULL;
+  }
```
justsmth, line 130: NP: Since this is only called internally, it just could be an `assert` instead.

justsmth (dismissed):

crypto/bio/bio.c
kingstjo: `bio_ret `having a negative value would allow `BIO_CB_RETURN ` operations not including `BIO_CB_CTRL `to set `processed `to `bio_ret`.

You are correct, currently this could overwrite error values of `bio_ret `with `processed ` masking the original error condition.

crypto/bio/bio.c
smittals2: Sure sure, it's not called in the current code but future development efforts may not keep this in mind. Just good to future proof the code as well. 

Also the concern wasn't null pointer but more so the function contract says this should only be called when the ex variant is not defined, so just to be explicit we can add that assert. 

But non-critical for sure

crypto/bio/bio.c
```diff
@@ -145,6 +227,8 @@ int BIO_free(BIO *bio) {
       }
     }
 
+
```
smittals2, line 230: nit: extra new lines

crypto/bio/bio.c
```diff
@@ -343,11 +418,13 @@ int BIO_puts(BIO *bio, const char *in) {
     }
   }
 
+
```
smittals2, line 421: nit: extra new line

crypto/bio/bio.c
```diff
@@ -68,34 +68,113 @@
...
+    }
+    // |callback_ex| receives the number of bytes processed via the |processed| parameter,
+    // while the legacy callback receives this information through both |argi| and |ret|.
+    // When using the legacy callback, the |processed| value will be mapped back to |ret|.
+    processed = ret;
+    ret = 1;
+  }
+
+  BIO_callback_fn_ex cb = get_callback(bio);
+  if (cb != NULL) {
```
smittals2, line 161: it's okay to have this check here but just want to confirm, this should never actually be true inside this function right? And if it is true (that cb == null), then is the current logic appropriately returning the right value (I guess we  assign processed to ret above, and then reassign it to ret below)

crypto/bio/bio.c
```diff
@@ -68,34 +68,113 @@
...
+    long callback_ret = cb(bio, oper | BIO_CB_RETURN, buf, len, 0, 0L, ret, &processed);
+    if (ret > INT_MAX || ret < INT_MIN) {
```
smittals2, line 163: ret is an int var, these conditions will never be true?

crypto/bio/bio.c
kingstjo: (cb = NULL) when there is no callback set

Potentially they could meet underlying method req's `if (bio == NULL || bio->method == NULL || bio->method->bread == NULL)` while still having no callback set.
- In this case we return `int ret = bio->method->bread(bio, buf, len);` for applicable method
- This could be replaced by processed `if (ret > 0)`

crypto/bio/bio.c
kingstjo: You are correct!
This should be `callback_ret`

crypto/bio/bio.c
kingstjo: You are correct, this has been updated to `ret > 0`

justsmth (approved):

smittals2 (approved):

PR #2289, opened 2025-03-24
Reject DSA trailing garbage in EVP layer, add test cases

skmcgrail:
# Description
The EVP API layer was not rejecting signatures that contained trailing garbage after the `DSA_SIG` ASN.1 encoding (note it correctly reject invalid signatures regardless). This did not match the behavior of the `DSA_verify` and `DSA_check_signature` APIs which explicitly check and reject trailing garbage. This aligns the EVP layer with this behavior.


crypto/evp_extra/evp_extra_test.cc
```diff
@@ -3402,3 +3402,81 @@ TEST(EVPExtraTest, RawKeyUnsupported) {
       EVP_PKEY_new_raw_private_key(EVP_PKEY_RSA, nullptr, kKey, sizeof(kKey)));
 }
 
+TEST(EVPExtraTest, DSATrailingSignatureGarbage) {
```
nebeid, line 3405: It would be good to add comments to the test case on what fails and why.

skmcgrail: Is the `EXPECT_FALSE` and on calls to verify a test case name not sufficient. What can use clarification?

justsmth (dismissed):

crypto/evp_extra/evp_extra_test.cc
skmcgrail: Added a comment above the signature value explaining that it has extra bytes after the DER encoding.

justsmth (approved):

crypto/evp_extra/evp_extra_test.cc
nebeid: Thanks. Yes, that's right the `EXPECT_FALSE` is clear. The return values of `DSA_check_signature` and `DSA_verify` are what I needed to ensure they're indicating failures, so I checked them in the code.

nebeid (approved):

PR #2290, opened 2025-03-24
Add Libwebsockets to our CI

smittals2:
### Issues:
`CryptoAlg-2992`

### Description of changes: 
Add libwebsockets master to our CI along with a patch file.

The patch file simplifies many of the ifdefs in libwebsockets code that were previously repetitive and convoluted by creating helped macros/simplifying logic. I have tested this variant of libwebsockets locally with AWS-LC and OpenSSL, WolfSSL remains pending.  


tests/ci/integration/libwebsockets_patch/libwebsocket_master.patch
```diff
@@ -0,0 +1,420 @@
...
++#if defined(LWS_WITH_BORINGSSL) || defined(LWS_WITH_AWSLC)
+ 				(uint32_t)
+ #endif
+ 					error,
+@@ -238,7 +238,7 @@ lws_tls_server_certs_load(struct lws_vhost *vhost, struct lws *wsi,
+ 				const char *s;
+ 				error = ERR_get_error();
+ 				s = ERR_error_string(
+-	#if defined(LWS_WITH_BORINGSSL)
++	#if defined(LWS_WITH_BORINGSSL) || defined(LWS_WITH_AWSLC)
```
darylmartin100, line 203: It will be worth while to try and simplify this upstream for future releases.  This #ifdef could simply turn into something like the below, which will allow us to remove a lot of these #ifdefs in libwebsockets.  What we have here is more than fine, it follows the pattern the project set, but I think we should minimize the total number of #ifdefs.  

`ERR_error_string(ERR_get_error(), (char*)vhost->context->pt[0].serv_buf;)`

tests/ci/integration/run_libwebsockets_integration.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+set -ex
+
```
justsmth, line 6: NP:
```
set -o pipefail
```

justsmth (approved):

justsmth (approved):

samuel40791765 (approved):

PR #2295, opened 2025-03-28
Implemented no-op CRYPTO_mem_ctrl

kingstjo:
### Issues:
Addresses #CryptoAlg-2905

### Description of changes: 
Applications requiring CRYPTO_mem_ctrl can now build successfully. However, we define `OPENSSL_NO_CRYPTO_MDEBUG` in AWS-LC, which disables this functionality. As a result, the implemented function intentionally does nothing.

### Call-outs:
We've added a defensive compilation error that will trigger if `OPENSSL_NO_CRYPTO_MDEBUG` is removed in the future. This will allow us to fully implement the function if needed.

### Testing:
No testing is required for this intentional no-operation function. This approach is consistent with OpenSSL 1.1.1's handling of `OPENSSL_NO_CRYPTO_MDEBUG`.

### DIFF
--- a/crypto/mem.c
+++ b/crypto/mem.c
@@ -122,6 +122,16 @@ WEAK_SYMBOL_FUNC(void, OPENSSL_memory_free, (void *ptr))
 WEAK_SYMBOL_FUNC(size_t, OPENSSL_memory_get_size, (void *ptr))
 WEAK_SYMBOL_FUNC(void *, OPENSSL_memory_realloc, (void *ptr, size_t size))
 
+int CRYPTO_mem_ctrl(int mode) {
+  #ifdef OPENSSL_NO_CRYPTO_MDEBUG
+  // Since AWS-LC doesn't support CRYPTO_MDEBUG, always return 0
+  (void)mode;  // Silence unused parameter warning
+  return 0;
+  #else
+    #error "Must compile with OPENSSL_NO_CRYPTO_MDEBUG defined."
+  #endif
+}
+
 // Below can be customized by |CRYPTO_set_mem_functions| only once.
 static void *(*malloc_impl)(size_t, const char *, int) = NULL;
 static void *(*realloc_impl)(void *, size_t, const char *, int) = NULL;
--- a/crypto/mem.c
+++ b/crypto/mem.c
@@ -122,11 +122,13 @@ WEAK_SYMBOL_FUNC(void, OPENSSL_memory_free, (void *ptr))
 WEAK_SYMBOL_FUNC(size_t, OPENSSL_memory_get_size, (void *ptr))
 WEAK_SYMBOL_FUNC(void *, OPENSSL_memory_realloc, (void *ptr, size_t size))
 
-int CRYPTO_mem_ctrl(int mode) {
-  #ifdef OPENSSL_NO_CRYPTO_MDEBUG
-  // Since AWS-LC doesn't support CRYPTO_MDEBUG, always return 0
-  (void)mode;  // Silence unused parameter warning
-  return 0;
+// Control function for crypto memory debugging Always returns 0 in AWS-LC
+// |mode| Unused parameter AWS-LC doesn't support |CRYPTO_MDEBUG|
+// Always returns 0
+int CRYPTO_mem_ctrl(const int mode) {
+  #if defined (OPENSSL_NO_CRYPTO_MDEBUG)
+    (void)mode;  // Silence unused parameter warning
+    return 0;
   #else
     #error "Must compile with OPENSSL_NO_CRYPTO_MDEBUG defined."
   #endif
--- a/crypto/mem.c
+++ b/crypto/mem.c
@@ -125,7 +125,7 @@ WEAK_SYMBOL_FUNC(void *, OPENSSL_memory_realloc, (void *ptr, size_t size))
 // Control function for crypto memory debugging Always returns 0 in AWS-LC
 // |mode| Unused parameter AWS-LC doesn't support |CRYPTO_MDEBUG|
 // Always returns 0
-int CRYPTO_mem_ctrl(const int mode) {
+int CRYPTO_mem_ctrl(int mode) {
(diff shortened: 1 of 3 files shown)


crypto/mem.c
```diff
@@ -122,6 +122,16 @@ WEAK_SYMBOL_FUNC(void, OPENSSL_memory_free, (void *ptr))
 WEAK_SYMBOL_FUNC(size_t, OPENSSL_memory_get_size, (void *ptr))
 WEAK_SYMBOL_FUNC(void *, OPENSSL_memory_realloc, (void *ptr, size_t size))
 
+int CRYPTO_mem_ctrl(int mode) {
+  #ifdef OPENSSL_NO_CRYPTO_MDEBUG
```
samuel40791765, line 126: Nit:
```suggestion
  #if defined(OPENSSL_NO_CRYPTO_MDEBUG)
```

Spacing's also a bit off below.

include/openssl/crypto.h
```diff
@@ -175,6 +175,11 @@ OPENSSL_EXPORT unsigned long OpenSSL_version_num(void);
 
 OPENSSL_EXPORT unsigned long awslc_api_version_num(void);
 
+// |CRYPTO_mem_ctrl| intentionally does nothing.
+// AWS-LC does not support the related
+// memory debugging with |OPENSSL_NO_CRYPTO_MDEBUG| defined.
+OPENSSL_EXPORT OPENSSL_DEPRECATED int CRYPTO_mem_ctrl(int mode);
```
samuel40791765, line 181: More of a nitpick, but deprecated functions would be more suitable at the bottom of the header file.

include/openssl/crypto.h
```diff
@@ -175,6 +175,11 @@ OPENSSL_EXPORT unsigned long OpenSSL_version_num(void);
 
 OPENSSL_EXPORT unsigned long awslc_api_version_num(void);
 
+// |CRYPTO_mem_ctrl| intentionally does nothing.
```
samuel40791765, line 178:
```suggestion
// |CRYPTO_mem_ctrl| intentionally does nothing and returns 0.
```

include/openssl/crypto.h
```diff
@@ -175,6 +175,11 @@ OPENSSL_EXPORT unsigned long OpenSSL_version_num(void);
...
+// AWS-LC does not support the related
+// memory debugging with |OPENSSL_NO_CRYPTO_MDEBUG| defined.
```
samuel40791765, line 180: Could be clearer that this is defined by default in AWS-LC. The original wording leaves some room for interpretation that this is adjustable by the user.
```suggestion
// AWS-LC defines `OPENSSL_NO_CRYPTO_MDEBUG` by default. These are related to
// memory debugging functionalities provided by OpenSSL, but are not supported in AWS-LC..
```

samuel40791765 (commented):
Great find and attention to detail! I like the way that we'll error out if we remove the surrounding definitions.

Just some minor nitpicks and documentation improvements, but lgtm once that's done.

crypto/mem.c
```diff
@@ -122,6 +122,16 @@ WEAK_SYMBOL_FUNC(void, OPENSSL_memory_free, (void *ptr))
...
+  // Since AWS-LC doesn't support CRYPTO_MDEBUG, always return 0
+  (void)mode;  // Silence unused parameter warning
+  return 0;
+  #else
+    #error "Must compile with OPENSSL_NO_CRYPTO_MDEBUG defined."
+  #endif
```
justsmth, line 132: Why is the `#ifdef` needed? `OPENSSL_NO_CRYPTO_MDEBUG` is [defined in our headers](https://github.com/aws/aws-lc/blob/cea04a616b340838a69b3628b9b7b41f83e3e7a5/include/openssl/opensslconf.h#L36). 

Is it possible for compilation to fail here?

include/openssl/crypto.h
darylmartin100: We should also update the documentation we have listing no-op functions in our repo.  

https://github.com/aws/aws-lc/blob/3d3d3a3cf4f714e7c3d0189c3619ee3e2bd2ccbe/docs/porting/functionality-differences.md#libcrypto-no-ops

crypto/mem.c
kingstjo: We intentionally want compilation to fail if `OPENSSL_NO_CRYPTO_MDEBUG` is no longer defined in a future build, at which time we can choose to implement or keep as is.

This covers us in the case of "this does not prevent us from providing these APIs in the future and removing the OPENSSL_NO_CRYPTO_MDEBUG define." seen here [PR](https://github.com/aws/aws-lc/pull/868)

include/openssl/crypto.h
justsmth: @kingstjo - in this PR, would you also add this function to the listing of no-op functions in our documentation linked above? Thanks!

include/openssl/crypto.h
```diff
@@ -217,6 +217,12 @@ OPENSSL_EXPORT void OPENSSL_cleanup(void);
 // |BORINGSSL_FIPS| and zero otherwise.
 OPENSSL_EXPORT int FIPS_mode_set(int on);
 
+// |CRYPTO_mem_ctrl| intentionally does nothing and returns 0.
```
samuel40791765, line 220: CI is failing on the documentation check. 
```suggestion
// CRYPTO_mem_ctrl intentionally does nothing and returns 0.
```

include/openssl/crypto.h
kingstjo: Thank you both!
I have updated functionality-differences.md to reflect this no-op in addition to previous no-op `X509_TRUST_cleanup`

justsmth (approved):

samuel40791765 (approved):

samuel40791765 (commented):
There's a CI failure going on. PR looks good to me, but we should fix it before merging this in.

samuel40791765 (approved):

PR #2302, opened 2025-03-31
SSL_in_*_init macros

justsmth:
### Description of changes: 
Provide functions `SSL_in_connect_init` and `SSL_in_accept_init`.

### Call-outs:
Function definitions match OpenSSL macros: https://github.com/openssl/openssl/blob/a3fc812c0c78e2f5db8b9d45bddaff62dfc958ae/include/openssl/ssl.h#L1071-L1073

### DIFF
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -1774,6 +1774,13 @@ OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_ciphers(const SSL *ssl);
 
 // Connection information.
 
+// SSL_in_connect_init returns 1 if |ssl| is a client and has a pending
+// handshake. Otherwise, it returns 0.
+# define SSL_in_connect_init(a)          (SSL_in_init(a) && !SSL_is_server(a))
+// SSL_in_accept_init returns 1 if |ssl| is a server and has a pending
+// handshake. Otherwise, it returns 0.
+# define SSL_in_accept_init(a)           (SSL_in_init(a) && SSL_is_server(a))
+
 // SSL_is_init_finished returns one if |ssl| has completed its initial handshake
 // and has no pending handshake. It returns zero otherwise.
 OPENSSL_EXPORT int SSL_is_init_finished(const SSL *ssl);
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -1776,10 +1776,11 @@ OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_ciphers(const SSL *ssl);
 
 // SSL_in_connect_init returns 1 if |ssl| is a client and has a pending
 // handshake. Otherwise, it returns 0.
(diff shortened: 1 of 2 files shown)


smittals2 (dismissed):

include/openssl/ssl.h
```diff
@@ -1774,6 +1774,13 @@ OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_ciphers(const SSL *ssl);
 
 // Connection information.
 
+// SSL_in_connect_init returns 1 if |ssl| is a client and has a pending
+// handshake. Otherwise, it returns 0.
+# define SSL_in_connect_init(a)          (SSL_in_init(a) && !SSL_is_server(a))
+// SSL_in_accept_init returns 1 if |ssl| is a server and has a pending
+// handshake. Otherwise, it returns 0.
+# define SSL_in_accept_init(a)           (SSL_in_init(a) && SSL_is_server(a))
```
samuel40791765, line 1782: I know OpenSSL defines these as macros, but could we define these as functions here instead? We already do something similar where all the `SSL_CTRL` related functions are redefined as functions in AWS-LC.
https://github.com/aws/aws-lc/blob/92e1332f940e89dcf5e693c1954f436e8237e290/include/openssl/ssl.h#L6028-L6036

include/openssl/ssl.h
```diff
@@ -1774,6 +1774,14 @@ OPENSSL_EXPORT STACK_OF(SSL_CIPHER) *SSL_get_ciphers(const SSL *ssl);
 
 // Connection information.
 
+// SSL_in_connect_init returns 1 if |ssl| is a client and has a pending
+// handshake. Otherwise, it returns 0.
+OPENSSL_EXPORT int SSL_in_connect_init(const SSL *ssl);
+
+// SSL_in_accept_init returns 1 if |ssl| is a server and has a pending
```
smittals2, line 1781: Not blocking, is there benefit in documenting these are macros in OpenSSL and making funcs is intentional?

smittals2 (approved):

include/openssl/ssl.h
andrewhop: If these are OpenSSL macros we probably want to define macros for them like we do other OpenSSL macros, e.g.`#define SSL_CTX_get_mode SSL_CTX_get_mode` so customers can check with `#if defined(SSL_in_connect_init)`.

samuel40791765: Ah yes I'd forgotten about those. There's a longer list of all the macros in OpenSSL we did as functions here:
https://github.com/aws/aws-lc/blob/c540a7dedd642ff3ef6f338e77cbd2ec709d3ab9/include/openssl/ssl.h#L6099-L6172

We can just append these at the bottom.

smittals2 (approved):

samuel40791765 (approved):

PR #2306, opened 2025-04-01
Reinstate indefinite length and [UNIVERSAL 0] support in crypto/asn1

samuel40791765:
### Issues:
Addresses `CryptoAlg-3037`

### Description of changes: 
This is the first step of trying to reinstate support for BER in our OpenSSL ASN1 macro parsers. Proper support for BER consists of two parts, "indefinite length BER" and "implicitly tagged constructed BER strings". This is being done to properly support BER in our PKCS7 parsers, as we've made the transition to switch from CBB -> ASN1 macros for PKCS7 for now: https://github.com/aws/aws-lc/commit/621bceb86dcad30ee5f2b42d2ffdf83115908d54. We were using `CBS_asn1_ber_to_der` to translate BER -> DER for `d2i_PKCS7`, but the function has a caveat with implicitly-tagged constructed strings that causes encoding translations to be incomplete. We've decided to restore BER support in the macro parsers to resolve the PKCS7 compatibility issue and provide better OpenSSL compatibility for possible applications using OpenSSL's ASN.1 functions with BER encoding. 

This commit is a revert of the two prior commits below. All "new logic" in `crypto/asn1/tasn_dec.c` was taken from these two commits. 
* https://github.com/aws/aws-lc/commit/edbdc240ecb6a2d5a500b8e2eedfe3e6a2423c0a
* https://github.com/aws/aws-lc/commit/ee510f58895c6a88b59f1b66a94939d08ebdfe5b

The only smaller new changes are the following:
1. I've only translated the comment code styling and added a few more comments clarifying some of the historic ASN1 shenanigans.
2. There's also a change to the call to `asn1_get_object_maybe_indefinite` to call `ASN1_get_object`. This aligns with the original code in OpenSSL. I think we can clean up some of the changes done in https://github.com/aws/aws-lc/commit/cfe86a3830d5212099b2284a964be066c653dcbc, but I'd prefer to do so in a separate commit.

### Testing:
* I've reused the BER tests in our CBB parsers here so we're confident that the functionalities 1:1. This commit is reinstating indefinite length so I've only borrowed those corresponding tests for now.
* I've generated a few new ASN1 test files with indefinite length as well. Indefinite length is more common among container types like SET and SEQUENCE, so we only pass those tests for now. Other indefinite length scenarios involve constructed form, but we don't have support for that just yet.


crypto/asn1/asn1_test.cc
```diff
@@ -2356,45 +2356,81 @@ TEST(ASN1Test, GetObject) {
...
+              {0x30, 0x11, 0x06, 0x0c, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x12, 0x04,
+               0x01, 0x84, 0xb7, 0x09, 0x01, 0x00, 0x01, 0x61},
+              true);
+
+
+  // The following test cases are rejected by OpenSSL with their type specific
+  // counterparts. They are parsable with |d2i_ASN1_TYPE| however, and we test
+  // that later.
+  const std::vector<uint8_t> zero_tag_sequence = {0x30, 0x02, 0x00, 0x00};
+  const std::vector<uint8_t> zero_tag_set_any = {0x31, 0x02, 0x00, 0x00};
```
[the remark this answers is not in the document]
samuel40791765: Ahh nice, I hadn't thought of looking into OpenSSL's tests. I'll scour for more if there's any.

crypto/asn1/tasn_dec.c
```diff
@@ -855,19 +911,84 @@ static int asn1_ex_c2i(ASN1_VALUE **pval, const unsigned char *cont, long len,
...
+    }
+    q = p;
+    // Just read in a header: only care about the length.
+    if (!asn1_check_tlen(&plen, NULL, NULL, &inf, NULL, &p, len,
+                         -1, 0, 0)) {
+      OPENSSL_PUT_ERROR(ASN1, ASN1_R_NESTED_ASN1_ERROR);
+      return 0;
+                         }
+    if (inf) {
+      expected_eoc++;
```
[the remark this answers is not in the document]
samuel40791765: Good catch. I was mainly focused on reverting the commit, but new changes around the code in OpenSSL definitely need to be taken note of as well. I'll look around to see if there's more.

crypto/asn1/tasn_dec.c
```diff
@@ -855,19 +911,89 @@ static int asn1_ex_c2i(ASN1_VALUE **pval, const unsigned char *cont, long len,
...
+    }
+    q = p;
+    // Just read in a header: only care about the length.
+    if (!asn1_check_tlen(&plen, NULL, NULL, &inf, NULL, &p, len,
+                         -1, 0, 0)) {
+      OPENSSL_PUT_ERROR(ASN1, ASN1_R_NESTED_ASN1_ERROR);
+      return 0;
+    }
+    if (inf) {
+      if (expected_eoc == UINT32_MAX) {
```
WillChilds-Klein, line 949: nit: suggest adding a comment explaining that this is an underflow, and the significance of that condition in the loop

crypto/asn1/tasn_dec.c
```diff
@@ -679,8 +727,16 @@ static int asn1_d2i_ex_primitive(ASN1_VALUE **pval, const unsigned char **in,
     }
 
     cont = *in;
-    len = p - cont + plen;
-    p += plen;
+    // If indefinite length constructed find the real end.
+    if (inf) {
+      if (!asn1_find_end(&p, plen, inf)) {
+        goto err;
```
WillChilds-Klein, line 733: this is uncovered. can we add a test case that drops the `00 00` trailer?

crypto/asn1/tasn_dec.c
```diff
@@ -855,19 +911,89 @@ static int asn1_ex_c2i(ASN1_VALUE **pval, const unsigned char *cont, long len,
...
+  // found. If more indefinite length constructed headers are encountered
+  // increment the expected eoc count otherwise just skip to the end of the
+  // data.
+  while (len > 0) {
+    if (asn1_check_eoc(&p, len)) {
+      expected_eoc--;
+      if (expected_eoc == 0) {
+        break;
+      }
+      len -= 2;
```
WillChilds-Klein, line 938: this is uncovered. is there a nested-indefinite test SEQUENCE we can construct that will exercise this?

crypto/asn1/tasn_dec.c
```diff
@@ -855,19 +911,89 @@ static int asn1_ex_c2i(ASN1_VALUE **pval, const unsigned char *cont, long len,
...
+      continue;
+    }
+    q = p;
+    // Just read in a header: only care about the length.
+    if (!asn1_check_tlen(&plen, NULL, NULL, &inf, NULL, &p, len,
+                         -1, 0, 0)) {
+      OPENSSL_PUT_ERROR(ASN1, ASN1_R_NESTED_ASN1_ERROR);
+      return 0;
+    }
+    if (inf) {
```
WillChilds-Klein, line 948: the `true` case here is uncovered. can we add a test case to exercise it?

WillChilds-Klein (commented):
Nice, much cleaner revert than I was expecting! Changes are looking good, just a few requests to bump up coverage.

samuel40791765 (commented):
> just a few requests to bump up coverage.

Added a couple more test cases which should add coverage against the places we were looking at!

justsmth (approved):

WillChilds-Klein (approved):

PR #2310, opened 2025-04-03
Reinstate support for constructed strings in crypto/asn1

samuel40791765:
### Issues:
Resolves `CryptoAlg-3037`

### Description of changes: 
This is the last step of trying to reinstate support for BER in our OpenSSL ASN1 macro parsers. Proper support for BER consists of two parts, "indefinite length BER" and "implicitly tagged constructed BER strings". The first step of supporting indefinite length BER can be found in https://github.com/aws/aws-lc/commit/eafd9404887d5982dd647bd4f914664281130060.

This commit is a revert of the prior commit below. All "new logic" in `crypto/asn1/tasn_dec.c` was taken from this.
* https://github.com/aws/aws-lc/commit/a70edd47a235b9b29c950b887f51f20635252d04

The only smaller new changes are the following:
* The original commit uses an additional `free_cont` parameter to pass into `asn1_ex_c2i`. This indicates whether a buffer should be reused or newly allocated. This made things a bit funky and the actual performance upgrade for reusing a buffer seems negligible. Removing the use of `free_cont` altogether helps isolate the ASN1 code changes to just one static function in `crypto/asn1/tasn_dec.c` and cleans up logic.

### Testing:
* I've reused the BER tests in our CBB parsers here so we're confident that the functionalities 1:1. https://github.com/aws/aws-lc/commit/eafd9404887d5982dd647bd4f914664281130060 only borrowed the ones that were indefinite length, while this commit takes and tests against constructed string test cases.
*  https://github.com/aws/aws-lc/commit/eafd9404887d5982dd647bd4f914664281130060 documented some test cases  that were failing due to no support for constructed strings. I've update the documentation and turned on the tests, as this commit now allows proper support of them.


justsmth (dismissed):

crypto/asn1/tasn_dec.c
```diff
@@ -738,11 +744,29 @@ static int asn1_d2i_ex_primitive(ASN1_VALUE **pval, const unsigned char **in,
...
+    // may get this wrong. The relevant specs say that constructed string
+    // types should be OCTET STRINGs internally irrespective of the type.
+    // So instead just check for UNIVERSAL class and ignore the tag.
+    if (!asn1_collect(&buf, &p, plen, inf, -1, V_ASN1_UNIVERSAL, 0)) {
+      goto err;
+    }
+    len = buf.length;
+    // Append a final null to string.
+    if (!BUF_MEM_grow_clean(&buf, len + 1)) {
+      OPENSSL_PUT_ERROR(ASN1, ERR_R_MALLOC_FAILURE);
```
WillChilds-Klein, line 764: nit: `OPENSSL_malloc` gets called under the hood, so setting the error here is redundant, right?

crypto/asn1/tasn_dec.c
```diff
@@ -967,6 +992,85 @@ static int asn1_find_end(const unsigned char **in, long len, char inf) {
...
+  *in = p;
+  return 1;
+}
+
+static int collect_data(BUF_MEM *buf, const unsigned char **p, long plen) {
+  int len;
+  if (buf) {
+    len = buf->length;
+    if (!BUF_MEM_grow_clean(buf, len + plen)) {
+      OPENSSL_PUT_ERROR(ASN1, ERR_R_MALLOC_FAILURE);
```
WillChilds-Klein, line 1064: same nit about redundant error

WillChilds-Klein (dismissed):

crypto/asn1/tasn_dec.c
samuel40791765: Correct, since the `OPENSSL_malloc` change was made after the commit that removed support for BER constructed strings. Will remove!

crypto/asn1/tasn_dec.c
```diff
@@ -738,11 +744,29 @@ static int asn1_d2i_ex_primitive(ASN1_VALUE **pval, const unsigned char **in,
...
+      goto err;
+    }
+    len = buf.length;
+    // Append a final null to string.
+    if (!BUF_MEM_grow_clean(&buf, len + 1)) {
+      OPENSSL_PUT_ERROR(ASN1, ERR_R_MALLOC_FAILURE);
+      goto err;
+    }
+    buf.data[len] = 0;
+    cont = (const unsigned char *)buf.data;
```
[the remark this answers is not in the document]
samuel40791765: I tried doing so, but `buf.data`'s memory is assigned to `cont` here and `cont` is used later in `asn1_ex_c2i` below. Defining `buf` here makes it harder to free it's corresponding data since it'll be out of scope :(.

WillChilds-Klein (approved):

crypto/asn1/tasn_dec.c
```diff
@@ -966,6 +990,84 @@ static int asn1_find_end(const unsigned char **in, long len, char inf) {
...
+    OPENSSL_PUT_ERROR(ASN1, ASN1_R_MISSING_EOC);
+    return 0;
+  }
+  *in = p;
+  return 1;
+}
+
+static int collect_data(BUF_MEM *buf, const unsigned char **p, long plen) {
+  int len;
+  if (buf) {
```
skmcgrail, line 1060: It seems like `collect_data` is never called with `buf == NULL`, should this return `0` instead?

skmcgrail (approved):

crypto/asn1/tasn_dec.c
samuel40791765: Great question, I did some digging around. But judging from the comment in `asn1_collect` which mentions "If no buffer and not indefinite length constructed just pass over the encoded data", it does seem like there's the possibility of `buf` being `NULL` and there being operations that need to be done under that circumstance.

I'll leave it as is for the time being, OpenSSL's ASN1 implementation is also like this today.

PR #2315, opened 2025-04-07
Add AES CBC cipher to speed.cc

andrewhop:
### Description of changes: 
1. Add CBC mode by itself to the speed tool
2. Update SpeedEvpGenericChunk to handle CBC which adds more padding

### Testing:
Tested locally, will also be covered in CI:
```
bssl speed -filter CBC -chunks 128
Did 7850000 EVP-AES-128-CBC encrypt init operations in 1000096us (7849246.5 ops/sec)
Did 5617000 EVP-AES-128-CBC encrypt (128 bytes) operations in 1000042us (5616764.1 ops/sec): 718.9 MB/s
Did 7293000 EVP-AES-128-CBC decrypt init operations in 1000015us (7292890.6 ops/sec)
Did 8290000 EVP-AES-128-CBC decrypt (128 bytes) operations in 1000005us (8289958.6 ops/sec): 1061.1 MB/s
```


tool/speed.cc
```diff
@@ -2875,6 +2877,9 @@ bool Speed(const std::vector<std::string> &args) {
        !SpeedEvpCipherGeneric(EVP_aes_128_ctr(), "EVP-AES-128-CTR", kTLSADLen, selected) ||
        !SpeedEvpCipherGeneric(EVP_aes_192_ctr(), "EVP-AES-192-CTR", kTLSADLen, selected) ||
        !SpeedEvpCipherGeneric(EVP_aes_256_ctr(), "EVP-AES-256-CTR", kTLSADLen, selected) ||
+       !SpeedEvpCipherGeneric(EVP_aes_128_cbc(), "EVP-AES-128-CBC", kTLSADLen, selected) ||
```
WillChilds-Klein, line 2880: what padding is used here? ISO10126 or PKCSv1.5?

Not a concern for this change, but ACCP [supports ISO10126 for CBC](https://github.com/corretto/amazon-corretto-crypto-provider/blob/4db841b3dbcb8d841c60c6cdd666a5c07f77c06e/csrc/aes_cbc.cpp#L251), and if we could offload that padding to an AWS-LC API, it'd be a nice simplification.

WillChilds-Klein (approved):

tool/speed.cc
andrewhop: Turns out this isn't obvious what is used: [cipher.h](https://github.com/aws/aws-lc/blob/main/include/openssl/cipher.h#L199-L205) says standard padding is used. [Reading the code](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/cipher/cipher.c#L389-L402) this uses [RFC 5652/PKCS 7](https://datatracker.ietf.org/doc/html/rfc5652#section-6.3) padding (each padding byte value is the number of padding bytes).

smittals2 (approved):

PR #2328, opened 2025-04-10
[DRAFT] Fix clang-tidy lints

justsmth:
### Description of changes: 
* Correct several issues that are reported by clang-tidy.

### Call-outs:
* I'm not sure whether the `OPENSSL_cleanse`s at the ends of `BF_cbc_encrypt` and `CAST_cbc_encrypt` are really needed, but the zero-assignments that were there before would trivially be optimized away.


crypto/pkcs7/pkcs7.c
```diff
@@ -915,9 +915,11 @@ int PKCS7_set_digest(PKCS7 *p7, const EVP_MD *md) {
         OPENSSL_PUT_ERROR(PKCS7, PKCS7_R_UNKNOWN_DIGEST_TYPE);
         return 0;
       }
-      if (p7->d.digest->digest_alg) {
-        OPENSSL_free(p7->d.digest->digest_alg->parameter);
+      if (p7->d.digest->digest_alg == NULL) {
+        OPENSSL_PUT_ERROR(PKCS7, ERR_R_ASN1_LIB);
+        return 0;
```
torben-hansen, line 920: behaviour change?

tool/client.cc
```diff
@@ -729,7 +729,7 @@ bool DoClient(std::map<std::string, std::string> args_map, bool is_openssl_s_cli
   if (!certPathFlag.empty()) {
     if (!SSL_CTX_load_verify_locations(
             ctx.get(), nullptr, args_map[certPathFlag].c_str())) {
-      fprintf(stderr, "Failed to load root certificates.\n");
+      fputs("Failed to load root certificates.\n", stderr);
```
smittals2, line 732: Non blocking, but we use fprintf in a lot of places (even in this same file). Why is clang-tidy only flagging some of these calls?

smittals2 (dismissed):

crypto/dynamic_loading_test.c
```diff
@@ -113,7 +113,7 @@ int main(int argc, char *argv[]) {
 #else
 
 int main(int argc, char **argv) {
-  fprintf(stdout, "PASS\n");
+  fprintf(stdout, "PASS\n"); // NOLINT(clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling)
```
justsmth, line 116: I actually don't know why this line was flagged:
```
/Users/justsmth/repos/aws-lc/crypto/dynamic_loading_test.c:116:3: warning: Call to function 'fprintf' is insecure as it does not provide security checks introduced in the C11 standard. Replace with analogous functions that support length arguments or provides boundary checks such as 'fprintf_s' in case of C11 [clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling]
  116 |   fprintf(stdout, "PASS\n");
      |   ^~~~~~~
```

torben-hansen (approved):

crypto/dynamic_loading_test.c
justsmth: BTW: I looked into this, but GLIBC doesn't support these secure variants (like `fprintf_s`): https://sourceware.org/glibc/manual/latest/html_node/Formatted-Output-Functions.html

It seems to mostly be an MSVC extension that's (only optionally) part of the C11 standard.

smittals2 (approved):

PR #2331, opened 2025-04-11
Implement BIO_dump

kingstjo:
# Issues:
Addresses #CryptoAlg-3030

# Description of changes:

Implemented BIO_dump() function in AWS-LC, matching OpenSSL 1.1.1's functionality. This function enables hexadecimal dumping of data through the BIO interface, providing compatibility for applications that rely on this OpenSSL API.

- Uses existing hexdump infrastructure in crypto/bio/hexdump.c
- Maintains OpenSSL 1.1.1 API compatibility
- Returns total bytes written to match OpenSSL behavior

# Call-outs:

- Function maintains compatibility with OpenSSL 1.1.1 API specification
- Returns total bytes written, matching OpenSSL's behavior

# Testing:

Added test in crypto/bio/bio_test.cc


crypto/bio/hexdump.c
```diff
@@ -192,3 +192,17 @@ int BIO_hexdump(BIO *bio, const uint8_t *data, size_t len, unsigned indent) {
...
+    return -1;
+  }
+  
+  // Use existing hexdump functionality with no indentation
+  if (!BIO_hexdump(bio, (const uint8_t *)data, (size_t)len, 0)) {
+    return -1;
+  }
+  
+  // Return estimate of bytes written (each line is ~80 bytes)
+  return ((len + 15) / 16) * 80;
```
justsmth, line 207: Instead of needing to estimate how many bytes were written, this could be exact. Why not modify `hexdump_write` above to return the actual number of bytes written (just bump a counter on each of its calls to `BIO_write`), call `hexdump_write` directly instead of calling it via `BIO_hexdump`, then return whatever value it returns?

kingstjo: Thank you for your suggestion about making the byte count exact. I considered modifying `hexdump_write `as you suggested, but went with the memory BIO approach for a few reasons:

Since `hexdump_write `is a static function with multiple write points (both it and the finish function write data), I was concerned about maintaining proper encapsulation.

The memory BIO approach seemed cleaner by using existing APIs rather than modifying internal functions, especially as a newcomer to the codebase.

This implementation accounts for all bytes written while being resilient to any future changes in the hexdump formatting logic.

The slight overhead of an extra allocation seemed acceptable for a debugging function like `BIO_dump`. I believe this approach offers the best balance of correctness and maintainability, but I'm open to discussing further if you have additional insights on this implementation.

samuel40791765 (approved):

justsmth (approved):

Issue #487, opened 2022-05-11
GCC12 support for Yocto Linux


thomas-roos:
### Problem:

Yocto mainline switched from GCC11 to GCC12 and now aws-lc is not buildable anymore in [meta-aws ](https://github.com/aws4embeddedlinux/meta-aws)
Wondering if this is a real bug or just a false positive.
When are you planning to support GCC12?

#### Relevant details
AWS-LC commit: 0.0.2 to 1.1.0

System information: for linux, below info can be collected by running `uname -srvmp`
 * CPU architecture: x86-64

Build log: 
```text
| FAILED: crypto/CMakeFiles/crypto_objects.dir/x509v3/v3_utl.c.o
[...]

PR #2340, opened 2025-04-18
Add Support for IPv4 and IPv5 X.509 Certificate Name Constraints

skmcgrail:

### DIFF
--- a/crypto/x509/v3_ncons.c
+++ b/crypto/x509/v3_ncons.c
@@ -564,38 +564,45 @@ static int nc_uri(const ASN1_IA5STRING *uri, const ASN1_IA5STRING *base) {
 #define IPV6_CIDR_LEN (IPV6_ADDR_LEN * 2)
 
 static int nc_ip(const ASN1_OCTET_STRING *ip, const ASN1_OCTET_STRING *base) {
-  int ip_len, cidr_len, i;
-  const unsigned char *ip_ptr, *cidr_ptr, *cidr_prefix_ptr;
-  ip_ptr = ip->data;
-  ip_len = ip->length;
-  cidr_ptr = base->data;
-  cidr_len = base->length;
+  CBS ip_cbs, cidr_cbs, cidr_addr, cidr_mask;
 
-  // The length should be either the length of an IPv4 or IPv6 address.
-  if (!((ip_len == IPV4_ADDR_LEN) || (ip_len == IPV6_ADDR_LEN))) {
-    return X509_V_ERR_UNSUPPORTED_NAME_SYNTAX;
-  }
+  CBS_init(&ip_cbs, ip->data, ip->length);
+  CBS_init(&cidr_cbs, base->data, base->length);
 
-  // The CIDR length should either be the length of an IPv4 Address+mask or IPv6
-  // Address+mask.
-  if (!((cidr_len == IPV4_CIDR_LEN) || (cidr_len == IPV6_CIDR_LEN))) {
-    return X509_V_ERR_UNSUPPORTED_NAME_SYNTAX;
-  }
+  size_t ip_len = CBS_len(&ip_cbs);
+  size_t cidr_len = CBS_len(&cidr_cbs);
 
-  // Validate we have an IPv4 address and IPv4 CIDR or IPv6 Address and IPv6
-  // CIDR
+  // Validate that the CIDR length is twice the size of the provided IP (address
+  // + mask).
   if (ip_len * 2 != cidr_len) {
     return X509_V_ERR_PERMITTED_VIOLATION;
   }
 
-  // Pointer to the CIDR prefix mask
-  cidr_prefix_ptr = cidr_ptr + ip_len;
+  // Next the IP length should be either the length of an IPv4 or IPv6 address.
+  if (!((ip_len == IPV4_ADDR_LEN) || (ip_len == IPV6_ADDR_LEN))) {
+    return X509_V_ERR_UNSUPPORTED_NAME_SYNTAX;
+  }
 
-  /* Considering possible not aligned base ipAddress */
-  /* Not checking for wrong mask definition: i.e.: 255.0.255.0 */
-  for (i = 0; i < ip_len; i++) {
-    if ((ip_ptr[i] & cidr_prefix_ptr[i]) !=
-        (cidr_ptr[i] & cidr_prefix_ptr[i])) {
+  // Finally the CIDR length should either be the length of an IPv4 address+mask
+  // or IPv6 address+mask.
+  if (!((cidr_len == IPV4_CIDR_LEN) || (cidr_len == IPV6_CIDR_LEN))) {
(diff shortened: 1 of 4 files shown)


skmcgrail:
Need to dig into the nature of this issue and associated fix https://github.com/aws/aws-lc/issues/487 because it's impacting being able to specify `::` as an IPv6 address. (Which is a valid address where all 128-bits are zeros).

crypto/x509/v3_ncons.c
```diff
@@ -556,3 +560,55 @@ static int nc_uri(const ASN1_IA5STRING *uri, const ASN1_IA5STRING *base) {
...
+
+  uint8_t ip_byte, cidr_addr_byte, cidr_mask_byte;
+
+  // Considering possible not aligned base ipAddress
+  // Not checking for wrong mask definition: i.e.: 255.0.255.0
+  for (size_t i = 0; i < ip_len; i++) {
+    if (!CBS_get_u8(&ip_cbs, &ip_byte) ||
+        !CBS_get_u8(&cidr_addr, &cidr_addr_byte) ||
+        !CBS_get_u8(&cidr_mask, &cidr_mask_byte) ||
+        ((ip_byte & cidr_mask_byte) != (cidr_addr_byte & cidr_mask_byte))) {
```
justsmth, line 608: Requested action:
1) Comment as to "why" we don't verify the mask.
2) Add test case with an invalid mask. Maybe have cidr-addr, cidr-mask and ip-addr all be the same?

---- 

* The comment indicates "Not checking for wrong mask definition: i.e.: 255.0.255.0", but it doesn't say *why* it's not.  Do we not check in order to be more compatible with OpenSSL?   
* If we must accept invalid masks such as `255.0.255.0` for compatibility, is there a test that covers it? (My implementation below doesn't cause any tests to fail, so I'm assuming that case is not covered.)
* I was curious about how we might verify the mask (were it required) and came up with this:
```
  // Considering possible not aligned base ipAddress
  for (size_t i = 0; i < ip_len; i++) {
    if (!CBS_get_u8(&ip_cbs, &ip_byte) ||
        !CBS_get_u8(&cidr_addr, &cidr_addr_byte) ||
        !CBS_get_u8(&cidr_mask, &cidr_mask_byte)) {
      return X509_V_ERR_PERMITTED_VIOLATION;
    }
    // Verify ip is within subnet
    if ((ip_byte & cidr_mask_byte) != (cidr_addr_byte & cidr_mask_byte)) {
      return X509_V_ERR_PERMITTED_VIOLATION;
    }
    // Verify mask is valid
    if (found_zero) {
      if (0 != cidr_mask_byte) {
        return X509_V_ERR_PERMITTED_VIOLATION;
      }
    } else if (255 != cidr_mask_byte) {
      // found the split
      if (cidr_mask_byte & (~cidr_mask_byte >> 1)) {
        return X509_V_ERR_PERMITTED_VIOLATION;
      }
      found_zero = 1;
    }
  }
```

crypto/x509/v3_utl.c
```diff
@@ -1186,25 +1186,58 @@ int x509v3_a2i_ipadd(unsigned char ipout[16], const char *ipasc) {
...
 }
 
-static int ipv4_from_asc(unsigned char v4[4], const char *in) {
-  int a0, a1, a2, a3;
-  if (sscanf(in, "%d.%d.%d.%d", &a0, &a1, &a2, &a3) != 4) { // NOLINT(clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling)
+// get_ipv4_component consumes one IPv4 component, terminated by either '.' or
+// the end of the string, from |*str|. On success, it returns one, sets |*out|
+// to the component, and advances |*str| to the first unconsumed character. On
+// invalid input, it returns zero.
+static int get_ipv4_component(uint8_t *out_byte, const char **str) {
```
justsmth, line 1193: NP: This implementation seems "ugly" to me, but I don't see any bugs.

skmcgrail: Taking it for now since it was a cherry-pick from upstream :).

crypto/x509/v3_ncons.c
```diff
@@ -566,6 +567,62 @@ static int nc_uri(const ASN1_IA5STRING *uri, const ASN1_IA5STRING *base) {
...
+
+  uint64_t mask_low =
+      (((uint64_t)mask_ptr[8]) << 56) | ((uint64_t)mask_ptr[9] << 48) |
+      ((uint64_t)mask_ptr[10] << 40) | ((uint64_t)mask_ptr[11] << 32) |
+      ((uint64_t)mask_ptr[12] << 24) | ((uint64_t)mask_ptr[13] << 16) |
+      ((uint64_t)mask_ptr[14] << 8) | ((uint64_t)mask_ptr[15]);
+
+  if (!(mask_low == 0 || ((~mask_low + 1) & ~mask_low) == 0)) {
+    return 0;
+  }
```
justsmth, line 609: If the lowest-bit of the high mask is zero, then the low mask must be 0.

crypto/x509/x509_compat_test.cc
```diff
@@ -1620,16 +1621,82 @@ TEST(X509CompatTest, EECertificateWithIPv6SANIssuedByCAWithBadNameConstraint) {
...
+                   /*intermediates=*/{inter.get()},
+                   /*crls=*/{}, /*flags=*/0));
+}
+
+struct IpCidrNetmaskTestParam {
+    std::string mask_hex;
+    bool valid;
+};
+
+static const IpCidrNetmaskTestParam kIpCidrNetmaskTestParams[] = {
```
justsmth, line 1643: I love this! Thanks!

justsmth (approved):

crypto/x509/v3_ncons.c
```diff
@@ -556,3 +561,118 @@ static int nc_uri(const ASN1_IA5STRING *uri, const ASN1_IA5STRING *base) {
...
+  return (mask_low == 0 || ((~mask_low + 1) & ~mask_low) == 0);
+}
+
+int validate_cidr_mask(CBS *cidr_mask) {
+  size_t mask_len = CBS_len(cidr_mask);
+  switch (mask_len) {
+    case IPV4_ADDR_LEN:
+      return validate_ipv4_cidr_mask(CBS_data(cidr_mask), mask_len);
+    case IPV6_ADDR_LEN:
+      return validate_ipv6_cidr_mask(CBS_data(cidr_mask), mask_len);
```
samuel40791765, line 626: I was able to cross reference `nc_ip` with OpenSSL's implementation to verify things are correct. Is there a similar implementation here to cross reference or are these two functions new logic?

skmcgrail: OpenSSL did not handle validation of the mask per this comment: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/x509v3/v3_ncons.c#L698

But a valid CA shouldn't have name constraints with an invalid CIDR mask/prefix, so I think it makes sense to sanity check that here.

PR #2366, opened 2025-04-25
Mark fallible container operations as `nodiscard`

justsmth:
It's evidently too easy to ignore the return values of these operations. After all, ignoring allocation failures is the normal pattern in most C++ code.

Thus this change marks these functions as `nodiscard`. This does require a few CHECKs in test code, but it also catches one real instance of a problem.

Change-Id: I24432506283145fc2f459336fe1035cbca27bd4f
Reviewed-on: https://boringssl-review.googlesource.com/c/boringssl/+/75187
Commit-Queue: Adam Langley <agl@google.com>
Auto-Submit: Adam Langley <agl@google.com>
Reviewed-by: David Benjamin <davidben@google.com>

### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 

### Call-outs:

### Testing:

### DIFF
--- a/ssl/handoff.cc
+++ b/ssl/handoff.cc
@@ -669,8 +669,10 @@ bool SSL_apply_handback(SSL *ssl, Span<const uint8_t> handback) {
   }
   s3->session_reused = session_reused;
   hs->channel_id_negotiated = channel_id_negotiated;
-  s3->next_proto_negotiated.CopyFrom(next_proto);
-  s3->alpn_selected.CopyFrom(alpn);
+  if (!s3->next_proto_negotiated.CopyFrom(next_proto) ||
+      !s3->alpn_selected.CopyFrom(alpn)) {
+    return false;
+  }
 
   const size_t hostname_len = CBS_len(&hostname);
   if (hostname_len == 0) {


ssl/handoff.cc
```diff
@@ -669,8 +669,10 @@ bool SSL_apply_handback(SSL *ssl, Span<const uint8_t> handback) {
   }
   s3->session_reused = session_reused;
   hs->channel_id_negotiated = channel_id_negotiated;
-  s3->next_proto_negotiated.CopyFrom(next_proto);
-  s3->alpn_selected.CopyFrom(alpn);
```
smittals2, line 673: I'm a bit confused on this change, how does checking ret value here mark the function as nodiscard? 

I'm sure we use CopyFrom in a bunch of other places, why are we only changing this here?

justsmth: Sorry, this wasn't clear.  This change is just a small part of a larger change that BoringSSL was making relating to [nodiscard](https://en.cppreference.com/w/cpp/language/attributes/nodiscard), which is a C++17 feature that we're not prepared to fully take.

This PR is just a few lines from the upstream commit were they add return value checking for the calls to `CopyFrom`

smittals2 (approved):

skmcgrail (approved):

PR #2376, opened 2025-04-28
Reintroduce support for validating DNS commonName subjects when name constraints are present.

skmcgrail:
### Description of changes:
* Reverts https://github.com/aws/aws-lc/commit/c67076d653f7501136f7b208df4b011a7275e8f5 which replaced how certificates are handled when there is a DNS-like name in the certificates subject commonName. (Note this doesn't fully restore the original OpenSSL behavior, as that change seems to have been made before upstream made their source available).
* Reintroduction of the OpenSSL 1.1.1 for identifying DNS-like names in the subject commonName and the ability to use the `X509_CHECK_FLAG_NEVER_CHECK_SUBJECT` and `X509_CHECK_FLAG_ALWAYS_CHECK_SUBJECT` flags to check this value.


crypto/x509/v3_ncons.c
```diff
@@ -305,6 +305,134 @@ int NAME_CONSTRAINTS_check(X509 *x, NAME_CONSTRAINTS *nc) {
...
+  
+  // XXX: Deviation from strict DNS name syntax, also check names with '_'
+  // Check DNS name syntax, any '-' or '.' must be internal,
+  // and on either side of each '.' we can't have a '-' or '.'.
+  //
+  // If the name has just one label, we don't consider it a DNS name.  This
+  // means that "CN=sometld" cannot be precluded by DNS name constraints, but
+  // that is not a problem.
+  for (int i = 0; i < utf8_length; ++i) {
+    unsigned char c = utf8_value[i];
```
justsmth, line 361: NP:
```
const unsigned char c = utf8_value[i];
```

crypto/x509/v3_ncons.c
```diff
@@ -305,6 +305,134 @@ int NAME_CONSTRAINTS_check(X509 *x, NAME_CONSTRAINTS *nc) {
...
+    return X509_V_OK;
+  }
+  OPENSSL_free(utf8_value);
+  return X509_V_OK;
+}
+
+// Check CN against DNS-ID name constraints.
+int NAME_CONSTRAINTS_check_CN(X509 *x, NAME_CONSTRAINTS *nc) {
+  int ret;
+  X509_NAME *nm = X509_get_subject_name(x);
```
justsmth, line 399: NP:
```
const X509_NAME *nm = X509_get_subject_name(x);
```

crypto/x509/v3_ncons.c
```diff
@@ -305,6 +305,134 @@ int NAME_CONSTRAINTS_check(X509 *x, NAME_CONSTRAINTS *nc) {
   return X509_V_OK;
 }
 
+static int cn2dnsid(ASN1_STRING *cn, unsigned char **dnsid, size_t *idlen) {
+  if (dnsid == NULL || idlen == NULL) {
+    return X509_V_ERR_UNSPECIFIED;
+  }
```
justsmth, line 311: NP: Since this is only called internally, these checks could just be `assert`s.

crypto/x509/v3_ncons.c
```diff
@@ -305,6 +305,134 @@ int NAME_CONSTRAINTS_check(X509 *x, NAME_CONSTRAINTS *nc) {
...
+    break;
+  }
+
+  if (isdnsname) {
+    *dnsid = utf8_value;
+    *idlen = (size_t)utf8_length;
+    return X509_V_OK;
+  }
+  OPENSSL_free(utf8_value);
+  return X509_V_OK;
```
justsmth, line 393: NP: For names containing NULs we return `X509_V_ERR_UNSUPPORTED_NAME_SYNTAX` to kick us out of the for loop in the caller. For all other invalid names, we return `X509_V_OK`.
* DNS name are limited to like ~255 characters -- should we also reject names that are too long?
* Besides containing a NUL, are there other *egrecious* issues we might find with a name for which we should return that as well?

crypto/x509/x509_test.cc
```diff
@@ -3601,99 +3507,6 @@ TEST(X509Test, CommonNameFallback) {
...
-        ".",
-        "*",
-        "*.",
-        "example..com",
-        ".example.com",
-        "example.com..",
-        "*foo.example.com",
-        "foo.*.example.com",
-        "foo,bar",
-    };
```
justsmth, line 3625: Could some of the test cases be repurposed to test either `cn2dnsid` or `equal_wildcard` (both of which are currently `static` so not directly tested)?

justsmth (approved):

samuel40791765 (approved):

PR #2377, opened 2025-04-29
Support allowing specific unknown critical extensions

samuel40791765:
### Issues:
Resolves `V1726629971`

### Description of changes: 
As of today, AWS-LC returns an error as part of verification if there are any critical extensions present in the certificate being validated. There have been asks to set a custom OID on the issued certificates to ensure that additional validation is performed by customers after or during the handshake. The intention is to prevent accidental mis-use of these certificates without that extra validation.

To support this, we've decided to add two new APIs for this use case. 
1. `X509_STORE_CTX_add_custom_crit_oid` adds  an oid as an `ASN1_OBJECT` to the list of "known" critical extension OIDs in `ctx`. Typical OpenSSL/AWS-LC behavior returns an error if there are any unknown critical extensions present within the certificates being validated. This function lets users specify custom OIDs of any critical extensions that are within the certificates being validated, that they wish to allow. The callback mechanism enabled with `X509_STORE_CTX_set_verify_crit_oids` must be set for this feature to enabled.

2. `X509_STORE_CTX_set_verify_crit_oids` enables the `X509_STORE_CTX_verify_crit_oids_cb` with `X509_STORE_CTX`. Consumers should be performing additional validation against the custom extension oids after or during the handshake. This callback forces users to validate their custom OIDs when processing unknown custom critical extensions. The `X509_STORE_CTX_verify_crit_oids_cb` callback function gives the user the current certificate being validated as `x509` and a stack of `ASN1_OBJECT`s representing unknown critical extension OIDs that were found in `x509` and match those previously registered via`|X509_STORE_CTX_add_custom_crit_oid` as  `oids`.

This should not effect any existing consumers of `X509_verify_cert`. Any existence of an unknown critical extension will still cause the entire verification to be aborted. Only consumers that have enabled the callback and set specific OIDs with `ASN1_OBJECT` can circumvent the check and trigger the verification to pass.

### Testing:
Test certs were generated by the team asking for this feature. 

### DIFF
--- a/crypto/x509/internal.h
+++ b/crypto/x509/internal.h
@@ -344,8 +344,8 @@ struct x509_store_ctx_st {
   X509_STORE_CTX_verify_cb verify_cb;       // error callback
   X509_STORE_CTX_get_crl_fn get_crl;        // retrieve CRL
   X509_STORE_CTX_check_crl_fn check_crl;    // Check CRL validity
-  X509_STORE_CTX_verify_crit_oids
-      verify_crit_oids;  // Check custom critical oids
+  X509_STORE_CTX_verify_crit_oids_cb
+      verify_custom_crit_oids;  // Check custom critical oids
 
   // The following is built up
 
@@ -361,7 +361,7 @@ struct x509_store_ctx_st {
 
   int current_crl_score;         // score of current CRL
 
-  // Array of allowed custom critical extension oids.
+  // Stack of allowed custom critical extension oids.
   STACK_OF(ASN1_OBJECT) *custom_crit_oids;
 
   CRYPTO_EX_DATA ex_data;
--- a/crypto/x509/internal.h
+++ b/crypto/x509/internal.h
@@ -344,6 +344,8 @@ struct x509_store_ctx_st {
   X509_STORE_CTX_verify_cb verify_cb;       // error callback
   X509_STORE_CTX_get_crl_fn get_crl;        // retrieve CRL
   X509_STORE_CTX_check_crl_fn check_crl;    // Check CRL validity
+  X509_STORE_CTX_verify_crit_oids
+      verify_crit_oids;  // Check custom critical oids
 
   // The following is built up
 
@@ -358,6 +360,9 @@ struct x509_store_ctx_st {
   X509_CRL *current_crl;  // current CRL
 
   int current_crl_score;         // score of current CRL
+
+  // Array of allowed custom critical extension oids.
+  STACK_OF(ASN1_OBJECT) *custom_crit_oids;
 
   CRYPTO_EX_DATA ex_data;
 } /* X509_STORE_CTX */;
--- a/crypto/x509/x509_test.cc
+++ b/crypto/x509/x509_test.cc
@@ -8279,7 +8279,6 @@ static int verify_crit_oids_callback(X509_STORE_CTX *ctx, X509 *x509,
     return 0;  // Fail if no OIDs provided
   }
   size_t known_oid_count = sk_ASN1_OBJECT_num(oids);
-  ;
   size_t unknown_ext_count = 0;
   int last_pos = X509_get_ext_by_critical(x509, 1, -1);
   while (last_pos >= 0) {
--- a/crypto/x509/x509_test.cc
+++ b/crypto/x509/x509_test.cc
@@ -62,6 +62,67 @@ j2kCAwG+LLpGNmNwcHBwcHBwcHBwcHBwcHBwcHBwcHBwcHBw5lmgITTEvXIj+8ls
 -----END CERTIFICATE-----
 )";
 
+static const char kX509CustomExtensionsCA[] = R"(
+-----BEGIN CERTIFICATE-----
+MIIBxzCCAW2gAwIBAgIFAQAAAAAwCgYIKoZIzj0EAwIwJjEPMA0GA1UECgwGQW1h
+em9uMRMwEQYDVQQpDAo0Mjk0OTY3Mjk2MCIYDzIwMjUwMzI5MjA0OTE5WhgPOTk5
+OTEyMzEyMzU5NTlaMCYxDzANBgNVBAoMBkFtYXpvbjETMBEGA1UEKQwKNDI5NDk2
+NzI5NjBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABL3eDQzFx4cherBJdIQsxMzZ
+rCtzXBTB3f/rRMLrtjxpk2/6h3ZbE4t8MmDbwVepAKYQgT1bjPUFn+edG2U8kRej
+gYMwgYAwEgYDVR0TAQH/BAgwBgEB/wIBADBLBgNVHSMERDBCgBTMNuas7QD5hDXY
+KS7k0WDN6ckMFqEqpCgwJjEPMA0GA1UECgwGQW1hem9uMRMwEQYDVQQpDAo0Mjk0
+OTY3Mjk2MB0GA1UdDgQWBBTMNuas7QD5hDXYKS7k0WDN6ckMFjAKBggqhkjOPQQD
+AgNIADBFAiB3MJLK86+JyyoBr2s1Ugjvc7gWAHSk9OgXfyfsVmBV9gIhAPIUiYo8
+Jx+IbRyNj2WfeCbn8v3fob0wkGsKf1TSVcZ8
+-----END CERTIFICATE-----
+)";
+
+static const char kX509CustomExtensionsCert[] = R"(
+-----BEGIN CERTIFICATE-----
+MIIB6zCCAZGgAwIBAgIFAQAAAAAwCgYIKoZIzj0EAwIwJjEPMA0GA1UECgwGQW1h
+em9uMRMwEQYDVQQpDAo0Mjk0OTY3Mjk2MCIYDzIwMjUwMzI5MjA0OTE5WhgPOTk5
+OTEyMzEyMzU5NTlaMBExDzANBgNVBAoMBkFtYXpvbjBZMBMGByqGSM49AgEGCCqG
+SM49AwEHA0IABNbNswB+jmoICPKu567Odfq83s9P0N82kFYnyANgmztgHqoK7yIX
+0meBn5N9Y4m3wAmvokYeK7dU1oRSM397unmjgbwwgbkwDAYDVR0TAQH/BAIwADAO
+BgNVHQ8BAf8EBAMCB4AwFgYDVR0lAQH/BAwwCgYIKwYBBQUHAwEwFQYHK4E7gcR0
+BQEB/wQHcHJlc2VudDBLBgNVHSMERDBCgBTMNuas7QD5hDXYKS7k0WDN6ckMFqEq
+pCgwJjEPMA0GA1UECgwGQW1hem9uMRMwEQYDVQQpDAo0Mjk0OTY3Mjk2MB0GA1Ud
+DgQWBBRiexFm2K7Ou2dx4+c0LjQOsuqHJjAKBggqhkjOPQQDAgNIADBFAiAxH63Q
+eK26A9QPOkqi+5Hvrptpb9HRstSC6emJdaEB1QIhAKyhyLBPrG85QDoXrFcVZUA2
++StWnDVDGtgWM6tPz4Uw
+-----END CERTIFICATE-----
+)";
+
+static const char kX509MultipleCustomExtensionsCA[] = R"(
+-----BEGIN CERTIFICATE-----
+MIIByDCCAW2gAwIBAgIFAQAAAAAwCgYIKoZIzj0EAwIwJjEPMA0GA1UECgwGQW1h
+em9uMRMwEQYDVQQpDAo0Mjk0OTY3Mjk2MCIYDzIwMjUwMzMwMjExMzA1WhgPOTk5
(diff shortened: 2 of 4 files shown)


include/openssl/x509.h
```diff
@@ -2903,6 +2903,33 @@ OPENSSL_EXPORT int X509_STORE_CTX_set_purpose(X509_STORE_CTX *ctx, int purpose);
...
+// OIDs of any critical extensions that are within the certificates being
+// validated, that they wish to allow.
+//
+// To properly consume this feature, both |X509_V_FLAG_ALLOW_CUSTOM_CRIT_EXT|
+// and |X509_STORE_CTX_set_verify_crit_oids| must be set. See their
+// specific documentation for more details.
+OPENSSL_EXPORT int X509_STORE_CTX_add_custom_crit_oid(X509_STORE_CTX *ctx,
+                                                      ASN1_OBJECT *oid);
+
+// X509_STORE_CTX_verify_crit_oids is the callback signature for
```
Mark-Simulacrum, line 2919: I'm not sure how I feel about this callback shape. A caller setting this to validate the oid list *they also configured* is still the same feels relatively useless. A caller setting this to validate the rest of the certificate as part of the AWS-LC validation is plausible, but also not particularly necessary -- since they can just do their validation after or before the AWS-LC validation.

For example, s2n-tls already has a similar verification callback (https://github.com/aws/s2n-tls/blob/1c5798b82442067bace943f748f4f24ae1770bed/api/unstable/crl.h#L175) that's also a bit more powerful (exposes the full cert chain and optionally allows asynchronous validation).

I'd personally expect s2n to just set this callback to always return true and then apply any validation it wants to separately from the AWS-LC X509_verify call. So maybe we should just drop it from the interface to avoid the extra complexity?

Let's get s2n thoughts before any decision, of course.

goatgoose: Yeah, I agree. I'm not opposed to the callback, but I think we're planning on having s2n-tls consumers just set the list of OIDs and do their custom verification in our existing cert verification callback or after the handshake, so we'd likely implement this how Mark described and always return true here. Even if we were to expose a similar callback in s2n-tls, I'd be hesitant to forward the AWS-LC callback for this purpose since I think we'd struggle to make it async compatible if we ever needed to.

samuel40791765: We initially decided to do a callback because we're not just going to be exposing this feature to only s2n-tls users, but to general users of AWS-LC as well. Since this feature is actually relaxing a check against unknown critical extensions, there's the caveat that we're relaying on the consumer to do additional validation against their OIDs. We do trust that s2n-tls will be doing the correct validation during/before/after the handshake. 
However without the callback, there's not a particular mechanism in AWS-LC enforcing general consumers of the feature to actually do their OID validation. We want to put the fate of this check relaxation into the user's hands, so that they understand what they're doing. 

I do agree that there's a redundancy on with the `X509_V_FLAG_ALLOW_CUSTOM_CRIT_EXT` flag vs callback though. Ideally whatever we decided upon, I'd like to just keep 1 of either. If we decided against a callback to enforce validation, there has to be explicit documentation in AWS-LC asking the consumer to be aware of what they're doing and there's the slight concern that the general consumer may overlook this.

samuel40791765: True that certificate verification would still fail if encountering an unknown critical extension and the callback would force the checking mechanism into the users hands. I do have the concern where there's a more dangerous scenario where a consumer could just simply set the callback to return 1 and circumvent **all** checks instead.

Do let me know what you folks think about the current iteration, I've removed the flag in the current PR iteration so this is a clearer example of what we were initially envisioning. If we're still not keen on doing a callback with the OID setters in AWS-LC, we can consider falling back to the flag with the OID setters in AWS-LC. The flag's documentation will have a much stronger "DO NOT USE" caveat for all general consumers though.

crypto/x509/internal.h
```diff
@@ -344,6 +344,8 @@ struct x509_store_ctx_st {
   X509_STORE_CTX_verify_cb verify_cb;       // error callback
   X509_STORE_CTX_get_crl_fn get_crl;        // retrieve CRL
   X509_STORE_CTX_check_crl_fn check_crl;    // Check CRL validity
+  X509_STORE_CTX_verify_crit_oids
+      verify_crit_oids;  // Check custom critical oids
```
smittals2, line 348: Since we also name the stack/array below as `custom_crit_oids`, maybe instead of `verify_crit_oids` we can use `maybe verify_custom_crit_oids`?

crypto/x509/internal.h
```diff
@@ -359,6 +361,9 @@ struct x509_store_ctx_st {
 
   int current_crl_score;         // score of current CRL
 
+  // Array of allowed custom critical extension oids.
```
smittals2, line 364: nit: Array, should be stack?

crypto/x509/x509_vfy.c
```diff
@@ -122,6 +122,10 @@ static int cert_crl(X509_STORE_CTX *ctx, X509_CRL *crl, X509 *x);
 static int internal_verify(X509_STORE_CTX *ctx);
 
 static int null_callback(int ok, X509_STORE_CTX *e) { return ok; }
+static int null_verify_crit_oids_callback(X509_STORE_CTX *ctx, X509 *x509,
```
smittals2, line 125: maybe null_verify_custom_crit_oids_callback instead?

smittals2 (commented):
General question/comment around naming convention here. I'm seeing both "*_custom_crit_*" and "*_custom_known_crit*_" in the code. Is there a distinction between the two that makes a given crit oid "known", or should this naming be standardized?

crypto/x509/internal.h
smittals2: and maybe also add a _cb at the end of the name so we know its a callback

crypto/x509/x509_vfy.c
```diff
@@ -571,8 +614,11 @@ static int check_chain_extensions(X509_STORE_CTX *ctx) {
   // Check all untrusted certificates
   for (int i = 0; i < ctx->last_untrusted; i++) {
     X509 *x = sk_X509_value(ctx->chain, i);
-    if (!(ctx->param->flags & X509_V_FLAG_IGNORE_CRITICAL) &&
-        (x->ex_flags & EXFLAG_CRITICAL)) {
+    if ((!(ctx->param->flags & X509_V_FLAG_IGNORE_CRITICAL) &&
+         (x->ex_flags & EXFLAG_CRITICAL)) &&
```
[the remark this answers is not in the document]
smittals2: Can we also add a comment here about this maybe, I was also confused at first. I read the if condition as "if `IGNORE_CRITICAL` is not set, and `EXFLAG_CRITICAL` is set (which to me implied any critical extensions, not just unhandled ones), AND either `check_custom_known_critical_extension` or `ctx->verify_crit_oids` fails", we are in an error state. 

Since `check_custom_known_critical_extension` fails immediately if `ctx->custom_crit_oids` is NULL, if `EXFLAG_CRITICAL` were to mean any critical ext and not just unhandled ones, this if condition would be erroneous. 

Just to confirm for my own clarity, at this point in code, are we expecting that any "known" critical oids have already been parsed/validated. So if EXFLAG_CRITICAL is still set here, we know that it's only possible because the customer has set custom oids and can therefore fail otherwise?

crypto/x509/x509_vfy.c
```diff
@@ -571,8 +614,11 @@ static int check_chain_extensions(X509_STORE_CTX *ctx) {
...
+        // Do check for enabling custom unknown critical extensions.
+        (!check_custom_known_critical_extension(ctx, x) ||
+         !ctx->verify_crit_oids(ctx, x, ctx->custom_crit_oids))) {
```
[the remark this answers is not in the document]
smittals2: is ctx->custom_crit_oids not the subset already? Or can ctx->custom_crit_oids have certain custom extensions that are not applicable to the given cert x (and might be to another cert in the chain), so you are recommending filtering further here?

crypto/x509/internal.h
samuel40791765: I originally wanted to name the callback function that, but the function name that we're exposing was getting way too long for my liking (`X509_STORE_CTX_set_verify_custom_crit_oids`).

`_cb`'s a good idea, I'll add that to the callback name. As for `verify_custom_crit_oids`, it should be fine to name the internal callback variable name that.

crypto/x509/x509_test.cc
```diff
@@ -8208,3 +8269,132 @@ TEST(X509Test, Trust) {
                               {intermediate.normal.get()}, {},
                               /*flags=*/0, set_server_trust));
 }
+
+// A brief validation against the |oids| expected to be done by the consumer.
+// This example simulates the consumer checking that the certificate has the
+// correct number of unknown extensions and there aren't any duplicates.
+static int verify_crit_oids_callback(X509_STORE_CTX *ctx, X509 *x509,
```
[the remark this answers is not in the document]
samuel40791765: Not relevant and I only picked this testing purposes. The expectations of the consumer are to validate that these are the OIDs that they wish to set.

crypto/x509/x509_vfy.c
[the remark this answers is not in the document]
samuel40791765: `ctx->custom_crit_oids` contains **all** of the custom critical oids that the user has set. I believe @skmcgrail's asking me to change some logic here to return only the custom critical oids that were found for this specific cert. 

Agreed that this would be more valuable than a list the user has set themselves. I'll rework some of the logic around here to do that.

crypto/x509/x509_vfy.c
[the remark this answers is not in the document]
samuel40791765: I was a bit hesitant to do so, mainly because the state of `EXFLAG_CRITICAL` is maintained along with the `x` certificate. If that state is cleared, there is now an implication that critical extension is supported and the certificate can be accepted. I was initially afraid of the user not being aware that there were typically unknown critical extensions present in the cert. But if they're already running `X509_verify_cert` with this feature, I guess it would make sense that that state is cleared later on.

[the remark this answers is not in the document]
samuel40791765: `EXFLAG_CRITICAL` simply indicates an unsupported critical extension, so a critical extension that is already supported wouldn't trigger this. This flag [is added](https://github.com/aws/aws-lc/blob/e4885d5e22f7dc482dd6bfa713b0e1b763b5c538/crypto/x509/v3_purp.c#L402) during the initial parsing of the certificate.

`check_custom_known_critical_extension` is supposed to fail under the circumstance you mentioned. Typical OpenSSL/AWS-LC behavior is to fail when an unsupported critical extension is found (hence the flag). The user has to set up `ctx->custom_crit_oids` with all their "known" unsupported critical extensions they wish to work around here.

At this point in the code, we only know if there are unknown critical extensions present with the `EXFLAG_CRITICAL` flag. We could remove the flag after the first validation against this has succeeded, but I'd prefer to do the custom OID validation regardless of whether the flag has been set.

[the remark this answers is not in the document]
smittals2: Thank you for those details, that clears up a lot for me! It would be great if we can add a comment here explaining EXFLAG_CRITICAL/what we are checking to improve readability in this func :)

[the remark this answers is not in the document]
samuel40791765: No problem, I think it's valuable. Will do :).

samuel40791765 (commented):
> General question/comment around naming convention here. I'm seeing both "custom_crit" and "_custom_known_crit_" in the code. Is there a distinction between the two that makes a given custom crit oid "known", or should this naming be standardized?

Sorry for the confusion, I was fighting myself with using `custom_crit` and `custom_known_crit` since the exported function names were getting a bit too long in the initial drafts... Both mean the same thing, I believe we'll be moving forward with `custom_crit`, so I'll change all places to reflect that.

crypto/x509/x509_vfy.c
```diff
@@ -561,6 +568,63 @@ static int get_issuer(X509 **issuer, X509_STORE_CTX *ctx, X509 *x) {
...
+  // properly handled and we pass the ones that were found to the caller.
+  if (!ctx->verify_custom_crit_oids(ctx, x, found_exts)) {
+    return 0;
+  }
+
+  // Remove the |EXFLAG_CRITICAL| flag from |x|, now that all unknown
+  // critical extensions have been handled.
+  x->ex_flags &= ~EXFLAG_CRITICAL;
+
+  return 1;
```
skmcgrail, line 625: Missing free of `found_exts`. I'm not sure if it should be the caller's responsibility, nor do I see the value in it outliving this function?

samuel40791765: Fixed, should be on us to free it.

skmcgrail (approved):

crypto/x509/x509_test.cc
```diff
@@ -8208,3 +8269,142 @@ TEST(X509Test, Trust) {
...
+                                     STACK_OF(ASN1_OBJECT) *oids) {
+  if (oids == nullptr) {
+    return 0;  // Fail if no OIDs provided
+  }
+  size_t known_oid_count = sk_ASN1_OBJECT_num(oids);
+  ;
```
skmcgrail, line 8282: np
```suggestion
```

skmcgrail (approved):

smittals2 (approved):

PR #2380, opened 2025-04-30
Rework memory BIOs and implement BIO_seek

nhatnghiho:
### Issues:
Resolves #CryptoAlg-3111

### Description of changes: 
This PR creates a wrapper around `BUF_MEM` and adds a new field `off` to indicate the current offset of the read pointer. This enables us to implement `BIO_seek`, which is also in this PR.

### Testing:
Unit tests are modified and added to cover more test cases.


crypto/bio/bio_mem.c
```diff
@@ -98,56 +105,87 @@ BIO *BIO_new_mem_buf(const void *buf, ossl_ssize_t len) {
...
     b->data = NULL;
   }
   BUF_MEM_free(b);
   bio->ptr = NULL;
+
+  OPENSSL_free(bbm);
   return 1;
 }
 
+static int mem_buf_sync(BIO *bio) {
```
WillChilds-Klein, line 151: suggest adding either a comment or check that `bio` must not be r/o

crypto/bio/bio_mem.c
```diff
@@ -222,6 +273,18 @@ static long mem_ctrl(BIO *bio, int cmd, long num, void *ptr) {
         }
       }
       break;
+    case BIO_C_FILE_SEEK:
+      if (num < 0 || (size_t)num > bbm->off + b->length)
```
WillChilds-Klein, line 277: nit: prefer `{ ... }` around conditional bodies

crypto/bio/bio_mem.c
```diff
@@ -231,6 +294,9 @@ static long mem_ctrl(BIO *bio, int cmd, long num, void *ptr) {
     case BIO_CTRL_INFO:
       ret = (long)b->length;
       if (ptr != NULL) {
+        if (!(bio->flags & BIO_FLAGS_MEM_RDONLY)) {
+          mem_buf_sync(bio);
```
WillChilds-Klein, line 298: why do we `mem_buf_sync` here? OpenSSL [doesn't](https://github.com/openssl/openssl/blob/995e9489e62ff1965553c4e127183565ccfe4265/crypto/bio/bss_mem.c#L294).

crypto/bio/bio_mem.c
```diff
@@ -98,56 +105,87 @@ BIO *BIO_new_mem_buf(const void *buf, ossl_ssize_t len) {
...
     ret = (int)b->length;
   }
 
   if (ret > 0) {
-    OPENSSL_memcpy(out, b->data, ret);
+    size_t off = bbm->off;
+    if (bio->flags & BIO_FLAGS_MEM_RDONLY) {
+      off = 0;
+    }
+    OPENSSL_memcpy(out, &b->data[off], ret);
```
WillChilds-Klein, line 184: would it be simpler if we just always use `bbm->off` on L184 and eliminate the `b->data += ret` advancement on L188 as well as other manipulation of `b->data`?

it seems a bit odd that we modify the value of `b->data`, but only in r/o mode.

if not, suggest adding a comment why you have to override `off = 0` in the r/o case

crypto/bio/bio_mem.c
```diff
@@ -65,10 +65,16 @@
 
 #include "../internal.h"
 
+typedef struct bio_buf_mem_st {
+  struct buf_mem_st *buf;   /* allocated buffer */
+  size_t off;
```
WillChilds-Klein, line 70: minor nit: suggest a more descriptive name like `read_off`

crypto/bio/bio_test.cc
```diff
@@ -265,21 +265,30 @@ TEST(BIOTest, MemReadOnly) {
...
+
+  ASSERT_TRUE(BIO_mem_contents(bio.get(), &contents, &len));
+  EXPECT_EQ(Bytes(contents, len), Bytes("pqrs"));
   EXPECT_EQ(BIO_eof(bio.get()), 0);
 
   // Read the remainder of the buffer.
   ret = BIO_read(bio.get(), buf, sizeof(buf));
   ASSERT_GT(ret, 0);
-  EXPECT_EQ(Bytes(buf, ret), Bytes("mno"));
+  EXPECT_EQ(Bytes(buf, ret), Bytes("pqrs"));
```
WillChilds-Klein, line 291: should we ` EXPECT_EQ(BIO_eof(bio.get()), 1);` here?

crypto/bio/bio_mem.c
nhatnghiho: This behavior of modifying `b->data` in R/O mode is pretty consistent in both BoringSSL and OpenSSL. 

In R/O mode, the getter functions (`BIO_get_mem_ptr`, `BIO_get_mem_data`, and `BIO_mem_contents`) expect that the modified `b->data` to be returned. This is especially important for `BIO_get_mem_ptr` that's returning the entire `BUF_MEM` struct, not just `b->data`. Therefore, I decided to keep `b->data += ret` for better backward compatibility.

crypto/bio/bio_mem.c
nhatnghiho: Good catch! Glossed over this part for OpenSSL

crypto/bio/bio_mem.c
nhatnghiho: Added comments + refactored the code a little to make it clear that `read_off` is to keep track of the number of bytes already read from the original buffer, not necessarily the relative offset from the current `b->data` ptr

crypto/bio/bio_mem.c
```diff
@@ -65,10 +65,16 @@
 
 #include "../internal.h"
 
+typedef struct bio_buf_mem_st {
+  struct buf_mem_st *buf;   /* allocated buffer */
+  size_t read_off;          /* number of bytes already read from the buffer */
+} BIO_BUF_MEM;
+
 
 BIO *BIO_new_mem_buf(const void *buf, ossl_ssize_t len) {
```
justsmth, line 74: We cast `len` to a `size_t` below. Although unlikely, we should verify that `len <= SIZE_MAX` to avoid overflow.

crypto/bio/bio_mem.c
```diff
@@ -81,7 +87,8 @@ BIO *BIO_new_mem_buf(const void *buf, ossl_ssize_t len) {
     return NULL;
   }
 
-  b = (BUF_MEM *)ret->ptr;
+  bbm = (BIO_BUF_MEM *)ret->ptr;
+  b = bbm->buf;
   // BIO_FLAGS_MEM_RDONLY ensures |b->data| is not written to.
   b->data = (void *)buf;
   b->length = size;
```
justsmth, line 94: There are several places that we cast `b->length` to an `int`.  We should verify that `size < INT_MAX` here.

crypto/bio/bio_mem.c
```diff
@@ -98,56 +105,96 @@ BIO *BIO_new_mem_buf(const void *buf, ossl_ssize_t len) {
...
     b->data = NULL;
   }
   BUF_MEM_free(b);
   bio->ptr = NULL;
+
+  OPENSSL_free(bbm);
   return 1;
 }
 
+static int mem_buf_sync(BIO *bio) {
```
justsmth, line 151: * This function should return void. Caller never check the return value (and this function always returns 0 anyways).
* Is the impact of a failure of this function that the caller would be unable to append new data after the `b->length` offset?

crypto/bio/bio_mem.c
```diff
@@ -98,56 +105,96 @@ BIO *BIO_new_mem_buf(const void *buf, ossl_ssize_t len) {
...
 
   if (ret > 0) {
-    OPENSSL_memcpy(out, b->data, ret);
+    // If the buffer is read-only, |data| is already pointing to the correct
+    // read position. If the buffer is writeable, |data| is pointing to the start
+    // of the buffer, so we need to add |read_off| to get the correct read
+    // position.
+    if (bio->flags & BIO_FLAGS_MEM_RDONLY) {
+      OPENSSL_memcpy(out, b->data, ret);
+    } else {
```
justsmth, line 190: Should the `b->data += ret;` below be moved up here?

crypto/bio/bio_mem.c
nhatnghiho: > Is the impact of a failure of this function that the caller would be unable to append new data after the b->length offset?

This function shouldn't fail though? If the buffer is empty/null or is read-only, there's nothing for `mem_buf_sync` to sync. Otherwise, `memmove` would be executed, which shouldn't fail either. If the caller wants to append new data, it already has checks to make sure the buffer is neither null nor read-only

crypto/bio/bio_mem.c
nhatnghiho: Looked at the code again. We only cast b->length to an int in `mem_read` and `mem_gets` when the requested number of bytes (which is an `int`) is larger than `b->length`. So I think it's safe not to have this check. 

In `BUF_MEM_append`, the value of `b->length` is also modified, so it's better to have the caster handling the overflow anyways.

crypto/bio/bio_mem.c
```diff
@@ -120,34 +134,64 @@ static int mem_free(BIO *bio) {
...
   if ((size_t)ret > b->length) {
     ret = (int)b->length;
   }
 
   if (ret > 0) {
-    OPENSSL_memcpy(out, b->data, ret);
+    // If the buffer is read-only, |data| is already pointing to the correct
+    // read position. If the buffer is writeable, |data| is pointing to the start
+    // of the buffer, so we need to add |read_off| to get the correct read
+    // position.
```
justsmth, line 184: Treating the pointer differently depending on whether the buffer is read-only or not isn't necessary. This adds some complexity to the logic that I think can be removed.

crypto/bio/bio_mem.c
nhatnghiho: Nvm! Talked to Justin and refactored the code to simplify the way we handle the memory buffer in R/O mode

include/openssl/bio.h
```diff
@@ -1116,7 +1116,7 @@ struct bio_st {
   // |BIO_set_close|. For example, whether an fd BIO closes the underlying fd
   // when it, itself, is closed.
   int shutdown;
-  int flags;
+  unsigned int flags;
```
justsmth, line 1119: Sorry, this change [breaks rust-openssl](https://github.com/aws/aws-lc/actions/runs/15011694254/job/42181514522?pr=2380#step:7:2088) consumers: 😞
```
 error[E0308]: mismatched types
   --> openssl/src/ssl/bio.rs:271:28
    |
271 |             (*bio).flags = flags;
    |             ------------   ^^^^^ expected `u32`, found `i32`
    |             |
    |             expected due to the type of this binding
```

justsmth (dismissed):

WillChilds-Klein (dismissed):

WillChilds-Klein (approved):

PR #2382, opened 2025-04-30
No need to replace semicolons in s2n-bignum

justsmth:
### Description of changes: 
After [recent changes in s2n-bignum](https://github.com/awslabs/s2n-bignum/pull/213), replacing the semi-colons during the build is no longer necessary.


justsmth:
#2385 does more and does it better.

PR #2386, opened 2025-05-01
Fix ImplDispatchTest for 32-bit x86 build

dkostic:
### Description of changes: 
ImplDispatchTest for 32-bit x86 architectures for SHA256 and AES-GCM
was not correct. The issue started manifesting itself when the CI
underlying hardware upgraded to a newer x86 CPU generation.

The fix is that the test shouldn't expect SHA and AES HW instructions
to be used for 32-bit x86 build.

### Call-outs:

### Testing:


justsmth (approved):
Tests are now passing.

skmcgrail (approved):

PR #2397, opened 2025-05-06
Fix PRNG model test

torben-hansen:
### Description of changes: 

The PRNG model has been disabled for a while in the RAGDOLL branch. This PR fixes the PRNG model test for the new implementation. It only captures system calls, so it was pretty straightforward to model the correct flow. Some changes were required to capture the real trace due to Jitter Entropy and other stuff... This should all be explained in inline comments in the source code. Few details below.

The following assumption is no longer true

> It's assumed that any arguments to open(2) are constants in read-only memory

because of system calls originating from Jitter Entropy. This requires a bit more care when parsing the filename argument to `open/openat`. Take the opportunity to simplify the string handling a bit by changing to c++ native std strings.

When reading registers through ptrace, sometimes they are sign-extended(?). Need to cater for that.

### DIFF
--- a/crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy.c
+++ b/crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy.c
@@ -151,7 +151,7 @@ static void tree_jitter_drbg_maybe_get_pred_resistance(
 
   if (tree_jitter_drbg->is_global == 0 &&
       tree_jitter_drbg->ube_protection != 1) {
-    CRYPTO_sysrand_for_seed(pred_resistance, RAND_PRED_RESISTANCE_LEN);
+    CRYPTO_sysrand(pred_resistance, RAND_PRED_RESISTANCE_LEN);
     *pred_resistance_len = RAND_PRED_RESISTANCE_LEN;
   }
 }
--- a/crypto/rand_extra/urandom_test.cc
+++ b/crypto/rand_extra/urandom_test.cc
@@ -24,6 +24,7 @@
 #include "../ube/snapsafe_detect.h"
 
 #if defined(OPENSSL_RAND_URANDOM) && \
+    defined(OPENSSL_X86_64) && \
     !defined(BORINGSSL_SHARED_LIBRARY) && \
     !defined(BORINGSSL_UNSAFE_DETERMINISTIC_MODE) && \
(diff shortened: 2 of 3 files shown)


skmcgrail (approved):

crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy.c
```diff
@@ -151,7 +151,7 @@ static void tree_jitter_drbg_maybe_get_pred_resistance(
 
   if (tree_jitter_drbg->is_global == 0 &&
       tree_jitter_drbg->ube_protection != 1) {
-    CRYPTO_sysrand_for_seed(pred_resistance, RAND_PRED_RESISTANCE_LEN);
+    CRYPTO_sysrand(pred_resistance, RAND_PRED_RESISTANCE_LEN);
```
dkostic, line 154: what's the motivation for this change?

crypto/rand_extra/urandom_test.cc
```diff
@@ -193,6 +174,34 @@ static const unsigned GETRANDOM_ERROR = 16;
 static const unsigned URANDOM_ERROR = 32;
 static const unsigned NEXT_FLAG = 64;
 
+// ReadString parses string at address |addr| in child process |pid|.
+static std::string ReadString(pid_t pid, unsigned long addr) {
+  std::string result;
+  size_t i = 0;
+  while (i < 4096) { // Don't accept paths longer than this.
+    long data = ptrace(PTRACE_PEEKDATA, pid, addr + i, NULL);
+    if (data == -1 && errno) {
```
dkostic, line 183: what's `errno`??

crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy.c
torben-hansen: Because `CRYPTO_sysrand_for_seed` is now equivalent to `CRYPTO_sysrand` and I didn't want to factor that in to the PRNG model. Later, I plan on removing `CRYPTO_sysrand_for_seed` completely.

crypto/rand_extra/urandom_test.cc
torben-hansen: It's defined by the std library. It carries information about which error was observed by the system call: https://man7.org/linux/man-pages/man3/errno.3.html

PR #2419, opened 2025-05-13
Add password prompting support & EVP_read_pw_string

smittals2:
### Issues:
`CryptoAlg-3060`

### Description of changes: 
1. Add console* suite of functions to interact with the terminal on both linux and windows with fallbacks to stdin and stdout
2. Add `EVP_read_pw_string`
3. Align `PEM_def_callback`'s behavior with OpenSSL. Previously, this function would fail if the `userdata` param was NULL. This is different from openSSL which would prompt a user for a password in the same scenario. Now `PEM_def_callback` will correctly prompt the user if `userdata` is NULL. 
4. Remove temp workaround in the OpenSSL req tool and use prompting from `PEM_def_callback`. Add carriage return support so this tool works on windows.

### Call-outs:
- The tests are disabled on android, tmpfile() doesn't work on androids when running form APK. 

### Testing:
We only test the stdin stdout routes in GTest. Console functionality was manually tested. 


crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,301 @@
...
+// set SIGWINCH handler to default so our workflow is not
+// triggered on user resizing of window
+# if defined(SIGWINCH) && defined(SIG_DFL)
+    signal(SIGWINCH, SIG_DFL);
+# endif
+}
+
+static void popsig(void) {
+# if !defined(_WIN32)
+    int i;
```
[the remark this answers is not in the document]
justsmth: Or just initialize it in the for loop:
```
for (int i = 1; i < NUM_SIG; i++) {
```

crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,301 @@
...
+  TTY_STRUCT tty_orig, tty_new;
+#endif
+
+FILE *tty_in, *tty_out;
+int is_a_tty;
+
+
+/* Internal functions to read a string without echoing */
+static int read_till_nl(FILE *in) {
+#  define SIZE 4
```
justsmth, line 65:
```
const size_t SIZE = 4;
```

crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,301 @@
...
+#else
+  TTY_STRUCT tty_orig, tty_new;
+#endif
+
+FILE *tty_in, *tty_out;
+int is_a_tty;
+
+
+/* Internal functions to read a string without echoing */
+static int read_till_nl(FILE *in) {
```
justsmth, line 64: NP: Rename to something like `discard_line_remainder`? This function's purpose appears to be to discard any remaining characters until it reaches a newline.

crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,301 @@
...
+# ifdef SIGACTION
+static struct sigaction savsig[NUM_SIG];
+# else
+static void (*savsig[NUM_SIG]) (int);
+# endif
+
+static int read_till_nl(FILE * in);
+static void recsig(int signal);
+static void pushsig(void);
+static void popsig(void);
```
justsmth, line 51: NP: These function declarations aren't needed if you reorder the functions below to have `recsig` first.

crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,301 @@
...
+# endif
+}
+
+static void recsig(int signal) {
+    intr_signal = signal;
+}
+
+/* Console management functions */
+int openssl_console_open(void) {
+    CRYPTO_STATIC_MUTEX_lock_write(&console_global_mutex);
```
justsmth, line 131: It would be better to have the caller manage the locking so that both the locking and unlocking of this are visible in the same scope.

crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,301 @@
...
+    if ((tty_out = fopen(DEV_TTY, "w")) == NULL) {
+        tty_out = stderr;
+    }
+    if (TTY_get(fileno(tty_in), &tty_orig) == -1) {
+        if (errno == ENOTTY || errno == EINVAL || errno == ENXIO || errno == EIO
+          || errno == EPERM || errno == ENODEV) {
+            is_a_tty = 0;
+          } else {
+              OPENSSL_PUT_ERROR(PEM, ERR_R_INTERNAL_ERROR);
+              return 0;
```
justsmth, line 147: If you return an error, then you must release the lock as it appears that the caller is unaware that a write lock was acquired. (See comment above about moving the locking up to the caller.)

crypto/pem/pem_passwd.c
justsmth: Were we to move the locking to the caller, we might still want to have an `assert` statement that verifies this lock has been acquired. We could add the following function to our library and call it from an `assert` here:
```
int CRYPTO_STATIC_MUTEX_is_write_locked(struct CRYPTO_STATIC_MUTEX *lock) {
  int result = pthread_rwlock_tryrdlock(&lock->lock);
    
  if (result == 0) {
    // If we successfully acquired the lock, it wasn't locked
    // Release it immediately and return false
    pthread_rwlock_unlock(&lock->lock);
    return 0;  // Not write-locked
  } else if (result == EBUSY) {
    // Lock is held by another thread
    return 1;  // Is write-locked
  } else {
    // Handle other errors (EINVAL, etc.)
    return -1; // Error condition
  }
}
```

crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,301 @@
...
+# else
+static void (*savsig[NUM_SIG]) (int);
+# endif
+
+static int read_till_nl(FILE * in);
+static void recsig(int signal);
+static void pushsig(void);
+static void popsig(void);
+
+#if defined(_WIN32)
```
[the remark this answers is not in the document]
smittals2: You are right it only ends up getting used once. I will change it to just work with OpenSSL WINDOWS

crypto/pem/pem_passwd.c
smittals2: Fair enough, these console APIs aren't being exposed to the user yet (like the UI api) so they don't necessarily need to manage the locking and we can offload that to the calling function for now. This would make the design cleaner for our current use case and ensure we appropriately let go of the lock.

crypto/pem/pem_passwd.c
smittals2: Converting it to a var made the compiler complain for some reason about variable length arr. Since this isn't really needed as a var or macro, I just got rid of it.

crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,309 @@
...
+    CRYPTO_STATIC_MUTEX_lock_write(&console_global_mutex);
+}
+
+void openssl_console_release_mutex(void) {
+    CRYPTO_STATIC_MUTEX_unlock_write(&console_global_mutex);
+}
+
+int openssl_console_open(void) {
+    is_a_tty = 1;
+    if (CRYPTO_STATIC_MUTEX_is_write_locked(&console_global_mutex) != 1) {
```
justsmth, line 131: Since this is an internal function and this is just a "sanity" check for the caller, this condition should be in an `assert`.

crypto/internal.h
```diff
@@ -705,6 +705,14 @@ OPENSSL_EXPORT void CRYPTO_STATIC_MUTEX_unlock_read(
 OPENSSL_EXPORT void CRYPTO_STATIC_MUTEX_unlock_write(
     struct CRYPTO_STATIC_MUTEX *lock);
 
+// CRYPTO_STATIC_MUTEX_is_write_locked checks whether |lock| has an active write
```
justsmth, line 708: I would put macros guarding this declaration so it's only available when `NDEBUG` is defined (i.e., for `assert` statements). This function is not needed when locks are properly handled (at least in our current use-case). We only need it to sanity check lock handling while testing.

smittals2: I'm getting symbol not found errors in the assert statements if I guard the declaration in NDEBUG (although per some research assert statements should be compiled out in non-debug builds, not sure what the issue is).

crypto/pem/pem_test.cc
```diff
@@ -473,3 +478,226 @@ TEST(PEMTest, WriteReadTraditionalPem) {
...
+    close(original_stdin);
+    close(original_stderr);
+    
+    // Close temp files
+    if (stdin_file) {
+      fclose(stdin_file);
+    }
+    if (stderr_file) {
+      fclose(stderr_file);
+    }
```
justsmth, line 526: * I believe it's essentially [undefined behavior](https://man7.org/linux/man-pages/man3/stdin.3.html) to close file-descriptors (e.g., `original_stdin` and `original_stderr`) that are wrapped by a FILE struct. Can't these just be left open?
```
Note that mixing use of FILEs and raw file descriptors can produce
unexpected results and should generally be avoided.  (For the
masochistic among you: POSIX.1, section 8.2.3, describes in detail
how this interaction is supposed to work.)  A general rule is that
file descriptors are handled in the kernel, while stdio is just a
library.
```
* The [documentation for `dup2`](https://man7.org/linux/man-pages/man2/dup.2.html) indicates that it will close `stdin_file` and `stderr_file` as part of it operation.
```
If the file descriptor newfd was previously open, it is closed
before being reused; the close is performed silently (i.e., any
errors during the close are not reported by dup2()).
```
* I would use the `STDIN_FILENO` and `STDERR_FILENO` when referring to their file descriptors. These values are constant.

crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,304 @@
...
+#include <openssl/buf.h>
+#include <openssl/err.h>
+#include <openssl/evp.h>
+#include <openssl/pem.h>
+#include <openssl/x509.h>
+
+#include "internal.h"
+#include "../internal.h"
+
+# include <signal.h>
```
skmcgrail, line 18: Minor bit but you can correct the white space (remove it) for these preprocessor macros like`# include`? I know this is an OpenSSL thing but it doesn’t much our style and is inconsistent throughout this file.

crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,304 @@
...
+        tty_new = tty_orig;
+        SetConsoleMode(GetStdHandle(STD_INPUT_HANDLE), tty_new);
+    }
+#endif
+    return 1;
+}
+
+int openssl_console_write(const char *str) {
+    assert(CRYPTO_STATIC_MUTEX_is_write_locked(&console_global_mutex));
+    if (fputs(str, tty_out) < 0 || fflush(tty_out) != 0) {
```
skmcgrail, line 221: Might be worth putting an OpenSSL error on the queue to indicate we got an error from the system? Not sure if it's possible to output the reason retried from `ferror` either?

crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,304 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
skmcgrail, line 1: Would it make sense to put all these console API implementations into a separate hierarchy folder then pem? I know pem is the primary use case currently, but it feels off to have all this terminal management code in here. I also don't care of the `ui` moniker used by OpenSSL either. `console`, `terminal`, `prompt` are maybe some suggestions? It's not a major deal, just feels like a weirdly out of place in this code area. :)

crypto/pem/pem_passwd.c
skmcgrail: Similar for the `fgets` terminal/device interactions.

crypto/pem/pem_passwd.c
smittals2: Yea that's fair. I don't really have a strong preference on this. The reason I put it in with PEM is like you noted, that is the primary use case. And that should remain the primary use case unless for some reason we decided to fully support the UI interface - at which point a lot more functionality would be needed and a separate folder structure is certainly justified. Happy to move this to a separate folder if that makes organization clearer in the current state as well.

crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,318 @@
...
+}
+
+int openssl_console_write(const char *str) {
+    assert(CRYPTO_STATIC_MUTEX_is_write_locked(&console_global_mutex));
+    if (fputs(str, tty_out) < 0 || fflush(tty_out) != 0) {
+        OPENSSL_PUT_ERROR(PEM, PEM_R_PROBLEMS_GETTING_PASSWORD);
+        if (ferror(tty_out)) {
+            ERR_add_error_data(2, "System error: ", strerror(errno));
+        }
+        return 0;
```
justsmth, line 232: Here and other places we use `ferror`, we should call `clearerr` after to avoid it affecting subsequent attempts.

crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,318 @@
...
+        tty_in = stdin;
+        tty_out = stderr;
+        is_a_tty = 0;
+    }
+#endif
+    return 1;
+}
+
+int openssl_console_close(void) {
+    if (tty_in != stdin) {
```
justsmth, line 179: This function should also assert that the lock is being held:
```
    assert(CRYPTO_STATIC_MUTEX_is_write_locked(&console_global_mutex));
```

crypto/pem/pem_passwd.c
```diff
@@ -0,0 +1,318 @@
...
+            }
+            wresult[numread] = '\0';
+            if (WideCharToMultiByte(CP_UTF8, 0, wresult, -1, buf, maxsize + 1, NULL, 0) > 0) {
+                p = buf;
+            }
+            OPENSSL_cleanse(wresult, sizeof(wresult));
+        }
+    } else
+#endif
+    p = fgets(buf, maxsize, tty_in);
```
justsmth, line 279: NP: The syntax here is confusing. Could this be made more explicit, even though it would be redundant?

crypto/pem/pem_test.cc
smittals2: I'll just revert to fileno since we have to use it anyway :)

crypto/pem/pem_passwd.c
smittals2: Do you just mean add all the logic there to an else block? So that there's two copies of the reading from terminal code?

crypto/pem/pem_passwd.c
smittals2: Moved to a new file, I think its the cleaner route - might as well set it up separated incase we have to expand later.

crypto/pem/pem_lib.c
```diff
@@ -785,13 +784,37 @@ int PEM_read_bio(BIO *bp, char **name, char **header, unsigned char **data,
 }
 
 int PEM_def_callback(char *buf, int size, int rwflag, void *userdata) {
-  if (!buf || !userdata || size < 0) {
+  if (!buf || size < 0) {
```
justsmth, line 787: * Fail on `size <= 0`?
* Since this is a public function, should we proactively zero the `buf` to help prevent its current contents from being considered password were a failure to go unchecked by the caller?

crypto/fipsmodule/evp/evp.c
```diff
@@ -156,6 +158,68 @@ int EVP_PKEY_cmp(const EVP_PKEY *a, const EVP_PKEY *b) {
...
+}
+
+int EVP_read_pw_string(char *buf, int length, const char *prompt, int verify) {
+  return EVP_read_pw_string_min(buf, 0, length, prompt, verify);
+}
+
+int EVP_read_pw_string_min(char *buf, int min_length, int length,
+                           const char *prompt, int verify) {
+  int ret = -1;
+  char verify_buf[1024];
```
justsmth, line 172: * Proactively zero the caller's `buf`?
* Also `OPENSSL_cleanse(verify_buf, sizeof(verify_buf`)
* Fail if `min_length >= 1024`
* Fail if `min_length >= length`

crypto/fipsmodule/evp/evp.c
```diff
@@ -156,6 +158,68 @@ int EVP_PKEY_cmp(const EVP_PKEY *a, const EVP_PKEY *b) {
...
+  }
+
+  if (verify) {
+    openssl_console_write("Verifying - ");
+    openssl_console_write(prompt);
+
+    ret = openssl_console_read(verify_buf, min_length, sizeof(verify_buf), 0);
+
+    if (ret == 0) {
+      if (strcmp(buf, verify_buf) != 0) {
```
justsmth, line 205: * Use `strncmp`
```
if (strncmp(buf, verify_buf, length >= 1024 ? 1024 : length) != 0) {
```

crypto/console/console.c
```diff
@@ -0,0 +1,325 @@
...
+        goto error;
+    }
+    phase = 2;
+
+    buf[0] = '\0';
+#if defined(OPENSSL_WINDOWS)
+    if (is_a_tty) {
+        DWORD numread;
+        // for now assuming UTF-8....
+        WCHAR wresult[BUFSIZ];
```
justsmth, line 265: zero-initialize

crypto/console/console.c
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    return 1;
+}
+
+// returns 0 on success, -1 on error, -2 if interrupt signal received
+int openssl_console_read(char *buf, int minsize, int maxsize, int echo) {
+    int ok = 0;
+    char *p = NULL;
+    int echo_eol = !echo;
```
justsmth, line 245: Fail if `minsize > maxsize`.

crypto/console/console.c
```diff
@@ -0,0 +1,325 @@
...
+        WCHAR wresult[BUFSIZ];
+
+        if (ReadConsoleW(GetStdHandle(STD_INPUT_HANDLE),
+                     wresult, maxsize, &numread, NULL)) {
+            if (numread >= 2 && wresult[numread-2] == L'\r' &&
+                wresult[numread-1] == L'\n') {
+                wresult[numread-2] = L'\n';
+                numread--;
+            }
+            wresult[numread] = '\0';
```
justsmth, line 274: NP: Since this is `WCHAR`, the literal should be `L'\0'`.

crypto/console/console.c
```diff
@@ -0,0 +1,325 @@
...
+
+static void recsig(int signal) {
+    intr_signal = signal;
+}
+
+static int discard_line_remainder(FILE *in) {
+    char buf[5];
+
+    do {
+        if (!fgets(buf, 4, in)) {
```
justsmth, line 83: NP: you can give `fgets` the actual buffer size:
```
if (!fgets(buf, sizeof(buf), in)) {
```

crypto/console/console.c
```diff
@@ -0,0 +1,325 @@
...
+    return 1;
+}
+
+int openssl_console_write(const char *str) {
+    assert(CRYPTO_STATIC_MUTEX_is_write_locked(&console_global_mutex));
+    if (fputs(str, tty_out) < 0 || fflush(tty_out) != 0) {
+        OPENSSL_PUT_ERROR(PEM, PEM_R_PROBLEMS_GETTING_PASSWORD);
+        if (ferror(tty_out)) {
+            ERR_add_error_data(2, "System error: ", strerror(errno));
+            clearerr(tty_in);
```
justsmth, line 233:
```
clearerr(tty_out)
```

include/openssl/pem.h
```diff
@@ -410,11 +410,15 @@ OPENSSL_EXPORT int PEM_ASN1_write(i2d_of_void *i2d, const char *name, FILE *fp,
                                   const unsigned char *pass, int pass_len,
                                   pem_password_cb *callback, void *u);
 
-// PEM_def_callback treats |userdata| as a string and copies it into |buf|,
-// assuming its |size| is sufficient. Returns the length of the string, or 0
-// if there is not enough room. If either |buf| or |userdata| is NULL, 0 is
-// returned. Note that this is different from OpenSSL, which prompts for a
-// password.
+// PEM_def_callback provides a password for PEM encryption/decryption operations.
```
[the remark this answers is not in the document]
smittals2: I see, there's more use cases for the PEM_def_callback. I'll update the comments here to talk about the other 2 funcs as well. Didn't touch PEM_do_header but I'll add documentation for it while I'm at it.

crypto/console/console.c
smittals2: Sorry about these, I think I was copy pasting and forgot to change what's being cleared :(

crypto/fipsmodule/evp/evp.c
```diff
@@ -156,6 +158,76 @@ int EVP_PKEY_cmp(const EVP_PKEY *a, const EVP_PKEY *b) {
...
+  if (!buf || min_length <= 0 || min_length >= length) {
+    return -1;
+  }
+
+  if (prompt == NULL) {
+    prompt = EVP_get_pw_prompt();
+  }
+
+  // Proactively zeroize |buf| and verify_buf
+  OPENSSL_cleanse(buf, sizeof(buf));
```
skmcgrail, line 183:
```suggestion
  OPENSSL_cleanse(buf, length);
```
This should be `length` I think? Otherwise you are just getting the size of the pointer no?

crypto/console/console.c
```diff
@@ -0,0 +1,332 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include <assert.h>
+#include <ctype.h>
+#include <stdio.h>
+#include <string.h>
+#include <signal.h>
+#include <string.h>
```
skmcgrail, line 9:
```suggestion
```
Already imported on line 7

crypto/fipsmodule/evp/evp.c
```diff
@@ -156,6 +158,76 @@ int EVP_PKEY_cmp(const EVP_PKEY *a, const EVP_PKEY *b) {
...
+
+  // Write initial password prompt
+  if (!openssl_console_write(prompt)) {
+    goto err;
+  }
+
+  // Read password with echo disabled, returns 1 on success, 0 on error, -2 on interrupt
+  ret = openssl_console_read(buf, min_length, length, 0);
+  if (ret != 0) {
+    OPENSSL_cleanse(buf, sizeof(buf));
```
skmcgrail, line 201:
```suggestion
    OPENSSL_cleanse(buf, length);
```

Same question if this should be `length`

crypto/fipsmodule/evp/evp.c
```diff
@@ -156,6 +158,76 @@ int EVP_PKEY_cmp(const EVP_PKEY *a, const EVP_PKEY *b) {
...
+    openssl_console_write("Verifying - ");
+    openssl_console_write(prompt);
+
+    ret = openssl_console_read(verify_buf, min_length, sizeof(verify_buf), 0);
+
+    if (ret == 0) {
+      if (strncmp(buf, verify_buf, length >= 1024 ? 1024 : length) != 0) {
+        openssl_console_write("Verify failure\n");
+        ret = -1;
+      }
```
skmcgrail, line 216: This seems like a weird behavior to me? We only verify the first 1024 characters if the input length is longer then 1024? Should we just set a hard limit that `length` can't exceed 1024?

skmcgrail: Also I guess is there any constant time concerns around comparing this value in memory?

smittals2: This will be local execution, requiring human interaction. I don't think timing attacks are likely in this scenario (unlike a server handling a high load of auth requests that could be timed and exploited). Let me know if you foresee a scenario, I can look into solutions.

crypto/fipsmodule/evp/evp.c
```diff
@@ -156,6 +158,81 @@ int EVP_PKEY_cmp(const EVP_PKEY *a, const EVP_PKEY *b) {
...
+  }
+
+  // Write initial password prompt
+  if (!openssl_console_write(prompt)) {
+    goto err;
+  }
+
+  // Read password with echo disabled, returns 1 on success, 0 on error, -2 on interrupt
+  ret = openssl_console_read(buf, min_length, length, 0);
+  if (ret != 0) {
```
justsmth: This comment is wrong. `openssl_console_read` returns 0 on success.

justsmth (approved):

PR #2429, opened 2025-05-19
Resolve SSL_PRIVATE_METHOD and certificate slots functionality

skmcgrail:
### Description of changes: 
There was an issue with trying to use `SSL_PRIVATE_KEY_METHOD` functionality that was working for the most part, but had some edge scenarios that had stopped working. For example setting the `privkey_method` argument when calling `SSL_CTX_set_chain_and_key` or `SSL_set_chain_and_key`. Originally I thought this feature had been in a complete broken state, but upon further investigation realized that the bssl runner framework was using this functionality for testing async and hints functionality. The caveat was the framework was using alternative functions to directly set the certificate and private key method without using this entrypoint. In theory this entrypoint should work as it is for the most part coded correctly, but had a bug in that is tried to use the private key to determine the slot index when it could possibly be null. This corrects that behavior to correctly use the public key, to allow the chain to be set correctly, but still delegate the private key operations to the registered `SSL_PRIVATE_KEY_METHOD` provided.

### DIFF
--- a/ssl/ssl_cert.cc
+++ b/ssl/ssl_cert.cc
@@ -262,7 +262,7 @@ static enum leaf_cert_and_privkey_result_t do_leaf_cert_and_privkey_checks(
 // returns |leaf_cert_and_privkey_mismatch|. Otherwise it returns
 // |leaf_cert_and_privkey_ok|.
 static enum leaf_cert_and_privkey_result_t check_leaf_cert_and_privkey(
-    CRYPTO_BUFFER *leaf_buffer, EVP_PKEY *privkey) {
+    CRYPTO_BUFFER *leaf_buffer, EVP_PKEY *privkey, int *out_pubslot_idx) {
   CBS cert_cbs;
   CRYPTO_BUFFER_init_CBS(leaf_buffer, &cert_cbs);
   UniquePtr<EVP_PKEY> pubkey = ssl_cert_parse_pubkey(&cert_cbs);
@@ -270,7 +270,16 @@ static enum leaf_cert_and_privkey_result_t check_leaf_cert_and_privkey(
     OPENSSL_PUT_ERROR(SSL, SSL_R_DECODE_ERROR);
     return leaf_cert_and_privkey_error;
   }
-   return do_leaf_cert_and_privkey_checks(&cert_cbs, pubkey.get(), privkey);
+  enum leaf_cert_and_privkey_result_t result = do_leaf_cert_and_privkey_checks(
+      &cert_cbs, pubkey.get(), privkey);
+  if(out_pubslot_idx != nullptr) {
+    if(result == leaf_cert_and_privkey_ok) {
+      *out_pubslot_idx = ssl_get_certificate_slot_index(pubkey.get());
+    } else {
+      *out_pubslot_idx = -1;
+    }
+  }
+  return result;
 }
 
 static int cert_set_chain_and_key(
@@ -291,7 +300,9 @@ static int cert_set_chain_and_key(
     return 0;
   }
 
-  switch (check_leaf_cert_and_privkey(leaf_buf, privkey)) {
+  int slot_idx = -1;
+
+  switch (check_leaf_cert_and_privkey(leaf_buf, privkey, &slot_idx)) {
     case leaf_cert_and_privkey_error:
       return 0;
     case leaf_cert_and_privkey_mismatch:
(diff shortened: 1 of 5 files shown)


ssl/ssl_test.cc
```diff
@@ -13637,5 +13637,164 @@ TEST(SSLTest, MixContextAndConnection) {
...
+  int cert_len = i2d_X509(x509, &buf);
+  if (cert_len <= 0) {
+    return 0;
+  }
+
+  UniquePtr<CRYPTO_BUFFER> buffer(CRYPTO_BUFFER_new(buf, cert_len, NULL));
+  OPENSSL_free(buf);
+
+  return buffer;
+}
```
samuel40791765, line 13705: Should we just export the one in [`ssl/ssl_x509.cc`](https://github.com/aws/aws-lc/blob/4e97131369571c269e2e6690fd4dc99edfc1d427/ssl/ssl_x509.cc#L173) to use here?

ssl/ssl_test.cc
```diff
@@ -13637,5 +13637,164 @@ TEST(SSLTest, MixContextAndConnection) {
...
+    };
+
+    // Load a certificate chain containg the leaf but set private key method
+    ASSERT_TRUE(SSL_CTX_set_chain_and_key(server_ctx.get(), &ed_chain[0],
+                                          ed_chain.size(), ed_key.get(),
+                                          nullptr));
+
+    // Index must be ECC key now, but key_method must be set.
+    ASSERT_EQ(server_ctx->cert->cert_private_key_idx, SSL_PKEY_ED25519);
+    ASSERT_EQ(server_ctx->cert->key_method, nullptr);
```
samuel40791765, line 13774: I could be missing something, but I don't believe a private key method is being set or checked here.

ssl/ssl_test.cc
```diff
@@ -13637,5 +13637,164 @@ TEST(SSLTest, MixContextAndConnection) {
...
+        client_ctx.get(), sigalgs.data(), sigalgs.size()));
+
+    bssl::UniquePtr<SSL> client, server;
+    ASSERT_TRUE(ConnectClientAndServer(&client, &server, client_ctx.get(),
+                                       server_ctx.get(), ClientConfig(),
+                                       false));
+
+    ASSERT_TRUE(CompleteHandshakes(client.get(), server.get()));
+
+    ASSERT_EQ(test_ecc_privkey_calls, (size_t)1);
```
samuel40791765, line 13790: I'm slightly unsure why this is being incremented here, especially if we're not setting `test_ecc_private_key_method` in this test..

Should we add this assertion in the other tests as well?

ssl/ssl_test.cc
skmcgrail: Bad comment from copy-paste that I forgot to correct

skmcgrail: 😮‍💨 It's one from before in the earlier half of the test in the first test with the ECC key type, but I forgot to add the assert up in the first half to assert it was one then. It shouldn't have changed since then, which is what this is trying to reflect. Will fix it up.

ssl/ssl_test.cc
skmcgrail: 😮‍💨 It's 1 from the previous server invocation from the first block (since it's a global variable). Except the first section I forgot to assert it so it's a bit unclear. This was meant to verify it hadn't changed since then.

ssl/ssl_test.cc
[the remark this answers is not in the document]
skmcgrail: Not just you, file might just be getting that large 🫠

samuel40791765 (approved):

ssl/ssl_cert.cc
```diff
@@ -300,27 +311,35 @@ static int cert_set_chain_and_key(
...
+  // If privatekey is currently set then reset it.
+  // We either set a new |privatekey| or |privkey_method|
+  // later below.
   if (cert_pkey->privatekey) {
     cert_pkey->privatekey.reset();
   }
-  cert_pkey->privatekey = UpRef(privkey);
-  cert->key_method = privkey_method;
+
+  if(privkey != nullptr) {
```
smittals2, line 331: Just confused about the logic here. If privkey is null, then we should set privkey_method. But if it's not null, then reset it and don't set the privkey_method? 

It looks like before, regardless of whether cert_pkey-> privatekey was null, it would be set to privkey. And privkey_method would also be assigned. Why this distinction now?

ssl/ssl_privkey.cc
```diff
@@ -104,6 +104,7 @@ static bool ssl_set_pkey(CERT *cert, EVP_PKEY *pkey) {
 
   // Update certificate slot index once all checks have passed.
   cert->cert_private_keys[idx].privatekey = UpRef(pkey);
+  cert->key_method = nullptr; // key_method should be cleared since we've set a private key
```
smittals2, line 107: I guess same question here as above, what is the reason behind key_method being set to Null if a privkey has been set?

skmcgrail: You can't use both features at the same time, and if a slot private key is set, then that will be used if not NULL, otherwise it will use `key_method`. They are mutually exclusive features so it doesn't really make sense to keep a state where both are present. IMO

smittals2: Ahhh okay got it, so they do get used exclusively but for whatever reason our code didn't set them exclusively previously? 

Thanks!

smittals2 (approved):

ssl/ssl_cert.cc
skmcgrail: I think this got answered for you in the other comment, but the two are mutually exclusive and only one of them can be passed to this function. So this just ensures that state is kept and reflected correctly to avoid confusion or future bugs around the state of the CERT type instance.

PR #2430, opened 2025-05-20
Simplify Compiler CI jobs

justsmth:
### Description of changes: 
* Simplify CI for testing various compilers by using prepared Docker images for each.
* Fix build failures affecting GCC 15/16 and Clang 21.

### Call-outs:
* The FIPS build appears to be failing with Clang 20/21.
  * Clang 20/21 warn when both `-S` and `-c` are used together as options to the compiler.  This is potentially related to the FIPS built failure. (?)
* GCC 16 and Clang 21 are not yet released, but thought it would be useful to proactively build against them in our CI.

### DIFF
--- a/.github/workflows/actions-ci.yml
+++ b/.github/workflows/actions-ci.yml
@@ -274,6 +274,10 @@ jobs:
             compiler: 'gcc15'
           - fips: 1
             compiler: 'gcc16'
+          - fips: 1
+            compiler: 'clang20'
+          - fips: 1
+            compiler: 'clang21'
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/mattkretz/cplusplus-ci/${{ matrix.compiler }}
@@ -313,11 +317,19 @@ jobs:
       matrix:
         fips: [0, 1]
         compiler:
+          - 'gcc13'
           - 'gcc15'
+          - 'clang19'
           - 'clang20'
         exclude:
           - fips: 1
             compiler: 'gcc15'
+          - fips: 1
+            compiler: 'clang20'
+          - fips: 0
+            compiler: 'gcc13'
+          - fips: 0
+            compiler: 'clang19'
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/mattkretz/cplusplus-ci/${{ matrix.compiler }}
--- a/.github/workflows/actions-ci.yml
+++ b/.github/workflows/actions-ci.yml
@@ -251,81 +251,102 @@ jobs:
       matrix:
         fips: [0, 1]
         compiler:
-          - ['gcc9',    1, 'gcc',   'g++']
-          - ['gcc10',   1, 'gcc',   'g++']
-          - ['gcc11',   1, 'gcc',   'g++']
-          - ['gcc12',   1, 'gcc',   'g++']
-          - ['gcc13',   1, 'gcc',   'g++']
-          - ['gcc14',   0, 'gcc',   'g++']
-          - ['gcc15',   0, 'gcc',   'g++']
-          - ['gcc16',   0, 'gcc',   'g++']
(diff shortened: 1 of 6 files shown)


skmcgrail (approved):

.github/workflows/actions-ci.yml
```diff
@@ -433,64 +243,109 @@ jobs:
       - name: Run tests
         run: cmake --build ./build --target run_tests
 
-  clang-ubuntu-2404-sanity:
-    if: github.repository_owner == 'aws'
-    needs: [sanity-test-run]
+  compiler-tests:
```
andrewhop, line 246: Can you add a nice name like:
`name: ${{ matrix.compiler}} FIPS=${{ matrix.fips }}` so it's easier to read than `compiler-tests (0, clang16)`

.github/workflows/actions-ci.yml
```diff
@@ -433,64 +243,109 @@ jobs:
...
+          echo "CC=gcc" >> $GITHUB_ENV
+          echo "CXX=g++" >> $GITHUB_ENV
+      - name: Setup Build
+        if: matrix.fips == 0
+        run:
+          cmake -G Ninja -B ./build -DCMAKE_BUILD_TYPE=Release
+      - name: Setup FIPS Build
+        if: matrix.fips == 1
+        run:
+          cmake -G Ninja -B ./build -DCMAKE_BUILD_TYPE=Release -DFIPS=1
```
andrewhop: Can you use the matrix values in commands or are they special? `cmake -G Ninja -B ./build -DCMAKE_BUILD_TYPE=Release -DFIPS=matrix.fips`

crypto/fipsmodule/CMakeLists.txt
```diff
@@ -331,6 +331,11 @@ if(FIPS_DELOCATE)
   set_target_properties(bcm_c_generated_asm PROPERTIES COMPILE_OPTIONS "-S")
   set_target_properties(bcm_c_generated_asm PROPERTIES POSITION_INDEPENDENT_CODE ON)
 
+  # Clang 20+ warns when both "-S" and "-c" are used as options to the compiler.
```
andrewhop, line 334: Instead of ignoring unused arguments can we fix whatever is incorrectly passing -S and -c?

.github/workflows/actions-ci.yml
```diff
@@ -433,64 +243,109 @@ jobs:
...
+          echo "CXX=clang++" >> $GITHUB_ENV
+      - name: Setup GCC
+        if: startsWith(matrix.compiler, 'gcc')
+        run: |
+          echo "CC=gcc" >> $GITHUB_ENV
+          echo "CXX=g++" >> $GITHUB_ENV
+      - name: Setup Build
+        if: matrix.fips == 0
+        run:
+          cmake -G Ninja -B ./build -DCMAKE_BUILD_TYPE=Release CMAKE_C_FLAGS=-pedantic CMAKE_CXX_FLAGS=-pedantic
```
andrewhop, line 341: Can this be turned on by default in our CMakeLists.txt? Why is this a special CI dimension?

crypto/fipsmodule/CMakeLists.txt
justsmth: I tried this approach, but I could not find a way to prevent the "-c" from being passed. There's a deeper problem with the FIPS build using Clang 20/21 that this is likely related to. We have an internal tracking issue for it.

.github/workflows/actions-ci.yml
justsmth: pedantic builds are very fragile. We would not want it as a default in our build. However, it should be part of our CI so we have awareness.

andrewhop (approved):

skmcgrail (approved):

PR #2433, opened 2025-05-22
Rework memory BIOs and implement BIO_seek (2nd try)

nhatnghiho:
### Issues:
Resolves #CryptoAlg-3111

### Description of changes: 
Same as #2380. 

Fixed the state bug that was causing Ruby failures. The failures came from `BIO_reset`, which tries to "rewind" the R/O buffer to its original state. The old implementation only works when the buffer pointer and read pointer are aligned. But in the new memory BIO, this is not always the case. The buffer pointer (`b->data`) does not need to align with the read pointer (since we can have a positive `bbm->read_off`). Our tests didn't cover this because the buffer pointer was synced with read pointer after every test case. 

### Testing: 
Added tests for `BIO_reset` and `BIO_seek` functions that attempt to rewind the buffer internally. The tests cover both cases -- when buffer pointer and read pointer are synced and when they are not.

### DIFF
--- a/crypto/bio/bio_mem.c
+++ b/crypto/bio/bio_mem.c
@@ -358,7 +358,7 @@ int BIO_mem_contents(const BIO *bio, const uint8_t **out_contents,
   mem_buf_sync((BIO *)bio);
 
   if (out_contents != NULL) {
-    *out_contents = (b->data != NULL) ? (uint8_t *)&b->data[bbm->read_off] : NULL;
+    *out_contents = (uint8_t *)b->data;
   }
   if(out_len) {
     *out_len = b->length;
--- a/crypto/bio/bio_mem.c
+++ b/crypto/bio/bio_mem.c
@@ -256,7 +256,7 @@ static long mem_ctrl(BIO *bio, int cmd, long num, void *ptr) {
       if (b->data != NULL) {
         // For read only case reset to the start again
         if (bio->flags & BIO_FLAGS_MEM_RDONLY) {
-          b->data -= b->max - b->length;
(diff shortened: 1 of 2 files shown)


crypto/bio/bio_mem.c
```diff
@@ -65,11 +65,18 @@
...
+} BIO_BUF_MEM;
+
 
 BIO *BIO_new_mem_buf(const void *buf, ossl_ssize_t len) {
   BIO *ret;
   BUF_MEM *b;
-  const size_t size = len < 0 ? strlen((char *)buf) : (size_t)len;
+  BIO_BUF_MEM *bbm;
+
+  const size_t size = (len < 0 || (size_t)len > SIZE_MAX) ? strlen((char *)buf) : (size_t)len;
```
skmcgrail, line 79: I'm not familiar with these APIs but the use of `strlen` here seems wrong? Is buffer really expected to be a null terminated value?

nhatnghiho: Apparently, this is ancient implementation from both OpenSSL and BoringSSL. Per [OpenSSL's documentation](https://docs.openssl.org/3.0/man3/BIO_s_mem/#description),

> BIO_new_mem_buf() creates a memory BIO using len bytes of data at buf, if len is -1 then the buf is assumed to be null terminated and its length is determined by strlen.

But if we don't have that assumption, I can't really think of a way to determine if `buf` is null terminated, especially since we don't have its size.

skmcgrail (approved):

justsmth (approved):

PR #2438, opened 2025-05-23
Verify leaf certificate public key rather then leaving it to the caller

skmcgrail:
OpenSSL's behavior when performing X509 certificate validation is to build the up certificate chain, verify the extensions of the constructed chain, etc, finally proceeding with handling the subjectPublicKeyInfo algorithm and parameters inheritance as appropriate through the chain when applicable per [rfc5280](https://www.rfc-editor.org/rfc/inline-errata/rfc5280.html). This process will cause OpenSSL to reject certificates when encountering public keys that it doesn't understand, for example unknown / unsupported EC curve names, invalid curve points etc. This includes intermediate and root certificates with such public keys (which would need to be usable to validate the issuer signatures on subjects in the chain etc), but also includes the leaf certificate as it's public key is parsed to determine if it inherits parameters etc.

For AWS-LC this parameter inheritance is not supported (as it was removed by BoringSSL from which we forked), particularly because for key types like EC we only support specific named curves and no custom parameters. So this step of key inheritance does not occur. This means that the certificate subject public keys are not used until validating the issuer signatures through the chain. Root and intermediate certificates with public keys that are not parseable / supported are then rejected at this step, except for the leaf certificate. There was a comment left by BoringSSL in the tests to indicate that the subject public key (for which the EVP_PKEY may be NULL if it was not parseable) is left to be validated by the caller invoking `X509_verify_cert`. This means that such a leaf certificate passes validated (correctly validated through the constructed chain, reasonably well formed etc), but still might not actually be useable. But such certificates would have never returned successfully by `X509_verify_cert` if using OpenSSL. This PR realigns AWS-LC to behave similarly to OpenSSL by checking the final leaf certificate for the user, and verifying that it contains a supported public key type.

### DIFF
--- a/crypto/x509/x509_vfy.c
+++ b/crypto/x509/x509_vfy.c
@@ -1432,10 +1432,10 @@ static int internal_verify(X509_STORE_CTX *ctx) {
   // has a public key we actually support. Otherwise
   // we shouldn't consider the certificate okay.
   //
-  // This covers situtations like invalid curves or points.
+  // This covers situations like invalid curves or points.
   EVP_PKEY *pkey = X509_get0_pubkey(xs);
   if(pkey == NULL) {
-    ctx->error = X509_V_ERR_UNABLE_TO_DECODE_LEAF_PUBLIC_KEY;
+    ctx->error = X509_R_UNABLE_TO_GET_CERTS_PUBLIC_KEY;
     ctx->current_cert = xi;
     ok = 0;
     goto end;
--- a/crypto/x509/x509_vfy.c
+++ b/crypto/x509/x509_vfy.c
@@ -1427,6 +1427,20 @@ static int internal_verify(X509_STORE_CTX *ctx) {
       xs = sk_X509_value(ctx->chain, n);
     }
   }
+
+  // Check that the final certificate subject (leaf)
+  // has a public key we actually support. Otherwise
+  // we shouldn't consider the certificate okay.
+  //
+  // This covers situtations like invalid curves or points.
+  EVP_PKEY *pkey = X509_get0_pubkey(xs);
+  if(pkey == NULL) {
+    ctx->error = X509_V_ERR_UNABLE_TO_DECODE_LEAF_PUBLIC_KEY;
+    ctx->current_cert = xi;
+    ok = 0;
+    goto end;
+  }
+
   ok = 1;
 end:
   return ok;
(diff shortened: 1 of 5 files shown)


ssl/ssl_x509.cc
```diff
@@ -1586,6 +1586,9 @@ int SSL_alert_from_verify_result(long result) {
     case X509_V_ERR_INVALID_PURPOSE:
       return SSL_AD_UNSUPPORTED_CERTIFICATE;
 
+    case X509_V_ERR_UNABLE_TO_DECODE_LEAF_PUBLIC_KEY:
+      return SSL_AD_UNSUPPORTED_CERTIFICATE;
```
justsmth, line 1590: NP: These can be combined (similar to the others above):
```
    case X509_V_ERR_INVALID_PURPOSE:
    case X509_V_ERR_UNABLE_TO_DECODE_LEAF_PUBLIC_KEY:
      return SSL_AD_UNSUPPORTED_CERTIFICATE;
```

crypto/x509/x509_vfy.c
```diff
@@ -1427,6 +1427,20 @@ static int internal_verify(X509_STORE_CTX *ctx) {
       xs = sk_X509_value(ctx->chain, n);
     }
   }
+
+  // Check that the final certificate subject (leaf)
+  // has a public key we actually support. Otherwise
+  // we shouldn't consider the certificate okay.
+  //
+  // This covers situtations like invalid curves or points.
```
justsmth, line 1435: NP: "situations"

justsmth (approved):

include/openssl/x509.h
```diff
@@ -2736,6 +2736,7 @@ OPENSSL_EXPORT void X509_STORE_CTX_set_cert(X509_STORE_CTX *c, X509 *x);
 #define X509_V_ERR_EE_KEY_TOO_SMALL 68
 #define X509_V_ERR_CA_KEY_TOO_SMALL 69
 #define X509_V_ERR_CA_MD_TOO_WEAK 70
+#define X509_V_ERR_UNABLE_TO_DECODE_LEAF_PUBLIC_KEY 71
```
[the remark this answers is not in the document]
skmcgrail: This specific error type doesn't occur in OpenSSL, as errors around bad key for an unknown curve / invalid point would be returned as `X509_R_UNABLE_TO_GET_CERTS_PUBLIC_KEY` in this instance, or potentially `X509_R_UNABLE_TO_FIND_PARAMETERS_IN_CHAIN`.

`X509_R_UNABLE_TO_GET_CERTS_PUBLIC_KEY` seemed fairly unspecific to me (and in our case this would always be the leaf), but maybe we should stick with it for compatibility?

justsmth (approved):

samuel40791765 (approved):

PR #2454, opened 2025-06-02
Order tool output by options provided - x509

justsmth:
### Context
* The OpenSSL CLI tool orders its output based on the order of the options passed to it.
* Our tool was using a static order that was unaffected by the order of the options.

### Description of changes: 
* This change adds an alternative suite of functions to parse options in a way that preserves their order.
* The new argument parser was integrated into the "X509" subcommand logic for the tool.

### Call-outs:
* The new parser still needs to be integrated with the other subcommands as well.

### Testing:
Added new tests that vary the order of the parameters provided.

### DIFF
--- a/tool-openssl/ordered_args.cc
+++ b/tool-openssl/ordered_args.cc
@@ -1,0 +1,134 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include <algorithm>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <errno.h>
+#include <limits.h>
+#include <stdio.h>
+#include <stdlib.h>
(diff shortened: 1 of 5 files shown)


tool-openssl/ordered_args.cc
```diff
@@ -0,0 +1,134 @@
...
+    return true;
+  }
+
+  const std::string &value = it->second;
+  *out = value;
+
+  return true;
+}
+
+bool GetBoolArgument(bool *out, const std::string &arg_name,
```
skmcgrail, line 120: Perhaps just return void and forgo the `amazonq` comment telling it to ignore the always true nature of the function. Also do we want to be committing these tool specific comments?

justsmth: I'll remove the comment. I think I kept it returning a bool just to keep it consistent with the other functions.

PR #2464, opened 2025-06-04
Add timeouts to PQ TLS Integ Tests

alexw91:
### Issues:
V1795657921

### Description of changes: 
Some TLS CLI tools listen on stdin for data to send over a TLS connection, other TLS CLI tools immediately close the TLS connection. To ensure that PQ TLS integration tests are not broken if TLS command line tools are updated to await input from stdin instead of immediately closing the TLS connection, run the CLI tools in the background (so that stdin is closed and they should not wait forever for input), and add a timeout (so that even if they do wait, they will be closed, and test progress is ensured).

### Call-outs:
None

### Testing:
Ran locally and confirmed passing in CI. 

### DIFF
--- a/tests/ci/integration/run_pq_tls_integration.sh
+++ b/tests/ci/integration/run_pq_tls_integration.sh
@@ -58,7 +58,7 @@ for GROUP in X25519MLKEM768 SecP256r1MLKEM768; do
     &> "$AWS_LC_BUILD_FOLDER"/s_server_out &
   sleep 5 # to allow for the server to startup in the background thread
   S_PID=$!
-  ${CLIENT_TIMEOUT_CMD} ${S2NC_CMD} -c default_pq -i localhost 45000 &> "$S2N_TLS_BUILD_FOLDER"/s2nc_out
+  ${CLIENT_TIMEOUT_CMD} ${S2NC_CMD} -c default_pq -i localhost 45000 &> "$S2N_TLS_BUILD_FOLDER"/s2nc_out &
   wait $S_PID || true
   cat "$AWS_LC_BUILD_FOLDER"/s_server_out
   cat "$S2N_TLS_BUILD_FOLDER"/s2nc_out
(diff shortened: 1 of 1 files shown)


tests/ci/integration/run_pq_tls_integration.sh
```diff
@@ -51,12 +54,11 @@ echo "s2n build succeeded. Found: ${S2NC_CMD} ${S2ND_CMD}"
...
 for GROUP in X25519MLKEM768 SecP256r1MLKEM768; do
   echo "TLS Handshake: aws-lc server (bssl) with s2n-tls client (s2nc) for group $GROUP"
-  ${AWS_LC_CMD} s_server -curves $GROUP -accept 45000 -debug \
+  ${SERVER_TIMEOUT_CMD} ${AWS_LC_CMD} s_server -curves $GROUP -accept 45000 -debug \
     &> "$AWS_LC_BUILD_FOLDER"/s_server_out &
   sleep 5 # to allow for the server to startup in the background thread
   S_PID=$!
-  # Relying on s2nc behavior that it exits after the first handshake
-  ${S2NC_CMD} -c default_pq -i localhost 45000 &> "$S2N_TLS_BUILD_FOLDER"/s2nc_out
+  ${CLIENT_TIMEOUT_CMD} ${S2NC_CMD} -c default_pq -i localhost 45000 &> "$S2N_TLS_BUILD_FOLDER"/s2nc_out
```
chockalingamc: Perhaps we should run this client in the background as well, in case this hangs and the timeout command ends up  terminating this, which will result in a non-zero exit code.

alexw91: Added `--preserve-status` flag to timeout command

chockalingamc: Even with `--preserve-status`, when timeout command itself is in the foreground the exit code would be non-zero (130/143 instead of 124) and would terminate the script. If we add `&` to the end of this, that would make timeout itself run in background and make sure the exitcode this script sees is always 0 irrespective of what the background processes returns.

chockalingamc (approved):
Looks good

WillChilds-Klein (approved):

PR #2475, opened 2025-06-06
Prepare Docker images for upcoming CI changes

nhatnghiho:
### Issues:
Addresses CryptoAlg-2907, CryptoAlg-3113, and P257046914

### Description of changes:
This PR bundles a few Docker image changes together:
- Added a few packages to our Linux images that's needed for Unbound and Bind9 CI.
- Updated all Windows images to Windows 2022 so that we can upgrade our CodeBuild compute type and migrate [MSVC-SDE-64-bit](https://github.com/aws/aws-lc/blob/main/.github/workflows/actions-ci.yml#L139-L167) to CodeBuild. This is because MSVC-SDE-64-bit needs a lot of compute power. Also added an MSVC 2022 image to be used for this test migration.

### Callouts:
- Updating all Windows images to Windows 2022 means that our existing Windows CI tests in CodeBuild will now be run on Windows 2022, which should be okay because the MSVC version is still the same.
- Running MSVC-SDE-64-bit on Windows 2022 will require disabling 5 CPUs that don't support AVX instructions, which Windows 2022 natively use. But Windows 2022 is our only easy path to migration because MSVC-SDE-64-bit is resource-intensive and requires at least a compute type of [BUILD_GENERAL1_2XLARGE](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-compute-types.html#environment.types) that is only available for Windows Server 2022 on CodeBuild.


tests/ci/cdk/pipeline/scripts/util.sh
```diff
@@ -145,7 +151,7 @@ function win_docker_img_build_status_check() {
     elif [[ ${command_run_status} == "InProgress" ]]; then
       echo "${i}: Wait 5 min for build job finish."
       sleep 300
-    elif echo "${command_run_status}" | grep -q "ExpiredTokenException"; then
+    elif echo "${command_run_status}" | grep -q "RequestExpired"; then
```
samuel40791765: A comment on what `RequestExpired` means here might help. Do we still need to handle `ExpiredTokenException`s?

tests/ci/docker_images/windows/build_images.ps1
```diff
@@ -1,6 +1,9 @@
...
 
-docker build -t aws-lc/windows_base:2019 .\windows_base
-docker build -t vs2015 .\vs2015
-docker build -t vs2017 .\vs2017
+. .\common.ps1
+
+Invoke-Command { docker build -t aws-lc/windows-2022:base .\windows-2022_base }
+Invoke-Command { docker build -t windows-2022:vs2022-sde .\windows-2022_vs2022-sde }
+Invoke-Command { docker build -t windows-2022:vs2017-sde .\windows-2022_vs2017-sde }
+Invoke-Command { docker build -t windows-2022:vs2015 .\windows-2022_vs2015 }
```
samuel40791765: Nit: Would be great if ordering could align with `tests/ci/docker_images/windows/push_images.ps1`.

Also, should we add the sde tag to these docker images? No strong opinions on my end.

Context: I added SDE tools since we could be looking to some docker image consolidation (`CryptoAlg-1841`). We're not actually running Windows SDE tests through this image though, we run the Windows SDE tests through Github Actions. I included the setup here since I had already investigated it (setting up Windows is a pain...) and it could be useful for future testing purposes."

nhatnghiho: We're actually trying to migrate the Windows SDE tests to CodeBuild per CryptoAlg-3113. I mentioned this more clearly in the description/callouts

nhatnghiho: > should we add the sde tag to these docker images? No strong opinions on my end.

The full tag for sde images are either `windows-2022_vs2019-sde` or `windows-2022_vs2022-sde`. Do you mean adding more tags or renaming them in a way?

tests/ci/cdk/pipeline/scripts/util.sh
nhatnghiho: When our temporary credentials expires, some cli commands will fire a `RequestExpired` or `ExpiredTokenException`. So we only need to handle 1 depending on the cli command. But this one should be `ExpiredTokenException`, my bad. Comments added

tests/ci/docker_images/windows/windows-2022_vs2022-sde/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+# Add MSBuild and Windows SDK tools to PATH
+RUN setx /M PATH "%PATH%;C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\MSBuild\17.0\Bin;C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\x64"
+
+# Make sure the Windows SDK is properly configured
+RUN if not exist "C:\Program Files (x86)\Windows Kits\10\bin\x64\rc.exe" ( `
+        if exist "C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\x64\rc.exe" ( `
+            mkdir "C:\Program Files (x86)\Windows Kits\10\bin\x64" 2>nul & `
+            copy "C:\Program Files (x86)\Windows Kits\10\bin\10.0.19041.0\x64\rc.exe" "C:\Program Files (x86)\Windows Kits\10\bin\x64\" `
+        ) `
+    )
```
skmcgrail: I get this feeling this path component `10.0.19041.0` is likely to change overtime with image rebuilds? Is there not an existing batch script or environment setup we can call to configure this correctly for us? Or discover the paths? Also seems weird that we need to shift these items around.

tests/ci/docker_images/windows/build_images.ps1
samuel40791765: The `sde` tag just seemed sort of redundant, since we only have one `windows-2022_vs2019` or `windows-2022_vs2022` image. The `sde` images aren't really special now since these images are the only types of `vs2019` & `vs2022` images we've got. 

I was leaning towards dropping the tag, but it's fine if we want to keep them.

tests/ci/docker_images/windows/windows-2022_vs2022-sde/Dockerfile
nhatnghiho: You're right. We don't need this part. I removed it.

tests/ci/cdk/pipeline/scripts/util.sh
```diff
@@ -99,8 +100,13 @@ function start_windows_img_build() {
...
     instance_id=$(aws ec2 describe-instances \
-      --filters "Name=tag:${WIN_EC2_TAG_KEY},Values=${WIN_EC2_TAG_VALUE}" | jq -r '.Reservations[0].Instances[0].InstanceId')
-    if [[ "${instance_id}" == "null" ]]; then
+      --filters "Name=tag:${WIN_EC2_TAG_KEY},Values=${WIN_EC2_TAG_VALUE}" \
+      --query 'Reservations[0].Instances[0].InstanceId' \
+      --output text 2>&1)
+    # if the build runs for too long, AWS sessions may expire and need to be refreshed
+    if echo "${instance_id}" | grep -q "RequestExpired"; then
+      refresh_session
+    elif [[ "${instance_id}" == "null" ]]; then
```
andrewhop, line 109: Did you consider extending the duration in `assume_role` instead of using the short default time period?

nhatnghiho: I tried but it's not possible to increase beyond the default time period because we're doing role chaining. According to this [documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html):

> Role chaining limits your AWS CLI or AWS API role session to a maximum of one hour. [...] However, if you assume a role using role chaining and provide a DurationSeconds parameter value greater than one hour, the operation fails.

samuel40791765 (approved):

skmcgrail (approved):

PR #2483, opened 2025-06-11
Add production stage to CI pipeline

nhatnghiho:
### Issues:
Resolves #CryptoAlg-2985

### Description of changes: 
Enable production stage for the CI pipeline with a few upgrades:
- The transition from pre-prod to prod requires manual approval
- Push Docker images built in pre-production to our production ECR directly rather than rebuilding. Right now only available for Linux Docker images (only takes 10 minutes for Linux Docker images vs 1.5hr to rebuild from scratch)
- The pipeline's execution mode is QUEUED, which means executions will get stuck and piled on at the manual approval step, regardless of what type of changes. I added a skip condition to skip production stage (including the manual approval step) to avoid unnecessary executions

### Testing:
Test deploy the pipeline using instructions in README


skmcgrail (approved):

tests/ci/cdk/pipeline/pipeline_stack.py
```diff
@@ -188,6 +193,45 @@ def __init__(
             targets=[targets.CodePipeline(pipeline=base_pipeline)],
         )
 
+        # Pipeline is run everytime we push to main branch. Avoid unnecessary hold up if these updates are non-CI related
+        if not IS_DEV:
+            start_index = next(
+                (i for i, stage in enumerate(base_pipeline.stages) if stage.stage_name == "PromoteToProduction"), #TODO: Revert this
+                None
+            )
```
andrewhop: I don't follow what this is doing or what the impact is, also do you have an issue for this TODO so we don't forget?

tests/ci/cdk/pipeline/scripts/copy_images.sh
```diff
@@ -0,0 +1,104 @@
...
+  for repo in "${!IMAGE_TAGS[@]}"; do
+    tag="${IMAGE_TAGS[$repo]}"
+    echo "Processing ${repo}:${tag} for prod"
+
+    # Create date-based tag
+    img_push_date=$(date +%Y-%m-%d)
+    if [[ -n "${PLATFORM:-}" && ${PLATFORM} == "windows" ]]; then
+      img_push_date=$(date +%Y-%m-%d-%H)
+    fi
+    date_tag="${tag%_latest}_${img_push_date}"
```
andrewhop: This creates today's tag which will be different than the pre-prod image, is there anyway to connect a prod image to the pre-prod image it came from?

tests/ci/cdk/pipeline/pipeline_stack.py
nhatnghiho: The TODO is redundant, I'll remove that. 

But the idea is -- the pipeline is run everytime we push to main branch. With every stage, we would check to see if we need to redeploy. However, the manual approval step doesn't have such check. The pipeline will just sit and wait at the manual approval step until we approve it. This will lead to a pile of irrelevant pipeline executions stuck waiting at this step. That's why I put in the skip condition for this manual approval step so we can silently skip it (and since I'm at it already, when the manual approval step is skipped, I also configure the pipeline to skip all the subsequent production stages)

tests/ci/cdk/pipeline/scripts/copy_images.sh
nhatnghiho: Updated code to reflect this. Since our images have a lifecycle of 30 days though, which depends on the time the images are published to ECR, the date tag and the expiration time will not correspond to each other. But preserving the date tag would make it easier for debugging.

andrewhop (approved):

skmcgrail (approved):

skmcgrail (approved):

PR #2484, opened 2025-06-12
Implement HMAC over SHA3 truncated variants

WillChilds-Klein:
### Issues:
n/a

### Description of changes: 

This PR is currently a DRAFT for exercising CI. Substantial cleanup is required, please do not review yet.

### Call-outs:

n/a

### Testing:
- forthcoming CI tests

---


crypto/hmac_extra/hmac_test.cc
```diff
@@ -87,6 +87,14 @@ static const EVP_MD *GetDigest(const std::string &name) {
     return EVP_sha512_224();
   } else if (name == "SHA512/256") {
     return EVP_sha512_256();
+  } else if (name == "SHA3/224") {
```
picnixz: I'd recommend naming them "SHA3-224" instead. The / is usually for true truncated variants such as SHA2-512/224 (SHA512/t is the same as SHA512 but with different IVs and with the output truncated at t bits).

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -583,6 +615,10 @@ int HMAC_Init_from_precomputed_key(HMAC_CTX *ctx,
   }
 
   const HmacMethods *methods = ctx->methods;
+  if (methods != NULL && ctx->methods->init_from_state == NULL) {
```
picnixz: I assume that `hmac_ctx_set_md_methods` is guaranteein that methods is not NULL, so you can just remove `methods != NULL` (if this is not the case, the existing code is already wrong as `methods->chaining_length` may crash).

crypto/hmac_extra/hmac_test.cc
WillChilds-Klein: Ah, good catch. agreed.

crypto/fipsmodule/hmac/hmac.c
WillChilds-Klein: Yes, it looks like [that's ensured](https://github.com/aws/aws-lc/blob/587cf97c404a6c32f212e5c44ccb3503cb5ff430/crypto/fipsmodule/hmac/hmac.c#L335-L337), so this null check is redundant.

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
```diff
@@ -19,3 +19,31 @@ index a066982..3d7fbc3 100644
...
  #
+diff --git a/configure b/configure
+index 029bf52..3fada48 100755
+--- a/configure
++++ b/configure
+@@ -30873,7 +30873,6 @@ main (void)
+       OBJ_nid2sn(NID_md5);
+       OBJ_nid2sn(NID_sha1);
+       OBJ_nid2sn(NID_sha3_512);
+-      OBJ_nid2sn(NID_blake2b512);
```
samuel40791765, line 30: Nit: Not a blocker for this PR, but I guess we'll have to handle this more gracefully if we're going to upstream this patch?

crypto/hmac_extra/hmac_test.cc
```diff
@@ -253,6 +261,87 @@ TEST(HMACTest, TestVectors) {
...
+    OPENSSL_memset(mac.get(), 0, expected_mac_len); // Clear the prior correct answer
+
+    // Test that an HMAC_CTX may be reset with the same key.
+    ASSERT_TRUE(HMAC_Init_ex(ctx.get(), nullptr, 0, digest, nullptr));
+    ASSERT_TRUE(HMAC_Update(ctx.get(), input.data(), input.size()));
+    ASSERT_TRUE(HMAC_Final(ctx.get(), mac.get(), &mac_len));
+    EXPECT_EQ(Bytes(output), Bytes(mac.get(), mac_len));
+    OPENSSL_memset(mac.get(), 0, expected_mac_len); // Clear the prior correct answer
+
+    // Test that an HMAC_CTX may be reset with the same key and a null md
```
andrewhop, line 280: Why did you move these tests? Did you change anything else?

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -505,6 +531,12 @@ int HMAC_set_precomputed_key_export(HMAC_CTX *ctx) {
 }
 
 int HMAC_get_precomputed_key(HMAC_CTX *ctx, uint8_t *out, size_t *out_len) {
+  GUARD_PTR(ctx);
+  if (ctx->methods != NULL && ctx->methods->get_state == NULL) {
```
andrewhop: Should this also return 0 if methods is null? We unconditionally dereference it down L550.

crypto/fipsmodule/sha/sha3.c
```diff
@@ -270,11 +270,67 @@ int SHA3_Final(uint8_t *md, KECCAK1600_CTX *ctx) {
 
   Keccak1600_Squeeze(ctx->A, md, ctx->md_size, ctx->block_size, ctx->state);
   ctx->state = KECCAK1600_STATE_FINAL;
-  
+
   FIPS_service_indicator_update_state();
   return 1;
 }
 
+int SHA3_224_Init(KECCAK1600_CTX *ctx) {
```
andrewhop, line 278: We already have all of these in digest.c, can you just update those to not be static?

include/openssl/digest.h
```diff
@@ -181,7 +181,7 @@ OPENSSL_EXPORT int EVP_DigestUpdate(EVP_MD_CTX *ctx, const void *data,
 
 // EVP_MAX_MD_BLOCK_SIZE is the largest digest block size supported, in
 // bytes.
-#define EVP_MAX_MD_BLOCK_SIZE 128  // SHA-512 is the longest so far.
+#define EVP_MAX_MD_BLOCK_SIZE 144      // SHA3-224 has the largest block size so far
```
andrewhop, line 184: Does anything check/enforce this?

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -118,62 +126,75 @@ struct hmac_methods_st {
...
+    out->methods[idx-1].get_state =                                          \
+        AWS_LC_TRAMPOLINE_##HASH_NAME##_get_state;                           \
+  }
+
 DEFINE_LOCAL_DATA(struct hmac_method_array_st, AWSLC_hmac_in_place_methods) {
   OPENSSL_memset((void*) out->methods, 0, sizeof(out->methods));
   int idx = 0;
   // Since we search these linearly it helps (just a bit) to put the most common ones first.
   // This isn't based on hard metrics and will not make a significant different on performance.
   // FIXME: all hashes but SHA256 have been disabled to check first SHA256
```
andrewhop: Do you know what this FIXME is referencing?

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -110,6 +106,18 @@ struct hmac_methods_st {
...
+  OPENSSL_STATIC_ASSERT(HASH_CBLOCK % 8 == 0,                                 \
+                        HASH_NAME##_has_blocksize_not_divisible_by_eight_t)   \
+  OPENSSL_STATIC_ASSERT(HASH_CBLOCK <= EVP_MAX_MD_BLOCK_SIZE,                 \
+                        HASH_NAME##_has_overlarge_blocksize_t)                \
+  OPENSSL_STATIC_ASSERT(sizeof(HASH_CTX) <= sizeof(union md_ctx_union),       \
+                        HASH_NAME##_has_overlarge_context_t)
+
+// For merkle-daamgard constructions, we also define functions for  manipulating
+// hash state |h| to calculate or use precomputed keys. These are not applicable
+// to Keccak/SHA3.
```
andrewhop: Is that not allowed or just not implemented (yet)?

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -118,62 +126,75 @@ struct hmac_methods_st {
...
+    out->methods[idx].chaining_length = 0UL;                                 \
+    out->methods[idx].init_from_state = NULL;                                \
+    out->methods[idx].get_state = NULL;                                      \
     idx++;                                                                   \
     assert(idx <= HMAC_METHOD_MAX);                                          \
   }
 
+// Use |idx-1| because DEFINE_IN_PLACE_METHODS has already incremented it.
+#define DEFINE_IN_PLACE_METHODS_PRECOMPUTED(EVP_MD, HASH_NAME)  {            \
+    DEFINE_IN_PLACE_METHODS(EVP_MD, HASH_NAME);                              \
```
samuel40791765, line 172: Nit: Should we add `assert(idx -1 >=0)` here? Just to align with `DEFINE_IN_PLACE_METHODS` above.
```suggestion
    DEFINE_IN_PLACE_METHODS(EVP_MD, HASH_NAME);                              \
    assert(idx -1 >=0);
```

include/openssl/hmac.h
```diff
@@ -250,12 +250,17 @@ OPENSSL_EXPORT int HMAC_CTX_copy(HMAC_CTX *dest, const HMAC_CTX *src);
 // Private functions
 typedef struct hmac_methods_st HmacMethods;
 
-// We use a union to ensure that enough space is allocated and never actually bother with the named members.
+// We use a union to ensure that enough space is allocated and never actually
+// bother with the named members. We do not externalize SHA3 ctx definition,
```
samuel40791765, line 254: Should we mention that SHA3 ctx is referring to `KECCAK1600_CTX` here? It doesn't seem immediately obvious.

include/openssl/hmac.h
```diff
@@ -250,12 +250,17 @@ OPENSSL_EXPORT int HMAC_CTX_copy(HMAC_CTX *dest, const HMAC_CTX *src);
...
+// so hard-code ctx size below and use a compile-time assertion where that ctx
+// is defined to ensure it does not exceed size bounded by |md_ctx_union|. This
+// is OK because union members are never referenced, they're only used for sizing.
 union md_ctx_union {
   MD5_CTX md5;
   SHA_CTX sha1;
   SHA256_CTX sha256;
   SHA512_CTX sha512;
+  uint8_t sha3[400];
 };
```
samuel40791765, line 264: A bit of a bolder change, but could we move the `md_ctx_union` union to our internal headers? I don't see any existing OpenSSL reference to the union. That way we could abstract `400` to a `#define` variable instead of a hard-coded number.

If we're not comfortable doing so, we can leave this as is though.

crypto/fipsmodule/sha/internal.h
```diff
@@ -61,6 +63,11 @@ extern "C" {
 #define SHA3_BLOCKSIZE(bitlen) (KECCAK1600_WIDTH - bitlen * 2) / 8
 #define SHA3_PAD_CHAR 0x06
 
+#define SHA3_224_CBLOCK SHA3_BLOCKSIZE(SHA3_224_DIGEST_BITLENGTH)
+#define SHA3_256_CBLOCK SHA3_BLOCKSIZE(SHA3_256_DIGEST_BITLENGTH)
+#define SHA3_384_CBLOCK SHA3_BLOCKSIZE(SHA3_384_DIGEST_BITLENGTH)
+#define SHA3_512_CBLOCK SHA3_BLOCKSIZE(SHA3_512_DIGEST_BITLENGTH)
```
samuel40791765: Nit: Should we move this with each `SHA3_*_CAPACITY_BYTES`, `SHA3_*_DIGEST_BITLENGTH`, etc. for consistency?

include/openssl/digest.h
samuel40791765, line 184: Had to cross reference the math being done in `crypto/fipsmodule/sha/internal.h`. Seems a bit magical that SHA3-224 has the largest block size now.

WillChilds-Klein: Indirectly, for digests used with HMAC. It's checked [here](https://github.com/aws/aws-lc/blob/f44d9780a8e184daa227f2ae8cd354235a924ffb/crypto/fipsmodule/hmac/hmac.c#L123) at compile time.

crypto/fipsmodule/hmac/hmac.c
WillChilds-Klein: Yes, good call.

crypto/fipsmodule/hmac/hmac.c
WillChilds-Klein: The internal hash state is not readily exportable in SHA3/Keccak in the way that it is in merkle-damgard, so these functions are not applicable. I'll clarify the comment.

crypto/fipsmodule/hmac/hmac.c
WillChilds-Klein: I do not. It was added [with the pre-computed keys change](https://github.com/aws/aws-lc/commit/98ccf4a316401112943bed604562102ad52efac6). Do you know @fabrice102?

crypto/fipsmodule/sha/sha3.c
WillChilds-Klein: I don't think so. The signatures are different (`KECCAK1600_CTX` vs. `EVP_MD_CTX`), so `digests.c` has those static functions to wrap SHA1 and SHA2's ctx-specific public functions. For SHA3, we follow the same pattern (e.g. [`SHA512_256_Update`](https://github.com/aws/aws-lc/blob/e2a5299e2eda27cae589e1077e454e0a1de782bc/crypto/fipsmodule/sha/sha512.c#L297) also has a corresponding static function [in `digests.c`](https://github.com/aws/aws-lc/blob/e2a5299e2eda27cae589e1077e454e0a1de782bc/crypto/fipsmodule/digest/digests.c#L341)), only the ctx-specific functions here aren't exposed in public headers.

crypto/hmac_extra/hmac_test.cc
WillChilds-Klein: I moved them here to retain full coverage on all HMAC variants, trying to minimize duplication/redundancy in `TestVectorsPrecomputedKey`. I changed the comments, but the test code changes here are strictly code motion -- teasing out precomputed-specific cases and copying some portions of the test for setup and coverage.

tests/ci/integration/python_patch/main/aws-lc-cpython.patch
WillChilds-Klein: Yeah, once [cpython PR 135402](https://github.com/python/cpython/pull/135402) has been merged, we can just delete this hunk from the patch file.

crypto/fipsmodule/hmac/hmac.c
WillChilds-Klein: yeah, good call.

include/openssl/hmac.h
WillChilds-Klein: I've intentionally kept mentions of `KECCAK1600_CTX` out of the public headers. Developers will be able to read explanatory comments in `sha/internal.h` to figure out what's going on.

include/openssl/digest.h
WillChilds-Klein: SHA3's block size [reduces](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-224.ipd.pdf#section.3) as its output size grows. I would link to that SP here, but it's not applicable to all digests -- just those approved for HMAC by NIST.

crypto/fipsmodule/sha/internal.h
WillChilds-Klein: sure, makes sense.

include/openssl/hmac.h
WillChilds-Klein: I agree that it would be preferable to keep this internal and make `HMAC_CTX` fully opaque, but it's in our public headers today, so moving it has the potential to break downstream builds. I'd rather not make a breaking change unless we need to.

crypto/fipsmodule/hmac/hmac.c
fabrice102: My mistake. It should be removed. It was for an intermediary commit to show how the changes would look like. But this is no more true.

fabrice102:
This looks good to me.

Minor comment:

> SHA3 is unique in supported HMAC digests in that, due to differences between its sponge construction and others' Merkle-Dåmgard constructions, it does not support pre-computed keys. To accommodate this difference, we refactor the relevant code generation macros and relevant unit tests.

Technically, I think we could do the same and support pre-computed keys by exporting the state after absorbing the HMAC key. However, as mentioned by @WillChilds-Klein  in https://github.com/aws/aws-lc/pull/2484#discussion_r2165040720, exporting the state is not readily available. Furthermore, I don't think there is a good use case to support pre-computed keys for HMAC-SHA-3. If there ever is in the future, it would not be difficult to add this support and still be backward compatible.

So I agree that we should not be supporting HMAC-SHA-3 pre-computed keys.

fabrice102:
To complement my comment above, there is an important difference between HMAC-SHA-3 and HMAC-SHA-2 which renders HMAC-SHA-3 pre-computed keys much less useful.

Under one-wayness of the compression function of SHA-2, it is not possible to recover the original HMAC key from an HMAC-SHA-2 precomputed key (assuming the original HMAC key had enough entropy).

This is not the case for HMAC-SHA-3: it is possible to invert the last permutation of the sponge construction of SHA-3 and recover the original HMAC key from an HMAC-SHA-3 precomputed key.

crypto/fipsmodule/hmac/hmac.c
WillChilds-Klein: Thanks Fabrice. Removed in 4553a6db6.

samuel40791765 (approved):

andrewhop (approved):

PR #2490, opened 2025-06-17
Implement SSL_CTX_set_client_hello_cb for ClientHello callback

justsmth:
### Description
This pull request adds support for the `SSL_CTX_set_client_hello_cb` API which allows applications to register a callback that will be invoked when a TLS ClientHello message is received. 

- Implementation of `SSL_CTX_set_client_hello_cb` and supporting functions
- Added `SSL_client_hello_isv2` - always returns 0, compatibility with OpenSSL
- Added `SSL_client_hello_get0_ext` - extracts specific extensions from the ClientHello

### Call-outs
This does not support the `SSL_CLIENT_HELLO_RETRY` callback return value. It's currently treated the same as an error. 

#### File modified
- include/openssl/ssl.h: Added function declarations and documentation
- ssl/internal.h: Added internal callback structures
- ssl/handshake_server.cc: Implemented callback invocation during handshake
- ssl/ssl_lib.cc: Implemented the new API functions
- ssl/ssl_client_hello_test.cc: Added tests for the new functionality
- ssl/CMakeLists.txt: Updated to include new test file

### Testing
Tests exercise the callback functionality with various extension types.

### DIFF
--- a/include/openssl/ssl.h
+++ b/include/openssl/ssl.h
@@ -1121,7 +1121,7 @@ OPENSSL_EXPORT size_t SSL_get0_peer_delegation_algorithms(
 // The callback should return one of:
 // - SSL_CLIENT_HELLO_SUCCESS to continue the handshake
 // - SSL_CLIENT_HELLO_ERROR to abort the handshake with alert |*al|
-// - SSL_CLIENT_HELLO_RETRY to pause the handshake (not fully supported)
+// - SSL_CLIENT_HELLO_RETRY (not supported) is handled like SSL_CLIENT_HELLO_ERROR
 typedef int (*SSL_client_hello_cb_fn)(SSL *s, int *al, void *arg);
 
 // SSL_CTX_set_client_hello_cb configures a callback that is called when a
--- a/ssl/ssl_lib.cc
+++ b/ssl/ssl_lib.cc
@@ -3036,14 +3036,12 @@ void SSL_set_msg_callback_arg(SSL *ssl, void *arg) {
 }
 
 void SSL_CTX_set_client_hello_cb(SSL_CTX *c, SSL_client_hello_cb_fn cb,
-                                 void *arg)
-{
+                                 void *arg) {
   c->client_hello_cb = cb;
   c->client_hello_cb_arg = arg;
 }
 
-int SSL_client_hello_isv2(SSL *s)
-{
+int SSL_client_hello_isv2(SSL *s) {
   // SSLv2 not supported
   return 0;
 }


ssl/handshake_server.cc
```diff
@@ -715,6 +715,25 @@ static enum ssl_hs_wait_t do_read_client_hello_after_ech(SSL_HANDSHAKE *hs) {
...
+  /* Give the ClientHello callback a crack at things */
+  const SSL_CTX* sctx = ssl->ctx.get();
+  if (sctx != NULL && sctx->client_hello_cb != NULL) {
+    int al = SSL_AD_INTERNAL_ERROR;
+    /* A failure in the ClientHello callback terminates the connection. */
+    switch (sctx->client_hello_cb(ssl, &al, sctx->client_hello_cb_arg)) {
+      case SSL_CLIENT_HELLO_SUCCESS:
+        break;
+      case SSL_CLIENT_HELLO_RETRY:
+        // This case is treated as failure.
```
samuel40791765, line 728:
```suggestion
        // This case is treated as failure.
        
        OPENSSL_FALLTHROUGH;
```

include/openssl/ssl.h
```diff
@@ -1099,6 +1099,63 @@ SSL_get0_peer_verify_algorithms(const SSL *ssl, const uint16_t **out_sigalgs);
...
+// implementation.
+OPENSSL_EXPORT void SSL_CTX_set_client_hello_cb(SSL_CTX *c, SSL_client_hello_cb_fn cb,
+                                 void *arg);
+
+// SSL_client_hello_isv2 returns one if the current client hello was in SSLv2 format
+// and zero otherwise. This function can only be called from within a client hello
+// callback (see |SSL_CTX_set_client_hello_cb|) or during server certificate selection 
+// (see |SSL_CTX_set_select_certificate_cb|).
+OPENSSL_EXPORT int SSL_client_hello_isv2(SSL *s);
+// SSL_client_hello_get0_ext searches the extensions in the ClientHello for an
```
samuel40791765, line 1147: Not sure why the prepush checker isn't failing, but there should be a single blank line between the documentation for each separate definition/function

ssl/handshake_server.cc
samuel40791765, line 1: Comment styling should use `//` instead of `/* */` here and `internal.h`

ssl/ssl_lib.cc
```diff
@@ -3035,6 +3035,42 @@ void SSL_set_msg_callback_arg(SSL *ssl, void *arg) {
...
+    return 0;
+  }
+
+  CBS cbs;
+  if (!ssl_client_hello_get_extension(&client_hello, &cbs, (uint16_t)type)) {
+    return 0;  // Extension not found
+  }
+
+  *out = CBS_data(&cbs);
+  *outlen = CBS_len(&cbs);
```
samuel40791765: We should do null checks against `*out` and `*outlen` here.

include/openssl/ssl.h
```diff
@@ -1099,6 +1099,63 @@ SSL_get0_peer_verify_algorithms(const SSL *ssl, const uint16_t **out_sigalgs);
...
+// SSL_CLIENT_HELLO_SUCCESS, SSL_CLIENT_HELLO_ERROR, or SSL_CLIENT_HELLO_RETRY.
+// Note that SSL_CLIENT_HELLO_RETRY is not properly supported in the current
+// implementation.
+OPENSSL_EXPORT void SSL_CTX_set_client_hello_cb(SSL_CTX *c, SSL_client_hello_cb_fn cb,
+                                 void *arg);
+
+// SSL_client_hello_isv2 returns one if the current client hello was in SSLv2 format
+// and zero otherwise. This function can only be called from within a client hello
+// callback (see |SSL_CTX_set_client_hello_cb|) or during server certificate selection 
+// (see |SSL_CTX_set_select_certificate_cb|).
```
samuel40791765: We only return `0` in AWS-LC since SSLv2 isn't supported and the documentation should probably reflect that.

andrewhop (commented):
What is the configuration expectation from OpenSSL? Should this callback be copied into the SSL_CONFIG or SSL struct? Looking at SSK_new we copy some things out so the SSL_CTX can be modified later without modifying all existing SSL objects which might have a reference to that SSL_CTX.

include/openssl/ssl.h
```diff
@@ -1099,6 +1099,62 @@ SSL_get0_peer_verify_algorithms(const SSL *ssl, const uint16_t **out_sigalgs);
...
+// the handshake proceeds.
+//
+// The callback is invoked before most ClientHello processing and before the
+// decision whether to resume a session is made. It may inspect the ClientHello
+// and configure the connection accordingly.
+//
+// The callback function is passed the |SSL| object, a pointer to an alert value,
+// and the argument provided when the callback was set. It should return either
+// SSL_CLIENT_HELLO_SUCCESS, SSL_CLIENT_HELLO_ERROR, or SSL_CLIENT_HELLO_RETRY.
+// Note that SSL_CLIENT_HELLO_RETRY is not properly supported in the current
```
smittals2: do we want to note what exactly are the gaps with SSL_CLIENT_HELLO_RETRY?

ssl/ssl_lib.cc
```diff
@@ -3035,6 +3035,46 @@ void SSL_set_msg_callback_arg(SSL *ssl, void *arg) {
...
+{
+  // SSLv2 not supported
+  return 0;
+}
+
+int SSL_client_hello_get0_ext(SSL *s, unsigned int type, const unsigned char **out,
+                              size_t *outlen) {
+  GUARD_PTR(s);
+  GUARD_PTR(s->s3);
+  SSL_HANDSHAKE* hs = s->s3->hs.get();
```
smittals2, line 3053: hs.get() could segfault here? Or do we know that if s->s3 is valid, s->s3->hs is valid?

smittals2 (approved):

include/openssl/ssl.h
smittals2: After reading through PR it looks like its simply not supported (we treat it as an error in the switch statement). Should we just say as much here?

ssl/ssl_lib.cc
justsmth: It looks good to me. We check that neither `s` nor `s->s3` are NULL above. `s->s3->hs` is our smart pointer type, which should always have a valid state, then we check whether `hs` is NULL on the next line.

include/openssl/ssl.h
justsmth: Good call. It should indicate that `SSL_CLIENT_HELLO_RETRY` is treated the same as `SSL_CLIENT_HELLO_ERROR`. I'll fix in next commit.

include/openssl/ssl.h
```diff
@@ -1099,6 +1099,62 @@ SSL_get0_peer_verify_algorithms(const SSL *ssl, const uint16_t **out_sigalgs);
...
+# define SSL_CLIENT_HELLO_RETRY   (-1)
+
+// SSL_client_hello_cb_fn is a callback function that is called when a ClientHello
+// is received. The callback may inspect the ClientHello and make decisions 
+// about the connection based on its contents.
+//
+// The callback should return one of:
+// - SSL_CLIENT_HELLO_SUCCESS to continue the handshake
+// - SSL_CLIENT_HELLO_ERROR to abort the handshake with alert |*al|
+// - SSL_CLIENT_HELLO_RETRY to pause the handshake (not fully supported)
```
smittals2: is this true? Will RETRY pause the handshake or abort like ERROR?

ssl/ssl_lib.cc
```diff
@@ -3035,6 +3035,46 @@ void SSL_set_msg_callback_arg(SSL *ssl, void *arg) {
...
 
+void SSL_CTX_set_client_hello_cb(SSL_CTX *c, SSL_client_hello_cb_fn cb,
+                                 void *arg)
+{
+  c->client_hello_cb = cb;
+  c->client_hello_cb_arg = arg;
+}
+
+int SSL_client_hello_isv2(SSL *s)
+{
```
smittals2: nit: Move the opening brace in line with func def? Here and above

smittals2 (approved):

ssl/internal.h
```diff
@@ -3935,6 +3935,10 @@ struct ssl_ctx_st : public bssl::RefCounted<ssl_ctx_st> {
   // more than this due to padding and MAC overheads.
   uint16_t max_send_fragment = SSL3_RT_MAX_PLAIN_LENGTH;
 
+  /* ClientHello callback.  Mostly for extensions, but not entirely. */
```
samuel40791765, line 3938:
```suggestion
  // ClientHello callback.  Mostly for extensions, but not entirely.
```

samuel40791765 (approved):

PR #2493, opened 2025-06-19
Fix Console Test Suite Execution Locally

smittals2:
### Description of changes: 
PemPasswdTest suite of tests hang when running them locally on a terminal. Currently the code directly tries to connect to /dev/tty and if not available falls back to stdin and stdout. In the CI env, there is no /dev/tty so the redirection in the test suite works. But when running locally, the code bypasses this since a /dev/tty is available. 

This PR introduces an env var to force using stdin/stdout when running the test suite. This ensures in both CI and local environments, the stdin/stdout redirection works as intended with PemPasswdTest. 

### DIFF
--- a/crypto/console/console.c
+++ b/crypto/console/console.c
@@ -146,7 +146,6 @@ int openssl_console_open(void) {
     }
 
 #if !defined(OPENSSL_WINDOWS)
-    // Unix/Linux path
     if ((tty_in = fopen(DEV_TTY, "r")) == NULL) {
         tty_in = stdin;
     }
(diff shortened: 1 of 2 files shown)


crypto/console/console.c
```diff
@@ -136,6 +136,15 @@ int openssl_console_open(void) {
     is_a_tty = 1;
     assert(CRYPTO_STATIC_MUTEX_is_write_locked(&console_global_mutex));
 
+    // Check for test environment variable first (platform-independent)
+    const char* test_mode = getenv("OPENSSL_CONSOLE_TEST_MODE");
```
justsmth: NP: I'm not sure about the name. There might be use-cases for this outside of just our testing. I'd prefer a more general name like `AWSLC_NO_TTY_DETECT`.

justsmth (dismissed):

skmcgrail (dismissed):

skmcgrail (approved):

justsmth (approved):

PR #2501, opened 2025-06-23
Migrate Openssl-tool parameter parsing

justsmth:
### Description of changes: 
Follow-up from previous PR: https://github.com/aws/aws-lc/pull/2454
* Migrates the remainder of the subcommands to using `ordered_args` for parameter processing.
* Updates tests for rsa and crl to use reordered arguments.

### DIFF
--- a/tool-openssl/req.cc
+++ b/tool-openssl/req.cc
@@ -457,9 +457,10 @@ static bool generate_serial(X509 *cert) {
 }
 
 bool reqTool(const args_list_t &args) {
-  ordered_args::ordered_args_map_t parsed_args;
+  using namespace ordered_args;
+  ordered_args_map_t parsed_args;
   args_list_t extra_args;
-  if (!ordered_args::ParseOrderedKeyValueArguments(parsed_args, extra_args, args, kArguments) ||
+  if (!ParseOrderedKeyValueArguments(parsed_args, extra_args, args, kArguments) ||
       extra_args.size() > 0) {
     PrintUsage(kArguments);
     return false;
@@ -469,16 +470,16 @@ bool reqTool(const args_list_t &args) {
   unsigned int days;
   bool help = false, new_flag = false, x509_flag = false, nodes = false;
 
-  ordered_args::GetBoolArgument(&help, "-help", parsed_args);
-  ordered_args::GetBoolArgument(&new_flag, "-new", parsed_args);
-  ordered_args::GetBoolArgument(&x509_flag, "-x509", parsed_args);
-  ordered_args::GetBoolArgument(&nodes, "-nodes", parsed_args);
+  GetBoolArgument(&help, "-help", parsed_args);
+  GetBoolArgument(&new_flag, "-new", parsed_args);
+  GetBoolArgument(&x509_flag, "-x509", parsed_args);
+  GetBoolArgument(&nodes, "-nodes", parsed_args);
 
-  ordered_args::GetString(&newkey, "-newkey", "", parsed_args);
-  ordered_args::GetUnsigned(&days, "-days", 30u, parsed_args);
-  ordered_args::GetString(&subj, "-subj", "", parsed_args);
-  ordered_args::GetString(&keyout, "-keyout", "", parsed_args);
-  ordered_args::GetString(&out, "-out", "", parsed_args);
+  GetString(&newkey, "-newkey", "", parsed_args);
+  GetUnsigned(&days, "-days", 30u, parsed_args);
+  GetString(&subj, "-subj", "", parsed_args);
+  GetString(&keyout, "-keyout", "", parsed_args);
+  GetString(&out, "-out", "", parsed_args);
 
   if (help) {
     PrintUsage(kArguments);
--- a/tool-openssl/req.cc
+++ b/tool-openssl/req.cc
@@ -7,6 +7,8 @@
 #include <openssl/evp.h>
 #include <openssl/pem.h>
 #include <openssl/rsa.h>
+#include <algorithm>
+#include <iostream>
 #include <stdio.h>
 #include <string.h>
 #include "../tool/internal.h"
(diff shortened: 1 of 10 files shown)


tool-openssl/rsa_test.cc
```diff
@@ -227,3 +227,34 @@ TEST_F(RSAComparisonTest, RSAToolCompareModulusOutNooutOpenSSL) {
   trim(openssl_output_str);
   ASSERT_EQ(tool_output_str, openssl_output_str);
 }
+
+// Test against OpenSSL output "openssl rsa -in file -noout -modulus"
+// Only modulus is printed to stdin (reordered parameters)
+TEST_F(RSAComparisonTest, RSAToolCompareReorderedModulusNooutOpenSSL) {
+  std::string tool_command = std::string(tool_executable_path) + " rsa -in " + in_path + " -noout -modulus > " + out_path_tool;
+  std::string openssl_command = std::string(openssl_executable_path) + " rsa -in " + in_path + " -noout -modulus > " + out_path_openssl;
```
smittals2, line 235: since modulus is the only thing being printed out here can we add another option that outputs smthg to test the ordering?

tool-openssl/rsa_test.cc
```diff
@@ -227,3 +227,34 @@ TEST_F(RSAComparisonTest, RSAToolCompareModulusOutNooutOpenSSL) {
...
+
+  RunCommandsAndCompareOutput(tool_command, openssl_command, out_path_tool, out_path_openssl, tool_output_str, openssl_output_str);
+
+  ASSERT_EQ(tool_output_str, openssl_output_str);
+}
+
+// Test against OpenSSL output "openssl rsa -in file -noout -modulus -out out_file"
+// Only modulus is printed to output file (reordered parameters)
+TEST_F(RSAComparisonTest, RSAToolCompareReorderedModulusOutNooutOpenSSL) {
+  std::string tool_command = std::string(tool_executable_path) + " rsa -in " + in_path + " -noout -modulus -out " + out_path_tool;
```
smittals2, line 245: same as above I think

tool-openssl/rsa_test.cc
justsmth: I agree that these two tests are rather pointless given that only one thing can be output from this subcommand, so the ordering doesn't really matter. So I could delete the test, but I see little harm in having it as place holder were we to add other output options.

smittals2 (commented):
This isn't very consistent on openssl's part, but I don't think RSA is one of the cli tools that honors ordering in output.  I haven't verified the other tools though


Screenshot 2025-06-24 at 3 27 51 PM

samuel40791765 (dismissed):

tool-openssl/req.cc
```diff
@@ -9,6 +9,8 @@
 #include <openssl/rsa.h>
 #include <stdio.h>
 #include <string.h>
+#include <algorithm>
```
smittals2, line 10: req doesn't seem to do ordering either. I used `openssl req -in client.csr -noout -text -subject` and `openssl req -in client.csr -noout -subject -text` to test. (We don't support a lot of the output generating options yet so may not have an impact yet, but probably revert this tool too?)

tool-openssl/rsa.cc
```diff
@@ -1,38 +1,62 @@
...
 
-  GetBoolArgument(&help, "-help", parsed_args);
-  GetString(&in_path, "-in", "", parsed_args);
-  GetString(&out_path, "-out", "", parsed_args);
-  GetBoolArgument(&noout, "-noout", parsed_args);
-  GetBoolArgument(&modulus, "-modulus", parsed_args);
+  ordered_args::GetBoolArgument(&help, "-help", parsed_args);
+  ordered_args::GetString(&in_path, "-in", "", parsed_args);
+  ordered_args::GetString(&out_path, "-out", "", parsed_args);
+  ordered_args::GetBoolArgument(&noout, "-noout", parsed_args);
```
smittals2: why are we not doing a GetBoolArgument for modulus and instead handling it specially below? And are all tools now using ordered_args_map_t even if they don't need ordered args (I see the comment below on line 94 but should we just revert this to using the original args_map_t)?

tool-openssl/s_client.cc
```diff
@@ -29,20 +29,26 @@ static const argument_t kArguments[] = {
 };
 
 bool SClientTool(const args_list_t &args) {
-  std::map<std::string, std::string> args_map;
+  ordered_args::ordered_args_map_t parsed_args;
```
smittals2: Similar question as above, what is the value of migrating this to ordered_args_map_t when we have to convert it back anyway for DoClient?

tool-openssl/verify.cc
```diff
@@ -2,8 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0 OR ISC
 
 #include <openssl/base.h>
-#include <openssl/x509.h>
```
smittals2, line 5: I also don't see this tool in OpenSSL honoring order of input args.

tool-openssl/s_client.cc
justsmth: It would be good for our tools/commands to all parse their options in the same way. We previously had this consistency (i.e., `args_map_t`), until we discovered that it can't properly handle some use-cases that we have for it. 

So I'd like to migrate our tools over to the new mechanism incrementally through a series of PRs, instead of doing it all at once. The "bssl" tool might be next for migration, but it's not urgent.

tool-openssl/rsa.cc
justsmth: I considered using `GetBoolArgument`, but for this use-case it seems more noisy.  It would look like this:
```
  bool modulus_option = false;
  ordered_args::GetBoolArgument(&modulus_option, "-modulus", parsed_args);
  if (modulus_option) {
    ...
```

But I think the current logic is (slightly) more clear:
```
  if (ordered_args::HasArgument(parsed_args, "-modulus")) {
```

tool-openssl/verify.cc
justsmth: With commands for which we (currently) only produce a singular output, I think we can ignore for now. If (or when) we decide to support more options for this subcommand, we will ensure that the command output is consistent with OpenSSL whatever the parameter order might be.

tool-openssl/req.cc
justsmth: See my other responses. Essentially, my thinking is that we should migrate subcommands away from `args_map_t` and toward use of `ordered_args` everywhere -- it supports a superset of the use-cases supported by `args_map_t`.

smittals2 (approved):

samuel40791765 (approved):

PR #2504, opened 2025-06-24
Fix logical bug and add test for tree-DRBG fork

torben-hansen:
### Issues:

V1827078279

### Description of changes: 

Fixes a logical bug in the tree-DRBG `tree_jitter_check_drbg_must_reseed()` that would prevent a reseed under UBE.

Adds test coverage for tree-DRBG under UBE's. 

Must amend the retry logic for mingw due to the bug https://sourceforge.net/p/mingw-w64/bugs/883/. The existing mitigation for that issue wasn't sufficient anymore (for unknown reasons). I counted up to thousands of retries required.
Instead of relying on tight retries, we implement time-based backoff. This added another complication that `nanosleep()` might not be available on Arm64. To get around that on Arm64, use `Sleep`. Buuut, this function has millisecond resolution, so we don't apply exponential backoff, instead we backoff with 1 ms at a time, to avoid it exploding. x86 still use nanosleep.

### DIFF
--- a/crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
+++ b/crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
@@ -279,6 +279,7 @@ TEST_F(treeDrbgJitterentropyTest, BasicFork) {
       }
     );
 
+    tree_jitter_free_thread_drbg(&entropy_source);
     exit(exit_code ? 0 : 1);
   };
 
@@ -309,6 +310,7 @@ TEST_F(treeDrbgJitterentropyTest, BasicFork) {
             }, "child")
           )
 
+          tree_jitter_free_thread_drbg(&thread_entropy_source);
           *result = true;
         };
 
@@ -343,6 +345,7 @@ TEST_F(treeDrbgJitterentropyTest, BasicFork) {
       }
     );
 
+    tree_jitter_free_thread_drbg(&entropy_source);
     exit(exit_code ? 0 : 1);
   };
 
--- a/crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
+++ b/crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
@@ -235,6 +235,8 @@ static bool assertSeedOrReseed(const struct entropy_source_t *entropy_source,
     expect_reseed_thread, expect_reseed_global, error_text);
 }
 
+#if !defined(OPENSSL_WINDOWS)
+
 TEST_F(treeDrbgJitterentropyTest, BasicFork) {
 
   if (runtimeEmulationIsIntelSde() && addressSanitizerIsEnabled()) {
@@ -352,6 +354,8 @@ TEST_F(treeDrbgJitterentropyTest, BasicFork) {
   EXPECT_EXIT(testFuncSingleForkThenThread(), ::testing::ExitedWithCode(0), "");
 
 }
+
+#endif // !defined(OPENSSL_WINDOWS)
 
 TEST_F(treeDrbgJitterentropyTest, TreeDRBGThreadReseedInterval) {
 
--- a/crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
+++ b/crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
@@ -250,12 +250,12 @@ TEST_F(treeDrbgJitterentropyTest, BasicFork) {
           expect_reseed = 1;
         }
 
-        return assertReseed(&entropy_source, expect_reseed, [this, entropy_source]() {
+        return assertReseed(&entropy_source, expect_reseed, [entropy_source]() {
           uint8_t child_out[CTR_DRBG_ENTROPY_LEN];
           return tree_jitter_get_seed(&entropy_source, child_out);
         }, "child");
       },
-      [this, entropy_source]() {
+      [entropy_source]() {
         // In child. If UBE detection is supported, we expect a reseed.
         // No UBE detection is handled via prediction resistance.
         return assertReseed(&entropy_source, 0, [entropy_source]() {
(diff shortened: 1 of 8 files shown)


crypto/thread_pthread.c
```diff
@@ -66,20 +67,115 @@ void CRYPTO_MUTEX_cleanup(CRYPTO_MUTEX *lock) {
...
 // Some MinGW pthreads implementations might fail on first use of
 // locks initialized using PTHREAD_RWLOCK_INITIALIZER.
 // See: https://sourceforge.net/p/mingw-w64/bugs/883/
 typedef int (*pthread_rwlock_func_ptr)(pthread_rwlock_t *);
 static int rwlock_EINVAL_fallback_retry(const pthread_rwlock_func_ptr func_ptr, pthread_rwlock_t* lock) {
+
   int result = EINVAL;
 #ifdef __MINGW32__
-  const int MAX_ATTEMPTS = 10;
+  const int MAX_ATTEMPTS = 50;
```
dkostic: just curious, why the increase?

crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
```diff
@@ -202,6 +201,298 @@ TEST_F(treeDrbgJitterentropyTest, BasicReseed) {
...
+            )
+
+            return true;
+          }
+        );
+
+        return exit_code_child;
+      },
+      [entropy_source]() {
+        // In Parent we expect no reseed, even if UBE detection is not supported.
```
dkostic, line 406:
```suggestion
        // In Parent we expect no reseed, even if UBE detection is supported.
```
?

crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
```diff
@@ -202,6 +201,298 @@ TEST_F(treeDrbgJitterentropyTest, BasicReseed) {
...
+          assertSeedOrReseed(&entropy_source, expect_reseed, 0, [entropy_source]() {
+            uint8_t child_out[CTR_DRBG_ENTROPY_LEN];
+            return tree_jitter_get_seed(&entropy_source, child_out);
+          }, "child")
+        )
+
+        return true;
+      },
+      [entropy_source]() {
+        // In Parent we expect no reseed, even if UBE detection is not supported.
```
dkostic, line 338:
```suggestion
        // In Parent we expect no reseed, even if UBE detection is supported.
```
?

crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
```diff
@@ -202,6 +201,298 @@ TEST_F(treeDrbgJitterentropyTest, BasicReseed) {
...
+          assertSeedOrReseed(&entropy_source, expect_reseed, expect_reseed, [entropy_source]() {
+            uint8_t child_out[CTR_DRBG_ENTROPY_LEN];
+            return tree_jitter_get_seed(&entropy_source, child_out);
+          }, "child")
+        )
+
+        return true;
+      },
+      [entropy_source]() {
+        // In Parent we expect no reseed, even if UBE detection is not supported.
```
dkostic, line 272:
```suggestion
        // In Parent we expect no reseed, even if UBE detection is supported.
```
?

dkostic (dismissed):

crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
torben-hansen: Existing comment is correct: For DRBG not in the tree-DRBG, no UBE detection implies a reseed of the state. Not in the tree-DRBG though.

crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
torben-hansen: See previous comment.

crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
torben-hansen: See previous comment.

crypto/thread_pthread.c
torben-hansen: Arbitrarily, I wanted to hit about 1/2 second total backoff. 
That's 50 * 10ms.
So, actually I need to change these values and be dependent on the architecture...

torben-hansen: So, I made it about 1 second now for both cases.

crypto/thread_pthread.c
```diff
@@ -66,20 +67,115 @@ void CRYPTO_MUTEX_cleanup(CRYPTO_MUTEX *lock) {
...
+#ifndef MIN
+#define MINGW_AWSLC_MIN(X,Y) (((X) < (Y)) ? (X) : (Y))
+#else
+#define MINGW_AWSLC_MIN(X,Y) MIN(X,Y)
+#endif
+
+// MINGW implements nanosleep on X86_64 but not on other architectures.
+// Prefer using nanosleep because of its higher resolution.
+// https://gist.github.com/scivision/dbbbf33c2faf5a16f31fd6d144adc314 claims
+// that MINGW only implements nanpsleep on X86, but not Arm64. The linked
```
samuel40791765, line 81:
```suggestion
// that MINGW only implements nanosleep on X86, but not Arm64. The linked
```

crypto/thread_pthread.c
```diff
@@ -66,20 +67,115 @@ void CRYPTO_MUTEX_cleanup(CRYPTO_MUTEX *lock) {
...
+  //    9         100,000,000 nsec
+  //    ...
+  struct timespec sleep_time = {.tv_sec = 0, .tv_nsec = 0 };
+
+  // Cap backoff at 100,000,000 nsec (100ms).
+  *backoff = MINGW_AWSLC_MIN((*backoff) * 10, MINGW_MILLISECONDS_100);
+  // |nanosleep| can mutate |sleep_time|. Hence, we use |backoff| for state.
+  sleep_time.tv_nsec = *backoff;
+
+  nanosleep(&sleep_time, &sleep_time);
```
samuel40791765, line 118: For the second parameter, we pass in `sleep_time` to capture the remaining time, but we don't really handle the returned value (if there is any). Can we just pass in `NULL` here instead?
```suggestion
  nanosleep(&sleep_time, NULL);
```

crypto/thread_pthread.c
torben-hansen: follow-up item

crypto/thread_pthread.c
torben-hansen: handle later in follow-up

PR #2498, opened 2025-06-22
ML-KEM: Import AArch64 backend from mlkem-native

hanno-becker:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

---------

**Context:** The ML-KEM implementation in AWS-LC is imported from
mlkem-native. mlkem-native comes in a "C-only" version, but also
offers AArch64 and x86_64 backends for (a) arithmetic,
and (b) FIPS-202. Currently, only the "C-only" version is
imported into AWS-LC.

**Summary:** This commit extends the mlkem-native->AWS-LC import to include the AArch64 arithmetic backend.

**Details:**

- `crypto/fipsmodule/ml_kem/importer.sh` now imports
  the arithmetic backend API header `native/api.h` as well
  as the native backend `native/aarch64/*`.

- The backend is imported as-is, with one exception:
  `importer.sh` converts the preprocessor directives used by
  mlkem-native into the ones used by s2n-bignum. This is to
  piggy-back on the adjustments made to the delocator to work
  with s2n-bignum assembly; otherwise, similar adjustments would
  likely be needed for mlkem-native assembly files.

- All imported functions are formally verified for functional
  correctness using HOL-Light. The proofs run as part of
  mlkem-native's CI. The HOL-Light specs are manually translated
  into CBMC specs in the header accompanying the ASM, and
  all higher level CBMC proofs conducted against those specs.
  Again, those are part of the mlkem-native CI.

- A backend header crypto/fipsmodule/ml_kem/mlkem_native_backend.h
  is added, activating the AArch64 arithmetic backend on Linux and
  MacOS AArch64 system, except if the NO_ASM directive is set
  (same as for s2n-bignum).

  Once the x86_64 arithmetic backend is ready for integration,
  it will be added to `mlkem_native_backend.h` as well.

- The backend header is registered in the configuration file
  `crypto/fipsmodule/ml_kem/mlkem_native_config.h`.

- The importer.sh is re-run.

----------

Performance
```
Platform      | Algorithm   | Operation | New (ops/sec) | Old (ops/sec) | Improvement
--------------|-------------|-----------|---------------|---------------|-------------
Graviton 4    | ML-KEM-512  | keygen    |    29,145.9   |    19,666.0   |   +48.2%
              |             | encaps    |    91,043.3   |    64,240.2   |   +41.7%
              |             | decaps    |    75,242.9   |    52,486.8   |   +43.4%
[...]

PR #2510, opened 2025-06-26
s2n-bignum: Add prefix header to _s2n_bignum_internal.h

hanno-becker:

------------

AWS-LC's prefix build works by providing a 'prefix header' which prefixes global symbols through a series of `#define ...` directives. This header needs to be included ahead of the source files it applies to.

1. For C files, the prefix header is pulled in through openssl/base.h.
2. For perl-generated ASM files, it comes from openssl/asm_base.h which is `#include`d through `xxx-xlate.pl`.
3. In contrast to 1. and 2., for s2n-bignum assembly, the prefix header is indirectly included through an `--include` directive in the CMakeLists.txt. Similarly, the cc_builder in aws-lc-rs is patched to use such `--include` clause when building s2n-bignum files.

This commit simplifies and improves the transparency of prefix handling for s2n-bignum files (case 3): Instead of using an `--include` directive in the build files, it explicitly adds the prefix header in the header _s2n_bignum_internal.h. This header is included by all s2n-bignum assembly. This is functionally equivalent to the previous --include directive, but avoids having to patch CMakeLists.txt or cc_builder.

The cc_builder patch can be removed from aws-lc-rs when convenient -- this commit renders the `--include` directive redundant, but not harmful.

### DIFF
--- a/third_party/s2n-bignum/s2n-bignum-imported/include/_internal_s2n_bignum.h
+++ b/third_party/s2n-bignum/s2n-bignum-imported/include/_internal_s2n_bignum.h
@@ -1,3 +1,5 @@
+// Auto-added during import for AWS-LC symbol prefixing support
+#include <openssl/boringssl_prefix_symbols_asm.h>
 
 #ifdef __APPLE__
 #   define S2N_BN_SYMBOL(NAME) _##NAME
--- a/crypto/fipsmodule/CMakeLists.txt
+++ b/crypto/fipsmodule/CMakeLists.txt
@@ -293,14 +293,6 @@ if((((ARCH STREQUAL "x86_64") AND NOT MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX) OR
     )
   endif()
 
-  if(BORINGSSL_PREFIX)
-    # s2n-bignum is third-party code and therefore doesn't have an explicit
-    # definition of symbol prefixes under the prefix build. Inject prefix
-    # definitions instead. One could set this property for just the s2n-bignum
-    # source. But for simplicity, do it for all. The pre-processor will remove
-    # any duplicate header files.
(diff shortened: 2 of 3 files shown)


third_party/s2n-bignum/s2n-bignum-imported/include/_internal_s2n_bignum.h
```diff
@@ -1,3 +1,5 @@
+// Auto-added during import for AWS-LC symbol prefixing support
```
torben-hansen, line 1: It's because the general model is to not touch the imported files. I don't see a strong reason to change that.

In addition, if we were to ever change the import method (e.g. use a submodule), we wouldn't be able to change the imported files. The build have to work with them as-is.

hanno-becker: @torben-hansen Fair enough, but there is indeed good reason: It is more transparent to make the include explicit rather than hiding it in build system patches -- even one per build system. Concretely, here we need to patch both the CMake build, and the external build in aws-lc-rs. If you want to add further files to the import, such as considered in #2498, you can only do this while also patching aws-lc-rs, which I consider undesirable. You don't have this dependency with the proposed change.

Please reconsider and discuss with @dkostic @nebeid @andrewhop . I don't see a reason to be dogmatic about the 'no touching of imported files' policy. Also, functionally you still touch them through the `--include` as much as you touch them through the explicit `#include`.

dkostic: I think this is a reasonable change if it allows us to remove special handling of s2n-bignum files both in our CMake and the aws-lc-rs build. As for Torben's concern about modifying imported files -- would it work if we add something like:
```
#if defined(AWS_LC)
#include <openssl/the-prefix-header-file.h>
#endif 
```
to `_internal_s2n_bignum.h` in the upstream s2n-bignum repo?

dkostic (approved):

PR #2519, opened 2025-07-02
Add pkey command to CLI tool 

smittals2:
### Issues:
`CryptoAlg-3385`

### Description of changes: 
Added pkey command to the CLI tool with: 
-in, -out: Input/output file handling (supports stdin/stdout)
-inform, -outform: Format conversion between PEM and DER
-pubin, -pubout: Public key input/output modes
-text, -text_pub: Display key components in human-readable format
-noout: Suppress encoded key output
-help: Display usage information

This PR also updates RSA private key print format from `Private-Key: (N bit)` to `Private-Key: (N bit, 2 primes)`. This is for OpenSSL compatibility, AWS-LC does not support multi-prime RSA keys.

### Testing:
Comparison tests and option validation testing. 

### DIFF
--- a/tool-openssl/pkey.cc
+++ b/tool-openssl/pkey.cc
@@ -64,9 +64,11 @@ static bool WritePublicKey(EVP_PKEY *pkey, bssl::UniquePtr<BIO> &output_bio, int
 
 
 bool pkeyTool(const args_list_t &args) {
-  args_map_t parsed_args;
+  using namespace ordered_args;
+  ordered_args_map_t parsed_args;
   args_list_t extra_args;
-  if (!ParseKeyValueArguments(parsed_args, extra_args, args, kArguments) ||
+
+  if (!ParseOrderedKeyValueArguments(parsed_args, extra_args, args, kArguments) ||
       extra_args.size() > 0) {
     PrintUsage(kArguments);
     return false;
@@ -74,11 +76,10 @@ bool pkeyTool(const args_list_t &args) {
 
   std::string in_path, out_path, inform, outform;
   bool pubin = false, pubout = false;
-  bool noout = false, text = false, textpub = false, help = false;
+  bool noout = false, text = false, textpub = false;
   int input_format = FORMAT_PEM;
   int output_format = FORMAT_PEM;
 
-  GetBoolArgument(&help, "-help", parsed_args);
   GetString(&in_path, "-in", "", parsed_args);
   GetString(&out_path, "-out", "", parsed_args);
   GetString(&inform, "-inform", "", parsed_args);
@@ -90,7 +91,7 @@ bool pkeyTool(const args_list_t &args) {
   GetBoolArgument(&textpub, "-text_pub", parsed_args);
 
   // Display pkey tool option summary
-  if (help) {
+  if (HasArgument(parsed_args, "-help")) {
     PrintUsage(kArguments);
     return false;
   }


tool-openssl/pkey.cc
```diff
@@ -0,0 +1,215 @@
...
+  { "", kOptionalArgument, "" }
+};
+
+static bool isCharUpperCaseEqual(char a, char b) {
+  return ::toupper(a) == ::toupper(b);
+}
+
+static bool isStringUpperCaseEqual(const std::string &a, const std::string &b) {
+  return a.size() == b.size() && std::equal(a.begin(), a.end(), b.begin(), isCharUpperCaseEqual);
+}
```
skmcgrail: Any reason not to just put these in a common file, seems like we are repeating them often even if they are simple.

tool-openssl/pkey.cc
```diff
@@ -0,0 +1,215 @@
...
+
+#include <openssl/pem.h>
+#include <openssl/err.h>
+#include <openssl/evp.h>
+#include "internal.h"
+#include <string>
+
+// Define format constants
+#define FORMAT_PEM 1
+#define FORMAT_ASN1 2
```
skmcgrail:
```suggestion
#define FORMAT_DER 2
```
nit: This seems more accurate. PEM is just base64 encoded ASN1 (DER) with start and end block indicators. So feels like a misnomer to just call this ASN.1 IMO. Also aligns with th actual argument name.

tool-openssl/pkey.cc
```diff
@@ -0,0 +1,215 @@
...
+  GetBoolArgument(&pubin, "-pubin", parsed_args);
+  GetBoolArgument(&pubout, "-pubout", parsed_args);
+  GetBoolArgument(&noout, "-noout", parsed_args);
+  GetBoolArgument(&text, "-text", parsed_args);
+  GetBoolArgument(&textpub, "-text_pub", parsed_args);
+
+  // Display pkey tool option summary
+  if (help) {
+    PrintUsage(kArguments);
+    return false;
```
skmcgrail, line 96: Asking for help shouldn't return `false` right? Does this cause a non-zero exit code? That's not inline with my quick test of the OpenSSL CLI.

tool-openssl/pkey.cc
```diff
@@ -0,0 +1,215 @@
...
+}
+
+static bool WritePrivateKey(EVP_PKEY *pkey, bssl::UniquePtr<BIO> &output_bio, int output_format) {
+  if (output_format == FORMAT_ASN1) {
+    if (!i2d_PrivateKey_bio(output_bio.get(), pkey)) {
+      fprintf(stderr, "Error: error writing private key in DER format\n");
+      ERR_print_errors_fp(stderr);
+      return false;
+    }
+  } else {
```
samuel40791765, line 34: Nit: Would be better if this was `else if (output_format == FORMAT_PEM)`, since the logic within here is specific to that. And we should bail on `else` here, just like you're doing in the main function `pkeyTool`.

tool-openssl/pkey.cc
```diff
@@ -0,0 +1,215 @@
...
+    }
+  }
+
+  // Set up output BIO
+  bssl::UniquePtr<BIO> output_bio;
+  if (out_path.empty()) {
+    output_bio.reset(BIO_new_fp(stdout, BIO_NOCLOSE));
+  } else {
+    output_bio.reset(BIO_new(BIO_s_file()));
+    BIO_write_filename(output_bio.get(), out_path.c_str());
```
samuel40791765: Should check the return value here.

tool-openssl/pkey_test.cc
```diff
@@ -0,0 +1,365 @@
...
+  bssl::UniquePtr<RSA> rsa(RSA_new());
+  bssl::UniquePtr<BIGNUM> bn(BN_new());
+  if (!bn || !BN_set_word(bn.get(), RSA_F4) ||
+      !RSA_generate_key_ex(rsa.get(), 2048, bn.get(), nullptr) ||
+      !EVP_PKEY_assign_RSA(pkey.get(), rsa.release())) {
+    return nullptr;
+  }
+  
+  return pkey.release();
+}
```
samuel40791765: Seems like the function from the `pkcs8_test.cc`: https://github.com/aws/aws-lc/blob/06d88cfbd6cb8b8416757647b841bda7ccd54fce/tool-openssl/pkcs8_test.cc#L18C18-L36 

Could we abstract these? I like this version better though.

samuel40791765 (commented):
Agree with @skmcgrail `FORMAT_DER` comment, the PR mostly lgtm

tool-openssl/pkey.cc
smittals2: Changed it so only x509.c has the implementation, re-used in pkey. We can also move the implementation itself to internal.h as inline (I see some in test_util.h), not sure what the convention there is/which would be preferred?

tool-openssl/pkey.cc
smittals2: That's what we seem to do with all our tooling in tool-openssl. Idt return code on help option is something that would necessarily impact consuming applications, but in the spirit of being 1:1 with OpenSSL will look into this.

tool-openssl/pkey.cc
smittals2: We already validate output_format in main but it would be better defensively to be specific for any future additions. Made the change!

samuel40791765: Yeah I understand that part. I'm just nitpicking though, it's mainly because it felt like the logic was more "PEM" specific rather than `else` logic, but I don't think there's anything actually off.

samuel40791765 (dismissed):

skmcgrail (approved):

PR #2525, opened 2025-07-07
Export BF_cfb64_encrypt

samuel40791765:
### Issues:
Resolves `CryptoAlg-3304`

### Description of changes: 
`BF_cfb64_encrypt` was historically a public symbol in OpenSSL until BoringSSL made it internal. This exports the symbol for applications still dependent on it from OpenSSL.

### Call-outs:
N/A

### Testing:
Only exports the symbol, our existing test coverage should already have this covered.

### DIFF
--- a/include/openssl/blowfish.h
+++ b/include/openssl/blowfish.h
@@ -85,6 +85,9 @@ OPENSSL_EXPORT void BF_cbc_encrypt(const uint8_t *in, uint8_t *out,
                                    size_t length, const BF_KEY *schedule,
                                    uint8_t *ivec, int enc);
 
+OPENSSL_EXPORT void BF_cfb64_encrypt(const uint8_t *in, uint8_t *out, size_t length,
+                                     const BF_KEY *schedule, uint8_t *ivec, int *num,
+                                     int encrypt);
 
 #ifdef  __cplusplus
 }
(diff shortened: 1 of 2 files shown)


include/openssl/blowfish.h
```diff
@@ -85,6 +85,9 @@ OPENSSL_EXPORT void BF_cbc_encrypt(const uint8_t *in, uint8_t *out,
                                    size_t length, const BF_KEY *schedule,
                                    uint8_t *ivec, int enc);
 
+OPENSSL_EXPORT void BF_cfb64_encrypt(const uint8_t *in, uint8_t *out, size_t length,
```
WillChilds-Klein: not specific to your change, but should these functions be marked `OPENSSL_DEPRECATED`?

WillChilds-Klein (dismissed):

include/openssl/blowfish.h
samuel40791765: Great point, these are all deprecated in 3.0, I can add the marker to all of them here: https://docs.openssl.org/3.4/man3/BF_encrypt/#name.

Might as well do so now, since we almost never touch these functions.

WillChilds-Klein (approved):

skmcgrail (approved):

PR #2526, opened 2025-07-08
Implement SSL_set_verify_result

nhatnghiho:
### Issues:
Resolves CryptoAlg-3061

### Description of changes: 
Implement `SSL_set_verify_result`, which requires adding the field `verify_result` to the SSL struct. According to OpenSSL's implementation, the value of `verify_result` in the SSL struct and `SSL_SESSION` are always synced during the handshake. I mirrored closely that behavior. See document in CryptoAlg-3061's comment section for line-by-line functional parity comparison

### Testing:
Unit tests added


ssl/handshake.cc
```diff
@@ -390,6 +391,8 @@ enum ssl_verify_result_t ssl_verify_peer_cert(SSL_HANDSHAKE *hs) {
               : ssl_verify_invalid;
   }
 
+  ssl->verify_result = hs->new_session->verify_result;
```
justsmth: The documentation (and the OpenSSL implementation) seem to indicate that `ssl->verify_result` is only updated when non-fatal certificate errors occur.

Should the setting of it here be moved to the end of the function and conditioned on `ret == ssl_verify_ok`?
```
...
  if (ret == ssl_verify_ok) {
    ssl->verify_result = hs->new_session->verify_result;
  }

  return ret;
}
```

ssl/ssl_session.cc
```diff
@@ -800,6 +800,10 @@ void ssl_set_session(SSL *ssl, SSL_SESSION *session) {
   }
 
   ssl->session = UpRef(session);
+
+  if (session != NULL) {
+    ssl->verify_result = session->verify_result;
+  }
```
justsmth: NP: I think it would be cleaner to condition on and use the `ssl->session` (`UniquePtr`) value:
```
  if (ssl->session) {
    ssl->verify_result = ssl->session->verify_result;
  }
```

ssl/ssl_x509.cc
```diff
@@ -744,13 +744,14 @@ int SSL_CTX_load_verify_locations(SSL_CTX *ctx, const char *ca_file,
...
+
 long SSL_get_verify_result(const SSL *ssl) {
   check_ssl_x509_method(ssl);
-  SSL_SESSION *session = SSL_get_session(ssl);
-  if (session == NULL) {
-    return X509_V_ERR_INVALID_CALL;
-  }
-  return session->verify_result;
+  return ssl->verify_result;
 }
```
justsmth: Is it possible that we have consumers that depend on our current behavior, i.e., they expect this function to return `X509_V_ERR_INVALID_CALL` when there is no session? Should that behavior be preserved?
```
long SSL_get_verify_result(const SSL *ssl) {
  check_ssl_x509_method(ssl);
  if (SSL_get_session(ssl) == nullptr) {
    return X509_V_ERR_INVALID_CALL;
  }
  return ssl->verify_result;
}
```

ssl/handshake.cc
nhatnghiho: Hmm this is a good point. Could you point me to where in the documention/code that OpenSSL does that?

From my understanding, OpenSSL sets [`ssl->verify_result` first](https://github.com/openssl/openssl/blob/36614faa98c5a947a635d3f44e78c7c36b722534/ssl/ssl_cert.c#L513) then [`ssl->session->verify_result` later](https://github.com/openssl/openssl/blob/36614faa98c5a947a635d3f44e78c7c36b722534/ssl/statem/statem_clnt.c#L2169) if non-fatal certificate errors occur. But AWS-LC sets the session's `verify_result` first. Since OpenSSL users would expect `ssl->verify_result` to be the source of truth, I think keeping the SSL_SESSION and SSL struct's `verify_result` as in sync as possible would allow us to be more compatible with OpenSSL

justsmth: Sorry, I was confused.

My question is whether the `ssl->verify_result` assignment should be moved to be immediately before the return on line 417. Is it possible that the "verify_result" value changes due to the result of OCSP?

nhatnghiho: Gotcha. I was more focused about the condition on `ret == ssl_verify_ok`. I agree -- `ssl->verify_result` should be moved to before the return line. 

Though I would argue it doesn't need the `ret == ssl_verify_ok` condition. Let me know if you think otherwise!

```
...
ssl->verify_result = hs->new_session->verify_result;

return ret;
```

justsmth (approved):

smittals2 (commented):
as we're mucking around here, i noticed ssl_get_new_session in OpenSSL initializes SSL_SESSION with X509_V_OK but we do it with X509_V_ERR_INVALID_CALL. There's a comment there stating " The session is marked not resumable until it is completely filled in." Just want to verify the changes we're making now won't affect that workflow

nhatnghiho:
>  as we're mucking around here, i noticed ssl_get_new_session in OpenSSL initializes SSL_SESSION with X509_V_OK but we do it with X509_V_ERR_INVALID_CALL. There's a comment there stating " The session is marked not resumable until it is completely filled in." Just want to verify the changes we're making now won't affect that workflow

Yeah I noticed that too but we initializes SSL_SESSION's `verify_result` with `X509_V_ERR_INVALID_CALL`, so I thought we should do the same for SSL struct for better consistency. If none of our consumers have issues before, I suppose initializing SSL struct's `verify_result` with `X509_V_ERR_INVALID_CALL` should still be safe?

ssl/ssl_lib.cc
```diff
@@ -687,6 +688,7 @@ SSL *SSL_new(SSL_CTX *ctx) {
   ssl->config->permute_extensions = ctx->permute_extensions;
   ssl->config->aes_hw_override = ctx->aes_hw_override;
   ssl->config->aes_hw_override_value = ctx->aes_hw_override_value;
+  ssl->verify_result = X509_V_OK;
```
smittals2, line 691: I guess we end up overwriting the X509_V_ERR_INVALID_CALL when we initialize SSL here anyway

justsmth (approved):

ssl/ssl_lib.cc
nhatnghiho: I missed this. That's not what we want I think. I tried removing it and found a few new spots where I have to sync SSL's and SSL_SESSION's `verify_result`. Will update the code without this line

PR #2531, opened 2025-07-08
Enable X25519MLKEM769 by default

alexw91:
### Issues:
N/A

### Description of changes: 
Adds `X25519MLKEM768` KeyShare to AWS-LC's default SupportedGroups list. 

### Call-outs:

### Testing:


ssl/extensions.cc
```diff
@@ -307,6 +307,7 @@ bool ssl_client_hello_get_extension(const SSL_CLIENT_HELLO *client_hello,
 }
 
 static const uint16_t kDefaultGroups[] = {
+    SSL_GROUP_X25519_MLKEM768,
```
WillChilds-Klein, line 310: should we add `SSL_GROUP_SECP256R1_MLKEM768` here? perhaps also `SSL_GROUP_SECP384R1_MLKEM1024`?

ssl/ssl_version_test.cc
```diff
@@ -2328,7 +2328,27 @@ TEST_P(SSLVersionTest, PeerTmpKey) {
...
+  ASSERT_TRUE(Connect());
+  for (SSL *ssl : {client_.get(), server_.get()}) {
+    SCOPED_TRACE(SSL_is_server(ssl) ? "server" : "client");
+    EVP_PKEY *key = nullptr;
+
+    if (getVersionParam().version == TLS1_3_VERSION) {
+        // TLS 1.3 default should be using X5519MLKEM768 as the key exchange.
+        ERR_clear_error();
+        EXPECT_FALSE(SSL_get_peer_tmp_key(ssl, &key));
+        ErrorEquals(ERR_get_error(), ERR_LIB_SSL, SSL_R_UNKNOWN_KEY_EXCHANGE_TYPE);
```
WillChilds-Klein: You have a good explanation for this in the PR overview:

> ssl_version_test.cc was updated to expect SSL_R_UNKNOWN_KEY_EXCHANGE_TYPE when calling SSL_get_peer_tmp_key() on an SSL connection that offered X25519MLKEM768. This is because there is no EVP_PKEY type for hybrid keys, only individual X25519 or MLKEM768 keys.

can we distill that into a source comment here?

ssl/extensions.cc
alexw91: Done

ssl/ssl_version_test.cc
alexw91: Added comment.

darylmartin100:
FYI: I want to hold off on merging this in until we can better quantify the impact.

WillChilds-Klein (approved):

PR #2534, opened 2025-07-10
Integrating LIEF parser for SHARED builds

prikhap:
Title: Add CI workflow for C++ inject_hash implementation

Description:
This PR adds CI workflow and testing infrastructure for the C++ inject_hash implementation.

Changes:
- Add GitHub Actions workflow for testing
- Add test script for building and verifying functionality
- Test on both Ubuntu and macOS environments

Testing:
- Verified workflow locally
- Tests build process
- Tests FIPS integrity verification


util/fipstools/inject_hash_cpp/test_sha256.cpp
samuel40791765, line 1: Is this file used right now? I understand that it might come up in later implementations, but we can leave until then to pull this in.

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -0,0 +1,58 @@
...
+fi
+
+# First try to build just inject_hash_cpp
+echo "Building inject_hash_cpp..."
+ninja inject_hash_cpp
+
+# Test inject_hash_cpp basic functionality
+echo "Testing inject_hash_cpp basic functionality..."
+./util/fipstools/inject_hash_cpp/inject_hash_cpp --version || echo "Version check failed"
+./util/fipstools/inject_hash_cpp/inject_hash_cpp --help || echo "Help check failed"
```
samuel40791765: This doesn't seem to be used, could we remove this?

CMakeLists.txt
```diff
@@ -1,5 +1,31 @@
 cmake_minimum_required(VERSION 3.5..3.31)
 
+
+####################
```
samuel40791765: The comment/text here is redundant.

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# Build everything
+echo "Building full project..."
+ninja
+
+# Test actual FIPS injection
```
samuel40791765: Redundant comment here as well.

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -0,0 +1,58 @@
...
+        -DFIPS_SHARED=1 \
+        -DBUILD_SHARED_LIBS=1 \
+        -DUSE_CPP_INJECT_HASH=ON \
+        -DCMAKE_OSX_SYSROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk \
+        ..
+else
+    cmake -GNinja \
+        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DFIPS=1 \
+        -DFIPS_SHARED=1 \
```
samuel40791765: We should also be testing the shared build with ubuntu. And we should use `-DCMAKE_BUILD_TYPE=Release` since that's what we validate for FIPS :)

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -0,0 +1,58 @@
...
+        -DBUILD_SHARED_LIBS=1 \
+        -DUSE_CPP_INJECT_HASH=ON \
+        ..
+fi
+
+# First try to build just inject_hash_cpp
+echo "Building inject_hash_cpp..."
+ninja inject_hash_cpp
```
samuel40791765: I don't think `inject_hash_cpp` is necessarily used here. Calling `ninja` with the build flags you've set up should be sufficient enough.

We could also probably drop the `ninja` call on L49 below in favor of this line.

util/fipstools/inject_hash_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+        ${CMAKE_CURRENT_BINARY_DIR}
+        ${CMAKE_SOURCE_DIR}/include
+    )
+
+    if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+        target_compile_options(inject_hash_cpp PRIVATE
+            -Wno-overloaded-virtual
+            -Wno-unused-parameter
+        )
+    endif()
```
samuel40791765, line 21: Are these additional linker flags needed? AWS-LC already enables all warning linker flags in the build, so this could be redundant.

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -0,0 +1,37 @@
...
+        }
+    } else {
+        std::cerr << "❌ No binary path provided" << std::endl;
+        return 1;
+    }
+
+    std::cout << "\nReceived arguments:" << std::endl;
+    for (int i = 1; i < argc; i++) {
+        std::cout << "  Arg " << i << ": " << argv[i] << std::endl;
+    }
```
samuel40791765: This is probably useful for debugging, but we should drop it since we don't want debugging prints in most production code.

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -0,0 +1,58 @@
...
+./util/fipstools/inject_hash_cpp/inject_hash_cpp --help || echo "Help check failed"
+
+# Build everything
+echo "Building full project..."
+ninja
+
+# Test actual FIPS injection
+echo "Testing FIPS injection..."
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    ./util/fipstools/inject_hash_cpp/inject_hash_cpp -o ./crypto/libcrypto.dylib -in-object ./crypto/libcrypto.dylib -apple
```
samuel40791765: `-o ./crypto/libcrypto.dylib` and `-apple` don't seem to actually be used here. Same for below.

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -0,0 +1,37 @@
...
+    std::cout << "\n=== C++ inject_hash starting ===" << std::endl;
+    std::cout << "Testing LIEF integration..." << std::endl;
+
+    const char* binary_path = nullptr;
+    for (int i = 1; i < argc - 1; i++) {
+        if (strcmp(argv[i], "-in-object") == 0) {
+            binary_path = argv[i + 1];
+            break;
+        }
+    }
```
samuel40791765: There's probably a more graceful way to handle this. Could you look into pulling in the argument parsing functions in `tool/internal.h`?
* https://github.com/aws/aws-lc/blob/main/tool/internal.h#L107-L114

`ParseKeyValueArguments` should help you parse what you need in a cleaner and more easily extendable fashion.

tests/ci/run_inject_hash_cpp.sh
prikhap: static *

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: I just kept this for testing purpose. will keep this in mind while making the final file

util/fipstools/inject_hash_cpp/CMakeLists.txt
prikhap: LIEF has it's own warning, which were resulting in an error. Therefore used these flags to disable -WError

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -0,0 +1,61 @@
...
+echo "Environment:"
+go version
+cmake --version
+ninja --version
+
+# Build AWS-LC with C++ inject_hash
+mkdir -p build
+cd build
+
+# Configure based on build type and OS
```
andrewhop: Nothing is actually building aws-lc, this is only running CMake to configure the build.

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -0,0 +1,61 @@
...
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    # macOS only supports shared builds
+    cmake -GNinja \
+        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DFIPS=1 \
+        -DFIPS_SHARED=1 \
+        -DBUILD_SHARED_LIBS=1 \
+        -DUSE_CPP_INJECT_HASH=ON \
+        -DCMAKE_OSX_SYSROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk \
+        ..
```
andrewhop: There are existing helper functions defined in `common_posix_setup.sh` to help simplify this.

util/fipstools/inject_hash_cpp/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+
+    target_link_libraries(inject_hash_cpp PRIVATE
+        LIEF::LIEF
+    )
+
+    set_target_properties(inject_hash_cpp PROPERTIES
+        CXX_STANDARD 17
+        CXX_STANDARD_REQUIRED ON
+    )
```
andrewhop: Why 17?

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -0,0 +1,34 @@
+#include <LIEF/LIEF.hpp>
+#include <iostream>
+#include <string>
+
+int main(int argc, char** argv) {
+    std::cout << "\n=== C++ inject_hash starting ===" << std::endl;
```
andrewhop, line 20: Nothing appears to be running this, I don't see it in any of the GitHub action output.

.github/workflows/test-cpp-inject-hash.yml
```diff
@@ -0,0 +1,49 @@
...
+      if: runner.os == 'macOS'
+      run: |
+        brew install ninja cmake perl go
+        
+    - name: Build and Test
+      env:
+        BUILD_TYPE: ${{ matrix.build-type }}
+      run: |
+        chmod +x tests/ci/run_inject_hash_cpp.sh
+        ./tests/ci/run_inject_hash_cpp.sh
```
andrewhop: Why not check the bash script in with the correct file permissions?

util/fipstools/inject_hash_cpp/inject_hash.cpp
samuel40791765: I think it's not running because the last few lines in `tests/ci/run_inject_hash_cpp.sh` were accidentally commented out.

util/fipstools/inject_hash_cpp/test_sha256.cpp
prikhap: Removed it!

tests/ci/run_inject_hash_cpp.sh
prikhap: Removed it!

tests/ci/run_inject_hash_cpp.sh
prikhap: missed adding the ninja command.

util/fipstools/inject_hash_cpp/CMakeLists.txt
prikhap: Had kept it since LIEF requires this version of C++. However, we don't have to bother about the version while building aws-lc, hence deleted the properties from CMake.

.github/workflows/test-cpp-inject-hash.yml
prikhap: done

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: I mistakenly commented commands to run inject_hash, should work now.

tests/ci/run_inject_hash_cpp.sh
prikhap: Tried handling static files, which require creating an archive file parser (left it for some other PR)

tests/ci/run_inject_hash_cpp.sh
prikhap: Done

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: the inject_hash.cpp currently is just a test file, added some test functionalities and std:cout to see if the file was actually called.

tests/ci/run_inject_hash_cpp.sh
prikhap: Modified according to the CMake compiler flags

tests/ci/run_inject_hash_cpp.sh
prikhap: included in the recent commit.

CMakeLists.txt
```diff
@@ -1,5 +1,28 @@
...
+    set(LIEF_DOC OFF CACHE BOOL "")
+    set(LIEF_INSTALL ON CACHE BOOL "")
+    set(LIEF_STATIC ON CACHE BOOL "")
+    set(LIEF_ELF ON CACHE BOOL "Enable ELF format")
+    set(LIEF_PE ON CACHE BOOL "Enable PE format")
+    set(LIEF_MACHO ON CACHE BOOL "Enable MACHO format")
+    set(LIEF_DEX OFF CACHE BOOL "Disable DEX format")
+    set(LIEF_VDEX OFF CACHE BOOL "Disable VDEX format")
+    set(LIEF_ART OFF CACHE BOOL "Disable ART format")
+    set(LIEF_OAT OFF CACHE BOOL "Disable OAT format")
```
andrewhop: It is much more useful if your comment explains what these options are and why you chose these specific values.

CMakeLists.txt
```diff
@@ -1,5 +1,28 @@
...
+    
+    # Add LIEF subdirectory
+    add_subdirectory(third_party/lief)
```
andrewhop: It is not necessary to comment every line, especially when they are self explanatory like this. 
```suggestion
    add_subdirectory(third_party/lief)
```

CMakeLists.txt
```diff
@@ -1,5 +1,28 @@
...
+    
+    # Make LIEF headers available
+    include_directories(${CMAKE_CURRENT_SOURCE_DIR}/third_party/lief/include)
```
andrewhop: This makes the LIEF headers available to every target, can you scope this down to just the inject hash tool?

crypto/CMakeLists.txt
```diff
@@ -652,14 +652,28 @@ if(FIPS_SHARED)
       set(INJECT_HASH_APPLE_FLAG "-apple")
     endif()
 
+    if(USE_CPP_INJECT_HASH)
+    # Add dependency on C++ implementation
```
andrewhop: Indentation looks off for the if and else block, more superfluous comments

crypto/CMakeLists.txt
```diff
@@ -652,14 +652,28 @@ if(FIPS_SHARED)
...
+    add_dependencies(crypto inject_hash_cpp)
+    
+    # C++ implementation for hash injection
+    add_custom_command(
+      TARGET crypto POST_BUILD
+      COMMAND ${CMAKE_BINARY_DIR}/util/fipstools/inject_hash_cpp/inject_hash_cpp
+      -o $<TARGET_FILE:crypto>
+      -in-object $<TARGET_FILE:crypto>
+      ${INJECT_HASH_APPLE_FLAG}
```
andrewhop: What are you using this flag for?

crypto/fipsmodule/CMakeLists.txt
```diff
@@ -389,15 +389,26 @@ if(FIPS_DELOCATE)
...
+#  if(USE_CPP_INJECT_HASH)
+#     add_custom_command(
+#         OUTPUT bcm.o
+#         COMMAND ../../util/fipstools/inject_hash_cpp/inject_hash_cpp
+#         -o bcm.o
+#         -in-archive $<TARGET_FILE:bcm_hashunset>
+#         DEPENDS bcm_hashunset inject_hash_cpp
+#         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+#     )
+# else()
```
andrewhop: I prefer not checking in unused/commented out code. It's likely to get out of date and require rewriting in the future.
```suggestion
# else()
```

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -0,0 +1,60 @@
...
+# Source common setup
+source tests/ci/common_posix_setup.sh
+
+# Build AWS-LC with C++ inject_hash
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    # macOS only supports shared builds
+    run_build \
+        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DFIPS=1 \
+        -DFIPS_SHARED=1 \
```
andrewhop: Our CMakeList figures this out automatically
```suggestion
```

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -0,0 +1,60 @@
...
+        -DBUILD_SHARED_LIBS=1 \
+        -DUSE_CPP_INJECT_HASH=ON \
+        -DCMAKE_OSX_SYSROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
```
andrewhop: Why are you specifying this?

crypto/fipsmodule/CMakeLists.txt
```diff
@@ -389,15 +389,26 @@ if(FIPS_DELOCATE)
...
-
-  go_executable(inject_hash
-                boringssl.googlesource.com/boringssl/util/fipstools/inject_hash)
-  add_custom_command(
-    OUTPUT bcm.o
-    COMMAND ./inject_hash -o bcm.o -in-archive $<TARGET_FILE:bcm_hashunset>
-    DEPENDS bcm_hashunset inject_hash
-    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
-  )
+  #ignoring static builds for now
```
samuel40791765: It would be helpful to document the reason in the comments, but we can do that in the next PR.

samuel40791765 (approved):

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -0,0 +1,60 @@
...
+        -DBUILD_SHARED_LIBS=1 \
+        -DUSE_CPP_INJECT_HASH=ON
+    # fi
+fi
+
+cd "${BUILD_ROOT}"
+
+# Test actual FIPS injection
+echo "TESTING INJECT_HASH.CPP WITH EDGE CASES..." 
+./util/fipstools/inject_hash_cpp/inject_hash_cpp || echo "--Expected failure with no args--"
```
andrewhop: Should this fail the build if this isn't an error? Also I prefer moving more of this test logic into actual test files like our tools.

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -0,0 +1,60 @@
...
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    ./util/fipstools/inject_hash_cpp/inject_hash_cpp \
+        -o ./crypto/libcrypto.dylib \
+        -in-object ./crypto/libcrypto.dylib \
+        -apple
+else
+    ./util/fipstools/inject_hash_cpp/inject_hash_cpp \
+        -o ./crypto/libcrypto.so \
+        -in-object ./crypto/libcrypto.so
+fi
```
andrewhop: Isn't this run (with the correct paths) automatically during the normal FIPS build?

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -0,0 +1,61 @@
...
+            i++; // Skip next argument
+        }
+        else if (strcmp(argv[i], "-apple") == 0) {
+            is_apple = true;
+        }
+    }
+
+    // Validate arguments
+    if (!input_file || !output_file) {
+        std::cerr << "❌ Missing required arguments" << std::endl;
```
andrewhop: It is more helpful to say which ones are missing to help future customers/developers debug the issue.

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -0,0 +1,61 @@
...
+            input_file = argv[i + 1];
+            i++; // Skip next argument
+        }
+        else if (strcmp(argv[i], "-o") == 0) {
+            output_file = argv[i + 1];
+            i++; // Skip next argument
+        }
+        else if (strcmp(argv[i], "-apple") == 0) {
+            is_apple = true;
+        }
```
andrewhop: Why do we need to handle Apple in a special way?

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -0,0 +1,61 @@
...
+    std::cout << "Usage: inject_hash_cpp -o output_file -in-object input_file [-apple]" << std::endl;
+}
+
+int main(int argc, char** argv) {
+    std::cout << "\n=== C++ inject_hash starting ===" << std::endl;
+
+    // Parse command line arguments
+    const char* input_file = nullptr;
+    const char* output_file = nullptr;
+    bool is_apple = false;
```
andrewhop: Is it possible to re-use any of our existing argument parsing from our tools?

CMakeLists.txt
prikhap: check LIEF_BUILD_OPTIONS at : https://lief.re/doc/latest/installation.html#cmake-integration

CMakeLists.txt
prikhap: LIEF headers are scoped to inject_hash_cpp using the util/fipstools/inject_hash_cpp/CMakeLists.txt. I added it in the global CMake, to make the LIEF available throughout, but it surely can be removed and it is better directed in it's desired scope.

crypto/CMakeLists.txt
prikhap: amdended!

crypto/CMakeLists.txt
prikhap: these flags currently aren't of much usage, but was trying maintain similarity with the flag format used by the inject_hash.go script. Currently these flags are just printed as a part of testing the integration of the C++ script, however while creating a proper parsing and hashing script, these flags would be required to have the input path of the binary, output path and if it is suited for apple or linux.

tests/ci/run_inject_hash_cpp.sh
prikhap: this was added to the test script just to show contrast between "expected failure" cases and a normal case.

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: It doesen't serve any purpose right now, was just setting foundations for the future builds.

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: modified.

crypto/fipsmodule/CMakeLists.txt
prikhap: modified.

util/fipstools/inject_hash_cpp/inject_hash.cpp
samuel40791765: I also brought this up in https://github.com/aws/aws-lc/pull/2534#discussion_r2202011937. The answer seemed to imply that this would part of a larger future refactor/PR?

CMakeLists.txt
```diff
@@ -1,5 +1,32 @@
...
+if(USE_CPP_INJECT_HASH)
+    # Configure LIEF build options first :  check LIEF_BUILD_OPTIONS at : https://lief.re/doc/latest/installation.html#cmake-integration
+    set(LIEF_PYTHON_API OFF CACHE BOOL "") # disabling LIEF's Python API
+    # Disabling examples and tests as we only need the core library
+    set(LIEF_EXAMPLES OFF CACHE BOOL "") 
+    set(LIEF_TESTS OFF CACHE BOOL "")
+    # disabling LIEF's documentation generation
+    set(LIEF_DOC OFF CACHE BOOL "")
+    # enabling installation of LIEF
+    set(LIEF_INSTALL ON CACHE BOOL "")
```
andrewhop: Why do we need to install LIEF?

crypto/CMakeLists.txt
```diff
@@ -652,14 +652,27 @@ if(FIPS_SHARED)
...
-      TARGET crypto POST_BUILD
-      COMMAND ${GO_EXECUTABLE} run
-      ${AWSLC_SOURCE_DIR}/util/fipstools/inject_hash/inject_hash.go
-      -o $<TARGET_FILE:crypto> -in-object $<TARGET_FILE:crypto> ${INJECT_HASH_APPLE_FLAG}
-      WORKING_DIRECTORY ${AWSLC_SOURCE_DIR}
-    )
-
+    if(USE_CPP_INJECT_HASH)
+      # Add dependency on C++ implementation
+      add_dependencies(crypto inject_hash_cpp)
```
andrewhop: Do you think this needs a comment?

crypto/fipsmodule/CMakeLists.txt
```diff
@@ -389,15 +389,17 @@ if(FIPS_DELOCATE)
...
+    #ignoring static builds for now as archive files require a seperate parser to extract .o files from it. The lief parser does not support archive files.
+    go_executable(inject_hash
+        boringssl.googlesource.com/boringssl/util/fipstools/inject_hash)
+    add_custom_command(
+        OUTPUT bcm.o
+        COMMAND ./inject_hash -o bcm.o -in-archive $<TARGET_FILE:bcm_hashunset>
+        DEPENDS bcm_hashunset inject_hash
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+    )
+# endif()
```
andrewhop: Besides the comment did anything change here? If not we shouldn't change any of the whitespace because it makes tracking actual changes in git history harder.

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -0,0 +1,99 @@
...
+        -DFIPS=1 \
+        -DBUILD_SHARED_LIBS=1 \
+        -DUSE_CPP_INJECT_HASH=ON 
+else
+    run_build \
+        -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+        -DFIPS=1 \
+        -DBUILD_SHARED_LIBS=1 \
+        -DUSE_CPP_INJECT_HASH=ON
+fi
```
andrewhop: Is there any difference between the if and else block?

andrewhop (approved):

PR #2536, opened 2025-07-10
Use SP 800-56Arev3 Section 5.6.2.1.4.b instead of ECDSA PCT method

torben-hansen:
### Description of changes: 

The FIPS build type is currently employing an "ECDSA PCT" on the EC key generation code-path and EC key check code-path. This has explosive performance impact.

Our key generation is agnostic (and totally unaware) to which algorithm type a key is being generated for. Hence switch method to "SP 800-56Arev3 Section 5.6.2.1.4.b PCT" (the "New PCT"). For key generation primitives, this will increase performance by up 4-5x. For an ECDH workflow, performance could increase by 2-3x.

Funnily, the current implementation performs the ECDSA PCT but also does the New PCT in `EC_KEY_check_key()` irrespective of FIPS build type or not. Basic impl is to just delete the ECDSA PCT.

Hence, if things were easy, the only thing we must do is to remove the existing ECDSA PCT. But, the `ECDSA_PWCT` test is embedded in the ECDSA PCT in the function `EVP_EC_KEY_check_fips()`. So, we must find a new place for that. Do this by flipping a bit in the private key instead. Also change the name to `EC_PWCT` which seems a tad more precise.

### Call outs

While I was here:
* `EC_KEY_keygen_verify_service_indicator((EC_KEY*)key);` does not need to be cast to non-const.
* Fix-up the service indicator comments and explicitly include the service indicator function declarations.
* Document various optimizations that might be possible by doing more bookkeeping.

### DIFF
--- a/crypto/fipsmodule/ec/ec_key.c
+++ b/crypto/fipsmodule/ec/ec_key.c
@@ -80,6 +80,7 @@
 
 #include "internal.h"
 #include "../delocate.h"
+#include "../service_indicator/internal.h"
 #include "../../internal.h"
 
 
@@ -292,18 +293,19 @@ void EC_KEY_set_conv_form(EC_KEY *key, point_conversion_form_t cform) {
   }
 }
 
+// ec_key_gen_pct computes the PCT: SP 800-56Arev3 Section 5.6.2.1.4 option ‘b’.
+// Should only be used for NIST P-curves.
 static int ec_key_gen_pct(const EC_KEY *key) {
 
-  // SP 800-56Arev3 Section 5.6.2.1.4 option ‘b’
-  //
-  // Technically, not needed for ECDH in non-FIPS. However, other APIs use this
-  // function for validations e.g. the EVP private key parsing. This part is
-  // also open for optimization later.
   if (key->priv_key != NULL) {
 
     EC_SCALAR *priv_key_scalar = &key->priv_key->scalar;
 
-    EC_SCALAR priv_key_scalar_bit_flipped = {0};
+    // In theory, we could just flip a bit in the existing private key. The
+    // test below is supposed to hard abort and the code-path is only active
+    // during testing the break KAT framework. But prefer to keep it local to
+    // the PCT to isolate errors.
+    EC_SCALAR priv_key_scalar_bit_flipped = {{0}};
     if (boringssl_fips_break_test("EC_PWCT")) {
       OPENSSL_memcpy(&priv_key_scalar_bit_flipped, priv_key_scalar,
         sizeof(priv_key_scalar->words));
@@ -321,10 +323,10 @@ static int ec_key_gen_pct(const EC_KEY *key) {
     // Leaking this comparison only leaks whether |key|'s public key was
     // correct.
     //
-    // |ec_GFp_simple_points_equal| is quite expensive (relatively). If we were
-    // sure that points aren't in jacobian, but in affine representation, then
-    // the comparison reduces to simple byte array equality. The current
-    // comparison function does a non-negligible amount of field arithmetic.
+    // |ec_GFp_simple_points_equal| is quite expensive (relatively); The current
+    // comparison function does a non-negligible amount of field arithmetic If
+    // we were sure that points aren't in jacobian, but in affine
+    // representation, then the comparison reduces to fast byte array equality.
     if (!constant_time_declassify_int(ec_GFp_simple_points_equal(
           key->group, &point, &key->pub_key->raw))) {
       OPENSSL_PUT_ERROR(EC, EC_R_INVALID_PRIVATE_KEY);
@@ -352,6 +354,12 @@ int EC_KEY_check_key(const EC_KEY *eckey) {
     return 0;
   }
 
+  // Many code-paths end up in |EC_KEY_check_key|. For example, ECDH operations.
+  // Technically, the PCT is not needed for ECDH in non-FIPS. However, other
+  // APIs use this function for validations e.g. the EVP private key parsing (to
+  // validate that public key is indeed generate from the private key) and there
+  // is currently no way to distinguish the code-paths at this level. This could
+  // be optimized.
   if (!ec_key_gen_pct(eckey)) {
     return 0;
   }
@@ -360,8 +368,12 @@ int EC_KEY_check_key(const EC_KEY *eckey) {
 }
 
 int EC_KEY_check_fips(const EC_KEY *key) {
-  FIPS_service_indicator_lock_state();
+
   int ret = 0;
+
+  // Nothing obvious will falsely increment the service indicator. But lock to
+  // be safe.
+  FIPS_service_indicator_lock_state();
 
   if (EC_KEY_is_opaque(key)) {
     // Opaque keys can't be checked.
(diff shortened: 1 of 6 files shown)


justsmth:
There's a [test failure](https://us-west-2.codebuild.aws.amazon.com/project/eyJlbmNyeXB0ZWREYXRhIjoiWk5IUWJGRGxBcTJ2Mkp4WGF3dnBwYjc5V0ZZYSt5SVVGbkwvODkydTNTaVQ2V2FMN3hwa0tjSWNFemw2QWtCWW5welFWV3lpRFpKVitwejgvelhpRWh3NDNqcWhKalpPYW9hL2tLMDlJSDFPT1NkNyIsIml2UGFyYW1ldGVyU3BlYyI6Ilp4VXRNQXFGM1BZYVlaRkIiLCJtYXRlcmlhbFNldFNlcmlhbCI6MX0%3D/build/fe8e9f56-0a3f-495d-9c60-0bb64806ed5b):
```

[ RUN      ] FIPSCallback.PWCT
--
1977 | ../crypto/fips_callback_test.cc:142: Failure
1978 | Value of: EC_KEY_generate_key_fips(key.get())
1979 | Actual: true
1980 | Expected: false
1981 | Google Test trace:
1982 | ../crypto/fips_callback_test.cc:133: ECDSA_PWCT
1983 | [  FAILED  ] FIPSCallback.PWCT (135 ms)
```

torben-hansen:
> There's a [test failure](https://us-west-2.codebuild.aws.amazon.com/project/eyJlbmNyeXB0ZWREYXRhIjoiWk5IUWJGRGxBcTJ2Mkp4WGF3dnBwYjc5V0ZZYSt5SVVGbkwvODkydTNTaVQ2V2FMN3hwa0tjSWNFemw2QWtCWW5welFWV3lpRFpKVitwejgvelhpRWh3NDNqcWhKalpPYW9hL2tLMDlJSDFPT1NkNyIsIml2UGFyYW1ldGVyU3BlYyI6Ilp4VXRNQXFGM1BZYVlaRkIiLCJtYXRlcmlhbFNldFNlcmlhbCI6MX0%3D/build/fe8e9f56-0a3f-495d-9c60-0bb64806ed5b):
> 
> ```
> 
> [ RUN      ] FIPSCallback.PWCT
> --
> 1977 | ../crypto/fips_callback_test.cc:142: Failure
> 1978 | Value of: EC_KEY_generate_key_fips(key.get())
> 1979 | Actual: true
> 1980 | Expected: false
> 1981 | Google Test trace:
> 1982 | ../crypto/fips_callback_test.cc:133: ECDSA_PWCT
> 1983 | [  FAILED  ] FIPSCallback.PWCT (135 ms)
> ```

Indeed. I hadn't yet taken care of that part. Now my proposed changes should pushed.

justsmth (approved):

crypto/fipsmodule/ec/ec_key.c
```diff
@@ -332,49 +337,44 @@ int EC_KEY_check_key(const EC_KEY *eckey) {
...
+    OPENSSL_PUT_ERROR(EC, ERR_R_PASSED_NULL_PARAMETER);
+    return 0;
   }
-  sig_der = OPENSSL_malloc(sign_len);
-  if (!sig_der ||
-      !EVP_DigestSign(&ctx, sig_der, &sign_len, msg, msg_len)) {
-    goto err;
+
+  if (EC_POINT_is_at_infinity(eckey->group, eckey->pub_key)) {
+    OPENSSL_PUT_ERROR(EC, EC_R_POINT_AT_INFINITY);
```
WillChilds-Klein, line 347: minor non-blocking nit: this line [isn't covered](https://app.codecov.io/gh/aws/aws-lc/pull/2536?src=pr&el=tree&filepath=crypto%2Ffipsmodule%2Fec%2Fec_key.c&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws#ca56d08b044e64d556fe6f76cb59aeb5-R347) by testing. are the PCTs executed during deserialization for ECDSA keys (obviously not applicable to ECDHE)? if so, we could cover this case with a simple KAT w/ point at infinity.

torben-hansen: Not sure what codecov actually does, but some tests take this code-path.

crypto/fipsmodule/ec/ec_key.c
```diff
@@ -292,38 +293,42 @@ void EC_KEY_set_conv_form(EC_KEY *key, point_conversion_form_t cform) {
...
 }
 
-int EC_KEY_check_key(const EC_KEY *eckey) {
-  if (!eckey || !eckey->group || !eckey->pub_key) {
-    OPENSSL_PUT_ERROR(EC, ERR_R_PASSED_NULL_PARAMETER);
-    return 0;
-  }
+// ec_key_gen_pct computes the PCT: SP 800-56Arev3 Section 5.6.2.1.4 option ‘b’.
+// Should only be used for NIST P-curves.
+static int ec_key_gen_pct(const EC_KEY *key) {
```
dkostic, line 298: seems like the function returns true when `key->priv_key == NULL`, is that intended?

crypto/fipsmodule/ec/ec_key.c
```diff
@@ -292,38 +293,42 @@ void EC_KEY_set_conv_form(EC_KEY *key, point_conversion_form_t cform) {
...
+    // Leaking this comparison only leaks whether |key|'s public key was
     // correct.
+    //
+    // |ec_GFp_simple_points_equal| is quite expensive (relatively); The current
+    // comparison function does a non-negligible amount of field arithmetic If
+    // we were sure that points aren't in jacobian, but in affine
+    // representation, then the comparison reduces to fast byte array equality.
     if (!constant_time_declassify_int(ec_GFp_simple_points_equal(
-            eckey->group, &point, &eckey->pub_key->raw))) {
+          key->group, &point, &key->pub_key->raw))) {
```
dkostic, line 331: should we check `key->pub_key != NULL` before dereferencing?

crypto/fipsmodule/ec/ec_key.c
```diff
@@ -332,49 +337,44 @@ int EC_KEY_check_key(const EC_KEY *eckey) {
...
-  EVP_MD_CTX ctx;
-  EVP_MD_CTX_init(&ctx);
-  const EVP_MD *hash = EVP_sha256();
-  size_t sign_len;
-  if (!evp_pkey ||
-      !EVP_PKEY_set1_EC_KEY(evp_pkey, key) ||
-      !EVP_DigestSignInit(&ctx, NULL, hash, NULL, evp_pkey) ||
-      !EVP_DigestSign(&ctx, NULL, &sign_len, msg, msg_len)) {
-    goto err;
+int EC_KEY_check_key(const EC_KEY *eckey) {
```
dkostic, line 340: does it make sense to simplify the code and leave only `EC_KEY_check_fips` now that the performance difference in FIPS/non_FIPS is not that big?

crypto/fipsmodule/ec/ec_key.c
torben-hansen: Yes, because we can't execute the PCT on such a key. Same as existing behaviour.

crypto/fipsmodule/ec/ec_key.c
torben-hansen: We could possibly do that. Would leave it for a future PR though.

crypto/fipsmodule/ec/ec_key.c
torben-hansen: Right now the code-path will never hit this if public key doesn't exist. Maintained by non-static callers.

dkostic (approved):

PR #2538, opened 2025-07-13
Fix PKCS12 Error Code

smittals2:
### Issues:
`CryptoAlg-3347`

### Description of changes: 
Serf forces a failure and checks for the PKCS12 error library and  PKCS12_R_MAC_VERIFY_FAILURE  error code in a specific test. This change aligns our PKCS12_Parse implementation to align with OpenSSL for this test case. 

### DIFF
--- a/include/openssl/pkcs8.h
+++ b/include/openssl/pkcs8.h
@@ -307,8 +307,9 @@ BSSL_NAMESPACE_END
 #define PKCS8_R_UNSUPPORTED_OPTIONS 132
 #define PKCS8_R_AMBIGUOUS_FRIENDLY_NAME 133
 
-
-// PKCS#12 Error code defined for compatibility
-#define PKCS12_R_MAC_VERIFY_FAILURE 113
+// PKCS12_R_MAC_VERIFY_FAILURE is an error code defined for
+// compatability. It points to our equivalent for this OpenSSL error,
+// |PKCS8_R_INCORRECT_PASSWORD|
+#define PKCS12_R_MAC_VERIFY_FAILURE PKCS8_R_INCORRECT_PASSWORD
 
 #endif  // OPENSSL_HEADER_PKCS8_H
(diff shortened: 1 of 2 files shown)


include/openssl/pkcs8.h
```diff
@@ -307,9 +307,8 @@ BSSL_NAMESPACE_END
 #define PKCS8_R_UNSUPPORTED_OPTIONS 132
 #define PKCS8_R_AMBIGUOUS_FRIENDLY_NAME 133
 
-// PKCS12_R_MAC_VERIFY_FAILURE is an error code defined for
-// compatability. It points to our equivalent for this OpenSSL error,
-// |PKCS8_R_INCORRECT_PASSWORD|
-#define PKCS12_R_MAC_VERIFY_FAILURE PKCS8_R_INCORRECT_PASSWORD
+
+// PKCS#12 Error code defined for compatibility
+#define PKCS12_R_MAC_VERIFY_FAILURE 113
```
WillChilds-Klein: > Serf forces a failure and checks for the PKCS12 error library and PKCS12_R_MAC_VERIFY_FAILURE error code in a specific test.

Do you have a link to the test? My inclination is that test should reference the macro-defined constant, not [the underlying numerical value from OpenSSL](https://github.com/openssl/openssl/blob/4337989667b003bbf7ddc10984129d3cba298e8b/include/openssl/pkcs12err.h#L40). If that's the case, then the underlying numerical value shouldn't matter.

smittals2: You are correct, the only change we needed is changing the err lib from pkcs8 to pkcs12 (since we define both and they have different numerical values). The actual error codes' value does not need to change, I validated this internally and will revert this part of the change. Will send links on slack

smittals2: Responded offline :)

WillChilds-Klein (approved):

PR #2540, opened 2025-07-14
Improve OpenSSL compatibility

justsmth:
### Description of changes: 
A few small changes to improve OpenSSL compatibility:
* Provide an (empty) "openssl/ui.h" file to prevent compiler from loading OpenSSL's "openssl/ui.h" from system headers.
* Define `PEMerr` as alias for `OPENSSL_PUT_ERROR`
```
// For compatibility with OpenSSL. First argument ignored.
#define PEMerr(f, r) OPENSSL_PUT_ERROR(PEM, (r))
```
* Define `SSL_verify_client_post_handshake` as a no-op function:
```
int SSL_verify_client_post_handshake(SSL *ssl) {
  return 0;
}
```

### DIFF
--- a/ssl/ssl_lib.cc
+++ b/ssl/ssl_lib.cc
@@ -3601,7 +3601,6 @@ OPENSSL_EXPORT int SSL_get_write_traffic_secret(
 
 // No-op function for compatibility with OpenSSL.
 int SSL_verify_client_post_handshake(SSL *ssl) {
-  (void)ssl;
   return 0;
 }
 
--- a/ssl/ssl_lib.cc
+++ b/ssl/ssl_lib.cc
@@ -3599,3 +3599,9 @@ OPENSSL_EXPORT int SSL_get_write_traffic_secret(
   return 1;
 }
 
+// No-op function for compatibility with OpenSSL.
+int SSL_verify_client_post_handshake(SSL *ssl) {
+  (void)ssl;
+  return 0;
+}
+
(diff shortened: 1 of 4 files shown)


ssl/ssl_lib.cc
```diff
@@ -3599,3 +3599,9 @@ OPENSSL_EXPORT int SSL_get_write_traffic_secret(
   return 1;
 }
 
+// No-op function for compatibility with OpenSSL.
+int SSL_verify_client_post_handshake(SSL *ssl) {
+  (void)ssl;
```
WillChilds-Klein: what's the purpose of this?

justsmth: It's to avoid an "unused parameter" warning/error (i.e., build failure) that compilers might produce when parameters go unused by a function. I did this preemptively, as I'm not certain whether any of the compilers in our CI produce this warning.

For example: https://github.com/aws/aws-lc/blob/d7d349949d208f9bb5c2417bae017df55df5285f/crypto/mem.c#L130

justsmth: Actually, I think that warning would only be generated on a static function, so it's not necessary here.

WillChilds-Klein: That makes sense for static functions where the compiler can just inline `return 0` within the compilation unit.

smittals2 (approved):

PR #2555, opened 2025-07-17
Centralize password handling tool-openssl

kingstjo:
### Issues:

Addresses CryptoAlg-3387
Addresses CryptoAlg-3383
Addresses AWS-LC-863

### Description of changes:

This PR creates new centralized password handling utility for AWS-LC tool-openssl commands:

1. **New Password Utility Implementation:**
   - Created `pass_util.cc` and `pass_util.h` providing unified password functionality for tool-openssl
   - Implemented `ExtractPassword` for single password extraction from various sources
   - Implemented `ExtractPasswords` for dual password extraction with same-file support
   - Added `ValidateSource` helper for consistent validation logic
   - Implemented `SensitiveStringDeleter` for secure memory cleanup
   - Uses `PEM_BUFSIZE` for consistent buffer sizing

2. **API Features:**
   - Supports `pass:`, `file:`, `env:`, `stdin`, and `fd` password sources for tool-openssl commands
   - `fd` is not supported for Windows
   - Consistent in-place parameter modification
   - Same-file handling: reads first line for passin, second line for passout
   - Comprehensive validation with clear error messages

3. **Design:**
   - Modular architecture with dedicated helper functions
   - Centralized validation logic for tool-openssl password handling
   - Memory-safe with proper cleanup and secure string deletion

### Call-outs:

- Follows OpenSSL's password handling behavior and conventions
- Same-file optimization matches OpenSSL's dual password file format
- Updates pkcs8.cc `validate_bio_size` parameter to take reference over pointer

### Testing:

- Created comprehensive `pass_util_test.cc` with full coverage:
  - `ExtractPassword` with various sources (direct, file, env) and edge cases
  - `ExtractPasswords` with different files, same file, and mixed sources
  - Same-file logic validation (first/second line reading)
  - Error handling for invalid formats, missing files, and null pointers
  - Memory safety and `SensitiveStringDeleter` validation
  - Cross-platform environment variable handling


tool-openssl/password.cc
```diff
@@ -0,0 +1,142 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
smittals2: pls rename this file to util, tool-openssl-util, or smthg similar. We don't want to create separate source files per util - and also theres an actual openssl tool called passwd we may have to add later :).

tool-openssl/internal.h
```diff
@@ -24,13 +24,40 @@ struct Tool {
...
+X509_CRL *createTestCRL();
 bool isStringUpperCaseEqual(const std::string &a, const std::string &b);
 
+// Password handling namespace
+namespace password {
+// Custom deleter for sensitive strings that clears memory before deletion
+// This ensures passwords are securely removed from memory when no longer needed
+void SensitiveStringDeleter(std::string *str);
+
+// Handles both passin and passout options, with optimization for when the same source is used
```
smittals2: does optimization need to be called out here? if it is important for the comments, please specify what the optimization is?

tool-openssl/internal.h
```diff
@@ -24,13 +24,40 @@ struct Tool {
 
 bool IsNumeric(const std::string &str);
 
-X509* CreateAndSignX509Certificate();
-X509_CRL* createTestCRL();
+X509 *CreateAndSignX509Certificate();
+X509_CRL *createTestCRL();
 bool isStringUpperCaseEqual(const std::string &a, const std::string &b);
 
+// Password handling namespace
```
smittals2: feel like we're missing a place where we state this is a common utility for handling passin passout options across all openssl cli commands somehwehre.

tool-openssl/password.cc
```diff
@@ -0,0 +1,142 @@
...
+    if (!extracted_passin) {
+      return false;
+    }
+    *passin_arg = std::move(extracted_passin);
+  }
+  
+  // Handle passout if provided
+  if (passout_arg && *passout_arg) {
+    // Check if both arguments point to the same source and passin was already processed
+    bool same_source = (passin_arg && *passin_arg && passout_arg && *passout_arg &&
```
smittals2: I understand the optimization is trying to avoid redundant operations, and correct me if im wrong, but the optimization I'm seeing here is that at max we prevent one extra file read or env var lookup? You still have to create a copy of the string, and if password is a string ("pass:") then this optimization doesn't really help?

My take is we should remove this from here. Adding a special case and making the code more complex to prevent 1 extra file read/env var lookup in a non-perfomance constrained CLI tooling context doesn't seem that beneficial? Simpler code would improve maintainability and reduce testing overhead. 

Feel free to disagree, just my opinion.

tool-openssl/internal.h
```diff
@@ -24,13 +24,40 @@ struct Tool {
...
+// Returns true on success, false on failure.
+bool HandlePassOptions(bssl::UniquePtr<std::string> *passin_arg,
+                      bssl::UniquePtr<std::string> *passout_arg);
+
+// Extracts password from various sources (direct input, file, environment)
+// Takes a secure string as input and returns a secure string with the extracted
+// password. Supports formats:
+// - pass:password (direct password)
+// - file:/path/to/file (password from file)
+// - env:VAR_NAME (password from environment variable)
```
smittals2: pls add info about return values here on failure

tool-openssl/password.cc
```diff
@@ -0,0 +1,142 @@
...
+        return result;
+    }
+
+    // Password from file: file:/path/to/file
+    if (source_str.compare(0, 5, "file:") == 0) {
+        std::string path = source_str.substr(5);
+        bssl::UniquePtr<BIO> bio(BIO_new_file(path.c_str(), "r"));
+        if (!bio) {
+            fprintf(stderr, "Cannot open password file\n");
+            return bssl::UniquePtr<std::string>();
```
smittals2: why this and not just return NULL?

tool-openssl/password_test.cc
```diff
@@ -0,0 +1,234 @@
...
+      EXPECT_FALSE(passout) << "Output password should be null for " << params.description;
+    }
+  } else {
+    EXPECT_FALSE(result) << "HandlePassOptions should fail for " << params.description;
+  }
+}
+
+// Instantiate password source test cases
+INSTANTIATE_TEST_SUITE_P(
+    PasswordSources,
```
smittals2: can we move this up belwo the relevant test

tool-openssl/password_test.cc
```diff
@@ -0,0 +1,234 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include <gtest/gtest.h>
```
smittals2: can we add some testing for edge cases like truncation of file, empty env var, etc. Also i dont think we're actually testing reading from a valid file here, only invalid file?

tool-openssl/internal.h
kingstjo: HandlePassOptions is the entry point for this usage, and has comments for this.
I can better encapsulate this in internal.h making it easier to use.

tool-openssl/password.cc
kingstjo: updated ExtractPassword signature to return bool and update source string

tool-openssl/password.cc
kingstjo: You are correct, this optimization was at the expense of complex implementation.
As I was not taking advantage of all optimizations OSSL, I have reverted these extra pieces

tool-openssl/internal.h
kingstjo: Optimization removed, as the gain did not warrant complexity added

tool-openssl/password_test.cc
kingstjo: You are correct, this is a good callout. Thank you
Adding more cases

tool-openssl/pass_util.cc
```diff
@@ -0,0 +1,262 @@
...
+// compatibility with PEM functions and password callbacks throughout AWS-LC
+
+// Implementation details - not exposed in header
+static const uint8_t kPasswordSourceNone = 0;
+static const uint8_t kPasswordSourcePass = 1;
+static const uint8_t kPasswordSourceFile = 2;
+static const uint8_t kPasswordSourceEnv = 3;
+
+// Detect the type of password source
+static uint8_t DetectSource(const bssl::UniquePtr<std::string> &source) {
```
andrewhop: Why is this function taking a unique pointer to a std::string? std::string will handle memory for you automatically.

tool-openssl/pass_util.cc
```diff
@@ -0,0 +1,262 @@
...
+  return true;
+}
+
+// Extract password directly provided with pass: prefix
+static bool ExtractDirectPassword(bssl::UniquePtr<std::string> &source) {
+  // Check for additional colons in password portion after prefix
+  if (source->find(':', 5) != std::string::npos) {
+    fprintf(stderr, "Invalid password format (use pass:, file:, or env:)\n");
+    return false;
+  }
```
andrewhop, line 98: Is this a pkcs8 or OpenSSL limitation? Why not support all special characters in the password?

tool-openssl/pass_util.cc
```diff
@@ -0,0 +1,262 @@
...
+
+  // Check length before modification
+  if (source->length() - 5 > PEM_BUFSIZE) {
+    fprintf(stderr, "Password exceeds maximum allowed length\n");
+    return false;
+  }
```
andrewhop, line 105: It would be helpful to tell the caller what the limit is when they hit it.

tool-openssl/pass_util.cc
```diff
@@ -0,0 +1,262 @@
...
+  int len = BIO_gets(bio.get(), buf, sizeof(buf));
+  if (len <= 0) {
+    OPENSSL_cleanse(buf, sizeof(buf));
+    fprintf(stderr, "Cannot read password file\n");
+    return false;
+  }
+
+  const bool possible_truncation =
+      (static_cast<size_t>(len) == PEM_BUFSIZE - 1 && buf[len - 1] != '\n' &&
+       buf[len - 1] != '\r');
```
andrewhop: This looks weird but I think the logic is correct, I would suggest removing the outermost parenthesis and adding one pair around the length check. Initially I thought you were bitwise anding the length and last byte together.

tool-openssl/pass_util.cc
```diff
@@ -0,0 +1,262 @@
...
+    fprintf(stderr, "Password file content too long\n");
+    return false;
+  }
+
+  // Trim trailing newlines
+  size_t buf_len = len;
+  while (buf_len > 0 &&
+         (buf[buf_len - 1] == '\n' || buf[buf_len - 1] == '\r')) {
+    buf[--buf_len] = '\0';
+  }
```
andrewhop: Is this consistent with OpenSSL? I could imagine a world where someone generates a random password in a file that might include a byte with value `10`

tool-openssl/pass_util.cc
```diff
@@ -0,0 +1,262 @@
...
+#include "internal.h"
+
+// Use PEM_BUFSIZE (defined in openssl/pem.h) for password buffer size to ensure
+// compatibility with PEM functions and password callbacks throughout AWS-LC
+
+// Implementation details - not exposed in header
+static const uint8_t kPasswordSourceNone = 0;
+static const uint8_t kPasswordSourcePass = 1;
+static const uint8_t kPasswordSourceFile = 2;
+static const uint8_t kPasswordSourceEnv = 3;
```
andrewhop: Why not an enum?

tool-openssl/pass_util_test.cc
```diff
@@ -0,0 +1,369 @@
...
+  
+  EXPECT_TRUE(pass_util::ExtractPasswords(passin, passout));
+  EXPECT_EQ(*passin, "testpassword");
+  EXPECT_EQ(*passout, "anotherpassword");
+}
+
+// Test ExtractPasswords with same file (critical functionality!)
+TEST_F(PassUtilTest, ExtractPasswordsSameFile) {
+  // Create file with two lines
+  WriteTestFile(pass_path, "firstpassword\nsecondpassword\n", true);
```
andrewhop, line 308: Do you have tests for Windows carriage return behavior?

tool-openssl/pass_util_test.cc
```diff
@@ -0,0 +1,369 @@
...
+  // Verify the buffer contains our password
+  ASSERT_EQ(memcmp(buffer_ptr, test_password, str->length()), 0)
+      << "Buffer doesn't contain expected password data";
+
+  // Call the deleter
+  pass_util::SensitiveStringDeleter(str);
+
+  // The original string content should still be intact for comparison
+  EXPECT_EQ(original_content, test_password)
+      << "Original content was unexpectedly modified";
```
andrewhop: All of these comments are redundant.

tool-openssl/pass_util_test.cc
```diff
@@ -0,0 +1,369 @@
...
+  const char *buffer_ptr = str->data();
+
+  // Verify the buffer contains our password
+  ASSERT_EQ(memcmp(buffer_ptr, test_password, str->length()), 0)
+      << "Buffer doesn't contain expected password data";
+
+  // Call the deleter
+  pass_util::SensitiveStringDeleter(str);
+
+  // The original string content should still be intact for comparison
```
andrewhop: Is this test missing the most important test that checks the data in str is all 0?

tool-openssl/pkcs8.cc
```diff
@@ -84,141 +77,59 @@ static const std::unordered_set<std::string> kSupportedPRFs = {
...
-        OPENSSL_cleanse(&(*str)[0], str->size());
-        str->clear();
-    }
-    delete str;
-}
-
-BSSL_NAMESPACE_BEGIN
-BORINGSSL_MAKE_DELETER(std::string, SensitiveStringDeleter)
-BSSL_NAMESPACE_END
+// The SensitiveStringDeleter is now defined in password.cc
```
andrewhop: Comment is out of date, also no need to say where code used to be Git history will track all of that.

tool-openssl/pkcs8.cc
```diff
@@ -84,141 +77,59 @@ static const std::unordered_set<std::string> kSupportedPRFs = {
...
+    if (p8inf) {
+      pkey.reset(EVP_PKCS82PKEY(p8inf.get()));
+      if (pkey) {
+        return pkey;
+      }
+    }
+    BIO_reset(in_bio);
+    pkey.reset(d2i_PrivateKey_bio(in_bio, nullptr));
+    return pkey;
+  } else {
```
andrewhop: This could use some comments, I don't follow the logic for keys with and without a PKCS8_PRIV_KEY_INFO. Can any of this be combined with the logic above in the `if(passin)` branch. Are there any other utility functions in AWS-LC that supports parsing all these keys with and without passwords?

tool-openssl/pkcs8.cc
```diff
@@ -230,153 +141,160 @@ static const argument_t kArguments[] = {
     {"-topk8", kBooleanArgument, "Convert traditional format to PKCS#8"},
     {"-nocrypt", kBooleanArgument, "Use unencrypted private key"},
     {"-v2", kOptionalArgument, "Use PKCS#5 v2.0 and specified cipher"},
-    {"-v2prf", kOptionalArgument, "Use specified PRF algorithm with PKCS#5 v2.0"},
+    {"-v2prf", kOptionalArgument,
+     "Use specified PRF algorithm with PKCS#5 v2.0"},
```
andrewhop, line 113: Please add a new interoperability test for this new parameter.

tool-openssl/pass_util.cc
kingstjo: This is to match OpenSSL behavior. The colon would need to be escaped for input and would be treated as part of the password string. 
Using pass: with -passin or -passout is the least preferred option and users can use file: if a colon char is needed in the password string.

tool-openssl/pass_util.cc
kingstjo: Good point. I can add PEM_BUFSIZE into error output

tool-openssl/pkcs8.cc
kingstjo: -v2prf was not added with this PR and was part of pkcs8.cc implementation #2342 
Not sure why GitHub is showing this as an addition and deletion for this.

tool-openssl/pass_util.cc
kingstjo: -passin and -passout will be using the smart pointer with the custom deleter created for sensitive strings.

This matches all functions in pass_util.cc taking in a reference to the same pointer from the origin tool call.

tool-openssl/pass_util.cc
kingstjo: Had an enum but FILE was shadowing stdio.h
I can better implement this within pass_util namespace and even it's own class as this will avoid the shadow issue.

Good call on questioning enum over uint8_t as enum will be safer at compile time.

tool-openssl/pass_util.cc
kingstjo: This will remove newline and carriage returns from the buffer and numerical values of 10 in the file will be preserved as the password.

tool-openssl/pass_util_test.cc
kingstjo: You're right. We aren't verifying the data has been updated to 0000000.

I was running into issues testing this having issues viewing data after it had already been freed.
We can remove this test from CLI tests as it is testing `OPENSSL_cleanse` and `bssl:UniquePtr `which if tested should be tested on the library side.

tool-openssl/pass_util_test.cc
kingstjo: No I am not testing Windows carriage return behavior with this.
Thank you for this call-out. I can update setup and teardown to use carriage returns if in Windows environments.

tool-openssl/pkcs8.cc
kingstjo: Some of this complexity stemmed from users that would be giving a private key not already in PKCS8 format along with trying an empty password for incoming keys that did not have -passin but may be encrypted.

With #2419 I now need to refactor pkcs8.cc to stop blocking encrypted paths that should prompt for password

tool-openssl/pkcs8.cc
```diff
@@ -230,153 +141,160 @@ static const argument_t kArguments[] = {
...
+      !validate_prf(parsed_args["-v2prf"])) {
+    return false;
+  }
+
+  GetString(passin_arg.get(), "-passin", "", parsed_args);
+  GetString(passout_arg.get(), "-passout", "", parsed_args);
+
+  // Extract passwords if provided
+  if (!passin_arg->empty()) {
+    if (!pass_util::ExtractPassword(passin_arg)) {
```
nhatnghiho: I wonder if this line and line 216 should use `ExtractPasswords` instead? Since they may use the same file

nhatnghiho:
Not sure if I missed something but OpenSSL looks like they also support `stdin` and `fd`. Did we make a decision to not include those?

tool-openssl/pkcs8.cc
kingstjo: Yes, this is correct. 
As tool pkcs8 is using both `-passin` and `-passout` it should use `ExtractPasswords` instead of `ExtractPassword`.

Thank you for this call-out!

tool-openssl/pkcs8.cc
kingstjo: Additionally as this may come up in the future.
AWS-LC only uses `hmacWithSha1`for `v2prf `and function `validate_prf` throws error if anything else is used in pkcs8 CLI

smittals2:
Can we add some tests for stdin and fd options as well?

tool-openssl/pkcs8.cc
```diff
@@ -293,31 +187,32 @@ bool pkcs8Tool(const args_list_t& args) {
...
     }
-    
+
     in.reset(BIO_new_file(in_path.c_str(), "rb"));
     if (!in) {
         fprintf(stderr, "Cannot open input file\n");
         return false;
     }
-    if (!validate_bio_size(in.get())) {
+    if (!validate_bio_size(*in)) {
```
smittals2: Curious about the change to dereference the unique_ptr directly instead of calling .get() as is the general paradigm in our codebase. I see you also rewrote validate_bio_size to accept a BIO& instead of BIO*.

tool-openssl/pkcs8.cc
```diff
@@ -2,80 +2,73 @@
...
+static bool validate_format(const std::string &format) {
+  if (format != "PEM" && format != "DER") {
+    fprintf(stderr, "Format must be PEM or DER\n");
+    return false;
+  }
+  return true;
 }
 
 // Supported cipher algorithms for key encryption
 static const std::unordered_set<std::string> kSupportedCiphers = {
```
smittals2: How are you picking the supported values here?

tool-openssl/pass_util_test.cc
```diff
@@ -0,0 +1,530 @@
...
+  source.reset(new std::string("fd:-1"));
+  EXPECT_FALSE(pass_util::ExtractPassword(source));
+  
+  source.reset(new std::string("fd:invalid"));
+  EXPECT_FALSE(pass_util::ExtractPassword(source));
+}
+#endif
+
+#ifndef _WIN32
+TEST_F(PassUtilTest, StdinExtraction) {
```
smittals2, line 491: I know FD is not supported on windows, but why are we not testing stdin there?

smittals2 (approved):

tool-openssl/pkcs8.cc
kingstjo: On initial implementation we needed at minimum these ciphers.
Investigating `EVP_get_cipherbyname` and `kCiphersOIDs[]` to better match all ciphers supported in library.

tool-openssl/pass_util_test.cc
kingstjo: This is a good call-out, thank you!
Looking into adding this testing for Win environments

tool-openssl/pkcs8.cc
kingstjo: After digging in a bit more. 
My understanding is that OpenSSL simply attempts the encryption with the provided cipher and relies on the library itself to provide the error handling of a non-supported cipher for the operation.

We can also take this approach by verifying the cipher is at minimum contained by `EVP_get_cipherbyname` and then the specific operation in this case for PEM is `PEM_write_bio_PKCS8PrivateKey` will only work with supported ciphers.

tool-openssl/pkcs8.cc
kingstjo: This change was made due to static analysis alerts, but you are right and this will be better matching the surrounding codebase.

smittals2 (approved):
do we need to change/add any pkcs8 tests as a part of refactoring that tool? Otherwise looks good!

PR #2560, opened 2025-07-18
No-op impls for PHA functions

justsmth:
### Description of changes: 
* Provide no-op implementations for functions related to post-handshake authentication (PHA):
  * SSL_CTX_set_post_handshake_auth
  * SSL_set_post_handshake_auth
* Update `docs/porting/functionality-differences.md` to document behavior of PHA functions.

### Call-outs:
* When a project attempts to enable PHA (but AWS-LC ignores it), the connection proceeds securely without PHA (using standard TLS 1.3 authentication during the initial handshake). There is no downgrade or exposure to weaker security, as PHA is an optional enhancement for deferred client authentication. If the project strictly requires PHA, it will fail functionally (e.g., the server won't request post-handshake auth), but not in a way that compromises the core TLS session.
* I considered alternatives like aborting or failing the handshake if val == 1. I chose a no-op implementation as this doesn't introduce crashes or errors in unrelated code paths. It keeps the library usable for projects that call the function but don't critically depend on PHA working.

### DIFF
--- a/docs/porting/functionality-differences.md
+++ b/docs/porting/functionality-differences.md
@@ -240,6 +240,38 @@ libssl is the portion of OpenSSL which supports TLS. AWS-LC does not have suppor
   <p><span>Returns zero.</span></p>
   </td>
  </tr>
+ <tr>
+  <td rowspan=3>
+    <p><span>Post-handshake authentication</span></p>
+  </td>
+  <td rowspan=3>
+    <p><span>
+<a href="https://github.com/aws/aws-lc/blob/b6063413c8cfa2302f750a8cf90de550a598671d/include/openssl/ssl.h#L5981-L5995">Post-handshake authentication</a>
+</span></p>
+  </td>
+  <td>
+    <p><span>SSL_verify_client_post_handshake</span></p>
+  </td>
+  <td>
+    <p><span>Returns zero.</span></p>
+  </td>
+ </tr>
+ <tr>
+  <td>
+    <p><span>SSL_CTX_set_post_handshake_auth</span></p>
+  </td>
+  <td>
+    <p><span>Does nothing.</span></p>
+  </td>
+ </tr>
+ <tr>
+  <td>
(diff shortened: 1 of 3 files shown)


smittals2 (commented):
Just curious, is there a project that needs these functions / motivating reason to add these functions now? 

I read the callouts in the description, and while I agree that the security of the TLS connection is not necessarily weakened, I'm concerned that application level behavior may be compromised.

For example, an AWS-LC client calls SSL_CTX_set_post_handshake_auth to enable PHA. Since it's a no-op, no PHA extension will be sent in the client hello. The client application may have expectations around PHA like custom verification callbacks or other logic that it expects to use after the handshake. 

There could be 2 issues here: 
1) The non LC-server/app is configured to verify the LC-client with SSL_VERIFY_PEER (mTLS) & SSL_VERIFY_POST_HANDSHAKE. Since PHA extension is not sent, server will try to verify the client during the handshake. This could be problematic at the client application level if there are assumptions that verification only happens after handshake for a given use case. A client working correctly for PHA but not for normal TLS is unlikely, but we could be violating some additional checks/assumptions on the application layer? 
2) A non-LC server/application might require PHA for its connections as part of its security model. Some app/client builds with LC not knowing about PHA support, assumes all is kosher and that PHA is enabled with the no-op. But obviously no PHA ext is sent. The non-LC server may initiate PHA at any time after the handshake - but this is only possible if the client sent the PHA ext. So now the non-LC server is unable to do PHA with the LC-client (for this use case we assume not being able to do PHA with a client violates the server's security model). This may lead to unexpected behavior for the LC client. 

As a LC-server, to indicate support/enable PHA you need to set a flag SSL_VERIFY_POST_HANDSHAKE or call SSL_verify_client_post_handshake, so I think these changes aren't problematic on that end.

WillChilds-Klein (approved):

WillChilds-Klein (approved):

PR #2561, opened 2025-07-21
Impl `SSL_client_hello_get1_extensions_present` and friends

justsmth:
### Description of changes: 
Implementation and documentation for:
* `SSL_client_hello_get1_extensions_present`
* `SSL_client_hello_get_extension_order`
* `SSL_client_hello_get0_legacy_version`

### Testing:
Added new tests for coverage.

### DIFF
--- a/ssl/ssl_client_hello_test.cc
+++ b/ssl/ssl_client_hello_test.cc
@@ -367,7 +367,7 @@ int callback_SSL_client_hello_get1_extensions_present_impl(
   }
 
   EXPECT_GT(extensions_len, 0u);
-  EXPECT_NE(nullptr, extensions);
+  EXPECT_TRUE(extensions);
 
   unsigned legacy_version = SSL_client_hello_get0_legacy_version(ssl);
   EXPECT_EQ(legacy_version, (unsigned)TLS1_2_VERSION);
@@ -469,32 +469,37 @@ TEST(SSLClientHelloTest, GetExtensionOrder) {
         // First, call with a null buffer to get the count of extensions.
         if (SSL_client_hello_get_extension_order(ssl, nullptr,
                                                  &num_extensions) != 1) {
+          ADD_FAILURE()
+              << "Failed initial call to SSL_client_hello_get_extension_order";
           return SSL_CLIENT_HELLO_ERROR;
         }
         EXPECT_GT(num_extensions, 0u);
 
         // Allocate a buffer of the correct size and get the extensions.
-        uint16_t* exts =
-            (uint16_t *)OPENSSL_zalloc(sizeof(uint16_t) * num_extensions);
+        uint16_t *exts = static_cast<uint16_t *>(
+            OPENSSL_zalloc(sizeof(uint16_t) * num_extensions));
         if (exts == nullptr) {
+          ADD_FAILURE() << "Failed to allocate extensions";
           return SSL_CLIENT_HELLO_ERROR;
         }
-        if (SSL_client_hello_get_extension_order(
-                ssl, exts, &num_extensions) != 1) {
+        if (SSL_client_hello_get_extension_order(ssl, exts, &num_extensions) !=
+            1) {
+          ADD_FAILURE()
+              << "Failed call to SSL_client_hello_get_extension_order";
           OPENSSL_free(exts);
           return SSL_CLIENT_HELLO_ERROR;
         }
 
         unsigned legacy_version = SSL_client_hello_get0_legacy_version(ssl);
-        EXPECT_EQ(legacy_version, (unsigned)TLS1_2_VERSION);
+        EXPECT_EQ(legacy_version, static_cast<unsigned>(TLS1_2_VERSION));
 
         // Call with a buffer that is too small and confirm it fails.
         size_t too_small_num_extensions = num_extensions - 1;
-        uint16_t* too_small_exts =
-            (uint16_t *)OPENSSL_zalloc(sizeof(uint16_t) *
-                                       too_small_num_extensions);
+        uint16_t *too_small_exts = static_cast<uint16_t *>(
+            OPENSSL_zalloc(sizeof(uint16_t) * too_small_num_extensions));
         if (!too_small_exts) {
           OPENSSL_free(exts);
+          ADD_FAILURE() << "Failed to allocate too small buffer";
           return SSL_CLIENT_HELLO_ERROR;
         }
         // Expect failure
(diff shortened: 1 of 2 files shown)


ssl/ssl_lib.cc
```diff
@@ -3081,6 +3081,111 @@ int SSL_client_hello_get0_ext(SSL *s, unsigned int type, const unsigned char **o
...
+  *out = ext_types;
+  *outlen = num_extensions;
+
+  return 1;
+}
+
+int SSL_client_hello_get_extension_order(SSL *s, uint16_t *exts, size_t *num_exts) {
+  GUARD_PTR(s);
+  GUARD_PTR(s->s3);
+  SSL_HANDSHAKE *hs = s->s3->hs.get();
```
smittals2, line 3138: this pattern feels familiar so maybe I'm forgetting smthg, but should we be checking s->s3->hs before calling get() on hs? since knowing s3 is valid doesn't imply hs will be valid, and calling get() on a null hs could lead to segfault. 
Same q for SSL_client_hello_get0_legacy_version below

include/openssl/ssl.h
```diff
@@ -1124,6 +1124,34 @@ OPENSSL_EXPORT size_t SSL_get0_peer_delegation_algorithms(
...
+// to point to this array and |*outlen| to the number of extensions. The ints
+// in the array correspond to the type of each extension. The caller is
+// responsible for releasing the array with OPENSSL_free. If no extensions are
+// found, it sets |*out| to NULL and |*outlen| to 0. The function returns 1 on
+// success and returns 0 on error.
+//
+// This function can only be called from within a client hello callback (see
+// |SSL_CTX_set_client_hello_cb|) or during server certificate selection (see
+// |SSL_CTX_set_select_certificate_cb|).
+OPENSSL_EXPORT int SSL_client_hello_get1_extensions_present(SSL *s, int **out, size_t *outlen);
```
smittals2, line 1138: for all these funcs we say they can only be called from within a client hello or server certificate selection callback. Do we verify this in the functions themselves, or can we? Possibly by asserting some property/field on the SSL object is either initialized/uninitialized/or a specific value? Or is this more of a documentation thing. What if someone didn't shed the hs object at the end of the handshake and then calls this func, would that execution technically succeed?

ssl/ssl_lib.cc
```diff
@@ -3081,6 +3081,111 @@ int SSL_client_hello_get0_ext(SSL *s, unsigned int type, const unsigned char **o
...
+  GUARD_PTR(hs);
+
+  SSLMessage msg_unused;
+  SSL_CLIENT_HELLO client_hello;
+  if (!hs->GetClientHello(&msg_unused, &client_hello)) {
+    return 0;
+  }
+  return client_hello.version;
```
smittals2, line 3185: is legacy_version and version here the same?

smittals2: after reading looks like legacy_version is only applicable to TLS1.3 while older handshakes just had a version field. Not sure if we're testing this, but can we add a TLS 1.3 handshake and check if this correctly returns?

ssl/ssl_lib.cc
justsmth: `hs` is a [UniquePtr](https://github.com/aws/aws-lc/blob/8135987a46084fc1bc732095a6216203e6ceab59/ssl/internal.h#L3061) field on `s->s3` struct. Since `s->s3` is not NULL, this line shouldn't segfault. Then we check whether `hs` holds a NULL value on the line after this.

include/openssl/ssl.h
justsmth: > Do we verify this in the functions themselves, or can we? 

We don't verify. We could, but it's not worth the overhead. The [handshake object is destroyed](https://github.com/aws/aws-lc/blob/main/ssl/ssl_lib.cc#L905) when the handshake completes, so this is just warning consumers to not expect this to provide you with information after the handshake is complete.

ssl/ssl_lib.cc
justsmth: I updated the test of `SSL_client_hello_get0_legacy_version` to be specific about the expected return value.

ssl/ssl_lib.cc
```diff
@@ -3081,6 +3081,111 @@ int SSL_client_hello_get0_ext(SSL *s, unsigned int type, const unsigned char **o
...
+  SSL_CLIENT_HELLO client_hello;
+  if (!hs->GetClientHello(&msg_unused, &client_hello)) {
+    return 0;
+  }
+
+  CBS extensions;
+  CBS_init(&extensions, client_hello.extensions, client_hello.extensions_len);
+
+  size_t num_extensions = 0;
+  size_t i = 0;
```
samuel40791765: Seems like `num_extensions` and `i` can be consolidated into 1 variable? Not a blocker though and I could be missing something.

ssl/ssl_client_hello_test.cc
```diff
@@ -345,6 +348,175 @@ TEST(SSLClientHelloTest, ClientHelloKnownExtensions) {
...
+  int *extensions = nullptr;
+  size_t extensions_len = 0;
+  if (!SSL_client_hello_get1_extensions_present(ssl, &extensions,
+                                                &extensions_len)) {
+    ADD_FAILURE() << "SSL_client_hello_get1_extensions_present failed";
+    return SSL_CLIENT_HELLO_ERROR;
+  }
+
+  EXPECT_GT(extensions_len, 0u);
+  EXPECT_NE(nullptr, extensions);
```
samuel40791765:
```suggestion
  EXPECT_TRUE(extensions);
```

ssl/ssl_client_hello_test.cc
```diff
@@ -345,6 +348,175 @@ TEST(SSLClientHelloTest, ClientHelloKnownExtensions) {
...
+          OPENSSL_free(exts);
+          return SSL_CLIENT_HELLO_ERROR;
+        }
+        // Expect failure
+        if (SSL_client_hello_get_extension_order(
+                ssl, too_small_exts, &too_small_num_extensions) != 0) {
+          OPENSSL_free(exts);
+          OPENSSL_free(too_small_exts);
+          return SSL_CLIENT_HELLO_ERROR;
+        }
```
samuel40791765: Is there any particular reason why we're using if/else statements in the callback instead of the typical `ASSERT/EXPECT` macros? If it's due to the additional memory we have to free, maybe we could consider using smart pointers? 

No strong opinions here, just thought I'd ask since it seemed like there were two ways of testing going on.

justsmth: The ASSERT_xxx macros can't be used in the callback, but it can use the EXPECT_xxx and ADD_FAILURE macros. I added more checks to ensure the test will fail if any errors occur in the callback.

smittals2 (approved):

samuel40791765 (approved):

PR #2562, opened 2025-07-21
Update bind9 CI test to use meson

nhatnghiho:
### Issues:
Resolves P257046914

### Description of changes: 
Bind9 recently switched from automake to meson for its build system, leading to broken CI test on our end. This PR fixes that.

### Callouts:
I disabled jemalloc for bind9 build because I ran into memory allocator conflict with jemalloc. We didn't have this problem when bind9 was still using automake so I suspect it's probably due to how meson is set up in bind9. Disabling jemalloc shouldn't affect the test outcome.


tests/ci/integration/run_bind9_integration.sh
```diff
@@ -53,6 +59,7 @@ ls
 
 aws_lc_build ${SRC_ROOT} ${AWS_LC_BUILD_FOLDER} ${AWS_LC_INSTALL_FOLDER} -DBUILD_TESTING=OFF -DBUILD_TOOL=OFF -DBUILD_SHARED_LIBS=1
 export LD_LIBRARY_PATH="${AWS_LC_INSTALL_FOLDER}/lib"
+export LD_PRELOAD=libjemalloc.so.2
```
WillChilds-Klein: `LD_PRELOAD` represents a path. is there a jemalloc lib in `BIND9_SRC_FOLDER`? if so, why do we need it?

nhatnghiho: There's no jemalloc lib in `BIND9_SRC_FOLDER`. I added it because I ran into a segmentation fault, which came from a memory allocator conflict between bind9 and aws-lc. Adding `LD_PRELOAD` solves the issue by forcing both to use the same memory allocator at runtime. But to be completely honest I don't like this approach too much.

We didn't have this problem building with jemalloc when bind9 was still using automake so I suspect it's probably due to how meson is set up in bind9. 

Another option is to disable jemalloc altogether. Does that seem like a better way to go?

WillChilds-Klein: I see. This work-around effectively disables jemalloc by pointing `LD_PRELOAD` to a non-existent file. Doing that explicitly in the build configuration seems like a better way to go.

WillChilds-Klein (approved):

PR #2565, opened 2025-07-21
Add EVP_PKEY_check and EVP_PKEY_public_check

smittals2:
### Description of changes: 
This PR implements two key validation functions from OpenSSL - EVP_PKEY_check and EVP_PKEY_public_check - with support for EC and RSA keys. Rather than expanding our ameth function table (as OpenSSL does), we directly invoke the appropriate validation functions for each key type. This provides the same functionality with a cleaner implementation.

#### Key-Specific Handling
RSA Keys:
- For EVP_PKEY_check, we use RSA_check_key which validates both public and private components
- For EVP_PKEY_public_check, we leverage the same function as OpenSSL doesn't provide a distinct public-only verification

EC Keys:
- Both functions ultimately call EC_KEY_check_key, which primarily validates the public key
- For EVP_PKEY_check, we add verification that the private key component exists before calling EC_KEY_check_key (like OpenSSL)


include/openssl/evp.h
```diff
@@ -755,6 +755,23 @@ OPENSSL_EXPORT int EVP_PKEY_derive_set_peer(EVP_PKEY_CTX *ctx, EVP_PKEY *peer);
...
+// key component exists and calls EC_KEY_check_key. For RSA keys, it calls
+// RSA_check_key which validates both public and private key relationships.
+//
+// It returns one on success and zero on error.
+OPENSSL_EXPORT int EVP_PKEY_check(EVP_PKEY_CTX *ctx);
+
+// EVP_PKEY_public_check validates at least the public components of a key.
+// For EC keys, this calls EC_KEY_check_key without requiring a private key.
+// For RSA keys, this reuses RSA_check_key (which does require a private key)
+// as OpenSSL doesn't provide a distinct public-only verification function for
```
dkostic: OpenSSL?

smittals2: I'll write the comment a bit better. What I meant to say here is RSA keys have no public key only verification function (in OpenSSL and AWS-LC). So if you called EVP_PKEY_public_check with OpenSSL it would error saying the operation is not supported for RSA keys. We can do this too, but I just redirected this to our existing RSA_check_key function for cleaner behavior.

smittals2:
> I guess we are adding these for compatibility with OpenSSL. Can you please comment on if the behavior between AWS-LC and OpenSSL is the same for these two functions?

Yep almost the same. EVP_PKEY_check is meant to validate both public and private components of they key while EVP_PKEY_public_check only needs a public key component. The only difference in our implementation is instead of throwing an unsupported for this key type error in EVP_PKEY_public_check for RSA keys (what happens in OpenSSL), we call our existing RSA_check_key function (which validates both private and public components).

PR #2573, opened 2025-07-23
Add an AVX-512-optimized single-buffer MD5 implementation

pittma:
This PR adds a block order implementation optimized with AVX-512's `vpternlog` instruction for the mixing functions.

The table below shows the results of `bssl speed` for a **Intel(R) Xeon(R) 6787P** (Granite Rapids) CPU using the command:

```sh
taskset -c 0 ./tool/bssl speed -filter MD5
```

| Bytes | MB/s (`main`) | MB/s (`md5`) | % diff |
| ----- | ------------- | ------------ | ------ |
| 16    | 103.8         | 106.5        | 2.6%     |
| 256   | 566.6         | 611.5        | 7.9%     |
| 1350  | 770.3         | 856.6        | 11.2%    |
| 8192  | 822.6         | 925.7        | 12.5%    |
| 16384 | 830.8         | 933.9        | 12.4%    |

---


pittma:
Hi all, I tried several versions of this but this is the most straightforward. If there is any interest in optimizing a single-buffer MD5 implementation, I thought this was the best place to start. If we'd like to see what else we can squeeze out in terms of performance with the trade-off of a slightly less clean API or more complex implementation, I'm happy to keep churning on it.

crypto/fipsmodule/md5/asm/md5-avx512.pl
```diff
@@ -0,0 +1,306 @@
...
+  .L_main_loop:
+___
+
+  one_round($data);
+
+  $code .= <<___;
+  add \$64, $data
+  sub \$1, $num
+  cmp \$0, $num
+  jg .L_main_loop
```
tkanteck: it's very minor but these two lines could be replaced with `jnz .L_main_loop` only

crypto/fipsmodule/md5/asm/md5-avx512.pl
```diff
@@ -0,0 +1,306 @@
...
+  md5_x86_64_avx512:
+  .cfi_startproc
+  endbranch
+___
+  if ($win64) {
+    $code .= <<___;
+    push  %rbp
+    mov %rsp,%rbp
+    sub $XMM_STORAGE, %rsp
+    and	\$0xffffffffffffffc0,%rsp
```
tkanteck: I believe this aligns the stack frame to 64 bytes and probably only 16 byte alignment is needed here

dkostic:
Hi Dan, thanks for your contribution! We are discussing internally, trying to figure out if we have any customer demand for faster single buffer MD5. We'll get back to you.

PR #2574, opened 2025-07-24
ABI monitoring GitHub workflow improvements

skmcgrail:
### Description of changes: 
Improve the ABI diff monitoring workflow to also compute whether the current pull request or push event to a branch breaks (or has broken) ABI compatibility when compared to the latest tagged release.

Previously the workflow jobs (now called incremental diffs) compared the current pull request or push event to the target branch / previous commit. This is problematic in the push event scenario when trying to track whether the current main branch has broken ABI comparability with the previous tagged release. For example:

`A -> B -> C -> C_1 -> C_2` where each is a commit on the main branch.

Assume `B` is the latest tagged release commit. Commit C does not break ABI compatibility, `C_1` is then merged from a pull request commit that breaks ABI compatibility (shows on the branch checks). If `C_2` makes changes that do not break ABI compatibility from the changes in `C_1` then the checks on the branch would pass.  The additions to this workflow will also ensure that all commits following release tag `B` will also diff themselves to `B` and report what that result is in addition the current incremental diffs.

### DIFF
--- a/.github/actions/latest-semver-tag/README.md
+++ b/.github/actions/latest-semver-tag/README.md
@@ -50,12 +50,12 @@ This action:
 1. Validates the provided `release` input parameter (must be one of: `main`, `fips1`, `fips2`, `fips3`)
 2. Validates that the specified path exists and is a git repository if provided
 3. Fetches all tags from the repository
-3. Based on the release type:
+4. Based on the release type:
    - For `main`: Filters tags to only include those that follow the format `vX.Y.Z`
    - For `fipsN`: Filters tags to only include those that follow the format `AWS-LC-FIPS-N.Y.Z` where N matches the FIPS version
-4. Sorts the tags according to semantic versioning rules
-5. Determines the latest tag (fails if no matching tag is found)
-6. Outputs the latest tag and version (without the prefix)
+5. Sorts the tags according to semantic versioning rules
+6. Determines the latest tag (fails if no matching tag is found)
+7. Outputs the latest tag and version (without the prefix)
 
 ## Requirements
 
--- a/.github/actions/latest-semver-tag/README.md
+++ b/.github/actions/latest-semver-tag/README.md
@@ -1,0 +1,65 @@
+# Latest Semantic Version Tag Action
+
+A GitHub Action that determines the latest tagged git release following semantic versioning and outputs that value for subsequent steps to use.
+
+## Inputs
+
+| Input | Description | Required | Default | Allowed Values |
+|-------|-------------|----------|---------|----------------|
+| `release` | The release type to filter tags by | Yes | `main` | `main`, `fips1`, `fips2`, `fips3` |
+| `path` | The path containing the git repository to check | No | `.` | Any valid directory path |
+
+## Outputs
+
+| Output | Description |
+|--------|-------------|
+| `latest-tag` | The latest semantic version tag (format depends on release type) |
+| `latest-version` | The latest semantic version without the prefix (format depends on release type) |
+
+### Tag Formats by Release Type
+
+- For `main` release: Tags in `vX.Y.Z` format (e.g., `v1.2.3`)
+- For `fipsN` releases: Tags in `AWS-LC-FIPS-N.Y.Z` format (e.g., `AWS-LC-FIPS-1.2.3` for `fips1`)
+
+## Usage
+
+```yaml
+jobs:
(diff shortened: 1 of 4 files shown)


.github/actions/latest-semver-tag/README.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+## How It Works
+
+This action:
+
+1. Validates the provided `release` input parameter (must be one of: `main`, `fips1`, `fips2`, `fips3`)
+2. Validates that the specified path exists and is a git repository if provided
+3. Fetches all tags from the repository
+3. Based on the release type:
```
torben-hansen: Should be (4) here and increment all numbered bullets below

.github/actions/latest-semver-tag/get-latest-tag.sh
```diff
@@ -0,0 +1,65 @@
...
+echo "Using git repository in directory: $REPO_PATH"
+
+# Ensure we have all tags
+git -C "$REPO_PATH" fetch --tags
+
+# Different tag format based on release type
+if [[ "$RELEASE" == "main" ]]; then
+  # For main release, use vX.Y.Z format
+  echo "Looking for tags in vX.Y.Z format..."
+  LATEST_TAG=$(git -C "$REPO_PATH" tag -l 'v[0-9]*.[0-9]*.[0-9]*' | sort -V | tail -n 1)
```
torben-hansen, line 37:
```suggestion
  LATEST_TAG=$(git -C "$REPO_PATH" tag -l 'v[0-9]+\.[0-9]+\.[0-9]+' | sort -V | tail -n 1)
```
Only match numbers and periods? Not sure we would allow anything else than a numerical.

.github/actions/latest-semver-tag/get-latest-tag.sh
```diff
@@ -0,0 +1,65 @@
...
+  
+  # Extract version without the 'v' prefix
+  LATEST_VERSION=${LATEST_TAG#v}
+else
```
torben-hansen, line 41: Matching algorithm generally assumes the tag is of a specific format and finds the latest tag with that format. Doesn't have context to derive whether a newer release has a tag that is not the correct format.

But I don't see a way to make this script smarter and error if a tag has been miss-formatted.
Should probably be another process.

What do you think?

.github/actions/latest-semver-tag/get-latest-tag.sh
skmcgrail: As discussed offline this is unfortunately shell globing syntax not regex.

.github/actions/latest-semver-tag/get-latest-tag.sh
skmcgrail: I think that would be the responsibility for another process to make sure we are tagging appropriately (incrementing in a forward direction in time).  Justin already added a workflow that does already validates tags are targeting the proper branches https://github.com/aws/aws-lc/tree/main/util/git-tag-check following the format, and would just need to be expanded to validate the forward movement. I'm going to say its unnecessary right now for this PR, as this "get latest tag" concept hasn't been an issue for aws-lc-rs which follows similar logic.

torben-hansen: makes sense!

torben-hansen (approved):

.github/actions/latest-semver-tag/action.yml
```diff
@@ -0,0 +1,28 @@
...
+inputs:
+  release:
+    description: 'The release type to filter tags by'
+    required: true
+    default: 'main'
+    # Allowed values: main, fips1, fips2, fips3
+  path:
+    description: 'The path containing the git repository to check'
+    required: false
+    default: '.'
```
samuel40791765, line 12: Are there any examples of circumstances where this would come in handy? It seems like we'll be running the checks against this repo's root directory in most cases.

samuel40791765 (approved):

.github/actions/latest-semver-tag/action.yml
skmcgrail: This path is mainly for the instance where an alternative checkout path is provided to `actions/checkout`, which is the case for the ABI workflows since the repo is checked out twice (previous & next). But normally it would just default to current path `.` since thats also what `actions/checkout` does.

PR #2576, opened 2025-07-25
Implement SSL_set_verify_result

nhatnghiho:
### Issues:
Resolves #CryptoAlg-3061

### Description of changes: 
(A continuation of #2526)
Implement SSL_set_verify_result, which requires adding the field verify_result to the SSL struct. According to OpenSSL's implementation, the value of verify_result in the SSL struct and SSL_SESSION are always synced during the handshake. I mirrored closely that behavior. See document in CryptoAlg-3061's comment section for line-by-line functional parity comparison.

### Callouts:
Additions since last PR:
- Make sure SSL struct's `verify_result` is initialized to `X509_V_ERR_INVALID_CALL`, which surfaced several places where we need to sync the value of `verify_result` in SSL struct and SSL_SESSION. I have patched all the spots that I know of.
- Updated the SSL de/serializer with a new optional field `verify_result`. No version bump is needed for now.

### Testing:
Unit tests added.


ssl/ssl_transfer_asn1.cc
```diff
@@ -963,6 +963,8 @@ static const unsigned kSSLQuietShutdownTag =
     CBS_ASN1_CONSTRUCTED | CBS_ASN1_CONTEXT_SPECIFIC | 0;
 static const unsigned kSSLConfigTag =
     CBS_ASN1_CONSTRUCTED | CBS_ASN1_CONTEXT_SPECIFIC | 1;
+static const unsigned kSSLVerifyResultTag =
+    CBS_ASN1_CONSTRUCTED | CBS_ASN1_CONTEXT_SPECIFIC | 3;
```
skmcgrail: Is there a reason why this skips over `2`?

ssl/ssl_transfer_asn1.cc
```diff
@@ -1066,11 +1075,18 @@ static int SSL_parse(SSL *ssl, CBS *cbs, SSL_CTX *ctx) {
     ssl->config.reset();
   }
 
+  if (!CBS_get_optional_asn1_int64(&ssl_cbs, &verify_result, kSSLVerifyResultTag, X509_V_ERR_INVALID_CALL)) {
+      OPENSSL_PUT_ERROR(SSL, SSL_R_SERIALIZATION_INVALID_SSL);
+    return 0;
+  }
```
skmcgrail: So a TLS connection can't be serialized until after the handshake is completed. So is there ever a case where the certificate result would have been anything but `X509_V_OK`? So if the handshake is complete (TLS serialization feature is only supported for the server side, so I assume this `verify_result` is the result of validating the client certificate if the connection was using mTLS), then the result would always have to be X509_V_OK right? Or am I mistaken here? AWS-LC doesn't support post-handshake authentication, which means the CertificateRequest and CertificateVerify messages would have been sent as part of the handshake process......so how could the result be something that was not OK? This is making me question whether we actually even need to serialize this data?

If it's indeed needed, then I kind of question the default status used here.

ssl/ssl_transfer_asn1.cc
nhatnghiho: No it should be 2. My bad.

ssl/ssl_transfer_asn1.cc
nhatnghiho: > So is there ever a case where the certificate result would have been anything but X509_V_OK?

Short answer -- Since we're adding `SSL_set_verify_result` (which is part of the public API), I suppose anyone can change the value of `verify_result` to whatever they want. Therefore, we should serialize it just as a precaution. 

Long answer -- I haven't seen cases where `SSL_set_verify_result` is used to set the result to anything other than `X509_V_OK`. However, even without that, a completed handshake may carry something other than `X509_V_OK` if the certification verification mode is none (`SSL_VERIFY_NONE`), since the verification error will be considered non-fatal. And a `SSL_VERIFY_NONE` mode is always possible since we also have `SSL{_CTX}_set_custom_verify`.

> If it's indeed needed, then I kind of question the default status used here.

Regarding the default status, I suppose using `X509_V_ERR_INVALID_CALL` can cause errors down the line if say the serialized SSL is of an older version, which lacks a serialized `verify_result` field. How about using the value of `verify_result` in the established session as a default fallback?

skmcgrail: Out of curiosity what is the use case for using `SSL_set_verify_result`? I see OpenSSL provides it as an ability to override the result that is stored on a session, but does that actually impact an on-going handshake or anything? Is that just metadata? What integration were we working on where this is required? What is their use case?

justsmth: > What integration were we working on where this is required?

I know it's [needed to support Apache httpd](https://github.com/apache/httpd/blob/0c0a5a323b1b250f821e87ac6bc3953742fa4e2b/modules/ssl/mod_ssl.c#L649). I also see it used in [Postfix](https://github.com/vdukhovni/postfix/blob/9756d67d1aebe23fae4cdfa4090de950fff5e843/postfix/src/tls/tls_dane.c#L1084).

justsmth: > I haven't seen cases where SSL_set_verify_result is used to set the result to anything other than X509_V_OK.

[Here's an example](https://github.com/apache/httpd/blob/trunk/modules/ssl/ssl_engine_kernel.c#L875) of that from httpd. Also, [from postfix](https://github.com/vdukhovni/postfix/blob/9756d67d1aebe23fae4cdfa4090de950fff5e843/postfix/src/tls/tls_dane.c#L1084).

justsmth: My stance on the discussion above is that we should align closely with the behavior of OpenSSL, unless there is a good (i.e., security-related) reason to do otherwise.

As another data point for serialization, Openssl serializes the "verify_result" field as part of its [SSL_SESSION_ASN1](https://github.com/openssl/openssl/blob/defe51c178e3dc9f07514c179121021fd78691b4/ssl/ssl_asn1.c#L59).

justsmth (approved):

skmcgrail (approved):

PR #2582, opened 2025-07-28
adding hmac in inject_hash.cpp

prikhap:
### Description
This pull request implements a C++ version of inject_hash for testing the integration of HMAC-SHA256 hash calculation and LIEF integration for handling shared libraries on Linux (.so) and MacOS (.dylib) platforms.
Here for testing we simply just calculate hash of the path provided and if the path provided has the extension .so or .dylib then we also parse it using the LIEF parser.

### Key Implementation Changes

Added HMAC-SHA256 hash calculation for file paths
Integrated LIEF library for parsing shared library files

### Files Modified

.github/workflows/test-cpp-inject-hash.yml: Updated workflow configuration
crypto/CMakeLists.txt: Added fips_hashing subdirectory
crypto/fips_hashing/CMakeLists.txt: New file for FIPS hashing configuration
crypto/fips_hashing/fips_hashing.c: Added core hashing functionality
crypto/fipsmodule/digest/digest.c: Added HMAC support functions
crypto/fipsmodule/evp/p_hmac.c: Modified HMAC functionality
tests/ci/run_inject_hash_cpp.sh: Updated test script with new test cases
util/fipstools/inject_hash_cpp/CMakeLists.txt: Added LIEF dependency
util/fipstools/inject_hash_cpp/inject_hash.cpp: Main implementation file

### Testing

Added new test cases in run_inject_hash_cpp.sh
Tests for invalid files and edge cases
HMAC calculation verification
Shared library parsing tests


.github/workflows/test-cpp-inject-hash.yml
```diff
@@ -1,8 +1,8 @@
-name: Test C++ inject_hash Implementation
+name: Test C++ inject_hash Implementation with LIEF, HMAC and SHA-256 integrated.
 
 on:
   push:
-    branches: [ implement-inject-hash-clean ]
+    branches: [ hmac-sha-lief-integration ]
```
samuel40791765: I think we can just remove this entire line. `pull_request` below already handles the event.

crypto/fipsmodule/evp/p_hmac.c
```diff
@@ -183,9 +183,9 @@ DEFINE_METHOD_FUNCTION(EVP_PKEY_METHOD, EVP_PKEY_hmac_pkey_meth) {
   out->ctrl_str = hmac_ctrl_str;
 }
 
-int used_for_hmac(EVP_MD_CTX *ctx) {
-  return ctx->flags == EVP_MD_CTX_HMAC && ctx->pctx != NULL;
-}
+// int used_for_hmac(EVP_MD_CTX *ctx) {
+//   return ctx->flags == EVP_MD_CTX_HMAC && ctx->pctx != NULL;
+// }
```
samuel40791765: Just remove this rather than commenting it out.

crypto/fips_hashing/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+
+    fips_hashing.c
+
+    ../mem.c
+    ../thread_none.c
+    ../thread_pthread.c
+
+    ../err/err.c
+    $<TARGET_OBJECTS:generated_err_data>
+    ../decrepit/ripemd/ripemd.c
```
samuel40791765, line 18: 1. Is this needed?
2. Can we try moving all the files from `crypto/fips_hashing/fips_hashing.c` to  `CMakeLists.txt` here? There's the potential that we can remove `fips_hashing.c` altogether and point the source files here instead.

util/fipstools/inject_hash_cpp/inject_hash.cpp
samuel40791765, line 1: I think and trust that the HMAC functionality in this works. Could we still add a brief test within your test script (`run_inject_hash_cpp.sh`) to verify? Any input file would do.

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -53,18 +53,43 @@ int main(int argc, char** argv) {
...
+    std::vector<uint8_t> calculate_hash(HMAC_size(&ctx));
+    
+    unsigned int calculate_hash_len;
+    if (!HMAC_Final(&ctx, calculate_hash.data(), &calculate_hash_len)) {
+        std::cerr << "HMAC_Final failed" << std::endl;
         return 1;
     }
+    std::cout << "HMAC hash calculated successfully, the hash is: " << std::endl;     
+    for (unsigned int i = 0; i < calculate_hash_len; i++) {
+        printf("%02x", calculate_hash[i]);
```
samuel40791765: Can we align the printing method here with the rest of the file?

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -53,18 +53,43 @@ int main(int argc, char** argv) {
     if (is_apple) {
         std::cout << "Platform: macOS" << std::endl;
     }
-
-    // Parse binary
-    std::cout << "\nParsing binary..." << std::endl;
+    std::cout << "=== C++ inject_hash started ===" << std::endl;
     if (auto binary = LIEF::Parser::parse(input_file)) {
```
samuel40791765: `binary` doesn't seem to be used here.

```suggestion
    if (LIEF::Parser::parse(input_file)) {
```

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -53,18 +53,43 @@ int main(int argc, char** argv) {
...
+    unsigned int calculate_hash_len;
+    if (!HMAC_Final(&ctx, calculate_hash.data(), &calculate_hash_len)) {
+        std::cerr << "HMAC_Final failed" << std::endl;
         return 1;
     }
+    std::cout << "HMAC hash calculated successfully, the hash is: " << std::endl;     
+    for (unsigned int i = 0; i < calculate_hash_len; i++) {
+        printf("%02x", calculate_hash[i]);
+    } 
+    std::cout << std::endl;
```
samuel40791765: We should call `HMAC_CTX_cleanse` at the end as best practice.

crypto/fipsmodule/evp/p_hmac.c
prikhap: I let the comment be to indicate migration of code. However I have modified it now.

crypto/fips_hashing/CMakeLists.txt
prikhap: for "Can we try moving all the files from crypto/fips_hashing/fips_hashing.c to CMakeLists.txt here? There's the potential that we can remove fips_hashing.c altogether and point the source files here instead." : I tried doing this, by creating a static fips_hashing with all the needed scripts in the CMake, but it ends up giving huge compilation issues. 
Reason for this is that when we `#include` scripts into fips_hashing.c then all the scripts are laid out together and compiled together, while add_library in the CMake file would compile each of the scripts individually. Given the convoluted nature of the scripts there are many inter-dependencies between each of the files, and compiling them individually would obviously cause errors. Even if the flow of each of the scripts are intricately studied and they are laid out in order there are different memory allocation issues too. In short, when I tried creating it directly through CMake it needed require extensive refactoring of internal functions

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: done

prikhap: done

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: removed.

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: here since we are formatting on the hex ("%02x") , std::cout command becomes more verbose and moreover requires #include <iomanip> for setfill and setw which is an additional dependency.

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: added.

crypto/fips_hashing/CMakeLists.txt
samuel40791765: We should add a comment in `fips_hashing.c` about this and why we went this workaround instead.

util/fipstools/inject_hash_cpp/inject_hash.cpp
samuel40791765: I'm not seeing the call to `HMAC_CTX_cleanse`.

crypto/fips_hashing/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+    STATIC
+
+    fips_hashing.c
+
+    ../mem.c
+    ../thread_none.c
+    ../thread_pthread.c
+
+    ../err/err.c
+    $<TARGET_OBJECTS:generated_err_data>
```
samuel40791765: Is it possible to directly reference `${GENERATE_CODE_ROOT}/err_data.c` here and get rid of `$<TARGET_OBJECTS:generated_err_data>` (and the additional `if(FIPS)` logic in `crypto/CMakeLists.txt`) altogether?

prikhap: This comes down to the nature of CMake. If I give a direct reference to using `${GENERATE_CODE_ROOT}/err_data.c`, it will start looking for err_data.c which does not exist when the CMake runs (during the configuration), but it does not exist then. err_data.c is only created during the build time.
Hence `$<TARGET_OBJECTS:generated_err_data>` Records that generated_err_data will provide objects and Doesn't start look for files just yet. In short direct dependency would look for the file right away but target reference just records a dependency.

samuel40791765: I'm not sure if that's true or if it's due to the nature of CMake.

Could you try doing something similar to what `crypto/CMakeLists.txt` is doing here: https://github.com/aws/aws-lc/blob/d9003392b72c22d96039a7d54c774bc32a6c9c55/crypto/CMakeLists.txt#L324-L325

And for the record, `err_data.c` isn't created at build time. We have a copy of `err_data.c` here: https://github.com/aws/aws-lc/blob/main/generated-src/err_data.c. The `${GENERATE_CODE_ROOT}` points to the path of the directory. 

Feel free to experiment a bit more, so we can try removing the extra `if(FIPS)` logic in `crypto/CMakeLists.txt`. It's perfectly fine if we still need to go with this workaround. But I'd like us to try a bit more instead of adding new CMake logic that doesn't align with the rest of our build.

prikhap: I missed seeing the file already existed. They were creating another err_data.c if it failed to upload. I have accordingly made the changes

crypto/fips_hashing/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+if(FIPS AND (UNIX AND NOT ANDROID)) 
+  add_definitions(-DOPENSSL_NO_ASM=1)
+  remove_definitions(-DBORINGSSL_FIPS -DFIPS_ENTROPY_SOURCE_JITTER_CPU -DFIPS_ENTROPY_SOURCE_PASSIVE)
```
andrewhop, line 3: Does this affect all other targets?

crypto/fips_hashing/fips_hashing.c
```diff
@@ -0,0 +1,19 @@
+// The scripts need to be included in fips_hashing.c rather than compiled separately through CMake because it maintains a 
+// shared context for all files and preserves critical dependencies between them. Compiling files individually would cause 
+// memory allocation issues and symbol resolution problems due to their interconnected nature. Even if the flow of each 
+// script were intricately studied and laid out in order in CMake, it would require extensive refactoring of internal
+//  functions to resolve these dependencies.
```
andrewhop:
```suggestion
// The C files need to be included in fips_hashing.c rather than compiled separately through CMake because it maintains a 
// shared context for all files and preserves dependencies between them like bcm.c.
```

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -76,6 +79,13 @@ else
     ((ERRORS+=$?))
 fi
 
+#Test 4: HMAC calvulation of a test file path (should succeed)
```
andrewhop:
```suggestion
#Test 4: HMAC calculation of a test file path (should succeed)
```

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -76,6 +79,13 @@ else
...
+run_test "HMAC calculation test" false \
+    ./util/fipstools/inject_hash_cpp/inject_hash_cpp \
+    -in-object "util/fipstools/inject_hash_cpp/CMakeLists.txt" \
```
andrewhop, line 85: What is this test doing? Also how is the test passing, you check for .so and .dylib extensions.

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -53,19 +56,49 @@ int main(int argc, char** argv) {
...
-    if (auto binary = LIEF::Parser::parse(input_file)) {
-        std::cout << "✅ LIEF parser successfully loaded: " << input_file << std::endl;
-        
-        // TODO: Add hash calculation and injection
-        
-    } else {
-        std::cerr << "❌ LIEF parser failed to load: " << input_file << std::endl;
+    if (input_file.substr(input_file.length() - 3) == ".so" || input_file.substr(input_file.length() - 6) == ".dylib")
+        {if (LIEF::Parser::parse(input_file)) {
+            std::cout << "LIEF parser successfully loaded: " << input_file << std::endl;
```
andrewhop: I prefer using C++ standard functions like `ends_with`, what about Windows?
```suggestion
    if (input_file.substr(input_file.length() - 3) == ".so" || input_file.substr(input_file.length() - 6) == ".dylib") {
        if (LIEF::Parser::parse(input_file)) {
            std::cout << "LIEF parser successfully loaded: " << input_file << std::endl;
```

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -53,19 +56,49 @@ int main(int argc, char** argv) {
...
+        } 
+        else {
+            std::cerr << "LIEF parser failed to load: " << input_file << std::endl;
+            return 1;
+                    }
```
andrewhop: More funky indentation. Not in this PR, but a follow up please run clang format on all of these files and review just the formatting changes.

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -53,19 +56,49 @@ int main(int argc, char** argv) {
...
         return 1;
     }
 
+    const std::string& input_str = input_file;
+    if (!HMAC_Update(&ctx, 
+                     reinterpret_cast<const uint8_t*>(input_str.data()), 
+                     input_str.length())) {
+        std::cerr << "HMAC_Update failed" << std::endl;
+        return 1;
+    }
```
andrewhop: I'm not following what this is doing or why it's mac'ing the file name string. I think this is just for testing, please update the PR title and description to explain what this overall change is doing.

prikhap: inject_hash.cpp for testing purpose calculates hash of the file path (as a string) using HMAC-SHA256. 
Later on if the file is .so or .dylib, it will use LIEF to parse the file (because we are currently dealing with shared builds for linux and MacOS) 
I have added these as a comment to inject_hash.cpp too.

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: inject_hash.cpp is only for linux and macos. And since inject_hash.cpp is currently only a testing script to see if all the dependencies are well integrated. Hence just trying parsing on file paths if they are shared builds for linux and macos.

tests/ci/run_inject_hash_cpp.sh
prikhap: this test comes from https://github.com/aws/aws-lc/pull/2582#discussion_r2240523219, hence a random existing path. And the parsing if/else condition was added just to see if hmac and lief are actually integrated together. Since this is not a path to the binary, it is not getting parsed. 
It was just to test that hmac can indeed parse any string (in this case any file path)

tests/ci/run_inject_hash_cpp.sh
prikhap: modified.

crypto/fips_hashing/fips_hashing.c
prikhap: description modified.

crypto/fips_hashing/CMakeLists.txt
prikhap: It comes from https://github.com/aws/aws-lc/pull/2582#issuecomment-3133473831. 
CodeBuild BuildBatch us-west-2 (aws-lc-ci-windows-x86) was failing since the files included in fips-hashing were referencing symbols in some other files, which were resolved for macOS and linux, but some conditional inconsistency with Windows.
Moreover, we would be going with the capture_hash route for windows, hence fips_hashing is not required for Windows, hence made sure that fips_hashing is used for 'UNIX AND NOT ANDROID'

samuel40791765 (approved):

crypto/fips_hashing/CMakeLists.txt
andrewhop: What are the side affects of calling `remove_definitions(-DBORINGSSL_FIPS ...)`? What is and is not impacted by these compiler options?

util/fipstools/inject_hash_cpp/inject_hash.cpp
andrewhop: This would be useful information to include in the PR description

crypto/fips_hashing/CMakeLists.txt
prikhap: this flag is turned of to reduce the scripts that hmac and sha depend on. 
I am pretty sure that even after this everything is up and working in the FIPS mode itself since integrity_check is working.

andrewhop (approved):

PR #2583, opened 2025-07-29
Migrate MSVC tests to CodeBuild

nhatnghiho:
### Issues:
Resolves #CryptoAlg-3113 and #P274195896

### Description of changes: 
Migrate the following MSVC tests from GHA to CodeBuild:
- MSVC-2019
- MSVC-2022
- MSVC-SDE-64-bit

### Callouts:
The tests were running on Windows 2019 in GHA, but will be upgraded to run in Windows 2022. `MSVC-SDE-64-bit` is the only test to be affected, since some old CPUs don't support AVX instructions that Windows 2022 may use. I updated `all_tests.go` to exclude those CPUs if Windows 2022 or newer is detected.


util/all_tests.go
```diff
@@ -95,8 +85,43 @@ var sdeCPUs = []string{
 	"cpx", // Cooper Lake
 	"icx", // Ice Lake server
 	"tgl", // Tiger Lake
+    "snb", // Sandy Bridge
```
m271828: Indentation is inconsistent in this file.

nhatnghiho: Thanks for pointing it out! It's indented just fine in my IDE so I didn't notice it's inconsistent on Github

util/all_tests.go
```diff
@@ -95,6 +85,40 @@ var sdeCPUs = []string{
...
+		cmd := exec.Command("cmd", "/C", "ver")
+		output, err := cmd.Output()
+		if err != nil {
+			return
+		}
+
+		verOutput := strings.ToLower(string(output))
+
+		// Windows Server 2022 (or 10.0.20348) may natively use some AVX instructions that old CPUs do not have
+		if !strings.Contains(verOutput, "10.0.20348") {
```
andrewhop: There is an interesting choice here: this is checking for an exact version. My gut says all future versions will continue to use new instructions that aren't supported. Did you consider making this a comparison such that when Windows Server 2022 updates to say 10.0.20349 this filter still works?

util/all_tests.go
```diff
@@ -95,6 +85,40 @@ var sdeCPUs = []string{
 	"cpx", // Cooper Lake
 	"icx", // Ice Lake server
 	"tgl", // Tiger Lake
+	"snb", // Sandy Bridge
+	"ivb", // Ivy Bridge
+}
+
+// CPUs that are NOT compatible with MSVC
+var cpusWithoutAVX = []string{
```
andrewhop: Comment and name seem out of date. The root cause is really modern MSVC assumes a minimum level of CPU support.

util/all_tests.go
nhatnghiho: You're right. I updated the code to check for all versions older than `10.0.20*`. Windows Server 2019 is `10.0.17763` so this check should work

util/all_tests.go
nhatnghiho: Updated! Let me know if the new comment/name is clearer now

Issue #2459, opened 2025-06-03
symbol attribution in stack traces


jmayclin:
### Problem:
AWS-LC assembly methods like `__bn_sqrt8x_mont` are poorly attributed in flamegraphs. This makes it difficult to figure out what the calls are actually coming from.

### Solution:

I would expect to see this methods attributes to the higher level libcrypo APIs (sign/verify/etc).

andrewhop:
Did some digging. This might only impact the assembly functions on Arm Linux hosts:
|Algorithm|Impacted on Arm Mac|Impacted on x86 Linux|Impacted on Arm Linux|
|---|---|---|---|
|RSA|No|No|Yes|
|AES GCM|No|No|Yes|
|ChaCha|No|No|Yes|

```
cmake -GNinja -DCMAKE_BUILD_TYPE=RelWithDebInfo ../
ninja
flamegraph -- tool/bssl speed -filter RSA
```

Mac info:
* OS: Sequoia 15.5
* Compiler: AppleClang 16.0.0.16000026
* CPU M1 Pro

Arm Linux info:
* OS: AL 2023
* Compiler: GCC 11.5.0
* CPU: Graviton 3
* perf version: 6.1.144-170.251.amzn2023.x86_64

x86 Linux info:
* OS: AL 2023
* Compiler: GCC 11.5.0
* CPU Intel(R) Xeon(R) Platinum 8488C
* perf version: 6.1.144-170.251.amzn2023.aarch64

## RSA M1 Mac example
<img width="1487" height="380" alt="Image" src="https://github.com/user-attachments/assets/c434b740-2b95-4cca-9f42-f947e9c5484f" />

## RSA x86 Linux

<img width="1202" height="563" alt="Image" src="https://github.com/user-attachments/assets/17eb4009-3e85-4fd2-a679-a06b5083503c" />

## RSA Arm Linux

<img width="960" height="454" alt="Image" src="https://github.com/user-attachments/assets/219e395f-4e2a-4307-bccc-9cc1a8d25ae0" />

andrewhop:
Prototype fix for RSA is on my fork https://github.com/andrewhop/aws-lc/commit/b11b6a031e92e3e14e81d236c4eb6131416f16fb. What is weird is `bn_div_consttime` still shows up in the wrong spot, but that is an entirely C function only called from C code.

## RSA Arm Linux updated
[...]

PR #2584, opened 2025-07-29
Add CFI directives to armv8-mont

andrewhop:
### Issues:
Addresses https://github.com/aws/aws-lc/issues/2459

### Description of changes: 
Some of our assembly functions are missing Call Frame Information (CFI) directives. These are used to generate DWARF (Debug With Arbitrary Record Formats) entries that are used for stack unwinding. This is also important for profilers that rely on stack unwinding to identify code host spots.

### Call-outs:
This change does not introduce any new CPU instructions or impact the performance at all. These CFI directives are assembler directives that generate metadata that explains how to unwind the stack. 

### Testing:
I am still working on adding/updating the libunwind tests to check this. Looking at the flamegraphs things are better but not perfect yet. 

#### Before
image

#### After
image


crypto/fipsmodule/bn/asm/armv8-mont.pl
```diff
@@ -76,17 +76,28 @@
...
 	AARCH64_SIGN_LINK_REGISTER
 	tst	$num,#7
 	b.eq	__bn_sqr8x_mont
 	tst	$num,#3
 	b.eq	__bn_mul4x_mont
 .Lmul_mont:
 	stp	x29,x30,[sp,#-64]!
+.cfi_def_cfa_offset 64
+.cfi_offset x30, -56
+.cfi_offset x29, -64
```
dkostic: nit: for all other pairs the order is opposite
```suggestion
.cfi_offset x29, -64
.cfi_offset x30, -56
```

crypto/fipsmodule/bn/asm/armv8-mont.pl
```diff
@@ -292,18 +313,33 @@
...
+.cfi_startproc
 	// Not adding AARCH64_SIGN_LINK_REGISTER here because __bn_sqr8x_mont is jumped to
 	// only from bn_mul_mont which has already signed the return address.
 	cmp	$ap,$bp
 	b.ne	__bn_mul4x_mont
 .Lsqr8x_mont:
 	stp	x29,x30,[sp,#-128]!
+.cfi_def_cfa_offset 128
+.cfi_offset x30, -120
+.cfi_offset x29, -128
```
dkostic: nit:
```suggestion
.cfi_offset x29, -128
.cfi_offset x30, -120
```

crypto/fipsmodule/bn/asm/armv8-mont.pl
```diff
@@ -269,13 +280,23 @@
...
 	ldp	x21,x22,[x29,#32]
+.cfi_restore x21
+.cfi_restore x22
 	mov	x0,#1
 	ldp	x23,x24,[x29,#48]
+.cfi_restore x23
+.cfi_restore x24
 	ldr	x29,[sp],#64
+.cfi_restore x29
+.cfi_def_cfa_offset 0
```
dkostic, line 296: I don't see x30 restored?

crypto/fipsmodule/bn/asm/armv8-mont.pl
```diff
@@ -1007,6 +1043,7 @@
 .Lsqr8x8_post_condition:
 	adc	$carry,xzr,xzr
 	ldr	x30,[x29,#8]		// pull return address
+.cfi_restore x30
```
dkostic, line 1046: not related to your change but this `ldr` is weirdly placed. shouldn't it be after the `.Lsqr8x_done` label?

crypto/fipsmodule/bn/asm/armv8-mont.pl
```diff
@@ -1485,6 +1550,7 @@
 	// $acc0-3,$carry hold result, $m0-7 hold modulus
 	subs	$a0,$acc0,$m0
 	ldr	x30,[x29,#8]		// pull return address
+.cfi_restore	x30
```
dkostic, line 1553: same as above, seems to me this load is in a weird spot?

crypto/fipsmodule/bn/asm/armv8-mont.pl
andrewhop: Does the order matter? I was under the impression it was the offset that mattered unless my understanding of stp is backwards so it should be:
```
.cfi_offset x29, -56            
.cfi_offset x30, -64
```

crypto/fipsmodule/bn/asm/armv8-mont.pl
andrewhop: Oh that is weird, it looks like `bn_mul_mont` stores `x30`, never changes/restores it, and can calls `ret` to that original `x30` address. It also calls `__bn_sqr8x_mont`/`__bn_mul4x_mont` which also store `x30`, and they restore it before returning. But those functions also never use/change `x30`. So it seems like I should load it in  `bn_mul_mont` before returning, or should I update everything to not save it on the stack at all?

crypto/fipsmodule/bn/asm/armv8-mont.pl
dkostic: the order doesn't matter, it's just a visual nit :)

crypto/fipsmodule/bn/asm/armv8-mont.pl
andrewhop: Turns out x30 is modified in the 4x and 8x functions but not in the top level bn_mul_mont, I was looking at uses in the perl code and missed `my ($cnt,$carry,$topmost)=("x27","x28","x30");`in the 8x and 4x versions. Looking at armv8-mont.S it is clear x30 is modified during execution in the 

I did an experiment to remove the store x30 in the top level bn_mul_mont but that caused all the uses of it to fail. I may have messed up the store/restore logic. I tried to remove the store x30 and shift all the other stores over by 8 bytes 
```asm
bn_mul_mont:
.cfi_startproc
	AARCH64_SIGN_LINK_REGISTER
	tst	$num,#7
	b.eq	__bn_sqr8x_mont
	tst	$num,#3
	b.eq	__bn_mul4x_mont
.Lmul_mont:
	stp	x29,x19,[sp,#-56]!
.cfi_def_cfa_offset 56
.cfi_offset x19, -48
.cfi_offset x29, -56
	add	x29,sp,#0
.cfi_def_cfa x29, 56
	stp	x20,x21,[sp,#16]
.cfi_offset x20, -40
.cfi_offset x21, -32
	stp	x22,x23,[sp,#32]
.cfi_offset x22, -24
.cfi_offset x23, -16
	str	x24,[sp,#48]
.cfi_offset x24, -8

...

	ldp	x20,x21,[x29,#16]
.cfi_restore x20
.cfi_restore x21
	mov	sp,x29
.cfi_def_cfa sp, 56
	ldp	x22,x23,[x29,#32]
.cfi_restore x22
.cfi_restore x23
	mov	x0,#1
	ldr	x24,[x29,#48]
.cfi_restore x24
	ldp	x29,x19,[sp],#56
.cfi_restore x19
.cfi_restore x29
.cfi_def_cfa_offset 0
	AARCH64_VALIDATE_LINK_REGISTER
	ret
.cfi_endproc
.size	bn_mul_mont,.-bn_mul_mont
```

crypto/fipsmodule/bn/asm/armv8-mont.pl
andrewhop: Oh I see it now, updated.

PR #2592, opened 2025-08-01
Move dynamic dispatching to C for x86_64-mont5.pl

m271828:
### Issues:
Resolves P280855694
Addresses CryptoAlg-2964

### Description of changes: 
Move dynamic dispatching from assembly to C.

### DIFF
--- a/crypto/fipsmodule/bn/asm/x86_64-mont5.pl
+++ b/crypto/fipsmodule/bn/asm/x86_64-mont5.pl
@@ -54,7 +54,7 @@ open OUT,"| \"$^X\" \"$xlate\" $flavour \"$output\"";
 $addx = 1;
 for (@ARGV) { $addx = 0 if (/-DMY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX/); }
 
-# int bn_mul_mont_gather5(
+# int bn_mul_mont_gather5_nohw(
 $rp="%rdi";	# BN_ULONG *rp,
 $ap="%rsi";	# const BN_ULONG *ap,
 $bp="%rdx";	# const BN_ULONG *bp,
@@ -76,31 +76,15 @@ $m1="%rbp";
 $code=<<___;
 .text
 
-.extern	OPENSSL_ia32cap_P
-
-.globl	bn_mul_mont_gather5
-.type	bn_mul_mont_gather5,\@function,6
+.globl	bn_mul_mont_gather5_nohw
+.type	bn_mul_mont_gather5_nohw,\@function,6
 .align	64
-bn_mul_mont_gather5:
+bn_mul_mont_gather5_nohw:
 .cfi_startproc
 	_CET_ENDBR
 	mov	${num}d,${num}d
 	mov	%rsp,%rax
 .cfi_def_cfa_register	%rax
-	test	\$7,${num}d
-	jnz	.Lmul_enter
-___
-$code.=<<___ if ($addx);
-#ifndef MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX
-	leaq	OPENSSL_ia32cap_P(%rip),%r11
-	mov	8(%r11),%r11d
-#endif
-___
-$code.=<<___;
-	jmp	.Lmul4x_enter
-
-.align	16
-.Lmul_enter:
 	movd	`($win64?56:8)`(%rsp),%xmm5	# load 7th argument
 	push	%rbx
 .cfi_push	%rbx
@@ -460,29 +444,21 @@ $code.=<<___;
 .Lmul_epilogue:
 	ret
 .cfi_endproc
-.size	bn_mul_mont_gather5,.-bn_mul_mont_gather5
+.size	bn_mul_mont_gather5_nohw,.-bn_mul_mont_gather5_nohw
 ___
 {{{
 my @A=("%r10","%r11");
 my @N=("%r13","%rdi");
 $code.=<<___;
+.global bn_mul4x_mont_gather5
 .type	bn_mul4x_mont_gather5,\@function,6
 .align	32
 bn_mul4x_mont_gather5:
 .cfi_startproc
+    _CET_ENDBR
 	.byte	0x67
 	mov	%rsp,%rax
 .cfi_def_cfa_register	%rax
-.Lmul4x_enter:
-___
-$code.=<<___ if ($addx);
-#ifndef MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX
-	and	\$0x80108,%r11d
-	cmp	\$0x80108,%r11d		# check for AD*X+BMI2+BMI1
-	je	.Lmulx4x_enter
-#endif
-___
-$code.=<<___;
 	push	%rbx
 .cfi_push	%rbx
 	push	%rbp
@@ -1087,7 +1063,7 @@ ___
 }}}
 {{{
 ######################################################################
-# void bn_power5(
+# void bn_power5_nohw(
 my $rptr="%rdi";	# BN_ULONG *rptr,
 my $aptr="%rsi";	# const BN_ULONG *aptr,
 my $bptr="%rdx";	# const BN_ULONG *table,
@@ -1102,25 +1078,14 @@ my @A1=("%r12","%r13");
 my ($a0,$a1,$ai)=("%r14","%r15","%rbx");
 
 $code.=<<___;
-.globl	bn_power5
-.type	bn_power5,\@function,6
+.globl	bn_power5_nohw
+.type	bn_power5_nohw,\@function,6
 .align	32
-bn_power5:
+bn_power5_nohw:
 .cfi_startproc
 	_CET_ENDBR
 	mov	%rsp,%rax
 .cfi_def_cfa_register	%rax
-___
-$code.=<<___ if ($addx);
-#ifndef MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX
-	leaq	OPENSSL_ia32cap_P(%rip),%r11
-	mov	8(%r11),%r11d
-	and	\$0x80108,%r11d
-	cmp	\$0x80108,%r11d		# check for AD*X+BMI2+BMI1
-	je	.Lpowerx5_enter
-#endif
-___
-$code.=<<___;
 	push	%rbx
 .cfi_push	%rbx
 	push	%rbp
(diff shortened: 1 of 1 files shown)


crypto/fipsmodule/bn/asm/x86_64-mont5.pl
```diff
@@ -1085,9 +1061,9 @@
 .size	mul4x_internal,.-mul4x_internal
 ___
 }}}
-{{{
```
m271828, line 1088: Todo: find why this is marked as a change

crypto/fipsmodule/bn/asm/x86_64-mont5.pl
```diff
@@ -2119,13 +2084,14 @@
 
 $code.=<<___;
 #ifndef MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX
```
m271828, line 2086: Todo: remove

m271828: We actually need to leave these in place because it's related to the compiler, not the processor.

crypto/fipsmodule/bn/internal.h
```diff
@@ -418,18 +418,23 @@ OPENSSL_INLINE int bn_mul4x_mont_capable(size_t num) {
...
+  return 0;
+#endif // !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX)
   return bn_mul4x_mont_capable(num) && bn_mulx_adx_capable();
 }
 int bn_mulx4x_mont(BN_ULONG *rp, const BN_ULONG *ap, const BN_ULONG *bp,
                    const BN_ULONG *np, const BN_ULONG *n0, size_t num);
 OPENSSL_INLINE int bn_sqr8x_mont_capable(size_t num) {
+#if !defined(MY_ASSEMBLER_IS_TOO_OOLD_FOR_512AVX)
+  return 0;
+#endif // !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX)
```
justsmth: The macro is misspelled: `..._TOO_OOLD_...` -- and shouldn't this condition be positive?

crypto/fipsmodule/bn/internal.h
```diff
@@ -418,18 +418,23 @@ OPENSSL_INLINE int bn_mul4x_mont_capable(size_t num) {
 }
 int bn_mul4x_mont(BN_ULONG *rp, const BN_ULONG *ap, const BN_ULONG *bp,
                   const BN_ULONG *np, const BN_ULONG *n0, size_t num);
-#if !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX)
+
 OPENSSL_INLINE int bn_mulx4x_mont_capable(size_t num) {
+#if defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX)
+  return 0;
+#endif // !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX)
```
justsmth: NP: The comment has a `!` that's not in the macro condition above. (With such a small `#if` block, I'm ok with no comment here.

crypto/fipsmodule/bn/internal.h
```diff
@@ -445,18 +450,27 @@ int bn_sqr8x_mont(BN_ULONG *rp, const BN_ULONG *ap, BN_ULONG mulx_adx_capable,
...
+    return (num & 7) == 0;
+  }
+  void bn_mul4x_mont_gather5(BN_ULONG *rp, const BN_ULONG *ap,
+                             const BN_ULONG *table, const BN_ULONG *np,
+                             const BN_ULONG *n0, int num, int power);
+
+  OPENSSL_INLINE int bn_mulx4x_mont_gather5_capable(int num) {
+  #if defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX)
+    return 0;
+  #endif // !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX)
```
justsmth: NP: Comment not matching condition.

crypto/fipsmodule/bn/internal.h
```diff
@@ -470,17 +484,23 @@ void bn_scatter5(const BN_ULONG *inp, size_t num, BN_ULONG *table,
...
+                      const BN_ULONG *np, const BN_ULONG *n0, int num, int power);
+
+  OPENSSL_INLINE int bn_power5_capable(int num) {
+    return (num & 7) == 0;
+  }
+
+  OPENSSL_INLINE int bn_powerx5_capable(int num) {
+  #if defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX)
+    return 0;
+  #endif // !defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX)
```
justsmth: NP

crypto/fipsmodule/bn/internal.h
```diff
@@ -418,18 +418,23 @@ OPENSSL_INLINE int bn_mul4x_mont_capable(size_t num) {
...
 OPENSSL_INLINE int bn_mulx4x_mont_capable(size_t num) {
+#if defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX)
+  return 0;
+#endif // defined(MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX)
   return bn_mul4x_mont_capable(num) && bn_mulx_adx_capable();
 }
 int bn_mulx4x_mont(BN_ULONG *rp, const BN_ULONG *ap, const BN_ULONG *bp,
                    const BN_ULONG *np, const BN_ULONG *n0, size_t num);
 OPENSSL_INLINE int bn_sqr8x_mont_capable(size_t num) {
+#if defined(MY_ASSEMBLER_IS_TOO_OOLD_FOR_512AVX)
```
justsmth: This is still misspelled.

crypto/impl_dispatch_test.cc
```diff
@@ -140,6 +140,11 @@ class ImplDispatchTest : public ::testing::Test {
   bool is_assembler_too_old = false;
   bool is_assembler_too_old_avx512 = false;
   bool ifma_avx512 = false;
+  bool is_bn_mulx4x_mont_gather5 = false;
```
dkostic, line 143: I'm not sure that we need these five values defined here, we only define capabilities here?

crypto/impl_dispatch_test.cc
```diff
@@ -179,6 +189,85 @@ constexpr size_t kFlag_sha3_keccak4_f1600_alt = 13;
...
+        BN_MONT_CTX_new_for_modulus(m.get(), ctx()));
+    ASSERT_TRUE(mont);
+
+    std::vector<BN_ULONG> r(words), a(words), b(words), table(words * 32);
+    a[0] = 1;
+    b[0] = 42;
+
+    bn_mul_mont(r.data(), a.data(), b.data(), mont->N.d, mont->n0, words);
+
+    is_bn_mulx4x_mont_gather5 = bn_mulx4x_mont_gather5_capable(words);
```
dkostic, line 219: if you use the same function `bn_mulx4x_mont_gather5_capable` for the test and the actual implementation then this test doesn't make much sense, it will always pass. The idea of dispatch tests is to determine if a specific combination of CPU capabilities (like ADX, BMI, BMI2, etc.) triggers the expected code path.

crypto/fipsmodule/bn/exponentiation.c
```diff
@@ -163,6 +163,56 @@ static void exponentiation_s2n_bignum_copy_from_prebuf(BN_ULONG *dest, int width
...
+// 16 bytes. |power| must be less than 32 and is treated as secret.
+//
+// WARNING: This function implements Almost Montgomery Multiplication from
+// https://eprint.iacr.org/2011/239. The inputs do not need to be fully reduced.
+// However, even if they are fully reduced, the output may not be.
+static void bn_mul_mont_gather5(
+    BN_ULONG *rp, const BN_ULONG *ap, const BN_ULONG *table, const BN_ULONG *np,
+    const BN_ULONG *n0, int num, int power) {
+  if (bn_mulx4x_mont_gather5_capable(num)) {
+    keccak_log_dispatch(15);
```
dkostic: can we please re-define `keccak_log_dispatch` to a generic log_dispatch function?

crypto/impl_dispatch_test.cc
m271828: This matches how we do it for the other dynamic dispatching tests. The only difference is they're not value dependent like these are.

crypto/impl_dispatch_test.cc
m271828: We can pull it out but we'll also need to pull out the tests on the numeric value as well.

crypto/impl_dispatch_test.cc
dkostic: not sure I understand. All other variables here refer to specific CPU capabilities like AES-HW, SHA-EXT, AVX, IFMA, etc., while `bn_mulx4x_mont_gather5` is not a capability. Could you please explain why you think you need these variables defined instead of specifica CPU capabilities that are required for those functions to run?

crypto/impl_dispatch_test.cc
dkostic: I didn't mean to remove the test but to modify it such that it actually tests the dispatching logic. Take a look at this test for example: https://github.com/m271828/aws-lc/blob/64fcb96bb5764a02e4e2d79ff17cf202b8a53a18/crypto/impl_dispatch_test.cc#L486

include/openssl/cpu.h
```diff
@@ -14,5 +14,12 @@
 
 // This header is provided for compatibility with older revisions of BoringSSL.
 // TODO(davidben): Remove this header.
+#ifndef OPENSSL_HEADER_CPU_H
+#define OPENSSL_HEADER_CPU_H
 
 #include "crypto.h"
+
+__attribute__((unused))
+void log_dispatch(size_t id);
```
dkostic, line 23: this should not be in a public header. maybe it can go in `crypto/internal.h`?

crypto/impl_dispatch_test.cc
m271828: The dispatching function isn't externally reachable. We'd need to either duplicate the logic, export the dispatching function or call from a higher level (figuring out the right values to use).

crypto/impl_dispatch_test.cc
m271828: These features are CPU capability dependent but also value dependent. We would do the CPU capabilities as global booleans like the others and leave the value as a dynamic booleans inside the test.

PR #2594, opened 2025-08-02
Complete integrity check for Linux and MacOS shared builds

prikhap:
Description
This pull request implements a C++ version of inject_hash for AWS-LC's FIPS module integrity verification. The implementation focuses on using LIEF for binary parsing for Linux and MacOS shared builds and the calculating hashes of non-relocatable module sections. It also maintains compatibility with existing FIPS integrity verification workflow

Key Implementation Changes
Added comprehensive platform-specific binary parsing
Implemented HMAC-SHA256 calculation for static module sections
Added proper error handling and validation

Files Modified
tests/ci/run_inject_hash_cpp.sh: Updated test script with new test cases
util/fipstools/inject_hash_cpp/inject_hash.cpp: Main implementation file with hash injection functionality implemented

Testing
Basic argument validation
Platform-specific behavior
Invalid file handling
Successful hash injection
Hash verification using crypto_test

### DIFF
--- a/tests/ci/run_inject_hash_cpp.sh
+++ b/tests/ci/run_inject_hash_cpp.sh
@@ -102,8 +102,14 @@ END_OFFSET=$((16#$END_OFFSET))
 RANDOM_OFFSET=$((START_OFFSET + RANDOM % (END_OFFSET - START_OFFSET)))
 echo "Corrupting FIPS module at offset $RANDOM_OFFSET"
 
-# Corrupt one byte
-printf '\x00' | dd of=./test_corrupted/libcrypto.${LIB_EXT} bs=1 seek=$RANDOM_OFFSET count=1 conv=notrunc
+# Read current byte
+CURRENT_BYTE=$(dd if=./test_corrupted/libcrypto.${LIB_EXT} bs=1 skip=$RANDOM_OFFSET count=1 2>/dev/null | xxd -p)
+
+# Flip the bit (XOR with 1)
+FLIPPED_BYTE=$(printf "%02x" "$((0x$CURRENT_BYTE ^ 1))")
+
+# Write back the flipped byte
+echo $FLIPPED_BYTE | xxd -r -p | dd of=./test_corrupted/libcrypto.${LIB_EXT} bs=1 seek=$RANDOM_OFFSET count=1 conv=notrunc
 
 # Change to test directory and set library path
 cd ./test_corrupted
--- a/util/fipstools/inject_hash_cpp/inject_hash.cpp
+++ b/util/fipstools/inject_hash_cpp/inject_hash.cpp
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0 OR ISC
 
 // currently only supports shared builds for Linux and MacOS
-#include <getopt.h>
 #include <inttypes.h>
 #include <openssl/base.h>
 #include <openssl/hmac.h>
@@ -341,7 +340,6 @@ static bool calculate_hash(const std::vector<uint8_t> &text_module,
 
 static bool process(const std::string &input_path,
                     const std::string &output_path, bool is_apple) {
-  // Read the input binary file
   std::ifstream input_file(
       input_path, std::ios::binary |
                       std::ios::ate);  // Open the file in binary mode and move
@@ -356,6 +354,7 @@ static bool process(const std::string &input_path,
   input_file.seekg(0, std::ios::beg);  // Move back to the beginning of the file
                                        // to read its content
 
+   // Read the input binary file
   std::vector<uint8_t> binary_data(file_size);
   if (!input_file.read(reinterpret_cast<char *>(binary_data.data()),
                        file_size)) {


util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+    GetBoolArgument(&is_apple, "-apple", args_map);
+
+    // Display configuration
+    std::cout << "Input file:  " << input_file << std::endl;
+    std::cout << "Output file: " << output_file << std::endl;
+    if (is_apple) {
+        std::cout << "Platform: macOS" << std::endl;
+    }
+
+    InjectHash injector(input_file, output_file, is_apple);
```
samuel40791765: I'm not sure if using a class is really applicable here. If feels like overkill imo.

We could achieve the same thing with just a single `process` function calling either `extract_apple_modules` or `extract_linux_modules` depending on whether `is_apple` is `true` or not. You can take note of what `inject_hash.go` was originally doing.

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
-      input_file.substr(input_file.length() - 6) == ".dylib") {
-    if (LIEF::Parser::parse(input_file)) {
-      std::cout << "LIEF parser successfully loaded: " << input_file
-                << std::endl;
-    } else {
-      std::cerr << "LIEF parser failed to load: " << input_file << std::endl;
-      return 1;
+class InjectHash {
+private:
+    // Placeholder hash pattern from fips_shared_support.c
```
samuel40791765:
```suggestion
    // UninitHashValue is the default hash value that we inject into the module.
    // This value need only be distinct, i.e. so that we can safely
    // search-and-replace it in an object file.
```
The description in `const.go` is better imo: https://github.com/aws/aws-lc/blob/main/util/fipstools/fipscommon/const.go#L20

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+    std::string input_path_;
+    std::string output_path_;
+    bool is_apple_;
+
+    // Convert a size to little-endian representation
+    static void size_to_little_endian(size_t size, uint8_t* result) {
+        for (int i = 0; i < 8; i++) {
+            result[i] = (size >> (i * 8)) & 0xFF;
+        }
+    }
```
samuel40791765: Why does the little endian conversion here need to be done?

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+        0x68, 0x27, 0xcb, 0xa0, 0xa0, 0x9f, 0x6b, 0x6f, 
+        0xde, 0x52, 0xcd, 0xe2, 0xcd, 0xff, 0x31, 0x80
+    };
+
+    std::string input_path_;
+    std::string output_path_;
+    bool is_apple_;
+
+    // Convert a size to little-endian representation
+    static void size_to_little_endian(size_t size, uint8_t* result) {
```
samuel40791765: 1. We should add an additional field for `result_len` here. Passing a pointer without a specified length in C/C++ is dangerous. 
2. The meaning of `size` here is fairly vague. What does it mean and can we change the naming?

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+    }
+
+    // Calculate HMAC hash
+    std::vector<uint8_t> calculate_hash(const std::vector<uint8_t>& text_module,
+                                       const std::vector<uint8_t>& rodata_module) {
+        std::vector<uint8_t> hash_result(32); // SHA256 size
+        uint8_t zero_key[64] = {0};
+        uint8_t length_bytes[8];
+        
+        HMAC_CTX* ctx = HMAC_CTX_new();
```
samuel40791765: We can use a smart pointer here to maintain the memory for us. Then we won't have to call `HMAC_CTX_free` at every code branch and C++ will free the memory for us.

Here's an example within our tests: https://github.com/aws/aws-lc/blob/31bc08745a6212bd7d81a29d001a2e93b4de1521/crypto/hmac_extra/hmac_test.cc#L265

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+                rodata_module.clear();
+                }
+
+            return true;
+    }
+
+    // Calculate HMAC hash
+    std::vector<uint8_t> calculate_hash(const std::vector<uint8_t>& text_module,
+                                       const std::vector<uint8_t>& rodata_module) {
+        std::vector<uint8_t> hash_result(32); // SHA256 size
```
samuel40791765: There's something handy for us to use here :) https://github.com/aws/aws-lc/blob/main/include/openssl/sha.h#L139-L140
```suggestion
        std::vector<uint8_t> hash_result(SHA256_DIGEST_LENGTH);
```

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+        std::ifstream input_file(input_path_, std::ios::binary | std::ios::ate); // Open the file in binary mode and move to the end to get size (the ate pointer gets to the end of the file)
+        if (!input_file) {
+            LOG_ERROR("Failed to open input file: %s", input_path_.c_str());
+            return false;
+        }
+        
+        size_t file_size = input_file.tellg();
+        input_file.seekg(0, std::ios::beg); // Move back to the beginning of the file to read its content
+        
+        // ??
```
samuel40791765:
```suggestion
```

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+    if (!GetString(&input_file, "-in-object", "", args_map) ||
+        !GetString(&output_file, "-o", "", args_map)) {
+        std::cerr << "❌ Error getting file paths" << std::endl;
+        print_usage();
+        return 1;
+    }
+
+    // Get the apple flag
+    bool is_apple = false;
+    GetBoolArgument(&is_apple, "-apple", args_map);
```
samuel40791765: We should check the return value here.

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+        }
+
+        HMAC_CTX_free(ctx);
+
+        // Print calculated hash for reference
+        LOG_INFO("Calculated hash: ");
+        for (unsigned int i = 0; i < hash_len; i++) {
+            printf("%02X", hash_result[i]);
+        }
+        printf("\n");
```
samuel40791765: We can probably use `LOG_INFO` here?

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+    std::vector<std::string> args;
+    for (int i = 1; i < argc; i++) {
+        args.push_back(argv[i]);
+    }
+
+    args_map_t args_map;
+    args_list_t extra_args;
+
+    if (!ParseKeyValueArguments(args_map, extra_args, args, kArguments)) {
+        std::cerr << "❌ Failed to parse arguments" << std::endl;
```
samuel40791765: We can use `LOG_ERROR` here. Same for the rest of the `std::cout/cerr` calls in `main` here.

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+            }
+        }
+
+        if (hash_pos == std::string::npos) {
+            LOG_ERROR("Could not find placeholder hash in binary");
+            return false;
+        }
+
+        //  Calculate the hash
+        auto calculated_hash = calculate_hash(text_module, rodata_module);
```
samuel40791765: If `calculate_hash` fails, we should be able to account for that error and `return false` to abort the process. As of now, we're directly copying an empty value into the binary if `calculate_hash` fails.

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+
+            // Find boundary symbols
+            const auto* text_start = elf->get_symbol("BORINGSSL_bcm_text_start");
+            if (!text_start) {
+                LOG_ERROR("Could not find BORINGSSL_bcm_text_start symbol");
+            }
+            const auto* text_end = elf->get_symbol("BORINGSSL_bcm_text_end");
+            if (!text_end) {
+                LOG_ERROR("Could not find BORINGSSL_bcm_text_end symbol");
+            }
```
samuel40791765: It's great that we're checking for the value before proceeding here! Could we apply the same to `rodata_start/end` below and to the same values in `extract_apple_modules` above?

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+                    text_section = &section;
+                    LOG_INFO("Found .text section at virtual address: 0x%llx", section.virtual_address());
+                }
+                if (section.name() == ".rodata") {
+                    rodata_section = &section;
+                    LOG_INFO("Found .rodata section at virtual address: 0x%llx", section.virtual_address());
+                }
+            }
+
+            if (!text_section) {
```
samuel40791765:
```suggestion
            if (!text_section || !rodata_section) {
```

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+                LOG_ERROR("Could not locate .text section");
+                return false;
+                                 }
+
+            // Calculate module boundaries within sections
+            auto text_sec_addr = text_section->virtual_address();
```
samuel40791765: We should check the return value before actually using it.

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+                if ((rodata_start == nullptr) != (rodata_section == nullptr)) {
+                LOG_ERROR("rodata start marker inconsistent with rodata section presence");
+                return false;
+                }
+                if ((rodata_start != nullptr) != (rodata_end != nullptr)) {
+                    LOG_ERROR("rodata marker presence inconsistent");
+                    return false;
+                }
+                rodata_module.clear();
+                }
```
samuel40791765: Should we `return false` here?

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +31,382 @@ static void print_usage() {
...
+                    LOG_ERROR("Invalid rodata module boundaries");
+                    return false;
+                }
+
+                auto rodata_content = rodata_section->content();
+                rodata_module.resize(rodata_end_offset - rodata_start_offset);
+                std::memcpy(rodata_module.data(), rodata_content.data() + rodata_start_offset, rodata_module.size());
+            }
+            else {
+                rodata_module.clear();
```
samuel40791765: Should we return false here?

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -60,31 +60,54 @@ run_test "No arguments test" true \
...
 ((ERRORS+=$?))
 
-# Test 3: Actual library (should succeed)
+# Platform Specific Tests
 if [[ "$OSTYPE" == "darwin"* ]]; then
-    run_test "Valid library test (macOS)" false \
+    echo "MacOS specific tests..."
+    
+    # Test 3: macOS without -apple flag (should fail)
+    run_test "macOS missing apple flag test" true \
```
andrewhop: Instead of requiring the user to pass in -apple can inject_hash/LIEF figure this out automatically? Also will we need to handle Windows in a special way? If so I prefer a generic option that is an enum instead of multiple booleans.

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -60,31 +60,54 @@ run_test "No arguments test" true \
...
         -o ./crypto/libcrypto.dylib \
-        -in-object ./crypto/libcrypto.dylib \
-        -apple
+        -in-object ./crypto/libcrypto.dylib
+    ((ERRORS+=$?))
+
+    # Test 4: Create corrupted library copy and invalid hash check
+    echo "Creating corrupted copy of library..."
+    cp ./crypto/libcrypto.dylib ./crypto/libcrypto_corrupted.dylib
+    printf '\x00' | dd of=./crypto/libcrypto_corrupted.dylib bs=1 seek=1024 count=1 conv=notrunc
```
andrewhop: 1. What happens if the FIPS module is not 1024 bytes into the library?
2. Are you sure this is corrupting the FIPS module and the self check is failing and not overriding some other part of the library?

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -60,31 +60,54 @@ run_test "No arguments test" true \
...
+    export DYLD_LIBRARY_PATH="./crypto"
+    run_test "Corrupted hash verification test" true \
+        ./crypto/crypto_test
```
andrewhop: How does this work? Won't crypto_test load the valid libcrypto.dylib while you corrupted libcrypto_corrupted.dylib?

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -60,31 +60,54 @@ run_test "No arguments test" true \
...
         -o ./crypto/libcrypto.so \
-        -in-object ./crypto/libcrypto.so
+        -in-object ./crypto/libcrypto.so \
+        -apple
+    ((ERRORS+=$?))
+    # Test 6: Create corrupted library copy and invalid hash check
+    echo "Creating corrupted copy of library..."
+    cp ./crypto/libcrypto.so ./crypto/libcrypto_corrupted.so
+    printf '\x00' | dd of=./crypto/libcrypto_corrupted.so bs=1 seek=1024 count=1 conv=notrunc
+    export LD_LIBRARY_PATH="./crypto_corrupted"
```
andrewhop: Same question, there is nothing in the crypto_corrupted folder.

util/fipstools/inject_hash_cpp/inject_hash.cpp
andrewhop: This code is not endian aware, what happens if this runs on a big endian CPU and the order is reversed? Can this use a utility function from AWS-LC like `CRYPTO_store_u64_le`?

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -27,84 +32,394 @@ static void print_usage() {
...
+                    LOG_ERROR("rodata marker presence inconsistent");
+                    return false;
+                }
+                rodata_module.clear();
+                }
+
+            return true;
+    }
+
+    bool extract_linux_modules(std::vector<uint8_t>& text_module, std::vector<uint8_t>& rodata_module) {
```
andrewhop: Can the majority of this linux and apple functions be combined if there were two small functions that used the ELF/MACHO parser from LIEF and then call a common function that does the majority of the work to find the symbols?

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: I understand, I was trying to use standard C++ practice to use classes and also to reduce the number of parameter passing. But your point is correct. Hence I have switched to the use of functions (in the way go was doing it)

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: modified.

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: Different platforms either use big endian or little endian by default. By specifying what endian we are going with, it would make the hash calculation consistent. Moreover, this also used in bcm.c for text section's size and rodata section's size

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: changed

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: changed

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: changed

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: changed

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: rodata_start/end even in inject_hash.go and bcm.c file is indicated as optional.

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: rodata_start/end even in inject_hash.go and bcm.c file is indicated as optional.

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: changed.

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: changed

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: changed return type of calculated_hash to bool and added the hash_result as out parameter.

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: No, we shouldn't return false here because rodata is optional

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: No, we shouldn't return false here because rodata is optional

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: I think this would complicate the entire function, although most of the operations in the functions seem similar, but there are slight differences in the way of parsing the both files, potential placeholder for static build in the linux function and names of linker markers. 
Too many conditional statements to in a function would convolute it, moreover gives difficulty in tracking potential LOG ERRORS and INFO

prikhap: Also, on our discussion on why does extract_apple_modules() functions has this statement: `auto macho = binary->at(0); ` and extract_linux_modules() does not, the reason is that LIEF supports producing a FAT binary for macOS where there are binaries for the ARM architecture and X86 architectures at auto macho = binary->at(0) and at(1). 

However on experimenting more with it, I realised that AWS-LC's CMake configuration explicitly checks for and prevents universal (fat) binary builds, hence the resulting libcrypto.dylib contains exactly one architecture. Hence even if LIEF can support universal binary builds, but AWS-LC restricts it.
So, when I try accessing 'auto macho = binary->at(1);', it fails since binary for one of the builds is created. So irrespective of which architecture we are referring to, `auto macho = binary->at(0);` points to it's binary.

tests/ci/run_inject_hash_cpp.sh
prikhap: I understand your point, in this approach we are not locating the FIPS module, hence I have changes the approach in the latest commit

tests/ci/run_inject_hash_cpp.sh
prikhap: I understand your point, in this approach we are not locating the FIPS module, hence I have changed the approach in the latest commit

tests/ci/run_inject_hash_cpp.sh
prikhap: I understand your point, in this approach we are not locating the FIPS module, hence I have changed the approach in the latest commit

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -21,14 +27,402 @@ static const argument_t kArguments[] = {
...
+      "Usage: inject_hash_cpp -o output_file -in-object input_file [-apple]");
   PrintUsage(kArguments);
 }
 
-int main(int argc, char **argv) {
-  // Convert argv to vector of strings, skipping program name (argv[0])
+static constexpr uint8_t UNINIT_HASH[32] = {
+    0xae, 0x2c, 0xea, 0x2a, 0xbd, 0xa6, 0xf3, 0xec, 0x97, 0x7f, 0x9b,
+    0xf6, 0x94, 0x9a, 0xfc, 0x83, 0x68, 0x27, 0xcb, 0xa0, 0xa0, 0x9f,
+    0x6b, 0x6f, 0xde, 0x52, 0xcd, 0xe2, 0xcd, 0xff, 0x31, 0x80};
```
samuel40791765, line 41: We're missing the comment for https://github.com/aws/aws-lc/pull/2594#discussion_r2252574897

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -21,14 +27,402 @@ static const argument_t kArguments[] = {
...
+    LOG_ERROR("HMAC_Init failed");
+    // HMAC_CTX_free(ctx);
+    return false;
+  }
+
+  // Add text module size and content
+  if (!size_to_little_endian(text_module.size(), length_bytes,
+                             sizeof(length_bytes))) {
+    LOG_ERROR("Buffer too small for little-endian conversion");
+    // HMAC_CTX_free(ctx);
```
samuel40791765: Please remove the `HMAC_CTX_free(ctx);` that were commented out.

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -21,14 +27,402 @@ static const argument_t kArguments[] = {
...
+                           std::vector<uint8_t> &out_hash) {
+  out_hash.resize(SHA256_DIGEST_LENGTH);
+  uint8_t zero_key[64] = {0};
+  uint8_t length_bytes[8];
+
+  bssl::ScopedHMAC_CTX ctx;
+  // if (!ctx) {
+  //     LOG_ERROR("Failed to create HMAC context");
+  //     return hash_result;
+  // }
```
samuel40791765: We should still check if the context was initialized correctly here.

util/fipstools/inject_hash_cpp/inject_hash.cpp
```diff
@@ -21,14 +27,402 @@ static const argument_t kArguments[] = {
...
+    LOG_ERROR("Could not find BORINGSSL_bcm_text_start symbol");
+    return false;
+  }
+  const auto *text_end = macho->get_symbol("_BORINGSSL_bcm_text_end");
+  if (!text_end) {
+    LOG_ERROR("Could not find BORINGSSL_bcm_text_end symbol");
+    return false;
+  }
+  const auto *rodata_start = macho->get_symbol("_BORINGSSL_bcm_rodata_start");
+  const auto *rodata_end = macho->get_symbol("_BORINGSSL_bcm_rodata_end");
```
samuel40791765: I remember you mentioning that these are allowed to be `nullptr`, could we leave a brief explanation as to why in a comment above?

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -60,32 +59,56 @@ run_test "No arguments test" true \
...
+    # Try multiple offsets to increase chance of hitting FIPS module
+    for OFFSET in 1024 2048 4096 8192 16384; do
+        echo "Corrupting at offset $OFFSET"
+        printf '\x00' | dd of=./test_corrupted/libcrypto.${LIB_EXT} bs=1 seek=$OFFSET count=1 conv=notrunc
+    done
+fi
+
+# Run test with the corrupted library
+run_test "Corrupted hash verification test" true \
+    bash -c "cd ./test_corrupted && ${LIB_PATH_VAR}=./ ./crypto_test 2>&1 | grep -q 'integrity'"
```
samuel40791765: Is there a test that verifies that cpp inject hash is working?

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: added.

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: modified.

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: yes I can probably check the context using if(!ctx.get()), that can return the context pointer. if(!ctx) couldn't do a boolean comparison earlier, hence I had commented it earlier.

util/fipstools/inject_hash_cpp/inject_hash.cpp
prikhap: I have recently actually changed the logic to let rodata be optional. 
code in bcm.c indicates that shared builds do not have optional rodata sections, and this PR is concerned with shared builds only. For static builds however, there are no bcm_rodata* markers because the rodata is embedded in the text section. ref: https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/bcm.c#L319-L326. This was probably the reason the go script (which supported shared and static builds both) had kept rodata extraction optional.

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -60,32 +59,70 @@ run_test "No arguments test" true \
...
+# Convert hex to decimal
+START_OFFSET=$((16#$START_OFFSET))
+END_OFFSET=$((16#$END_OFFSET))
+
+# Pick a random offset within the module
+RANDOM_OFFSET=$((START_OFFSET + RANDOM % (END_OFFSET - START_OFFSET)))
+echo "Corrupting FIPS module at offset $RANDOM_OFFSET"
+
+# Corrupt one byte
+printf '\x00' | dd of=./test_corrupted/libcrypto.${LIB_EXT} bs=1 seek=$RANDOM_OFFSET count=1 conv=notrunc
```
andrewhop: What if that byte was already all zero?

tests/ci/run_inject_hash_cpp.sh
```diff
@@ -60,32 +59,70 @@ run_test "No arguments test" true \
...
-run_test "HMAC calculation test" false \
-    ./util/fipstools/inject_hash_cpp/inject_hash_cpp \
-    -in-object "util/fipstools/inject_hash_cpp/CMakeLists.txt" \
-    -o "util/fipstools/inject_hash_cpp/CMakeLists.txt"
+# Test 3: Sanity check - verify tests pass with injected hash
+run_test "Sanity check" false \
+    ./crypto/crypto_test
 ((ERRORS+=$?))
 
+# Test 4: Create corrupted library copy and invalid hash check
```
andrewhop, line 81: Using the break hash or KAT go programs would simplify a lot of this code.

tests/ci/run_inject_hash_cpp.sh
prikhap: I think you're right. I could probably just flip the byte to corrupt it.

tests/ci/run_inject_hash_cpp.sh
prikhap: I dont think directly importing them into the test file would be a good practice, adding more go dependencies at test time. Moreover, the current test case more or less does the same thing.

smittals2 (approved):

PR #2603, opened 2025-08-06
Support other field for PKCS7

samuel40791765:
### Issues:
Addresses `CryptoAlg-3279`

### Description of changes: 
OpenSSL's `PKCS7` structure has historically been non-opaque. This causes issues like projects depending on the `PKCS7` internals to be of a certain format. OpenSSL's `PKCS7` struct has a field called `other`, which various projects internally have found to have a dependence on. The usage spans multiple projects and seeing that OpenSSL 3.x still supports the field, this isn't going away anytime soon.

examples:
* https://github.com/tianocore/edk2/blob/686f1e3ea8373db48ea3a8457cc758c986455de0/CryptoPkg/Library/BaseCryptLib/Pk/CryptPkcs7VerifyBase.c#L69-L71
* https://github.com/krb5/krb5/blob/820a7277ed1c553d33b2a7f3efb48655fb993b77/src/plugins/preauth/pkinit/pkinit_crypto_openssl.c#L1615-L1636

### Call-outs:
N/A

### Testing:
New tests

### DIFF
--- a/crypto/pkcs7/pkcs7_asn1.c
+++ b/crypto/pkcs7/pkcs7_asn1.c
@@ -10,7 +10,7 @@
 #include "../internal.h"
 #include "internal.h"
 
-ASN1_ADB_TEMPLATE(p7default) = ASN1_EXP_OPT(PKCS7, d.data, ASN1_ANY, 0);
+ASN1_ADB_TEMPLATE(p7default) = ASN1_EXP_OPT(PKCS7, d.other, ASN1_ANY, 0);
 
 ASN1_ADB(PKCS7) = {
     ADB_ENTRY(NID_pkcs7_data,
--- a/crypto/pkcs7/pkcs7_test.cc
+++ b/crypto/pkcs7/pkcs7_test.cc
@@ -2170,7 +2170,7 @@ iWjrtmwM/HRbFEg2THS9b/vkiTsNSRCR9goaq9KPqXuJJsjJIoMA8IBHSLVvFnLf
   // Run |PKCS7_verify| again to check that we're consuming a copy of the
   // underlying |EVP_MD_CTX|.
   EXPECT_TRUE(PKCS7_verify(pkcs7.get(), nullptr, store.get(), nullptr,
-                         out.get(), /*flags*/ 0));
+                           out.get(), /*flags*/ 0));
 }
 
 TEST(PKCS7Test, PKCS7SignedAttributesRuby) {
@@ -2222,7 +2222,7 @@ L6qflZ9YCU5erE4T5U98hCQBMh4nOYxgaTjnZzhpkKQuEiKq/755cjzTzlI/eok=
   // The test has an expected output. Check that we output the same contents
   // as OpenSSL.
   static const char kPKCS7RubyOpenSSLOutput[] =
-    R"(-----BEGIN PKCS7-----
+      R"(-----BEGIN PKCS7-----
 MIIDawYJKoZIhvcNAQcDoIIDXDCCA1gCAQAxggEQMIIBDAIBADB1MHAxEDAOBgNV
 BAoMB2V4YW1wbGUxFzAVBgNVBAMMDlRBUk1BQyBST09UIENBMSIwIAYJKoZIhvcN
 AQkBFhNzb21lb25lQGV4YW1wbGUub3JnMQswCQYDVQQGEwJVUzESMBAGA1UEBwwJ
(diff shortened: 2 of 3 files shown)


crypto/pkcs7/pkcs7_asn1.c
```diff
@@ -10,7 +10,7 @@
 #include "../internal.h"
 #include "internal.h"
 
-ASN1_ADB_TEMPLATE(p7default) = ASN1_EXP_OPT(PKCS7, d.data, ASN1_ANY, 0);
+ASN1_ADB_TEMPLATE(p7default) = ASN1_EXP_OPT(PKCS7, d.other, ASN1_ANY, 0);
```
justsmth, line 13: Do we know whether any projects might depend on our previous behavior, expecting such values to be placed in `data`.

crypto/pkcs7/pkcs7_test.cc
```diff
@@ -2272,3 +2272,31 @@ tcH961onq8Tme2ICaCzk
   BIO_get_mem_ptr(out3.get(), &buf);
   EXPECT_EQ(Bytes(buf->data, buf->length), Bytes(kPKCS7RubyOpenSSLOutput));
 }
+
+TEST(PKCS7Test, VerifyOtherField) {
```
justsmth: Could we also add some "roundtrip" tests that serialize, then deserialize to verify that the data comes back in the expected form?

crypto/pkcs7/pkcs7_asn1.c
samuel40791765: I don't think so. The functionality's relatively new and the only projects dependent on this were looking to consume the same implementation as OpenSSL.

crypto/pkcs7/pkcs7_test.cc
samuel40791765: Yeah great call, I'll do so.

samuel40791765 (commented):
> Can we also seed our "pkcs7 verify" fuzz test with some example inputs that use the new "other" field?

I'll try seeing if I can put this in the same PR.

crypto/pkcs7/pkcs7_asn1.c
WillChilds-Klein: Confirmed that [this is what OpenSSL does](https://github.com/openssl/openssl/blob/389728876b51de0df9f97b6a295948ebec1e0f0c/crypto/pkcs7/pk7_asn1.c#L21).

crypto/pkcs7/pkcs7_asn1.c
```diff
@@ -25,7 +25,7 @@ ASN1_ADB(PKCS7) = {
     ADB_ENTRY(
         NID_pkcs7_encrypted,
         ASN1_EXP_OPT(PKCS7, d.encrypted, PKCS7_ENCRYPT,
-                     0))} ASN1_ADB_END(PKCS7, 0, type, 0, &p7default_tt, NULL);
+                     0))} ASN1_ADB_END(PKCS7, 0, type, 0, &p7default_tt, &p7default_tt);
```
WillChilds-Klein, line 28: why do we diverge from [what OpenSSL does](https://github.com/openssl/openssl/blob/389728876b51de0df9f97b6a295948ebec1e0f0c/crypto/pkcs7/pk7_asn1.c#L30) here?

samuel40791765: This was the key to the memory leak I had to resolve. The last field in `ASN1_ADB_END` points to `null_tt` within the `ASN1_ADB_st` structure.
https://github.com/aws/aws-lc/blob/04875dbbd6610a91855dcdc8edc268da289cb6d9/include/openssl/asn1t.h#L374-L375

`type` is the selector field for the ASN1 structure. `default_tt` returns the designated default type for the selector field, while `null_tt` returns the type to return if the selector field is `NULL`. 

I was testing against a circumstance where only `d.other` had been designated a value, but where `type` is `NULL`. Interestingly enough, this causes a memory leak once `PKCS7` goes out of scope in AWS-LC. I pinned the reason to us returning `NULL` for the selector field (since `type` hasn't been defined) and our ASN1 macros think that nothing has to be freed here. Said ASN1 code is used below:
* https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/asn1/tasn_utl.c#L209-L213
* https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/asn1/tasn_fre.c#L106-L110

This doesn't happen in OpenSSL for some reason. `type` seems to already have a value assigned and cases default to `default_tt` to free all related memory. I don't necessarily think this is the "correct behavior" though, it's a bit sketchy since we haven't really initialized the `type` field. My assumption is that the different behavior might have something to do with us calling `OpenSSL_zalloc` in more places and most undefined fields are actually `NULL`.

Changing `null_tt` to a proper default value allows us to properly free values in `d.other`, even if `type` has no value assigned to it (like OpenSSL). There won't be double free issues, mainly because `OpenSSL_free` already does ` NULL` check for us. I added a test called `TEST(PKCS7Test, OtherFieldMemoryLeak)` mainly to test this behavior.

WillChilds-Klein: >  I pinned the reason to us returning NULL for the selector field (since type hasn't been defined) and our ASN1 macros think that nothing has to be freed here.

that makes sense.

> This doesn't happen in OpenSSL for some reason.

perhaps it does, and there isn't similar valgrind CI?

anyway, changes look good to me.

WillChilds-Klein (approved):

crypto/pkcs7/pkcs7_asn1.c
samuel40791765: I ran a mini test with OpenSSL locally with valgrind and verified that there was no memory leakage with OpenSSL. I dove deeper and found that reason being `type` having a non-NULL value and ASN1 cases default to `default_tt` to free the memory because of that. In AWS-LC, we default to `null_tt` instead since uninitialized memory is all NULL.

PR #2610, opened 2025-08-13
Implement read/write timeouts for BIO datagram

smittals2:
### Issues:
`CryptoAlg-3287`
`CryptoAlg-3331`

### Description of changes: 
Added 
- BIO_CTRL_DGRAM_SET_RECV_TIMEOUT
- BIO_CTRL_DGRAM_GET_RECV_TIMEOUT
- BIO_CTRL_DGRAM_SET_SEND_TIMEOUT
- BIO_CTRL_DGRAM_GET_SEND_TIMEOUT

### Testing:
- Read timeout test added
- For Write timeout, the APIs are tested but the behavior is not. Filling the send buffer to force a timeout with UDP is unreliable across different OS (buffers can be very large, or instead of timing out OS starts dropping packets). 

### Callouts: 
- We also change how we track error states for datagram sockets. Previously, on read/write errors, we would only store certain errors if they were considered "retryable" or non-fatal. In the case of timeouts, WSAETIMEDOUT on Windows is not considered retriable and therefore was never tracked in our error handling, making that code path unreachable. This PR changes this behavior to store the error value regardless of whether it is retriable, so we can always track the error state.


crypto/bio/bio_socket_test.cc
```diff
@@ -807,6 +807,96 @@ TEST_P(BIODgramTest, SocketDatagramConnect) {
...
+    EXPECT_EQ(1, BIO_dgram_recv_timedout(bio.get()));
+    EXPECT_EQ(0, BIO_dgram_send_timedout(bio.get()));
+  }
+
+  // Part 2: Test send timeout
+  // Triggering a send timeout requires filling the kernel's send buffer, which
+  // is unreliable. The OS may drop packets or have a very large buffer,
+  // preventing the send call from blocking. The API's get/set functionality
+  // is verified in the |SocketDatagramTimeouts| test.
+}
```
justsmth: Missing part 2?

smittals2: Its intentionally not there, i wasn't able to simulate forcing a write timeout. I'll get rid of the pt. 2 from the comments. In SocketDatagramTimeouts (the test above this one) we just do a simple get/set check to ensure API works correctly. But we are not able to test the behavior.

crypto/bio/dgram.c
```diff
@@ -80,9 +93,11 @@ static int dgram_write(BIO *bp, const char *in, const int in_len) {
...
 
   BIO_clear_retry_flags(bp);
-  if (ret <= 0 && bio_socket_should_retry(ret)) {
-    BIO_set_retry_write(bp);
-    data->_errno = bio_sock_error_get_and_clear(bp->num);
+  if (ret <= 0) {
+    if (bio_socket_should_retry(ret)) {
+      BIO_set_retry_write(bp);
+    }
+    data->_errno = dgram_get_errno();
```
justsmth, line 100: Why did we switch from using `bio_sock_error_get_and_clear(bp->num)`, both here and below?

crypto/bio/dgram.c
```diff
@@ -255,6 +272,116 @@ static long dgram_ctrl(BIO *bp, const int cmd, const long num, void *ptr) {
       }
       break;
     }
+#if defined(SO_RCVTIMEO)
+    case BIO_CTRL_DGRAM_SET_RECV_TIMEOUT: {
+      GUARD_PTR(ptr);
+#ifdef OPENSSL_WINDOWS
+      struct timeval *tv = (struct timeval *)ptr;
+      int timeout = tv->tv_sec * 1000 + tv->tv_usec / 1000;
```
justsmth: * The `tv_sec` and `tv_usec` [values are signed](https://learn.microsoft.com/en-us/windows/win32/api/winsock/ns-winsock-timeval), so can potentially be negative. We should check for and reject negative values.
* I believe `timeout` [should be of type `DWORD`](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-setsockopt), which is unsigned.

crypto/bio/dgram.c
```diff
@@ -255,6 +272,116 @@ static long dgram_ctrl(BIO *bp, const int cmd, const long num, void *ptr) {
...
+      if (setsockopt(bp->num, SOL_SOCKET, SO_RCVTIMEO,
+                     (void *)&timeout, sizeof(timeout)) < 0) {
```
justsmth: On error, the [value returned on Windows](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-setsockopt#return-value) is `SOCKET_ERROR`. I'm not sure whether that's guaranteed to be negative.  I think we should just check for `!= 0`.

crypto/bio/dgram.c
smittals2: The synchronous errors we get from sendto or recvfrom are reflected in errno/WSAGetLastError (like EAGAIN, WASETIMEDOUT, etc.), while async connection based errors are what we get from SO_ERROR. OpenSSL also reflects errors in this way.

justsmth (approved):

include/openssl/bio.h
```diff
@@ -726,6 +726,10 @@ OPENSSL_EXPORT int BIO_get_md(BIO *b, EVP_MD **md);
 #define BIO_CTRL_DGRAM_CONNECT       31 // BIO dgram special
 #define BIO_CTRL_DGRAM_SET_CONNECTED 32 /* allow for an externally connected
                                           * socket to be passed in */
+#define BIO_CTRL_DGRAM_SET_RECV_TIMEOUT 33 /* setsockopt, essentially */
+#define BIO_CTRL_DGRAM_GET_RECV_TIMEOUT 34 /* getsockopt, essentially */
+#define BIO_CTRL_DGRAM_SET_SEND_TIMEOUT 35 /* setsockopt, essentially */
+#define BIO_CTRL_DGRAM_GET_SEND_TIMEOUT 36 /* getsockopt, essentially */
```
samuel40791765: Nit:
```suggestion
#define BIO_CTRL_DGRAM_SET_RECV_TIMEOUT 33 // setsockopt, essentially
#define BIO_CTRL_DGRAM_GET_RECV_TIMEOUT 34 // getsockopt, essentially
#define BIO_CTRL_DGRAM_SET_SEND_TIMEOUT 35 // setsockopt, essentially
#define BIO_CTRL_DGRAM_GET_SEND_TIMEOUT 36 // getsockopt, essentially
```

samuel40791765 (approved):

samuel40791765 (approved):

PR #2616, opened 2025-08-15
TLS Transfer Serialization Improvements

skmcgrail:
### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 

### Call-outs:

### Testing:


ssl/ssl_transfer_asn1.cc
```diff
@@ -717,20 +811,28 @@ static int SSL3_STATE_from_bytes(SSL *ssl, CBS *cbs, const SSL_CTX *ctx) {
...
+        pending_app_data_ptr + pending_app_data_size >
+            out->read_buffer.buf_ptr() + out->read_buffer.buf_size()) {
       OPENSSL_PUT_ERROR(SSL, SSL_R_SERIALIZATION_INVALID_SSL3_STATE);
       return 0;
     }
     out->pending_app_data =
-        MakeSpan(out->read_buffer.buf_ptr() + pending_app_data_offset,
-                 pending_app_data_size);
+        MakeSpan(pending_app_data_ptr, pending_app_data_size);
   }
```
skmcgrail: On a fresh re-read, I realize this weeds to special case the previous format version behavior from the new behavior.

ssl/ssl_lib.cc
```diff
@@ -1063,6 +1078,8 @@ static int ssl_read_impl(SSL *ssl) {
 }
 
 int SSL_read_ex(SSL *ssl, void *buf, size_t num, size_t *read_bytes) {
+  GUARD_SUSPENDED_STATE(ssl, -1);
```
samuel40791765: The `SSL_*_ex` functions and a few others here (ex: `SSL_key_update`) have a function contract that only returns `0` and `1`. I think we should tweak those accordingly, otherwise this could lead to issues with consumers checking with the return values as booleans.

ssl/ssl_transfer_asn1.cc
```diff
@@ -669,17 +707,40 @@ static int SSL3_STATE_from_bytes(SSL *ssl, CBS *cbs, const SSL_CTX *ctx) {
...
+         CBS_len(&read_traffic_secret) != read_traffic_secret_len) ||
+        (serde_version < SSL3_STATE_SERDE_VERSION_THREE &&
+         CBS_len(&read_traffic_secret) != SSL_MAX_MD_SIZE) ||
+        !exporter_secret_present || !exporter_secret_len ||
+        exporter_secret_len > UINT8_MAX ||
+        exporter_secret_len > SSL_MAX_MD_SIZE ||
+        (serde_version >= SSL3_STATE_SERDE_VERSION_THREE &&
+         CBS_len(&exporter_secret) != exporter_secret_len) ||
+        (serde_version < SSL3_STATE_SERDE_VERSION_THREE &&
+         CBS_len(&exporter_secret) != SSL_MAX_MD_SIZE) ||
```
samuel40791765, line 738: Wish there was a easier way to parse the divergent logic from `>= SSL3_STATE_SERDE_VERSION_THREE` and `< SSL3_STATE_SERDE_VERSION_THREE`, but I can't think of any better options other than using multiple if statements instead. Not a blocker though.

ssl/internal.h
```diff
@@ -4237,6 +4262,10 @@ struct ssl_st {
   // as will fit in the SSLBuffer from the BIO, or just enough to read the record
   // header and then the length of the body
   bool enable_read_ahead : 1;
+
+  // is_suspended_state indicates that the |SSL| object has been serialized and
+  // operations should not be performed on the connection.
+  bool is_suspended_state : 1;
```
samuel40791765, line 4270: Should the default of this be `false`?

ssl/ssl_buffer.cc
```diff
@@ -192,13 +276,190 @@ bool SSLBuffer::DoDeserialization(CBS *cbs) {
...
+  return x >= PTRDIFF_MIN && x <= PTRDIFF_MAX;
+#else
+  // ptrdiff_t is narrower than int64_t — cast to int64_t for safe compare
+  return x >= (int64_t)PTRDIFF_MIN && x <= (int64_t)PTRDIFF_MAX;
+#endif
+}
+
+static bool deserialize_buffer_view_from_data_offset(CBS &cbs,
+                                                     SSLBuffer *buffer,
+                                                     Span<uint8_t> &view) {
```
samuel40791765, line 490: Did we consider consolidating this with `deserialize_buffer_view_from_buf_ptr_offset` and just passing `kBufferViewOffsetFromDataPtr` as the parameter?

ssl/ssl_buffer.cc
```diff
@@ -192,13 +276,190 @@ bool SSLBuffer::DoDeserialization(CBS *cbs) {
...
+      !CBS_get_asn1_uint64(&cbs, &cap) ||
+      !CBS_get_asn1(&cbs, &buf, CBS_ASN1_OCTETSTRING) || CBS_len(&cbs) != 0) {
+    OPENSSL_PUT_ERROR(SSL, SSL_R_SERIALIZATION_INVALID_SSL_BUFFER);
+    return false;
+  }
+
+  bool buf_allocated = !!buf_allocated_int;
+  if (buf_allocated) {
+    buf_size_ = compute_buffer_size(buf_cap);
+    buf_ = (uint8_t *)malloc(buf_size_);
```
samuel40791765, line 353:
```suggestion
    buf_ = (uint8_t *)OPENSSL_malloc(buf_size_);
```

I understand that this might be due to align with previous decisions to use `malloc/free` here. Should we replace those as well?

ssl/internal.h
skmcgrail: I believe the `: 1` is a bit field semantic no? This just indicates that this will be just a 1 bit field, and does not represent the default.

ssl/internal.h
```diff
@@ -1432,28 +1439,46 @@ class SSLBuffer {
...
   // inline_buf_ is a static buffer for short reads.
   uint8_t inline_buf_[SSL3_RT_HEADER_LENGTH];
-  // buf_allocated_ is true if |buf_| points to allocated data and must be freed
-  // or false if it points into |inline_buf_|.
-  bool buf_allocated_ = false;
-  // buf_size_ is how much memory allocated for |buf_|. This is needed by
-  // |DoSerialization|.
-  size_t buf_size_ = 0;
+    
+  // The V1 version has some intricacies were solved in later serialization versions.
```
justsmth, line 1472: NP: "... has some intricacies THAT were solved ..."

ssl/ssl_buffer.cc
justsmth: We must always trust these inputs. Were an adversary able to manipulate it, they could quickly bring down a server.

Should we do some sanity checking? Possibly some debug assert statements around the values we parse?

ssl/ssl_buffer.cc
```diff
@@ -125,51 +139,121 @@ void SSLBuffer::DiscardConsumed() {
...
+      !CBB_add_asn1_uint64(&seq, header_len_) ||
+      !CBB_add_asn1_uint64(&seq, rel_offset) ||
       !CBB_add_asn1_uint64(&seq, cap_) ||
-      (buf_allocated_ && !CBB_add_asn1_octet_string(&seq, buf_, buf_size_)) ||
-      (!buf_allocated_ && !CBB_add_asn1_octet_string(&seq, inline_buf_, SSL3_RT_HEADER_LENGTH))) {
+      (buf_allocated_ && !CBB_add_asn1_octet_string(&seq, buf_ + align_offset,
+                                                    rel_offset + size_)) ||
+      (!buf_allocated_ &&
+       !CBB_add_asn1_octet_string(&seq, inline_buf_ + align_offset,
+                                  rel_offset + size_))) {
```
justsmth, line 194: Since `inline_buf_` is a fixed size, I think it would be more clear to do the same as V1 here:
```
(!buf_allocated_ &&
   !CBB_add_asn1_octet_string(&seq, inline_buf_, SSL3_RT_HEADER_LENGTH))) {
...
```

ssl/ssl_buffer.cc
```diff
@@ -192,13 +276,190 @@ bool SSLBuffer::DoDeserialization(CBS *cbs) {
...
+  max_serialization_version_ = SSL_BUFFER_SERDE_VERSION_TWO;
+  return true;
+}
+
+static const unsigned kBufferViewOffsetFromDataPtr =
+    CBS_ASN1_CONSTRUCTED | CBS_ASN1_CONTEXT_SPECIFIC | 0;
+
+bool SSLBuffer::SerializeBufferView(CBB &cbb, Span<uint8_t> &view) {
+  // View must be a span that points into this buffer.
+  if (!buf_ptr() || buf_ptr() > view.data()) {
```
justsmth: * Add more constraints?
```
  if (!buf_ptr() || !view.data() || buf_ptr() > view.data() || view.data() > (buf_ptr() + buf_size())) {
```
* Should some of these checks be in `assert` statements?

ssl/ssl_buffer.cc
```diff
@@ -192,13 +276,190 @@ bool SSLBuffer::DoDeserialization(CBS *cbs) {
...
+}
+
+static int fits_in_ptrdiff_int64(int64_t x) {
+#if PTRDIFF_MAX > INT64_MAX
+  // ptrdiff_t is wider than int64_t — all int64_t fit
+  (void)x;
+  return 1;
+#elif PTRDIFF_MAX == INT64_MAX
+  // ptrdiff_t is exactly int64_t — no cast needed
+  return x >= PTRDIFF_MIN && x <= PTRDIFF_MAX;
```
justsmth: We know `PTRDIFF_MAX == INT64_MAX` -- so is this check even needed?

ssl/ssl_buffer.cc
skmcgrail: Next push includes some improvements and fuzzing to harden and sanity check the final constructed buffer.

ssl/ssl_buffer.cc
skmcgrail: This intentionally avoids doing the whole buffer to avoid serializing uninitialized memory.

ssl/ssl_buffer.cc
skmcgrail: This matches the way `EnsureCap` works today, and it seems to have been done to avoid the penalty of zeroing the buffer. I think it makes sense to treat it as a separate exploration as it would require performance evaluation.

ssl/ssl_buffer.cc
skmcgrail: The anchor point of the offset contained in the view is completely different, which is the reason for the separation.

ssl/ssl_buffer.cc
```diff
@@ -125,54 +139,145 @@ void SSLBuffer::DiscardConsumed() {
...
-  if (!cbs) {
+bool SSLBuffer::ValidateBuffersState() {
+  const uint8_t *start_ptr = buf_;
+  const uint8_t *end_ptr = buf_ + buf_size_;
+  const uint8_t *data_start_ = buf_ + offset_;
+  const uint8_t *data_end_ptr = data_start_ + size_;
+  const uint8_t *remaining_ptr = data_end_ptr + (cap_ - size_);
+  if (data_start_ > end_ptr || data_start_ < start_ptr ||
+      data_end_ptr > end_ptr || data_end_ptr < start_ptr || size_ > cap_ ||
+      remaining_ptr > end_ptr || remaining_ptr < start_ptr) {
```
justsmth, line 235: Can any of these potentially be equal?

ssl/ssl_lib.cc
```diff
@@ -168,6 +168,14 @@
 
 BSSL_NAMESPACE_BEGIN
 
+#define GUARD_SUSPENDED_STATE(ptr,code)                         \
+  do {                                                           \
+    if (ptr->is_suspended_state) {                               \
+      OPENSSL_PUT_ERROR(SSL, ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED); \
+      return code;                                               \
+    }                                                            \
+  } while (0)
```
justsmth, line 177: NP: Use parentheses: `(ptr)->is_suspended_state`.

PR #2619, opened 2025-08-15
Add x86 Keccak implementation

manastasova:
### Description of changes: 
Add x86 optimized Keccak implementation from s2n-bignum

This PR integrates optimized and formally verified x86 Keccak code from the s2n-bignum library. The code has been imported as-is using our importer script, adding all changes up to commit 717b57ac643327489d2ac7dd4022a64b5dfb2d8f.

Key aspects:
- Adds hardware-accelerated Keccak implementation for x86 platforms
- No changes made to the imported code
- Enables SHA3/SHAKE to use Keccak assembly on x86 (with UNIX) platforms

### Testing:
 ninja && ./crypto/crypto_test
./tool/bssl speed -filter {SHA3-224, ...}

# SHA3 Performance Comparison

## ASM Implementation vs C Implementation
| Algorithm | Input Size | ASM Implementation |  | C Implementation |  | Speedup |
|-----------|------------|-------------------|--|-----------------|--|---------|
|           |            | ops/sec | MB/s    | ops/sec | MB/s    | % |
| **SHA3-224**  | **16 bytes**   | **3,367,963.0** | **53.9** | **3,084,636.0** | **49.4** | **9.2%** |
|           | 256 bytes  | 1,748,043.8 | 447.5 | 1,629,982.1 | 417.3 | 7.2% |
|           | 1350 bytes | 363,247.7 | 490.4 | 343,366.1 | 463.5 | 5.8% |
|           | 8192 bytes | 63,937.3 | 523.8 | 60,643.4 | 496.8 | 5.4% |
|           | 16384 bytes | 31,957.9 | 523.6 | 30,315.5 | 496.7 | 5.4% |
| **SHA3-256** | **16 bytes**   | **3,400,705.8** | **54.4** | **3,106,263.8** | **49.7** | **9.5%** |
|           | 256 bytes  | 1,762,683.0 | 451.2 | 1,634,490.0 | 418.4 | 7.8% |
|           | 1350 bytes | 365,010.1 | 492.8 | 342,906.0 | 462.9 | 6.4% |
|           | 8192 bytes | 60,317.3 | 494.1 | 56,481.2 | 462.7 | 6.8% |
|           | 16384 bytes | 30,437.0 | 498.7 | 28,552.4 | 467.8 | 6.6% |
| **SHA3-384** | **16 bytes**   | **3,416,610.5** | **54.7** | **3,103,572.4** | **49.7** | **10.1%** |
[...]

PR #2628, opened 2025-08-19
Remove BIT_INTERLEAVE support

manastasova:
Bit Interleave is used for performance optimizations on 32-bit platforms. Bit Interleave adds unnecessary complexity.

### Issues:
Some Windows compiler, e.g., old versions of Microsoft Visual C++ (MSVC), do not support some preprocessor directives and expressions, e.g., of the type:

```
// Double-check that bit-interleaving is not used on AArch64
#if BIT_INTERLEAVE != 0
#error Bit-interleaving of Keccak1600 states should be disabled for AArch64
#endif
```

in https://github.com/aws/aws-lc/blob/d781046a99638d1466ec912cf0191d0564de2084/crypto/fipsmodule/sha/keccak1600.c#L422

A solution could be:

```
#if defined(BIT_INTERLEAVE) && BIT_INTERLEAVE
  #error Bit-interleaving of Keccak1600 states should be disabled for AArch64
#endif
```

However, BIT_INTERLEAVE is intended for only optimizing 32-bit platforms, i.e., it adds unnecessary complexity to the code without providing many benefits. 

Therefore, removing BIT_INTERLEAVE support is the better solution for clarity and maintainability.

### Description of changes: 
Remove all support for BIT_INTERLEAVE.

### Call-outs:
This change is needed/motivated by the integration of x86 Keccak to aws-lc https://github.com/aws/aws-lc/pull/2619 which fails when running on x86 Windows platform.

### DIFF
--- a/crypto/fipsmodule/sha/keccak1600.c
+++ b/crypto/fipsmodule/sha/keccak1600.c
@@ -12,43 +12,31 @@
 #include "../../internal.h"
 #include "../cpucap/internal.h"
 
-#if defined(__x86_64__) || defined(__aarch64__) || \
-    defined(__mips64) || defined(__ia64) || defined(__loongarch_lp64) || \
-    (defined(__VMS) && !defined(__vax))
-
- // These are available even in ILP32 flavours, but even then they are
- // capable of performing 64-bit operations as efficiently as in *P64.
- // Since it's not given that we can use sizeof(void *), just shunt it.
-# define BIT_INTERLEAVE (0)
-#else
-# define BIT_INTERLEAVE (sizeof(void *) < 8)
-#endif
-
 static const uint64_t iotas[] = {
-    BIT_INTERLEAVE ? 0x0000000000000001ULL : 0x0000000000000001ULL,
-    BIT_INTERLEAVE ? 0x0000008900000000ULL : 0x0000000000008082ULL,
-    BIT_INTERLEAVE ? 0x8000008b00000000ULL : 0x800000000000808aULL,
-    BIT_INTERLEAVE ? 0x8000808000000000ULL : 0x8000000080008000ULL,
-    BIT_INTERLEAVE ? 0x0000008b00000001ULL : 0x000000000000808bULL,
-    BIT_INTERLEAVE ? 0x0000800000000001ULL : 0x0000000080000001ULL,
-    BIT_INTERLEAVE ? 0x8000808800000001ULL : 0x8000000080008081ULL,
-    BIT_INTERLEAVE ? 0x8000008200000001ULL : 0x8000000000008009ULL,
-    BIT_INTERLEAVE ? 0x0000000b00000000ULL : 0x000000000000008aULL,
-    BIT_INTERLEAVE ? 0x0000000a00000000ULL : 0x0000000000000088ULL,
-    BIT_INTERLEAVE ? 0x0000808200000001ULL : 0x0000000080008009ULL,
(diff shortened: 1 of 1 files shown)


dkostic (approved):

hanno-becker:
@manastasova I am a bit hesitant removing this -- this is an important optimization for 32-bit platforms. Why not use your suggestion
```c
#if defined(BIT_INTERLEAVE) && BIT_INTERLEAVE
  #error Bit-interleaving of Keccak1600 states should be disabled for AArch64
#endif
```
to work around the issue? Or is there more?

manastasova:
Some data on how the performance is impacted when compiled for 32-bit platforms _but ran on 64-bit machine_. (x86 c7i.2xlarge EC2 instance  using the https://github.com/aws/aws-lc/blob/main/util/32-bit-toolchain.cmake toolchain to build).

# SHA3 Performance Comparison: No Bit Interleave vs. With Bit Interleave

## SHA3-224

| Input Size | No Bit Interleave (MB/s) | With Bit Interleave (MB/s) | Improvement (%) |
|------------|--------------------------|----------------------------|-----------------|
| 16 bytes   | 4.7                      | 3.4                        | 38.2%           |
| 256 bytes  | 38.9                     | 28.0                       | 39.0%           |
| 1350 bytes | 42.2                     | 29.6                       | 42.6%           |
| 8192 bytes | 45.2                     | 31.8                       | 42.1%           |
| 16384 bytes| 45.2                     | 31.7                       | 42.6%           |

## SHA3-256

| Input Size | No Bit Interleave (MB/s) | With Bit Interleave (MB/s) | Improvement (%) |
|------------|--------------------------|----------------------------|-----------------|
| 16 bytes   | 4.8                      | 3.4                        | 41.2%           |
| 256 bytes  | 39.4                     | 27.7                       | 42.2%           |
| 1350 bytes | 42.7                     | 30.0                       | 42.3%           |
| 8192 bytes | 42.8                     | 30.0                       | 42.7%           |
| 16384 bytes| 43.3                     | 29.8                       | 45.3%           |

## SHA3-384

| Input Size | No Bit Interleave (MB/s) | With Bit Interleave (MB/s) | Improvement (%) |
|------------|--------------------------|----------------------------|-----------------|
| 16 bytes   | 4.8                      | 3.4                        | 41.2%           |
| 256 bytes  | 26.7                     | 18.8                       | 42.0%           |
| 1350 bytes | 33.2                     | 23.4                       | 41.9%           |
| 8192 bytes | 33.5                     | 23.3                       | 43.8%           |
| 16384 bytes| 33.4                     | 23.3                       | 43.3%           |

## SHA3-512

| Input Size | No Bit Interleave (MB/s) | With Bit Interleave (MB/s) | Improvement (%) |
|------------|--------------------------|----------------------------|-----------------|
| 16 bytes   | 4.8                      | 3.4                        | 41.2%           |
| 256 bytes  | 20.4                     | 14.4                       | 41.7%           |
| 1350 bytes | 23.0                     | 16.2                       | 42.0%           |
| 8192 bytes | 23.1                     | 16.3                       | 41.7%           |
| 16384 bytes| 23.2                     | 16.3                       | 42.3%           |

hanno-becker:
Ok, I disagree and commit.

manastasova:
@hanno-becker do you have some performance data? 
I don't mind either keeping or removing the bit interleave support but it will be good to know what is the actual performance gain if we keep the code more complex. (I was only able to measure on a 64-bit platform where bit interleave shows worse performance).

PR #2638, opened 2025-08-25
Implement dgst CLI command

nhatnghiho:
### Issues:
Resolves #CryptoAlg-3383

### Description of changes: 
Re-implement dgst cli commands with the following options:
-binary
-sha256 (and a handful of other digest algorithms)
-sigopt
-passin
-sign
-out
-verify
-signature
-keyform

### Call-outs:
Waiting for https://github.com/aws/aws-lc/pull/2555 to implement `-passin`

### Testing:

### DIFF
--- a/tool-openssl/dgst.cc
+++ b/tool-openssl/dgst.cc
@@ -7,214 +7,534 @@
 
 #include <openssl/evp.h>
 #include <openssl/hmac.h>
+#include <openssl/pem.h>
+
 #include "internal.h"
 
 // MD5 command currently only supports stdin
 static const argument_t kArguments[] = {
     {"-help", kBooleanArgument, "Display option summary"},
+    {"-binary", kBooleanArgument, "Output in binary form"},
+    {"-hex", kBooleanArgument, "Output as hex dump"},
+    {"-md5", kExclusiveBooleanArgument, "Supported digest function"},
+    {"-ripemd160", kExclusiveBooleanArgument, "Supported digest function"},
+    {"-sha1", kExclusiveBooleanArgument, "Supported digest function"},
+    {"-sha256", kExclusiveBooleanArgument,
+     "Supported digest function (default)"},
+    {"-sha224", kExclusiveBooleanArgument, "Supported digest function"},
+    {"-sha384", kExclusiveBooleanArgument, "Supported digest function"},
+    {"-sha512", kExclusiveBooleanArgument, "Supported digest function"},
     {"-hmac", kOptionalArgument,
      "Create a hashed MAC with the corresponding key"},
+    {"-sigopt", kOptionalArgument, "Signature parameter in n:v form"},
+    {"-passin", kOptionalArgument, "Input file pass phrase source"},
+    {"-sign", kOptionalArgument, "Sign digest using private key"},
+    {"-out", kOptionalArgument, "Output to filename rather than stdout"},
+    {"-verify", kOptionalArgument, "Verify a signature using public key"},
+    {"-signature", kOptionalArgument, "File with signature to verify"},
+    {"-keyform", kOptionalArgument, "key file format (only PEM is supported)"},
     {"", kOptionalArgument, ""}};
 
-static bool dgst_file_op(const std::string &filename, const int fd,
-                         const EVP_MD *digest) {
-  static const size_t kBufSize = 8192;
-  std::unique_ptr<uint8_t[]> buf(new uint8_t[kBufSize]);
+static bool LoadPrivateKey(const std::string &key_file_path,
+                           bssl::UniquePtr<EVP_PKEY> &pkey) {
+  ScopedFILE key_file(fopen(key_file_path.c_str(), "rb"));
 
-  bssl::ScopedEVP_MD_CTX ctx;
-  if (!EVP_DigestInit_ex(ctx.get(), digest, nullptr)) {
-    fprintf(stderr, "Failed to initialize EVP_MD_CTX.\n");
+  if (!key_file) {
+    fprintf(stderr, "Failed to open %s", key_file_path.c_str());
     return false;
   }
 
-  for (;;) {
-    size_t n;
-    if (!ReadFromFD(fd, &n, buf.get(), kBufSize)) {
-      fprintf(stderr, "Failed to read from %s: %s\n", filename.c_str(),
-              strerror(errno));
-      return false;
-    }
+  // TODO: use -passin to read the key, if applicable
+  const char *password = nullptr;
 
-    if (n == 0) {
-      break;
-    }
+  pkey.reset(PEM_read_PrivateKey(key_file.get(), nullptr, nullptr,
+                                 const_cast<char *>(password)));
 
-    if (!EVP_DigestUpdate(ctx.get(), buf.get(), n)) {
-      fprintf(stderr, "Failed to update hash.\n");
-      return false;
-    }
-  }
-
-  uint8_t hash[EVP_MAX_MD_SIZE];
-  unsigned hash_len;
-  if (!EVP_DigestFinal_ex(ctx.get(), hash, &hash_len)) {
-    fprintf(stderr, "Failed to finish hash.\n");
+  if (!pkey) {
+    fprintf(stderr, "Failed to read private key from %s",
+            key_file_path.c_str());
     return false;
   }
 
-  // Print digest output. OpenSSL outputs the digest name with files, but not
-  // with stdin.
-  if (fd != 0) {
-    fprintf(stdout, "%s(%s)= ", EVP_MD_get0_name(digest), filename.c_str());
-  } else {
-    fprintf(stdout, "(%s)= ", filename.c_str());
-  };
-  for (size_t i = 0; i < hash_len; i++) {
-    fprintf(stdout, "%02x", hash[i]);
-  }
-  fprintf(stdout, "\n");
   return true;
 }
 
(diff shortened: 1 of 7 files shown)


tool-openssl/dgst_test.cc
```diff
@@ -76,125 +117,275 @@ TEST_F(DgstComparisonTest, HMAC_default_files) {
...
+}
+
+TEST_F(DgstTest, DigestSHA256) {
+  std::ofstream ofs(in_path);
+  ofs << "test data";
+  ofs.close();
+
+  args_list_t args = {"-sha256", in_path};
+  EXPECT_TRUE(dgstTool(args));
+}
```
justsmth: The "msys2 ucrt64" test seems to be [failing here](https://github.com/aws/aws-lc/actions/runs/17251362461/job/48954136964?pr=2638#step:6:1307). I suspect it might be related to how `ofstream` is being used?

```
[----------] 6 tests from DgstTest
[ RUN      ] DgstTest.HMAC
HMAC-SHA256(C:\Users\RUNNER~1\AppData\Local\Temp\aws9953.tmp)=a77e9db4c923bb8da10106a492db17a1c792d7c09a860004e57abb7fb8b0e784
[       OK ] DgstTest.HMAC (140 ms)
[ RUN      ] DgstTest.Sign
[       OK ] DgstTest.Sign (119 ms)
[ RUN      ] DgstTest.Verify
Verified OK
[       OK ] DgstTest.Verify (125 ms)
[ RUN      ] DgstTest.DigestSHA256
tool-openssl/tool_openssl_test
tool-openssl/tool_openssl_test failed to complete: exit status 0xc0000005

1 of 26 tests failed:
	tool-openssl/tool_openssl_test
```

justsmth: I believe we've seen this before -- it's a runtime issue with, e.g., required DLLs not being found. 

Unfortunately, the easiest solution has been to use BIO or libc functions, instead of C++ standard library, for I/O operations in these tests. :-(

tool-openssl/dgst.cc
```diff
@@ -7,214 +7,534 @@
...
+  if (in_path.empty()) {
+    in_file.reset(stdin);
+  } else {
+    in_file.reset(fopen(in_path.c_str(), "rb"));
+    if (!in_file) {
+      fprintf(stderr, "Error: unable to open input file '%s'\n",
+              in_path.c_str());
       return false;
     }
+  }
```
justsmth: I see these same 11 lines multiple places in this file. Should this be its own function?  Or perhaps the caller should prepare the file?

tool-openssl/dgst.cc
```diff
@@ -7,214 +7,534 @@
...
+    fprintf(stderr, "Error opening signature file %s.\n",
+            signature_file_path.c_str());
+    return false;
+  }
+
+  if (!ReadAll(&signature, signature_file.get())) {
+    fprintf(stderr, "Error reading from signature file %s.\n",
+            signature_file_path.c_str());
+    return false;
+  }
```
justsmth, line 305: NP: I think we should open the signature file (to verify that we actually have a signature to verify) prior to reading through the entire input.

tool-openssl/dgst.cc
```diff
@@ -7,214 +7,534 @@
...
-      // Empty input. OpenSSL continues processing the next file even when
-      // provided an invalid file.
-      fprintf(stderr, "Failed to read from empty input.");
+      BIO_printf(out_bio.get(), "(stdin)=");
+    }
+
+    for (size_t i = 0; i < data.size(); i++) {
+      BIO_printf(out_bio.get(), "%02x", data[i]);
     }
+    BIO_printf(out_bio.get(), "\n");
```
justsmth: Could you just use `BIO_hexdump` here?

tool-openssl/dgst.cc
```diff
@@ -7,214 +7,534 @@
...
+  // Validate arguments
+  if (HasArgument(parsed_args, "-hex") && HasArgument(parsed_args, "-binary")) {
+    fprintf(stderr, "Error: -hex and -binary cannot both be specified");
+    return false;
+  }
+
+  if ((!verify_key_file.empty() + !sign_key_file.empty() + !hmac.empty()) > 1) {
+    fprintf(stderr, "Error: MAC and signing key cannot both be specified\n");
+    return false;
+  }
```
justsmth, line 391: NP: It might also be that `sign_key_file` and `verify_key_file` are not empty.

tool-openssl/dgst.cc
```diff
@@ -7,214 +7,534 @@
...
+      fprintf(stderr, "Error reading from '%s'.\n", in_path.c_str());
+      return false;
+    }
+
+    if (!EVP_DigestVerifyUpdate(ctx.get(), buf, n)) {
+      fprintf(stderr, "Failed to update signature.\n");
+      return false;
+    }
+  }
+  std::vector<uint8_t> signature(EVP_PKEY_size(pkey));
```
justsmth: NP: Sizing the vector is not needed. `ReadAll` will size the vector based on the contents read.

tool-openssl/dgst.cc
```diff
@@ -7,214 +7,534 @@
...
+
+  if (HasArgument(parsed_args, "-help")) {
+    PrintUsage(kArguments);
+    return true;
+  }
+
+  GetBoolArgument(&binary, "-binary", parsed_args);
+  GetBoolArgument(&hex, "-hex", parsed_args);
+  GetString(&hmac, "-hmac", "", parsed_args);
+  GetString(&passin, "-passin", "", parsed_args);
```
justsmth, line 373: Since we don't support `-passin`, we shouldn't advertise it as an option.

tool-openssl/dgst.cc
```diff
@@ -7,214 +7,534 @@
...
+
+  // Default digest is SHA-256.
+  if (digest == nullptr) {
+    GetExclusiveBoolArgument(&digest_name, kArguments, "-sha256", parsed_args);
+    digest = EVP_get_digestbyname(digest_name.substr(1).c_str());
+  }
+
+  if (hex) {
+    out_bin = false;
+  } else if (binary || !sign_key_file.empty() || !verify_key_file.empty()) {
```
justsmth, line 413: NP: `out_bin` defaults  to false, so the `if (hex)` is not needed.

tool-openssl/dgst.cc
```diff
@@ -7,214 +7,534 @@
...
+      std::vector<uint8_t> mac(EVP_MD_size(digest));
+
+      if (!GenerateHMAC(in_path, hmac_key, hmac_key_len, digest, mac)) {
         return false;
       }
-    } else {
-      if (!dgst_file_op(file_name, fd, digest)) {
+
+      WriteOutput(out_bio, mac, "HMAC", EVP_MD_get0_name(digest), in_path,
+                  out_bin);
```
justsmth: Shouldn't our output match that of openssl?
```
❯ echo "hello world" | openssl dgst -sha256 -hmac "secret_key"
SHA2-256(stdin)= 074c3f9a7b27557c0996d5b1ef9e66ee9d7341870d23e85b62980d3418352471
```

```
❯ echo "hello world" | ~/lib/AWS-LC-DEBUG-INSTALL/bin/openssl dgst -sha256 -hmac "secret_key"
HMAC-SHA256(stdin)=f304c11274cabc93abe79f3abb848b94026d3e1c9a49071ea96fbece9b9f2bb0
```

tool-openssl/dgst.cc
nhatnghiho: No unfortunately. We need a raw hex dump but `BIO_hexdump` has a special formatting that looks something like this -- `OFFSET HEX_BYTES |ASCII|`

tool-openssl/dgst.cc
nhatnghiho: I was trying to fix a clang tidy warning before opening the PR, which led to the mismatch. I fixed that and the output should match now

tool-openssl/dgst_test.cc
nhatnghiho: Should be fixed now

justsmth (approved):

tool-openssl/internal.h
```diff
@@ -124,6 +134,10 @@ bool GetString(std::string *out, const std::string &arg_name,
                std::string default_value, const ordered_args_map_t &args);
 bool GetBoolArgument(bool *out, const std::string &arg_name,
                      const ordered_args_map_t &args);
+
+bool GetExclusiveBoolArgument(std::string *out_arg, const argument_t *templates,
```
smittals2, line 138: just curious on reasoning here. Why do we want to do exclusive bool arguments at the parsing layer, as opposed to for example writing a func at tool level to validate parsed options?

smittals2 (approved):

tool-openssl/internal.h
nhatnghiho: Since the supported digest suite is fairly large, I think that doing it the parsing layer makes more sense. Otherwise, we'd need to 1) maintain a list of supported digests (in addition to the arg templates), 2) parse the arguments a 2nd time to get the digest names, and 3) check for mutual exclusivity. Computationally, it's the same amount of work either way but doing it at the parsing layer is much cleaner. 

And since we're already relying on the `kExclusiveBooleanArgument` type for the parsing, might as well make it a util function rather than a one-off function in the digest tool. I thought it'd be useful in the future if we ever need to do handle multiple exclusive bool args again.

PR #2642, opened 2025-08-26
Support for explicit curve parameter on EC public keys where parameters match supported curves

skmcgrail:
### Description of changes: 

This feature is supported for private keys already, but EC public key types did not support explicitly encoded curve parameters. This updates the logic to support such types to use the same logic as EC private keys, meaning that we will only accept explicit curve parameters that match the named curve types we support. This does mean that subject public keys in X.509 certificates, or that use ASN.1 encoding will now support explicit parameters as well. This was a feature that OpenSSL 1.1.1 supported, but OpenSSL 3+ now will fail certificate validation on, due to this being a non RFC conforming behavior and shouldn't happen in publicly issued certificate by conforming certificate authorities. We are adding this only for compatability with the former. I am still investigating separately whether to disable / reject such keys in libssl as to not open the door to much, and possibly whether we should add an X509 verification flag to disable this behavior as well. Will follow that up in a separate PR.


crypto/ecdh_extra/ecdh_test.cc
```diff
@@ -261,14 +261,25 @@ static void RunWycheproofTest(FileTest *t) {
...
   // This is non-standard and error-prone.
   CBS cbs;
   CBS_init(&cbs, peer_spki.data(), peer_spki.size());
   bssl::UniquePtr<EVP_PKEY> peer_evp(EVP_parse_public_key(&cbs));
   if (!peer_evp || CBS_len(&cbs) != 0) {
+    if (!peer_evp && result.raw_result == WycheproofRawResult::kAcceptable &&
+        std::find(result.flags.begin(), result.flags.end(), "UnusedParam") !=
+            result.flags.end()) {
+      // The proof flags are not granular enough, and we do validate some
+      // parameters for unnamed / explicit curves.
```
justsmth, line 276: I believe this could make the stronger statement: "we validate ALL parameters". (Are there any parameters that go unchecked?) 

The key observation for this is that `EC_KEY_parse_parameters` can only return one of the 5 curves we support: `EC_group_p224`, `EC_group_p256`, `EC_group_p384`, `EC_group_p521`, `EC_group_secp256k1` -- it iterates over those supported curves, looking for an exact match.

justsmth (approved):

crypto/ecdh_extra/ecdh_test.cc
samuel40791765: I would hesitate to say we validate ALL parameters. `EC_KEY_parse_parameters` only uses the OID if the OID is of a known curve.
https://github.com/aws/aws-lc/blob/6d2eb62ba375ebba7ab20ab277332f5bff9e13f0/crypto/ec_extra/ec_asn1.c#L368-L370

I was tweaking this function a while back, but I believe the parameter check is only a "fall back" mechanism for backwards compatibility with OpenSSL. Some older versions of OpenSSL were exporting the EC parameters of known curves as custom/explicit curves. It isn't inherently "wrong", but the ideal behavior should be just directly checking the OID to map to the relevant known curve in most cases.

samuel40791765 (approved):

crypto/ecdh_extra/ecdh_test.cc
justsmth: The next tag in the ASN1 is different when it's a named curve. Can it contain both the curve name and the parameters?  I thought the parsing would fail if it had both -- and this comment is specifically referring to "unnamed / explicit" curves.

PR #2644, opened 2025-08-26
Migrate from CodeBuild account actor filter to pull request comment filter based on GitHub permissions

andrewhop:
### Issues:
Addresses P292389560

### Description of changes: 
Instead of managing an opaque list of GitHub account IDs this migrates the CI to he new [CodeBuild pull request comment approval](https://docs.aws.amazon.com/codebuild/latest/userguide/pull-request-build-policy.html). Any user that has admin, write, triage, or maintainer permission to the repository will have the CI run automatically when they open a PR. If any other user opens a pull request the CodeBuild jobs will not run until someone in the above permission list comments on the PR with `codebuild_run(<SHA_OF_THE_LATEST_COMMIT>)` e.g. `codebuild_run(0faf56fee7b72f842408930bfbdab9e90e4d174d)`.

### Call-outs:
As far as I can tell this isn't modeled in the CDK yet which requires manually configuring the CFN property on each job. This also isn't a property of `codebuild.Source.git_hub` but is a property of the overall [CFN Project object](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_codebuild/CfnProject.html#aws_cdk.aws_codebuild.CfnProject.PullRequestBuildPolicyProperty) which means we have to configure it on every project we create.  

### Testing:
Verified locally the CloudFormation template looks like what I expect:
```
      Source:
        BuildSpec: tests/ci/cdk/cdk/codebuild/github_ci_linux_x86_omnibus.yaml
        Location: https://github.com/aws/aws-lc.git
        ReportBuildStatus: true
        Type: GITHUB
        PullRequestBuildPolicy:
          requiresCommentApproval: ALL_PULL_REQUESTS
          approverRoles:
            - GITHUB_ADMIN
            - GITHUB_MAINTAINER
            - GITHUB_WRITE
            - GITHUB_TRIAGE
```

I am going to verify this change in pre-prod before merging this. 

### DIFF
--- a/tests/ci/cdk/cdk/aws_lc_android_ci_stack.py
+++ b/tests/ci/cdk/cdk/aws_lc_android_ci_stack.py
@@ -77,6 +77,9 @@ class AwsLcAndroidCIStack(AwsLcBaseCiStack):
             ),
             build_spec=BuildSpecLoader.load(spec_file_path, env),
         )
+        cfn_project = project.node.default_child
+        cfn_project.add_property_override("Source.PullRequestBuildPolicy", self.pull_request_policy)
+
         project.enable_batch_builds()
 
         PruneStaleGitHubBuilds(
--- a/tests/ci/cdk/cdk/aws_lc_android_ci_stack.py
+++ b/tests/ci/cdk/cdk/aws_lc_android_ci_stack.py
@@ -78,7 +78,7 @@ class AwsLcAndroidCIStack(AwsLcBaseCiStack):
             build_spec=BuildSpecLoader.load(spec_file_path, env),
         )
         cfn_project = project.node.default_child
-        cfn_project.add_property_override("Source.PullRequestBuildPolicy", self.pull_request_policy)
+        cfn_project.add_property_override("Triggers.PullRequestBuildPolicy", self.pull_request_policy)
 
         project.enable_batch_builds()
 
--- a/tests/ci/cdk/cdk/aws_lc_base_ci_stack.py
+++ b/tests/ci/cdk/cdk/aws_lc_base_ci_stack.py
@@ -57,12 +57,12 @@ class AwsLcBaseCiStack(Stack):
             webhook_triggers_batch_build=True,
         )
         # P292389560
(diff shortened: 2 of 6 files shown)


tests/ci/cdk/cdk/aws_lc_android_ci_stack.py
```diff
@@ -77,6 +77,9 @@ def __init__(
             ),
             build_spec=BuildSpecLoader.load(spec_file_path, env),
         )
+        cfn_project = project.node.default_child
+        cfn_project.add_property_override("Source.PullRequestBuildPolicy", self.pull_request_policy)
```
nhatnghiho: According to this [CFN example](https://docs.aws.amazon.com/codebuild/latest/userguide/pull-request-build-policy.html#pull-request-build-policy.cloudformation), I think the property path here should be "Triggers.PullRequestBuildPolicy"

Same applies to the same occurrences in other stacks.

tests/ci/cdk/cdk/aws_lc_base_ci_stack.py
```diff
@@ -49,16 +49,20 @@ def __init__(
...
+        # P292389560
+        self.pull_request_policy = codebuild.CfnProject.PullRequestBuildPolicyProperty(
+            requires_comment_approval="ALL_PULL_REQUESTS",
+            approver_roles=[
+                "GITHUB_ADMIN",
+                "GITHUB_MAINTAINER",
+                "GITHUB_WRITE",
+                "GITHUB_TRIAGE"
+            ]
+        )
```
nhatnghiho: I'm curious to see if this would work with the current CDK library version that we use, as a simple search within my local environment doesn't have `PullRequestBuildPolicyProperty`. I think we might need to either upgrade to a newer version of CDK library or use a hard-coded JSON structure like [here](https://github.com/aws/aws-lc/blob/main/tests/ci/cdk/pipeline/pipeline_stack.py#L203-L225)

samuel40791765 (approved):

PR #2647, opened 2025-08-28
Fix FIPS linux clang 20

justsmth:
### Context

The FIPS module build was failing to build with Clang 20 and later versions. 
* The original implementation attempted to compile `bcm.c` to assembly using CMake's `add_library()` with the `-S` compiler flag set via `set_target_properties()`.
* Newer Clang versions (20+) became stricter about conflicting compiler flags, warning when both `-S` (assembly output) and `-c` (object output) flags are present, and ultimately producing object files instead of assembly.

### Description of changes:
This change replaces the problematic `add_library()` target approach:
* Uses an `add_custom_command()` that invokes the compiler directly for "bcm.c" with the `-S` flag, ensuring assembly output is generated.
* Passes the generated `bcm.s` file directly to the `delocate` tool as a regular positional argument (not via the `-a` archive flag).

### DIFF
--- a/crypto/fipsmodule/CMakeLists.txt
+++ b/crypto/fipsmodule/CMakeLists.txt
@@ -397,31 +397,40 @@ if(FIPS_DELOCATE)
     enable_language(ASM)
   endif()
 
-  add_library(
-    bcm_c_generated_asm
-
-    STATIC
-
-    bcm.c
-  )
-  target_compile_definitions(bcm_c_generated_asm PRIVATE BORINGSSL_IMPLEMENTATION S2N_BN_HIDE_SYMBOLS)
-
-  # Important: We do not want to add the generated prefix symbols to the include path here!
-  # Delocator expects symbols to not be prefixed.
-  target_add_awslc_include_paths(TARGET bcm_c_generated_asm SCOPE PRIVATE EXCLUDE_PREFIX_HEADERS)
-  target_include_directories(bcm_c_generated_asm PRIVATE "${S2N_BIGNUM_INCLUDE_DIR}")
-  set_target_properties(bcm_c_generated_asm PROPERTIES COMPILE_OPTIONS "-S")
-  set_target_properties(bcm_c_generated_asm PROPERTIES POSITION_INDEPENDENT_CODE ON)
-
-  # Clang 20+ warns when both "-S" and "-c" are used as options to the compiler.
-  if (CLANG AND (CMAKE_ASM_COMPILER_ID MATCHES "Clang" OR CMAKE_ASM_COMPILER MATCHES "clang") AND (CMAKE_C_COMPILER_VERSION VERSION_GREATER "19.99.99"))
-    set_target_properties(bcm_c_generated_asm PROPERTIES COMPILE_OPTIONS "-Wno-unused-command-line-argument")
-  endif()
-
   set(TARGET "")
-  if(CMAKE_ASM_COMPILER_TARGET)
-    set(TARGET "--target=${CMAKE_ASM_COMPILER_TARGET}")
+  if(CMAKE_C_COMPILER_TARGET)
+    set(TARGET "--target=${CMAKE_C_COMPILER_TARGET}")
   endif()
+
+  # Listify the current arguments
+  separate_arguments(BCM_ALL_FLAGS UNIX_COMMAND "${CMAKE_C_FLAGS}")
+
+  # Compilation flags + defines + include dirs + target
+  list(APPEND BCM_ALL_FLAGS "-fPIC" "-DBORINGSSL_IMPLEMENTATION" "-DS2N_BN_HIDE_SYMBOLS" "-DBORINGSSL_FIPS")
+  list(APPEND BCM_ALL_FLAGS "-I${AWSLC_SOURCE_DIR}/include" "-I${S2N_BIGNUM_INCLUDE_DIR}")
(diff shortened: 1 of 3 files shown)


crypto/fipsmodule/CMakeLists.txt
```diff
@@ -394,32 +394,45 @@ if(FIPS_DELOCATE)
...
-  set(TARGET "")
-  if(CMAKE_ASM_COMPILER_TARGET)
-    set(TARGET "--target=${CMAKE_ASM_COMPILER_TARGET}")
+  if(CMAKE_C_STANDARD)
+    list(APPEND BCM_ALL_FLAGS "-std=c${CMAKE_C_STANDARD}")
+  endif()
+
+  if(CMAKE_C_STANDARD)
+    list(APPEND BCM_ALL_FLAGS "-std=c${CMAKE_C_STANDARD}")
+  endif()
```
nebeid: Is this redundant?

crypto/fipsmodule/CMakeLists.txt
```diff
@@ -394,32 +394,45 @@ if(FIPS_DELOCATE)
...
-
-  # Clang 20+ warns when both "-S" and "-c" are used as options to the compiler.
-  if (CLANG AND (CMAKE_ASM_COMPILER_ID MATCHES "Clang" OR CMAKE_ASM_COMPILER MATCHES "clang") AND (CMAKE_C_COMPILER_VERSION VERSION_GREATER "19.99.99"))
-    set_target_properties(bcm_c_generated_asm PROPERTIES COMPILE_OPTIONS "-Wno-unused-command-line-argument")
+  # Compilation flags + defines + include dirs + target
+  list(APPEND BCM_ALL_FLAGS "-fPIC" "-DBORINGSSL_IMPLEMENTATION" "-DS2N_BN_HIDE_SYMBOLS" "-DBORINGSSL_FIPS")
+  list(APPEND BCM_ALL_FLAGS "-I${AWSLC_SOURCE_DIR}/include" "-I${S2N_BIGNUM_INCLUDE_DIR}")
+  list(APPEND BCM_ALL_FLAGS "${TARGET}")
+  if (TEST_SYSGENID_PATH)
+    list(APPEND BCM_ALL_FLAGS "-DAWSLC_SNAPSAFE_TESTING=1" -DAWSLC_SYSGENID_PATH=\"${TEST_SYSGENID_PATH}\")
```
nebeid, line 413: Is this newly added and wasn't tested before?

util/fipstools/delocate/delocate.go
```diff
@@ -2529,41 +2525,21 @@ func parseInputs(inputs []inputFile, cppCommand []string) error {
...
-			}
-			defer arFile.Close()
+		var inBytes []byte
+		var err error
 
-			ar, err := ar.ParseAR(arFile)
-			if err != nil {
-				return err
-			}
+        // Assembly file with lower-case *.s suffix do not get preprocessed
```
nebeid, line 2531: Maybe we can be specific here and add that, specifically, we don't want to preprocess `bcm.s`, but I'm curious as to why? Was the delocator tripping when `bcm.s` was preprocessed?

util/fipstools/delocate/delocate.go
```diff
@@ -2666,7 +2630,7 @@ func main() {
 
 		inputs = append(inputs, inputFile{
 			path:  path,
-			index: i + 1,
+			index: i,
```
nebeid, line 2633: I'm curious about what this change does.

crypto/fipsmodule/CMakeLists.txt
justsmth: No. Because it's using `add_custom_command` to produce "bcm.s", many of the compiler options configured in the top-level CMakeLists.txt need to be specified here as well.

crypto/fipsmodule/CMakeLists.txt
justsmth: This aligns with what is configured here: https://github.com/aws/aws-lc/blob/main/CMakeLists.txt#L201-L206

util/fipstools/delocate/delocate.go
justsmth: I noticed that when a file with [a lower-case `*.s` extension](https://developer.arm.com/documentation/100066/0611/assembling-assembly-code/preprocessing-assembly-code) is passed to the preprocessor, it seems to produce no output. The build subsequently would fail during linking b/c it couldn't find the symbols defined in "bcm.s".

util/fipstools/delocate/delocate.go
justsmth: A few lines up from here, you see that I removed the logic that explicitly set the value at index = 0:
```
	if len(*arInput) > 0 {
		inputs = append(inputs, inputFile{
			path:      *arInput,
			index:     0,
			isArchive: true,
		})
	}
```

The logic on this line previously set the index as `i+1` to avoid a conflict with the above were `i=0`. Since this is no longer needed, I just removed the ` + 1`.

crypto/fipsmodule/CMakeLists.txt
justsmth: Let me look for a better way to do this.  This seems potentially problematic.

justsmth: ohh... redundant b/c the same lines are just above. So yes, it's redundant.

nebeid (approved):

nebeid (approved):

PR #2648, opened 2025-08-28
Gate EC explicit curve parameters for X.509 behind flag

skmcgrail:
### Description of changes: 
A follow-up to https://github.com/aws/aws-lc/pull/2642 which added supported for EC explicit curve parameters when parsing subject public key info structures. This adds `X509_VERIFY_PARAM_enable_ec_key_explicit_params` and `X509_VERIFY_PARAM_disable_ec_key_explicit_params` functions that can configure a `X509_STORE` or `X509_STORE_CTX`'s `X509_VERIFY_PARAM` to enable to disable the support for EC explicit curve parameters, and adds logic to `X509_verify_cert` which will reject such SPKI values in the certificate chain (2 or more certificates) that have explicit parameters if they corresponding enablement function has not been invoked. This ensures that the default behavior of AWS-LC does not change in libssl or consumers using our certificate verification functions, but they may choose to opt-in if they still support private certificate CAs that issue such certificates.

One logical implementation would have been to add support to the existing `X509_V_FLAG_*` values, and just utilize the `X509_STORE_CTX_set_flags`, but as these constants and associated values have mapped 1-1 with OpenSSL today  (minus the values mapped to no-op zero values), I opted it would be better to have functions that explicitly toggle the behavior on an internal set of flags, and would potentially allow us to make these functions fallible in the future if we remove the feature in the future.


crypto/ec_extra/internal.h
```diff
@@ -70,6 +70,13 @@ OPENSSL_EXPORT int ec_hash_to_scalar_p384_xmd_sha512_draft07(
     const EC_GROUP *group, EC_SCALAR *out, const uint8_t *dst, size_t dst_len,
     const uint8_t *msg, size_t msg_len);
 
+enum ECParametersType {
+  UNKNOWN_EC_PARAMETERS = 0,
+  NAMED_CURVE_EC_PARAMETERS = 1,
+  SPECIFIED_CURVE_EC_PARAMETERS = 2,
+};
+
+EC_GROUP *EC_KEY_maybe_parse_parameters(CBS *cbs, enum ECParametersType *paramType);
```
justsmth: NP: 
* A comment here that `paramType` is solely an OUT variable.
* The name `EC_KEY_maybe_parse_parameters` doesn't seem to indicate how its different than `EC_KEY_parse_parameters`. Perhaps include "type" in its name: `EC_KEY_parse_for_type_and_parameters`?

crypto/ec_extra/ec_asn1.c
```diff
@@ -365,8 +374,22 @@ int EC_KEY_marshal_curve_name(CBB *cbb, const EC_GROUP *group) {
...
   if (!CBS_peek_asn1_tag(cbs, CBS_ASN1_SEQUENCE)) {
-    return EC_KEY_parse_curve_name(cbs);
+    ret = EC_KEY_parse_curve_name(cbs);
+    if(ret) {
+      *paramType = NAMED_CURVE_EC_PARAMETERS;
+    }
+    return (EC_GROUP *)ret;
   }
 
   // OpenSSL sometimes produces ECPrivateKeys or ECPublicKey with explicitly-encoded versions
```
justsmth, line 395: I assume we no longer want to have a TODO to remove this support.

crypto/evp_extra/p_ec_asn1.c
```diff
@@ -160,6 +172,10 @@ static int eckey_priv_decode(EVP_PKEY *out, CBS *oid, CBS *params, CBS *key, CBS
     return 0;
   }
 
+  if (paramType == SPECIFIED_CURVE_EC_PARAMETERS) {
+    ec_key->group_decoded_from_explicit_params = 1;
+  }
+
```
justsmth: * No need for the `else` as on line 125?
```
... else {
    eckey->group_decoded_from_explicit_params = 0;
}
```
* Is it needed on line 125?

crypto/fipsmodule/ec/internal.h
```diff
@@ -830,6 +830,8 @@ struct ec_key_st {
 
   const EC_KEY_METHOD *eckey_method;
 
+  int group_decoded_from_explicit_params;
```
justsmth, line 833: Should we also use `ECParametersType` for this field?

crypto/x509/x509_vfy.c
```diff
@@ -633,6 +633,28 @@ static int check_custom_critical_extensions(X509_STORE_CTX *ctx, X509 *x) {
   return 1;
 }
 
+static int check_curve(X509 *x) {
```
justsmth: It's not clear what the return value means. A name like `check_curve_decoded_by_name` would be more clear.

crypto/fipsmodule/ec/internal.h
skmcgrail: We don't support implicit key types, and we don't ever successfully return a key for which there are no/unknown parameters. So it feels safe to treat this as a boolean value for the time being.

crypto/x509/x509_vfy.c
skmcgrail: I was keeping the name to match the OpenSSL corresponding function for simplicity in comparing implementations.

crypto/evp_extra/p_ec_asn1.c
skmcgrail: Technically no as `EC_KEY_new` zeros the structure, but I can explicitly set it here to be safe.

crypto/x509/x509_vfy.c
skmcgrail: On second thought I think being a little more explicit on the name for now makes sense, as it is only called if we should check that it was decoded by name. Rather then being called every time to check attributes about a PKEY if it is a curve.

justsmth (approved):

include/openssl/x509.h
```diff
@@ -3308,6 +3308,17 @@ OPENSSL_EXPORT int X509_VERIFY_PARAM_set_purpose(X509_VERIFY_PARAM *param,
 OPENSSL_EXPORT int X509_VERIFY_PARAM_set_trust(X509_VERIFY_PARAM *param,
                                                int trust);
 
+// X509_STORE_CTX_enable_ec_key_explicit_params enables X.509 subject public
+// keys to contain elliptic curve keys with explicit parameters. By default
+// AWS-LC rejects valiation of certificate chains containg public keys
```
samuel40791765:
```suggestion
// AWS-LC rejects validation of certificate chains containing public keys
```

include/openssl/x509.h
```diff
@@ -3308,6 +3308,17 @@ OPENSSL_EXPORT int X509_VERIFY_PARAM_set_purpose(X509_VERIFY_PARAM *param,
...
+// with explicit EC parameters. Returns 1 on success, or 0 on failure.
+OPENSSL_EXPORT int X509_VERIFY_PARAM_enable_ec_key_explicit_params(X509_VERIFY_PARAM *param);
+
+// X509_STORE_CTX_disable_ec_key_explicit_params disables X.509 subject public keys
+// to contain elliptic curve keys with explicit parameters. By default
+// AWS-LC rejects valiation of certificate chains containg public keys
```
samuel40791765:
```suggestion
// AWS-LC rejects validation of certificate chains containing public keys
```

crypto/x509/x509_test.cc
```diff
@@ -4923,26 +4923,38 @@ TEST(X509Test, SignatureVerification) {
...
-  EXPECT_EQ(X509_V_ERR_UNABLE_TO_DECODE_ISSUER_PUBLIC_KEY,
+  EXPECT_EQ(X509_V_ERR_UNSPECIFIED,
             Verify(leaf.valid.get(), {root.bad_key_type.get()},
                    {intermediate.valid.get()}, {}));
-  EXPECT_EQ(X509_V_ERR_UNABLE_TO_DECODE_ISSUER_PUBLIC_KEY,
+  EXPECT_EQ(X509_V_ERR_UNSPECIFIED,
             Verify(leaf.valid.get(), {root.valid.get()},
                    {intermediate.bad_key.get()}, {}));
-  EXPECT_EQ(X509_V_ERR_UNABLE_TO_DECODE_ISSUER_PUBLIC_KEY,
+  EXPECT_EQ(X509_V_ERR_UNSPECIFIED,
```
samuel40791765, line 4935: From a first glance, the error messages seem to become more unclear due to the changes. Is that intentional for more OpenSSL compatibility?

skmcgrail: This would actually be more inline with the OpenSSL behavior which returns unspecified at this point if the return code < 0.

PR #1829, opened 2024-09-05
ED25519 Service Indicator

skmcgrail:
Support for Ed25519 FIPS Service Indicator for KeyGen, SigGen, and SigVer.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

PR #2666, opened 2025-09-05
ML-DSA service indicator

jakemas:
### Issues:
Required for FIPS
Addresses ML-DSA FIPS 204

### Description of changes: 
Following https://github.com/aws/aws-lc/pull/1829, this PR adds the service indicator for FIPS 204 ML-DSA (both pure and pre-hash variants).

### Testing:
All EVP functions are tested, for keygen, sign, verify, in all levels of ml-dsa, and all pre-hash/pure variants.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -494,7 +494,7 @@ if(GCC OR CLANG)
     endif()
   endif()
 
-  if(MINGW)
+  if(MINGW AND NOT CLANG)
     # Some MinGW compilers set _WIN32_WINNT to an older version (Windows Server 2003)
     # See: https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-170
     # Support Windows 7 and later.


crypto/fipsmodule/ml_dsa/ml_dsa.c
```diff
@@ -44,7 +45,11 @@ int ml_dsa_44_keypair(uint8_t *public_key   /* OUT */,
   boringssl_ensure_ml_dsa_self_test();
   ml_dsa_params params;
   ml_dsa_44_params_init(&params);
-  return (ml_dsa_keypair(&params, public_key, private_key, seed) == 0);
+  int ret = (ml_dsa_keypair(&params, public_key, private_key, seed) == 0);
```
andrewhop: Any ML-DSA function that calls another approved service (e.g. SHA3) or has a self test needs to lock the service indicator before the call, and then unlock the service indicator after. Otherwise it's not clear to the caller if the ML-DSA call they made is approved or if it's the SHA3 call that was approved but ML-DSA was not approved. This is why the CI is failing with `Assertion failed: (before + 1 == after), function TestBody, file service_indicator_test.cc, line 5413.`. 

https://github.com/aws/aws-lc/pull/2112 has more information on the issue and examples.

crypto/fipsmodule/evp/p_pqdsa.c
```diff
@@ -49,7 +49,7 @@ static int pkey_pqdsa_keygen(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey) {
   PQDSA_KEY *key = PQDSA_KEY_new();
   if (key == NULL ||
       !PQDSA_KEY_init(key, pqdsa) ||
-      !pqdsa->method->pqdsa_keygen(key->public_key, key->private_key, key->seed) ||
+      pqdsa->method->pqdsa_keygen(key->public_key, key->private_key, key->seed) != 0 ||
```
jakemas: `ml_dsa_keypair: Returns 0 (success) -1 on failure`, (feels better than`!(-1)==0`) so cleaning this up, those flags from the reference implementation we inherited are a little bssl anti-pattern, and could change(I think they have before already?).

https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c#L137C2-L137C36

crypto/fipsmodule/ml_dsa/ml_dsa.c
jakemas: Thanks! I couldn't get CI to fire off at first, and my local FIPS build was in `RELEASE` so I wasn't getting the test output I wanted! That second PR helped!

crypto/fipsmodule/ml_dsa/ml_dsa.c
samuel40791765, line 1: Not within the scope of this PR, but is there a reason why all the internal ML-DSA functions return `0` on success and `-1` on failure? Seems a bit counter-intuitive to the general semantic of the rest of our functions.

crypto/fipsmodule/ml_dsa/ml_dsa.c
```diff
@@ -62,22 +69,34 @@ int ml_dsa_44_sign(const uint8_t *private_key /* IN */,
...
                    const uint8_t *ctx_string  /* IN */,
                    size_t ctx_string_len      /* IN */) {
+  FIPS_service_indicator_lock_state();
   boringssl_ensure_ml_dsa_self_test();
   ml_dsa_params params;
   ml_dsa_44_params_init(&params);
-  return ml_dsa_sign(&params, sig, sig_len, message, message_len,
-                     ctx_string, ctx_string_len, private_key) == 0;
+  int ret = ml_dsa_sign(&params, sig, sig_len, message, message_len,
+                        ctx_string, ctx_string_len, private_key) == 0;
```
samuel40791765, line 77: This or either 
```
  int ret = ml_dsa_sign(&params, sig, sig_len, message, message_len,
                        ctx_string, ctx_string_len, private_key);
``` 
is fine, but we should choose one and align the way we're doing things across all the functions within this file.

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -373,6 +390,21 @@ void EVP_PKEY_keygen_verify_service_indicator(const EVP_PKEY *pkey) {
...
+        case NID_MLDSA44:
+        case NID_MLDSA65:
+        case NID_MLDSA87:
+          // FIPS 204: ML-DSA Key Generation
+          FIPS_service_indicator_update_state();
+          break;
+        default:
+          break;
+      }
+    }
```
samuel40791765: Maybe we could consider abstracting this and above into a shared function?

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -5317,6 +5319,245 @@ TEST(ServiceIndicatorTest, ED25519SigGenVerify) {
...
+                               'M', 'E', 'S', 'S', 'A', 'G', 'E'};
+  const uint8_t CONTEXT[6] = {'A', 'W', 'S', '-', 'L', 'C'};
+
+  // Test raw ML-DSA functions for each parameter set
+  struct {
+    int (*keypair_func)(uint8_t *, uint8_t *, uint8_t *);
+    int (*sign_func)(const uint8_t *, uint8_t *, size_t *, const uint8_t *, size_t, const uint8_t *, size_t);
+    int (*verify_func)(const uint8_t *, const uint8_t *, size_t, const uint8_t *, size_t, const uint8_t *, size_t);
+    int (*extmu_sign_func)(const uint8_t *, uint8_t *, size_t *, const uint8_t *, size_t);
+    int (*extmu_verify_func)(const uint8_t *, const uint8_t *, size_t, const uint8_t *, size_t);
```
samuel40791765: These are great tests, but I've noticed that these ML-DSA functions are only used internally. This would be great for testing ML-DSA functional correctness, but for service indicator tests, testing against the ML-DSA functions that are exposed for typical consumers is more important.
I believe those would be the EVP functions and it seems like we're already testing them? If all the same code paths have already been covered, perhaps we could consider just removing the Service Indicator tests running against the internal functions.

crypto/fipsmodule/ml_dsa/ml_dsa.c
jakemas: I agree -- it's something we inherited from both reference mlkem and mldsa. It still shows up even after the mlkem-native import (https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/ml_kem/mlkem/kem.h#L124-L125 for example). 

We should go in and switch this all out for both algorithms, I agree, some return `0/1`, some `-1/0/1`, we can use more sensible library codes, and covert them to bssl style.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
jakemas: Ahh you are correct, I was looking at the ed25519 example above, but of course the APIs are hooked up differently there. I've removed the `mldsa_*` testing and kept all the `EVP` functions in [b21da25](https://github.com/aws/aws-lc/pull/2666/commits/b21da25a00cf86456f2df868d40fba63f38e7ca6), as they are the only FIPS path for ML-DSA -- as you say!

crypto/fipsmodule/service_indicator/service_indicator.c
jakemas: ok, done in [9119706](https://github.com/aws/aws-lc/pull/2666/commits/9119706e1cd1e5ae57c8515b582a4e8dc3df696b)

crypto/fipsmodule/ml_dsa/ml_dsa.c
jakemas: okay, unified to all use the approach 
```
int ret = ml_dsa_keypair(&params, public_key, private_key, seed) == 0;
  if (ret) {
  }
  return ret;
}
```
in
[ba1fcff](https://github.com/aws/aws-lc/pull/2666/commits/ba1fcff8c04850438c268aa5ab7f578af3687ffe)

samuel40791765 (approved):

samuel40791765 (approved):

skmcgrail (approved):

PR #2674, opened 2025-09-11
More arm64 CI tests

justsmth:
### Description of changes: 
Add more arm64 build targets to our CI.

### DIFF
--- a/.github/workflows/actions-ci.yml
+++ b/.github/workflows/actions-ci.yml
@@ -14,6 +14,7 @@ env:
   PACKAGE_NAME: aws-lc
   # Used to enable ASAN test dimension.
   AWSLC_NO_ASM_FIPS: 1
+  DEBIAN_FRONTEND: noninteractive
 
 jobs:
   # MacOS and Windows GHA runners are more expensive, so we do a sanity test run before proceeding.


crypto/fipsmodule/bcm.c
```diff
@@ -246,7 +246,7 @@ static void BORINGSSL_maybe_set_module_text_permissions(int permission) {
     perror("BoringSSL: mprotect");
   }
 }
-#else
+#elif !defined(OPENSSL_WINDOWS)
```
nebeid, line 249: Does this need to be excluding all Windows builds or only AArch64?

justsmth: Just Windows.  You can see that the only two places that `BORINGSSL_maybe_set_module_text_permissions` is called ([here](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/bcm.c#L357) and [here](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/bcm.c#L374)), it's guarded by:
```
#if !defined(OPENSSL_WINDOWS)
```

skmcgrail (approved):

PR #2702, opened 2025-09-18
Remove jitter entropy tests folder

torben-hansen:
### Issues:

P303750654

### Description of changes: 

This creates some annoyance downstream and we don't use the tests for anything. At least not right now. We can always reintroduce them.


third_party/jitterentropy/import.sh
```diff
@@ -62,6 +62,9 @@ echo "Remove upstream repository build scripts to avoid them being invoked ..."
 rm "${SRC}/CMakeLists.txt"
 rm "${SRC}/Makefile"
 
+echo "Remove test folder which is unused and contains recursive paths..."
+rm -rf ${SRC}/tests
+
```
TingDaoK, line 67: Maybe remove /doc and anything that is not needed as well? 
https://github.com/aws/aws-lc/tree/main/third_party/jitterentropy/jitterentropy-library/doc

Also, https://github.com/aws/aws-lc/tree/main/third_party/googletest/test the googletest also has a test folder, do we need it?

justsmth (approved):

samuel40791765 (approved):

PR #2708, opened 2025-09-23
Migrate linux-x86 jobs to self-hosted runners

skmcgrail:


tests/ci/cdk/cdk/codebuild/github_ci_linux_x86_omnibus.yaml
```diff
@@ -6,659 +6,10 @@ version: 0.2
...
-      buildspec: ./tests/ci/codebuild/common/run_simple_target.yml
-      env:
-        type: LINUX_CONTAINER
-        privileged-mode: true
-        compute-type: BUILD_GENERAL1_LARGE
-        image: 620771051181.dkr.ecr.us-west-2.amazonaws.com/aws-lc-docker-images-linux-x86:ubuntu-22.04_clang-14x-sde_latest
-        variables:
-          AWS_LC_CI_TARGET: "tests/ci/run_benchmark_build_tests.sh"
-
-    - identifier: ubuntu1004_gcc4_1x_x86_64_build
```
torben-hansen, line 636: missing from new list?

skmcgrail: That's referred to as the legacy_build in the new file.

.github/workflows/linux_x86_omnibus.yml
```diff
@@ -0,0 +1,467 @@
...
+          - ubuntu-22.04_gcc-11x_latest
+          - ubuntu-22.04_gcc-12x_latest
+        build32: [0] 
+        include:
+          - image: centos-7_gcc-4x_latest
+            build32: 1
+          - image: fedora-31_clang-9x_latest
+            cc: clang-9
+            cxx: clang++-9
+            build32: 0
```
justsmth, line 88: NP:

```
build32: [0]
```

Is this line needed? I suppose the line `${{ matrix.build32 || 0 }}` expression would fail otherwise.
* I ask b/c I noticed that you're able to reference `matrix.cc` below even though it's not listed as a dimension for the matrix.

justsmth (approved):

.github/workflows/linux_x86_omnibus.yml
skmcgrail: You can reference variables not in the matrix technically. But not defining this in the top-level creates the issue I described on the call where it was not creating the expected x86 and x86-64 test jobs for centos and ubuntu, but was instead collapsing it. So really the defaulting behavior is no longer needed. Can cleanup further later.

PR #2709, opened 2025-09-24
Add EVP_PKEY_check and EVP_PKEY_public_check for KEMs

jakemas:
### Issues:
N/A - ML-KEM support.

### Description of changes: 
mlkem-native provides `pk` and `sk` check functions [crypto_kem_check_pk](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/ml_kem/mlkem/kem.c#L52-L74) and [crypto_kem_check_sk](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/ml_kem/mlkem/kem.c#L77-L107), exposed as `ml_kem_{512/768/1024}_check_{pk/sk}`. This PR hooks them up to `EVP_PKEY_check` and `EVP_PKEY_public_check` so we can call them on a `PKEY` of type `KEM`.

This PR also implements a Pairwise Consistency Test (PCT) between the provided public and secret key. This means `KEM_check_key` enforces that the public key is present when the secret key is present -- that is intended behaviour, since both are needed to check consistency.

`EVP_PKEY_kem_check_key` now delegates to `KEM_check_key` rather than carrying its own copy of the encapsulate/decapsulate check. It is a public API, so two consequences are worth noting: it is now strictly stronger, because it also validates both key encodings, and a shared-secret mismatch now raises `EVP_R_KEM_PCT_FAILED` where it previously returned zero with nothing queued.

Five reason codes are added (`EVP_R_KEM_PCT_FAILED`, `EVP_R_MISSING_PRIVATE_KEY`, `EVP_R_MISSING_PUBLIC_KEY`, `EVP_R_INVALID_PUBLIC_KEY`, `EVP_R_INVALID_PRIVATE_KEY`, 142-146), so `generated-src/err_data.c` is regenerated. Each check failure queues exactly one reason code: the `kem_check_*` helpers raise their own, and `KEM_check_key` does not add another on top. In the PCT, an encapsulate/decapsulate fault or a length mismatch raises `ERR_R_INTERNAL_ERROR`, and only a shared-secret mismatch raises `EVP_R_KEM_PCT_FAILED`, so an allocation failure or a library fault is no longer reported as a failed consistency check.

The documented contracts of `EVP_PKEY_check` and `EVP_PKEY_public_check` in `include/openssl/evp.h` are updated to describe the KEM behaviour, including that `EVP_PKEY_check` requires the private key while `EVP_PKEY_public_check` does not.

### Call-outs:
`ml_kem_*_check_pk`/`check_sk` already exist on `main`, so this PR does not add the wrapper functions in `crypto/fipsmodule/ml_kem/ml_kem.c` that an earlier revision of it did.

Note what `check_sk` does and does not cover. It recomputes `H(ek_PKE)` and compares it against the copy embedded in the secret key, so it validates that binding only. Corruption elsewhere in the secret key behaves differently: a bit flipped in `dk_PKE` is caught by the PCT rather than by `check_sk`, and a bit flipped in `z`, the implicit-rejection seed, is not detected at all, since `z` is unused when decapsulation succeeds normally and nothing binds it to the rest of the key. This is why the negative test corrupts the hash region specifically.

### Testing:
KEM tests are modified to:
```
EXPECT_TRUE(EVP_PKEY_check(kem_key_ctx.get()));
EXPECT_TRUE(EVP_PKEY_public_check((kem_key_ctx.get())));
```

`PerKEMTest.KEMCheckKeyTests` covers `KEM_check_key` via `EVP_PKEY_check` and `EVP_PKEY_public_check`. Every negative case asserts the specific reason code, and that the error queue is empty afterwards, so the one-error-per-failure behaviour is covered rather than just the true/false result. The cases are:

- Valid Key Baseline - Verifies both `EVP_PKEY_check` and `EVP_PKEY_public_check` pass on a properly generated KEM keypair.
- Corrupted Public Key - Tests that validation fails when public key bytes are corrupted while secret key remains valid.
- Corrupted Secret Key - Tests that validation fails when the secret key's embedded hash is corrupted while public key remains valid.
- Mismatched Keypair - Tests Pairwise Consistency Test (PCT) failure detection when public and secret keys are from different keypairs.
- Public Key Only - Tests that `EVP_PKEY_public_check` passes when only public key material is provided, while `EVP_PKEY_check` fails with `EVP_R_MISSING_PRIVATE_KEY`, as with the EC and RSA cases.
- Public Key Only Corrupted - Tests that validation fails when only corrupted public key material is provided.
- Secret Key Only - Tests that validation fails when only secret key is provided without required public key.


crypto/fipsmodule/kem/kem.c
```diff
@@ -306,3 +306,75 @@ int KEM_KEY_set_raw_key(KEM_KEY *key, const uint8_t *in_public,
...
+  if (key->kem == NULL || key->kem->method == NULL) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_NO_KEY_SET);
+    return 0;
+  }
+
+  // Check that at least the public key exists
+  if (key->public_key == NULL) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_NO_KEY_SET);
+    return 0;
+  }
```
justsmth: NP: It's currently possible (although I wish it weren't) for the secret_key to be set but not the public_key. For those, this would always return an error.

crypto/fipsmodule/kem/kem.c
```diff
@@ -306,3 +306,75 @@ int KEM_KEY_set_raw_key(KEM_KEY *key, const uint8_t *in_public,
...
+      if (ml_kem_512_check_pk(key->public_key) != 0) {
+        OPENSSL_PUT_ERROR(EVP, EVP_R_DECODE_ERROR);
+        return 0;
+      }
+      // Check secret key validity if present
+      if (key->secret_key != NULL && ml_kem_512_check_sk(key->secret_key) != 0) {
+        OPENSSL_PUT_ERROR(EVP, EVP_R_DECODE_ERROR);
+        return 0;
+      }
+      break;
```
justsmth: We would also need a pair-consistency check, otherwise the secret and public keys might be unrelated. (?)

crypto/fipsmodule/kem/kem.c
jakemas: Currently intended behaviour, my idea is to merge the PRs first that ensure we can populate both. Then we fail this check if pub_key isn't set when secrect_key is? How does that sound?

crypto/fipsmodule/kem/kem.c
jakemas: Good idea -- I'll have to make changes upstream in mlkem-native to support this at it would require exposing/namspacing the pct tests, i.e. would need to add 

`#define crypto_kem_check_pct MLK_NAMESPACE_K(check_pct)`

to `/crypto/fipsmodule/ml_kem/mlkem/kem.h` and then I can hook it up just like I did the `check_{pk/sk}` tests that become `ml_kem_512_check_pk` etc.

and some changes to make sure they are available, as currently the PCTs are only defined in FIPS builds (when `MLK_CONFIG_KEYGEN_PCT`)

ooor we avoid all of that, and could just do a cheeky little encap decap in this function itself? I'll go check out the other algorithm variants of this function.

jakemas: Had a chat with Hanno today, he advocates for writing a (new) PCT within the function itself. Implemented in [d24a116](https://github.com/aws/aws-lc/pull/2709/commits/d24a1168a75b991abd09d61616607293ee48cc50)

crypto/fipsmodule/kem/kem.c
jakemas: Once caveat is that `EVP_PKEY_public_check` also calls the same function e.g. `EC_KEY_check_key` that just handles it. [See here](https://github.com/aws/aws-lc/blob/main/crypto/evp_extra/evp_asn1.c#L343-L364). Little bit strange, but if we follow the pattern, we'd have to make KEM_CHECK ok with pub only. Or we could implement two functions, the second `KEM_CHECK_PUBLIC_KEY` and call the it from `EVP_PKEY_public_check`: drafted that idea out here [0616873](https://github.com/aws/aws-lc/pull/2709/commits/0616873c1b6dd9f025ad3b86e59d816138478e93), but I don't think I like the anti-pattern. Looking at EC check for inspo, I will implement similar to how they do it,  If only public key: validates public key only, if secret key is present: requires public key and validates public key, secret key, and PCT.

jakemas: I've landed on [f30bfd4](https://github.com/aws/aws-lc/pull/2709/commits/f30bfd4ac9d3aaf7d39de2ade0cef255495e3e24) that preserves functionality with other `EVP_PKEY_public_check/EVP_PKEY_check` functions. As documented, it

```
// KEM_check_key: validates the key based on available key material
// - If only public key: validates public key only
// - If secret key is present: requires public key and validates public key, secret key, and PCT
```

I added negative testing for a bunch of cases, including the one you mention:

```
  // ---- 7. Test with secret key only (no public key) ----
  // Create EVP_PKEY with only secret key (no public key)
```

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -2780,6 +2777,156 @@ TEST_P(PerKEMTest, RawKeyOperations) {
...
+  ASSERT_TRUE(EVP_PKEY_get_raw_public_key(valid_pkey, pk_copy.data(), &pk_len));
+  ASSERT_TRUE(EVP_PKEY_get_raw_private_key(valid_pkey, sk_copy.data(), &sk_len));
+  
+  // Create a corrupted public key
+  std::vector<uint8_t> corrupted_pk = pk_copy;
+  corrupted_pk[0] = 0xFF;
+  corrupted_pk[1] = 0xFF;
+  if (pk_len > 2) {
+    corrupted_pk[2] = 0xFF;
+  }
```
dkostic: isn't it enough to corrupt a single bit? why do we change two more bytes?

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -2780,6 +2777,156 @@ TEST_P(PerKEMTest, RawKeyOperations) {
...
+  size_t sk_len = GetParam().secret_key_len;
+  std::vector<uint8_t> pk_copy(pk_len);
+  std::vector<uint8_t> sk_copy(sk_len);
+  
+  ASSERT_TRUE(EVP_PKEY_get_raw_public_key(valid_pkey, pk_copy.data(), &pk_len));
+  ASSERT_TRUE(EVP_PKEY_get_raw_private_key(valid_pkey, sk_copy.data(), &sk_len));
+  
+  // Create a corrupted public key
+  std::vector<uint8_t> corrupted_pk = pk_copy;
+  corrupted_pk[0] = 0xFF;
```
dkostic, line 3052: this will produce a flaky test because with some probability the valid keys will have the bytes equal to 0xFF :)
to be sure that you are corrupting the keys you can just XOR with 1 (`^ 1`), that always changes the least significant bit.

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -2780,6 +2777,156 @@ TEST_P(PerKEMTest, RawKeyOperations) {
...
+      EVP_PKEY_CTX_new(corrupted_pk_pkey.get(), nullptr));
+  ASSERT_TRUE(corrupted_pk_ctx);
+  EXPECT_FALSE(EVP_PKEY_check(corrupted_pk_ctx.get()));
+  EXPECT_FALSE(EVP_PKEY_public_check(corrupted_pk_ctx.get()));
+
+  // ---- 3. Test with corrupted secret key ----
+  // Create a corrupted secret key
+  std::vector<uint8_t> corrupted_sk = sk_copy;
+  
+  // Corrupt the last 64 bytes of secret key (in ML-KEM this is the hash)
```
dkostic: shouldn't corrupting any single bit in the secret key be sufficient?

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -2780,6 +2777,156 @@ TEST_P(PerKEMTest, RawKeyOperations) {
...
+  
+  // This should fail EVP_PKEY_check due to invalid secret key
+  bssl::UniquePtr<EVP_PKEY_CTX> corrupted_sk_ctx(
+      EVP_PKEY_CTX_new(corrupted_sk_pkey.get(), nullptr));
+  ASSERT_TRUE(corrupted_sk_ctx);
+  EXPECT_FALSE(EVP_PKEY_check(corrupted_sk_ctx.get()));
+  // Public key check will fail PCT since secret key is present, and corrupted
+  EXPECT_FALSE(EVP_PKEY_public_check(corrupted_sk_ctx.get()));
+
+  // ---- 4. Test mismatched keypair (PCT failure) ----
```
dkostic: is there a differentiation between mismatched keypair and corrupted key pare cases above? is a mismatched key pair going to fail differently than a corrupted key pair?

crypto/fipsmodule/kem/kem.c
```diff
@@ -356,3 +356,107 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+  // If we have a secret key, we must also have a public key for PCT
+  if (key->secret_key != NULL && key->public_key == NULL) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_NO_KEY_SET);
+    return 0;
+  }
+
+  // Call appropriate ML-KEM check functions based on KEM NID
+  switch (key->kem->nid) {
+    case NID_MLKEM512:
+    case NID_KYBER512_R3:
```
dkostic: I don't think we should mix-up Kyber with ML-KEM, technically they are not the same algorithms. I'd drop Kyber NIDs from here.

crypto/fipsmodule/ml_kem/ml_kem.c
```diff
@@ -394,3 +394,214 @@ int ml_kem_common_decapsulate(int (*decapsulate)(uint8_t *shared_secret, const u
...
+  }
+  // Call the underlying ML-KEM-1024 check function
+  return mlkem1024_check_sk(secret_key);
+}
+
+// ML-KEM Pairwise Consistency Test (PCT) functions
+// These functions perform FIPS 203 compliant PCT by performing encapsulation
+// and decapsulation operations and comparing the resulting shared secrets
+
+int ml_kem_512_check_pct(const uint8_t *public_key, const uint8_t *secret_key) {
```
dkostic: mlkem-native doesn't have pct already implemented?

jakemas:
Rewrote this on top of current main. Much simpler now since check_pk/check_sk already landed upstream — no more wrappers in ml_kem.c needed.

Addressed all the review feedback:
- @dkostic: Dropped Kyber NIDs (removed upstream), using XOR for byte corruption, non-canonical coefficient injection (>= q) for pk tests, and `CRYPTO_memcmp` for constant-time comparison in PCT. Re: your question about mlkem-native PCT — it's there but static/internal, so the encaps/decaps approach (per Hanno's recommendation) is what we use here through the KEM_METHOD pointers.
- @justsmth: PCT is now generic via encaps+decaps through the method table, handles the pub-key-only case gracefully.

@sgmenda should be ready for review now.

crypto/evp_extra/evp_asn1.c
justsmth, line 280: * The [expected behavior for `EVP_PKEY_check`](https://github.com/aws/aws-lc/blob/aaa3c16445480d861a787bb9d8edd83a631ca98e/include/openssl/evp.h#L710-L716) is that it must include the private key -- the current implementation doesn't enforce this.
```
// EVP_PKEY_check supports EC and RSA keys and validates both the public and
// private components of a key pair. For EC keys, it verifies that the private
// key component exists and calls EC_KEY_check_key. For RSA keys, it calls
// RSA_check_key which validates both public and private key relationships.
```
* The documentation in our public header `evp.h` also needs to be updated.

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +338,117 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+    if (!kem_check_secret_key(key)) {
+      OPENSSL_PUT_ERROR(CRYPTO, ERR_R_INTERNAL_ERROR);
+      return 0;
+    }
+
+    if (!kem_check_pct(key)) {
+      OPENSSL_PUT_ERROR(CRYPTO, ERR_R_INTERNAL_ERROR);
+      return 0;
+    }
+  }
```
justsmth, line 455: * These all result in `ERR_R_INTERNAL_ERROR`, which is not very helpful.  Should we introduce a `EVP_R_KEM_PCT_FAILED` or some other error that's a better  indicator of the failure?
* Should these be `OPENSSL_PUT_ERROR( EVP, ...` instead of `CRYPTO`?

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +338,117 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+static int kem_check_public_key(const KEM_KEY *key) {
+  switch (key->kem->nid) {
+    case NID_MLKEM512:
+      return ml_kem_512_check_pk(key->public_key, key->kem->public_key_len) == 0;
+    case NID_MLKEM768:
+      return ml_kem_768_check_pk(key->public_key, key->kem->public_key_len) == 0;
+    case NID_MLKEM1024:
+      return ml_kem_1024_check_pk(key->public_key, key->kem->public_key_len) == 0;
+    default:
+      return 0;
```
justsmth, line 358: Should we abort in this case, or at least report a `ERR_R_INTERNAL_ERROR`?  This branch should not be reachable.

crypto/evp_extra/evp_extra_test.cc
jakemas: addressed

crypto/fipsmodule/kem/kem.c
jakemas: addressed kyber out now

crypto/fipsmodule/ml_kem/ml_kem.c
jakemas: addressed, it does now

crypto/evp_extra/evp_extra_test.cc
jakemas: addressed

crypto/evp_extra/evp_extra_test.cc
jakemas: Good catch on the flakiness. There are two different corruption sites and they need different treatment:

- **Secret key** (`corrupted_sk[sk_len - 64] ^= 1`) — now uses `^ 1` as you suggested, so it deterministically flips a bit.
- **Public key** — `check_pk` is a *range/canonicity* check, not a hash check. A single bit flip on a coefficient usually yields another value still in range (`< q = 3329`), so the check would pass and we wouldn't actually be testing corruption. Instead we force the first 12-bit coefficient to `0xFFF` (4095 ≥ 3329), which is guaranteed out of range regardless of the original bytes — so it's deterministic, not probabilistic:

```c
corrupted_pk[0] = 0xFF;
corrupted_pk[1] = (corrupted_pk[1] & 0xF0) | 0x0F;
```

So both sites are now non-flaky; the public-key case just can't use `^ 1` because the failure condition is "coefficient ≥ q" rather than "bytes differ".

crypto/evp_extra/evp_asn1.c
jakemas: Good call — fixed in d88bf3d. EVP_PKEY_check now requires the secret key for KEM (via a new KEM_KEY_get0_secret_key accessor), same as the EC/RSA cases. public_check still passes with pk only. Updated the evp.h docs and the test to match.

crypto/fipsmodule/kem/kem.c
jakemas: Yeah, agreed. Added EVP_R_KEM_PCT_FAILED for the PCT failure, plus EVP_R_INVALID_PUBLIC_KEY / EVP_R_INVALID_PRIVATE_KEY / EVP_R_MISSING_PUBLIC_KEY for the other cases, and switched them all to OPENSSL_PUT_ERROR(EVP, ...). d88bf3d.

crypto/fipsmodule/kem/kem.c
jakemas: Done — it now puts ERR_R_INTERNAL_ERROR before returning. Left it returning failure rather than aborting since it's genuinely unreachable (we only build KEM_KEYs for the three ML-KEM NIDs), but this way we at least get a signal if that ever stops being true. d88bf3d.

crypto/fipsmodule/kem/kem.c
```diff
@@ -8,6 +8,7 @@
 #include "internal.h"
 #include <openssl/bytestring.h>
 #include <openssl/err.h>
+#include <openssl/mem.h>
```
justsmth, line 12: `kem.c` uses `EVP_R_*` and the `NID_MLKEM*` constants but doesn't include the header that declares them. It still compiles because `bcm.c` pulls in `evp/evp.c` before `kem/kem.c`.

```suggestion
#include <openssl/evp.h>
#include <openssl/mem.h>
```

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +342,121 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+  uint8_t *ss_enc = OPENSSL_malloc(kem->shared_secret_len);
+  uint8_t *ss_dec = OPENSSL_malloc(kem->shared_secret_len);
+
+  if (ciphertext == NULL || ss_enc == NULL || ss_dec == NULL) {
+    goto cleanup;
+  }
+
+  size_t ct_len = kem->ciphertext_len;
+  size_t ss_enc_len = kem->shared_secret_len;
+  size_t ss_dec_len = kem->shared_secret_len;
```
justsmth: Nit: The malloc-failure `goto cleanup` jumps over the initialization of `ct_len`/`ss_enc_len`/`ss_dec_len`. It's harmless b/c the `cleanup` block doesn't read them, but having the declarations first avoids potential compiler warnings about jumping past initializations.

```suggestion
  size_t ct_len = kem->ciphertext_len;
  size_t ss_enc_len = kem->shared_secret_len;
  size_t ss_dec_len = kem->shared_secret_len;

  if (ciphertext == NULL || ss_enc == NULL || ss_dec == NULL) {
    goto cleanup;
  }
```
`

justsmth (dismissed):
A couple of "nits" -- no blocking concerns.

justsmth (dismissed):

crypto/err/evp.errordata
```diff
@@ -22,14 +22,19 @@ EVP,114,INVALID_OPERATION
 EVP,115,INVALID_PADDING_MODE
 EVP,133,INVALID_PARAMETERS
 EVP,134,INVALID_PEER_KEY
+EVP,146,INVALID_PRIVATE_KEY
```
manastasova, line 25: How/Where are the numbers (e.g., 146) defined?

crypto/evp_extra/evp_extra_test.cc
```diff
@@ -2978,6 +2975,123 @@ TEST_P(PerKEMTest, RawKeyOperations) {
...
+  ASSERT_TRUE(EVP_PKEY_get_raw_private_key(valid_pkey, sk_copy.data(), &sk_len));
+
+  // Non-canonical public key fails (inject coefficient >= q=3329)
+  // ML-KEM coefficients are 12-bit little-endian packed.
+  // Set first coefficient to 0xFFF (4095 > 3329) to trigger check_pk failure.
+  std::vector<uint8_t> corrupted_pk = pk_copy;
+  corrupted_pk[0] = 0xFF;
+  corrupted_pk[1] = (corrupted_pk[1] & 0xF0) | 0x0F;
+
+  bssl::UniquePtr<EVP_PKEY> corrupted_pk_pkey(
```
manastasova, line 3008: Why not adding a check for correctness (i.e, the `EVP_PKEY_check` functionality) in the `EVP_PKEY_kem_new_raw_key` and avoid the creation of corrupted keys at all?

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +343,121 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+  }
+
+  if (CRYPTO_memcmp(ss_enc, ss_dec, kem->shared_secret_len) != 0) {
+    goto cleanup;
+  }
+
+  ret = 1;
+
+cleanup:
+  if (ciphertext != NULL) {
```
manastasova: nit: Are the NULL checks needed?

crypto/err/evp.errordata
jakemas: They come from the `EVP_R_*` defines in `include/openssl/evp_errors.h` (e.g. `EVP_R_INVALID_PRIVATE_KEY` is 146). `evp.errordata` and `generated-src/err_data.c` are both regenerated from those by `go run util/make_errors.go`, so the numbers aren't hand-written here.

crypto/evp_extra/evp_extra_test.cc
jakemas: The corrupted-key cases are there specifically to prove `EVP_PKEY_check` rejects bad keys, so we do want to feed it deliberately-broken material. Validating inside `EVP_PKEY_kem_new_raw_key` would also change its contract (raw-key construction currently doesn't run the check), so I'd rather keep the two separate.

crypto/fipsmodule/kem/kem.c
jakemas: Good catch, they're not - OPENSSL_cleanse and OPENSSL_free both no-op on NULL. Dropped the guards in 67b6ddc.

justsmth (commented):
---

crypto/evp_extra/evp_asn1.c
```diff
@@ -281,6 +281,16 @@ int EVP_PKEY_check(EVP_PKEY_CTX *ctx) {
     }
     case EVP_PKEY_RSA:
       return RSA_check_key(pkey->pkey.rsa);
+    case EVP_PKEY_KEM: {
+      KEM_KEY *kem_key = pkey->pkey.kem_key;
+      // For EVP_PKEY_check, ensure the private key exists for KEM keys, as with
+      // the EC and RSA cases above.
```
justsmth: A KEM |EVP_PKEY| can have its type set but no key material attached.
```suggestion
      if (kem_key == NULL) {
        OPENSSL_PUT_ERROR(EVP, EVP_R_NO_KEY_SET);
        return 0;
      }
      // For EVP_PKEY_check, ensure the private key exists for KEM keys, as with
      // the EC case above.
```

crypto/fipsmodule/kem/kem.c
```diff
@@ -249,6 +251,10 @@ const KEM *KEM_KEY_get0_kem(KEM_KEY* key) {
   return key->kem;
 }
 
+const uint8_t *KEM_KEY_get0_secret_key(const KEM_KEY *key) {
+  return key->secret_key;
```
justsmth, line 261: Callers use a NULL return to test for a secret key, so accept a NULL |key|.
```suggestion
  if (key == NULL) {
    return NULL;
  }
  return key->secret_key;
```

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +343,116 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+    case NID_MLKEM768:
+      return ml_kem_768_check_pk(key->public_key, key->kem->public_key_len) == 0;
+    case NID_MLKEM1024:
+      return ml_kem_1024_check_pk(key->public_key, key->kem->public_key_len) == 0;
+    default:
+      // Unreachable: KEM_KEY objects are only created for the NIDs above.
+      OPENSSL_PUT_ERROR(EVP, ERR_R_INTERNAL_ERROR);
+      return 0;
+  }
+}
```
justsmth, line 378: There's inconsistent expectations about where errors get queued -- is it here or in the caller (shouldn't be both).  

I suggest raising the error here in this function, not in the caller:

```suggestion
static int kem_check_public_key(const KEM_KEY *key) {
  int ok;
  switch (key->kem->nid) {
    case NID_MLKEM512:
      ok = ml_kem_512_check_pk(key->public_key, key->kem->public_key_len) == 0;
      break;
    case NID_MLKEM768:
      ok = ml_kem_768_check_pk(key->public_key, key->kem->public_key_len) == 0;
      break;
    case NID_MLKEM1024:
      ok = ml_kem_1024_check_pk(key->public_key, key->kem->public_key_len) == 0;
      break;
    default:
      // Unreachable: KEM_KEY objects are only created for the NIDs above.
      OPENSSL_PUT_ERROR(EVP, ERR_R_INTERNAL_ERROR);
      return 0;
  }
  if (!ok) {
    OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_PUBLIC_KEY);
    return 0;
  }
  return 1;
}
```

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +343,116 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+    case NID_MLKEM768:
+      return ml_kem_768_check_sk(key->secret_key, key->kem->secret_key_len) == 0;
+    case NID_MLKEM1024:
+      return ml_kem_1024_check_sk(key->secret_key, key->kem->secret_key_len) == 0;
+    default:
+      // Unreachable: KEM_KEY objects are only created for the NIDs above.
+      OPENSSL_PUT_ERROR(EVP, ERR_R_INTERNAL_ERROR);
+      return 0;
+  }
+}
```
justsmth, line 402: Same for this function:
```suggestion
 static int kem_check_secret_key(const KEM_KEY *key) {
  int ok;
  switch (key->kem->nid) {
    case NID_MLKEM512:
      ok = ml_kem_512_check_sk(key->secret_key, key->kem->secret_key_len) == 0;
      break;
    case NID_MLKEM768:
      ok = ml_kem_768_check_sk(key->secret_key, key->kem->secret_key_len) == 0;
      break;
    case NID_MLKEM1024:
      ok = ml_kem_1024_check_sk(key->secret_key, key->kem->secret_key_len) == 0;
      break;
    default:
      // Unreachable: KEM_KEY objects are only created for the NIDs above.
      OPENSSL_PUT_ERROR(EVP, ERR_R_INTERNAL_ERROR);
      return 0;
  }
  // Raise the error here, not in the caller, so only one error is queued.
  if (!ok) {
    OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_PRIVATE_KEY);
    return 0;
  }
  return 1;
}
```

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +343,116 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+  }
+
+  if (ss_enc_len != kem->shared_secret_len ||
+      ss_dec_len != kem->shared_secret_len) {
+    goto cleanup;
+  }
+
+  if (CRYPTO_memcmp(ss_enc, ss_dec, kem->shared_secret_len) != 0) {
+    goto cleanup;
+  }
```
justsmth, line 449:
```suggestion
  // An encaps/decaps failure is a library fault, not an invalid key pair.
  if (!kem->method->encaps(ciphertext, &ct_len, ss_enc, &ss_enc_len,
                           key->public_key)) {
    OPENSSL_PUT_ERROR(EVP, ERR_R_INTERNAL_ERROR);
    goto cleanup;
  }

  // Validate the encapsulation output before feeding it back into decaps.
  if (ct_len != kem->ciphertext_len) {
    OPENSSL_PUT_ERROR(EVP, ERR_R_INTERNAL_ERROR);
    goto cleanup;
  }

  if (!kem->method->decaps(ss_dec, &ss_dec_len, ciphertext, key->secret_key)) {
    OPENSSL_PUT_ERROR(EVP, ERR_R_INTERNAL_ERROR);
    goto cleanup;
  }

  if (ss_enc_len != kem->shared_secret_len ||
      ss_dec_len != kem->shared_secret_len) {
    OPENSSL_PUT_ERROR(EVP, ERR_R_INTERNAL_ERROR);
    goto cleanup;
  }

  // Only a shared-secret mismatch means the key pair itself failed the PCT.
  if (CRYPTO_memcmp(ss_enc, ss_dec, kem->shared_secret_len) != 0) {
    OPENSSL_PUT_ERROR(EVP, EVP_R_KEM_PCT_FAILED);
    goto cleanup;
  }
```

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +343,116 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+    if (!kem_check_secret_key(key)) {
+      OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_PRIVATE_KEY);
+      return 0;
+    }
+
+    if (!kem_check_pct(key)) {
+      OPENSSL_PUT_ERROR(EVP, EVP_R_KEM_PCT_FAILED);
+      return 0;
+    }
+  }
```
justsmth, line 485: (Assuming suggestion above taken.) The helpers above queue their own errors, so do not add another one here.
```suggestion
  if (!kem_check_public_key(key)) {
    return 0;
  }

  if (key->secret_key != NULL) {
    if (!kem_check_secret_key(key)) {
      return 0;
    }

    if (!kem_check_pct(key)) {
      return 0;
    }
  }
```

justsmth (commented):
Couldn't [this entire section](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/evp/p_kem.c#L541-L589) now be replaced with a call to `KEM_check_key`?
```c
  return KEM_check_key(key->pkey.kem_key);
```

crypto/evp_extra/evp_asn1.c
jakemas: Fixed.

crypto/fipsmodule/kem/kem.c
jakemas: Fixed.

crypto/fipsmodule/kem/kem.c
jakemas: Fixed. Used `int ok = 0;` — clang-tidy flags the uninitialized form.

crypto/fipsmodule/kem/kem.c
jakemas: Fixed.

crypto/fipsmodule/kem/kem.c
jakemas: Fixed. Left the allocation branch alone, since `OPENSSL_malloc` already queues its own error.

crypto/fipsmodule/kem/kem.c
jakemas: Fixed.

jakemas:
Done. Kept the NULL checks above it so both halves of the key pair are still required, which the existing tests rely on. Also added the reason code to each negative case in `KEMCheckKeyTests`, since none were asserted.

crypto/fipsmodule/kem/kem.c
jakemas: Fixed in `fa05b37e4` — `<openssl/evp.h>` is included directly now.

crypto/fipsmodule/kem/kem.c
jakemas: Fixed in `fa05b37e4` — the three length declarations now sit above the malloc-failure `goto`.

justsmth (dismissed):

include/openssl/evp.h
```diff
@@ -1001,7 +1008,9 @@ OPENSSL_EXPORT EVP_PKEY *EVP_PKEY_kem_new_raw_key(int nid,
                                                   size_t len_secret);
 
 // EVP_PKEY_kem_check_key validates that the public key in |key| corresponds
-// to the secret key in |key|.
+// to the secret key in |key|, and that both encodings are well-formed. Both
+// halves of the key pair must be present. Prefer |EVP_PKEY_check|, which
+// applies the same validation and accepts a public-only key.
```
justsmth: Should be:
```suggestion
// halves of the key pair must be present. Prefer |EVP_PKEY_check| when a
// private key is available, or |EVP_PKEY_public_check| for a public-only key.
```

crypto/evp_extra/evp_asn1.c
```diff
@@ -304,6 +319,8 @@ int EVP_PKEY_public_check(EVP_PKEY_CTX *ctx) {
       return EC_KEY_check_key(pkey->pkey.ec);
     case EVP_PKEY_RSA:
       return RSA_check_key(pkey->pkey.rsa);
+    case EVP_PKEY_KEM:
+      return KEM_check_key(pkey->pkey.kem_key);
```
justsmth: Nit: Null-check here as well? Otherwise, `KEM_check_key` will set `ERR_R_PASSED_NULL_PARAMETER` instead.
```suggestion
    case EVP_PKEY_KEM: {
      KEM_KEY *kem_key = pkey->pkey.kem_key;
      // A KEM |EVP_PKEY| can have its type set but no key material attached.
      if (kem_key == NULL) {
        OPENSSL_PUT_ERROR(EVP, EVP_R_NO_KEY_SET);
        return 0;
      }
      return KEM_check_key(kem_key);
```

include/openssl/evp.h
jakemas: Fixed.

crypto/evp_extra/evp_asn1.c
jakemas: Fixed. Added test coverage for the branch too.

justsmth (dismissed):

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +348,148 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+  if (CRYPTO_memcmp(ss_enc, ss_dec, kem->shared_secret_len) != 0) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_KEM_PCT_FAILED);
+    goto cleanup;
+  }
+
+  ret = 1;
+
+cleanup:
+  // OPENSSL_cleanse and OPENSSL_free both no-op on NULL, so no guards needed.
+  OPENSSL_cleanse(ciphertext, kem->ciphertext_len);
```
torben-hansen: `OPENSSL_free` calls `OPENSSL_cleanse`. So this is doing duplicate work.

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +348,148 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+  return ret;
+}
+
+int KEM_check_key(const KEM_KEY *key) {
+  if (key == NULL) {
+    OPENSSL_PUT_ERROR(EVP, ERR_R_PASSED_NULL_PARAMETER);
+    return 0;
+  }
+
+  if (key->kem == NULL || key->kem->method == NULL) {
```
torben-hansen: first and second if statements can be joined

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +348,148 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+    OPENSSL_PUT_ERROR(EVP, ERR_R_PASSED_NULL_PARAMETER);
+    return 0;
+  }
+
+  if (key->public_key == NULL) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_MISSING_PUBLIC_KEY);
```
torben-hansen, line 469: These error codes are defined in `evp_errors.h` which should be included.

crypto/evp_extra/evp_asn1.c
```diff
@@ -281,6 +281,21 @@ int EVP_PKEY_check(EVP_PKEY_CTX *ctx) {
...
       return RSA_check_key(pkey->pkey.rsa);
+    case EVP_PKEY_KEM: {
+      KEM_KEY *kem_key = pkey->pkey.kem_key;
+      // A KEM |EVP_PKEY| can have its type set but no key material attached.
+      if (kem_key == NULL) {
+        OPENSSL_PUT_ERROR(EVP, EVP_R_NO_KEY_SET);
+        return 0;
+      }
+      // For EVP_PKEY_check, ensure the private key exists for KEM keys, as with
+      // the EC and RSA cases above.
```
torben-hansen: RSA case doesn't do that.

crypto/evp_extra/evp_asn1.c
```diff
@@ -304,6 +319,15 @@ int EVP_PKEY_public_check(EVP_PKEY_CTX *ctx) {
       return EC_KEY_check_key(pkey->pkey.ec);
     case EVP_PKEY_RSA:
       return RSA_check_key(pkey->pkey.rsa);
+    case EVP_PKEY_KEM: {
+      KEM_KEY *kem_key = pkey->pkey.kem_key;
+      // A KEM |EVP_PKEY| can have its type set but no key material attached.
```
torben-hansen: redundant comment.

crypto/fipsmodule/evp/p_kem.c
```diff
@@ -538,53 +538,9 @@ int EVP_PKEY_kem_check_key(EVP_PKEY *key) {
...
-  // 0 or 1.
-  ret = constant_time_is_zero_8(res) & 1;
-
-end:
-  OPENSSL_free(ct);
-  OPENSSL_free(ss_a);
-  OPENSSL_free(ss_b);
-  EVP_PKEY_CTX_free(ctx);
-  return ret;
+  // |KEM_check_key| performs the same encapsulate/decapsulate consistency
```
torben-hansen: The same checks as what?

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +348,148 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+    return 0;
+  }
+
+  // The helpers above queue their own errors, so none are added here.
```
torben-hansen: helper functions below?

crypto/fipsmodule/kem/kem.c
jakemas: Fixed.

crypto/fipsmodule/kem/kem.c
jakemas: Fixed.

crypto/fipsmodule/kem/kem.c
jakemas: `kem.c` includes `<openssl/evp.h>`, which pulls in `evp_errors.h` with an `// IWYU pragma: export`, so that's the sanctioned way to get the `EVP_R_*` codes. Only `hpke.c` includes `evp_errors.h` directly. Happy to add it if you'd still prefer the explicit include.

crypto/evp_extra/evp_asn1.c
jakemas: Fixed — comment now says the EC case does it explicitly and `RSA_check_key` does it internally.

crypto/evp_extra/evp_asn1.c
jakemas: Fixed.

crypto/fipsmodule/evp/p_kem.c
jakemas: Fixed — now says which checks it shares with this function.

crypto/fipsmodule/kem/kem.c
jakemas: The helpers are above, but the comment above them already says this, so I dropped it.

crypto/fipsmodule/kem/kem.c
torben-hansen: you transitively get them. Rule-of-thumb is to include definitions that you use. Many places in our code, we only compile because the compilation unit is derived from the giant bcm.c

crypto/evp_extra/evp_asn1.c
```diff
@@ -281,6 +281,21 @@ int EVP_PKEY_check(EVP_PKEY_CTX *ctx) {
...
+      if (kem_key == NULL) {
+        OPENSSL_PUT_ERROR(EVP, EVP_R_NO_KEY_SET);
+        return 0;
+      }
+      // For EVP_PKEY_check, ensure the private key exists for KEM keys, as the
+      // EC case above does explicitly and RSA_check_key does internally.
```
torben-hansen: `RSA_check_key` doesn't require the private key to be present.

torben-hansen (commented):
just some precision nits.

crypto/fipsmodule/kem/kem.c
jakemas: Added it — fair point on bcm.c.

crypto/evp_extra/evp_asn1.c
jakemas: Right it returns 1 for a public-only key. Comment now says only the EC case enforces this.

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +349,140 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+    OPENSSL_PUT_ERROR(EVP, EVP_R_MISSING_PUBLIC_KEY);
+    return 0;
+  }
+
+  if (!kem_check_public_key(key)) {
+    return 0;
+  }
+
+  if (key->secret_key != NULL) {
+    if (!kem_check_secret_key(key)) {
```
justsmth, line 478: The hash/RNG calls inside `kem_check_secret_key` and the PCT run without an outer service-indicator lock, incrementing it even when validation fails.

In a FIPS build, wrapping a valid `EVP_PKEY_check` in `CALL_SERVICE_AND_CHECK_APPROVED` would abort because the counter increments 7-9 times rather than once. (The key-check tests call these functions directly, and the ML-KEM indicator tests cover keygen/encaps/decaps but not key validation.)

crypto/fipsmodule/kem/kem.c
```diff
@@ -337,3 +349,140 @@ int KEM_KEY_set_raw_keypair_from_seed(KEM_KEY *key, const CBS *seed) {
...
+  size_t ss_enc_len = kem->shared_secret_len;
+  size_t ss_dec_len = kem->shared_secret_len;
+
+  if (ciphertext == NULL || ss_enc == NULL || ss_dec == NULL) {
+    // |OPENSSL_malloc| has already queued the allocation failure.
+    goto cleanup;
+  }
+
+  // An encaps/decaps failure is a library fault, not an invalid key pair.
+  if (!kem->method->encaps(ciphertext, &ct_len, ss_enc, &ss_enc_len,
```
justsmth, line 422: `KEM_check_key` should establish `SET_DIT_AUTO_RESET` at entry so all callers retain that protection.

Direct method calls bypass `SET_DIT_AUTO_RESET` in `EVP_PKEY_encapsulate`/`EVP_PKEY_decapsulate`, and the internal ML-KEM wrappers do not replace it. This removes the opt-in timing protection around the PCT.

crypto/fipsmodule/evp/p_kem.c
```diff
@@ -538,53 +538,9 @@ int EVP_PKEY_kem_check_key(EVP_PKEY *key) {
...
-  OPENSSL_free(ct);
-  OPENSSL_free(ss_a);
-  OPENSSL_free(ss_b);
-  EVP_PKEY_CTX_free(ctx);
-  return ret;
+  // |KEM_check_key| performs the encapsulate/decapsulate consistency check this
+  // function documents above, and additionally validates the public and secret
+  // key encodings. The NULL checks above keep this function's stricter contract,
+  // which requires both halves of the key pair to be present.
+  return KEM_check_key(key->pkey.kem_key);
```
justsmth, line 545: The old EVP dispatch rejected non-KEM key types. This direct call bypasses that check, so `KEM_check_key` can interpret another algorithm's key bytes as a `KEM *`.

An X25519 public key filled with `0x41` previously returned an unsupported-type error. UBSan detects the invalid pointer dereference in the instrumented PR functions. A type check should precede any KEM union access, with a wrong-type regression test. Before the existing NULL checks:
```c
if (key != NULL && key->type != EVP_PKEY_KEM) {
  OPENSSL_PUT_ERROR(EVP, EVP_R_EXPECTING_A_KEM_KEY);
  return 0;
}
```

crypto/fipsmodule/kem/kem.c
jakemas: Fixed — KEM_check_key now locks the service indicator for its duration. Added a check to ServiceIndicatorTest.ML_KEM; verified on a FIPS build that it fails without the lock (approved reads APPROVED) and passes with it.

crypto/fipsmodule/kem/kem.c
jakemas: Fixed — SET_DIT_AUTO_RESET at entry to KEM_check_key.

crypto/fipsmodule/evp/p_kem.c
jakemas: Good catch, this segfaults rather than just tripping UBSan — exit 139 without the type check. Added the check before the union access, plus KEMTest.CheckKeyWrongKeyType covering X25519 and EC.

torben-hansen (approved):

justsmth (approved):

PR #2712, opened 2025-09-25
[DRAFT] Fix MUSL build

justsmth:
### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 

### Call-outs:

### Testing:


include/openssl/target.h
```diff
@@ -116,6 +116,20 @@
...
+#if defined(_GNU_SOURCE)
+#define GNU_SOURCE_IS_DEFINED
+#else
+#define _GNU_SOURCE
+#endif
+#include <features.h>
+#if !defined(__GLIBC__) && !defined(__USE_GNU)
+// MUSL does not provide an explicit macro for detection.
+// If it's not GLIBC (checked via __GLIBC__ or __USE_GNU), we assume MUSL.
+#define OPENSSL_LINUX_MUSL
```
davemo88, line 128: question about this - common for rust musl builds to use rustls instead of openssl, e.g. for statically linked binaries running in `scratch` docker images.

just want to make sure this doesn't imply presence of openssl somewhere downstream in the rust musl build process

justsmth: No, we just use these macros to guard platform-specific logic so it doesn't cause failures when being built for a different platform than it's intended for. (The "OPENSSL_xxx" macro names were inherited from BoringSSL, who inherited them from OpenSSL.)

davemo88: Thanks!

PR #2726, opened 2025-10-02
Move udiv and sencond tweak calculations to when needed

nebeid:
### Issues:
Addresses CryptoAlg-2941

### Description of changes: 
- We don't need to do `len DIV 0x50` until before the loop that processes 5 blocks at a time.
- We don't need the computation of the second block's tweak (iv) until we know we will use it.
  Note: that iv was not needed in the case of 1 block + k bytes, k < 16, i.e. cipher-stealing case of tail1x, because that iv is anyway computed before cipher-stealing, so it was actually computed twice before this move.


manastasova:
Also looks good to me!

PR #2733, opened 2025-10-08
Add build-time option to opt-out of CPU Jitter Entropy

torben-hansen:
### Issues:

D315146529

### Description of changes: 

Adds a build-time option `DISABLE_USAGE_OF_CPU_JITTER_ENTROPY` to opt-out of CPU Jitter Entropy. That is, with this option CPU Jitter Entropy will never be configured, the tree-DRBG will not be used. Furthermore, CPU Jitter Entropy will not even be included in any compilation units.

When using `DISABLE_USAGE_OF_CPU_JITTER_ENTROPY` we can no longer guarantee two entropy sources because only the Operating System type will be required to be available. rdrand and rndr will be used if supported in any given run-time, but it's not a guarantee. 

### Testing:

New stanza in build script executes a complete build and test with the option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,7 +112,7 @@ message(STATUS "PERFORM_SONAME_BUILD: ${PERFORM_SONAME_BUILD}")
 
 enable_language(C)
 
-if(DISABLE_USAGE_OF_CPU_JITTER_ENTROPY)
+if(DISABLE_CPU_JITTER_ENTROPY)
   if(FIPS)
     message(FATAL_ERROR "Cannot opt-out of CPU Jitter for the FIPS build")
   endif()
@@ -1124,7 +1124,7 @@ if(BUILD_TESTING)
 endif()
 
 # By default, include jitter entropy.
-if(NOT DISABLE_USAGE_OF_CPU_JITTER_ENTROPY)
+if(NOT DISABLE_CPU_JITTER_ENTROPY)
   add_subdirectory(third_party/jitterentropy)
 endif()
 
--- a/crypto/CMakeLists.txt
+++ b/crypto/CMakeLists.txt
@@ -620,7 +620,7 @@ function(build_libcrypto)
 
   add_library(${arg_NAME} $<TARGET_OBJECTS:crypto_objects> ${CRYPTO_FIPS_OBJECTS} ${arg_MODULE_SOURCE})
 
-  if(NOT DISABLE_USAGE_OF_CPU_JITTER_ENTROPY)
+  if(NOT DISABLE_CPU_JITTER_ENTROPY)
     target_sources(${arg_NAME} PRIVATE $<TARGET_OBJECTS:jitterentropy>)
   endif()
 


CMakeLists.txt
```diff
@@ -109,9 +110,20 @@ message(STATUS "PERFORM_SONAME_BUILD: ${PERFORM_SONAME_BUILD}")
...
 enable_language(C)
 
-# The validated entropy source will always be configured to be CPU Jitter
-# Entropy by default; because it's the root of the Tree DRBG.
-message(STATUS "Entropy source configured: Dynamic (default: CPU Jitter)")
+if(DISABLE_USAGE_OF_CPU_JITTER_ENTROPY)
+  if(FIPS)
+    message(FATAL_ERROR "Cannot opt-out of CPU Jitter for the FIPS build")
+  endif()
+  add_definitions(-DDO_NOT_USE_CPU_JITTER_ENTROPY)
```
justsmth: Should we align the name of the build flag with the name of the macro? 

We do for many others: `MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX`, `MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX`, `OPENSSL_SMALL`, `OPENSSL_NO_SSE2_FOR_TESTING`, ...

CMakeLists.txt
```diff
@@ -109,9 +110,20 @@ message(STATUS "PERFORM_SONAME_BUILD: ${PERFORM_SONAME_BUILD}")
...
+  message(STATUS "Entropy source configuration: CPU Jitter opt-out")
+  message(STATUS "Entropy source configured: Dynamic (default: Operating system)")
```
justsmth, line 121: You could bump this [from `STATUS` to `WARNING`](https://cmake.org/cmake/help/latest/command/message.html#general-messages) to increase visibility.

crypto/rand_extra/urandom_test.cc
```diff
@@ -27,7 +27,8 @@
     defined(OPENSSL_X86_64) && \
     !defined(BORINGSSL_SHARED_LIBRARY) && \
     !defined(BORINGSSL_UNSAFE_DETERMINISTIC_MODE) && \
-    defined(USE_NR_getrandom) && !defined(AWSLC_SNAPSAFE_TESTING)
+    defined(USE_NR_getrandom) && !defined(AWSLC_SNAPSAFE_TESTING) && \
+    !defined(DO_NOT_USE_CPU_JITTER_ENTROPY)
```
justsmth: How hard would it be to preserve this test even when the `DO_NOT_USE_CPU_JITTER_ENTROPY` flag is set? I see it uses some sophisticated `ptrace` mechanism to monitor process behavior -- perhaps we just need to change a few assumptions when this is set?

CMakeLists.txt
torben-hansen: Sure, I made a name change.

CMakeLists.txt
torben-hansen: It's informational IMO. Sticking to STATUS.

crypto/rand_extra/urandom_test.cc
torben-hansen: You are correct. I mean to do this after this PR. I was meant to write that in the description but forgot... my bad...

CMakeLists.txt
```diff
@@ -109,9 +110,20 @@ message(STATUS "PERFORM_SONAME_BUILD: ${PERFORM_SONAME_BUILD}")
...
 enable_language(C)
 
-# The validated entropy source will always be configured to be CPU Jitter
-# Entropy by default; because it's the root of the Tree DRBG.
-message(STATUS "Entropy source configured: Dynamic (default: CPU Jitter)")
+if(DISABLE_USAGE_OF_CPU_JITTER_ENTROPY)
+  if(FIPS)
+    message(FATAL_ERROR "Cannot opt-out of CPU Jitter for the FIPS build")
+  endif()
+  add_definitions(-DDISABLE_CPU_JITTER_ENTROPY)
```
justsmth: The intention is to align these both to be `DISABLE_CPU_JITTER_ENTROPY`.(?)

PR #2734, opened 2025-10-08
No need for round keys loading; they're still in registers.

nebeid:
### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 

### Call-outs:

### Testing:

### DIFF
--- a/third_party/s2n-bignum/s2n-bignum-to-be-imported/arm/aes/aes-xts-enc.S
+++ b/third_party/s2n-bignum/s2n-bignum-to-be-imported/arm/aes/aes-xts-enc.S
@@ -576,22 +576,22 @@ S2N_BN_SYMBOL(aes_hw_xts_encrypt):
         eor  v26.16b,v26.16b,v6.16b
 
         // Encrypt the composite block to get the last second encrypted text block
-        ldr  w6,[x3,#240]        // load key schedule...
-        ld1  {v0.16b},[x3],#16
-        sub  w6,w6,#2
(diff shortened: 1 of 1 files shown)


third_party/s2n-bignum/s2n-bignum-to-be-imported/arm/aes/aes-xts-dec.S
```diff
@@ -594,25 +594,23 @@ S2N_BN_SYMBOL(aes_hw_xts_decrypt):
...
-        aesimc    v26.16b,v26.16b
-        ld1    {v1.4s},[x3],#16     // load key schedule...
-        b.gt    .Loop_final_2nd_dec
-
-        aesd    v26.16b,v0.16b
-        aesimc    v26.16b,v26.16b
-        ld1    {v0.4s},[x3]
-        aesd    v26.16b,v1.16b
-        eor    v26.16b,v26.16b,v0.16b
+
```
manastasova, line 597: Is the number of loop iterations (in the original code) always 6? Would that allow, e.g., aes-128 support?

nebeid: Yes, the number of loop iterations, hence, of rounds has been constant in the original and the new implementations. The new one doesn't support AES-128-XTS. The original one may have supported it without any optimization but I didn't check or test it because we didn't (and still don't) support it via our API.

PR #2736, opened 2025-10-08
Simple script to build/run tests

justsmth:
### Description of changes: 
A simple script to build/run tests. I've been moving away from CLion and using the CLI more often for running builds/tests. This script simplifies that process.

### DIFF
--- a/util/build_and_test.sh
+++ b/util/build_and_test.sh
@@ -8,6 +8,7 @@ AWS_LC_BUILD="${BASE_DIR}/build"
 
 DEFAULT_CMAKE_FLAGS=("-DCMAKE_BUILD_TYPE=Debug" "-GNinja")
 
+BUILD_TARGET="all_tests"
 RUN_TARGET="run_tests"
 BUILD_ONLY=false
 CMAKE_ARGS=()
@@ -17,6 +18,10 @@ while [[ $# -gt 0 ]]; do
     --build-only)
       BUILD_ONLY=true
       shift
+      ;;
+    --build-target)
+      BUILD_TARGET="$2"
+      shift 2
       ;;
     --run-target)
       RUN_TARGET="$2"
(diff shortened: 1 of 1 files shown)


util/build_and_test.sh
```diff
@@ -0,0 +1,32 @@
...
+  fi
+done
+
+mkdir -p "${AWS_LC_BUILD}"
+
+cmake "${BASE_DIR}" -B "${AWS_LC_BUILD}" "${DEFAULT_CMAKE_FLAGS[@]}" "${CMAKE_ARGS[@]}"
+
+cmake --build "${AWS_LC_BUILD}" --target all_tests -j
+
+# Only run tests if --no-tests was not specified
```
skmcgrail:
```suggestion
# Only run tests if --build-only was not specified
```

util/build_and_test.sh
```diff
@@ -0,0 +1,32 @@
...
+    # Everything else gets passed to CMake
+    CMAKE_ARGS+=("$arg")
+  fi
+done
+
+mkdir -p "${AWS_LC_BUILD}"
+
+cmake "${BASE_DIR}" -B "${AWS_LC_BUILD}" "${DEFAULT_CMAKE_FLAGS[@]}" "${CMAKE_ARGS[@]}"
+
+cmake --build "${AWS_LC_BUILD}" --target all_tests -j
```
skmcgrail: Might be useful to override the default target to allow for explicitness? Not critical to change right now, but have a minor nit on the comment so feel free to ignore if you want.

util/build_and_test.sh
```diff
@@ -0,0 +1,42 @@
...
+      shift
+      ;;
+  esac
+done
+
+mkdir -p "${AWS_LC_BUILD}"
+
+cmake "${BASE_DIR}" -B "${AWS_LC_BUILD}" "${DEFAULT_CMAKE_FLAGS[@]}" "${CMAKE_ARGS[@]}"
+
+cmake --build "${AWS_LC_BUILD}" --target all_tests -j
```
skmcgrail:
```suggestion
cmake --build "${AWS_LC_BUILD}" --target ${RUN_TARGET} -j
```

Targets should probably be the same? Or maybe we need an option for build target too.....I guess I wasn't clear in my thoughts originally when I wrote it down. :| For simplicity I think they both can be the same target, but not sure if you want to change the name for the flag or anything.

justsmth: The first `cmake --build ...` command is intended to use a *build-only* target, like "all_tests" which builds the tests, but doesn't run them. The second `cmake --build ...` is intended to be a runnable-target (like "run_tests").

We could have a parameter for each of these.

PR #2746, opened 2025-10-14
Add Docker Image Build Workflows

skmcgrail:
### Description of changes: 
This just sets up the Linux x86 and ARM images (minus formal verification and Android) for now. To simplify I've setup new consolidated images in a new path from the old images. Next PR will transition the relevant workflows to use these images.

### DIFF
--- a/.github/docker_images/aws-lc/ubuntu/Dockerfile.2204
+++ b/.github/docker_images/aws-lc/ubuntu/Dockerfile.2204
@@ -1,0 +1,234 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+FROM public.ecr.aws/ubuntu/ubuntu:22.04 AS base
+
+SHELL ["/bin/bash", "-c"]
+
+ENV GOFLAGS=-buildvcs=false
+ENV DEBIAN_FRONTEND=noninteractive
+ENV DEPENDENCIES_DIR=/home/dependencies
+ENV ASAN_SYMBOLIZER_PATH=/usr/bin/llvm-symbolizer
+
+ENV BOOST_PACKAGE_NAME=boost_1_77_0
+ENV BOOST_TARBALL="${BOOST_PACKAGE_NAME}.tar.bz2"
+ENV BOOST_SRC_URL="https://d2yr98kym3baw0.cloudfront.net/${BOOST_TARBALL}"
+
+# llvm, llvm-dev, libcxx, and libcxxabi are needed for the sanitizer tests.
+# 11.1.0 is the latest stable release as of 2021-02-16.
+# See https://github.com/google/sanitizers/wiki/MemorySanitizerLibcxxHowTo
+RUN <<EOF
+set -ex
+apt-get update
+apt-get -y --no-install-recommends upgrade
+apt-get -y --no-install-recommends install \
+    autoconf \
+    autoconf-archive \
+    autogen \
+    automake \
+    binutils-dev \
+    bison \
+    build-essential \
+    ca-certificates \
+    check \
+    clang-14 \
+    clang++-14 \
+    clang-format \
+    cmake \
+    curl \
+    dnsutils \
+    doxygen \
+    dpkg-dev \
+    flex \
+    gcc-10 \
+    gcc-11 \
+    gcc-12 \
+    gcovr \
+    gettext \
+    git \
+    gnupg \
+    gnutls-dev \
+    gobject-introspection \
+    gperf \
+    groff \
+    g++-10 \
+    g++-11 \
+    g++-12 \
+    iproute2 \
+    jq \
+    lcov \
+    ldnsutils \
+    libcap-dev \
+    libcap-ng-dev \
+    libcmocka0 \
+    libcmocka-dev \
+    libcryptx-perl \
+    libcurl4-openssl-dev \
+    libev-dev \
+    libevent-dev \
+    libexpat1-dev \
+    libfstrm-dev \
+    libftdi-dev \
+    libgcrypt20-dev \
+    libglib2.0-dev \
+    libgmp-dev \
+    libicu-dev \
+    libini-config-dev \
+    libio-socket-ssl-perl \
+    libipc-run-perl \
+    libjemalloc-dev \
+    libjson-c-dev \
+    libjson-perl \
+    libltdl-dev \
+    liblua5.4-dev \
+    liblz4-dev \
+    liblzo2-dev \
+    libnet-dev \
+    libnet-ssleay-perl \
+    libnghttp2-dev \
+    libnl-3-dev \
+    libnl-genl-3-dev \
+    libpam-dev \
+    libpam0g-dev \
+    libpcap-dev \
+    libpcre2-dev \
+    libpcre3-dev \
+    libperl-dev \
+    libprocps-dev \
+    libprotobuf-c-dev \
+    libpsl-dev \
+    libreadline-dev \
+    libsqlite3-dev \
+    libssl-dev \
+    libsystemd-dev \
+    libtool \
+    libudev-dev \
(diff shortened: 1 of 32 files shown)


.github/docker_images/scripts/create-compiler-env.sh
```diff
@@ -0,0 +1,191 @@
...
+    local compiler_base="$1"  # e.g., "gcc" or "clang"
+    local cxx_base="$2"       # e.g., "g++" or "clang++"
+    local versions=()
+    
+    # Search for versioned compilers in standard paths
+    for search_path in "${SEARCH_PATHS[@]}"; do
+        if [[ -d "$search_path" ]]; then
+            # Find all versioned compilers (e.g., gcc-11, clang-14)
+            while IFS= read -r -d '' compiler_path; do
+                local basename_comp=$(basename "$compiler_path")
```
torben-hansen: The local command will mask return value of `basename` and ignore `-e` (I think?)

.github/docker_images/scripts/create-compiler-env.sh
```diff
@@ -0,0 +1,191 @@
...
+    # List created scripts
+    echo "Compiler environment setup complete!"
+    echo "Created scripts:"
+    for script in "${COMPILER_ENV_DIR}"/*.sh; do
+        if [[ -f "$script" ]]; then
+            echo "  $(basename "$script")"
+        fi
+    done
+    
+    echo "Usage: source /opt/compiler-env/setup-<compiler>.sh"
```
torben-hansen:
```suggestion
    echo "Usage: source ${COMPILER_ENV_DIR}/setup-<compiler>.sh"
```

.github/docker_images/scripts/create-compiler-env.sh
```diff
@@ -0,0 +1,191 @@
...
+    
+    if [[ ${#versions[@]} -gt 0 ]]; then
+        for version in "${versions[@]}"; do
+            if [[ -n "$version" ]]; then
+                create_compiler_script "setup-gcc-${version}.sh" "gcc-${version}" "g++-${version}" "GCC ${version}"
+            fi
+        done
+    fi
+    
+    # Handle base GCC compiler
```
torben-hansen, line 107: Would technically create scripts for the same compiler if `gcc` is symbolically linked to any of the versioned ones.

Probably not a blocker.

.github/docker_images/scripts/verify-compiler-version.sh
```diff
@@ -0,0 +1,169 @@
...
+        return 1
+    fi
+    
+    # Extract version number based on compiler type
+    local version=""
+    
+    # Check if it's G++ (g++ specifically) - check this first since g++ output contains both "g++" and "gcc"
+    if echo "$version_output" | grep -q "^g++"; then
+        if [[ "$expected_type" != "g++" ]]; then
+            echo "Error: Expected $expected_type compiler, but found g++" >&2
```
torben-hansen, line 44: clang++?

.github/docker_images/scripts/verify-compiler-version.sh
```diff
@@ -0,0 +1,169 @@
...
+            return 1
+        fi
+        # G++ format: "g++ (GCC) 4.8.5" or "g++ version 4.8.5"
+        version=$(echo "$version_output" | head -n1 | grep -oE '[0-9]+\.[0-9]+(\.[0-9]+)?' | head -n1)
```
torben-hansen, line 48: your second `head`s are redundant.

.github/docker_images/scripts/create-compiler-env.sh
skmcgrail: Yeah that was intentional, I wanted to support having a `setup-gcc.sh` if there was a top-level `gcc` command even if it points to `gcc-9` and we already setup a `setup-gcc-9.sh`. Though in all the workflows I've migrated for the PR to follow this I always explicitly choose the version compiler name variant.

.github/docker_images/scripts/verify-compiler-version.sh
skmcgrail: For clang++ it still outputs just `clang`.

```
$ clang++ --version
clang version 15.0.7 (AWS 15.0.7-3.amzn2023.0.4)
Target: x86_64-amazon-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

.github/docker_images/scripts/verify-compiler-version.sh
skmcgrail: Some compilers can have the version multiple times in the string, so this just grabs the first value. i.e. The below would output two lines both containing `11.5.0` because it appears twice in this string. So this grabs the first match and uses that for comparison.

```
g++ (GCC) 11.5.0 20240719 (Red Hat 11.5.0-5)
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

torben-hansen: ahh makes sense

.github/docker_images/aws-lc/ubuntu/Dockerfile.2204
```diff
@@ -0,0 +1,234 @@
...
+    autoconf-archive \
+    autogen \
+    automake \
+    binutils-dev \
+    bison \
+    build-essential \
+    ca-certificates \
+    check \
+    clang-14 \
+    clang++-14 \
```
justsmth, line 35: NP: Should we get other clang versions? https://documentation.ubuntu.com/ubuntu-for-developers/reference/availability/llvm

.github/docker_images/scripts/verify-compiler-version.sh
```diff
@@ -0,0 +1,169 @@
...
+        # GCC format: "gcc (GCC) 4.8.5" or "gcc version 4.8.5"
+        version=$(echo "$version_output" | head -n1 | grep -oE '[0-9]+\.[0-9]+(\.[0-9]+)?' | head -n1)
+    # Check if it's Clang
+    elif echo "$version_output" | grep -q "clang"; then
+        if [[ "$expected_type" != "clang" ]]; then
+            echo "Error: Expected $expected_type compiler, but found Clang" >&2
+            return 1
+        fi
+        # Clang format: "clang version 11.0.0" or "Apple clang version 12.0.0"
+        version=$(echo "$version_output" | head -n1 | grep -oE '[0-9]+\.[0-9]+(\.[0-9]+)?' | head -n1)
```
justsmth, line 64: NP: Should this limit the first component to at most 2 digits?  I've seen Apple's clang produce output that might match to "1500.1.0.2.5".


I collected a few compiler version test outputs here: https://github.com/aws/aws-lc-rs/blob/19f46c2891f97d794f11a3fac346502108cbff59/aws-lc-fips-sys/builder/cmake_builder.rs#L508-L512
```
("Apple clang version 14.0.0 (clang-1500.1.0.2.5)\n", (14, 0, 0)),
```

.github/docker_images/aws-lc/ubuntu/Dockerfile.2204
skmcgrail: I can follow-up in a new PR to add more into the image(s). Though I'm going back and forth whether we need to be extensively compiling for every compiler for a given distribution, but I think that's more of a broader discussion we need to have around our build matrix.

.github/docker_images/scripts/verify-compiler-version.sh
skmcgrail: Yeah this should be capturing the first number group only with the current head filter where it grabs the first match line. But the passed in "check value" / "expected value" can be any number of components. So we can say `14` or `14.0` or `14.0.0` and it works. Right now I just wanted to at least capture major+minor as those minor bumps in a compiler have been known to break us from time to time.

PR #2750, opened 2025-10-14
Implementation of XAES-256-GCM with EVP_CIPHER 

ttunglee:
### **Issues:**

1. Add XAES-256-GCM, which is extended AES-256-GCM with a derived key mode proposed by [Filippo Valsorda](https://words.filippo.io/xaes-256-gcm/) in 2023, followed by [a specification](https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM.md) released in 2024. 
2. This implementation supports EVP_CIPHER API, allows configuration to use either FIPS-compliant CMAC available in AWS-LC, or an optimized CMAC dedicated to the specific use case of XAES-256-GCM from https://github.com/aws/aws-lc/pull/2652. 
3. Support varying nonce sizes: 20 ≤ b ≤ 24 based on the extension: https://eprint.iacr.org/2025/758.pdf#page=24 

### **Description of Changes**
Implementation for API EVP_CIPHER of XAES-256-GCM is appended to [e_aes.c](https://github.com/ttungle96/aws-lc/blob/xaes-256-gcm/crypto/fipsmodule/cipher/e_aes.c#L1750), and the tests are appended to [cipher_test.cc](https://github.com/ttungle96/aws-lc/blob/xaes-256-gcm/crypto/cipher_extra/cipher_test.cc#L1459). 

### **Testing**
We use the test vectors provided here:  
https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM.md
https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM/openssl/openssl.c
https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM/go/XAES-256-GCM_test.go

### **Evaluation**
Performance evaluation with speedtool:
Performance_Evaluation


crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1455,308 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
   CHECK_ERROR(EVP_DecryptUpdate(ctx.get(), out_vec.data(), &out_len, in_vec.data(), in_len), ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
   CHECK_ERROR(EVP_DecryptFinal(ctx.get(), out_vec.data(), &out_len), ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
 }
+
+TEST(CipherTest, XAES_256_GCM_EVP_CIPHER) {
```
dkostic: I'm not sure I can follow all the tests this function is performing. Maybe you could break it into smaller test functions?

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1455,308 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
   CHECK_ERROR(EVP_DecryptUpdate(ctx.get(), out_vec.data(), &out_len, in_vec.data(), in_len), ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
   CHECK_ERROR(EVP_DecryptFinal(ctx.get(), out_vec.data(), &out_len), ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
 }
+
```
dkostic, line 1462: can we use `CipherFileTest` for test vectors?

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1455,308 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+        size_t plaintext_len = strlen((const char *)plaintext);
+
+        ciphertext.resize(20);
+
+        // XAES-256-GCM Encryption
+        bssl::UniquePtr<EVP_CIPHER_CTX> ctx(EVP_CIPHER_CTX_new());
+        ASSERT_TRUE(ctx);
+        ASSERT_TRUE(EVP_CipherInit_ex(ctx.get(), EVP_xaes_256_gcm(), NULL, NULL, NULL, 1));
+
+        // Invalid nonce size
```
dkostic: please write down the valid range for nonce size.

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1455,308 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+            0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
+            0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01};
+        const uint8_t nonce[24] = {
+            0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42,
+            0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42};
+
+        const uint8_t *plaintext = (const uint8_t *)"Hello, XAES-256-GCM!";
+        size_t plaintext_len = strlen((const char *)plaintext);
+
+        ciphertext.resize(20);
```
dkostic: why this?

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1455,308 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
   CHECK_ERROR(EVP_DecryptUpdate(ctx.get(), out_vec.data(), &out_len, in_vec.data(), in_len), ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
   CHECK_ERROR(EVP_DecryptFinal(ctx.get(), out_vec.data(), &out_len), ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
 }
+
+TEST(CipherTest, XAES_256_GCM_EVP_CIPHER) {
+    {
+        std::vector<uint8_t> ciphertext, tag, decrypted; 
+        const uint8_t key[32] = {
```
dkostic: why not std::vector<uint8_t>?

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1455,308 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+            0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
+            0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
+            0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01};
+        const uint8_t nonce[24] = {
+            0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42,
+            0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42};
+
+        const uint8_t *plaintext = (const uint8_t *)"Hello, XAES-256-GCM!";
```
dkostic: why not std::vector<uint8_t> ?

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1455,308 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+        // Encryption with 256-bit key
+        ctx.get()->key_len = 32;
+        ASSERT_TRUE(EVP_CipherInit_ex(ctx.get(), NULL, NULL, key, nonce, -1));
+        
+        int ciphertext_len = 0;
+        ASSERT_TRUE(EVP_CipherUpdate(ctx.get(), (uint8_t*)ciphertext.data(), &ciphertext_len, 
+                    plaintext, plaintext_len));
+        
+        int len = 0;
+        ASSERT_TRUE(EVP_CipherFinal_ex(ctx.get(), (uint8_t*)ciphertext.data() + ciphertext_len, &len));
```
dkostic: why `(uint8_t*)ciphertext.data() + ciphertext_len`?

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1455,308 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+        ciphertext_len += len;
+        
+        std::vector<uint8_t> output;
+
+        // Empty and invalid data
+        ASSERT_TRUE(ConvertToBytes(&output, ""));
```
dkostic: I'm not sure what are we testing here with all these `ConvertToBytes`?

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1455,308 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+        ASSERT_FALSE(ConvertToBytes(&output, "1z"));
+        ASSERT_FALSE(ConvertToBytes(&output, "abc"));
+        ASSERT_FALSE(ConvertToBytes(&output, "ABC"));
+        ASSERT_TRUE(ConvertToBytes(&output, "ABCD"));
+        ASSERT_FALSE(ConvertToBytes(&output, "GHIJ"));
+        ASSERT_FALSE(ConvertToBytes(&output, "GHIJK"));
+        ASSERT_FALSE(ConvertToBytes(&output, "ghijk")); 
+        
+        ASSERT_TRUE(ConvertToBytes(&output, "01e5f78bc99de880bd2eeff2870d361f0eab5b2f"));
+        ASSERT_EQ(Bytes(ciphertext), Bytes(output));
```
dkostic: ok, so it looks like you are implementing a KAT. as I mentioned above, I think we could use the infrastructure around `CipherFileTest` to avoid writing our own KAT implemetation.

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1455,308 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+        
+        tag.resize(16);
+        ASSERT_TRUE(EVP_CIPHER_CTX_ctrl(ctx.get(), EVP_CTRL_AEAD_GET_TAG, tag.size(), (void*)tag.data()));
+        ASSERT_TRUE(ConvertToBytes(&output, "c55268f34b14045878fe3668db980319"));
+        ASSERT_EQ(Bytes(tag), Bytes(output));
```
dkostic: what's tested here?

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1746,167 @@ int EVP_has_aes_hardware(void) {
...
+// ----------- XAES-256-GCM Auxiliary Data Structures and Subroutines -----------
+// ------------------------------------------------------------------------------
+
+#define XAES_256_GCM_CTX_OFFSET      (sizeof(EVP_AES_GCM_CTX) + EVP_AES_GCM_CTX_PADDING)
+#define XAES_256_GCM_KEY_COMMIT_SIZE (AES_BLOCK_SIZE * 2)
+#define XAES_256_GCM_CMAC_INPUT_SIZE (AES_BLOCK_SIZE * 2)
+#define USE_OPTIMIZED_CMAC                  
+
+struct xaes_256_gcm_ctx {
+#ifdef USE_OPTIMIZED_CMAC
```
dkostic: I think we'll need a bit more documentation here. What is "optimized cmac", etc.? Also, why don't we always use optimized cmac, why do we need non-optimized cmac?

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1746,167 @@ int EVP_has_aes_hardware(void) {
...
+    AES_KEY xaes_key; 
+    uint8_t k1[AES_BLOCK_SIZE]; 
+#else 
+    uint8_t xaes_key[AES_BLOCK_SIZE * 2];
+#endif 
+};
+
+struct xaes_256_gcm_key_commit_ctx {
```
dkostic: this can be removed.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1746,167 @@ int EVP_has_aes_hardware(void) {
...
+#ifdef USE_OPTIMIZED_CMAC 
+    AES_KEY xaes_key; 
+    uint8_t k1[AES_BLOCK_SIZE]; 
+#else 
+    uint8_t xaes_key[AES_BLOCK_SIZE * 2];
+#endif 
+    uint8_t kc[XAES_256_GCM_KEY_COMMIT_SIZE];
+};
+
+#define BINARY_FIELD_MUL_X_128(out, in)             \
```
dkostic, line 1797: please add a reference for this algorithm.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1746,167 @@ int EVP_has_aes_hardware(void) {
 }
 
 OPENSSL_MSVC_PRAGMA(warning(pop))
+
+// ------------------------------------------------------------------------------
+// ----------- XAES-256-GCM Auxiliary Data Structures and Subroutines -----------
```
dkostic: add reference to the spec please.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1746,167 @@ int EVP_has_aes_hardware(void) {
...
+    AES_encrypt(M1, derived_key, &xaes_ctx->xaes_key);
+    AES_encrypt(M2, derived_key + AES_BLOCK_SIZE, &xaes_ctx->xaes_key);
+#else 
+    AES_CMAC(derived_key, xaes_ctx->xaes_key, AES_BLOCK_SIZE * 2, M1, AES_BLOCK_SIZE);
+    AES_CMAC(derived_key + AES_BLOCK_SIZE, xaes_ctx->xaes_key, AES_BLOCK_SIZE * 2, M2, AES_BLOCK_SIZE);
+#endif 
+    return 1;
+}
+
+static int xaes_256_gcm_set_gcm_key(EVP_CIPHER_CTX *ctx, const uint8_t *nonce, int enc) {
```
dkostic, line 1844: we need documentation here to explain nonce being NULL vs. non-NULL, the IV length rquirements, etc.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1746,167 @@ int EVP_has_aes_hardware(void) {
...
+    if(nonce == NULL) {
+        return aes_gcm_init_key(ctx, NULL, NULL, enc);
+    }
+
+    EVP_AES_GCM_CTX *gctx = aes_gcm_from_cipher_ctx(ctx);
```
dkostic: we can't get the IV length without `EVP_AES_GCM_CTX`?

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1746,167 @@ int EVP_has_aes_hardware(void) {
...
+
+    xaes_256_gcm_CMAC_derive_key(xaes_ctx, nonce, derived_key);
+    
+    int ivlen = gctx->ivlen;
+    
+    gctx->ivlen = 12;
+    
+    aes_gcm_init_key(ctx, derived_key, nonce + ivlen - 12, enc);
+
+    gctx->ivlen = ivlen;
```
dkostic, line 1871: we need explanation for these magic ivlen operations

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1746,167 @@ int EVP_has_aes_hardware(void) {
...
+    
+    if(gctx->ivlen < 20 || gctx->ivlen > 24) {
+        OPENSSL_PUT_ERROR(CIPHER, CIPHER_R_INVALID_NONCE_SIZE);
+        return 0;
+    }
+    
+    struct xaes_256_gcm_ctx *xaes_ctx =
+        (struct xaes_256_gcm_ctx *)((uint8_t*)ctx->cipher_data + XAES_256_GCM_CTX_OFFSET);
+
+    uint8_t derived_key[(256 >> 3)];
```
dkostic: magic numbers?

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1746,167 @@ int EVP_has_aes_hardware(void) {
...
+    gctx->ivlen = 12;
+    
+    aes_gcm_init_key(ctx, derived_key, nonce + ivlen - 12, enc);
+
+    gctx->ivlen = ivlen;
+
+    return 1;
+}
+
+static const uint8_t kZeroIn[AES_BLOCK_SIZE] = {0};
```
dkostic: we can move inside the function it's using it.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1746,167 @@ int EVP_has_aes_hardware(void) {
...
+
+static int xaes_256_gcm_ctx_init(struct xaes_256_gcm_ctx *xaes_ctx, 
+                            const uint8_t *key, const unsigned key_len) {
+    if ((key_len << 3) != 256) {
```
dkostic: magic numbers?

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1746,167 @@ int EVP_has_aes_hardware(void) {
...
+    char *ptr = ctx->cipher_data;
+#if defined(OPENSSL_32_BIT)
+    assert((uintptr_t)ptr % 4 == 0);
+    ptr += (uintptr_t)ptr & 4;
+#endif
+    assert((uintptr_t)ptr % 8 == 0);
+    ptr += (uintptr_t)ptr & 8;
+
+    OPENSSL_memcpy(ptr, gctx, sizeof(EVP_AES_GCM_CTX));
+    OPENSSL_free(temp);
```
dkostic: why?

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1746,167 @@ int EVP_has_aes_hardware(void) {
...
+    return xaes_256_gcm_ctx_init(xaes_ctx, key, ctx->key_len);
+}
+
+// ------------------------------------------------------------------------------
+// --------------- EVP_CIPHER XAES-256-GCM Without Key Commitment ---------------
+// ------------------------------------------------------------------------------
+
+static int xaes_256_gcm_init(EVP_CIPHER_CTX *ctx, const uint8_t *key,
+                            const uint8_t *iv, int enc) {
+    if(key != NULL && !xaes_256_gcm_init_common(ctx, key,
```
dkostic: I think this is a bug -- if key is NULL than you'll fall through and call `xaes_256_gcm_set_gcm_key`, but I don't think you want to do that, do you?

crypto/cipher_extra/cipher_test.cc
ttunglee: I copied the test case directly here: https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM/openssl/openssl.c

crypto/cipher_extra/cipher_test.cc
ttunglee: I copied the test case directly here https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM/openssl/openssl.c

crypto/fipsmodule/cipher/e_aes.c
ttunglee: - When key is NOT NULL, it invokes xaes_256_gcm_init_common() to initializes xaes_256_gcm_ctx with the main key. 
- When key is NULL but nonce is NOT NULL, it invokes xaes_256_gcm_set_gcm_key() to derive a subkey, then invokes aes_gcm_init_key() to initialize the underlying AES-GCM with the subkey. 
- Function aes_gcm_init_key() is called every time EVP_CipherInit_ex is called, and even when key is NULL. See this: https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/cipher/e_aes.c#L369, it only does nothing if both key and iv are NULL. And since the flag EVP_CIPH_ALWAYS_CALL_INIT is set for AES-GCM (see https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/cipher/e_aes.c#L971), it always calls this init function every time when EVP_CipherInit_ex is called (see https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/cipher/cipher.c#L232). The idea is to maintain the behaviors of XAES-256-GCM the same as those of underlying AES-GCM.

crypto/fipsmodule/cipher/e_aes.c
ttunglee: - cipher_data is allocated memory space of the same size as the context of AES-GCM. But we need to have space for storing xaes_256_gcm_ctx. I allocate another memory space to accommodate both the existing context of underlying AES-GCM and also struct xaes_256_gcm_ctx. The reason that I cannot change the context size of xaes_256_gcm_ctx right in the beginning is because of this: https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/cipher/e_aes.c#L356, which prevents the initialization of underlying AES-GCM when wrapping it inside XAES-256-GCM. Furthermore, EVP_CIPHER *cipher in ctx is declared as const (https://github.com/aws/aws-lc/blob/main/include/openssl/cipher.h#L622), so I cannot change ctx->cipher->ctx_size once it was initialized. 
- The alignment part is for compatibility with this https://github.com/aws/aws-lc/blob/xaes-256-gcm/crypto/fipsmodule/cipher/e_aes.c#L360

crypto/fipsmodule/cipher/e_aes.c
ttunglee: I'll fix this. n << 3 ~ n * 2^3.

crypto/fipsmodule/cipher/e_aes.c
ttunglee: I'll fix this

crypto/fipsmodule/cipher/e_aes.c
ttunglee: - AES-GCM uses 96-bit nonce size but XAES-256-GCM uses 192-bit (or 20 -> 24 bytes per https://eprint.iacr.org/2025/758.pdf#page=24) nonce size. 
- aes_gcm_init_key() is for initializing the underlying EVP_AES_GCM_CTX of AES-GCM to do encryption/decryption, so I set the ivlen to its standard size of AES-GCM (12-byte), then put it the original nonce size of XAES-256-GCM after that (20 -> 24 bytes). 
- The part (nonce + ivlen - 12) is from the extension here: https://eprint.iacr.org/2025/758.pdf#page=24, which serves the cases using < 24-byte nonce size.

crypto/fipsmodule/cipher/e_aes.c
ttunglee: I'll fix this. 256 >> 3 = 256 / 2^3 = 32

crypto/fipsmodule/cipher/e_aes.c
ttunglee: Since EVP_CIPHER_CTX does not have ivlen field?

crypto/fipsmodule/cipher/e_aes.c
ttunglee: Just for comparison and performance test. You can comment/uncomment to choose whether to use optimized CMAC or CMAC already provided by AWS-LC. Should I remove and only keep optimized CMAC?

crypto/cipher_extra/cipher_test.cc
ttunglee: It is for comparison with the ciphertext output I obtained here: https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM/openssl/openssl.c

crypto/cipher_extra/cipher_test.cc
ttunglee: I can use CipherFileTest. But how about the loop test over 10,000 and 1,000,000 random data generated by hash function here: https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM/go/XAES-256-GCM_test.go. Should I keep it like that in my current code?

crypto/cipher_extra/cipher_test.cc
ttunglee: This is just to increase the test coverage in the report. I'll remove these.

crypto/cipher_extra/cipher_test.cc
ttunglee: I'm not sure what you mean, redundant casting (uint8_t*) or ciphertext.data() + ciphertext_len? For the latter, it is the same as https://github.com/aws/aws-lc/blob/main/crypto/cipher_extra/cipher_test.cc#L345.

crypto/cipher_extra/cipher_test.cc
ttunglee: Thanks! I'll follow your suggestion!

crypto/fipsmodule/cipher/e_aes.c
ttunglee: Thanks! I'll add reference.

crypto/fipsmodule/cipher/e_aes.c
ttunglee: Thank you! I'll add it. 
It was explained in the description of my PR: https://github.com/aws/aws-lc/pull/2750 
The algorithm is from here: https://github.com/aws/aws-lc/pull/2652. It's not something special, just to left-shift a 128-bit register as described by the line #2 of XAES-256-GCM here: https://words.filippo.io/xaes-256-gcm/

crypto/cipher_extra/cipher_test.cc
ttunglee: Current tests w/o using CipherFileTest can achieve better test coverage, since I can manually put test cases for all possible cases, similar to this test: https://github.com/aws/aws-lc/blob/main/crypto/cipher_extra/cipher_test.cc#L1117. I can switch to using CipherFileTest, but it might not reach and cover all edge cases that only appear in XAES-256-GCM if using CipherFileTest.

crypto/cipher_extra/cipher_test.cc
ttunglee: I'll fix this

crypto/cipher_extra/cipher_test.cc
dkostic: I think we should use the test files for the test vectors that we have and then we can use this approach for edge cases. How does that sound?

crypto/cipher_extra/cipher_test.cc
dkostic: in aws-lc, we tend to use C++ style for unit tests so let's please convert to that.

crypto/cipher_extra/cipher_test.cc
dkostic: yes, I have two questions:
- why do we need the cast?
- why do we need the `+ ciphertext_len`?

crypto/cipher_extra/cipher_test.cc
dkostic: we can use the test files for KATs and implement a unit test for X random encrypt/decrypt. I'm not sure what's the right number of random tests though, I don't want to significantly increase the runtime of `crypto_test`...

crypto/cipher_extra/cipher_test.cc
dkostic: ok, I think we need a different approach here. If I understand correctly you have used the code in https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM/openssl/openssl.c to generate a test vector and then you hard-coded this test vector in this unit test. I would suggest we do it differently:
1. generate X test vectors with the code from C2SP (and I'd suggest to use the Go implementation instead of the one that depends on OpenSSL)
2. decide if we can use the existing CipherFileTest infrastructure for these test vectors
3. if not, implement a unit test for XAES-256-GCM that goes over test vectors in a file and performs the required tests.

crypto/fipsmodule/cipher/e_aes.c
dkostic: what's the performance difference between the two?

crypto/fipsmodule/cipher/e_aes.c
dkostic: please add this explanation to the code.

crypto/fipsmodule/cipher/e_aes.c
dkostic: > cipher_data is allocated memory space of the same size as the context of AES-GCM.

where is this allocated? why can't we allocate enough space in the first place?

crypto/fipsmodule/cipher/e_aes.c
dkostic: I'm not sure I completely understand the explanation but I would suggest we simplify the code and add comments explaining what is done when. We can start by adding this at the beginning of the function:
```
 if (iv == NULL && key == NULL) {
    return;
  }
```

crypto/fipsmodule/cipher/e_aes.c
ttunglee: I already explained it above: 

"The reason that I cannot change the context size of xaes_256_gcm_ctx right in the beginning is because of this: https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/cipher/e_aes.c#L356, which prevents the initialization of underlying AES-GCM when wrapping it inside XAES-256-GCM."

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1750,154 @@ int EVP_has_aes_hardware(void) {
...
+    // Key length: 32 bytes
+    if (ctx->key_len != XAES_256_GCM_KEY_LENGTH) {
+        OPENSSL_PUT_ERROR(CIPHER, CIPHER_R_BAD_KEY_LENGTH);
+        return 0;
+    }
+
+    struct xaes_256_gcm_ctx *xaes_ctx =
+            (struct xaes_256_gcm_ctx*)((uint8_t*)ctx->cipher_data + XAES_256_GCM_CTX_OFFSET);
+
+    // Initialize the main key
```
dkostic: I'm a bit too dumb to figure out what is happening with different key and iv values. Can you please simplify? For example, it seems to me that we can do
```
if (key == NULL) {
  if (iv == NULL) ...
} else {
  if (xaes_256_gcm_ctx_init(...))...
}

File: crypto/fipsmodule/cipher/e_aes.c [hunk 24]
```diff
@@ -1745,3 +1750,154 @@ int EVP_has_aes_hardware(void) {
...
+    return 1;
+}
+
+static int xaes_256_gcm_set_gcm_key(EVP_CIPHER_CTX *ctx, const uint8_t *nonce, int enc) {
+
+    EVP_AES_GCM_CTX *gctx = aes_gcm_from_cipher_ctx(ctx);
+
+    // Nonce size: 20 bytes <= |N| <= 24 bytes
+    if(gctx->ivlen < XAES_256_GCM_MIN_NONCE_SIZE || 
+    gctx->ivlen > XAES_256_GCM_MAX_NONCE_SIZE) {
```
dkostic:
```suggestion
    if(gctx->ivlen < XAES_256_GCM_MIN_NONCE_SIZE || 
       gctx->ivlen > XAES_256_GCM_MAX_NONCE_SIZE) {
```

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1750,154 @@ int EVP_has_aes_hardware(void) {
...
+    struct xaes_256_gcm_ctx *xaes_ctx =
+        (struct xaes_256_gcm_ctx *)((uint8_t*)ctx->cipher_data + XAES_256_GCM_CTX_OFFSET);
+
+    uint8_t derived_key[XAES_256_GCM_KEY_LENGTH];
+
+    xaes_256_gcm_CMAC_derive_key(xaes_ctx, nonce, derived_key);
+
+    int ivlen = gctx->ivlen;
+
+    // AES-GCM uses 12-byte nonce
```
dkostic:
```suggestion
    // AES-GCM uses a different size nonce than XAES-GCM,
    // so to be able to call aes_gcm_init_key with ctx we temporarily
    // set the nonce (iv) length to AES_GCM_NONCE_LENGTH.
```

include/openssl/nid.h
nebeid, line 1: - Was this file modified by `objects.go` by editing `objects.txt`as in https://github.com/aws/aws-lc/blob/main/crypto/obj/README?

crypto/test/test_util.h
```diff
@@ -34,6 +34,7 @@
 // hexdump writes |msg| to |fp| followed by the hex encoding of |len| bytes
 // from |in|.
 void hexdump(FILE *fp, const char *msg, const void *in, size_t len);
+bool ConvertToBytes(std::vector<uint8_t> *out, const std::string &value);
```
nebeid: Can you please add comments on what the function does and the two cases it handles?

crypto/obj/obj_dat.h
```diff
@@ -11711,6 +11726,7 @@ static const uint16_t kNIDsInOIDOrder[] = {
     901 /* 2.16.840.1.101.3.4.1.46 (OBJ_aes_256_gcm) */,
     902 /* 2.16.840.1.101.3.4.1.47 (OBJ_aes_256_ccm) */,
     903 /* 2.16.840.1.101.3.4.1.48 (OBJ_id_aes256_wrap_pad) */,
+    999 /* 2.16.840.1.101.3.4.1.49 (OBJ_xaes_256_gcm) */,
```
nebeid, line 11729: We're coming up with the OID or it was introduced by the author?

crypto/cipher_extra/cipher_test.cc
nebeid: Was this (or the original) test function broken down?
If yes, to which functions?

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1750,154 @@ int EVP_has_aes_hardware(void) {
...
+    uint8_t k1[AES_BLOCK_SIZE]; 
+#else 
+    uint8_t xaes_key[XAES_256_GCM_KEY_LENGTH];
+#endif 
+};
+
+/* 
+Left-shift a 128-bit register: https://words.filippo.io/xaes-256-gcm/ (line 2)
+If MSB₁(L) = 0: K1 = L << 1;
+Else: K1 = (L << 1) ⊕ (0x00, ..., 0x00, 0x87)
```
nebeid: I prefer we don't have non-ASCII characters in the comments.

crypto/fipsmodule/cipher/e_aes.c
nebeid: - I think the reference should be to the particular step in the original CMAC algorithm in its spec. The one to Filippo's can be in addition. 
- Also, it's better to state why we only need K1
- Would this be enclosed in the define of the optimised algorithm? (if we keep a define for it as per my comment above).

crypto/fipsmodule/cipher/e_aes.c
nebeid: It's good to know the answer to this question before we decide.

I personally think we won't be keeping both, unless we want to make a separate EVP API for each (using original CMAC call and using necessary-only CMAC steps).
If users want to build using existing APIs, they can do for example what's in the speed test `SpeedXAES256GCM` in this PR https://github.com/aws/aws-lc/pull/2652.

On another note, we won't be using a compile flag to use either one, so at some point we need to decide and/or create two APIs as mentioned.

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1459,96 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+        ASSERT_TRUE(ctx);
+        ASSERT_TRUE(EVP_CipherInit_ex(ctx.get(), EVP_xaes_256_gcm(), NULL, NULL, NULL, 1));
+        ASSERT_TRUE(EVP_CIPHER_CTX_ctrl(ctx.get(), EVP_CTRL_AEAD_SET_IVLEN, 24, NULL));
+
+        bssl::UniquePtr<EVP_CIPHER_CTX> dctx(EVP_CIPHER_CTX_new());
+        ASSERT_TRUE(dctx);
+        ASSERT_TRUE(EVP_DecryptInit_ex(dctx.get(), EVP_xaes_256_gcm(), NULL, NULL, NULL));
+        ASSERT_TRUE(EVP_CIPHER_CTX_ctrl(dctx.get(), EVP_CTRL_AEAD_SET_IVLEN, 24, NULL));
+
+        std::vector<uint8_t> key(32), nonce(24), plaintext_len(1), plaintext(256);
```
dkostic: I don't think `plaintext_len` should be of type `std::vector<uint8_t>`. it's just length so probably `size_t`? same for `aad_len`.

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1459,96 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+
+            // XAES-256-GCM Decryption
+            ASSERT_TRUE(EVP_DecryptInit_ex(dctx.get(), NULL, NULL, key.data(), nonce.data()));
+            ASSERT_TRUE(EVP_CIPHER_CTX_ctrl(dctx.get(), EVP_CTRL_AEAD_SET_TAG, tag_size, tag.data()));
+
+            std::vector<uint8_t> decrypted;
+            decrypted.resize(plaintext_len[0]);
+            len = 0;
+            EVP_DecryptUpdate(dctx.get(), NULL, &len, aad.data(), aad_len[0]);
+            ASSERT_TRUE(EVP_DecryptUpdate(dctx.get(), (uint8_t*)decrypted.data(), &len, ciphertext.data(), ciphertext_len));
```
dkostic: is the cast necessary?

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -353,9 +354,13 @@ static EVP_AES_GCM_CTX *aes_gcm_from_cipher_ctx(EVP_CIPHER_CTX *ctx) {
 
   // |malloc| guarantees up to 4-byte alignment on 32-bit and 8-byte alignment
   // on 64-bit systems, so we need to adjust to reach 16-byte alignment.
-  assert(ctx->cipher->ctx_size ==
-         sizeof(EVP_AES_GCM_CTX) + EVP_AES_GCM_CTX_PADDING);
-
+  if(ctx->cipher->iv_len > 12) {
+    // Placeholder for XAES-256-GCM
```
dkostic: is something missing here? what's the placeholder for?

include/openssl/nid.h
ttunglee: Yeah. Dusan showed me how to add it.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1750,147 @@ int EVP_has_aes_hardware(void) {
...
+Else: K1 = (L << 1) ⊕ (0x00, ..., 0x00, 0x87)
+*/
+#define BINARY_FIELD_MUL_X_128(out, in)             \
+do {                                                \
+    unsigned i;                                     \
+    for (i = 0; i < 15; i++) {                      \
+        out[i] = (in[i] << 1) | (in[i+1] >> 7);     \
+    }                                               \
+    const uint8_t carry = in[0] >> 7;               \
+    out[i] = (in[i] << 1) ^ ((0 - carry) & 0x87);   \
```
dkostic, line 1806: is this supposed to be constant-time? because the compiler could probably turn this code into a branch. If it has to be constant-time then we'll have to use the constant-time functions from `crypto/internal.h`

crypto/test/test_util.h
ttunglee: It is this function: https://github.com/aws/aws-lc/blob/main/crypto/test/file_test.cc#L341
I separate it there to use in my test since this method is declared as private in class FileTest.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1750,146 @@ int EVP_has_aes_hardware(void) {
...
+    AES_encrypt(M1, derived_key, &xaes_ctx->xaes_key);
+    AES_encrypt(M2, derived_key + AES_BLOCK_SIZE, &xaes_ctx->xaes_key);
+
+    return 1;
+}
+
+static int xaes_256_gcm_set_gcm_key(EVP_CIPHER_CTX *ctx, const uint8_t *nonce, int enc) {
+
+    struct xaes_256_gcm_ctx *xaes_ctx =
+        (struct xaes_256_gcm_ctx *)aes_gcm_from_cipher_ctx(ctx);
```
dkostic: I don't think we should abuse `aes_gcm_from_cipher_ctx`, we should implement a new function for xaes.

crypto/obj/obj_dat.h
ttunglee: I made it up for temporary usage.

crypto/fipsmodule/cipher/e_aes.c
ttunglee: Thank you. I'll repalce ⊕ with ^

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1750,146 @@ int EVP_has_aes_hardware(void) {
...
+    AES_set_encrypt_key(key, XAES_256_GCM_KEY_LENGTH << 3, &xaes_ctx->xaes_key);
+    AES_encrypt(kZeroIn, L, &xaes_ctx->xaes_key);
+    BINARY_FIELD_MUL_X_128(xaes_ctx->k1, L);
+    return 1;
+}
+
+// ------------------------------------------------------------------------------
+// --------------- EVP_CIPHER XAES-256-GCM Without Key Commitment ---------------
+// ------------------------------------------------------------------------------
+static int xaes_256_gcm_init(EVP_CIPHER_CTX *ctx, const uint8_t *key,
```
dkostic, line 1888: please add a comment that says something along these lines:
```
when |key| is provided the function initializes the context and derives a subkey.
when |key| is not provided, but |iv| is then...
```

crypto/fipsmodule/cipher/e_aes.c
ttunglee: Since the ctx size of xaes-256-gcm is no longer `sizeof(EVP_AES_GCM_CTX) + EVP_AES_GCM_CTX_PADDING)`, it is `sizeof(struct xaes_256_gcm_ctx) + EVP_AES_GCM_CTX_PADDING` instead? How to avoid this to prevent test case failed due to 
`assert(ctx->cipher->ctx_size == sizeof(EVP_AES_GCM_CTX) + EVP_AES_GCM_CTX_PADDING)`
?

crypto/fipsmodule/cipher/e_aes.c
ttunglee: I reuse to avoid code duplicate. I can implement another for xaes, but it just copies the entire aes_gcm_from_cipher_ctx, and replace the final `return (EVP_AES_GCM_CTX *)ptr;` by `return (struct xaes_256_gcm_ctx *)ptr;`

crypto/cipher_extra/cipher_test.cc
ttunglee: It does not matter. I'll change

crypto/fipsmodule/cipher/e_aes.c
ttunglee: I assume you would like it to be constant-time function to avoid leakage the value of in[0], which could be 0 or 1. But could you suggest how to use functions in `crypto/internal.h` to do that? In internal.h, I see this function but for 64-bit numbers:
`
// constant_time_msb_w returns the given value with the MSB copied to all the
// other bits.
static inline crypto_word_t constant_time_msb_w(crypto_word_t a) {
    return 0u - (a >> (sizeof(a) * 8 - 1));
}
`

The way it works is totally similar the current code: `(0 - carry)`, with `carry = in[0] >> 7`.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1756,154 @@ int EVP_has_aes_hardware(void) {
...
+    XAES_256_GCM_CTX *xaes_ctx = xaes_256_gcm_from_cipher_ctx(ctx);
+    
+    // When main key is provided, initialize the context and derive a subkey  
+    if(key != NULL) { 
+        if(!xaes_256_gcm_ctx_init(xaes_ctx, key)) {
+            return 0;
+        }
+    }
+
+    // If iv is provided, even if main key is not, derive a subkey
```
dkostic, line 1903: do we have a unit test that makes sure that we:
- When main key is provided, initialize the context and derive a subkey  
- If iv is provided, even if main key is not, derive a subkey?

ttunglee: I had that test previously. However, CipherFileTest does not support that kind of test so I removed it. Currently, speedtool only initializes the main key once, and measure the performance with varying nonces, i.e., different subkeys are derived from the main key in speedtool test.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1756,153 @@ int EVP_has_aes_hardware(void) {
...
 
 OPENSSL_MSVC_PRAGMA(warning(pop))
+
+/* ---------------------------- XAES-256-GCM ----------------------------
+Specification: https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM.md 
+-----------------------------------------------------------------------*/
+#define XAES_256_GCM_CTX_OFFSET      (sizeof(EVP_AES_GCM_CTX) + EVP_AES_GCM_CTX_PADDING)
+#define XAES_256_GCM_KEY_LENGTH      (AES_BLOCK_SIZE * 2)
+#define XAES_256_GCM_KEY_COMMIT_SIZE (AES_BLOCK_SIZE * 2)
+#define XAES_256_GCM_CMAC_INPUT_SIZE (AES_BLOCK_SIZE * 2)
```
nebeid: I thought this would be
```suggestion
#define XAES_256_GCM_CMAC_INPUT_SIZE (AES_BLOCK_SIZE)
```

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1756,153 @@ int EVP_has_aes_hardware(void) {
 }
 
 OPENSSL_MSVC_PRAGMA(warning(pop))
+
+/* ---------------------------- XAES-256-GCM ----------------------------
+Specification: https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM.md
```
nebeid: Since we're allowing the nonce size to vary, we need to reference the description that allows a nonce smaller than 24 bytes. I see you have it further below on l. 1849, but it would be good to add that reference here too.

crypto/fipsmodule/cipher/e_aes.c
dkostic: let's get that test back in, we don't have to use `CipherFileTest` for it.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1756,153 @@ int EVP_has_aes_hardware(void) {
...
+    }
+
+    AES_encrypt(M1, derived_key, &xaes_ctx->xaes_key);
+    AES_encrypt(M2, derived_key + AES_BLOCK_SIZE, &xaes_ctx->xaes_key);
+
+    return 1;
+}
+
+static XAES_256_GCM_CTX *xaes_256_gcm_from_cipher_ctx(EVP_CIPHER_CTX *ctx) { 
+    // XAES_256_GCM_CTX data is put at the rear of ctx->cipher_data
```
dkostic: where? I can't find this. and why would we do it like that, why not store the context at ctx->cipher_data?

crypto/obj/objects.txt
```diff
@@ -893,7 +893,8 @@ aes 44			: AES-256-CFB		: aes-256-cfb
 aes 45			: id-aes256-wrap
 aes 46			: id-aes256-GCM		: aes-256-gcm
 aes 47			: id-aes256-CCM		: aes-256-ccm
-aes 48			: id-aes256-wrap-pad
+aes 48			: id-aes256-wrap-pad 
+aes 49			: id-xaes256-GCM    : xaes-256-gcm
```
nebeid, line 898: I think we should add a comment here that this is a new OID we're introducing.

crypto/test/test_util.h
nebeid: Thank you, can you add this comment here with a TODO so we can decide later whether we want to make it public in file_test.h?

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1459,98 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+            ASSERT_TRUE(EVP_CipherInit_ex(ctx.get(), NULL, NULL, key.data(), nonce.data(), -1));
+            ASSERT_EQ(aad_len, EVP_Cipher(ctx.get(), NULL, aad.data(), aad_len));
+            int ciphertext_len = 0;
+            ASSERT_TRUE(EVP_CipherUpdate(ctx.get(), ciphertext.data(), &ciphertext_len, 
+                        plaintext.data(), plaintext_len));
+
+            int len = 0;
+            ASSERT_TRUE(EVP_CipherFinal_ex(ctx.get(), ciphertext.data() + ciphertext_len, &len));
+            ciphertext_len += len;
+            ASSERT_TRUE(EVP_DigestUpdate(d.get(), ciphertext.data(), ciphertext_len));
```
nebeid, line 1643: What is this line for as one as l.1532? It would be good if the flow here has comments as it is not very clear to me.

crypto/cipher_extra/test/cipher_tests.txt
```diff
@@ -506,6 +506,32 @@ AAD = feedfacedeadbeeffeedfacedeadbeefabaddad2
 Tag = a44a8266ee1c8eb0c8b5d4cf5ae9f19b
 Operation = InvalidDecrypt
 
+# Source of test vectors: 
+# https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM.md 
+Cipher = XAES-256-GCM
+Key = 0101010101010101010101010101010101010101010101010101010101010101
```
nebeid, line 512: I assume you ensured these test vectors are being used, e.g. by inserting an error in one of them.

crypto/fipsmodule/cipher/e_aes.c
ttunglee: It will fail a test case MAC non-fips 16. I cannot get a docker to debug the error as there is no available docker link in that test. I guest there needs to have some padding space after EVP_AES_GCM_CTX to pass that test. If I add a field just to add space for passing that test, there will be more confusions.  

Why cannot do like that? What is the error that may occur by doing `(XAES_256_GCM_CTX *)((uint8_t*)ctx->cipher_data + XAES_256_GCM_CTX_OFFSET);`?

crypto/fipsmodule/cipher/e_aes.c
ttunglee: This const is no longer used as I removed the code using CMAC available in AWS-LC. I'll remove it.

crypto/obj/objects.txt
ttunglee: Thanks! I'll add a comment like you suggested.

crypto/fipsmodule/cipher/e_aes.c
ttunglee: The test has been put back in cipher_test.cc

crypto/fipsmodule/cipher/e_aes.c
ttunglee: I'll add it. It is a simple extension in the case without key commitment. We use the first 12 bytes of nonce for key derivation, and the bytes at [b-12: b] for IV in encryption, where 20 <= b <= 24 is the size of the input nonce.

crypto/cipher_extra/test/cipher_tests.txt
ttunglee: They are being used in cipher_test.cc. I verified like what you said. I edited the test case to make it different and when I run test, it shows errors and fails the test.

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1459,186 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+        ASSERT_TRUE(EVP_DecryptInit_ex(dctx.get(), nullptr, nullptr, nullptr, iv.data()));
+        
+        // Test encryption and decryption again with anotherother plaintext
+        plaintext = (const uint8_t *)"XAES-256-GCM";
+        ConvertToBytes(&ciphertext, "ce546ef63c9cc60765923609");
+        ConvertToBytes(&tag, "b33a9a1974e96e52daf2fcf7075e2271");
+        test(plaintext, strlen((const char *)plaintext), ciphertext, tag);
+    }   
+
+    // Source of multi-loop tests:
```
nebeid, line 1602: - Are there KATs (known-answer tests) that test the streaming API, i.e. calling CipherUpdate multiple times? 
- are the shorter nonces tested at least with a self-consistency test, i.e. we can decrypt what we encrypt?
- what does multi-loop here stand for?

crypto/cipher_extra/cipher_test.cc
ttunglee: I already added the reference. The source of this test is at:
https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM/go/XAES-256-GCM_test.go

I just rewrite it in C++. It uses hash function SHAKE128 to generate data. Also, verifying the ciphertext output is done by that hash function. It verifies the ciphertext output of each iteration, then compares the final digest output after 10K iterations, 1M iterations. 

EVP_DigestUpdate(d.get(), ciphertext.data(), ciphertext_len);
It accumulates `ciphertext.data()` of `ciphertext_len` bytes into the digest output.

crypto/cipher_extra/cipher_test.cc
ttunglee: - As you see in my implementation, all functions of aes-gcm are reused, except the init one (aes_gcm_init is replaced by xaes_256_gcm_init). Invoking CipherUpdate again will have the same behavior as the first time it is called. The KATs available with too small plaintext sizes, so I do not need to call it multiple times. Maybe speedtool tests that behavior, and it is still fine. 
- I do not have KATs with shorter nonces. I can write tests if required. I'm not sure whether that is necessary since it is a small change: 
From: 
`aes_gcm_init_key(ctx, derived_key, nonce + AES_GCM_NONCE_LENGTH, enc);`
To: 
`aes_gcm_init_key(ctx, derived_key, nonce + ivlen - AES_GCM_NONCE_LENGTH, enc);`

- I tried to get all KATs available. Multi-loop tests are from: https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM/go/XAES-256-GCM_test.go. I just rewrite it in C/C++.

crypto/fipsmodule/cipher/e_aes.c
ttunglee: I added all recommended comments.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -353,9 +354,19 @@ static EVP_AES_GCM_CTX *aes_gcm_from_cipher_ctx(EVP_CIPHER_CTX *ctx) {
...
-
+  switch(ctx->cipher->nid) { 
+    // only execute assert checking with aes-gcm
+    case NID_aes_128_gcm:
+    case NID_aes_192_gcm:
+    case NID_aes_256_gcm:
+        assert(ctx->cipher->ctx_size ==
+            sizeof(EVP_AES_GCM_CTX) + EVP_AES_GCM_CTX_PADDING);
+        break;
+    // not execute assert checking with xaes-256-gcm
```
dkostic: I still don't understand why would we call this function with XAES?

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1756,153 @@ int EVP_has_aes_hardware(void) {
...
+}
+
+DEFINE_METHOD_FUNCTION(EVP_CIPHER, EVP_xaes_256_gcm) {
+    OPENSSL_memset(out, 0, sizeof(EVP_CIPHER));
+    out->nid = NID_xaes_256_gcm;
+    out->block_size = AES_BLOCK_SIZE;
+    out->key_len = XAES_256_GCM_KEY_LENGTH;
+    out->iv_len = XAES_256_GCM_MAX_NONCE_SIZE;
+    out->ctx_size = sizeof(EVP_AES_GCM_CTX) + EVP_AES_GCM_CTX_PADDING 
+                + sizeof(XAES_256_GCM_CTX);
```
dkostic: is this correct? why are we not storing everything in `XAES_256_GCM_CTX`?

crypto/fipsmodule/cipher/e_aes.c
dkostic: can you please show me where you "put data at the rear of ctx->cipher_data"? If storing the data where it should be stored (ctx->cipher_data) is causing a test failure then we should investigate what's the issue before implementing a work around.

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1459,186 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
   CHECK_ERROR(EVP_DecryptUpdate(ctx.get(), out_vec.data(), &out_len, in_vec.data(), in_len), ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
   CHECK_ERROR(EVP_DecryptFinal(ctx.get(), out_vec.data(), &out_len), ERR_R_SHOULD_NOT_HAVE_BEEN_CALLED);
 }
+
+TEST(CipherTest, XAES_256_GCM_EVP_CIPHER) {
```
dkostic: can you please separate out these tests:
- Invalid nonce and key length
- deriving many subkeys from a main key but with different nonces
- multi-loop test

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1459,186 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+        ASSERT_TRUE(EVP_CIPHER_CTX_ctrl(ctx.get(), EVP_CTRL_AEAD_SET_IVLEN, 24, nullptr));
+
+        // Valid key length: 32 bytes 
+        // Test invalid key length 
+        ctx.get()->key_len = 24;
+        ASSERT_FALSE(EVP_CipherInit_ex(ctx.get(), nullptr, nullptr, key.data(), nonce.data(), -1));
+    }
+    
+    /* 
+    * Test deriving many subkeys from a main key but with different nonces
```
dkostic, line 1500: is there a simpler way to implement this test? Maybe it doesn't have to be a KAT?

crypto/fipsmodule/cipher/e_aes.c
ttunglee: Line 1901, the context size is defined as:
`out->ctx_size = sizeof(EVP_AES_GCM_CTX) + EVP_AES_GCM_CTX_PADDING 
                + sizeof(XAES_256_GCM_CTX); `
It means I would like to allocate more space for cipher_data to accommodate a struct XAES_256_GCM_CTX, apart from EVP_AES_GCM_CTX of standard AES-GCM. 

I would like to reuse the function: aes_gcm_from_cipher_ctx, as this function is also called in many other places (cipher, ctrl, etc.). For this reason, XAES_256_GCM_CTX should be put at the end of cipher_data. 

If I follow your code, I declare a struct that includes EVP_AES_GCM_CTX inside, then write another function similar to aes_gcm_from_cipher_ctx to specify the start position of EVP_AES_GCM_CTX. Then I also have to see why a test case failed? Is that too cumbersome?

crypto/fipsmodule/cipher/e_aes.c
ttunglee: aes_gcm_from_cipher_ctx() is to specify the position of EVP_AES_GCM_CTX. It is used in many other places: cipher, ctrl, etc.

In this function, there is check: 
`assert(ctx->cipher->ctx_size ==
            sizeof(EVP_AES_GCM_CTX) + EVP_AES_GCM_CTX_PADDING);`

The context size is no longer true with xaes-256-gcm, since we need more space for XAES_256_GCM_CTX, apart from EVP_AES_GCM_CTX and EVP_AES_GCM_CTX_PADDING. 

So I rewrite the code like this to temporarily bypass this check for XAES-256-GCM, so that tests will not fail in CI.

crypto/fipsmodule/cipher/e_aes.c
ttunglee: It is just style. Like I explained, I can declare a field of type EVP_AES_GCM_CTX inside XAES_256_GCM_CTX. Then, like in previous review, you require another function similar to aes_gcm_from_cipher_ctx to specify the start position of EVP_AES_GCM_CTX. Then I also have to see why a test case failed? Is that too cumbersome?

crypto/cipher_extra/cipher_test.cc
ttunglee: What is the complication that needs to be simplified here? I declare one context for encryption, one context for decryption, one lambda expression for testing encryption and decryption with input contexts, and test with two different nonces?

crypto/cipher_extra/cipher_test.cc
ttunglee: Done

crypto/fipsmodule/cipher/e_aes.c
dkostic: my question is why do you use `aes_gcm_from_cipher_ctx` for XAES?

crypto/fipsmodule/cipher/e_aes.c
dkostic: I think the whole issue stems from the fact that you are trying to use an AES specific function for XAES. We should not do that, instead, we should define a new XAES specific function.

crypto/fipsmodule/cipher/e_aes.c
ttunglee: That function is used by `aes_gcm_init_key`, `aes_gcm_cipher`, and `aes_gcm_ctrl`. I do not use `aes_gcm_from_cipher_ctx` explicitly. I reuse `aes_gcm_init_key`, `aes_gcm_cipher`, `aes_gcm_ctrl`, and `aes_gcm_cleanup` for XAES-256-GCM.

crypto/fipsmodule/cipher/e_aes.c
ttunglee: XAES-256-GCM is AES-GCM in nature. Reusing already-verified functions from AES-GCM makes sense. The only difference is a KDF, which receives a 24-byte nonce and a main key to derive a subkey, which is used to encrypt/decrypt with the underlying AES-GCM. Please elaborate specific ideas that you expect me to do.

Look at this implementation with EVP_AEAD:
https://github.com/aws/aws-lc/pull/2652/files

All functions from AES-GCM are reused for XAES-256-GCM.

crypto/fipsmodule/cipher/e_aes.c
nebeid: The way I understand it is that it differs from the AEAD implementation in #2652 in that, here, in the EVP API, we want to be able to do multiple cipher calls (complete with a tag, not streaming) using the same derived key but new (second half?) IV. In the AEAD implementation, every "seal" or "open" operation had a new IV that derived a new key so the gcm context was temporary. So here, the `cipher` call is pointing to `aes_gcm_cipher` and uses a gcm context. 
I think it would be clearer if that context were part of the `xaes` context as dkostic suggests.
On another note, can the interface take a half IV for GCM and not re-derive the key? Or I'm wrong in assuming that that use-case is implemented?

ttunglee: @nebeid I have included `EVP_AES_GCM_CTX` inside `XAES_256_GCM_CTX` in the newest update. Please review again based on that. This will make the code longer since in EVP_AEAD implementation, XAES_256_GCM_CTX does not need EVP_AES_GCM_CTX inside. I will have to define another struct for AEAD case, then another aead_xaes_256_gcm_ctx_init(), which is totally the same as xaes_256_gcm_ctx_init(), but different by only one argument. Similarly for aead_xaes_256_gcm_CMAC_derive_key(), totally the same as xaes_256_gcm_CMAC_derive_key().

For your second question, currently it has not implemented yet since people have not agreed on that design. Furthermore, I tried to keep this PR as simple as possible for review. If that case is required, I'll fix and create another PR.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1745,3 +1773,152 @@ int EVP_has_aes_hardware(void) {
...
+}
+
+static XAES_256_GCM_CTX *xaes_256_gcm_from_cipher_ctx(EVP_CIPHER_CTX *ctx) { 
+    char *ptr = ctx->cipher_data;
+#if defined(OPENSSL_32_BIT)
+    assert((uintptr_t)ptr % 4 == 0);
+    ptr += (uintptr_t)ptr & 4;
+#endif
+    assert((uintptr_t)ptr % 8 == 0);
+    ptr += (uintptr_t)ptr & 8;
```
dkostic, line 1840: do we need this for `XAES_256_GCM_CTX`?

ttunglee: I tried without it and many tests failed. Would you like me to temporarily remove it and you can have a look at tests?

dkostic: no need, can you just add a comment to say that we handle the alignment according to the way it's implemented for the AES-GCM context?

dkostic (approved):

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1459,198 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+            ASSERT_TRUE(EVP_DigestSqueeze(s.get(), key.data(), 32));
+            ASSERT_TRUE(EVP_DigestSqueeze(s.get(), nonce.data(), 24));
+            ASSERT_TRUE(EVP_DigestSqueeze(s.get(), &plaintext_len, 1));
+            ASSERT_TRUE(EVP_DigestSqueeze(s.get(), plaintext.data(), plaintext_len));
+            ASSERT_TRUE(EVP_DigestSqueeze(s.get(), &aad_len, 1));
+            ASSERT_TRUE(EVP_DigestSqueeze(s.get(), aad.data(), aad_len));
+
+            // XAES-256-GCM Encryption
+            ASSERT_TRUE(EVP_CipherInit_ex(ctx.get(), nullptr, nullptr, key.data(), nonce.data(), -1));
+            ASSERT_EQ(aad_len, EVP_Cipher(ctx.get(), nullptr, aad.data(), aad_len));
```
nebeid, line 1635: This assert is questionable.; it compares the return value of a function with the length of the aad. `ASSERT_TRUE` on the `EVP_Cipher` call is what I believe you intended.

crypto/test/test_util.h
nebeid: Sorry, I should have realised that you could probably use `DecodeHex` instead of exposing this function. It's used for example in https://github.com/aws/aws-lc/blob/d826708c3526889c6bb451063e62e1a03cd51629/crypto/fipsmodule/evp/evp_ctx_test.cc#L413 Would you be able to try it?

crypto/cipher_extra/cipher_test.cc
nebeid: This accumulation is necessary to match the final result in that linked file, right?

crypto/cipher_extra/cipher_test.cc
nebeid: I don't see tests for the following usecases, unless you're planning on doing them in a subsequent PR (or point me to them if I missed them):
- Are there KATs (known-answer tests) that test the streaming API, i.e. calling CipherUpdate multiple times?
- are the shorter nonces tested at least with a self-consistency test, i.e. we can decrypt what we encrypt?

crypto/cipher_extra/cipher_test.cc
ttunglee: I follow this: https://github.com/aws/aws-lc/blob/main/crypto/cipher_extra/cipher_test.cc#L258 
Should I need to change it to ASSERT_TRUE instead?

crypto/cipher_extra/cipher_test.cc
ttunglee: Yes. It is correct! The output of each iteration is accumulated into the final result of hash.

crypto/test/test_util.h
ttunglee: Yes. I removed ConvertToBytes() and replace it by that function. Thank you very much!

crypto/cipher_extra/cipher_test.cc
ttunglee: I added test cases. 
- Streaming APIs:
https://github.com/ttungle96/aws-lc/blob/xaes-256-gcm/crypto/cipher_extra/cipher_test.cc#L1539
https://github.com/ttungle96/aws-lc/blob/xaes-256-gcm/crypto/cipher_extra/cipher_test.cc#L1567
- Shorter nonces: 
https://github.com/ttungle96/aws-lc/blob/xaes-256-gcm/crypto/cipher_extra/cipher_test.cc#L1672

dkostic (approved):

crypto/cipher_extra/cipher_test.cc
nebeid: Thank you for the reference. I noticed this explanation https://github.com/aws/aws-lc/blob/93f9f73350b82a5dc088ff56087fe32826c139d9/include/openssl/cipher.h#L466-L475. It's a confusing function, but we can leave it as is.

nebeid (approved):
(Pending CI)

PR #2754, opened 2025-10-17
Support NetBSD

justsmth:
### Issues:
Addresses: aws/aws-lc-rs#932

### Description of changes: 
Adds library support for NetBSD.

### Call-outs:
* For aarch64 CPU feature detection, I adapted the logic I found here: https://github.com/NetBSD/src/blob/62c785e59d064070166dab5d2a4492055effba89/sys/arch/aarch64/aarch64/cpu.c#L363
* Both shared and static FIPS builds were failing for NetBSD.
* One (of the nine) OCSP "AmazonTrustServices" tests was failing -- the one that attempts connection to "ocsp.sectigo.com:80". The failure indicated `No route to host`. I suspect that it's attempting to connect via IPv6 and not finding appropriate routing for it. I disabled this test for NetBSD.

### Testing:
Adds tests for both x86-64 and aarch64.

### DIFF
--- a/crypto/fipsmodule/cpucap/cpu_aarch64_netbsd.c
+++ b/crypto/fipsmodule/cpucap/cpu_aarch64_netbsd.c
@@ -50,7 +50,7 @@ void OPENSSL_cpuid_setup(void) {
   int found_cpu = 0;
 
   // Query up to 256 CPUs (arbitrary but reasonable upper limit)
-  for (int cpu_num = 0; cpu_num < 256; cpu_num++) {
+  for (size_t cpu_num = 0; cpu_num < 256; cpu_num++) {
     if (get_cpu_id(cpu_num, &cpu_id) < 0) {
       // Failed to read this CPU - either it doesn't exist or is offline
       if (found_cpu > 0) {
@@ -126,6 +126,8 @@ void OPENSSL_cpuid_setup(void) {
       ID_AA64PFR0_EL1_DIT_IMPL) {
     OPENSSL_armcap_P |= (ARMV8_DIT | ARMV8_DIT_ALLOWED);
   }
+
+  OPENSSL_cpucap_initialized = 1;
 }
 
 #endif  // OPENSSL_AARCH64 && OPENSSL_NETBSD && !OPENSSL_STATIC_ARMCAP
--- a/crypto/rand_extra/urandom.c
+++ b/crypto/rand_extra/urandom.c
@@ -30,11 +30,6 @@
 #include <time.h>
 #include <unistd.h>
 
-#if defined(OPENSSL_NETBSD)
-#include <sys/ioctl.h>
-#include <sys/rndio.h>
-#endif
-
 #if defined(OPENSSL_LINUX)
 #if defined(AWS_LC_URANDOM_NEEDS_U32)
   // On old Linux OS: unknown type name '__u32' when include <linux/random.h>.
--- a/crypto/fipsmodule/cpucap/cpu_aarch64_freebsd.c
+++ b/crypto/fipsmodule/cpucap/cpu_aarch64_freebsd.c
@@ -56,6 +56,8 @@ void OPENSSL_cpuid_setup(void) {
   if (ID_AA64ISAR0_SHA2_VAL(id_aa64isar0) >= ID_AA64ISAR0_SHA2_512) {
     OPENSSL_armcap_P |= ARMV8_SHA512;
   }
+
+  OPENSSL_cpucap_initialized = 1;
 }
 
 #endif  // OPENSSL_AARCH64 && OPENSSL_FREEBSD && !OPENSSL_STATIC_ARMCAP
--- a/crypto/fipsmodule/cpucap/cpu_aarch64_openbsd.c
+++ b/crypto/fipsmodule/cpucap/cpu_aarch64_openbsd.c
@@ -53,6 +53,8 @@ void OPENSSL_cpuid_setup(void) {
 
   if (ID_AA64ISAR0_SHA2(cpu_id) >= ID_AA64ISAR0_SHA2_512)
     OPENSSL_armcap_P |= ARMV8_SHA512;
+
+  OPENSSL_cpucap_initialized = 1;
 }
 
 #endif  // OPENSSL_AARCH64 && OPENSSL_OPENBSD && !OPENSSL_STATIC_ARMCAP


crypto/fipsmodule/cpucap/cpu_aarch64_netbsd.c
MartinHusemann, line 1: I think you want machdep.cpu0.cpu_id, and pass a pointer to this struct  [aarch64_sysctl_cpu_id](https://nxr.netbsd.org/xref/src/sys/arch/aarch64/include/armreg.h#1894).
I suppose you are aware of different capabilities of different CPUs in a single SoC? So "cpu0" hardcoded may not be enough

justsmth: > I think you want machdep.cpu0.cpu_id, and pass a pointer to this struct [aarch64_sysctl_cpu_id](https://nxr.netbsd.org/xref/src/sys/arch/aarch64/include/armreg.h#1894).

Thanks!

> I suppose you are aware of different capabilities of different CPUs in a single SoC? So "cpu0" hardcoded may not be enough

Yeah, that might be problematic if we pick a particular CPU to detect capabilities with, when not all CPUs provide the same capabilities. After this value is initialized, our library assumes the cpu features are consistent for the lifetime of the process. (Where needed, our library can also be built with the [cpu capabilities determined statically by the build environment](https://github.com/aws/aws-lc/blob/7e6ef7b972e9f82f06ffbf39b05d00b273aa0069/crypto/fipsmodule/cpucap/cpucap.c#L44-L86).)

What would you recommend for this logic? One option would be for us to assume a very small aarch64 baseline (maybe just `ARMV7_NEON`?), and then require consumers to build with the `OPENSSL_STATIC_ARMCAP` options if they would like to unlock better performing implementations? What do you think?

MartinHusemann: Idealy you could find the CPU with the least capabilities and use that, but detecting the situation and using a minimum baseline sounds like a practical alternative.

justsmth: Hello!  In the latest commit the logic iterates over the available CPUs, taking the intersection of their capabilities. This seems like the safest way to enable all appropriate CPU features. 

Your feedback on this code is welcome. Please let us know if you foresee any potential issues with the latest approach. Thanks!

crypto/fipsmodule/cpucap/cpu_aarch64_netbsd.c
```diff
@@ -0,0 +1,131 @@
...
+  // directly, so it reflects that specific core's capabilities, not a
+  // system-wide minimum.
+
+  // Initialize with all features enabled (we'll AND them together)
+  uint64_t common_aa64isar0 = UINT64_MAX;
+  uint64_t common_aa64pfr0 = UINT64_MAX;
+  int found_cpu = 0;
+
+  // Query up to 256 CPUs (arbitrary but reasonable upper limit)
+  for (int cpu_num = 0; cpu_num < 256; cpu_num++) {
```
torben-hansen:
```suggestion
  for (size_t cpu_num = 0; cpu_num < 256; cpu_num++) {
```
For sanity? Feel free to ignore my pedantic :)

crypto/fipsmodule/cpucap/cpu_aarch64_netbsd.c
```diff
@@ -0,0 +1,131 @@
...
+    if (get_cpu_id(cpu_num, &cpu_id) < 0) {
+      // Failed to read this CPU - either it doesn't exist or is offline
+      if (found_cpu > 0) {
+        // We've already found at least one CPU, so assume we've enumerated all
+        break;
```
torben-hansen, line 58: The assumption is that the remaining CPUs can't be probed or it's some optimisation through simplification?

crypto/fipsmodule/cpucap/cpu_aarch64_netbsd.c
torben-hansen: What if the process is pinned to a subset of CPUs with the maximal set of capabilities? I my simple mental model of a SOC, CPUs, cores that's possible. But my mental model might not be correct on that(?).

This might not be a problem is practice, but at least wanted to test my mental model.

crypto/fipsmodule/cpucap/cpu_aarch64_netbsd.c
```diff
@@ -0,0 +1,131 @@
...
+    OPENSSL_armcap_P |= ARMV8_RNG;
+  }
+
+  // Check for DIT (Data Independent Timing) support across all cores
+  if (__SHIFTOUT(common_aa64pfr0, ID_AA64PFR0_EL1_DIT) >=
+      ID_AA64PFR0_EL1_DIT_IMPL) {
+    OPENSSL_armcap_P |= (ARMV8_DIT | ARMV8_DIT_ALLOWED);
+  }
+}
+
```
torben-hansen, line 132: missing support for mutating capabilities at runtime?

torben-hansen: and you need to set `OPENSSL_cpucap_initialized`

crypto/rand_extra/urandom.c
```diff
@@ -30,6 +30,11 @@
 #include <time.h>
 #include <unistd.h>
 
+#if defined(OPENSSL_NETBSD)
```
torben-hansen: This file should not be part of a compilation unit when building for NETBSD. That's decided by the change in `crypto/rand_extra/internal.h`.

tests/ci/run_bsd_tests.sh
```diff
@@ -6,6 +6,11 @@ set -ex
 
 source tests/ci/common_posix_setup.sh
 
+# Our NetBSD CI environment gives a "No route to host" error when connecting to `ocsp.sectigo.com:80`.
```
torben-hansen, line 9: Weird...

tests/ci/run_bsd_tests.sh
```diff
@@ -41,8 +46,11 @@ build_and_test -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=1
 echo "Testing AWS-LC static library in release mode."
 build_and_test -DCMAKE_BUILD_TYPE=Release
 
-echo "Testing AWS-LC shared library in FIPS Release mode."
-fips_build_and_test -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=1
+# The FIPS builds fail on NetBSD
```
torben-hansen, line 49: Because `OPENSSL_cpucap_initialized` is not set? What the reason it fails?

crypto/ube/fork_detect.c
```diff
@@ -114,7 +114,7 @@ static int init_fork_detect_wipeonfork(void *addr, long page_size) {
 #endif // defined(OPENSSL_LINUX)
 
 
-#if defined(OPENSSL_FREEBSD) || defined(OPENSSL_OPENBSD)
+#if defined(OPENSSL_FREEBSD) || defined(OPENSSL_OPENBSD) || defined(OPENSSL_NETBSD)
```
torben-hansen, line 117: Seems like we soon need a `OPENSSL_BSD_FLAVOR`... Something for another day.

crypto/fipsmodule/cpucap/cpu_aarch64_netbsd.c
justsmth: This is assuming that the CPUs are enumerated across a contiguous set of numbers, and that once an unavailable CPU is reached, no other CPUs will be found at the higher numbers. I guess we could always iterate over the 256 of them, but that seems like overkill.

crypto/fipsmodule/cpucap/cpu_aarch64_netbsd.c
justsmth: > missing support for mutating capabilities at runtime?

We support updating capabilities after initialization?

justsmth: > and you need to set OPENSSL_cpucap_initialized

🤦‍♂️ It looks like we don't do that for FreeBSD nor OpenBSD. I'll update all 3 files.

tests/ci/run_bsd_tests.sh
justsmth: * For the static build it looked like the delocator was unable to parse a specific combination of assembly instruction.
* For the shared build there was a linker error, but I forget the details.

torben-hansen: Added to P333208296

PR #2763, opened 2025-10-23
Implement -passin for dgst cli

nhatnghiho:
### Description of changes: 
Add -passin to dgst cli now that #2555 is merged in.

### Testing:
Unit tests

### DIFF
--- a/tool-openssl/dgst.cc
+++ b/tool-openssl/dgst.cc
@@ -27,8 +27,7 @@ static const argument_t kArguments[] = {
     {"-hmac", kOptionalArgument,
      "Create a hashed MAC with the corresponding key"},
     {"-sigopt", kDuplicateArgument, "Signature parameter in n:v form"},
-    // TODO: Implement -passin
-    // {"-passin", kOptionalArgument, "Input file pass phrase source"},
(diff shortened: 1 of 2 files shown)


tool-openssl/dgst.cc
```diff
@@ -37,6 +36,7 @@ static const argument_t kArguments[] = {
     {"", kOptionalArgument, ""}};
 
 static bool LoadPrivateKey(const std::string &key_file_path,
+                           bssl::UniquePtr<std::string> &passin,
```
samuel40791765, line 39: Nit: I don't think we need to wrap `std::string` with a `UniquePtr` here. `UniquePtr`s are used for custom memory allocations/deallocations, but `std::string` is already designed to manage its own memory automatically.

Unfortunately, I noticed that this is a recent pattern throughout the rest of the code base. Will probably open another PR to clean things up.

nhatnghiho: Yeah this pattern did strike me as a little odd. We did have someone pointed this out before and here is Jonathon's reply https://github.com/aws/aws-lc/pull/2555#discussion_r2238160513

PR #2775, opened 2025-10-29
Implement more options for req CLI

nhatnghiho:
### Issues:
Resolves #CryptoAlg-3380

### Description of changes: 
Implement the following options for x509 cli command:

- sha256
- config
- key
- passin
- extensions
- outform

### Testing:
Unit tests added, including missing coverage for some existing x509 options.

### DIFF
--- a/tool-openssl/req.cc
+++ b/tool-openssl/req.cc
@@ -559,7 +559,7 @@ static bool AddCertExtensions(X509 *cert, CONF *req_conf,
         "basicConstraints=critical,CA:true\n";
 
     // Create a BIO for the config
-    bssl::UniquePtr<BIO> bio(BIO_new_mem_buf(default_config, -1));
+    bssl::UniquePtr<BIO> bio(BIO_new_mem_buf(default_config, strlen(default_config)));
     if (!bio) {
       fprintf(stderr, "Failed to create memory BIO\n");
       return false;
--- a/tool-openssl/test_util.cc
+++ b/tool-openssl/test_util.cc
@@ -157,6 +157,10 @@ bool CompareCSRs(X509_REQ *csr1, X509_REQ *csr2) {
     return false;
   }
 
+  if(X509_NAME_cmp(name1, name2) != 0) {
+    return true;
+  }
+
   int count1 = X509_NAME_entry_count(name1);
   int count2 = X509_NAME_entry_count(name2);
   if (count1 < 0 || count2 < 0) {
@@ -501,12 +505,7 @@ bool CompareKeyEquality(EVP_PKEY *key1, EVP_PKEY *key2) {
 
   // Check if keys are RSA type
   if (EVP_PKEY_id(key1) != EVP_PKEY_RSA) {
-    std::cout << "AWS-LC key is not an RSA key" << std::endl;
-    return false;
-  }
-
-  if (EVP_PKEY_id(key2) != EVP_PKEY_RSA) {
-    std::cout << "OpenSSL key is not an RSA key" << std::endl;
+    std::cout << "Keys are not RSA keys" << std::endl;
     return false;
   }
 
@@ -549,12 +548,7 @@ bool CompareRandomGeneratedKeys(EVP_PKEY *key1, EVP_PKEY *key2,
 
   // Check if keys are RSA type
   if (EVP_PKEY_id(key1) != EVP_PKEY_RSA) {
-    std::cout << "AWS-LC key is not an RSA key" << std::endl;
-    return false;
-  }
-
-  if (EVP_PKEY_id(key2) != EVP_PKEY_RSA) {
-    std::cout << "OpenSSL key is not an RSA key" << std::endl;
+    std::cout << "Keys are not RSA keys" << std::endl;
     return false;
   }
 


tool-openssl/req.cc
```diff
@@ -56,22 +74,37 @@ static const argument_t kArguments[] = {
      "be formatted as /type0=value0/type1=value1/type2=.... "
      "Keyword characters may be escaped by \\ (backslash), and "
      "whitespace is retained."},
+    // AWS-LC does not support config files by design, but some of our
+    // dependencies
+    // still use this cli command with -config. Therefore, we decided to
+    // implement -config but only parse a MINIMAL set of fields (e.g.,
+    // default_md, distiguished_name, etc.). This set will be updated and
+    // re-evaluated on a need-based basis.
```
justsmth: NP: wording "... as-needed basis."

tool-openssl/req.cc
```diff
@@ -493,50 +722,161 @@ bool reqTool(const args_list_t &args) {
...
+    }
+  }
+
+  bool no_prompt = false;
+  const char *no_prompt_str = NULL;
+  if (req_conf.get()) {
+    no_prompt_str = NCONF_get_string(req_conf.get(), REQ_SECTION, PROMPT);
+  }
+
+  if (no_prompt_str != NULL && strcmp(no_prompt_str, "no") == 0) {
```
justsmth: Does this match need to be case-sensitive?  Can it use `isStringUpperCaseEqual`?

tool-openssl/req.cc
```diff
@@ -493,50 +722,161 @@ bool reqTool(const args_list_t &args) {
...
+  bool encrypt_key = true;
+  const char *encrypt_key_str = NULL;
+  if (req_conf.get()) {
+    encrypt_key_str =
+        NCONF_get_string(req_conf.get(), REQ_SECTION, ENCRYPT_KEY);
+  }
+
+  if (encrypt_key_str != NULL && strcmp(encrypt_key_str, "no") == 0) {
+    encrypt_key = true;
+  }
```
justsmth: So "no" means `encrypt_key = true`?  This seems backwards.

tool-openssl/req.cc
justsmth, line 117: `errno` should be cleared prior to calling `strtol` so that we know whether it was that call that set it.

tool-openssl/req.cc
```diff
@@ -493,50 +722,161 @@ bool reqTool(const args_list_t &args) {
...
+    const char *bits_str = NULL;
+    if (req_conf.get()) {
+      bits_str = NCONF_get_string(req_conf.get(), REQ_SECTION, BITS);
+    }
+    long default_keylen = DEFAULT_KEY_LENGTH;
+
+    if (bits_str) {
+      char *endptr = nullptr;
+      long val = strtol(bits_str, &endptr, 10);
+      if (*endptr == '\0' && endptr != bits_str) {
```
justsmth: Also clear `errno` then check its value after calling `strtol`.

tool-openssl/req.cc
```diff
@@ -493,50 +722,161 @@ bool reqTool(const args_list_t &args) {
...
+    encrypt_key = true;
+  }
+
   // Generate and write private key
   const EVP_CIPHER *cipher = NULL;
-  if (!nodes) {
+  if (!nodes && encrypt_key) {
     cipher = EVP_des_ede3_cbc();
   }
+  (void)cipher;  // Mark as potentially unused
```
justsmth: * The key was generated on line 798 above? Why comment here?
* Where is `cipher` used? Why are we declaring it? Should this be passed to `PEM_write_bio_PrivateKey`?

tool-openssl/req.cc
```diff
@@ -493,50 +722,161 @@ bool reqTool(const args_list_t &args) {
...
+    if (req_conf.get()) {
+      bits_str = NCONF_get_string(req_conf.get(), REQ_SECTION, BITS);
+    }
+    long default_keylen = DEFAULT_KEY_LENGTH;
+
+    if (bits_str) {
+      char *endptr = nullptr;
+      long val = strtol(bits_str, &endptr, 10);
+      if (*endptr == '\0' && endptr != bits_str) {
+        default_keylen = val;
```
justsmth, line 875: Are there no other constraints on what `val` can be? Too small or too large?

tool-openssl/req.cc
```diff
@@ -358,98 +446,231 @@ static int make_certificate_request(X509_REQ *req, EVP_PKEY *pkey,
...
+  if (!passin->empty() && !pass_util::ExtractPassword(passin)) {
+    fprintf(stderr, "Error: Failed to extract password\n");
+    return false;
+  }
+
+  pkey.reset(PEM_read_PrivateKey(key_file.get(), nullptr, nullptr,
+                                 const_cast<char *>(passin->c_str())));
+
+  if (!pkey) {
+    fprintf(stderr, "Error: Failed to read private key from %s",
```
justsmth: NP: Add newline to end of output.

tool-openssl/req.cc
nhatnghiho: I guess it doesn't need to be case-sensitive. Changed to use `isStringUpperCaseEqual` instead

tool-openssl/req.cc
nhatnghiho: This is more like if `encrypt_key` is anything but lowercase `no`, it will be set to true. But yeah this is a bit hard to read. I changed the logic a bit to make this clearer

tool-openssl/req.cc
nhatnghiho: > The key was generated on line 798 above? Why comment here?

Comment is now modified/moved for better clarity

> Where is cipher used? Why are we declaring it? Should this be passed to PEM_write_bio_PrivateKey?

I marked cipher as unused in the beginning to remind myself to implement it but then forgot. It should be used appropriately now.

tool-openssl/req.cc
nhatnghiho: The constraints will be applied later in `GenerateKey` [here](https://github.com/aws/aws-lc/blob/3eaeb1327375986bcf41593d3193c8f8d567cf96/tool-openssl/req.cc#L135-L144)

tool-openssl/req.cc
```diff
@@ -56,44 +74,64 @@ static const argument_t kArguments[] = {
...
     }
   } else if (OPENSSL_strcasecmp(keyspec, "rsa") == 0) {
-    keylen = DEFAULT_KEY_LENGTH;
+    keylen = default_keylen;
   } else {
     fprintf(
         stderr,
-        "Unknown key specification: %s, using RSA key with 2048 bit length\n",
-        keyspec);
+        "Unknown key specification: %s, using RSA key with %ld bit length\n",
```
justsmth: openssl exits with an error when provided with a known-type (like "ec" or "dha") when `-pkeyopt` is not provided. 
```
❯ openssl req -new -newkey dha -nodes -keyout my_new_key -out my_csr -subj "/name=justintest"
Error allocating keygen context
40A1B2F901000000:error:0308010C:digital envelope routines:inner_evp_generic_fetch:unsupported:crypto/evp/evp_fetch.c:375:Global default library context, Algorithm (dha : 0), Properties (<null>)
```

We don't support `-pkeyopt`, but we should still fail when the user is attempting to generate a key that's not RSA. Currently we proceed successfully, ignoring this option:
```
❯ ./build/tool-openssl/openssl req -new -newkey dha -nodes -keyout my_new_key -out my_csr -subj "/name=justintest"
Unknown key specification: dha, using RSA key with 2048 bit length
Writing private key to my_new_key
```

tool-openssl/req_test.cc
```diff
@@ -15,15 +15,428 @@
...
+
+  std::string key_content = ReadFileToString(output_key_path);
+  EXPECT_TRUE(key_content.find("ENCRYPTED") != std::string::npos);
+
+  bssl::UniquePtr<EVP_PKEY> key(DecryptPrivateKey(output_key_path, "testpass"));
+  ASSERT_TRUE(key);
+}
+
+TEST_F(ReqTest, DefaultKeyoutPath) {
+  // Test that key is written to privkey.pem when -keyout is not specified
```
justsmth, line 201: In some CI setups this test might be run in parallel, which creates a race condition since the "privkey.pem" created by one run could be deleted or overwritten by another.

justsmth (approved):

tool-openssl/test_util.cc
```diff
@@ -419,3 +474,109 @@ bool CompareCertificates(X509 *cert1, X509 *cert2, X509 *ca_cert,
...
+  if (EVP_PKEY_id(key1) != EVP_PKEY_RSA) {
+    std::cout << "AWS-LC key is not an RSA key" << std::endl;
+    return false;
+  }
+
+  if (EVP_PKEY_id(key2) != EVP_PKEY_RSA) {
+    std::cout << "OpenSSL key is not an RSA key" << std::endl;
+    return false;
+  }
+
```
WillChilds-Klein, line 511: aren't these unreachable due to L497?

tool-openssl/test_util.cc
```diff
@@ -419,3 +474,109 @@ bool CompareCertificates(X509 *cert1, X509 *cert2, X509 *ca_cert,
...
+  // Check if keys are RSA type
+  if (EVP_PKEY_id(key1) != EVP_PKEY_RSA) {
+    std::cout << "AWS-LC key is not an RSA key" << std::endl;
+    return false;
+  }
+
+  if (EVP_PKEY_id(key2) != EVP_PKEY_RSA) {
+    std::cout << "OpenSSL key is not an RSA key" << std::endl;
+    return false;
+  }
```
WillChilds-Klein, line 553: also unreachable?

tool-openssl/test_util.cc
```diff
@@ -134,30 +145,70 @@ bssl::UniquePtr<X509> LoadDERCertificate(const char *path) {
...
+    std::cout << "Invalid CSRs" << std::endl;
     return false;
   }
 
   // 1. Compare subjects
   X509_NAME *name1 = X509_REQ_get_subject_name(csr1);
   X509_NAME *name2 = X509_REQ_get_subject_name(csr2);
-  if (X509_NAME_cmp(name1, name2) != 0) {
+
+  if (!name1 || !name2) {
```
WillChilds-Klein, line 156: don't you still need to call `X509_NAME_cmp` on the subject names?

tool-openssl/pkeyutl.cc
```diff
@@ -298,11 +314,12 @@ bool pkeyutlTool(const args_list_t &args) {
 
     bssl::UniquePtr<BIO> output_bio;
     if (out_path.empty()) {
-      output_bio.reset(BIO_new_fp(stdout, BIO_CLOSE));
+      output_bio.reset(BIO_new_fp(stdout, BIO_NOCLOSE));
```
WillChilds-Klein, line 317: here and elsewhere, why change this flag?

tool-openssl/req.cc
```diff
@@ -56,44 +74,69 @@ static const argument_t kArguments[] = {
...
+static EVP_PKEY *GenerateKey(const char *keyspec, long default_keylen) {
   EVP_PKEY *pkey = NULL;
-  long keylen = DEFAULT_KEY_LENGTH;
+  long keylen = default_keylen;
   int pkey_type = EVP_PKEY_RSA;
 
   // Parse keyspec
   if (OPENSSL_strncasecmp(keyspec, "rsa:", 4) == 0) {
     char *endptr = NULL;
+    errno = 0;
```
WillChilds-Klein, line 115: hm, it's not generally a good practice to set errno directly in the application layer

> The <errno.h> header file defines the integer variable errno,
       which is set by system calls and some library functions in the
       event of an error to indicate what went wrong.

https://man7.org/linux/man-pages/man3/errno.3.html

why do we do so here?

tool-openssl/req.cc
```diff
@@ -358,128 +457,314 @@ static int make_certificate_request(X509_REQ *req, EVP_PKEY *pkey,
...
+    }
+  } else {
+    const char *default_config =
+        "[v3_ca]\n"
+        "subjectKeyIdentifier=hash\n"
+        "authorityKeyIdentifier=keyid:always,issuer:always\n"
+        "basicConstraints=critical,CA:true\n";
+
+    // Create a BIO for the config
+    bssl::UniquePtr<BIO> bio(BIO_new_mem_buf(default_config, -1));
```
WillChilds-Klein: do we have an upper bound on the buffer size here? negative length is not encouraged for new usages

https://github.com/aws/aws-lc/blob/981c79b69b930e17105b8ce1f20f7d81c97fe79c/include/openssl/bio.h#L461-L463

tool-openssl/test_util.cc
justsmth: Yeah, we know that key1 and key2 have the same ID, so the inside of the second if block is unreachable.

tool-openssl/test_util.cc
justsmth: `X509_NAME_cmp` compares the canonical DER encodings. The logic below this compares each entry to provide better diagnostic messages. I think we could call `X509_NAME_cmp` and return early if it says they're equal.

tool-openssl/pkeyutl.cc
justsmth: You shouldn't attempt to close `stdout` or other standard streams. On Windows, this might cause the process to abort.

tool-openssl/req.cc
justsmth: A program should generally only check the value of errno in the case that an IO operation failed, so there no need to clear previous errno values prior to an operation.  

However, when using `strtol` it's necessary in order to differentiate certain outcomes. See [here](https://man7.org/linux/man-pages/man3/strtol.3.html#CAVEATS):

> Since `strtol()` can legitimately return `0`, `LONG_MAX`, or `LONG_MIN`
> (`LLONG_MAX` or `LLONG_MIN` for `strtoll()`) on both success and
> failure, the calling program should set [errno](https://man7.org/linux/man-pages/man3/errno.3.html) to `0` before the
> call, and then determine if an error occurred by checking whether
> `errno == ERANGE` after the call.

tool-openssl/req.cc
justsmth: I updated this to provide the string length.

tool-openssl/req.cc
WillChilds-Klein: ah, POSIX std lib's API strikes again. thanks.

WillChilds-Klein (approved):

PR #2777, opened 2025-10-29
Support more "openssl rsa" options

justsmth:
### Description of changes: 
Adds support for several options to the "openssl rsa" subcommand:
* inform
* outform
* pubin
* pubout

### DIFF
--- a/tool-openssl/rsa.cc
+++ b/tool-openssl/rsa.cc
@@ -18,7 +18,7 @@
 
 static const argument_t kArguments[] = {
   { "-help", kBooleanArgument, "Display option summary" },
-  { "-in", kRequiredArgument, "RSA key input file" },
+  { "-in", kOptionalArgument, "RSA key input file" },
   { "-inform", kOptionalArgument, "Input format (PEM or DER), default PEM" },
   { "-out", kOptionalArgument, "Output file to write to" },
   { "-outform", kOptionalArgument, "Output format (PEM or DER), default PEM" },
@@ -82,6 +82,12 @@ bool rsaTool(const args_list_t &args) {
   if (help) {
     PrintUsage(kArguments);
     return true;
+  }
+
+  if (!HasArgument(parsed_args, "-in")) {
+    fprintf(stderr, "Missing value for required argument: -in\n");
+    PrintUsage(kArguments);
+    return false;
   }
 
   if (pubin) {
--- a/tool-openssl/rsa_test.cc
+++ b/tool-openssl/rsa_test.cc
@@ -107,6 +107,272 @@ TEST_F(RSAOptionUsageErrorsTest, InvalidFilePathTest) {
   ASSERT_FALSE(result);
 }
 
+// -------------------- RSA Functional Unit Tests -----------------------------
+
+class RSAFunctionalTest : public ::testing::Test {
+protected:
+  void SetUp() override {
+    ASSERT_GT(createTempFILEpath(in_path), 0u);
+    ASSERT_GT(createTempFILEpath(out_path), 0u);
+    ASSERT_GT(createTempFILEpath(pub_path), 0u);
+
+    rsa.reset(CreateRSAKey());
+    ASSERT_TRUE(rsa);
+
+    // Write private key in PEM format
+    ScopedFILE in_file(fopen(in_path, "wb"));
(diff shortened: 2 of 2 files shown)


tool-openssl/rsa.cc
```diff
@@ -3,15 +3,25 @@
...
+#include <openssl/err.h>
+#include <openssl/evp.h>
+#include <cstdio>
 #include <string>
 #include "internal.h"
+#include "openssl/x509.h"
+
+#define FORMAT_PEM 1
+#define FORMAT_DER 2
+#define FORMAT_UNKNOWN 3
```
samuel40791765, line 17: Might be better if we use an enum here instead.

tool-openssl/rsa.cc
```diff
@@ -65,21 +81,101 @@ bool rsaTool(const args_list_t &args) {
...
   if (!in_file) {
     fprintf(stderr, "Error: unable to load RSA key from '%s'\n", in_path.c_str());
     return false;
   }
 
-  bssl::UniquePtr<RSA> rsa(PEM_read_RSAPrivateKey(in_file.get(), nullptr, nullptr, nullptr));
+  bssl::UniquePtr<RSA> rsa;
+  // Load the key
+  if (pubin) {
+    if (input_format == FORMAT_DER || input_format == FORMAT_UNKNOWN) {
```
samuel40791765, line 137: Does OpenSSL also try both PEM and DER if the format isn't specified?

samuel40791765: Based on the tests, it does seem to be so.

samuel40791765 (approved):

tool-openssl/rsa.cc
justsmth: Yeah, I believe it supports either format on input unless you specify a specific one.

tool-openssl/rsa.cc
justsmth: Yeah, this is an enum that most of our commands could use. I'll defer this change to a subsequent PR.

tool-openssl/rsa.cc
```diff
@@ -65,48 +84,154 @@ bool rsaTool(const args_list_t &args) {
     return true;
   }
 
+  if (pubin) {
+    // If reading a public key, then we can only output a public key.
+    pubout = true;
```
WillChilds-Klein, line 95: looks like all the pubin/pubout branches are uncovered below. are they worth covering?

tool-openssl/rsa.cc
```diff
@@ -65,48 +84,154 @@ bool rsaTool(const args_list_t &args) {
...
+          return false;
+        }
+      } else {
+        // Output private key
+        if (output_format == FORMAT_DER) {
+          // For DER output, use PKCS#8 PrivateKeyInfo format to match OpenSSL
+          write_success = i2d_PKCS8PrivateKeyInfo_bio(out_file.get(), pkey.get());
+        } else {
+          write_success = PEM_write_bio_PrivateKey(out_file.get(), pkey.get(), nullptr, nullptr, 0, nullptr, nullptr);
+        }
```
WillChilds-Klein, line 233: regardless of IO success, can we zeroize any structs/buffers containing private key material here?

justsmth: For anything heap allocated, [OPENSSL_free does that](https://github.com/aws/aws-lc/blob/89569ed2f415ad1069f17380e26cc2e2e665c8a9/crypto/mem.c#L248) for us. Are there any buffers local to this code that would need to be cleansed?

tool-openssl/rsa.cc
justsmth: The code should be fully covered by the comparison tests, but [Codecov doesn't run them](https://app.codecov.io/gh/aws/aws-lc/pull/2777?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws) due to not having the environment variables set. I'll add a few non-comparison tests for more coverage.

tool-openssl/rsa.cc
WillChilds-Klein: read through and confirmed no local buffers that would hold private key material 👍

PR #2785, opened 2025-10-30
AES-XTS on AArch64: Set w19 earlier before cipher-stealing of 1 block + tail.

nebeid:
Fixes a bug in the AES-256-XTS AArch64 architecture implementation. For input lengths of 17-31 bytes, there was a 50% probability that an encryption operation would result in corrupted ciphertext. The resulting ciphertext would result in corrupted plaintext when performing a decrypt operation.

### Description of changes: 
w19 was set to its fixed value 0x87 after the code jumps to the case of cipher stealing of 1 block and a tail.
- This bug was introduced in [v1.62](https://github.com/aws/aws-lc/releases/tag/v1.62.0) in "Move udiv and sencond tweak calculations to when needed" by [@nebeid](https://github.com/nebeid) in [#2726](https://github.com/aws/aws-lc/pull/2726) where the setting of w19 was moved past the branch to `.Lxts_enc_tail1x`.
- Also the 3 test vectors that tested 1 block + tail were not lucky enough to exercise this issue which happens when the encrypted tweak, when shifted left by 1, results in a bit of 1, which, in turn, results in XORing the value 0x87 with the tweak.

### Testing:
The failing tests are now passing.

### DIFF
--- a/crypto/fipsmodule/modes/xts_test.cc
+++ b/crypto/fipsmodule/modes/xts_test.cc
@@ -302,7 +302,8 @@ static const XTSTestCase kXTSTestCases[] = {
         "6f229c1b60833e2a50a041b360d991814c6ec7f3199d8b2482f5b19b64c32013"
         "a679f1361a011bf37b2e1565"
     },
-
+    // Same key and tweak as ACCP test vectore above for
+    // len = 40 bytes = 2 blocks + 8 bytes
     {
         "000102030405060708090A0B0C0D0E0F101112131415161718191A1B1C1D1E1F"
         "202122232425262728292A2B2C2D2E2F303132333435363738393A3B3C3D3E3F",


WillChilds-Klein (approved):

skmcgrail (approved):

justsmth (approved):

third_party/s2n-bignum/s2n-bignum-to-be-imported/arm/aes/aes-xts-enc.S
```diff
@@ -112,7 +113,6 @@ S2N_BN_SYMBOL(aes_hw_xts_encrypt):
         // the five ivs stored into, v6.16b,v8.16b,v9.16b,v10.16b,v11.16b
         fmov  x9, d6
```
manastasova: Why are these `fmov`'s only needed before the 2 byte tail?

Issue #2776, opened 2025-10-29
Massive performance loss due to locking issue around use of ssl_update_counter() / ssl_read_counter()


wtarreau:
### Problem:

When redoing some tests on haproxy+awslc using TLS tickets, it appeared to my surprise that awslc was even slower than OpenSSL (1/4-1/5 of the perf), without eating CPU, as shown in the table below comparing the performance of the different libs in resumptions per second (and percent relative to 1.1.1):
|lib|TLS 1.2 resumptions/s|TLS 1.3 resumptions/s|
|---|-------------:|----------:|
|openssl-1.1.1|208k (100%)|144k (100%)|
|openssl-3.5.4|172k (83%)|58k (40%)|
|openssl-3.6.0|178k (86%)|~11-63k (44%)|
|**aws-lc-1.62.0**|49k (**24%**)|30k (**21%**)|

The machine (a 64-core EPYC) was 87% idle, suggeting some waiting locks. I rebuilt haproxy with  USE_PTHREAD_EMULATION=1 to replace the slow pthread rwlocks, and aws-lc's performance jumped to 91k/s at 100% CPU, clearly indicating a misuse of a pthread rwlock, which is also shown in perf top in this case:
```
Overhead  Shared Object         Symbol                                          
  43.34%  haproxy                   [.] pthread_rwlock_wrlock
   8.61%  [kernel]                  [k] native_queued_spin_lock_slowpath
   2.40%  haproxy                   [.] pthread_rwlock_rdlock
   1.94%  haproxy                   [.] sha512_block_data_order_nohw
```

By setting a break point in gdb on the lock, I easily tracked it down to `ssl_update_counter()` called from `ssl_process_ticket()` and its sibling `ssl_read_counter()`, which only take the lock to read or increment an integer (here the session hit counter)! (i.e. not even a case of consistency with other use cases since it's the only place where this counter is manipulated).

Worse, looking closer, I noticed that the `ssl_read_counter()` function is wrong: the counter is read **in the caller** out of any lock, then passed to the function, which takes the lock and returns it:
```
static int ssl_read_counter(const SSL_CTX *ctx, int counter) {
  MutexReadLock lock(const_cast<CRYPTO_MUTEX *>(&ctx->lock));
  return counter;
}
...
int SSL_CTX_sess_hits(const SSL_CTX *ctx) {
  return ssl_read_counter(ctx, ctx->stats.sess_hit);
}

```
[...]

PR #2786, opened 2025-10-30
Use C++11 atomics to update session stats

justsmth:
### Issues:
Addresses: #2776 
* This PR was tested/verified by [Willy Tarreau (w@1wt.eu)](https://github.com/aws/aws-lc/pull/2786#issuecomment-3480923204)  who reported this issue.

### Description of changes: 
- `ssl_update_counter` uses atomic [`fetch_add`](https://en.cppreference.com/w/cpp/atomic/atomic/fetch_add.html) to increment when atomic operations are available.
- `ssl_read_counter` uses atomic [`load`](https://en.cppreference.com/w/cpp/atomic/atomic/load.html) for lock-free reads when available.
- Falls back to existing mutex-based approach when atomic operations are unavailable or not guaranteed lock-free

### Call-outs:
The previous `ssl_read_counter` implementation was not taking a reference to `counter`, so was acquiring the lock for no reason.

### DIFF
--- a/ssl/ssl_lib.cc
+++ b/ssl/ssl_lib.cc
@@ -485,31 +485,17 @@ bool SSL_get_traffic_secrets(const SSL *ssl,
 }
 
 void ssl_update_counter(SSL_CTX *ctx, SSL_STATS_COUNTER_TYPE &counter, bool lock) {
-  if (lock) {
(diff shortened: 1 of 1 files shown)


wtarreau:
Thanks!

PR #2787, opened 2025-10-30
Update max polyz value

jakemas:
### Issues:
Resolves #V1982529084

### Description of changes: 
Fix a parameter issue that was not set as the max: ml-dsa-44 uses `576`, ml-dsa-65 uses `640`, ml-dsa-87 uses `640`, so max should be `640`

### Testing:
N/A


manastasova (approved):
Looks good to me!

justsmth (approved):

PR #2788, opened 2025-10-30
Make poly_chknorm constant flow

jakemas:
### Issues:
Resolves #V1982530715 and #V1982532566

### Description of changes: 
The reference implementation implements `poly_chknorm` in variables time. It argues that while the input coefficients itself are secret in some call sites, it is okay to leak which coefficient lead to rejection. It, hence, does absolute value computation in constant-time and then checks the bound using a conditional.

This approach appears safe, but somewhat unclean as it is still operating on secret data. When performing constant-time testing it also requires a number of declassifications.

This commit takes a more conservative approach and changes `poly_chknorm` to a constant-time implementation in the hope that the performance penalty is acceptable.

A minor change is that the API of `poly_chknorm` is changed to returning `0xFFFFFFFF` in the case of failure to be able to re-use existing constant-time primitives.

### Call-outs:
PR source adapted from: Upstream PR in https://github.com/pq-code-package/mldsa-native/pull/392.

### DIFF
--- a/.github/docker_images/aws-lc/ubuntu/Dockerfile.2404
+++ b/.github/docker_images/aws-lc/ubuntu/Dockerfile.2404
@@ -21,7 +21,6 @@ apt-get -y --no-install-recommends install \
     ca-certificates \
     cmake \
     curl \
-    gcc-12 \
     git \
     libffi-dev \
     libgmp-dev \
--- a/.github/workflows/actions-ci.yml
+++ b/.github/workflows/actions-ci.yml
@@ -41,7 +41,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - "macos-13"
+          - "macos-15-intel"
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v4
@@ -59,7 +59,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - "macos-13"
+          - "macos-15-intel"
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v4
--- a/.github/workflows/analytics.yml
+++ b/.github/workflows/analytics.yml
@@ -1,0 +1,35 @@
+name: aws-lc-analytics
+on:
+  push:
+    branches: [main]
+    tags-ignore: ['**']
+permissions:
+  contents: read
+env:
+  GOPROXY: https://proxy.golang.org,direct
+jobs:
+  analytics:
+    name: analytics-${{ matrix.arch }}
+    runs-on:
+      - codebuild-aws-lc-ci-github-actions-${{ github.run_id }}-${{ github.run_attempt }}
+        image:${{ matrix.arch == 'x86_64' && 'linux-5.0' || matrix.arch == 'aarch64' && 'arm-3.0' }}
+        instance-size:large
+    strategy:
+      matrix:
+        arch: [x86_64, aarch64]
+    env:
+      ANALYTICS_BRANCH: ${{ github.ref }}
+    steps:
+      - uses: actions/checkout@v5
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v2
+      - uses: ./.github/actions/codebuild-docker-run
+        name: Run Container
+        with:
+          image: ${{ steps.login-ecr.outputs.registry }}/aws-lc/amazonlinux:2
+          env: |
+            ANALYTICS_BRANCH
+          run: |
+            source /opt/compiler-env/setup-gcc-7.sh
+            ./tests/ci/run_analytics.sh
--- a/.github/workflows/integration_omnibus.yml
+++ b/.github/workflows/integration_omnibus.yml
@@ -462,5 +462,5 @@ jobs:
             FIPS
           ipv6: true
           run: |
-            source /opt/compiler-env/setup-gcc-12.sh
+            source /opt/compiler-env/setup-gcc-13.sh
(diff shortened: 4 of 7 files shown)


crypto/fipsmodule/ml_dsa/ml_dsa_ref/polyvec.c
```diff
@@ -223,16 +223,20 @@ void ml_dsa_polyvecl_pointwise_acc_montgomery(ml_dsa_params *params,
...
 **************************************************/
-int ml_dsa_polyvecl_chknorm(ml_dsa_params *params, const polyvecl *v, int32_t bound)  {
+uint32_t ml_dsa_polyvecl_chknorm(ml_dsa_params *params, const polyvecl *v, int32_t bound)  {
   unsigned int i;
+  uint32_t r = 0;
+  
+    /* Reference: Leaks which polynomial violates the bound via a conditional.
+     * We are more conservative to reduce the number of declassifications in
+     * constant-time testing.
+     */
```
dkostic:
```suggestion
```

crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c
```diff
@@ -258,15 +260,17 @@ int ml_dsa_sign_internal(ml_dsa_params *params,
   ml_dsa_polyveck_invntt_tomont(params, &h);
   ml_dsa_polyveck_sub(params, &w0, &w0, &h);
   ml_dsa_polyveck_reduce(params, &w0);
-  if(ml_dsa_polyveck_chknorm(params, &w0, params->gamma2 - params->beta)) {
+  w0_invalid = ml_dsa_polyveck_chknorm(params, &w0, params->gamma2 - params->beta);
+  if(w0_invalid) {
```
dkostic, line 269: this and the previous check (z_invalid) should not be done consecutively iiuc, because that would leak which check failed.

crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c
```diff
@@ -248,25 +249,28 @@ int ml_dsa_sign_internal(ml_dsa_params *params,
...
   ml_dsa_polyveck_pointwise_poly_montgomery(params, &h, &cp, &s2);
   ml_dsa_polyveck_invntt_tomont(params, &h);
   ml_dsa_polyveck_sub(params, &w0, &w0, &h);
   ml_dsa_polyveck_reduce(params, &w0);
-  if(ml_dsa_polyveck_chknorm(params, &w0, params->gamma2 - params->beta)) {
+  w0_invalid = ml_dsa_polyveck_chknorm(params, &w0, params->gamma2 - params->beta);
+
+  /* FIPS 204: Algorithm 7 line 23 - Reject if either check fails (constant-time to avoid leaking
+     which check failed) */
+  if(z_invalid | w0_invalid) {
```
dkostic: I think the compiler can turn this into non-CT code, please use an appropriate `constant_time_*` function to get the branch condition.

crypto/fipsmodule/ml_dsa/ml_dsa_ref/poly.c
```diff
@@ -228,29 +228,34 @@ void ml_dsa_poly_use_hint(ml_dsa_params *params,
...
-     the probability for each coefficient is independent of secret
-     data but we must not leak the sign of the centralized representative. */
+  /* Constant-time implementation to avoid timing side-channel attacks.
+     We accumulate violations using bitwise OR instead of early exit.
+     
+     Note: This uses constant_time_ge_w(t, B) instead of the mldsa-native
+     pattern constant_time_msb_w(B - 1 - t) because AWS-LC's constant-time
+     utilities handle the >= comparison correctly, while the subtraction
+     approach can cause unsigned underflow issues when t < B. Both are
+     mathematically equivalent: (t >= B) == (B - 1 - t < 0). */
```
hanno-becker: It is stated explicitly in Section 5.5 of https://pq-crystals.org/dilithium/data/dilithium-specification-round3-20210208.pdf that it is safe to leak which coefficient violates the bound. One could still decide to harden the code further as defense-in-depth, but the current comment reads as if the code is vulnerable otherwise, which it's not (from all we know).

crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c
hanno-becker: Similar comment as above. We may want to harden this as defense-in-depth, but separate checks are OK from all we know -- and what is explicitly stated in 5.5 of https://pq-crystals.org/dilithium/data/dilithium-specification-round3-20210208.pdf -- so the comment should be changed.

crypto/fipsmodule/ml_dsa/ml_dsa_ref/polyvec.c
jakemas: removed!

crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c
jakemas: Removed this check, added documentation and references to https://pq-crystals.org/dilithium/data/dilithium-specification-round3-20210208.pdf section 5.5

crypto/fipsmodule/ml_dsa/ml_dsa_ref/poly.c
jakemas: Ok updated documentation accordingly -- thanks for the reference!

torben-hansen (approved):

PR #2791, opened 2025-10-31
Additional options for "openssl c_client"

justsmth:
### Description of changes: 
Add support for `-tls1_1` and `-cipher` for "openssl s_client"

### Call-outs:
We previously were not build/running the "s_client" tests. 🤦‍♂️

### Testing:
Add tests for new options.

### DIFF
--- a/tool/client.cc
+++ b/tool/client.cc
@@ -478,7 +478,7 @@ static int verify_cb(int ok, X509_STORE_CTX *ctx)
   X509 *err_cert = X509_STORE_CTX_get_current_cert(ctx);
   int err = X509_STORE_CTX_get_error(ctx);
   int depth = X509_STORE_CTX_get_error_depth(ctx);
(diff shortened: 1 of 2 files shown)


tool/client.cc
```diff
@@ -474,7 +478,7 @@ static int verify_cb(int ok, X509_STORE_CTX *ctx)
   X509 *err_cert = X509_STORE_CTX_get_current_cert(ctx);
   int err = X509_STORE_CTX_get_error(ctx);
   int depth = X509_STORE_CTX_get_error_depth(ctx);
-  bssl::UniquePtr<BIO> bio_err(BIO_new_fp(stderr, BIO_CLOSE));
+  bssl::UniquePtr<BIO> bio_err(BIO_new_fp(stderr, BIO_NOCLOSE));
```
WillChilds-Klein, line 481: do we now need to explicitly close this for the various early-return branches below?

justsmth: No. On Windows, the process gets a abrupt termination if it attempts to close one of the standard IO streams.

PR #2795, opened 2025-11-03
AES-XTS Enc Dec test on rand incremental length inputs

manastasova:
Add incremental length random input message tests for AES-XTS encryption/decryption

## Issues:
N/A

## Description of changes:
AWS-LC's current AES-XTS tests use fixed-size test vectors but do not systematically verify that encryption and decryption work correctly across different message lengths.

This PR adds a new test that generates random inputs of incremental lengths and verify that `encrypt(decrypt(input)) == input` for each length. This ensures that AES-XTS correctly handles messages of varying sizes, particularly:
- Messages at and above the minimum length (16 bytes)
- Messages that require ciphertext stealing (non-block-aligned sizes)
- Messages of various sizes up to the practical maximum

The tests also add explicit length equality checks between plaintext and ciphertext to catch potential length-related bugs early.

## Call-outs:
The test provides better confidence in correctness across different input sizes.

## Testing:
`./crypto/crypto_test --gtest_filter=XTSTest.EncryptDecryptRand.`

### DIFF
--- a/crypto/fipsmodule/modes/xts_test.cc
+++ b/crypto/fipsmodule/modes/xts_test.cc
@@ -1268,11 +1268,13 @@ TEST(XTSTest, EncryptDecryptRand) {
     int len = 0;
     uint8_t *in_p = nullptr, *out_p = nullptr;
   #if defined(OPENSSL_LINUX)
+    std::unique_ptr<uint8_t[]> in, out;
+
     for (bool beg: {false, true}) {
       if (pagesize < (int)plaintext.size() && !beg) {
         // For small page sizes skip page bound edge cases
-        std::unique_ptr<uint8_t[]> in(new uint8_t[plaintext.size()]);
-        std::unique_ptr<uint8_t[]> out(new uint8_t[plaintext.size()]);
+        in.reset(new uint8_t[plaintext.size()]);
+        out.reset(new uint8_t[plaintext.size()]);
         in_p = in.get();
         out_p = out.get();
       } else if (pagesize < (int)plaintext.size() && beg) {


crypto/fipsmodule/modes/xts_test.cc
```diff
@@ -1240,6 +1244,90 @@ TEST(XTSTest, TestVectors) {
 #endif
 }
 
+TEST(XTSTest, EncryptDecyptRand) {
```
torben-hansen: Instead of being specific to XTS. Can expand to cover all the cipher implementations similar to tests in `cipher_test.cc` https://github.com/aws/aws-lc/blob/main/crypto/cipher_extra/cipher_test.cc#L543-L589

They use known input-output test files though. So, would need to extend that to randomised testing.

crypto/fipsmodule/modes/xts_test.cc
```diff
@@ -1240,6 +1244,90 @@ TEST(XTSTest, TestVectors) {
...
+        if (in_place) {
+          out_p = in_p;
+        }
+
+        bssl::ScopedEVP_CIPHER_CTX ctx;
+        ASSERT_TRUE(EVP_EncryptInit_ex(ctx.get(), cipher, nullptr, key.data(),
+                                      iv.data()));
+        ASSERT_TRUE(
+            EVP_EncryptUpdate(ctx.get(), out_p, &len, in_p, plaintext.size()));
+        OPENSSL_memcpy(ciphertext.data(), out_p, plaintext.size());
```
nebeid: Is this line, and more so `ciphertext`, used?

crypto/fipsmodule/modes/xts_test.cc
nebeid:
```suggestion
TEST(XTSTest, EncryptDecryptRand) {
```

crypto/fipsmodule/modes/xts_test.cc
manastasova: Indeed, not used. Thanks, Nevine. Updated.

crypto/fipsmodule/modes/xts_test.cc
manastasova: Thanks!

crypto/fipsmodule/modes/xts_test.cc
```diff
@@ -1240,6 +1244,95 @@ TEST(XTSTest, TestVectors) {
...
+#if defined(OPENSSL_LINUX)
+  int pagesize = sysconf(_SC_PAGE_SIZE);
+  ASSERT_GE(pagesize, 0);
+  uint8_t *in_buffer_beg = get_buffer_beg(pagesize);
+  uint8_t *out_buffer_beg = get_buffer_beg(pagesize);
+  uint8_t *in_buffer_end = in_buffer_beg + pagesize;
+  uint8_t *out_buffer_end = out_buffer_beg + pagesize;
+#endif
+
+const EVP_CIPHER *cipher = EVP_aes_256_xts();
```
torben-hansen: indentation

crypto/fipsmodule/modes/xts_test.cc
```diff
@@ -1240,6 +1244,95 @@ TEST(XTSTest, TestVectors) {
...
+      if (!beg) {
+        in_p = in_buffer_end - plaintext.size();
+        out_p = out_buffer_end - plaintext.size();
+      } else {
+        in_p = in_buffer_end - pagesize;
+        out_p = out_buffer_end - pagesize;
+      }
+      OPENSSL_memset(in_p, 0x00, plaintext.size());
+      OPENSSL_memset(out_p, 0x00, plaintext.size());
+    #else
```
torben-hansen: indentation

crypto/fipsmodule/modes/xts_test.cc
```diff
@@ -1240,6 +1244,95 @@ TEST(XTSTest, TestVectors) {
...
+        }
+
+        ctx.Reset();
+        len = 0;
+        ASSERT_TRUE(EVP_DecryptInit_ex(ctx.get(), cipher, nullptr, key.data(),
+                                      iv.data()));
+        ASSERT_TRUE(
+            EVP_DecryptUpdate(ctx.get(), in_p, &len, out_p, plaintext.size()));
+
+        decrypted_len = len;
```
torben-hansen: nitpicking, but variables `decrypted_len` and `encrypted_len` seems redundant

crypto/fipsmodule/modes/xts_test.cc
```diff
@@ -1240,6 +1244,95 @@ TEST(XTSTest, TestVectors) {
...
+        ASSERT_EQ(decrypted_len, plaintext.size());
+        EXPECT_EQ(Bytes(plaintext), Bytes(in_p, static_cast<size_t>(len)));
```
torben-hansen, line 1336: might as well also be an assert.

crypto/fipsmodule/modes/xts_test.cc
```diff
@@ -1240,6 +1244,95 @@ TEST(XTSTest, TestVectors) {
...
+  uint8_t *in_buffer_end = in_buffer_beg + pagesize;
+  uint8_t *out_buffer_end = out_buffer_beg + pagesize;
+#endif
+
+const EVP_CIPHER *cipher = EVP_aes_256_xts();
+
+  // Test AESXTS Encrypt and Decrypt with random messages of incremental lenghts
+  for (size_t msg_len = 16; msg_len < AESXTS_RAND_MSG_MAX_LEN ; msg_len += 1) {
+
+    std::vector<uint8_t> key(EVP_CIPHER_key_length(cipher)), iv(EVP_CIPHER_iv_length(cipher)),
```
torben-hansen: nitpicking, but don't need to allocate new memory per loop iteration. (same for `in` and `out` below). In the grand scheme of things, won't save much testing time though.

crypto/fipsmodule/modes/xts_test.cc
```diff
@@ -23,6 +25,8 @@
 #include "internal.h"
 #include "../../test/test_util.h"
 
+#define AESXTS_RAND_MSG_MAX_LEN 4096
```
torben-hansen:
```suggestion
static constexpr size_t AESXTS_RAND_MSG_MAX_LEN = 4096
```
for type-safety and more c++ idiomatic

crypto/fipsmodule/modes/xts_test.cc
```diff
@@ -1240,6 +1244,95 @@ TEST(XTSTest, TestVectors) {
...
+                                      iv.data()));
+        ASSERT_TRUE(
+            EVP_EncryptUpdate(ctx.get(), out_p, &len, in_p, plaintext.size()));
+
+        encrypted_len = len;
+        ASSERT_EQ(encrypted_len, plaintext.size());
+
+        // Test decryption.
+
+        if (!in_place) {
```
torben-hansen: if-condition not needed.

crypto/fipsmodule/modes/xts_test.cc
```diff
@@ -1240,6 +1244,95 @@ TEST(XTSTest, TestVectors) {
...
+      in_p = in.get();
+      out_p = out.get();
+    #endif
+      for (bool in_place : {false, true}) {
+        SCOPED_TRACE(in_place);
+
+        // Test encryption.
+
+        OPENSSL_memcpy(in_p, plaintext.data(), plaintext.size());
+        if (in_place) {
```
torben-hansen, line 1311: Just a note: The not in-place test case assumes that the loop initializer list being processed in-order from left-to-right. Which it is! If not, we would always test the in-place case.

crypto/fipsmodule/modes/xts_test.cc
```diff
@@ -1240,6 +1244,95 @@ TEST(XTSTest, TestVectors) {
...
+    RAND_bytes(iv.data(), EVP_CIPHER_iv_length(cipher));
+    RAND_bytes(plaintext.data(), msg_len);
+
+    SCOPED_TRACE(plaintext.size());
+
+    int len = 0;
+    size_t decrypted_len = 0, encrypted_len = 0;
+    uint8_t *in_p = nullptr, *out_p = nullptr;
+  #if defined(OPENSSL_LINUX)
+    ASSERT_GE(pagesize, (int)plaintext.size());
```
torben-hansen: Hopefully, we will never meet an system that requires small pagesizes (e.g. below 4kib). I understand the test here. But maybe we can just take the minimal of pagesize and max plaintext size and let that be the upper bound we test to? Slightly more robust.

crypto/fipsmodule/modes/xts_test.cc
```diff
@@ -15,6 +15,8 @@
 #include <openssl/cipher.h>
 #include <openssl/aes.h>
```
torben-hansen: No newline

crypto/fipsmodule/modes/xts_test.cc
manastasova: Thank you

crypto/fipsmodule/modes/xts_test.cc
manastasova: I agree

crypto/fipsmodule/modes/xts_test.cc
manastasova: `in`, `out` and `plaintext` are of different lengths based on the current loop iteration. I moved `key` and `iv` outside of the loop.

crypto/fipsmodule/modes/xts_test.cc
manastasova: I see your point (even though it is, indeed, very unlikely to have linux page < 4kib). 
I still believe it is useful to explicitly test the edge cases related to page size. I keep it as is with an additional check if OPENSSL_LINUX and pagesize < plaintext.size();

crypto/fipsmodule/modes/xts_test.cc
manastasova: I agree! Fixed! Thanks!

crypto/fipsmodule/modes/xts_test.cc
manastasova: I prefer to keep it as `EXPECT_EQ` so that all the failing tests can be listed and tracked (e.g., plaintext values between 17 and 31 bytes will all fail, but will pass if length is above 31 bytes - adds clues for debugging).

crypto/fipsmodule/modes/xts_test.cc
manastasova: thanks

crypto/fipsmodule/modes/xts_test.cc
manastasova: fixed

crypto/fipsmodule/modes/xts_test.cc
manastasova: thanks

crypto/fipsmodule/modes/xts_test.cc
torben-hansen: Sure - Just noted that all other tests are asserts.

crypto/fipsmodule/modes/xts_test.cc
torben-hansen: Right, could be allocated with a length upper bound (:
But can leave as-is.

crypto/fipsmodule/modes/xts_test.cc
```diff
@@ -1240,6 +1243,104 @@ TEST(XTSTest, TestVectors) {
...
+      for (bool in_place : {false, true}) {
+        SCOPED_TRACE(in_place);
+
+        // Generate random key and iv for each encryption test
+        RAND_bytes(key.data(), EVP_CIPHER_key_length(cipher));
+        RAND_bytes(iv.data(), EVP_CIPHER_iv_length(cipher));
+
+        // Test encryption.
+
+        OPENSSL_memcpy(in_p, plaintext.data(), plaintext.size());
```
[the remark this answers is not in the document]
manastasova: Not sure why this warning appears. `in_p` is used as the input message for the encryption function later on.

nebeid (approved):

PR #2801, opened 2025-11-04
Use right compiler with ruby CI

samuel40791765:
### Issues:
Resolves `P315730123`

### DIFF
--- a/.github/docker_images/aws-lc/ubuntu/Dockerfile.2404
+++ b/.github/docker_images/aws-lc/ubuntu/Dockerfile.2404
@@ -21,7 +21,6 @@ apt-get -y --no-install-recommends install \
     ca-certificates \
     cmake \
     curl \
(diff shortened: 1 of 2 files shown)


.github/docker_images/aws-lc/ubuntu/Dockerfile.2404
```diff
@@ -21,7 +21,6 @@ apt-get -y --no-install-recommends install \
     ca-certificates \
     cmake \
     curl \
-    gcc-12 \
```
nebeid, line 24: The docker image doesn't need a compiler specified?

samuel40791765: It already has one in gcc-13 as the default compiler. The gcc-12 installation was redundant.

nebeid (approved):

PR #2806, opened 2025-11-07
Ensure HMAC_Init_ex reinitializes data properly

samuel40791765:
### Description of changes: 
I discovered this edge case when working with an older Ruby version. This discrepancy only exists in older versions since Ruby openssl migrated from using the `HMAC_CTX` APIs to use the `EVP` layer in 3.1: https://github.com/ruby/ruby/commit/b91f62f3840582bb3e4fbece15654e1d852c829b.

[`test_reset_keep_key`](https://github.com/ruby/ruby/blame/cf4a034d59913fb71a7dd1b052164984be4a3d14/test/openssl/test_hmac.rb#L37-L43) was failing since we were MACing the data twice. It turns out the call to `h1.reset` wasn't working properly and this was due to our implementation of `HMAC_Init_ex` not reinitializing the data input when only `HMAC_Update` had been called. According to the original function contract, `HMAC_Init_ex` should reinitialize the inputted data, but the computed key should still be preserved.

This is a minor edge case due to how older versions of Ruby were consuming `HMAC_CTX`. [Their call](https://github.com/ruby/ruby/blob/ruby_2_7/ext/openssl/ossl_hmac.c#L167-L174) to `HMAC_Final` was called upon a copy of the original `HMAC_CTX`. The original `HMAC_CTX` was always within a `HMAC_Update` state and AWS-LC was not properly reinitializing these cases.

### Call-outs:
N/A

### Testing:
New tests

### DIFF
--- a/crypto/fipsmodule/hmac/hmac.c
+++ b/crypto/fipsmodule/hmac/hmac.c
@@ -378,16 +378,20 @@ int HMAC_Init_ex(HMAC_CTX *ctx, const void *key, size_t key_len,
     ctx->state = HMAC_STATE_INIT_NO_DATA;  // Mark that init has been called
   }
 
-  // TODO(davidben,eroman): Passing the previous |md| with a NULL |key| is
-  // ambiguous between using the empty key and reusing the previous key. There
-  // exist callers which intend the latter, but the former is an awkward edge
-  // case. Fix to API to avoid this.
-  if (key == NULL && (md == NULL || md == ctx->md)) {
-    // Reinitialize |md_ctx| from |i_ctx| to start fresh with the same key. This
-    // is the same behavior as Openssl.
-    OPENSSL_memcpy(&ctx->md_ctx, &ctx->i_ctx, sizeof(ctx->i_ctx));
-    // If nothing is changing then we can return without doing any further work.
-    return 1;
+  if (hmac_ctx_is_initialized(ctx)) {
+    // TODO(davidben,eroman): Passing the previous |md| with a NULL |key| is
+    // ambiguous between using the empty key and reusing the previous key. There
+    // exist callers which intend the latter, but the former is an awkward edge
+    // case. Fix to API to avoid this.
+    if (key == NULL && (md == NULL || md == ctx->md)) {
+      if(HMAC_STATE_IN_PROGRESS == ctx->state) {
+        // Reinitialize |md_ctx| from |i_ctx| to start fresh with the same key. This
+        // is the same behavior as Openssl.
+        OPENSSL_memcpy(&ctx->md_ctx, &ctx->i_ctx, sizeof(ctx->i_ctx));
+      }
+      // If nothing is changing then we can return without doing any further work.
+      return 1;
+    }
   }
 
   // At this point we *know* we need to change things and rekey because either the key has changed


crypto/fipsmodule/hmac/hmac.c
```diff
@@ -376,15 +376,18 @@ int HMAC_Init_ex(HMAC_CTX *ctx, const void *key, size_t key_len,
...
-      return 1;
+  // TODO(davidben,eroman): Passing the previous |md| with a NULL |key| is
+  // ambiguous between using the empty key and reusing the previous key. There
+  // exist callers which intend the latter, but the former is an awkward edge
+  // case. Fix to API to avoid this.
+  if (key == NULL && (md == NULL || md == ctx->md)) {
+    // Check if we should reinitialize with the same key
+    if (key_len == 0) {
+      // Reinitialize |md_ctx| from |i_ctx| to start fresh with the same key.
+      OPENSSL_memcpy(&ctx->md_ctx, &ctx->i_ctx, sizeof(ctx->i_ctx));
```
justsmth: Why is this conditioned on `key_len == 0` when we already know that `key == NULL`?

samuel40791765: Yeah I was trying to make sure that the same condition was being forced as the ruby test, but this doesn't really make much sense in hindsight.

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -376,15 +376,16 @@ int HMAC_Init_ex(HMAC_CTX *ctx, const void *key, size_t key_len,
...
-      return 1;
-    }
+  // TODO(davidben,eroman): Passing the previous |md| with a NULL |key| is
+  // ambiguous between using the empty key and reusing the previous key. There
+  // exist callers which intend the latter, but the former is an awkward edge
+  // case. Fix to API to avoid this.
+  if (key == NULL && (md == NULL || md == ctx->md)) {
+    // Reinitialize |md_ctx| from |i_ctx| to start fresh with the same key. This
+    // is the same behavior as Openssl.
+    OPENSSL_memcpy(&ctx->md_ctx, &ctx->i_ctx, sizeof(ctx->i_ctx));
```
WillChilds-Klein: are we sure that `ctx->i_ctx` is populated at this point?

samuel40791765: Both `i_ctx` and `md_ctx` are allocated on the stack, so there's no pointer to check. But I've added a general `NULL` check for `ctx` above.

https://github.com/aws/aws-lc/blob/981c79b69b930e17105b8ce1f20f7d81c97fe79c/include/openssl/hmac.h#L266-L273

justsmth (approved):

WillChilds-Klein (approved):

crypto/fipsmodule/hmac/hmac.c
```diff
@@ -371,20 +371,23 @@ int HMAC_Init_ex(HMAC_CTX *ctx, const void *key, size_t key_len,
...
```
nebeid: I suggest we keep the state machine check by checking instead for `hmac_ctx_is_initialized(ctx)` and later for `HMAC_STATE_IN_PROGRESS` before the copy because the copy also happens at the end of Final.
```suggestion
  if (hmac_ctx_is_initialized(ctx)) {
  // TODO(davidben,eroman): Passing the previous |md| with a NULL |key| is
  // ambiguous between using the empty key and reusing the previous key. There
  // exist callers which intend the latter, but the former is an awkward edge
  // case. Fix to API to avoid this.
  if (key == NULL && (md == NULL || md == ctx->md)) {
    // Reinitialize |md_ctx| from |i_ctx| to start fresh with the same key in case
    // Init was called after Update and not after Final.
    if (HMAC_STATE_IN_PROGRESS == ctx->state) {
    OPENSSL_memcpy(&ctx->md_ctx, &ctx->i_ctx, sizeof(ctx->i_ctx));
```

PR #2811, opened 2025-11-11
Add infrastructure for managing third-party test vectors

sgmenda:
### Issues:

N/A -- AWS-LC does not have a systematic way to track and sync cryptographic test vectors from external sources like Wycheproof.

### Description of changes: 

This PR adds `third_party/vectors/` with:
- `sync.py` - python script to sync existing vectors and add new ones
- `sources.toml` - straightforward config of upstream sources  
- `vectorslib/` - utility functions (will host upcoming conversion logic)
- `upstream/` - unmodified upstream test vectors (always checked in for transparency)
  - `upstream/wycheproof/`- initial import from Wycheproof including `LICENSE` and  `aes_gcm_test.json`

### Call-outs:

- conversion hasn't yet been implemented, `convert_sources()` is a placeholder - will be implemented in the next PR
- `--check` mode is documented but not yet implemented

### Testing:

Manually tested:
- `./sync.py --new wycheproof/testvectors_v1/aes_gcm_test.json` successfully adds new file
- `./sync.py` verifies existing files are in sync with upstream
- successfully handles error cases: invalid source names, missing files, and duplicate files


third_party/vectors/vectorslib/utils.py
```diff
@@ -0,0 +1,22 @@
+import sys
+
+def eprint(*args, **kwargs):
+    print(*args, file=sys.stderr, **kwargs)
+
+class Colors:
+    RED = "\033[31m" if sys.stdout.isatty() else ""
```
torben-hansen, line 7: Fancy. Yet probably redundant.

third_party/vectors/sync.py
```diff
@@ -0,0 +1,234 @@
...
+def main() -> int:
+    cwd = pathlib.Path.cwd()
+    if not (cwd.parts[-1] == "vectors" and cwd.parts[-2] == "third_party"):
+        print(
+            f"ERROR: must run from [REPO_ROOT]/third_party/vectors, not {cwd}",
+            file=sys.stderr,
+        )
+        return 1
+
+    from vectorslib import utils
```
torben-hansen: already import this

third_party/vectors/sync.py
```diff
@@ -0,0 +1,234 @@
...
+    cwd: pathlib.Path,
+    tmpdir: pathlib.Path,
+    sources: dict,
+    new_file: typing.Optional[str],
+    skip_fetch: bool,
+    skip_convert: bool,
+    using_custom_tmpdir: bool = False,
+):
+    if not fetch_sources(tmpdir, sources, skip_fetch, using_custom_tmpdir):
+        return False
```
torben-hansen: More pythonic way would be to raise exceptions and catch it here.

third_party/vectors/sync.py
```diff
@@ -0,0 +1,234 @@
...
+    cwd: pathlib.Path,
+    sources: dict,
+    new_file: typing.Optional[str],
+    skip_fetch: bool,
+):
+    if skip_fetch:
+        utils.info("skipping update")
+        return True
+
+    upstream_dir = cwd / "upstream"
```
torben-hansen, line 58: So, this is "existing" test vectors, right?

third_party/vectors/sync.py
```diff
@@ -0,0 +1,234 @@
...
+        local_path = source_info["local_path"]
+        
+        for upstream_file in upstream_path.rglob("*"):
+            if not upstream_file.is_file():
+                continue
+            
+            relative_path = upstream_file.relative_to(upstream_path)
+            local_file = local_path / relative_path
+            
+            if not local_file.exists():
```
torben-hansen: So, if I understand correctly: iterate over all existing files and fail closed if a vector exist locally but not from fresh cloned upstream?

Could consider a warning.

with this logic, you won't be able to add any new files unless existing files are all in sync. maybe that's what you want though.

third_party/vectors/sync.py
```diff
@@ -0,0 +1,234 @@
...
+            
+            if not filecmp.cmp(local_file, upstream_file, shallow=False):
+                shutil.copy2(local_file, upstream_file)
+                utils.info(f"updated upstream file: {source_name}/{relative_path}")
+
+    if new_file:
+        file_path = pathlib.Path(new_file)
+        source_name = file_path.parts[0]
+        
+        if source_name not in sources:
```
torben-hansen, line 70: input validation should be much earlier

third_party/vectors/sync.py
sgmenda: yes

third_party/vectors/sync.py
sgmenda: Yep, that's the intended behavior. If upstream touches files we use, I want this script to notify us so we can manually take a look. For example, they might soon split the ML-KEM test vector file into two, and that is worth a notification so we can update our tests.

I think failing closed is fine because this script will *not* be a part of building AWS-LC. In our design, this script will run once a week in CI and if it fails generate a notification.

third_party/vectors/sync.py
sgmenda: good catch, updated this function to add new file before updating existing files, so we can catch invalid inputs earlier.

third_party/vectors/sync.py
sgmenda: oops, good catch.

third_party/vectors/sync.py
sgmenda: C brain 😄, updated.

third_party/vectors/sync.py
torben-hansen: Fair enough, it's one choice. No right or wrong. A benefit on iterating through all files for the "sync" check (either by a warning of delayed error) is that we would get a list of all "missing" files immediately.

> For example, they might soon split the ML-KEM test vector file into two, and that is worth a notification so we can update our tests.

Presumably, this would only trigger a notification if the existing test vector file name also changes. But discovering new relevant test vector files is not straightforward (or clear how one would do that).

third_party/vectors/sync.py
torben-hansen: This could be even earlier actually. Because the name has to match a stanza name from toml. But it's nitpicking at this point :)

torben-hansen (dismissed):

third_party/vectors/sync.py
sgmenda: re list of missing files immediately: that's a good point, I hadn't thought of that, will update.

third_party/vectors/sync.py
```diff
@@ -0,0 +1,232 @@
...
+
+def update_sources(
+    cwd: pathlib.Path,
+    sources: dict,
+    new_file: typing.Optional[str],
+    skip_fetch: bool,
+):
+    if skip_fetch:
+        utils.info("skipping update")
+        return
```
skmcgrail: It feels like this type of check should be higher up in `sync_source`. It seems weird that you are calling a function that is explicitly meant to do a certain action (update sources) but then telling it not to? It feels like it shouldn't be the responsibility of the function to check whether it should action in this manner, but the caller of the function.

third_party/vectors/sync.py
```diff
@@ -0,0 +1,232 @@
...
+
+def convert_sources(
+    cwd: pathlib.Path,
+    tmpdir: pathlib.Path,
+    sources: dict,
+    skip_convert: bool,
+):
+    if skip_convert:
+        utils.info("skipping convert")
+        return
```
skmcgrail: Same here for, I think this should be pulled higher-up.

third_party/vectors/sync.py
```diff
@@ -0,0 +1,232 @@
...
+
+def fetch_sources(
+    tmpdir: pathlib.Path,
+    sources: dict,
+    skip_fetch: bool,
+    using_custom_tmpdir: bool = False,
+):
+    if skip_fetch:
+        utils.info("skipping fetch")
+        return
```
skmcgrail: See note further below

third_party/vectors/sync.py
```diff
@@ -0,0 +1,232 @@
...
+
+    for source_name, source_info in sources.items():
+        stmpdir = tmpdir / source_name
+
+        if stmpdir.is_dir():
+            assert using_custom_tmpdir
+            utils.warning(
+                f"using existing, potentially stale upstream clone of {source_name} at {stmpdir}"
+            )
```
skmcgrail: The naming here doesn't feel right which left me a bit confused. It's not that this is a "temporary directory" path that you are planning to clone into, instead this is an existing directory that is either a clone of upstream, or a source snapshot. I feel like maybe there is a better name for this? I think temporary is a built overloaded, perhaps offline_sources_path? Or something to that affect? It describes better how it is being used IMO.

third_party/vectors/sync.py
sgmenda: good point, moved the skipping logic one level up.

third_party/vectors/sync.py
sgmenda: You're right, it started off as a temporary directory but then morphed into this. I renamed it to `--clone-dir`.

PR #2826, opened 2025-11-14
Add EVP_CIPHER API for XAES-256-GCM with Key Commitment 

ttunglee:
1. Add XAES-256-GCM, which is extended AES-256-GCM with a derived key mode proposed by [Filippo Valsorda](https://words.filippo.io/xaes-256-gcm/) in 2023, followed by [a specification](https://github.com/C2SP/C2SP/blob/main/XAES-256-GCM.md) released in 2024. The key commitment extension follows https://eprint.iacr.org/2025/758.pdf#page=6. 
2. This implementation supports EVP_CIPHER API and uses an optimized CMAC dedicated to the specific use case of XAES-256-GCM from https://github.com/aws/aws-lc/pull/2652 to derive subkeys and key commitments. 
3. Support varying nonce sizes: 20 ≤ b ≤ 24 based on the extension: https://eprint.iacr.org/2025/758.pdf#page=24 

### **Description of Changes**
1. Implementation for API EVP_CIPHER of XAES-256-GCM with Key Commitment is appended to [e_aes.c](https://github.com/ttungle96/aws-lc/blob/xaes-256-gcm/crypto/fipsmodule/cipher/e_aes.c#L2056)
2. The tests are appended to [cipher_tests.txt](https://github.com/ttungle96/aws-lc/blob/xaes-256-gcm/crypto/cipher_extra/test/cipher_tests.txt#L535). 
3. TestCipherAPI (https://github.com/ttungle96/aws-lc/blob/xaes-256-gcm/crypto/cipher_extra/cipher_test.cc#L184) and TestCipher (https://github.com/ttungle96/aws-lc/blob/xaes-256-gcm/crypto/cipher_extra/cipher_test.cc#L473) in cipher_test.cc is added a parameter named `kc` at the end with default value to support testing cipher with key commitment without disrupting legacy code. 

### **Testing**
Reuse the test vectors in https://github.com/aws/aws-lc/pull/2750 with our own added key commitment values. 

### **Evaluation**
Performance evaluation with speedtool on c6i.2xlarge (8 vCPUs and 16 GB RAM):
KCXAES

`./crypto_test --gtest_filter='CipherTest.*'`

### DIFF
--- a/tool/speed.cc
+++ b/tool/speed.cc
@@ -2952,6 +2952,7 @@ bool Speed(const std::vector<std::string> &args) {
        !SpeedEvpCipherGeneric(EVP_aes_256_gcm(), "EVP-AES-256-GCM", kTLSADLen, selected) ||
 #if AWSLC_API_VERSION > 34
        !SpeedEvpCipherGeneric(EVP_xaes_256_gcm(), "EVP-XAES-256-GCM", kTLSADLen, selected) ||
+       !SpeedEvpCipherGeneric(EVP_xaes_256_gcm_kc(), "EVP-XAES-256-GCM-KC", kTLSADLen, selected) ||
 #endif 
        !SpeedEvpCipherGeneric(EVP_aes_128_ctr(), "EVP-AES-128-CTR", kTLSADLen, selected) ||
        !SpeedEvpCipherGeneric(EVP_aes_192_ctr(), "EVP-AES-192-CTR", kTLSADLen, selected) ||


crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -1781,7 +1801,6 @@ Extension to support nonce size less than 24 bytes:
 https://eprint.iacr.org/2025/758.pdf#page=24
 -----------------------------------------------------------------------*/
 #define XAES_256_GCM_KEY_LENGTH      (AES_BLOCK_SIZE * 2)
-#define XAES_256_GCM_KEY_COMMIT_SIZE (AES_BLOCK_SIZE * 2)
```
nebeid: Why was this removed?

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -2033,3 +2052,128 @@ DEFINE_METHOD_FUNCTION(EVP_AEAD, EVP_aead_xaes_256_gcm) {
...
+// ------------------------------------------------------------------------------
+static XAES_256_GCM_KC_CTX *xaes_256_gcm_kc_from_cipher_ctx(EVP_CIPHER_CTX *ctx) { 
+    // alignment to be consistent with aes_gcm_from_cipher_ctx()
+    char *ptr = ctx->cipher_data;
+#if defined(OPENSSL_32_BIT)
+    assert((uintptr_t)ptr % 4 == 0);
+    ptr += (uintptr_t)ptr & 4;
+#endif
+    assert((uintptr_t)ptr % 8 == 0);
+    ptr += (uintptr_t)ptr & 8;
```
nebeid, line 2068: I doubt we need these alignment requirements in addition to those of the gcm ctx. Can we try removing them here and on line 1858? Otherwise, do you know what would drive our need for them?

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -2033,3 +2052,128 @@ DEFINE_METHOD_FUNCTION(EVP_AEAD, EVP_aead_xaes_256_gcm) {
...
+    }
+
+    return 1;
+}
+
+static int xaes_256_gcm_kc_ctrl(EVP_CIPHER_CTX *ctx, int type, int arg, void *ptr) {
+
+    XAES_256_GCM_KC_CTX *xaes_ctx = xaes_256_gcm_kc_from_cipher_ctx(ctx);
+
+    switch(type) {
```
nebeid, line 2140: I think it's more prudent to check that ptr is not NULL.
I see that there are no checks in `aes_gcm_ctrl`, but it's good to verify here and we should probably change there as well wherever it's used. It's sometimes intentionally NULL as in the INIT case in https://github.com/aws/aws-lc/blob/88aec8967065cbba516e5720c001cde1d9cd3bdf/crypto/fipsmodule/cipher/cipher.c#L185

crypto/cipher_extra/test/cipher_tests.txt
```diff
@@ -532,6 +532,34 @@ Ciphertext = 986ec1832593df5443a17943
 AAD = 633273702e6f72672f584145532d3235362d47434d
 Tag = 7fd083bf3fdb41abd740a21f71eb769d
 
+# Note: KC are our own test values
```
nebeid, line 535: Is it possible to use the same set of tests as the ones without KC and the test harness just ignores it when it's plain XAES?

crypto/fipsmodule/cipher/e_aes.c
ttunglee: Yes. I'll add checking ptr for two new added ctrl commands.

crypto/cipher_extra/test/cipher_tests.txt
ttunglee: Does that mean I should remove all changes that were made to `cipher_test.cc` to support key commitment verification?

crypto/fipsmodule/cipher/e_aes.c
ttunglee: It will lead to many CI tests failed. I tried that. You can see the failed tests below. However, when I created and run with a docker, all tests still passed.

crypto/fipsmodule/cipher/e_aes.c
```diff
@@ -390,23 +390,11 @@ static EVP_AES_GCM_CTX *aes_gcm_from_cipher_ctx(EVP_CIPHER_CTX *ctx) {
     case NID_xaes_256_gcm: 
       assert(ctx->cipher->ctx_size == sizeof(XAES_256_GCM_CTX));
       ptr = ctx->cipher_data;
-#if defined(OPENSSL_32_BIT)
```
nebeid, line 384: In my previous comment, I meant keeping this alignment in `aes_gcm_from_cipher_ctx` and try and remove them from the ones below, but maybe it doesn't make sense because it's the same pointer to `ctx->cipher_data` and it would be different if you're not doing the same operations to get it.

include/openssl/cipher.h
```diff
@@ -744,7 +744,8 @@ BSSL_NAMESPACE_END
 #define CIPHER_R_ALIGNMENT_CHANGED 142
 #define CIPHER_R_SERIALIZATION_INVALID_SERDE_VERSION 143
 #define CIPHER_R_SERIALIZATION_INVALID_CIPHER_ID 144
-// We're introducing a new error code for invalid key commitment
+// We're introducing a new error code for invalid key commitment and invalid buffer
 #define CIPHER_R_KEY_COMMITMENT_INVALID 145
+#define CIPHER_R_BUFFER_INVALID 146
```
nebeid: You can use https://github.com/aws/aws-lc/blob/88aec8967065cbba516e5720c001cde1d9cd3bdf/crypto/internal.h#L1447
and no need to introduce a new error.

crypto/cipher_extra/test/cipher_tests.txt
nebeid: No, sorry, what I meant is if you were to unify the 2 sets of test vectors in this file (i.e. delete the ones above of XAES-256-GCM because they're superseded by the KC one), would they then work with both XAES and XAES-KC? It seems to me that XAES may not be tested in `CipherFileTest` because it needs its own "Cipher" string. So you can ignore the suggestion.

crypto/cipher_extra/cipher_test.cc
nebeid, line 1: Can the negative tests that were added for XAES (e.g. shorter nonces) be modified to also test XAES-KC?

crypto/fipsmodule/cipher/e_aes.c
ttunglee: Yeah. Since `aes_gcm_from_cipher_ctx() `is called in other places like `aes_gcm_cipher()`, `aes_gcm_ctrl()`, and `aes_gcm_init_key()`, it needs to be consistent everywhere.

crypto/cipher_extra/test/cipher_tests.txt
ttunglee: Yes. It needs to have "Cipher = XAES-256-GCM-KC" to specify the cipher.

crypto/cipher_extra/cipher_test.cc
ttunglee: I made changes in `cipher_test.cc` to test negative cases with invalid key/nonce, and shorter nonce for both `evp-xaes-256-gcm` and `evp-xaes-256-gcm-kc`.

include/openssl/cipher.h
ttunglee: Thank you! I replaced that directive in my update and deleted CIPHER_R_BUFFER_INVALID.

nebeid (approved):

PR #2830, opened 2025-11-14
Android Docker Image Build

skmcgrail:
### Description of changes: 
Migrate the Android image build to GitHub workflow managed docker image build. I also have added a helper action `./.github/actions/configure-aws-credentials` which will handle logging into the OIDC base-role and then will perform a chained assume role with those credentials to the target roleName. So this is the first workflow that is specifically using the OIDC docker image build role that I will also migrate the other workflows to use, and eventually all workflows that need privileged access to pull container images etc.

### DIFF
--- a/.github/docker_images/aws-lc/android/Dockerfile
+++ b/.github/docker_images/aws-lc/android/Dockerfile
@@ -1,7 +1,7 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0 OR ISC
 
-FROM ubuntu:24.04
+FROM public.ecr.aws/ubuntu/ubuntu:24.04 AS base
 
 SHELL ["/bin/bash", "-c"]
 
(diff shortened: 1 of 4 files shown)


torben-hansen (approved):

.github/docker_images/aws-lc/android/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+
+# Setup Android SDK
+RUN <<EOF
+# Set Java 17 as default
+export JAVA17_ALT=$(update-alternatives --list java | grep java-17 | head -1)
+update-alternatives --set java $JAVA17_ALT
+
+# Set Java 17 for SDK manager compatibility
+export JAVA_HOME=$(find /usr/lib/jvm -name "*java-17*" -type d | head -1)
+export PATH=$JAVA_HOME/bin:$PATH
```
sgmenda, line 44: It isn't immediately obvious to me how we selected Java 17. It might be worth linking to the source of this info. I see it here: https://developer.android.com/build/jdks

sgmenda-aws (dismissed):

justsmth (approved):

.github/docker_images/aws-lc/android/Dockerfile
skmcgrail: I actually don't have the context on this myself. Our Android build is in some dire need of some TLC. It's a bit lost due to how the diff is shown, but this was just a straight move of this file, and I converted it to use heredoc syntax. Otherwise content is the same as before with the addition of the adjustment to use the new Go install script like the other images we build. I also say the Android build needs some TLC as our Gradle tends to trend out of date, requiring us to use an older JDK, and then we still use old SDKs etc. Tends to become a problem when we go to debug the occasional failure and realize how much it has drifted in terms of Android development environment changes.

PR #2836, opened 2025-11-17
Match req CLI behavior with OpenSSL

nhatnghiho:
### Description of changes: 
Some of our current implementation doesn't match the OpenSSL default behavior, resulting in some package failures. The fixes include:
- Suppress private key write when `-keyout` is not provided and the config has no default output path
- Use default extensions when no valid extension section is found in config
- Add a fallback when no `req` section is found in config

### Testing:
Cleaned up and added extra tests

### DIFF
--- a/tool-openssl/req.cc
+++ b/tool-openssl/req.cc
@@ -531,17 +531,19 @@ static bool LoadConfig(const std::string &config_path,
 static bool AddCertExtensions(X509 *cert, CONF *req_conf,
                               std::string &ext_section) {
   // Set up X509V3 context for certificate
-  bool result = false;
+  bool result = true;
   X509V3_CTX ext_ctx;
   X509V3_set_ctx(&ext_ctx, cert, cert, NULL, NULL,
                  X509V3_CTX_REPLACE);  // self-signed
 
-  if (req_conf != NULL && !ext_section.empty()) {
+  if (req_conf != NULL) {
     X509V3_set_nconf(&ext_ctx, req_conf);
 
     // Add extensions from config to the certificate
-    result = X509V3_EXT_add_nconf(req_conf, &ext_ctx, ext_section.c_str(),
-                                  cert) != 0;
+    if (!ext_section.empty()) {
+      result = X509V3_EXT_add_nconf(req_conf, &ext_ctx, ext_section.c_str(),
+                                    cert) != 0;
+    }
 
     /* Prevent X509_V_ERR_MISSING_SUBJECT_KEY_IDENTIFIER */
     if (!AdaptKeyIDExtension(cert, &ext_ctx, "subjectKeyIdentifier", "hash",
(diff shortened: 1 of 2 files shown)


tool-openssl/req.cc
```diff
@@ -904,8 +901,21 @@ bool reqTool(const args_list_t &args) {
...
+    if (req_conf) {
+      const char *default_keyfile =
+          NCONF_get_string(req_conf.get(), REQ_SECTION, DEFAULT_KEYFILE);
+      keyout = default_keyfile != NULL ? default_keyfile : "";
+    } else {
+      keyout = "privkey.pem";
+    }
+  }
+
+  if (!keyout.empty() && !WritePrivateKey(keyout, passout, pkey, cipher)) {
```
justsmth: Be sure to call `SetUmaskForPrivateKey()` prior to writing the private key.  

Does the output file further down need a different umask?  If so, we might need to ensure it's written first, or otherwise find a way to undo this masking.

nhatnghiho: I removed this because I call `SetUmaskForPrivateKey()` inside `WritePrivateKey()` already ([link](https://github.com/aws/aws-lc/blob/2d29913019ff9286f00ddb0b9d3f25654e1a38a0/tool-openssl/req.cc#L696)) and this was a duplicate. The output file below doesn't write private key so I think it is okay? Not sure if we need to undo the masking though...

justsmth: > Not sure if we need to undo the masking though...


What permissions does Openssl place on the other file?  Is it the same permissions as are placed on the private key?

nhatnghiho: How do you check the permissions?

justsmth: In most shells, it's `ls -l` to get a list including the permissions. For individual files you can use the `stat` command.

tool-openssl/req.cc
```diff
@@ -694,13 +696,8 @@ static bool WritePrivateKey(std::string &out_path,
...
   SetUmaskForPrivateKey();
-  if (out_path.empty()) {
-    // Default to privkey.pem in the current directory
-    out_path = "privkey.pem";
-    fprintf(stderr, "Writing private key to %s (default)\n", out_path.c_str());
-  } else {
-    fprintf(stderr, "Writing private key to %s\n", out_path.c_str());
-  }
+
+  fprintf(stderr, "Writing private key to %s\n", out_path.c_str());
```
WillChilds-Klein, line 705: do we know that `out_path` will always be non-null/non-empty? looks like prior logic had a default value for this case. should we consider erroring out?

i see we move some of that logic below to L904, but are there any other call sites for WritePrivateKey?

tool-openssl/req.cc
```diff
@@ -904,8 +901,21 @@ bool reqTool(const args_list_t &args) {
...
-  if (!WritePrivateKey(keyout, passout, pkey, cipher)) {
+  // If keyout is not provided and we generated a new key (not loaded from -key):
+  // 1. If -config, use it to set keyout
+  // 2. If no -config, output key to privkey.pem (similar to how OpenSSL would
+  // default to the default openssl.conf file)
+  if (keyout.empty() && key_file_path.empty()) {
+    if (req_conf) {
+      const char *default_keyfile =
+          NCONF_get_string(req_conf.get(), REQ_SECTION, DEFAULT_KEYFILE);
+      keyout = default_keyfile != NULL ? default_keyfile : "privkey.pem";
```
WillChilds-Klein: can we cover this case?

tool-openssl/req.cc
nhatnghiho: I added a condition to only call `WritePrivateKey` when keyout  is non-empty

PR #2841, opened 2025-11-18
Verify size of mlen in ML-DSA external mu mode

jakemas:
### Issues:
Resolves #N/A

### Description of changes: 
When signing/verifying in ML-DSA the caller provides `mlen` the the length of the message `m` that is being signed/verified. Currently, we do no validation on the size of `mlen` when in pre-hash mode (called "external mu" in ML-DSA). Since the pre-hash is the output of a SHAKE256 hash function, it is of fixed size `ML_DSA_CRHBYTES`. This adds validation to check that, returning `-1` if not true to match the case where the context `ctx` has length too large.


crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c
```diff
@@ -205,6 +205,10 @@ int ml_dsa_sign_internal(ml_dsa_params *params,
     SHAKE_Final(mu, &state, ML_DSA_CRHBYTES);
   }
   else {
+    // When external_mu is true, m is expected to be exactly ML_DSA_CRHBYTES
+    if (mlen != ML_DSA_CRHBYTES) {
+      return -1;
```
dkostic: you'll skip the destruction of intermediate values with this.

crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c
```diff
@@ -492,6 +496,10 @@ int ml_dsa_verify_internal(ml_dsa_params *params,
...
```
dkostic: same as above

crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c
jakemas: Thank you, great spot, added in [2185013](https://github.com/aws/aws-lc/pull/2841/commits/21850132e1d4f6c0f4864302e9fbd0b7581ce1a2)

crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c
jakemas: Thank you, great spot, added in [2185013](https://github.com/aws/aws-lc/pull/2841/commits/21850132e1d4f6c0f4864302e9fbd0b7581ce1a2)

manastasova (commented):
Why not checking for the correctness of the length at the beginning of these two functions and, thus, avoid the memory allocations and function calls?

jakemas:
> Why not checking for the correctness of the length at the beginning of these two functions and, thus, avoid the memory allocations and function calls?

Ahh yeah that could work too, follow the exisiting siglen check with an mlen check... I'm happy with either

```
  if(siglen != params->bytes) {
    return -1;
  }
  
  if (mlen != ML_DSA_CRHBYTES) {
      return -1;
  }
```

manastasova:
I can see it being at the very beginning of the functions (along with the external mu check):
```
if (external_mu && mlen != ML_DSA_CRHBYTES) {
       return -1;
}
```

dkostic (approved):

crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c
```diff
@@ -161,7 +161,7 @@ int ml_dsa_keypair(ml_dsa_params *params, uint8_t *pk, uint8_t *sk, uint8_t *see
 *              - uint8_t *sk:     pointer to bit-packed secret key
 *              - int external_mu: indicates input message m is to be processed as mu
 *
-* Returns 0 (success) or -1 (context string too long)
+* Returns 0 (success) or -1 (context string or mlen too long)
```
manastasova: Or too short? I could not see where the length of context is checked.

jakemas: addressed in [5b1c5a5](https://github.com/aws/aws-lc/pull/2841/commits/5b1c5a5d89585775dfa2752e28b8984422ee5a6d)

crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c
```diff
@@ -469,6 +473,11 @@ int ml_dsa_verify_internal(ml_dsa_params *params,
   if(siglen != params->bytes) {
     return -1;
   }
+
+  if (external_mu && mlen != ML_DSA_CRHBYTES) {
+       return -1;
```
manastasova: alignment here is off

crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c
```diff
@@ -469,6 +473,11 @@ int ml_dsa_verify_internal(ml_dsa_params *params,
   if(siglen != params->bytes) {
```
manastasova, line 473: we could add space here `if <space> ( ....`

crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c
jakemas: ty fixed in [ff7fbb9](https://github.com/aws/aws-lc/pull/2841/commits/ff7fbb9da25dcf8469c85e585f1fc2ea9bc40e7f)

crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c
jakemas: There is a mix of if's with and without -- for now I can live with this

manastasova (approved):

PR #2851, opened 2025-11-19
Add EVP_bf_cfb64

nhatnghiho:
### Description of changes: 
Add `EVP_bf_cfb64` cipher to public API

### Testing:
Unit test added


crypto/decrepit/blowfish/blowfish.c
```diff
@@ -640,3 +640,5 @@ const EVP_CIPHER *EVP_bf_ecb(void) { return &bf_ecb; }
 const EVP_CIPHER *EVP_bf_cbc(void) { return &bf_cbc; }
 
 const EVP_CIPHER *EVP_bf_cfb(void) { return &bf_cfb; }
+
+const EVP_CIPHER *EVP_bf_cfb64(void) { return &bf_cfb; }
```
justsmth, line 644: I think this is the only required change in this file, with the others just being formatting. To reduce the noise, could the non-essential changes above be reverted?

nhatnghiho: It's my autoformatter using our `clang-format` config. But I'd imagine that anyone with the same formatting config will also inadvertently change the file later on?

Issue #1098, opened 2023-07-18
Add soname


bkmgit:
### Problem:

Would like to package this for Fedora.  For shared libraries, it is helpful to have a soname. Would a pull request to add a soname be required.

### Solution:

Add soname in CMake configuration for shared library builds

* **Does this change any public APIs?*** If yes, explain.
Does not change any APIs, though when using shared library, linking to an explicit version will be made easier.

* **Which algorithm(s) will this impact?**
Does not change any algorithms.

### Requirements / Acceptance Criteria:

What must a solution address in order to solve the problem? How do we know the solution is complete?
Shared libraries should have sonames,  should generate .so, .so.x and .so.x.x.x - may want to use the sonames used for libssl and libcrypto used by the associated boringssl library if expect to keep compatibility, otherwise use sonames based on aws-lc releases

* **RFC links:** No relevant RFCs
* **Related Issues:** Link any relevant issues
* **Will the Usage Guide or other documentation need to be updated?** No changes needed
* **Testing:** No changes to tests needed.
  * **Will this change trigger AWS LibCrypto Formal Verification changes?** No changes needed
  * **Should this change be fuzz tested?** No changes needed

### Out of scope:

Is there anything the solution will intentionally NOT address?

skmcgrail:
Thanks for reaching out @bkmgit , I'm interested in learning more about your use case and packaging plan for Fedora.

The AWS-LC API is not considered stable, and we have made breaking API changes that wouldn't be considered SemVer compliant. This obviously poses challenges for packaging for OS package managers like Fedora. Are you looking to package AWS-LC for your own consumption, or is this something you were planning to contribute directly to Fedora? As indicated in https://github.com/aws/aws-lc/issues/804, long-term we want to support packaging AWS-LC for operating systems (particularly Amazon Linux), but we don't have anything more specific that we can share in terms of timelines or a public roadmap.

As for the question of setting the `SONAME`, this is currently defaulting to an `libcrypto.so` and `libssl.so` respectively based on which library is built. We are open to setting `SONAME` to contain the release version number such as `libcrypto-1.12.1.so` which would correspond to the  [v1.12.1](https://github.com/aws/aws-lc/releases/tag/v1.12.1) of AWS-LC.

bkmgit:
It is a dependency of [Apache Arrow](https://github.com/apache/arrow/blob/main/cpp/cmake_modules/ThirdpartyToolchain.cmake#L4716)

[...]

PR #2852, opened 2025-11-20
RFC: Separate out the OpenSSL shim from AWS-LC

stewartsmith:
### Issues:
Part of addressing https://github.com/aws/aws-lc/issues/1098

### Description of changes: 
This enables `aws-lc` and `openssl` to be installed concurrently, including the `-devel` packages. We keep a shim around that can be packaged as a separate compat RPM in order to build unmodified software against `aws-lc` rather than `openssl`. I've also added/modified pkgconfig to have an explicit aws-lc one as well as an openssl compat one, and by putting these in separate RPMs it's possible to build packages in a way that allows individual packages to opt-in to using AWS-LC *or* to have the `-openssl-compat-devel` subpackage come in and let unmodified software link against it.

The headers are now `include/aws-lc/openssl` rather than `include/openssl` to clearly differentiate `aws-lc` from `openssl`, and enable a situation where both `-devel` packages can be installed simultaneously.

i.e. this results in AWS-LC very much having its own identity rather than when installed system-wide entirely pretending to be OpenSSL.

### Call-outs:
Probably the most important change is this one:
```
-          NAMELINK_SKIP)
+          NAMELINK_COMPONENT Development)
```
That enables the unversioned `.so` libraries to exist, and thus be installed correctly, and enables the explicit linking against `libssl-awslc.so` rather than `libssl.so`.

### Testing:
I've built an RPM with this change, and used it to do test builds of a number of packages in Amazon Linux 2023 against `aws-lc` rather than OpenSSL. I have not finished building everything, nor have I run any tests that aren't just the `%check` sections of said packages.

Testing I've done has focused on was getting the `-devel`, `-openssl-compat-devel`, and unversioned `.so` for the `-devel` package going.

This is very RFC for the moment, suitable for discussion.


DemiMarie:
Are there ways that the community could help with this?

PR #2863, opened 2025-11-24
Add CTR-DRBG derivation function

torben-hansen:
### Description of changes: 

IG D.L requires that the entropy argument has full-entropy. This won't happen using the passive method. Hence, implement a derivation function to compensate.

Derivation function is implemented a described in SP800-90A using a generation counter for the nonce. The nonce is then unique per process up to 2^128 life-time initializations.

The new ctr-drbg derivation function interface is not exported. Instead it's only used internally. I kept the existing ctr-drbg interface around in an attempt to make the change minimal. `RAND_bytes()` will continue to use the existing ctr-drbg implementation except if in FIPS and if not using the special CPU Jitter mode.

### Call-outs:

All FIPS related tests are adjusted to use the new ctr-drbg derivation functions.

#### performance on Arm64

Now
```
$ ./tool/bssl speed -filter RNG
Did 1737000 RNG (16 bytes) operations in 1000437us (1736241.3 ops/sec): 27.8 MB/s
Did 1597500 RNG (256 bytes) operations in 1000066us (1597394.6 ops/sec): 408.9 MB/s
Did 1155250 RNG (1350 bytes) operations in 1000043us (1155200.3 ops/sec): 1559.5 MB/s
Did 445000 RNG (8192 bytes) operations in 1000808us (444640.7 ops/sec): 3642.5 MB/s
Did 254000 RNG (16384 bytes) operations in 1000784us (253801.0 ops/sec): 4158.3 MB/s
```

Before
```
$ ./tool/bssl speed -filter RNG
Did 4173250 RNG (16 bytes) operations in 1000045us (4173062.2 ops/sec): 66.8 MB/s
Did 3481750 RNG (256 bytes) operations in 1000064us (3481527.2 ops/sec): 891.3 MB/s
Did 1899000 RNG (1350 bytes) operations in 1000167us (1898682.9 ops/sec): 2563.2 MB/s
Did 524000 RNG (8192 bytes) operations in 1001596us (523165.0 ops/sec): 4285.8 MB/s
Did 278000 RNG (16384 bytes) operations in 1001963us (277455.4 ops/sec): 4545.8 MB/s
```

### Testing:

Added lots of KATs for the new ctr-drbg with df interface.

### DIFF
--- a/crypto/fipsmodule/rand/ctrdrbg.c
+++ b/crypto/fipsmodule/rand/ctrdrbg.c
@@ -253,8 +253,9 @@ int CTR_DRBG_init(CTR_DRBG_STATE *drbg,
     seed_material[i] ^= kInitMask[i];
   }
 
+  // keylen is 32 bytes and block length is AES_BLOCK_SIZE.
   drbg->ctr = aes_ctr_set_key(&drbg->ks, NULL, &drbg->block, seed_material, 32);
-  OPENSSL_memcpy(drbg->counter, seed_material + 32, 16);
+  OPENSSL_memcpy(drbg->counter, seed_material + 32, AES_BLOCK_SIZE);
   drbg->reseed_counter = 1;
 
   return 1;
@@ -316,8 +317,9 @@ int CTR_DRBG_init_df(CTR_DRBG_STATE *drbg,
     seed_material[i] ^= kInitMask[i];
   }
 
+  // keylen is 32 bytes and block length is AES_BLOCK_SIZE.
   drbg->ctr = aes_ctr_set_key(&drbg->ks, NULL, &drbg->block, seed_material, 32);
-  OPENSSL_memcpy(drbg->counter, seed_material + 32, 16);
+  OPENSSL_memcpy(drbg->counter, seed_material + 32, AES_BLOCK_SIZE);
   drbg->reseed_counter = 1;
 
   return 1;


crypto/fipsmodule/rand/ctrdrbg.c
```diff
@@ -12,21 +12,205 @@
...
+  // Compute required zero-padding (might be zero). +1 because of 0x80
+  // delimiter in S.
+  size_t df_s_zero_padding = (-(df_s_prefix_length + entropy_input_len + 1)) & (AES_BLOCK_SIZE-1);
+
+  // Knowing the padding, we can compute the full length of S.
+  size_t df_s_length = df_s_prefix_length + entropy_input_len + 1 + df_s_zero_padding;
+
+  // Since the length of the IV is fixed when the output length is fixed, we can
+  // compute the length of IV || S.
+  static const size_t df_iv_length = 16;
```
skmcgrail:
```suggestion
  static const size_t df_iv_length = AES_BLOCK_SIZE;
```
Perhaps?

crypto/fipsmodule/rand/ctrdrbg.c
```diff
@@ -12,21 +12,205 @@
...
+
+  // Zeroise to avoid zero padding later.
+  uint8_t *IV_S = OPENSSL_zalloc(df_iv_s_length);
+  if (IV_S == NULL) {
+    return 0;
+  }
+
+  // IV_S = IV || L || N || entropy_input
+  CRYPTO_store_u32_be(IV_S + df_iv_length, (uint32_t) entropy_input_len);
+  CRYPTO_store_u32_be(IV_S + df_iv_length + 4, (uint32_t) CTR_DRBG_ENTROPY_LEN);
```
skmcgrail:
```
CRYPTO_store_u32_be(IV_S + df_iv_length + 4, (uint32_t) CTR_DRBG_ENTROPY_LEN); 
```

I understand it's `4` since `entropy_input_len` is a 4 byte value, but feels off to have a "magic number" here. Perhaps `sizeof(uint32_t)` would be sufficient?

crypto/fipsmodule/rand/ctrdrbg.c
```diff
@@ -12,21 +12,205 @@
...
+
+// Implementation of derivation function assumes CTR_DRBG_ENTROPY_LEN is 48.
+// That is, assumes the number of output bytes is 48. In addition, value must be
+// a multiple of the AES block size.
+OPENSSL_STATIC_ASSERT(CTR_DRBG_ENTROPY_LEN == 48, ctr_drbg_df_assumes_ctr_drbg_entropy_len_is_equal_to_48)
+OPENSSL_STATIC_ASSERT(CTR_DRBG_ENTROPY_LEN % AES_BLOCK_SIZE == 0,
+                      not_a_multiple_of_AES_block_size)
+
+static void init_df_bcc_once(void) {
+  static const uint8_t staticBccKey[32] = {
```
skmcgrail, line 61: Worth documenting that this key is the one defined `10.3.2 (8)`, and not arbitrarily chosen.

crypto/fipsmodule/self_check/self_check.c
```diff
@@ -1873,22 +1884,24 @@ static int boringssl_self_test_fast(void) {
...
+      0x2b, 0x65, 0x12, 0x50, 0x23, 0xac, 0x61, 0x3d,
+      0x64, 0xa9, 0xb6, 0x29, 0x5f, 0xf5, 0xcb, 0x1f,
+      0x9b, 0xd6, 0xe3, 0x92, 0x2f, 0x2f, 0x75, 0xda,
+      0x85, 0x81, 0xbe, 0xc4, 0x2b, 0xf9, 0x8a, 0x32,
   };
   CTR_DRBG_STATE drbg;
-  if (!CTR_DRBG_init(&drbg, kDRBGEntropy, kDRBGPersonalization,
-                     sizeof(kDRBGPersonalization)) ||
-      !CTR_DRBG_generate(&drbg, output, sizeof(kDRBGOutput), kDRBGAD,
+  if (!CTR_DRBG_init_df(&drbg, kDRBGEntropy, kDRBGPersonalization,
```
skmcgrail, line 1897: Do we not need the KAT/POST for the non-DF variant since it's used with jitterentropy still??

util/fipstools/acvp/modulewrapper/modulewrapper.cc
```diff
@@ -543,9 +543,9 @@ static bool GetConfig(const Span<const uint8_t> args[],
         "reseedImplemented": true,
```
skmcgrail, line 543: Confused on whether we need to be testing both (for passive df case, and for ctrdrbg with jitter)

crypto/fipsmodule/rand/rand.c
```diff
@@ -73,11 +73,12 @@
 #endif
 
 #if defined(FIPS_ENTROPY_SOURCE_JITTER_CPU)
-static const unsigned kReseedInterval = 16777216;
+  static const unsigned kReseedInterval = 16777216;
```
WillChilds-Klein, line 76: nitpick: not your change, but the variable naming here is a bit confusing w/r/t NIST.SP.800-90Ar1, which refers to "Maximum number of requests between reseeds" as `reseed_interval`, i.e. what we call `kMaxReseedCount` elsewhere

crypto/fipsmodule/rand/ctrdrbg_test.cc
```diff
@@ -127,3 +127,77 @@ TEST(CTRDRBGTest, TestVectors) {
...
+static bool nonceIsGreater(const uint8_t nonce1[CTR_DRBG_NONCE_LEN],
+  const uint8_t nonce2[CTR_DRBG_NONCE_LEN]) {
+
+  for (size_t i = 0; i < CTR_DRBG_NONCE_LEN; i++) {
+    if (nonce2[i] > nonce1[i]) {
+      return true;
+    }
+    if (nonce2[i] < nonce1[i]) {
+      return false;
+    }
```
WillChilds-Klein, line 188: does this make an implicit assumption about endianness? is it valid on both LE and BE systems?

crypto/fipsmodule/rand/ctrdrbg.c
```diff
@@ -12,21 +12,205 @@
...
+  // 10.3.2 (3): N = CTR_DRBG_ENTROPY_LEN
+  // 10.3.2 (4+5): S = L || N || input_string || 0x80 || 0x00-padded to block size
+  // 10.3.2 (9.1): IV = i || 0x00-padded to block size
+  //
+  // Using these values, we construct IV_S = IV || S used in 10.3.2 (9.2).
+
+  static const size_t df_s_prefix_length = 8; // 2x 32-bit integers
+  // Compute required zero-padding (might be zero). +1 because of 0x80
+  // delimiter in S.
+  size_t df_s_zero_padding = (-(df_s_prefix_length + entropy_input_len + 1)) & (AES_BLOCK_SIZE-1);
```
WillChilds-Klein, line 144: here and elsewhere, can these be `const`?

crypto/fipsmodule/rand/ctrdrbg_test.cc
WillChilds-Klein: hm, if we always represent the nonce as a byte array and don't do anything with int types, i suppose it's fine.

crypto/fipsmodule/rand/ctrdrbg.c
torben-hansen: Done

crypto/fipsmodule/rand/ctrdrbg.c
torben-hansen: They could, but it's not a general style choice.

crypto/fipsmodule/rand/ctrdrbg.c
torben-hansen: Done

crypto/fipsmodule/rand/ctrdrbg.c
torben-hansen: changed to `sizeof(uint32_t)`. And did other places as well.

crypto/fipsmodule/rand/ctrdrbg_test.cc
torben-hansen: I think this should work for both BE and LE. As you point out.

crypto/fipsmodule/rand/ctrdrbg.c
```diff
@@ -220,6 +493,86 @@ int CTR_DRBG_generate(CTR_DRBG_STATE *drbg, uint8_t *out, size_t out_len,
...
+
+    todo &= ~(AES_BLOCK_SIZE-1);
+    const size_t num_blocks = todo / AES_BLOCK_SIZE;
+
+    if (drbg->ctr) {
+      OPENSSL_memset(out, 0, todo);
+      ctr32_add(drbg, 1);
+      drbg->ctr(out, out, num_blocks, &drbg->ks, drbg->counter);
+      ctr32_add(drbg, (uint32_t)(num_blocks - 1));
+    } else {
```
WillChilds-Klein, line 546: how would this case get hit? when is the `drbg->ctr` function pointer ever not set?

crypto/fipsmodule/rand/ctrdrbg.c
```diff
@@ -12,21 +12,207 @@
...
+DEFINE_STATIC_MUTEX(generation_nonce_lock)
+
+// Implementation of derivation function assumes CTR_DRBG_ENTROPY_LEN is 48.
+// That is, assumes the number of output bytes is 48. In addition, value must be
+// a multiple of the AES block size.
+OPENSSL_STATIC_ASSERT(CTR_DRBG_ENTROPY_LEN == 48, ctr_drbg_df_assumes_ctr_drbg_entropy_len_is_equal_to_48)
+OPENSSL_STATIC_ASSERT(CTR_DRBG_ENTROPY_LEN % AES_BLOCK_SIZE == 0,
+                      not_a_multiple_of_AES_block_size)
+
+static void init_df_bcc_once(void) {
```
WillChilds-Klein, line 58: minor nitpick: i usually consider "once" functions to themselves be idempotent, but it looks like idempotency here is granted by `CRYPTO_once` and not this function. suggest dropping the `_once` suffix

crypto/fipsmodule/rand/ctrdrbg.c
```diff
@@ -12,21 +12,207 @@
...
+// only keeping the last resulting block. |input_len| must be a multiple of the
+// AES block size (16) and non-zero. |output| must be zeroised.
+// Return 1 on success, 0 otherwise.
+static int df_bcc(uint8_t *input, size_t input_len,
+  uint8_t output[AES_BLOCK_SIZE], AES_KEY *key) {
+
+  if (input_len < AES_BLOCK_SIZE || input_len % AES_BLOCK_SIZE != 0) {
+    return 0;
+  }
+
```
WillChilds-Klein, line 89: step one of "BCC Process" in 10.3.3 calls for initializing the chaining value (`output` here since you're doing it in-place) to all 0's

```
chaining_value = 0outlen. Comment: Set the first chaining value to outlen zeros.
```

looks like we're relying on `OPENSSL_cleanse` on L175 to do that. suggest adding a comment that this function expects `output` to already be zero'd before calling

crypto/fipsmodule/rand/ctrdrbg.c
```diff
@@ -12,21 +12,207 @@
...
+  if (IV_S == NULL) {
+    return 0;
+  }
+
+  // IV_S = IV || L || N || entropy_input
+  CRYPTO_store_u32_be(IV_S + df_iv_length, (uint32_t) entropy_input_len);
+  CRYPTO_store_u32_be(IV_S + df_iv_length + sizeof(uint32_t), (uint32_t) CTR_DRBG_ENTROPY_LEN);
+  OPENSSL_memcpy(IV_S + df_iv_length + df_s_prefix_length, entropy_input, entropy_input_len);
+
+  // IV_S = IV || L || N || entropy_input || 0x80
```
WillChilds-Klein, line 165: 10.3.2 Block_Cipher_df  step 9.1 in the spec says


```
IV = i || 0^(outlen - len (i))    Comment: The 32-bit integer representation of i is padded with zeros to outlen bits.
```

so, you're implicitly initializing `IV` as 0 IV in the `OPENSSL_zalloc` above, and intentionally not touching it here and below until you start doing the `df_bcc` calls below. is that right?

if so, where are you incrementing `i`, i.e. the value of `IV` between calls to `df_bcc` as the spec requires in step `9.3 i = i + 1`?

crypto/fipsmodule/rand/ctrdrbg.c
```diff
@@ -63,16 +249,69 @@ int CTR_DRBG_init(CTR_DRBG_STATE *drbg,
...
 
-  // kInitMask is the result of encrypting blocks with big-endian value 1, 2
-  // and 3 with the all-zero AES-256 key.
-  static const uint8_t kInitMask[CTR_DRBG_ENTROPY_LEN] = {
-      0x53, 0x0f, 0x8a, 0xfb, 0xc7, 0x45, 0x36, 0xb9, 0xa9, 0x63, 0xb4, 0xf1,
-      0xc4, 0xcb, 0x73, 0x8b, 0xce, 0xa7, 0x40, 0x3d, 0x4d, 0x60, 0x6b, 0x6e,
-      0x07, 0x4e, 0xc5, 0xd3, 0xba, 0xf3, 0x9d, 0x18, 0x72, 0x60, 0x03, 0xca,
-      0x37, 0xa6, 0x2a, 0x74, 0xd1, 0xa2, 0xf5, 0x8e, 0x75, 0x06, 0x35, 0x8e,
-  };
+  drbg->ctr = aes_ctr_set_key(&drbg->ks, NULL, &drbg->block, seed_material, 32);
```
WillChilds-Klein, line 257: nitpick: can we use a constant for AES 256 key size here instead of 32? same with 32 and 16 (presumably bock size) on line below

crypto/fipsmodule/rand/ctrdrbg.c
```diff
@@ -63,16 +249,69 @@ int CTR_DRBG_init(CTR_DRBG_STATE *drbg,
...
+  OPENSSL_memcpy(drbg->counter, seed_material + 32, 16);
+  drbg->reseed_counter = 1;
+
+  return 1;
+}
+
+static void nonce_increment(uint8_t counter[CTR_DRBG_NONCE_LEN]) {
+  size_t index = CTR_DRBG_NONCE_LEN, carry = 1;
```
WillChilds-Klein, line 265: why is `carry a size_t`? `counter` is 16 bytes long and you're casting `carry` to a single byte, lopping off the 48 high bits (assuming 64 bit system)

crypto/fipsmodule/rand/ctrdrbg.c
```diff
@@ -149,6 +385,43 @@ int CTR_DRBG_reseed(CTR_DRBG_STATE *drbg,
   return 1;
 }
 
+int CTR_DRBG_reseed_df(CTR_DRBG_STATE *drbg,
+                    const uint8_t entropy[CTR_DRBG_ENTROPY_LEN],
+                    const uint8_t *additional_data,
+                    size_t additional_data_len) {
```
WillChilds-Klein, line 393: is this stuff from 8.7.1 or 8.7.2? if neither where is defined in the spec?

crypto/fipsmodule/rand/ctrdrbg.c
```diff
@@ -220,6 +493,86 @@ int CTR_DRBG_generate(CTR_DRBG_STATE *drbg, uint8_t *out, size_t out_len,
...
+    }
+  }
+
+  // kChunkSize is used to interact better with the cache. Since the AES-CTR
+  // code assumes that it's encrypting rather than just writing keystream, the
+  // buffer has to be zeroed first. Without chunking, large reads would zero
+  // the whole buffer, flushing the L1 cache, and then do another pass (missing
+  // the cache every time) to “encrypt” it. The code can avoid this by
+  // chunking.
+  static const size_t kChunkSize = 8 * 1024;
```
WillChilds-Klein, line 530: how did you chose this value? does L1 cache size vary by architecture/platform?

crypto/fipsmodule/rand/ctrdrbg.c
torben-hansen: Keeping as-is because it's a naming convention used throughout the code-base.

crypto/fipsmodule/rand/ctrdrbg.c
torben-hansen: It's in the function documentation (:
> |output| must be zeroised.

crypto/fipsmodule/rand/ctrdrbg.c
torben-hansen: > so, you're implicitly initializing IV as 0 IV in the OPENSSL_zalloc above, and intentionally not touching it here and below until you start doing the df_bcc calls below. is that right?

Correct!

> if so, where are you incrementing i, i.e. the value of IV between calls to df_bcc as the spec requires in step 9.3 i = i + 1?

The increment step (only two increments are needed) is done using the function `u32_add_one(IV_S)` between the `df_bcc` function calls below L179+L183.

crypto/fipsmodule/rand/ctrdrbg.c
torben-hansen: added a comment instead.

crypto/fipsmodule/rand/ctrdrbg.c
torben-hansen: I think I had this to be `uint16_t` but then changed it to `size_t` because `index` is an array index. But true, it could be `uint16_t` and still be correct. But just keeping as-is.

crypto/fipsmodule/rand/ctrdrbg.c
torben-hansen: You mean the naming or the algorithm? Sorry not sure I understand the question.

Algorithm is 10.2.1.4.2.

crypto/fipsmodule/rand/ctrdrbg.c
torben-hansen: It's a pre-existing value that was copied from `CTR_DRBG_generate`. I haven't experimented with it.

Yes, caches can, and do, vary.

crypto/fipsmodule/rand/ctrdrbg.c
torben-hansen: This is also just copied from `CTR_DRBG_generate`. ~~It does look like dead code.~~ Should be handled on Tip of AWS-LC.

This derivation function change is a backport and not going into tip of main.

crypto/fipsmodule/rand/ctrdrbg.c
WillChilds-Klein: ah, unclear question on my part. i meant specifically `additional_data` and its length

crypto/fipsmodule/rand/ctrdrbg.c
WillChilds-Klein: oof, i totally missed that. thanks!

crypto/fipsmodule/rand/ctrdrbg.c
torben-hansen: I see, the (maximum allowed) length is governed by `Table 3: Definitions for the CTR_DRBG` in SP800-90A. The upper bound is 2^32 bytes.

torben-hansen:
Checked all failing dimensions. I believe they are due to outdated CI and unrelated to this change.

PR #2876, opened 2025-12-02
Speed up legacy AVX CI

samuel40791765:
Legacy AVX CI builds have been taking more 2 hours, especially for 32-bit runs. This is an attempt to speed things up.

### DIFF
--- a/.github/workflows/linux_x86_omnibus.yml
+++ b/.github/workflows/linux_x86_omnibus.yml
@@ -171,15 +171,20 @@ jobs:
             ./tests/ci/run_legacy_build.sh
 
   legacy_avx_tests:
-    name: legacy-avx-${{ matrix.image }}-${{ matrix.build32 && 'x86' || 'x86_64' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/linux_x86_omnibus.yml
```diff
@@ -213,5 +213,51 @@ jobs:
             AWSLC_32BIT
           run: |
             source /opt/compiler-env/setup-${{ matrix.compiler }}.sh
-            ./tests/ci/run_legacy_avx_tests.sh
+            ./tests/ci/run_legacy_avx_tests.sh MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX
```
justsmth: Instead of nearly identical separate jobs for this, how hard would it be to make this a "strategy" dimension that gets factored into a single job?

samuel40791765: Sure that was the next step, just wanted to see how much the speed up would be first :)

justsmth (approved):

PR #2877, opened 2025-12-03
Implement enc CLI

nhatnghiho:
### Description of changes: 
Implement enc CLI with the following options:
- e
- d
- aes-128-cbc
- K
- in
- out
- iv

### Testing:
Unit tests


tool-openssl/enc.cc
```diff
@@ -0,0 +1,183 @@
...
+
+  if (cipher_name.empty()) {
+    cipher_name = "aes-128-cbc";
+  } else {
+    cipher_name = cipher_name.substr(1);
+  }
+  const EVP_CIPHER *cipher = EVP_get_cipherbyname(cipher_name.c_str());
+
+  int iv_length = EVP_CIPHER_iv_length(cipher);
+  bssl::UniquePtr<uint8_t> iv((uint8_t *)OPENSSL_zalloc(EVP_MAX_IV_LENGTH));
```
justsmth: Shouldn't these pointers be to arrays?
```
bssl::UniquePtr<uint8_t[]> iv((uint8_t *)OPENSSL_zalloc(EVP_MAX_IV_LENGTH));
```

tool-openssl/enc.cc
```diff
@@ -0,0 +1,183 @@
...
+      return false;
+    }
+  }
+
+  if (cipher_name.empty()) {
+    cipher_name = "aes-128-cbc";
+  } else {
+    cipher_name = cipher_name.substr(1);
+  }
+  const EVP_CIPHER *cipher = EVP_get_cipherbyname(cipher_name.c_str());
```
justsmth, line 114: Check for NULL.
```
  if (cipher == nullptr) {
    fprintf(stderr, "Error: Unknown cipher %s\n",
            cipher_name.c_str());
    return false;
  }
```

tool-openssl/enc.cc
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  // Read from stdin if no -in path provided
+  ScopedFILE in_file;
+  if (in_path.empty()) {
+    in_file.reset(stdin);
+  } else {
+    in_file.reset(fopen(in_path.c_str(), "rb"));
+    if (!in_file) {
+      fprintf(stderr, "Error: unable to load certificate from '%s'\n",
```
justsmth: Why assume it's a certificate?

tool-openssl/enc.cc
```diff
@@ -0,0 +1,183 @@
...
+    {"", kOptionalArgument, ""}};
+
+static bool HexToBinary(bssl::UniquePtr<uint8_t> &buffer,
+                        const std::string &hex_string, int size) {
+  // First validate that the string contains only valid hex characters
+  for (char c : hex_string) {
+    if (!OPENSSL_isxdigit(c)) {
+      return false;
+    }
+  }
```
justsmth: We should also verify that the hex_string matches the expected size.

tool-openssl/enc.cc
```diff
@@ -0,0 +1,183 @@
...
+  int inlen, outlen;
+
+  while ((inlen = fread(inbuf, 1, sizeof(inbuf), in_file.get())) > 0) {
+    if (!EVP_CipherUpdate(ctx.get(), outbuf.get(), &outlen, inbuf, inlen)) {
+      fprintf(stderr, "Error: Cipher update failed\n");
+      return false;
+    }
+    BIO_write(output_bio.get(), outbuf.get(), outlen);
+  }
+
```
justsmth: [`fread` also returns 0 on error.](https://man7.org/linux/man-pages/man3/fread.3.html#RETURN_VALUE) This needs to call `feof` or `ferror` after the loop.

tool-openssl/enc.cc
```diff
@@ -0,0 +1,183 @@
...
+  if (!EVP_CipherInit_ex(ctx.get(), cipher, nullptr, key.get(), iv.get(),
+                         enc)) {
+    fprintf(stderr, "Error: Failed to initialize cipher\n");
+    return false;
+  }
+
+  // Process the input file
+  uint8_t inbuf[1024];
+  bssl::UniquePtr<uint8_t> outbuf(
+      (uint8_t *)OPENSSL_zalloc(1024 + EVP_CIPHER_block_size(cipher)));
```
justsmth: 1024 is "magic" -- this should declare it as a named constant and use that.

tool-openssl/enc.cc
```diff
@@ -0,0 +1,183 @@
...
+
+  bssl::UniquePtr<uint8_t> key((uint8_t *)OPENSSL_zalloc(EVP_MAX_KEY_LENGTH));
+
+  if (!hkey->empty()) {
+    if (!HexToBinary(key, *hkey, EVP_CIPHER_key_length(cipher))) {
+      fprintf(stderr, "Error: Invalid hex key value\n");
+      // pass_util::SensitiveStringDeleter(&hkey);
+      return false;
+    }
+    // pass_util::SensitiveStringDeleter(&hkey);
```
justsmth: Delete the commented-out lines.

tool-openssl/enc.cc
```diff
@@ -107,8 +113,13 @@ bool encTool(const args_list_t &args) {
...
 
-  int iv_length = EVP_CIPHER_iv_length(cipher);
-  bssl::UniquePtr<uint8_t> iv((uint8_t *)OPENSSL_zalloc(EVP_MAX_IV_LENGTH));
+  if (cipher == nullptr) {
+    fprintf(stderr, "Error: Unknown cipher %s\n", cipher_name.c_str());
+    return false;
+  }
+
+  unsigned int iv_length = EVP_CIPHER_iv_length(cipher);
+  bssl::UniquePtr<uint8_t[]> iv((uint8_t *)OPENSSL_zalloc(EVP_MAX_IV_LENGTH));
```
justsmth: If you unconditionally allocate `EVP_MAX_IV_LENGTH`, why not make `iv` a stack value?

tool-openssl/enc.cc
```diff
@@ -0,0 +1,204 @@
...
+                         enc)) {
+    fprintf(stderr, "Error: Failed to initialize cipher\n");
+    return false;
+  }
+
+  // Process the input file
+  uint8_t inbuf[BUF_SIZE];
+  bssl::UniquePtr<uint8_t[]> outbuf(
+      (uint8_t *)OPENSSL_zalloc(BUF_SIZE + EVP_CIPHER_block_size(cipher)));
+  int inlen = 0, outlen = 0;
```
justsmth, line 172: NP:  You could also keep `outbuf` on the stack as long as you assume a large enough block size. (I believe 16 is the largest block size we support.)

tool-openssl/enc.cc
```diff
@@ -0,0 +1,204 @@
...
+    }
+  } else {
+    if (iv_length != 0) {
+      fprintf(stderr, "Error: IV is required for cipher %s\n",
+              cipher_name.c_str());
+      return false;
+    }
+  }
+
+  bssl::UniquePtr<uint8_t[]> key((uint8_t *)OPENSSL_zalloc(EVP_MAX_KEY_LENGTH));
```
justsmth: Similarly here, `key` can be a stack value if you're always allocating a constant size, `EVP_MAX_KEY_LENGTH`.

justsmth (approved):

tool-openssl/enc.cc
```diff
@@ -0,0 +1,202 @@
...
+    if (!OPENSSL_isxdigit(c)) {
+      return false;
+    }
+  }
+
+  if (hex_string.size() != size * 2) {
+    return false;
+  }
+
+  BIGNUM *raw = NULL;
```
samuel40791765, line 39:
```suggestion
  bssl::UniquePtr<BIGNUM> raw;
```

tool-openssl/enc.cc
```diff
@@ -0,0 +1,202 @@
...
+  if (BN_hex2bn(&raw, hex_string.c_str()) == 0) {
+    return false;
+  }
+
+  int ret = BN_bn2binpad(raw, buffer, size);
```
samuel40791765, line 44: Can we use `BN_bn2bin_padded` instead? Seems like that's easier to use (judging from the function documentation).

https://github.com/aws/aws-lc/blob/ec39cb332cd0c01eab5ed8d17d282811f8f52a54/include/openssl/bn.h#L1015C9-L1016

tool-openssl/enc.cc
```diff
@@ -0,0 +1,202 @@
...
+
+  std::string in_path, out_path, hiv, cipher_name;
+  bssl::UniquePtr<std::string> hkey(new std::string);
+  bool help = false, e = false, d = false;
+
+  ordered_args::GetBoolArgument(&help, "-help", parsed_args);
+  ordered_args::GetString(&in_path, "-in", "", parsed_args);
+  ordered_args::GetString(&out_path, "-out", "", parsed_args);
+  ordered_args::GetBoolArgument(&e, "-e", parsed_args);
+  ordered_args::GetBoolArgument(&d, "-d", parsed_args);
```
samuel40791765, line 67: Can we use better naming variables for `e` and `d` here? It's a bit hard to understand.

tool-openssl/enc.cc
```diff
@@ -0,0 +1,202 @@
...
+  ordered_args::GetString(hkey.get(), "-K", "", parsed_args);
+  ordered_args::GetString(&hiv, "-iv", "", parsed_args);
```
samuel40791765, line 69:
```suggestion
  ordered_args::GetString(&hex_iv, "-iv", "", parsed_args);
```

`hex_iv` or simply `iv` would be easier to understand here. Same for `hkey` above.

tool-openssl/enc.cc
```diff
@@ -0,0 +1,202 @@
...
+
+  if (e && d) {
+    fprintf(stderr, "Error: -e and -d are mutually exclusive\n");
+    return false;
+  }
+
+  bool enc = true;
+  if (d) {
+    enc = false;
+  }
```
samuel40791765, line 94: Can we directly use `e`/`encrypt` here?

tool-openssl/enc.cc
```diff
@@ -0,0 +1,202 @@
...
+      return false;
+    }
+  }
+
+  if (cipher_name.empty()) {
+    cipher_name = "aes-128-cbc";
+  } else {
+    cipher_name = cipher_name.substr(1);
+  }
+  const EVP_CIPHER *cipher = EVP_get_cipherbyname(cipher_name.c_str());
```
samuel40791765, line 114: Do we support more ciphers than `aes-128-cbc` here? That would be awesome, but it's a bit hard to tell from the help options above. Might want to make adjustments or document that we support more than just `aes-128-cbc` somewhere

tool-openssl/enc.cc
```diff
@@ -0,0 +1,202 @@
...
+    if (ferror(in_file.get())) {
+      fprintf(stderr, "Error reading from '%s'.\n", in_path.c_str());
+      return false;
+    }
+
+    if (!EVP_CipherUpdate(ctx.get(), outbuf, &outlen, inbuf, inlen)) {
+      fprintf(stderr, "Error: Cipher update failed\n");
+      return false;
+    }
+    BIO_write(output_bio.get(), outbuf, outlen);
```
samuel40791765, line 190: We should check the return value here

tool-openssl/enc.cc
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  // Finalize
```
samuel40791765, line 193:
```suggestion
```
Bit redundant

tool-openssl/enc.cc
```diff
@@ -0,0 +1,202 @@
...
+  if (!EVP_CipherFinal_ex(ctx.get(), outbuf, &outlen)) {
+    fprintf(stderr, "Error: Cipher final failed\n");
+    return false;
+  }
+
+  BIO_write(output_bio.get(), outbuf, outlen);
```
samuel40791765, line 199: Same return value check should be done here

tool-openssl/internal.h
```diff
@@ -90,9 +90,10 @@ bool CRLTool(const args_list_t &args);
 bool asn1parseTool(const args_list_t &args);
 bool dgstTool(const args_list_t &args);
 bool dhparamTool(const args_list_t &args);
+bool ecTool(const args_list_t &args);
 bool ecparamTool(const args_list_t &args);
+bool encTool(const args_list_t &args);
```
samuel40791765, line 95: Always appreciate people who follow the alphabetical ordering

tool-openssl/enc.cc
```diff
@@ -0,0 +1,202 @@
...
+
+#include <openssl/evp.h>
+#include <stdio.h>
+#include <string.h>
+#include <algorithm>
+#include <iostream>
+#include "../tool/internal.h"
+#include "internal.h"
+
+#define BUF_SIZE 1024
```
samuel40791765, line 12: We should abstract `BUF_SIZE` from `req.cc` to a shared header.
https://github.com/aws/aws-lc/blob/ec39cb332cd0c01eab5ed8d17d282811f8f52a54/tool-openssl/req.cc#L31

Same for the other PR as well.

tool-openssl/enc.cc
```diff
@@ -0,0 +1,202 @@
...
+  uint8_t inbuf[BUF_SIZE];
+  uint8_t outbuf[BUF_SIZE + EVP_MAX_BLOCK_LENGTH];
+  int inlen = 0, outlen = 0;
+
+  for (;;) {
+    if (feof(in_file.get())) {
+      break;
+    }
+
+    inlen = fread(inbuf, 1, sizeof(inbuf), in_file.get());
```
samuel40791765, line 179: Can we use `ReadFromFD` here? And use `ScopedFD` for all the file descripters

nhatnghiho: Can we create a `ScopedFD` for `stdin`? My main reason for using `ScopedFILE` was because I couldn't find a straightforward way to use `ScopedFD` for `stdin`

tool-openssl/enc.cc
nhatnghiho: Only `aes-128-cbc` is supported for now. But I set up the logic to allow us to add new ciphers by simply adding a new argument like here https://github.com/aws/aws-lc/blob/c23b2ae88deec89091d8eeb37178fbef24c96919/tool-openssl/enc.cc#L23
Do you have any particular ciphers you want to add?

tool-openssl/enc.cc
nhatnghiho: I feel like the choice of buffer size among different commands can be a little arbitrary. Standardizing that across all commands might not be a good idea?

tool-openssl/enc.cc
nhatnghiho: `BN_hex2bn` requires `BIGNUM **outp` as a parameter so we can't use `bssl::UniquePtr<BIGNUM>` here :(

PR #2882, opened 2025-12-05
[tool-openssl] basic asn1parse support

skmcgrail:
### Issues:
Resolves P342458791

### Description of changes: 
Adds basic support for `openssl asn1parse` with the following arguments:
```
-in <inputFile>
-inform (PEM | DER)
```

The behavior of the command will default to OpenSSL's `-strictpem` flag behavior which was an optional feature. We can revisit if this determined to be required in order to relax the input restriction for valid PEM blocks.

### Call-outs:
The original asn1parse function used `ASN1_dump` rather then `ASN1_parse`, under the hood they are the same function, except that `ASN1_dump` can be given a flag to indicate that unknown data should be hex dumped out. As our BIO hexdump functions doesn't match OpenSSL's the output would be slightly different. For now I've opted to not support the `-dump` flag for the CLI, and chose to only expose `ASN1_parse` in the library. That way we could come back and add `-dump` and `ASN1_dump` with proper output without worrying about backwards compatibility concerns.

### Testing:
* Added a series of corpus files for some BER and DER encodings (the asn1parse tool / library components due allow some BER features e.g. indefinite length encoding).
* Used DER corpus files to seed a fuzzer for testing the `ASN1_parse` library function.

### DIFF
--- a/tool-openssl/asn1parse_test.cc
+++ b/tool-openssl/asn1parse_test.cc
@@ -12,9 +12,28 @@ struct TestCorpus {
   std::string name;
   std::string hex;
   std::string format;
-  bool awslc_success;
-  bool match_openssl;
+  bool awslc_success = false;
+  bool match_openssl = false;
+
+  // Constructor for explicit initialization
+  TestCorpus(const std::string &n, const std::string &h, const std::string &f,
+             bool aws_success, bool match_ssl)
+      : name(n),
+        hex(h),
+        format(f),
+        awslc_success(aws_success),
+        match_openssl(match_ssl) {}
+
+  // Default constructor - explicitly initialize all members
+  TestCorpus() : awslc_success(false), match_openssl(false) {}
 };
+
+// Custom printer for TestCorpus to avoid GoogleTest's byte-level printing
+static void PrintTo(const TestCorpus &corpus, std::ostream *os) {
+  *os << "TestCorpus{name=\"" << corpus.name << "\", format=\"" << corpus.format
+      << "\", awslc_success=" << (corpus.awslc_success ? "true" : "false")
+      << ", match_openssl=" << (corpus.match_openssl ? "true" : "false") << "}";
+}
 
 #define FORMAT_PEM "PEM"
 #define FORMAT_DER "DER"


crypto/asn1/asn1_par.c
```diff
@@ -101,3 +115,417 @@ const char *ASN1_tag2str(int tag) {
   }
   return tag2str[tag];
 }
+
+int ASN1_parse(BIO *bp, const unsigned char *pp, long len, int indent) {
+  return asn1_parse2(bp, &pp, len, 0, 0, indent, 0);
+}
+
```
justsmth, line 127: Since this is the public function, can we add some NULL checks here before we invoke the internal functions?

crypto/asn1/asn1_par.c
```diff
@@ -101,3 +115,417 @@ const char *ASN1_tag2str(int tag) {
...
+      if ((parse_result == 2) || (*current_pos >= total_end)) {
+        *object_length = *current_pos - start_pos;
+        break;
+      }
+    }
+  } else {
+    // Definite length constructed object
+    long remaining_length = *object_length;
+
+    while (*current_pos < constructed_end) {
```
justsmth, line 221: * The `constructed_end` declaration could be made (and validated) here.
* Is it possible for `*object_length > (total_end - *current_pos)`?

crypto/asn1/asn1_par.c
```diff
@@ -101,3 +115,417 @@ const char *ASN1_tag2str(int tag) {
   }
   return tag2str[tag];
 }
+
+int ASN1_parse(BIO *bp, const unsigned char *pp, long len, int indent) {
+  return asn1_parse2(bp, &pp, len, 0, 0, indent, 0);
+}
+
+// Constants
+#define ASN1_PARSE_MAXDEPTH 128
```
justsmth, line 129: 128 seems really high.  Is it realistic for it ever to reach depth 128? Would 64 be enough?

crypto/asn1/asn1_par.c
```diff
@@ -101,3 +115,417 @@ const char *ASN1_tag2str(int tag) {
...
+
+  if (!bp || !current_pos || !total_end || !original_start || !object_length) {
+    return 0;
+  }
+
+  if (BIO_write(bp, "\n", 1) <= 0) {
+    return 0;
+  }
+
+  if ((parse_flags == 0x21) && (*object_length == 0)) {
```
justsmth: What is `0x21`? `V_ASN1_CONSTRUCTED | V_ASN1_BOOLEAN`?

crypto/asn1/asn1_par.c
```diff
@@ -101,3 +115,417 @@ const char *ASN1_tag2str(int tag) {
...
+}
+
+// Helper function to parse constructed ASN.1 types (SEQUENCE, SET, etc.)
+static int asn1_parse_constructed_type(
+    BIO *bp, const unsigned char **current_pos, const unsigned char *total_end,
+    const unsigned char *original_start, long *object_length, int parse_flags,
+    int offset, int depth, int indent, int dump) {
+  const unsigned char *start_pos = *current_pos;
+  const unsigned char *constructed_end = *current_pos + *object_length;
+  int parse_result;
```
justsmth: If the declaration was inside the loops below, it could be declared `const`.

crypto/asn1/asn1_par.c
```diff
@@ -101,3 +115,417 @@ const char *ASN1_tag2str(int tag) {
...
+
+  if ((tag == V_ASN1_PRINTABLESTRING) || (tag == V_ASN1_T61STRING) ||
+      (tag == V_ASN1_IA5STRING) || (tag == V_ASN1_VISIBLESTRING) ||
+      (tag == V_ASN1_NUMERICSTRING) || (tag == V_ASN1_UTF8STRING) ||
+      (tag == V_ASN1_UTCTIME) || (tag == V_ASN1_GENERALIZEDTIME)) {
+    if (BIO_write(bp, ":", 1) <= 0) {
+      goto end;
+    }
+    if ((object_length > 0) &&
+        BIO_write(bp, (const char *)current_pos, (int)object_length) !=
```
justsmth: Is it possible for `object_length > INT_MAX`?

crypto/asn1/asn1_par.c
skmcgrail: To your second bullet point, I think we can validate that `constructed_end > total_end` is not true, and if it is we should stop because that would be invalid.

crypto/asn1/asn1_par.c
skmcgrail:
```
//   * 0x80: error occurred while parsing.
//   * 0x20: the encoding is constructed, not primitive.
//   * 0x01: indefinite-length constructed encoding.
```
Defined by `ASN1_get_object`, and documented in it's documentation.

So this would be `1 | V_ASN1_CONSTRUCTED`, technically.

It looks like that is how it is checked in most of the code, except for this one location. Will fix.

crypto/asn1/asn1_par.c
```diff
@@ -117,6 +118,8 @@ const char *ASN1_tag2str(int tag) {
 }
 
 int ASN1_parse(BIO *bp, const unsigned char *pp, long len, int indent) {
+  GUARD_PTR(bp);
+  GUARD_PTR(pp);
   return asn1_parse2(bp, &pp, len, 0, 0, indent, 0);
 }
```
justsmth, line 126: When `len` is negative, the logic in `asn1_parse2` moves `pp` backwards and return success.  Should this instead be an error?

crypto/asn1/asn1_par.c
```diff
@@ -224,244 +228,342 @@ static int asn1_parse_constructed_type(
...
+      d2i_ASN1_OCTET_STRING(NULL, &parse_pos, len);
+  if (octet_string != NULL && octet_string->length > 0) {
+    int printable = 1;
+    parse_pos = octet_string->data;
+
+    // Test whether the octet string is printable
+    for (int i = 0; i < octet_string->length; i++) {
+      if (((parse_pos[i] < ' ') && (parse_pos[i] != '\n') &&
+           (parse_pos[i] != '\r') && (parse_pos[i] != '\t')) ||
+          (parse_pos[i] > '~')) {
```
justsmth: NP: We could comment that, for our purposes, we only consider ASCII strings to be printable.

crypto/asn1/asn1_par.c
```diff
@@ -101,3 +118,527 @@ const char *ASN1_tag2str(int tag) {
...
+  const uint8_t *hex_data = object_start + header_len;
+
+  if (BIO_puts(bp, ":[") <= 0) {
+    return 0;
+  }
+  for (int i = 0; i < object_len; i++) {
+    if (BIO_printf(bp, "%02X", hex_data[i]) <= 0) {
+      return 0;
+    }
+  }
```
justsmth, line 517: This seems to align fairly closely with the OpenSSL v1.1 function [BIO_hex_string](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/include/openssl/bio.h#L640), implemented [here](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/bio/b_dump.c#L125-L148). Perhaps we could take this opportunity  to add this function, use it here, and improve our compatibility slightly?

crypto/asn1/asn1_par.c
```diff
@@ -101,3 +118,527 @@ const char *ASN1_tag2str(int tag) {
...
+  if (BIO_write(bp, "\n", 1) <= 0) {
+    return 0;
+  }
+
+  if ((parse_flags == (V_ASN1_CONSTRUCTED | 1)) && (*object_len == 0)) {
+    // Indefinite length constructed object
+    for (;;) {
+      const int parse_result = asn1_parse2(
+          bp, current_pos, (long)(total_end - *current_pos),
+          offset + (*current_pos - original_start), depth + 1, indent, dump);
```
justsmth, line 203: Although unlikely, it's possible for `offset + (*current_pos - original_start)` to overflow here. (As mentioned above, this can be mitigated by using type `long` for the offset.) We could check on the line above, but I think it might be better to add a few "sanity checks" at the beginning of `asn1_parse2` to verify that values like `length`, `offset` and `depth` are non-negative.

crypto/asn1/asn1_par.c
```diff
@@ -101,3 +118,527 @@ const char *ASN1_tag2str(int tag) {
...
+        ASN1_get_object(&current_pos, &object_length, &tag, &xclass, length);
+    if (parse_flags & 0x80) {
+      if (BIO_write(bp, "Error in encoding\n", 18) <= 0) {
+        goto end;
+      }
+      return_value = 0;
+      goto end;
+    }
+    header_length = (current_pos - object_start);
+    length -= header_length;
```
justsmth: NP: Could we rename `object_length` to `content_length` here -- `object_length` seems to imply that it includes the length of the header and content, but [it's only the content](https://github.com/aws/aws-lc/blob/8207492fafdd24c037e9b8f8ef227240c8653920/crypto/asn1/asn1_lib.c#L119-L120). On 586, we reduce `length` by the `header_length`, then on 611, compare `object_length` to the reduced `length` value, which made me think that we were double-counting the header (but we're not).

crypto/asn1/asn1_par.c
```diff
@@ -101,3 +118,527 @@ const char *ASN1_tag2str(int tag) {
...
+
+static int asn1_parse_octet_string_type(BIO *bp, const uint8_t *data, long len,
+                                        int dump, int *newline_printed) {
+  GUARD_PTR(bp);
+  GUARD_PTR(data);
+  GUARD_PTR(newline_printed);
+
+  const unsigned char *parse_pos = data;
+  int return_code = 0;
+  int dump_indent = 6;
```
justsmth, line 312: NP:
```
const int dump_indent = 6;
```

crypto/asn1/asn1_par.c
```diff
@@ -55,7 +55,24 @@
...
+static int asn1_parse_constructed_type(BIO *bp, const uint8_t **current_pos,
+                                       const uint8_t *total_end,
+                                       const uint8_t *original_start,
+                                       long *object_len, int parse_flags,
+                                       int offset, int depth, int indent,
+                                       int dump);
+static int asn1_parse_primitive_type(BIO *bp, const uint8_t *object_start,
+                                     const uint8_t *current_pos,
+                                     long object_len, int header_len, int tag,
+                                     int dump);
```
justsmth, line 75: NP: for these internal functions, there are a lot of conversions between pointer arithmetic and integer types. To avoid subtle overflow issues, it might be better to also use `long` for the "offset" parameters.
```
static int asn1_parse2(BIO *bp, const uint8_t **pp, long length,
                       long offset, int depth, int indent, int dump);
static int asn1_print_info(BIO *bp, int tag, int xclass, int constructed,
                           int indent);
static int asn1_parse_constructed_type(BIO *bp, const uint8_t **current_pos,
                                       const uint8_t *total_end,
                                       const uint8_t *original_start,
                                       long *object_len, int parse_flags,
                                       long offset, int depth, int indent,
                                       int dump);
static int asn1_parse_primitive_type(BIO *bp, const uint8_t *object_start,
                                     const uint8_t *current_pos,
                                     long object_len, long header_len,
                                     int tag, int dump);
```

crypto/asn1/asn1_par.c
```diff
@@ -101,3 +118,527 @@ const char *ASN1_tag2str(int tag) {
...
+      if (parse_result == 0) {
+        return 0;
+      }
+      remaining_length -= *current_pos - start_pos;
+    }
+  }
+  return 1;
+}
+
+static int asn1_parse_string_type(BIO *bp, const uint8_t *data, long len) {
```
justsmth, line 235: How paranoid should we be about the string. There are several different encodings that this could be, some of which could have an invalid length (e.g., multi-byte UTF8) or contain control characters.

Should we basically have this behave similar to `asn1_parse_octet_string_type`  in relation to validating ASCII content?

crypto/asn1/asn1_par.c
skmcgrail: It's not 100% the same, and I am going to bias towards not exposing it until we are asked for it.

crypto/asn1/asn1_par.c
skmcgrail: This function entry point is guarded by a check on the explicit ASN.1 types that are either numbers, digits, ascii, or subset of ascii, and UTF-8, which appears to be assumed that the output destination for the BIO, be that a terminal interface or a file supports UTF-8 encoding. I've checked OpenSSL 3+ and the behavior is the same there as well.

justsmth (dismissed):

samuel40791765 (dismissed):

justsmth (approved):

PR #2884, opened 2025-12-05
Route ML-DSA ACVP to the right APIs

samuel40791765:
### Description of changes: 
The previous ML-DSA change weren't routing against the right APIs. This fixes that.

Issues were mainly:
1. `extmu` is exclusive to `signatureInterface=internal`, but we were passing in false for `signatureInterface=external`.
2. `ml_dsa_*_sign` APIs are non-deterministic and we weren't passing in `rnd` as the random bytes. This routes the SignGen tests to the right places.

### Call-outs:
N/A

### Testing:
New ACVP demo vector, verified against expected outputs and added to this change. The new tests fail without the new changes made to ACVP.

### DIFF
--- a/util/fipstools/acvp/modulewrapper/modulewrapper.cc
+++ b/util/fipstools/acvp/modulewrapper/modulewrapper.cc
@@ -3430,7 +3430,10 @@ static bool ML_DSA_SIGGEN(const Span<const uint8_t> args[],
     }
   } else {
     // |context| is unique to signatureInterface: external.
-    // Prepare |pre| exactly how |ml_dsa_sign| is doing.
+    //
+    // Prepare |pre| exactly how |ml_dsa_sign| is doing. The maximum |context| size
+    // for ML-DSA is 255 bytes. We append a 0 and the size as two additional bytes
+    // before |context| to become the prefix string.
     uint8_t pre[257];
     pre[0] = 0;
     pre[1] = context.size();


skmcgrail (dismissed):

util/fipstools/acvp/modulewrapper/modulewrapper.cc
```diff
@@ -3434,6 +3428,17 @@ static bool ML_DSA_SIGGEN(const Span<const uint8_t> args[],
...
     }
+  } else {
+    // |context| is unique to signatureInterface: external.
+    // Prepare |pre| exactly how |ml_dsa_sign| is doing.
+    uint8_t pre[257];
+    pre[0] = 0;
+    pre[1] = context.size();
+    OPENSSL_memcpy(pre + 2 , context.data(), context.size());
+    if (!mldsa_funcs.sign_internal(sk.data(), signature.data(), &signature_len,
+                        msg.data(), msg.size(), pre, 2 + context.size(), rnd.data())) {
```
WillChilds-Klein, line 3442: suggest adding comments explaining magic constants here. the `+ 2` is fairly obvious given assignments to `pre`. stack size of `pre` at 257 is not obvious. where does this value come from?

also, we're sure that `context` is always a single byte? if so, L3436 can be `sizeof(uint8_t)` and save a runtime call.

samuel40791765: `context`s a pointer that could range from 0-255: https://pages.nist.gov/ACVP/draft-celi-acvp-ml-dsa.html.

I can add a comment explaining what the 257 size means here, but I mainly borrowed the same code from `ml_dsa_sign` here:https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/ml_dsa/ml_dsa_ref/sign.c#L330-L348

skmcgrail (approved):

PR #2885, opened 2025-12-06
Add sha1 CLI

nhatnghiho:
### Description of changes: 
Expose `openssl sha1` as a CLI command.

### Call-outs:
Relies on `dgstTool` so only minimal code change is needed.

### Testing:
Unit tests


tool-openssl/dgst_test.cc
```diff
@@ -246,6 +246,47 @@ TEST_F(DgstComparisonTest, SignAndVerify) {
   EXPECT_EQ(awslc_hash, openssl_hash);
 }
 
+class MD5ComparisonTest : public DgstComparisonTest {};
```
WillChilds-Klein, line 249: nice. thanks for adding coverage here :)

PR #2898, opened 2025-12-11
Several CLI Fixes

nhatnghiho:
### Description of changes: 
Modify a few of our CLI commands to match OpenSSL's behavior:
- `ecparam`:
   - Add secp256r1 as valid curve name
   - Enable write key when `-genkey` is provided despite `-noout`. Previously, writing is suppressed with `-noout`
- `x509`: Set serial number when signing
- `pkeyutl`: 
  - Add support for `-pkeyopt`
  - Accept `rsa_pss_saltlen:-1` as a valid pkeyopt value
- `req`: Suppress key write if `-key` is provided and no `-keyout` is specified

### Testing:
Unit tests

### DIFF
--- a/tool-openssl/req.cc
+++ b/tool-openssl/req.cc
@@ -919,7 +919,7 @@ bool reqTool(const args_list_t &args) {
   // 2. If no -config, output key to privkey.pem (this imitates how OpenSSL
   // would default to the default openssl.conf file, which has default_keyfile
   // set to privkey.pem)
-  if (keyout.empty()) {
+  if (keyout.empty() && key_file_path.empty()) {
     if (req_conf) {
       const char *default_keyfile = NCONF_get_string(
           req_conf.get(), req_section.c_str(), DEFAULT_KEYFILE);
--- a/tool-openssl/req_test.cc
+++ b/tool-openssl/req_test.cc
@@ -198,7 +198,8 @@ TEST_F(ReqTest, EncryptedPrivateKey) {
 }
 
 TEST_F(ReqTest, DefaultKeyoutPath) {
-  // Test that key is written to privkey.pem when -keyout is not specified
+  // Test that key is written to privkey.pem when -newkey is present even with
+  // no -keyout
   args_list_t args = {"-new", "-newkey", "rsa:2048", "-nodes",
                       "-out", csr_path,  "-subj",    "/CN=test.com"};
 
@@ -225,6 +226,30 @@ TEST_F(ReqTest, SuppressedKeyWrite) {
 
   args_list_t args = {"-new",   "-config", config_path,   "-out",
                       csr_path, "-subj",   "/CN=test.com"};
+
+  ASSERT_TRUE(reqTool(args));
+
+  // Verify that privkey.pem was NOT created
+  ScopedFILE f(fopen("privkey.pem", "r"));
+  EXPECT_FALSE(f) << "privkey.pem should not be created";
+}
+
+TEST_F(ReqTest, ExistingKeyNoWrite) {
+  // Verify if -key is provided, no key write should happen
+  ScopedFILE config_file(fopen(config_path, "w"));
+  ASSERT_TRUE(config_file);
+  fprintf(config_file.get(),
+          "[req]\n"
+          "default_keyfile = privkey.pem\n"
+          "distinguished_name = req_dn\n"
+          "encrypt_key = yes\n"
+          "[req_dn]\n"
+          "CN = Common Name\n");
+  fclose(config_file.release());
+
+  args_list_t args = {"-new",   "-config",      config_path,
+                      "-key",   input_key_path, "-out",
+                      csr_path, "-subj",        "/CN=primary"};
 
   ASSERT_TRUE(reqTool(args));
 


crypto/fipsmodule/evp/p_rsa.c
```diff
@@ -181,9 +182,7 @@ static int pkey_pss_init(EVP_PKEY_CTX *ctx) {
 // does not have the ability to handle ".data" sections. Splitting
 // |pkey_pss_init| into two functions: |pkey_pss_init_sign| and
 // |pkey_pss_init_verify|, gets around this undesired behaviour.
-static int pkey_pss_init_sign(EVP_PKEY_CTX *ctx) {
-  return pkey_pss_init(ctx);
-}
+static int pkey_pss_init_sign(EVP_PKEY_CTX *ctx) { return pkey_pss_init(ctx); }
```
justsmth: NP: I'm not sure how your clang-format is configured, but we generally avoid single-line functions.

crypto/fipsmodule/evp/p_rsa.c
```diff
@@ -734,12 +732,12 @@ static int pkey_rsa_ctrl_str(EVP_PKEY_CTX *ctx, const char *type,
     // "max" and "auto" are not supported
     long saltlen;
 
-    if (!strcmp(value, "digest")) {
+    if (!strcmp(value, "digest") || !strcmp(value, "-1")) {
```
justsmth: This will always be true, unless `value` can magically compare positively to both "digest" and "-1" simultaneously.

tool-openssl/ecparam.cc
```diff
@@ -2,69 +2,75 @@
...
-    bssl::UniquePtr<EC_GROUP> group;
-    if (nid != NID_undef) {
-        group.reset(EC_GROUP_new_by_curve_name(nid));
-    }
+  // We only allow 1 short name per NID but P-256 are synonymously addressible
+  // by secp256r1 and prime256v1
+  if (nid == NID_undef && isStringUpperCaseEqual(curve_name, "secp256r1")) {
+    nid = NID_X9_62_prime256v1;
+  }
+
```
justsmth: I think a better solution is to add it [here](https://github.com/aws/aws-lc/blob/main/crypto/obj/obj_dat.h#L9803) to our list of short-names:
```
❯ git diff
diff --git a/crypto/obj/obj_dat.h b/crypto/obj/obj_dat.h
index 6834ec795..d74f7ed0c 100644
--- a/crypto/obj/obj_dat.h
+++ b/crypto/obj/obj_dat.h
@@ -9800,6 +9800,7 @@ static const uint16_t kNIDsInShortNameOrder[] = {
     712 /* secp224k1 */,
     713 /* secp224r1 */,
     714 /* secp256k1 */,
+    415 /* secp256r1 */
     715 /* secp384r1 */,
     716 /* secp521r1 */,
     154 /* secretBag */,
```

With this, this name would resolve in other contexts as well.

crypto/fipsmodule/evp/p_rsa.c
```diff
@@ -734,7 +734,7 @@ static int pkey_rsa_ctrl_str(EVP_PKEY_CTX *ctx, const char *type,
...
```
justsmth, line 739: NP: A comment here about why "-1" is the same as `RSA_PSS_SALTLEN_DIGEST` would be helpful.

tool-openssl/ecparam.cc
```diff
@@ -2,69 +2,75 @@
...
+    nid = EC_curve_nist2nid(curve_name.c_str());
+  }
 
-    bssl::UniquePtr<EC_GROUP> group;
-    if (nid != NID_undef) {
-        group.reset(EC_GROUP_new_by_curve_name(nid));
-    }
+  // We only allow 1 short name per NID but P-256 are synonymously addressible
+  // by secp256r1 and prime256v1
+  if (nid == NID_undef && isStringUpperCaseEqual(curve_name, "secp256r1")) {
```
justsmth, line 22: It's unfortunate that we don't have a mechanism to have short-name aliases for NIDs.

skmcgrail: I mean we can? OpenSSL does, I think BoringSSL just omitted the duplicates from the file that generates the definitions: https://github.com/aws/aws-lc/blob/main/crypto/obj/objects.txt#L106-L107

From a generation perspective I don't think it changes anything or breaks anything. Would need a closer look.

tool-openssl/x509.cc
```diff
@@ -116,9 +116,91 @@ static int AdaptKeyIDExtension(X509 *cert, X509V3_CTX *ext_ctx,
...
+  BIGNUM *bn = NULL;
+
+  // 1. If no CA provided or serial file not found, generate a random serial
+  // number
+  // 2. Otherwise, read serial number from serial file
+  if (ca_file_path.empty() || !file) {
+    bn = BN_new();
+
+    // Randomly generate a serial number
+    constexpr int SERIAL_RAND_BITS = 159;
```
skmcgrail: This code comment from OpenSSL is useful to retain:
```
/*
 * IETF RFC 5280 says serial number must be <= 20 bytes. Use 159 bits
 * so that the first bit will never be one, so that the DER encoding
 * rules won't force a leading octet.
 */
```

tool-openssl/ecparam.cc
justsmth: I think we would need to make changes to objects.go to create another array for the aliases, or somehow integrate them into the current array for short-names.

nhatnghiho: I did try this out but unfortunately we can't have 2 short names for the same NID. Not sure if we have any mechanisms for aliases

justsmth (dismissed):

tool-openssl/pkeyutl.cc
```diff
@@ -112,7 +113,31 @@ static bool ReadInputData(const std::string &in_path,
   return true;
 }
 
+static bool ApplyPKeyParam(EVP_PKEY_CTX *ctx, const char *pkeyopt) {
+  char *stmp = OPENSSL_strdup(pkeyopt);
```
samuel40791765, line 117: Nit: we could wrap this in a uniqueptr

tool-openssl/pkeyutl_test.cc
```diff
@@ -147,7 +146,7 @@ TEST_F(PKeyUtlTest, PassinErrorHandlingTest) {
 }
 
 // Test that unprotected key works without passin
-TEST_F(PKeyUtlTest, NoPassinRequiredTest) {
+TEST_F(PKeyUtlTest, NoPassinRequired) {
```
samuel40791765, line 149: Love the simplification here, so much of our other tests in the CLI tool have the same redundant naming scheme

tool-openssl/req.cc
```diff
@@ -647,6 +647,13 @@ static bool GenerateSerial(X509 *cert) {
     return false;
   }
 
+  /*
+   * Randomly generate a serial number
+   *
+   * IETF RFC 5280 says serial number must be <= 20 bytes. Use 159 bits
+   * so that the first bit will never be one, so that the DER encoding
+   * rules won't force a leading octet.
+   */
```
samuel40791765, line 656:
```suggestion
  // Randomly generate a serial number
  //
  // IETF RFC 5280 says serial number must be <= 20 bytes. Use 159 bits
  // so that the first bit will never be one, so that the DER encoding
  // rules won't force a leading octet.
```

tool-openssl/x509.cc
```diff
@@ -116,9 +116,105 @@ static int AdaptKeyIDExtension(X509 *cert, X509V3_CTX *ext_ctx,
...
+
+  // If CA cert is provided, the serial file associated with the CA is
+  // <ca_file_name>.srl
+  if (!ca_file_path.empty()) {
+    size_t pos = ca_file_path.find_last_of('.');
+    serial_file_path = ca_file_path.substr(0, pos) + ".srl";
+    file.reset(fopen(serial_file_path.c_str(), "r"));  // Check if file exists
+  }
+
+  BIGNUM *bn = NULL;
```
samuel40791765, line 137:
```suggestion
  bssl::UniquePtr<BIGNUM> bn;
```

We should use UniquePtrs to maintain the memory here.

tool-openssl/x509_test.cc
```diff
@@ -1345,6 +1486,76 @@ TEST_F(X509ComparisonTest, X509ToolCompareExtensionsOpenSSL) {
...
+      ca_cert_path + " -CAkey " + ca_key_path + " -out " + out_path_openssl;
+
+  ExecuteCommand(tool_command);
+  ExecuteCommand(openssl_command);
+
+  auto cert_tool = LoadPEMCertificate(out_path_tool);
+  auto cert_openssl = LoadPEMCertificate(out_path_openssl);
+
+  ASSERT_TRUE(cert_tool != nullptr);
+  ASSERT_TRUE(cert_openssl != nullptr);
```
samuel40791765, line 1547:
```suggestion
  ASSERT_TRUE(cert_tool);
  ASSERT_TRUE(cert_openssl);
```

Nit: The additional comparison is a bit redundant

tool-openssl/x509.cc
```diff
@@ -116,9 +116,105 @@ static int AdaptKeyIDExtension(X509 *cert, X509V3_CTX *ext_ctx,
...
+      return false;
+    }
+
+    size_t len = OPENSSL_strnlen(buf, sizeof(buf));
+    if (len > 0 && buf[len - 1] == '\n') {
+      buf[len - 1] = '\0';
+      len--;
+    }
+
+#if defined(_WIN32)
```
samuel40791765, line 170:
```suggestion
#if defined(OPENSSL_WINDOWS)
```

Can we use this instead?

tool-openssl/x509.cc
```diff
@@ -116,9 +116,105 @@ static int AdaptKeyIDExtension(X509 *cert, X509V3_CTX *ext_ctx,
...
+  if (ca_file_path.empty() || !file) {
+    bn = BN_new();
+
+    /*
+     * Randomly generate a serial number
+     *
+     * IETF RFC 5280 says serial number must be <= 20 bytes. Use 159 bits
+     * so that the first bit will never be one, so that the DER encoding
+     * rules won't force a leading octet.
+     */
```
samuel40791765, line 151:
```suggestion
    // Randomly generate a serial number
    //
    // IETF RFC 5280 says serial number must be <= 20 bytes. Use 159 bits
    // so that the first bit will never be one, so that the DER encoding
    // rules won't force a leading octet.
```

tool-openssl/x509.cc
```diff
@@ -116,9 +116,105 @@ static int AdaptKeyIDExtension(X509 *cert, X509V3_CTX *ext_ctx,
...
+
+    bn = BN_new();
+
+    if (!BN_hex2bn(&bn, buf)) {
+      BN_free(bn);
+      return false;
+    }
+  }
+
+  if (!BN_add_word(bn, 1)) {
```
samuel40791765, line 185: Minor question: why do we need to add 1 here?

tool-openssl/req_test.cc
```diff
@@ -233,6 +234,30 @@ TEST_F(ReqTest, SuppressedKeyWrite) {
...
+  ScopedFILE config_file(fopen(config_path, "w"));
+  ASSERT_TRUE(config_file);
+  fprintf(config_file.get(),
+          "[req]\n"
+          "default_keyfile = privkey.pem\n"
+          "distinguished_name = req_dn\n"
+          "encrypt_key = yes\n"
+          "[req_dn]\n"
+          "CN = Common Name\n");
+  fclose(config_file.release());
```
samuel40791765, line 248: This is redundant, `ScopeFILE` already takes care of this for us.
https://github.com/aws/aws-lc/blob/ec39cb332cd0c01eab5ed8d17d282811f8f52a54/crypto/test/file_util.h#L33-L39

tool-openssl/ecparam_test.cc
```diff
@@ -200,101 +259,125 @@ TEST_F(EcparamOptionUsageErrorsTest, InvalidOutformTest) {
...
+          << "Skipping test: AWS_LC_TOOL_EXECUTABLE_PATH and/or "
+             "OPENSSL_EXECUTABLE_PATH environment variables are not set";
     }
     ASSERT_GT(createTempFILEpath(out_path_tool), 0u);
     ASSERT_GT(createTempFILEpath(out_path_openssl), 0u);
   }
 
   void TearDown() override {
-    if (strnlen(out_path_tool, PATH_MAX) > 0) {  // Only remove if path was created
+    if (strnlen(out_path_tool, PATH_MAX) >
```
samuel40791765, line 281: Not a priority for this PR, I understand this was only a minor adjustment in the format in this file.

But we should change this to use `OPENSSL_strnlen` for this and similar cases.

tool-openssl/x509.cc
nhatnghiho: To increment serial number (i.e., get a new number). Will add a comment

samuel40791765 (approved):

PR #2899, opened 2025-12-11
Support stdin for openssl rsa tool

samuel40791765:
### Issues:
Addresses `P327135637`

### Description of changes: 
There are users that expect the stdin format when using the `openssl rsa` tool. I also noticed a number of behavioral issues/differences while working on this and comparing it to Openssl's version.
1. OpenSSL prioritizes the PKCS#8 SubjectPublicKeyInfo format first, rather than the raw format. This is particularly problematic with `stdin`, since `stdin` doesn't have a FILE rewind mechanism. This PR aligns AWS-LC behavior with that.
2. ~~OpenSSL does not fallback to the raw format, but exposes it through a `RSAPublicKey_in` flag. Our fallback here isn't really an issue though, but I left a TODO to implement `RSAPublicKey_in`. We haven't noticed any actual consumption of such yet.~~ Implemented RSAPublicKey_in as a no-op.
3. ~~OpenSSL does not support reading the RSAPrivateKey format via the CLI. It's unclear whether this is actually needed in our code. I'm hesitant to remove it though, but I've left some notes to clean this up.~~ OpenSSL's PKCS8 parser falls back to PKCS1 if unsuccessful. We have to do so manually.

### Call-outs:
N/A

### Testing:
New RSA comparison test

### DIFF
--- a/tool-openssl/rsa.cc
+++ b/tool-openssl/rsa.cc
@@ -140,9 +140,8 @@ bool rsaTool(const args_list_t &args) {
       }
       if (!rsa && BIO_seek(in_file.get(), 0) == 0) {
         // Try raw RSAPublicKey format.
-        // TODO: This is behind the "RSAPublicKey_in" flag in OpenSSL and they don't
-        // use a fallback. We can keep this fallback, but we may want to support the
-        // flag as well.
+        // This is the "RSAPublicKey_in" flag in OpenSSL, but we support
+        // this behind a no-op flag and an automatic fallback.
         rsa.reset(d2i_RSAPublicKey_bio(in_file.get(), nullptr));
       }
     }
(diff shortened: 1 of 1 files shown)


justsmth (approved):

tool-openssl/rsa.cc
```diff
@@ -125,54 +126,53 @@ bool rsaTool(const args_list_t &args) {
...
-        // Try PKCS#8 SubjectPublicKeyInfo format
-        bssl::UniquePtr<EVP_PKEY> pkey(d2i_PUBKEY_bio(in_file.get(), nullptr));
-        if (pkey) {
-          rsa.reset(EVP_PKEY_get1_RSA(pkey.get()));
-        }
+        // Try raw RSAPublicKey format.
+        // TODO: This is behind the "RSAPublicKey_in" flag in OpenSSL and they don't
+        // use a fallback. We can keep this fallback, but we may want to support the
+        // flag as well.
+        rsa.reset(d2i_RSAPublicKey_bio(in_file.get(), nullptr));
```
skmcgrail, line 145: Would you just make the flag a no-op in the future then as to not remove the automatic fallback that is occurring here?

tool-openssl/rsa.cc
```diff
@@ -125,54 +126,53 @@ bool rsaTool(const args_list_t &args) {
...
+      if (pkey) {
+        rsa.reset(EVP_PKEY_get1_RSA(pkey.get()));
+      }
       if (!rsa && BIO_seek(in_file.get(), 0) == 0) {
-        bssl::UniquePtr<EVP_PKEY> pkey(PEM_read_bio_PrivateKey(in_file.get(), nullptr, nullptr, nullptr));
-        if (pkey) {
-          rsa.reset(EVP_PKEY_get1_RSA(pkey.get()));
-        }
+        // TODO: Can we remove this, OpenSSL doesn't really expect this format through the CLI?
+        rsa.reset(PEM_read_bio_RSAPrivateKey(in_file.get(), nullptr, nullptr, nullptr));
```
skmcgrail: If OpenSSL doesn't support it I think it's safer for us to do it either, but we can always reason about adding support for such fallback. The issue with including it now is that it is harder to reason about if people started to come to rely on the behavior not realizing the intent of the original interface.

skmcgrail (dismissed):

tool-openssl/rsa.cc
samuel40791765: I agree, just wasn't sure if there already consumers dependent on the current behavior, so wasn't comfortable removing it. Luckily, this hasn't existed for long and I think we can give it a shot.

tool-openssl/rsa.cc
samuel40791765: Yeah great idea. I could actually just add it as a no-op now.

justsmth (approved):

PR #2902, opened 2025-12-12
Import mldsa-native

jakemas:
### Issues:
Related PRs:
- Import-mldsa-native-NTT stress test on delocator [#2903](https://github.com/aws/aws-lc/pull/2903)
- AES-GCM: Add function pointer trampolines to avoid delocator issue [#2919](https://github.com/aws/aws-lc/pull/2919)
- Service Indicator: Add error call trampoline to avoid delocator issue [#2920](https://github.com/aws/aws-lc/pull/2920)

### Import mldsa-native

This imports mldsa-native (https://github.com/pq-code-package/mldsa-native) into AWS-LC.

This PR focuses on the minimal configuration of mldsa-native: No assembly and no FIPS-202 code are imported.

mldsa-native is a high-performance, high-assurance C90 implementation of ML-DSA developed under the Post-Quantum Cryptography Alliance (PQCA) and the Linux Foundation. It is a fork of the Dilithium reference implementation.

### Import Mechanism

The mldsa-native source code is unmodified and imported using the importer script `crypto/fipsmodule/ml_dsa/importer.sh;` the details of the import are in META.yml.

A custom config is provided for mldsa-native which in particular includes a small 'compatibility layer' between AWS-LC/OpenSSL and mldsa-native -- see below.

### Future imports (C-only)

Future updates of the C-only mldsa-native source tree should happen through a re-import of mldsa-native: That is, (a) delete `crypto/fipsmodule/ml_dsa/mldsa` and (b) re-run import.sh. This will re-import `mldsa-native/main`, though you can set the `GITHUB_SHA` and `GITHUB_REPOSITORY` environment variables to point to any other mldsa-native repository/fork.

### Future imports (native code)

Once we have verified meaningful parts of the mldsa-native assembly backends, PRs will be filed to integrate those. The details for this integration are TBD and not necessary to finalize for this PR. The options are (a) extending import.sh to import larger parts of the mldsa-native upstream source tree, including native backends, (b) writing custom backends, backed by sources living in the s2n-bignum source tree. Both is possible and compatible with this PR.

### Import Scope

mldsa-native has a C-only version as well as native 'backends' in AVX2 and Neon for high performance. This commit only imports the C-only version. Integration of native backends will be done separately.

mldsa-native offers its own FIPS-202 implementation, including fast versions of batched FIPS-202. However, this commit does not import those, but instead provides glue-code around AWS-LC's own FIPS-202 implementation. The path to leveraging the FIPS-202 performance improvements in mldsa-native would be to integrate them directly into `crypto/fipsmodule/sha`.

### Impact on build

None. No build-files are modified. The multilevel build process remains unchanged.

### Internal API changes
3 Removed functions:
```
  [D] 'function void ml_dsa_44_params_init(ml_dsa_params*)'    {ml_dsa_44_params_init}
  [D] 'function void ml_dsa_65_params_init(ml_dsa_params*)'    {ml_dsa_65_params_init}
  [D] 'function void ml_dsa_87_params_init(ml_dsa_params*)'    {ml_dsa_87_params_init}
```

### Compatibility layer

The configuration file `mldsa_native_config.h` includes a compatibility layer between AWS-LC/OpenSSL and mldsa-native, covering:

* FIPS/PCT: If AWSLC_FIPS is set, `MLD_CONFIG_KEYGEN_PCT` is enabled to include a PCT.
* FIPS/PCT: If `BORINGSSL_FIPS_BREAK_TESTS` is set, `MLD_CONFIG_KEYGEN_PCT_BREAKAGE_TEST` is set and `mld_break_pct` defined via `boringssl_fips_break_test("MLDSA_PWCT")`, to include runtime-breakage of the PCT for testing purposes.
* CT: If `BORINGSSL_CONSTANT_TIME_VALIDATION` is set, then `MLD_CONFIG_CT_TESTING_ENABLED` is set to enable valgrind testing.
* Zeroization: `MLD_CONFIG_CUSTOM_ZEROIZE` is set and `mld_zeroize` mapped to `OPENSSL_cleanse` to use OpenSSL's zeroization function.
* Randombytes: `MLD_CONFIG_CUSTOM_RANDOMBYTES` is set and `mld_randombytes` mapped to `RAND_bytes` to use AWS-LC's randombytes function.

### Side-channels

mldsa-native's CI uses a patched version of valgrind to check for various compilers and compile flags that there are no secret-dependent memory accesses, branches, or divisions. The relevant assertions have been kept but are unused unless `MLD_CONFIG_CT_TESTING_ENABLED` is set, which is the case if and only if `BORINGSSL_CONSTANT_TIME_VALIDATION` is set.

mldsa-native uses value barriers to block potentially harmful compiler reasoning and optimization. Where standard gcc/clang inline assembly is not available, mldsa-native falls back to a slower 'opt blocker' based on a volatile global -- both are described in ct.h.

### Formal Verification

All C-code imported in this commit is formally verified using the C Bounded Model Checker (CBMC) to be free of various classes of undefined behaviour, including out-of-bounds memory accesses and arithmetic overflow; the latter is of particular interest for ML-DSA because of the use of lazy modular reduction for improved performance.

The heart of the CBMC proofs are function contract and loop annotations to the C-code. Function contracts are denoted `__contract__(...)` clauses and occur at the time of declaration, while loop contracts are denoted `__loop__` and follow the for statement.

The function contract and loop statements are kept in the source, but removed by the preprocessor so long as the CBMC macro is undefined. Keeping them simplifies the import, and care has been taken to make them readable to the non-expert, and thereby serve as precise documentation of assumptions and guarantees upheld by the code.

### FIPS Compliance

mldsa-native unconditionally includes stack zeroization. mldsa-native's default secure memset is replaced by `OPENSSL_cleanse`.

mldsa-native conditionally includes a PCT, guarded by `MLD_CONFIG_KEYGEN_PCT`. This is set in the config if and only if `AWSLC_FIPS` is set.

While not part of the FIPS standard, the `pk_from_sk` function includes validation of both t0 (low-order bits) and tr (hash of public key) using constant-time comparison functions (`mld_ct_memcmp`), providing strong assurance of key consistency.

### Testing

We KAT ML-DSA with test vectors obtained from https://github.com/post-quantum-cryptography/KAT within `PQDSAParameterTest.KAT`. We select the KATs for the signing mode `hedged`, which derives the signing private random seed (rho) pseudorandomly from the signer's private key, the message to be signed, and a 256-bit string `rnd` which is generated at random. The `pure` variant of these KATs were used, as they provide test vector inputs for "pure" i.e., non-pre-hashed messages.

We also run the ACVP test vectors obtained from https://github.com/usnistgov/ACVP-Server within the three functions `PerMLDSATest.ACVPKeyGen`, `PerMLDSATest.ACVPSigGen` and `PerMLDSATest.ACVPSigVer`. These correspond to the tests found at ML-DSA-keyGen-FIPS204, ML-DSA-sigGen-FIPS204, and ML-DSA-sigVer-FIPS204.
To test ML-DSA pure, non-deterministic mode, we use `tgId = 19, 21, 23` of sigGen and `tgId = 7, 9, 11` of sigVer.
To test ML-DSA ExternalMu, non-deterministic mode, we use `tgId = 20, 22, 24` of sigGen and `tgId = 8, 10, 12` of sigVer.

**Test Results**:
- ML-DSA Tests: 100% passing (61/61 tests)

### Formatting

Code in `crypto/fipsmodule/ml_dsa/mldsa` is directly imported from mldsa-native and comes with its own `crypto/fipsmodule/ml_dsa/mldsa/.clang-format`.

### Prefix build

The prefix build should not be affected by the import, since no definitions of external linkage are imported (everything is tagged either static directly, or `MLD_EXTERNAL_API` or `MLD_INTERNAL_API`, both of which are set to static in the context of the import, too).

### Performance

Performance should be comparable to the previous integration as both are based on C-only code with AWS-LC's FIPS-202 implementation. The fast mldsa-native backends are not yet imported.

### Multilevel build

At the core, mldsa-native is currently a 'single-level' implementation of ML-DSA: A build of the main source tree provides an implementation of exactly one of ML-DSA-44/65/87, depending on the `MLD_CONFIG_PARAMETER_SET` parameter.

To build all security levels, level-specific sources are built 3 times, once per security level, and linked with a single build of the level-independent code. The single-compilation-unit approach pursued by AWS-LC makes this process fairly simple since one merely needs to include the single-compilation-unit file provided by mldsa-native three times, and configure it so that the level-independent code is included only once. The final include moreover #undef'ines all macros defined by mldsa-native, reducing the risk of name clashes with other parts of `crypto/fipsmodule/bcm.c`.

Note that this process is entirely internal to ml_dsa.c, and does not affect the AWS-LC build.

HashML-DSA: mldsa-native includes lots of HashML-DSA functionality that we dont need in aws-lc. Perhaps we should add config upstream to mldsa-native to choose which of pure/externalmu/hash modes are imported to reduce unused code.

### Main differences from reference implementation

mldsa-native is a fork of the ML-DSA reference implementation (Dilithium).

The following gives an overview of the major changes:

- CBMC and debug annotations, and minor code restructurings or signature changes to facilitate the CBMC proofs. For example, functions are structured to make loop bounds and memory access patterns explicit for formal verification.
- Introduction of 4x-batched versions of some functions from the reference implementation. This is to leverage 4x-batched Keccak-f1600 implementations if present. The batching happens at the C level even if no native backend for FIPS 202 is present.
- FIPS 204 compliance: Introduced optional PCT (FIPS 204, Section 4.4, Pairwise Consistency) and zeroization of stack buffers as required by (FIPS 204, Section 3.6.3, Destruction of intermediate values).
- Introduction of native backend implementations for AVX2. Those are drop-in replacements for the corresponding C functions and dispatched at compile-time. (Not in this PR, but the C code prep is in place).
- Restructuring of files to separate level-specific from level-generic functionality. This is needed to enable a multi-level build of mldsa-native where level-generic code is shared between levels.
- More pervasive use of value barriers to harden constant-time primitives, even when Link-Time-Optimization (LTO) is enabled. The use of LTO can lead to insecure compilation in case of the reference implementation.

### License
mldsa-native (everything under `crypto/fipsmodule/ml_dsa/mldsa/**`) is imported under the Apache 2.0 license and the ISC license. The LICENSE file remains unchanged.

Integration-specific code (everything with direct parent `crypto/fipsmodule/ml_dsa/*`) is made under the terms of the Apache 2.0 license and the ISC license.


jakemas:
Opened https://github.com/awslabs/aws-lc-verification/pull/180 for the SAW proof wrapper function changes.

crypto/fipsmodule/ml_dsa/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+The function contract and loop statements are kept in the source, but
+removed by the preprocessor so long as the CBMC macro is undefined. Keeping
+them simplifies the import, and care has been taken to make them readable
+to the non-expert, and thereby serve as precise documentation of
+assumptions and guarantees upheld by the code.
+
+## Testing
+
+We KAT ML-DSA with test vectors obtained from https://github.com/post-quantum-cryptography/KAT within `PQDSAParameterTest.KAT`. We select the KATs for the signing mode `hedged`, which derives the signing private random seed (rho) pseudorandomly from the signer's private key, the message to be signed, and a 256-bit string `rnd` which is generated at random. The `pure` variant of these KATs were used, as they provide test vector inputs for "pure" i.e., non-pre-hashed messages. The KAT files have been modified to insert linebreaks between each test vector set.
```
nebeid:
```suggestion
We test ML-DSA with Known Answer Test (KAT) vectors obtained from https://github.com/post-quantum-cryptography/KAT within `PQDSAParameterTest.KAT`. We select the KATs for the signing mode `hedged`, which derives the signing private random seed (rho) pseudorandomly from the signer's private key, the message to be signed, and a 256-bit string `rnd` which is generated at random. The `pure` variant of these KATs were used, as they provide test vector inputs for "pure" i.e., non-pre-hashed messages. The KAT files have been modified to insert linebreaks between each test vector set.
```

crypto/fipsmodule/ml_dsa/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+We also run the ACVP test vectors obtained from https://github.com/usnistgov/ACVP-Server within the three functions `PerMLDSATest.ACVPKeyGen`, `PerMLDSATest.ACVPSigGen` and `PerMLDSATest.ACVPSigVer`. These correspond to the tests found at [ML-DSA-keyGen-FIPS204](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-keyGen-FIPS204), [ML-DSA-sigGen-FIPS204](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-sigGen-FIPS204), and [ML-DSA-sigVer-FIPS204](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-sigVer-FIPS204).
+To test ML-DSA pure, non-deterministic mode, we use `tgId = 19, 21, 23` of sigGen and `tgId = 7, 9, 11` of sigVer.
+To test ML-DSA ExternalMu, non-deterministic mode, we use `tgId = 20, 22, 24` of sigGen and `tgId = 8, 10, 12` of sigVer.
+
+The test suite includes:
```
nebeid, line 104: Does it include testing the deterministic functions? Are these included in the KATs? The ACVP tests are not KATs? They test the non-deterministic version with random data?

crypto/fipsmodule/ml_dsa/README.md
```diff
@@ -0,0 +1,151 @@
...
+  changes to facilitate the CBMC proofs. For example, functions are structured
+  to make loop bounds and memory access patterns explicit for formal verification.
+- Introduction of 4x-batched versions of some functions from the reference
+  implementation. This is to leverage 4x-batched Keccak-f1600 implementations
+  if present. The batching happens at the C level even if no native backend
+  for FIPS 202 is present.
+- FIPS 204 compliance: Introduced optional PCT (FIPS 204, Section 4.4, Pairwise
+  Consistency) and zeroization of stack buffers as required by (FIPS 204, 
+  Section 3.6.3, Destruction of intermediate values).
+- Introduction of native backend implementations for AVX2. Those are drop-in
```
nebeid: Where is the introduction implemented?

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
```diff
@@ -0,0 +1,135 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#ifndef MLD_CONFIG_H
+#define MLD_CONFIG_H
+
+#if !defined(__ASSEMBLER__)
+#include "../../internal.h"
+
+// Define inline before mldsa-native headers are included
```
nebeid: I believe some of these definitions up until line 31 are in sys.h in the case of mlkem. Will they be put there when the assembly back end is imported? It doesn't seem to me they were moved in PR#2903.

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
```diff
@@ -0,0 +1,135 @@
...
+// AWS-LC's own FIPS 202 implementation.
+#define MLD_CONFIG_FIPS202_CUSTOM_HEADER "../fips202_glue.h"
+#define MLD_CONFIG_FIPS202X4_CUSTOM_HEADER "../fips202x4_glue.h"
+
+// Everything is built in a single CU, so both internal and external
+// mldsa-native API can have internal linkage.
+// Mark as unused to suppress warnings for functions we don't expose
+#if defined(__GNUC__) || defined(__clang__)
+#define MLD_CONFIG_INTERNAL_API_QUALIFIER static __attribute__((unused))
+#define MLD_CONFIG_EXTERNAL_API_QUALIFIER static __attribute__((unused))
```
nebeid, line 26: Same question

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
```diff
@@ -0,0 +1,135 @@
...
+#endif
+
+// Map the CPU capability function to the ones used by AWS-LC
+#define MLD_CONFIG_CUSTOM_CAPABILITY_FUNC
+#if !defined(__ASSEMBLER__) && !defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)
+#include <stdint.h>
+static MLD_INLINE int mld_sys_check_capability(int cap)
+{
+#if defined(MLD_SYS_X86_64)
+  if (cap == 1) // MLD_SYS_CAP_AVX2
```
nebeid: Related to the previous question, but also in PR#2903, it doesn't seem like the macro was used.

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
```diff
@@ -0,0 +1,135 @@
...
+#if !defined(__ASSEMBLER__) && !defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)
+#include <stdint.h>
+static MLD_INLINE void *mld_memcpy(void *dest, const void *src, size_t n) {
+    return OPENSSL_memcpy(dest, src, n);
+}
+#endif // !__ASSEMBLER__
+
+// Map memset function to the one used by AWS-LC
+#define MLD_CONFIG_CUSTOM_MEMSET
+#if !defined(__ASSEMBLER__) && !defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)
```
nebeid:
```suggestion
#if !defined(__ASSEMBLER__) 
```
as in mlkem_native_config.h

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
```diff
@@ -0,0 +1,135 @@
...
+#include <stdint.h>
+#include <openssl/rand.h>
+static MLD_INLINE void mld_randombytes(uint8_t *ptr, size_t len) {
+    RAND_bytes(ptr, len);
+}
+#endif // !__ASSEMBLER__
+
+// Map memcpy function to the one used by AWS-LC
+#define MLD_CONFIG_CUSTOM_MEMCPY
+#if !defined(__ASSEMBLER__) && !defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)
```
nebeid:
```suggestion
#if !defined(__ASSEMBLER__) 
```
as in mlkem_native_config.h

crypto/fipsmodule/ml_dsa/README.md
jakemas: This is referring to the deterministic variant of ML-DSA (as opposed to deterministically testing the APIs). We do not test the deterministic mode as we do not support the ML-DSA deterministic mode. Internally the API is available (we note this here: https://github.com/aws/aws-lc/blob/d7cd042f5f2150eaa3c7034be77a313918c232bb/crypto/fipsmodule/ml_dsa/mldsa/sign.c#L709-L710 but it isn't exposed as an external API anywhere.)

crypto/fipsmodule/ml_dsa/README.md
jakemas: Thank you, added in [afa0b3b](https://github.com/aws/aws-lc/pull/2902/commits/afa0b3b6c7801395e92d21454ec3525e4c8cb9dc)

crypto/fipsmodule/ml_dsa/README.md
jakemas: Good catch -- not added in this PR now. Removed this line in [afa0b3b](https://github.com/aws/aws-lc/pull/2902/commits/afa0b3b6c7801395e92d21454ec3525e4c8cb9dc)

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
jakemas: You are correct, they are already in `sys.h` https://github.com/aws/aws-lc/blob/79c9a4cbb4391232d075cdb6e0bd64baa0549fe9/crypto/fipsmodule/ml_dsa/mldsa/sys.h#L118-L136

I removed the duplicate definitions in [79c9a4c](https://github.com/aws/aws-lc/pull/2902/commits/79c9a4cbb4391232d075cdb6e0bd64baa0549fe9)

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
jakemas: I've aligned [crypto/fipsmodule/ml_dsa/mldsa_native_config.h](https://github.com/aws/aws-lc/pull/2902/files/fca7864e6a1b896f475371f9de89b77964bdc665#diff-4e926b4a923cb20183a2d3c5d510dff9f49e09e3c869ea9080667d22c4420bd0) to look more like `crypto/fipsmodule/ml_kem/mlkem_native_config.h` in [79c9a4c](https://github.com/aws/aws-lc/pull/2902/commits/79c9a4cbb4391232d075cdb6e0bd64baa0549fe9) and removed

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
jakemas: This was added to prevent errors regarding the unused APIs -- unlike ML-KEM there are some additional API options for ML-DSA. Many of the internals are shared, so there isn't much redundant code, but I have opened an issue upstream https://github.com/pq-code-package/mldsa-native/issues/740 to add to the import options to be more specific about exact APIs that are included.
```
aws-lc/crypto/fipsmodule/ml_dsa/mldsa/sign.c:1053:5: error: unused function 'mldsa87_verify_pre_hash_shake256' [-Werror,-Wunused-function]
```

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
jakemas: I've aligned [crypto/fipsmodule/ml_dsa/mldsa_native_config.h](https://github.com/aws/aws-lc/pull/2902/files/fca7864e6a1b896f475371f9de89b77964bdc665#diff-4e926b4a923cb20183a2d3c5d510dff9f49e09e3c869ea9080667d22c4420bd0) to look more like crypto/fipsmodule/ml_kem/mlkem_native_config.h in [79c9a4c](https://github.com/aws/aws-lc/pull/2902/commits/79c9a4cbb4391232d075cdb6e0bd64baa0549fe9) and removed.

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
jakemas: I've aligned [crypto/fipsmodule/ml_dsa/mldsa_native_config.h](https://github.com/aws/aws-lc/pull/2902/files/fca7864e6a1b896f475371f9de89b77964bdc665#diff-4e926b4a923cb20183a2d3c5d510dff9f49e09e3c869ea9080667d22c4420bd0) to look more like crypto/fipsmodule/ml_kem/mlkem_native_config.h in [79c9a4c](https://github.com/aws/aws-lc/pull/2902/commits/79c9a4cbb4391232d075cdb6e0bd64baa0549fe9) and removed.

util/fipstools/delocate/delocate.peg
```diff
@@ -52,7 +52,7 @@ SymbolArgs <- SymbolArg ((WS? ',' WS?) SymbolArg)*
 SymbolArg <- SymbolExpr
 SymbolExpr <- SymbolAtom (WS? SymbolOperator WS? SymbolExpr)?
 SymbolAtom <- Offset / SymbolType / LocalLabelRef / LocalSymbol TCMarker? / SymbolName Offset / SymbolName TCMarker? / Dot / OpenParen WS? SymbolExpr WS? CloseParen
-SymbolOperator <- '+' / '-' / '|' / '<<' / '>>'
+SymbolOperator <- '+' / '-' / '|' / '<<' / '>>' / '/'
```
jakemas, line 55: I had to add `/` as a new symbol operator due to the following code generating the use of one:

```
sed -n '185810,185815p' test_build_dir/crypto/fipsmodule/libbcm_c_generated_asm.a                                                                                                                                 
.Lrtx8021:
        .section        .rodata
        .align  0
        .align  2
.L8021:
        .word   (.L8032 - .Lrtx8021) / 4
```

The error was:
```
error while parsing "/codebuild/output/src482392576/src/actions-runner/_work/aws-lc/aws-lc/test_build_dir/crypto/fipsmodule/libbcm_c_generated_asm.a":
parse error near WS (line 185997 symbol 28 - line 185997 symbol 29): 
" "
```

crypto/fipsmodule/ml_dsa/README.md
nebeid: > This is referring to the deterministic variant of ML-DSA

Do you mean "non-deterministic"? If so, I'd like to know how it's tested in ACVP, with KATs or random data making sure it verifies the signature? If KATs, how are they set up to bypass randomness generation?

crypto/fipsmodule/ml_dsa/fips202_glue.h
```diff
@@ -0,0 +1,133 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#ifndef MLD_AWSLC_FIPS202_GLUE_H
+#define MLD_AWSLC_FIPS202_GLUE_H
+#include <stddef.h>
+#include <stdint.h>
+
+#include "../sha/internal.h"
+
+// Define MLD_INLINE if not already defined
```
dkostic: looks like the MLKEM counterpart is defined in `sys.h`. can we do the same here?

crypto/fipsmodule/ml_dsa/mldsa/common.h
```diff
@@ -0,0 +1,172 @@
+/*
+ * Copyright (c) The mldsa-native project authors
+ * Copyright (c) The mlkem-native project authors
+ * SPDX-License-Identifier: Apache-2.0 OR ISC OR MIT
+ */
+#ifndef MLD_COMMON_H
+#define MLD_COMMON_H
+
+#define MLD_BUILD_INTERNAL
```
dkostic, line 9: what's this used for?

crypto/fipsmodule/ml_dsa/mldsa/ct.c
```diff
@@ -0,0 +1,21 @@
+/*
```
dkostic, line 1: I don't think `ct.h` and `ct.c` are used? if that's the case let's skip importing them.

jakemas: There are uses:
- `mld_ct_sel_int32` of `ct.h` is used in `reduce.h` and `rounding.h`m
- `mld_ct_memcmp` of `ct.h` in `sign.c`, 
- `mld_zeroize` of `ct.h` is used in `poly_kl.c`, `poly.c`, `polyvec.`, `sign.c`
- `mld_ct_cmask_neg_i32`of `ct.h` is used in `poly.c`, `reduce.h`, `rounding.h`
- `mld_ct_abs_i32` of `ct.h` is used in `poly.c`

crypto/fipsmodule/ml_dsa/mldsa/common.h
jakemas: `MLD_BUILD_INTERNAL` is a marker macro that indicates the code is being compiled as part of the internal implementation build, as opposed to being used as an external library.

mlkem-native now has the same flag added see https://github.com/pq-code-package/mlkem-native/commit/62db0a5ba95fcc971780b89e8e739d6ba93d9cfd added in https://github.com/pq-code-package/mlkem-native/commit/62db0a5ba95fcc971780b89e8e739d6ba93d9cfd

crypto/fipsmodule/ml_dsa/fips202_glue.h
jakemas: ok, removed in [0bda0f1](https://github.com/aws/aws-lc/pull/2902/commits/0bda0f12cbc5aa58004b6920df42e2ed990d6c47) to now mimic mlkem-native

crypto/fipsmodule/ml_dsa/mldsa/ct.c
dkostic: ack, I think Github was loading properly yesterday so the cmd+f for `ct.h` didn't work. Anyway, I'm not sure is this supposed to work in AWS-LC? isn't everything in a single compilation unit?

crypto/fipsmodule/ml_dsa/README.md
jakemas: There are two versions of ml-dsa, deterministic and non-deterministic, we only implement the non-deterministic variant. To test the non-deterministic variant we do it two ways, one using the ACVP test vectors, and one using the KATs. In both cases, we use the internal APIs which allow the tester to provide the randomness for each signature via the provided `seed` argument. So there is no need to bypass the random generation, we use the provided seeds and supply them directly the the API to get the expected known result.

crypto/fipsmodule/ml_dsa/mldsa/ct.c
jakemas: Ok!
>Anyway, I'm not sure is this supposed to work in AWS-LC? isn't everything in a single compilation unit?

What do you mean by this? The SCU is autogenerated just like with mlkem-native and `ct.c` is included within it here https://github.com/aws/aws-lc/blob/b42fd206f66fb16aeb93473545cb07f297317229/crypto/fipsmodule/ml_dsa/mldsa/mldsa_native_bcm.c#L62

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
```diff
@@ -60,24 +36,25 @@
...
 #endif
-  (void)cap;
   return 0;
 }
 #endif
 
 #if defined(BORINGSSL_FIPS_BREAK_TESTS)
 #define MLD_CONFIG_KEYGEN_PCT_BREAKAGE_TEST
-#if !defined(__ASSEMBLER__) && !defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)
+#if !defined(__ASSEMBLER__)
```
nebeid: I'm a bit confused about `!defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)`. It's sometimes included and sometimes not in `crypto/fipsmodule/ml_kem/mlkem_native_config.h`. I understand its purpose in multilevel build, but here, I don't know why it's sometimes included and sometimes not. Can you explain when it needed to be included and when not in `mlkem_native_config.h`

crypto/fipsmodule/ml_dsa/mldsa/ct.c
dkostic: this mitigation works only when `volatile uint64_t mld_ct_opt_blocker_u64` is compiled in a separate compilation unit from the one it's used in. But in our case we are adding `ct.c` to `mldsa_native_bcm.c` together with the other files.

crypto/fipsmodule/ml_dsa/mldsa/sys.h
```diff
@@ -0,0 +1,246 @@
...
+/*
+ * C90 does not have the inline compiler directive yet.
+ * We don't use it in C90 builds.
+ * However, in that case the compiler warns about some inline functions in
+ * header files not being used in every compilation unit that includes that
+ * header. To work around it we silence that warning in that case using
+ * __attribute__((unused)).
+ */
+
+/* Do not use inline for C90 builds*/
```
nebeid: I think you need the latest version of mldsa-native, because I see a difference in the definitions of MLD_INLINE

jakemas: ok, ran the import script again to pick up newest changes upstream: [d37774e](https://github.com/aws/aws-lc/pull/2902/commits/d37774e25b4eb00455399f645a61144805807a5d)

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
jakemas: The `!defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)` guard is needed for parameter-set-specific functions but not for generic utility functions.

The guard is needed for `mlk_break_pct`, `mlk_zeroize`, and `mlk_randombytes` because these functions could potentially have parameter-set-specific behavior. In multilevel builds, without the guard, these would be defined multiple times (once per parameter set like MLKEM512/768/1024), which would cause "duplicate symbol" linker errors. The guard ensures they're only defined once in the shared portion of the multilevel build.

The guard is not needed for `mlk_memcpy` and `mlk_memset` because these are generic utility functions with identical implementations across all parameter sets. They don't depend on parameter set constants (K, L, ETA, etc.), so they can safely be defined in every compilation unit without conflicts. A simpler guard (`!defined(__ASSEMBLER__)` only) is sufficient for these functions.

In summary, use the guard for functions that could vary by parameter set or would create duplicate definitions in multilevel builds. We don't use it for truly generic utilities that are identical across all parameter sets.

crypto/fipsmodule/ml_dsa/mldsa/ct.c
jakemas: ~~Ok I see, so we could potentially exclude ct.c from the import, since it's essentially a no-op in our build configuration.~~

~~This is something I'd have to add as a config option upstream in mldsa-native. I will create an issue for it. For the scope of this PR, being the import code and a first pass at import, I feel there's no harm in keeping it for consistency with upstream, until I can make changes and then re-import later.~~

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
nebeid: Thank you for the satisfactory explanation. However, I am not sure how `mld_zeroize`, `mld_break_pct` and `mld_randombytes` are parameterised. This can wait for a subsequent PR but is to be checked in the case of mlkem as well.

crypto/fipsmodule/ml_dsa/mldsa/ct.c
jakemas: Offline discussion summarized:

Running through a comparison here with ML-KEM:

**mlkem-native**
Header: `mlkem/verify.h` - defines constant-time functions and value barriers
Source: `mlkem/verify.c `- contains `volatile uint64_t mlk_ct_opt_blocker_u64 = 0` (but effectively empty with `MLK_USE_ASM_VALUE_BARRIER`)


**mldsa-native**
Header: `mldsa/ct.h` - defines constant-time functions and value barriers
Source: `mldsa/ct.c` - contains `volatile uint64_t mld_ct_opt_blocker_u64 = 0` (but effectively empty with `MLD_USE_ASM_VALUE_BARRIER`)

The definition chain is then:

1. In `mldsa_native_config.h`: We don't set `MLD_CONFIG_NO_ASM` (only set if `OPENSSL_NO_ASM` is defined)
2. In `sys.h`: Since `MLD_CONFIG_NO_ASM` is not set:
```
#if !defined(MLD_CONFIG_NO_ASM) && (defined(__GNUC__) || defined(__clang__))
#define MLD_HAVE_INLINE_ASM
#endif
```
3. In `ct.h`: Since `MLD_HAVE_INLINE_ASM` is defined:
```
#if defined(MLD_HAVE_INLINE_ASM) && !defined(MLD_CONFIG_NO_ASM_VALUE_BARRIER)
#define MLD_USE_ASM_VALUE_BARRIER
#endif
```

i.e. both ct.h and ct.c are used. This is also how it works for mlkem-native.

dkostic: thanks Jake, I think I understand now!

nebeid (approved):

dkostic (approved):

crypto/fipsmodule/ml_dsa/mldsa/common.h
hanno-becker: @dkostic There has been a change in how the public headers (`mlkem_native.h` and `mldsa_native.h`) for mlkem-native and mldsa-native work.

Some config options affect only the header, not the implementation (e.g. `MLK_CONFIG_CONSTANTS_ONLY`); some config options affect only the implementation, not the header (e.g. `MLK_CONFIG_KEYGEN_PCT`).

Previously, the header used dedicated `MLK_CONFIG_API_XXX` options for configuration; the user had to set those directly, in alignment with `MLK_CONFIG_XXX`, before importing `mlkem_native.h`. This avoided the `mlkem_native.h` header being dependent on the full config, but we received feedback that this is confusing and cumbersome to use.

We have since changed the header so it, by default, consumes the full config file `mlkem_native_config.h`; there is no difference between header-config and build-config anymore.

However, as mentioned, some config options do not matter for the public header. In the header, those are behind a `MLK_BUILD_INTERNAL` guard. This guard is set for the implementation (in `common.h`, as you see above), but need not be set for consuming libraries/applications importing the header. They may, of course, but this pulls in things like the custom memset/zeroize/randombytes, which does not belong in the public header and may interfere for some reason with the consuming library.

The story is the same for mldsa-native.

Since AWS-LC does not consume the public header, IIRC, this is irrelevant for the present PR.

crypto/fipsmodule/ml_dsa/mldsa_native_config.h
hanno-becker: > The !defined(MLD_CONFIG_MULTILEVEL_NO_SHARED) guard is needed for parameter-set-specific functions but not for generic utility functions.

It's the opposite, see the documentation:
```
 *              If this is set, no MLK_CONFIG_PARAMETER_SET-independent code
 *              will be included in the build.
```

So, where the implementation uses level-independent code, that's guarded by `!defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)`.

The config is a special case, because that's fixes across the 'stages' of a monobuild: If you build 512, then 768, then 1024 -- as we do -- then the config will be included once at the beginning, and not again thereafter. The assumption is that you have a consistent configuration, but vary _only_ the level.

That being so, it doesn't matter if you add `!defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)` to the config or not; the code will only be used once in a context where `!defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)` holds anyway.

I agree this is confusing, and am open to ideas for how we can improve it.

For now, I would suggest that AWS-LC's config does consistently _not_ use `!defined(MLD_CONFIG_MULTILEVEL_NO_SHARED)`. We should also change this for the mlkem-native config used by AWS-LC.

crypto/fipsmodule/ml_dsa/fips202_glue.h
fadalialbaricar-dev: 🙏

fadalialbaricar-dev: 🙏

PR #2906, opened 2025-12-15
Iterate through all DNS entries in connect CLI

geedo0:
### Issues:
N/A

### Description of changes: 
The `openssl sclient -connect` CLI command would only attempt to connect to the first resolved DNS entry. Failing that,
it considers the entire connection attempt failed. This change updates the command to iterate through the linked list of
entries until it can successfully connect. This is in-line with upstream [OpenSSL](https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/apps/s_socket.c#L93).

It's particularly useful in cases where the first address resolved is either unavailable or otherwise unusable. This can
be common when trying to resolve an address such as `localhost` and the OS returns the ipv6 loopback address before the
ipv4 version and your web server is only listening with ipv4.

### Call-outs:
I did not exhaustively consider other uses of this function beyond this specific CLI tool. A glance check of the references suggests low usage and it's still the right change to make overall.

### Testing:
- Tested this in a build container which resolves `localhost` as `[::1, 127.0.0.1]` with a web server that listens on ipv4 and asserted that the CLI tool can connect successfully to that web server.

- CI

### DIFF
--- a/tool/transport_common.cc
+++ b/tool/transport_common.cc
@@ -165,43 +165,45 @@ bool Connect(int *out_sock, const std::string &hostname_and_port, bool quiet) {
 
   bool ok = false;
   char buf[256];
+  *out_sock = -1;
 
-  *out_sock =
-      socket(result->ai_family, result->ai_socktype, result->ai_protocol);
-  if (*out_sock < 0) {
-    PrintSocketError("socket");
-    goto out;
-  }
+  for (struct addrinfo *ai = result; ai != NULL; ai = ai->ai_next) {
+    *out_sock = socket(ai->ai_family, ai->ai_socktype, ai->ai_protocol);
+    if (*out_sock < 0) {
+      PrintSocketError("socket");
+      continue;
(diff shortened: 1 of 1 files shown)


tool/transport_common.cc
```diff
@@ -165,43 +165,45 @@ bool Connect(int *out_sock, const std::string &hostname_and_port, bool quiet) {
...
+      break;
+    } else {
+      PrintSocketError("connect");
+      closesocket(*out_sock);
+    }
   }
-  ok = true;
 
-out:
   freeaddrinfo(result);
```
WillChilds-Klein, line 207: do we also need to `free` subsequent `struct addrinfo`s as we iterate over `->ai_next`?

geedo0: Good question, the man page says that we're good.

> The freeaddrinfo() function frees the memory that was allocated for the dynamically allocated linked list res.

PR #2907, opened 2025-12-15
Add openssl genpkey cli utility tool

samuel40791765:
### Issues:
Resolves `P342458515`

### Description of changes: 
Internal consumers are depending on the `openssl genpkey` cli tool, so this adds the necessary functionality to support that. This is mainly used for generating a private key or key pair. 
I've only added the necessary options for now which are `-algorithm`, `-pkeyopt`, and `-out`. We can add more if needed, but there haven't been more use cases discovered yet. I've also reused and cleaned up a bit of the surrounding code while I was at it. Would like to refactor more, but that would be a separate PR.

### Call-outs:
N/A

### Testing:
new unit tests. I haven't added any comparison tests, since it didn't really make sense to compare generated private keys from two different libraries.

### DIFF
--- a/tool-openssl/genpkey.cc
+++ b/tool-openssl/genpkey.cc
@@ -18,8 +18,6 @@ static const argument_t kArguments[] = {
     {"", kOptionalArgument, ""}};
 
 static EVP_PKEY_CTX *init_gen_str(std::string algorithm) {
-  bssl::UniquePtr<EVP_PKEY_CTX> ctx;
-  int pkey_id = 0;
   const EVP_PKEY_ASN1_METHOD *ameth =
       EVP_PKEY_asn1_find_str(nullptr, algorithm.c_str(), algorithm.size());
   if (!ameth) {
@@ -27,10 +25,11 @@ static EVP_PKEY_CTX *init_gen_str(std::string algorithm) {
     return nullptr;
   }
 
-  ERR_clear_error();
-
-  EVP_PKEY_asn1_get0_info(&pkey_id, nullptr, nullptr, nullptr, nullptr, ameth);
-  ctx.reset(EVP_PKEY_CTX_new_id(pkey_id, nullptr));
+  int pkey_id = 0;
+  if (!EVP_PKEY_asn1_get0_info(&pkey_id, nullptr, nullptr, nullptr, nullptr, ameth)) {
+    return nullptr;
+  };
+  bssl::UniquePtr<EVP_PKEY_CTX> ctx(EVP_PKEY_CTX_new_id(pkey_id, nullptr));
   if (!ctx) {
     return nullptr;
   }
@@ -72,7 +71,7 @@ bool genpkeyTool(const args_list_t &args) {
     out.reset(BIO_new_fp(stdout, BIO_NOCLOSE));
   } else {
     out.reset(BIO_new(BIO_s_file()));
-    if (BIO_write_filename(out.get(), out_path.c_str()) <= 0) {
+    if (!BIO_write_filename(out.get(), out_path.c_str())) {
       fprintf(stderr, "Error: failed to open output file '%s'\n",
               out_path.c_str());
       return false;
@@ -90,7 +89,7 @@ bool genpkeyTool(const args_list_t &args) {
   }
 
   EVP_PKEY *tmp_key = nullptr;
-  if (EVP_PKEY_keygen(ctx.get(), &tmp_key) <= 0) {
+  if (!EVP_PKEY_keygen(ctx.get(), &tmp_key)) {
     fprintf(stderr, "Error generating key\n");
     return false;
   }
--- a/tool-openssl/internal.h
+++ b/tool-openssl/internal.h
@@ -135,11 +135,12 @@ BUCKET **get_table();
 void cleanup_hash_table();
 
 // ApplyPkeyCtrlString parses the options in |pkeyopt| and passes them to
-// |EVP_PKEY_CTX_ctrl_str|.
+// |EVP_PKEY_CTX_ctrl_str|. It returns false if the parsing or memory allocation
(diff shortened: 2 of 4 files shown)


tool-openssl/internal.h
```diff
@@ -133,6 +134,14 @@ void add_entry(enum Type type, uint32_t hash, const char *filename,
 BUCKET **get_table();
 void cleanup_hash_table();
 
+// ApplyPkeyCtrlString parses the options in |pkeyopt| and passes them to
+// |EVP_PKEY_CTX_ctrl_str|.
+bool ApplyPkeyCtrlString(EVP_PKEY_CTX *ctx, const char *pkeyopt);
+
+// WritePrivateKey writes the private key contents of |pkey| to |out| based on
+// |format|.
+bool WritePrivateKey(EVP_PKEY *pkey, bssl::UniquePtr<BIO> &out, int format);
```
WillChilds-Klein, line 144: nit: document failure modes on `false` return

tool-openssl/tool_util.cc
```diff
@@ -23,3 +26,24 @@ void SetUmaskForPrivateKey() {
...
+  if (!stmp) {
+    return false;
+  }
+
+  char *vtmp = strchr(stmp.get(), ':');
+  if (!vtmp) {
+    return false;
+  }
+
+  *vtmp = 0;
```
WillChilds-Klein: nit: clarity

```suggestion
  *vtmp = nullptr;
```

tool-openssl/genpkey.cc
```diff
@@ -0,0 +1,104 @@
...
+  bssl::UniquePtr<EVP_PKEY_CTX> ctx;
+  int pkey_id = 0;
+  const EVP_PKEY_ASN1_METHOD *ameth =
+      EVP_PKEY_asn1_find_str(nullptr, algorithm.c_str(), algorithm.size());
+  if (!ameth) {
+    fprintf(stderr, "Algorithm %s not found\n", algorithm.c_str());
+    return nullptr;
+  }
+
+  ERR_clear_error();
```
WillChilds-Klein: why do we need to clear errors here if no errors were encountered?

tool-openssl/genpkey.cc
```diff
@@ -0,0 +1,104 @@
...
+    {"-help", kBooleanArgument, "Display this summary"},
+    {"-out", kOptionalArgument, "Output file to write the key to"},
+    {"-algorithm", kRequiredArgument, "Public key algorithm to use."},
+    {"-pkeyopt", kOptionalArgument,
+     "Value to set the public key algorithm option to"},
+    {"", kOptionalArgument, ""}};
+
+static EVP_PKEY_CTX *init_gen_str(std::string algorithm) {
+  bssl::UniquePtr<EVP_PKEY_CTX> ctx;
+  int pkey_id = 0;
```
WillChilds-Klein: minor nit: declare next to first usage

tool-openssl/pkey.cc
```diff
@@ -49,17 +49,14 @@ static bool WritePublicKey(EVP_PKEY *pkey, bssl::UniquePtr<BIO> &output_bio, int
       return false;
     }
     return true;
```
WillChilds-Klein: nit: pull this and L58 out to L60

tool-openssl/tool_util.cc
```diff
@@ -23,3 +26,24 @@ void SetUmaskForPrivateKey() {
...
 #endif
 }
+
+bool ApplyPkeyCtrlString(EVP_PKEY_CTX *ctx, const char *pkeyopt) {
+  bssl::UniquePtr<char> stmp(OPENSSL_strdup(pkeyopt));
+  if (!stmp) {
+    return false;
+  }
+
+  char *vtmp = strchr(stmp.get(), ':');
```
WillChilds-Klein, line 37: for my understanding, what's the purpose of `:` here? is it not a delimiter?

suggest documenting this character's semantic in a source comment.

tool-openssl/genpkey.cc
```diff
@@ -0,0 +1,104 @@
...
+    fprintf(stderr, "Error setting %s parameter:\n", pkey_options.c_str());
+    return false;
+  }
+
+  EVP_PKEY *tmp_key = nullptr;
+  if (EVP_PKEY_keygen(ctx.get(), &tmp_key) <= 0) {
+    fprintf(stderr, "Error generating key\n");
+    return false;
+  }
+  pkey.reset(tmp_key);
```
WillChilds-Klein, line 96: does `EVP_PKEY_keygen`'s API contract ensure that as long as it returns >0, then `tmp_key` is non-null?

tool-openssl/tool_util.cc
samuel40791765: `nullptr` isn't allowed to be assigned to `char`, so I've used `'\0'` here.

tool-openssl/genpkey.cc
samuel40791765: My bad, this was remnant from code I took from other files. There are a lot of things I want to clean up in the other cli tool code though.

Also forgot a return value check below, will add.

tool-openssl/genpkey.cc
samuel40791765: Cleaned this up along with `ctx` above.

tool-openssl/genpkey.cc
samuel40791765: yes, it only allocates memory to it if the call was successful. But after reading the documentation, I believe it only return 1 or 0. I've fixed up the call here.

https://github.com/aws/aws-lc/blob/5a08a986819c4e871be686fdd0c47eb136e5c8fe/include/openssl/evp.h#L792-L796

WillChilds-Klein (approved):

PR #2908, opened 2025-12-16
Initial Framework for Using Doxygen to Document Public Header Files

m271828:
### Issues:
Resolves P276033715
Resolves P276038464
Addresses P276034041

### Description of changes: 
Start of moving AWS-LC over to Doxygen for a documentation framework and automatic publication of documentation with each update to main. This creates the initial configuration files, updates the style guide and includes one converted file as an example.

### Call-outs:
The documentation itself is not being updated as part of this work, it is being moved to the Doxygen framework. Updating documentation is future work. Doxygen layout has been customized since this is not a C++ project.

### Testing:
Doxygen configuration files and symbol linking have been verified correct with local testing, not yet at the stage for testing via GitHub.

### DIFF
--- a/include/openssl/base.h.in
+++ b/include/openssl/base.h.in
@@ -54,7 +54,10 @@
 #define OPENSSL_HEADER_BASE_H
 
 
-// This file should be the first included by all BoringSSL headers.
+/**
+ * @file
+ * @brief This file should be the first included by all AWS-LC headers.
+ */
 
 #include <stddef.h>
 #include <stdint.h>
@@ -420,6 +423,12 @@ typedef struct evp_aead_st EVP_AEAD;
 typedef struct evp_aead_ctx_st EVP_AEAD_CTX;
 typedef struct evp_cipher_ctx_st EVP_CIPHER_CTX;
 typedef struct evp_cipher_st EVP_CIPHER;
+
+/**
+ * @typedef EVP_ENCODE_CTX
+ * @copydoc evp_encode_ctx_st
+ * @see evp_encode_ctx_st
+ */
 typedef struct evp_encode_ctx_st EVP_ENCODE_CTX;
 typedef struct evp_hpke_aead_st EVP_HPKE_AEAD;
 typedef struct evp_hpke_ctx_st EVP_HPKE_CTX;


docs/api/Doxyfile
sgmenda, line 1: This file is *very very* long making it hard to review. Can we compress this to just the not-default options? it would also be helpful to summarize the high-level design in a short, friendly `docs/api/README.md`.

STYLE.md
sgmenda, line 1: This now seems to include two style guides, one for aws-lc and one for BoringSSL. Can we remove the boring one?

sgmenda: How many of these can we automate? I'd first put things that cannot be automated, and leave the syntax in an appendix.

m271828: The style guide covers more than documentation, it also covers code conventions. The portion covering documentation using BoringSSL's tool was removed, the rest was left in place (with a future task of updating the documentation style in the samples in the rest of this doc).

The automation would need to be done on the developer end. Most IDEs with generate a blank template if you type `/**` and hit enter. Depending on your specific IDE, you might need to modify the default templates. Since the public headers aren't updated often, automation is less of a concern than the initial change.

I don't think moving the more specific guidelines into a separate file is necessarily needed, it matches the level of detail that follows on what data types to use, single statement blocks aren't allowed, etc. The main reason this is longer than what was in place before is it referred to another style guide and Doxygen is a more generalized tool.

docs/api/Doxyfile
m271828: The current diff is:
```
# Difference with default Doxyfile 1.15.0
PROJECT_NAME           = aws-lc
PROJECT_BRIEF          = "API documentation may be out of date."
OUTPUT_DIRECTORY       = ./
TAB_SIZE               = 2
OPTIMIZE_OUTPUT_FOR_C  = YES
INLINE_GROUPED_CLASSES = YES
TIMESTAMP              = YES
EXTRACT_ALL            = YES
EXTRACT_STATIC         = YES
SHOW_INCLUDE_FILES     = NO
SHOW_USED_FILES        = NO
SHOW_NAMESPACES        = NO
LAYOUT_FILE            = ./DoxygenLayout.xml
WARN_IF_UNDOC_ENUM_VAL = YES
WARN_AS_ERROR          = YES
INPUT                  = ../../include/
FILE_PATTERNS          = *.c \
                         *.h
RECURSIVE              = YES
EXCLUDE                = ../../include/openssl/arm_arch.h \
                         ../../include/openssl/asm_base.h
DISABLE_INDEX          = YES
GENERATE_LATEX         = NO
CLASS_GRAPH            = NO
COLLABORATION_GRAPH    = NO
INCLUDE_GRAPH          = NO
INCLUDED_BY_GRAPH      = NO
```
Right now the diff is large since it's a new file to the repo but I have had to change some settings in other branches where I've converted header files to Doxygen (mostly related to linking and handling cross-platform compilation). It's easier figure out what's going on with the full Doxyfile.

I see a couple different options:
1. Accept that the initial commit it a large diff and use what I have here (or pull the branch and run `doxygen -x` on the Doxyfile) and know that future branches will only show the changes in PRs.
2. I keep the full Doxyfile locally and we only keep the diff version checked in (some risk of accidentally dropping changes across branches or loosing the Doxyfile).
3. Keep the full version for now, when all header files have been converted, replace the full Doxyfile with the diff version for long term ease of use. That does drop all of the variable descriptions, as seen above.

I'm not sure what type of high level design you'd be looking for as there isn't much. In the files here we're just setting it up for C and our public header files. Can you explain what you're looking for?

PR #2916, opened 2025-12-17
Fix extension processing order in x509 cli

nhatnghiho:
### Description of changes: 
Our `x509` cli currently does extension handling before generating a serial number for the certificate, which would lead to an incorrect AKID for self-signed certificates. This PR fixes that.

### Testing:
Unit tests

### DIFF
--- a/tool-openssl/x509.cc
+++ b/tool-openssl/x509.cc
@@ -221,7 +221,7 @@ static bool LoadExtensionsAndSignCertificate(const X509 *issuer, X509 *subject,
   bssl::UniquePtr<CONF> ext_conf(nullptr);
 
   // Determine if this is a self-signed certificate
-  int self_sign = X509_check_private_key(subject, pkey);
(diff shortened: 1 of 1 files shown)


tool-openssl/x509.cc
```diff
@@ -220,6 +220,23 @@ static bool LoadExtensionsAndSignCertificate(const X509 *issuer, X509 *subject,
   X509V3_CTX ext_ctx;
   bssl::UniquePtr<CONF> ext_conf(nullptr);
 
+  // Determine if this is a self-signed certificate
+  int self_sign = X509_check_private_key(subject, pkey);
```
justsmth: Based on how this function is used, wouldn't this be easier?
```
  // Determine if this is a self-signed certificate
  int self_sign = (issuer == subject);
```

nhatnghiho: Yeah you're right!

PR #2917, opened 2025-12-18
Rename SHAKE128_Init_x4 to the more general SHAKE_Init_x4

manastasova:
### Issues:
mldsa-native import needs SHAKE256_* functions. SHAKE256_Absorb_once_x4 and SHAKE256_Squeezeblocks_x4 are already supported. SHAKE256_Init_x4 is missing, however, SHAKE128_Init_x4 can be reused as-is for both SHAKE128 and SHAKE256. Therefore, generalizing the function avoids duplicates.

### Description of changes: 
Rename SHAKE128_Init_x4 to the more general function SHAKE_Init_x4, covering both SHAKE security levels.

### DIFF
--- a/crypto/fipsmodule/ml_kem/fips202x4_glue.h
+++ b/crypto/fipsmodule/ml_kem/fips202x4_glue.h
@@ -38,7 +38,7 @@ static MLK_INLINE void mlk_shake128x4_squeezeblocks(uint8_t *out0, uint8_t *out1
 static MLK_INLINE void mlk_shake128x4_init(mlk_shake128x4ctx *state) {
   // Return code check can be omitted
   // since mlkem-native adheres to call discipline
-  (void) SHAKE128_Init_x4(state);
+  (void) SHAKE_Init_x4(state);
 }
 
 static MLK_INLINE void mlk_shake128x4_release(mlk_shake128x4ctx *state) {
--- a/crypto/fipsmodule/sha/internal.h
+++ b/crypto/fipsmodule/sha/internal.h
@@ -567,11 +567,11 @@ int SHAKE_Final(uint8_t *md, KECCAK1600_CTX *ctx, size_t len);
  * detailed above each SHAKE128_x4_ function signature, is satisfied.
  */
 
-// SHAKE128_Init_x4 is a batched API that operates on four independent
+// SHAKE_Init_x4 is a batched API that operates on four independent
 // Keccak bitstates. It initialises all four |ctx| fields and returns
 // 1 on success and 0 on failure. When call-discipline is maintained,
 // this function never fails.
-OPENSSL_EXPORT int SHAKE128_Init_x4(KECCAK1600_CTX_x4 *ctx);
+OPENSSL_EXPORT int SHAKE_Init_x4(KECCAK1600_CTX_x4 *ctx);
 
 // SHAKE128_Absorb_once_x4 is a batched API that operates on four independent
 // Keccak bitstates. It absorbs all four inputs |data0|, |data1|, |data2|, |data3|
--- a/crypto/fipsmodule/sha/sha3.c
+++ b/crypto/fipsmodule/sha/sha3.c
@@ -454,7 +454,7 @@ int SHAKE_Squeeze(uint8_t *md, KECCAK1600_CTX *ctx, size_t len) {
 /*
  * SHAKE batched (x4) APIs implement SHAKE functionalities in batches of four on top of SHAKE API layer
  */
-int SHAKE128_Init_x4(KECCAK1600_CTX_x4 *ctx) {
+int SHAKE_Init_x4(KECCAK1600_CTX_x4 *ctx) {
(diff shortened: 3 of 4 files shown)


crypto/fipsmodule/sha/internal.h
```diff
@@ -567,11 +567,11 @@ int SHAKE_Final(uint8_t *md, KECCAK1600_CTX *ctx, size_t len);
  * detailed above each SHAKE128_x4_ function signature, is satisfied.
  */
 
-// SHAKE128_Init_x4 is a batched API that operates on four independent
+// SHAKE_Init_x4 is a batched API that operates on four independent
 // Keccak bitstates. It initialises all four |ctx| fields and returns
 // 1 on success and 0 on failure. When call-discipline is maintained,
 // this function never fails.
-OPENSSL_EXPORT int SHAKE128_Init_x4(KECCAK1600_CTX_x4 *ctx);
+OPENSSL_EXPORT int SHAKE_Init_x4(KECCAK1600_CTX_x4 *ctx);
```
nebeid, line 574: While the declaration is now generic, the block comment that governs it is still scoped to SHAKE128. Just above, the call-discipline preamble reads "the `SHAKE128_x4_` batched APIs ...", "`SHAKE128_x4_` context must go through the flow: (a) Init_x4, (b) Absorb_once_x4 [one time; maximum input length of `SHAKE128_BLOCKSIZE - 1`] ...", and "detailed above each `SHAKE128_x4_` function signature".

The point of the rename is that a future SHAKE256 x4 batched API will share this Init, but a caller arriving at `SHAKE_Init_x4` finds a SHAKE128-only flow and a SHAKE128-only input-length bound, with no stated discipline for the SHAKE256 rate.

Could we split the preamble so the generic Init's contract is not buried under a SHAKE128-specific one -- either hoisting the rate-independent parts (Init once, then Absorb_once, then Squeezeblocks) and leaving the `SHAKE128_BLOCKSIZE - 1` bound attached to `SHAKE128_Absorb_once_x4` where it actually applies?

nebeid (changes requested):

crypto/fipsmodule/ml_kem/fips202x4_glue.h
```diff
@@ -38,7 +38,7 @@ static MLK_INLINE void mlk_shake128x4_squeezeblocks(uint8_t *out0, uint8_t *out1
 static MLK_INLINE void mlk_shake128x4_init(mlk_shake128x4ctx *state) {
   // Return code check can be omitted
   // since mlkem-native adheres to call discipline
-  (void) SHAKE128_Init_x4(state);
+  (void) SHAKE_Init_x4(state);
```
nebeid, line 41: The ml_dsa sibling of this file was missed, so the tree does not build at this head. Two issues, plus a note.

**1. Build break: the renamed callee was not updated in `ml_dsa`.**

`crypto/fipsmodule/ml_dsa/fips202x4_glue.h:47` still calls the old name:

```c
static MLD_INLINE void mld_shake128x4_init(mld_shake128x4ctx *state) {
  // Return code check can be omitted
  // since mldsa-native adheres to call discipline
  (void) SHAKE128_Init_x4(state);   // -> SHAKE_Init_x4
}
```

It is live code, not dead: `ml_dsa/mldsa_native_config.h:19` wires the header in with `#define MLD_CONFIG_FIPS202X4_CUSTOM_HEADER "../fips202x4_glue.h"`, and `bcm.c:132` compiles `ml_dsa/ml_dsa.c` into the module. Since the declaration and definition are both renamed in this PR, `bcm.c` fails to compile. That file also exists at this PR's base (`eab96790b`), so it was reachable when the rename was written -- this glue header got updated and the ml_dsa one did not. Same one-line change.

**2. A rebase will also need the exported-symbol lists updated.**

`crypto/libcrypto.txt` and `crypto/libcrypto.map` did not exist at this PR's base but do on current `main`, and both still name `SHAKE128_Init_x4` (`libcrypto.txt:2092`, `libcrypto.map:2099`). Left unchanged, the version script names a symbol the library no longer defines -- a link failure under `--no-undefined-version`, and the symbol-check CI will flag the missing export. The entry is tagged `AWS_LC_1.0 PRIVATE`, so this is not a public ABI break and the outright rename is fine; no compatibility alias needed.

(A third point, on the SHAKE128-scoped call-discipline comment above `SHAKE_Init_x4`, is in a separate comment on `crypto/fipsmodule/sha/internal.h`.)

Note: this branch is ~6 months behind `main` (`sha/internal.h` has since shifted from line 567 to 589), so a rebase is needed regardless.

For what it is worth, the rename itself is sound: the Keccak rate is supplied per-call by `SHAKE128_Absorb_once_x4` / `SHAKE256_Absorb_once_x4` (`sha3.c:454-490`) and `SHAKE_Init_x4` only zeroes `KECCAK1600_CTX_x4`, so the function genuinely is security-level agnostic.

PR #2920, opened 2025-12-19
Service Indicator: Add error call trampoline to avoid delocator issue

jakemas:
### Problem

On AArch64, the delocator can patch up PC-relative addressing only if the references can be computed with a PC-relative offset in the range (-1MB, 1MB).

For the `OPENSSL_PUT_ERROR` macro calls in `crypto/fipsmodule/service_indicator/service_indicator.c`, the `__FILE__` string literal references were exceeding the ARM64 ADR instruction's ±1MB range limit as the FIPS module grew larger. This manifested as build failures in ARM64 FIPS builds:

```
error: fixup value out of range
 adr x3, .L.str.192
 ^
```

The error originated from string literals being placed too far from their usage sites (e.g., string at line ~365493, first use at line ~2597 in bcm-delocated.S).

### Solution

This commit fixes the issue by adding a non-inlined trampoline function for error calls:

```c
#if defined(_MSC_VER)
__declspec(noinline)
#else
__attribute__((noinline))
#endif
static void put_set_thread_local_error(void) {
  OPENSSL_PUT_ERROR(CRYPTO, ERR_R_INTERNAL_ERROR);
}
```

This ensures the `__FILE__` string literal stays close to its usage site and can be addressed using a PC-relative offset within the allowed range.

### Key Distinction from Function Pointer Trampolines

This fix differs from previous function pointer trampolines (e.g., #2165), which wrapped assembly functions and used branch instructions to call them. Here, the pointer is not to a function but to a string literal (`__FILE__` from error macros), so instead we wrap the function that uses the pointer to keep the string reference close to its usage. The `noinline` directive is critical - without it, the compiler will inline the function, placing the `__FILE__` reference back at the original call sites and recreating the distance problem.

### DIFF
--- a/.github/workflows/image-build-windows.yml
+++ b/.github/workflows/image-build-windows.yml
@@ -17,6 +17,7 @@ on:
     paths:
       - .github/docker_images/aws-lc/windows/Dockerfile.ltsc2022
       - .github/docker_images/scripts/**
+      - .github/workflows/image-build-windows.yml
 concurrency:
   group: ${{ inputs.concurrency_prefix || github.workflow }}-${{ github.ref_name }}
   cancel-in-progress: true
@@ -42,6 +43,8 @@ jobs:
       vs2022: ${{ steps.images.outputs.vs2022 }}
       sde: ${{ steps.images.outputs.sde }}
     steps:
+      - name: Enable long paths
+        run: git config --global core.longpaths true
       - uses: actions/checkout@v5
       - name: Login to Amazon ECR
         id: login-ecr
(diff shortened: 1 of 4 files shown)


justsmth (approved):

crypto/fipsmodule/service_indicator/service_indicator.c
```diff
@@ -20,6 +20,19 @@ int is_fips_build(void) {
 
 #if defined(AWSLC_FIPS)
 
+// Trampoline function to avoid ARM64 ADR range issues in large FIPS module.
+// This function is intentionally not inlined to ensure the __FILE__ string
+// literal reference stays close to the call site, avoiding the ±1MB PC-relative
+// addressing limit of the ARM64 ADR instruction.
+#if defined(_MSC_VER)
+__declspec(noinline)
```
justsmth, line 28: NP: this line is not strictly necessary since the delocator is not used on Windows; but I think it's fine for consistency.

PR #2922, opened 2025-12-29
Add randomized unit testing for EVP_CIPHERs

dkostic:
### Issues:
N/A

### Description of changes: 
Added a unit test that performs end-to-end testing of all
EVP_CIPHERs with random key, IV, plaintext, aad.

### Call-outs:
Locally, I reverted the XTS fix (https://github.com/aws/aws-lc/commit/3f282aa6fd16f6afce9e7d0d10ed5d9fdb713efa) and verified that this new unit test would catch the bug.

### Testing:

### DIFF
--- a/crypto/cipher_extra/cipher_test.cc
+++ b/crypto/cipher_extra/cipher_test.cc
@@ -1521,7 +1521,7 @@ TEST_P(RandomizedCipherTest, EncryptDecrypt) {
   const bool is_xts = mode == EVP_CIPH_XTS_MODE;
 
   for (size_t pt_len = 0; pt_len <= max_input_size; pt_len++) {
-    // Filter based on cipher mode constraints
+    // Filter based on cipher mode constraints.
     if (is_xts && pt_len < 16) {
       continue;  // XTS requires at least 16 bytes
     }


crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1454,188 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+    All, RandomizedCipherTest, testing::ValuesIn(kAllCiphers),
+    [](const testing::TestParamInfo<CipherInfo> &info) -> std::string {
+      std::string name = info.param.name;
+      std::replace(name.begin(), name.end(), '-', '_');
+      return name;
+    });
+
+TEST_P(RandomizedCipherTest, EncryptDecrypt) {
+
+  const size_t max_input_size = 1601;
```
samuel40791765, line 1509: Why's the max input size 1601 here? Should we document the reasoning?

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1454,188 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+  const size_t block_size = EVP_CIPHER_block_size(cipher);
+  const uint32_t mode = EVP_CIPHER_mode(cipher);
+  const bool is_aead = EVP_CIPHER_flags(cipher) & EVP_CIPH_FLAG_AEAD_CIPHER;
+  const bool is_ccm = mode == EVP_CIPH_CCM_MODE;
+  const bool is_xts = mode == EVP_CIPH_XTS_MODE;
+
+  for (size_t pt_len = 0; pt_len <= max_input_size; pt_len++) {
+    // Filter based on cipher mode constraints
+    if (is_xts && pt_len < 16) {
+      continue;  // XTS requires at least 16 bytes
```
samuel40791765, line 1526: Nit: Google test provides `GTEST_SKIP()` for test cases like these, we can consider using it

samuel40791765 (dismissed):

skmcgrail (approved):

crypto/cipher_extra/cipher_test.cc
dkostic: `GTEST_SKIP` skips the whole test, but I want only to skip the test for certain input sizes.

samuel40791765: Ahh I see what you're saying. The for loop should continue running.

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1454,3 +1454,188 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+      }
+
+      if (!aad.empty()) {
+        int len;
+        ASSERT_TRUE(EVP_DecryptUpdate(ctx.get(), nullptr, &len, aad.data(), aad.size()));
+      }
+
+      std::vector<uint8_t> decrypted(ct_len + block_size);
+      out_len = 0;
+      if (ct_len > 0) {
```
WillChilds-Klein, line 1630: when would `ct_len` be 0? in that case, would there still be an auth tag? if not, we can probably either skip this case earlier or at least avoid the allocation on L1623.

skmcgrail (approved):

samuel40791765 (dismissed):

WillChilds-Klein (dismissed):

crypto/cipher_extra/cipher_test.cc
dkostic: `ct_len` is 0 when the plaintext length is 0, but for AEADs like AES-GCM and ChaCha20-Poly1305 we still have the auth tag.

skmcgrail (approved):

PR #2925, opened 2026-01-01
Replace password string with proper class

samuel40791765:
The `bssl::UniquePtr<std::string>` pattern that we're using for passwords in the CLI tool is a bit confusing. C++ `std::string` class automatically manages its own memory.

This was an attempt to manage the cleansing of passwords/secrets once they were out of scope, but the existing scheme forces the developer to have to remember to wrap Password strings with these pointers. Replacing the `bssl::UniquePtr<std::string>` pattern with a proper Password class should enforce correct usage through stronger typing and make the code more readable.

### DIFF
--- a/tool-openssl/internal.h
+++ b/tool-openssl/internal.h
@@ -37,63 +37,63 @@ bool isStringUpperCaseEqual(const std::string &a, const std::string &b);
 // Password class that wraps std::string with secure memory clearing
 class Password {
 private:
-    std::string data_;
-    
-    // Helper method to securely clear current data
-    void secure_clear() {
-        if (!data_.empty()) {
-            OPENSSL_cleanse(&data_[0], data_.size());
-        }
+  std::string data_;
+
+  // Helper method to securely clear current data
+  void secure_clear() {
+    if (!data_.empty()) {
+      OPENSSL_cleanse(&data_[0], data_.size());
     }
+  }
 
 public:
-    // Default constructor
-    Password() = default;
-    
-    // String constructor
-    explicit Password(const std::string& str) : data_(str) {}
-    explicit Password(std::string&& str) : data_(std::move(str)) {}
-    
-    // Copy constructor and assignment
-    Password(const Password& other) : data_(other.data_) {}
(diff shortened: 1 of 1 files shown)


tool-openssl/internal.h
```diff
@@ -27,6 +27,68 @@ bool IsNumeric(const std::string &str);
...
+    // Implicit conversion for ease of use
+    operator std::string&() { return data_; }
+    operator const std::string&() const { return data_; }
+    
+    // Common string operations
+    bool empty() const { return data_.empty(); }
+    size_t size() const { return data_.size(); }
+    void clear() { 
+        secure_clear();
+        data_.clear();
```
WillChilds-Klein: is this necessary? if so, should we also call it in the destructor?

tool-openssl/internal.h
```diff
@@ -27,6 +27,68 @@ bool IsNumeric(const std::string &str);
...
+    // String constructor
+    explicit Password(const std::string& str) : data_(str) {}
+    explicit Password(std::string&& str) : data_(std::move(str)) {}
+    
+    // Copy constructor and assignment
+    Password(const Password& other) : data_(other.data_) {}
+    Password& operator=(const Password& other) {
+        if (this != &other) {
+            secure_clear();
+            data_ = other.data_;
```
WillChilds-Klein: making a note, no change requested -- confirmed that std::string's assignment operator [ensures](https://cplusplus.com/reference/string/string/operator=/) that `other.data_`'s

```
value is either copied (1) or moved (5) if different from *this (if moved, str is left in an unspecified but valid state).
```

so no need to worry about duplicate ownership after this assignment

tool-openssl/internal.h
samuel40791765: `clear()` leaves the password string in a usable state (so string must still exist, but be logically empty). The destructor destroys the object entirely, so only the security wipe matters since `std::string` will go out of scope.

I can still call it in the destructor for consistency though.

WillChilds-Klein (approved):

PR #2926, opened 2026-01-02
Fix ppc64le; Improve platform detection

justsmth:
### Issues:
Addresses: aws/aws-lc-rs#980

### Description of changes: 
* Fixes powerpc64 detection w/ certain compilers. See [here](https://github.com/aws/aws-lc-rs/issues/980#issuecomment-3706604663).
* Makes the detection of 64-bit and endian-ness consistent across all targets.
  * Adds sanity checks in "base.h" for what was detected.

### Testing:
All existing CI jobs (including those for cross-compilation) pass.

### DIFF
--- a/include/openssl/target.h
+++ b/include/openssl/target.h
@@ -22,8 +22,23 @@
 // <openssl/base.h> and <openssl/asm_base.h>. Prefer to include those headers
 // instead.
 
-#if (defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) &&  __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__)
-#define OPENSSL_BIG_ENDIAN
+#if defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__)
+#  if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+#    define OPENSSL_BIG_ENDIAN
+#  endif
+#elif defined(__has_include)
+#  if __has_include(<endian.h>)
+#    include <endian.h>
+#  elif __has_include(<sys/param.h>)
+#    include <sys/param.h>
+#  endif
+#  if (defined(__BYTE_ORDER) && defined(__BIG_ENDIAN) && __BYTE_ORDER == __BIG_ENDIAN) || \
+      (defined(_BYTE_ORDER) && defined(_BIG_ENDIAN) && _BYTE_ORDER == _BIG_ENDIAN) || \
+      (defined(BYTE_ORDER)  && defined(BIG_ENDIAN)  && BYTE_ORDER == BIG_ENDIAN)
+#    define OPENSSL_BIG_ENDIAN
+#  endif
+#elif defined(_M_PPC)
+#  define OPENSSL_BIG_ENDIAN
 #endif
 
 #if (defined(__SIZEOF_POINTER__) && __SIZEOF_POINTER__ == 8) || defined(__LP64__) || defined(_WIN64)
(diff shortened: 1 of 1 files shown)


include/openssl/type_check.h
```diff
@@ -63,40 +63,6 @@
 extern "C" {
 #endif
 
-
-// Previously we defined |OPENSSL_STATIC_ASSERT| to use one of two keywords:
```
torben-hansen, line 67: the first paragraph containing history is deleted when coded moved.

include/openssl/base.h
```diff
@@ -235,6 +235,53 @@ extern "C" {
 #define OPENSSL_UNUSED
 #endif
 
+// C99-compatible static assertion using bit-field width trick.
```
torben-hansen, line 238: In theory one could use macro's from type_check.h without importing base.h, which now doesn't work. Although, unlikely IMO and would be a build error.

include/openssl/target.h
```diff
@@ -22,68 +22,57 @@
...
-#define OPENSSL_64_BIT
+#elif defined(__MIPSEL__) || defined(__MIPSEB__)
+#  if defined(OPENSSL_64_BIT)
+#    define OPENSSL_MIPS64
+#  else
+#    define OPENSSL_MIPS
+#  endif
+#elif defined(__riscv) && defined(OPENSSL_64_BIT)
 #define OPENSSL_RISCV64
-#elif defined(__riscv) && __SIZEOF_POINTER__ == 4
```
torben-hansen, line 73: In new version this is no longer an option. Intentional?

include/openssl/base.h
justsmth: The macro would still be available. "type_check.h" includes "base.h": https://github.com/aws/aws-lc/blob/main/include/openssl/type_check.h#L60

include/openssl/target.h
justsmth:
```
#elif defined(__riscv) && __SIZEOF_POINTER__ == 4
#define OPENSSL_32_BIT
```

We don't have an `OPENSSL_RISCV` macro anywhere, but this is a valid concern as this was removing support for 32-bit riscv. I'll update.

include/openssl/base.h
```diff
@@ -235,6 +235,63 @@ extern "C" {
...
+        unsigned int AWSLC_CONCAT(static_assertion_, msg) : (cond) ? 1 : -1; \
+    } AWSLC_CONCAT(static_assertion_, msg) OPENSSL_UNUSED;
+#define AWSLC_STATIC_ASSERT_ADD_LINE0(cond, suffix) AWSLC_STATIC_ASSERT_DEFINE(cond, AWSLC_CONCAT(at_line_, suffix))
+#define AWSLC_STATIC_ASSERT_ADD_LINE1(cond, line, suffix) AWSLC_STATIC_ASSERT_ADD_LINE0(cond, AWSLC_CONCAT(line, suffix))
+#define AWSLC_STATIC_ASSERT_ADD_LINE2(cond, suffix) AWSLC_STATIC_ASSERT_ADD_LINE1(cond, __LINE__, suffix)
+#define AWSLC_STATIC_ASSERT_ADD_ERROR(cond, suffix) AWSLC_STATIC_ASSERT_ADD_LINE2(cond, AWSLC_CONCAT(_error_is_, suffix))
+#define OPENSSL_STATIC_ASSERT(cond, error) AWSLC_STATIC_ASSERT_ADD_ERROR(cond, error)
+
+// Sanity check of "target.h": OPENSSL_64_BIT/OPENSSL_32_BIT must match actual pointer size
+#if defined(OPENSSL_64_BIT)
```
torben-hansen, line 275: One comment on this is that these tests are duplicated many many times per build because this code is included in quite a few compilation units. Not blocking, just wanted to mention it.

torben-hansen (approved):

include/openssl/target.h
```diff
@@ -22,68 +22,59 @@
...
 // instead.
 
-#if defined(__x86_64) || defined(_M_AMD64) || defined(_M_X64)
+#if (defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) &&  __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__)
+#define OPENSSL_BIG_ENDIAN
+#endif
+
+#if (defined(__SIZEOF_POINTER__) && __SIZEOF_POINTER__ == 8) || defined(__LP64__) || defined(_WIN64)
 #define OPENSSL_64_BIT
+#else
```
samuel40791765, line 46: Nit: Should we check `__SIZEOF_POINTER__ == 4` here as well? And bail with a `#error "Unknown size` on `#else`.

samuel40791765 (dismissed):

include/openssl/target.h
justsmth: The problem is that `__SIZEOF_POINTER__` is not defined for every compiler, but we do verify in "base.h" that the `sizeof` a pointer is correct.

PR #2933, opened 2026-01-08
Add weekly automated check for outdated third-party test vectors

sgmenda:
### Issues:
We currently rely on manual updates of test vectors from third-party sources like Wycheproof. This PR automates the detection of outdated vectors.

### Description of changes: 
Adds a GitHub Actions workflow that runs weekly on Wednesdays at 10 AM PT to check if third-party test vectors need updating. When outdated vectors are detected, the workflow creates a GitHub issue with update instructions using a new issue template.

The workflow:
- Runs `sync.py --check` to detect outdated vectors without modifying files
- Checks for open issues with the "vectors" label
- If no issues, then creates an issue with the "vectors" label

The issue template provides step-by-step instructions for updating vectors locally, including running tests and regenerating build files.

### Call-outs:
- This only works for vectors onboarded to the `third_party/vectors` system
- The workflow detects and notifies but does not automatically update vectors
- It only checks vectors we actually use, not all available Wycheproof vectors
- fixes a bug in `sync.py --check`

### Testing:
The `sync.py --check` command was tested locally and confirmed to work correctly. The workflow can be manually triggered to debug any issues. The deduplication logic follows GitHub's recommended pattern for scheduled issue creation.

### DIFF
--- a/.github/workflows/check-test-vectors.yml
+++ b/.github/workflows/check-test-vectors.yml
@@ -40,6 +40,7 @@ jobs:
             echo "❌ Test vectors are outdated"
           fi
 
+      # https://docs.github.com/en/actions/tutorials/manage-your-work/schedule-issue-creation
       - name: Check for existing issue
         id: check_issues
         if: steps.check_vectors.outputs.vectors_outdated == 'true'
@@ -56,7 +57,6 @@ jobs:
             echo "❌ There is no issue"
           fi
 
-      # based on https://docs.github.com/en/actions/tutorials/manage-your-work/schedule-issue-creation
       - name: Create an issue for outdated vectors
         if: steps.check_issues.outputs.no_issues == 'true'
         run: |


.github/workflows/check-test-vectors.yml
```diff
@@ -0,0 +1,69 @@
...
+            echo "vectors_outdated=false" >> $GITHUB_OUTPUT
+            echo "✅ Test vectors are up-to-date"
+          else
+            echo "vectors_outdated=true" >> $GITHUB_OUTPUT
+            echo "❌ Test vectors are outdated"
+          fi
+
+      - name: Create Pull Request for outdated vectors
+        if: steps.check_vectors.outputs.vectors_outdated == 'true'
+        uses: peter-evans/create-pull-request@v8
```
WillChilds-Klein: why create a PR instead of an Issue?

.github/workflows/check-test-vectors.yml
```diff
@@ -0,0 +1,69 @@
...
+            ### Next Steps
+            - Run `cd third_party/vectors && python3 ./sync.py` locally to update vectors
+            - Review any new test vectors that are added
+            - Run `ninja -C build crypto_test` to ensure all tests pass
+            - Run `python3 util/generate_build_files.py` to update build files
+            - Verify `cd third_party/vectors && python3 ./sync.py --check` returns success
+
+            By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.
+          assignees: sgmenda
+          draft: false
```
WillChilds-Klein: if further action is required before the PR is ready for review (i.e. `./sync.py` locally), we should create it as a draft

.github/workflows/check-test-vectors.yml
sgmenda: good point, initially this was going to open a proper PR, but with the current version, an issue does seem more appropriate. what do you suggest?

.github/workflows/check-test-vectors.yml
sgmenda: good call, will change (if we choose to stick with opening a PR).

.github/workflows/check-test-vectors.yml
WillChilds-Klein: unless/until we can automate the full PR, I'd lean towards creating an Issue. i don't feel too strongly here, though.

could be a neat exercise in AI agent usage to have some bot actually cook up the PR here.

Issue #2283, opened 2025-03-20
AES-GCM slower than BoringSSL on Zen3


sticnarf:
### Problem:

Environment: EC2 c6a.large
AWS-LC version: https://github.com/aws/aws-lc/commit/652f7a94f98b98918d91641a0580dcf38b7ca07e
BoringSSL version: https://github.com/google/boringssl/commit/7c25e7308c3c322c2a0916ad03796945eb5b88eb

Run the benchmark suite of AWS-LC. 

AWS-LC result:

```
Did 88351000 AEAD-AES-128-GCM seal init operations in 10000023us (8835079.7 ops/sec)
Did 96643500 AEAD-AES-128-GCM seal (16 bytes) operations in 10000012us (9664338.4 ops/sec): 154.6 MB/s
Did 62820000 AEAD-AES-128-GCM seal (256 bytes) operations in 10000136us (6281914.6 ops/sec): 1608.2 MB/s
Did 24766500 AEAD-AES-128-GCM seal (1350 bytes) operations in 10000051us (2476637.4 ops/sec): 3343.5 MB/s
Did 5502000 AEAD-AES-128-GCM seal (8192 bytes) operations in 10000284us (550184.4 ops/sec): 4507.1 MB/s
Did 2844000 AEAD-AES-128-GCM seal (16384 bytes) operations in 10002055us (284341.6 ops/sec): 4658.7 MB/s
Did 88406750 AEAD-AES-128-GCM open init operations in 10000024us (8840653.8 ops/sec)
Did 91620000 AEAD-AES-128-GCM open (16 bytes) operations in 10000057us (9161947.8 ops/sec): 146.6 MB/s
Did 61806750 AEAD-AES-128-GCM open (256 bytes) operations in 10000005us (6180671.9 ops/sec): 1582.3 MB/s
Did 24261000 AEAD-AES-128-GCM open (1350 bytes) operations in 10000112us (2426072.8 ops/sec): 3275.2 MB/s
Did 5475000 AEAD-AES-128-GCM open (8192 bytes) operations in 10000813us (547455.5 ops/sec): 4484.8 MB/s
Did 2826000 AEAD-AES-128-GCM open (16384 bytes) operations in 10001913us (282545.9 ops/sec): 4629.2 MB/s
Did 82229250 AEAD-AES-256-GCM seal init operations in 10000032us (8222898.7 ops/sec)
[...]

PR #2934, opened 2026-01-09
Port AES-GCM AVX-2 implementation from BoringSSL

sticnarf:
### Issues:
Addresses #2283

### Description of changes: 

On x86_64 CPUs that support VAES + VPCLMULQDQ + AVX2 but do not support AVX-512 (notably AMD Zen 3 and some Intel client parts), AWS-LC would not take advantage of the newer VAES/VPCLMUL instructions.

This change ports BoringSSL’s AES-GCM AVX2 VAES + VPCLMULQDQ implementation (https://github.com/google/boringssl/blob/main/crypto/fipsmodule/aes/asm/aes-gcm-avx2-x86_64.pl) into AWS-LC.

### Call-outs:

The `aesni-gcm-avx2.pl` in this PR is mostly identical to BoringSSL's `aes-gcm-avx2-x86_64.pl` except:

- Some SEH directives are modified to make `x86_64-xlate.pl` work.
- `BORINGSSL_function_hit` index is set to 9.

### Testing:

`crypto_test` passes and `bssl` shows expected performance on my Zen 3 desktop.

### DIFF
--- a/crypto/fipsmodule/CMakeLists.txt
+++ b/crypto/fipsmodule/CMakeLists.txt
@@ -27,6 +27,7 @@ if(ARCH STREQUAL "x86_64")
   set(
     BCM_ASM_SOURCES
 
+    aesni-gcm-avx2.${ASM_EXT}
     aesni-gcm-avx512.${ASM_EXT}
     aesni-gcm-x86_64.${ASM_EXT}
     aesni-xts-avx512.${ASM_EXT}
@@ -119,6 +120,7 @@ endif()
 
 if(PERL_EXECUTABLE)
   perlasm(aesni-gcm-x86_64.${ASM_EXT} modes/asm/aesni-gcm-x86_64.pl)
+  perlasm(aesni-gcm-avx2.${ASM_EXT} modes/asm/aesni-gcm-avx2.pl)
   perlasm(aesni-gcm-avx512.${ASM_EXT} modes/asm/aesni-gcm-avx512.pl)
   perlasm(aesni-xts-avx512.${ASM_EXT} aes/asm/aesni-xts-avx512.pl)
   perlasm(aesni-x86_64.${ASM_EXT} aes/asm/aesni-x86_64.pl)
--- a/crypto/fipsmodule/modes/asm/aesni-gcm-avx2.pl
+++ b/crypto/fipsmodule/modes/asm/aesni-gcm-avx2.pl
@@ -1,0 +1,1176 @@
+#!/usr/bin/env perl
+# Copyright 2024 The BoringSSL Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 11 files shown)


sticnarf:
> Hi @sticnarf, thanks for your contribution! We are trying to prioritize reviewing and merging this PR, but we are missing some data on how impactful the performance improvement would be (we are weighing the performance benefit against the added complexity, code size, and potential for new bugs). What is your use case for this? Do you know of a service or app that uses Zen3 CPUs extensively?

My service involves heavy TLS data transfer and is subject to random scheduling across different nodes (including Zen 3 instances). I'm trying to optimize its CPU consumption. Currently, AES encryption/decryption accounts for 20% of the total CPU usage. That's why I hope this optimization can be added to AWS-LC.

By the way, I'm not so familiar with the wide variety of CI configurations here. It seems the failures mostly occur on older operating systems, which I suspect might be related to linker or compiler support for AVX2 instructions. Does it mean the code should also respect the `MY_ASSEMBLER_IS_TOO_OLD_FOR_512AVX`?

sticnarf:
@dkostic Tests on legacy Linux distributions pass now. The remaining CI failures are beyond my knowledge, so I’d appreciate some help if you’d like me to address them further.

dkostic:
> @dkostic Tests on legacy Linux distributions pass now. The remaining CI failures are beyond my knowledge, so I’d appreciate some help if you’d like me to address them further.

thanks! the formal verification CI job is failing. I'll have to find someone internally to take a look at this.

justsmth:
FYI -- I rebase this commit onto the latest from `main` and needed to resolve some conflicts.

justsmth:
Sorry, I had to do the rebase again. The keccak dispatch indices in `keccak1600.c` needed to be bumped to account for the new `BORINGSSL_function_hit` index 9 used by `aes_gcm_enc_update_vaes_avx2`. This was missed during my previous rebase.

namtran1812:
Hi @sticnarf — I was benchmarking the AVX2 AES-GCM implementation in this PR on an AMD EPYC 75F3 (Zen 3) and found a small scheduling optimization in the 48-byte encryption tail.

The current sequence computes the third-block XMM `VPCLMULQDQ` products and the first-two-block YMM products through the same temporary. I changed the schedule to issue the independent XMM/YMM products into separate destinations and accumulate each pair immediately. This keeps the GHASH algebra and number of CLMUL operations unchanged, while exposing more instruction-level parallelism.

On a pinned Zen 3 core using a direct `aes_gcm_enc_update_vaes_avx2` microbenchmark with 200M calls/run:

| | Baseline | Patched |
|---|---:|---:|
| Cycles/call (median) | 49.5304 | 47.3555 |
| Instructions/call | 149.0185 | 149.0185 |

That's a **4.39% reduction in cycles** for the 48-byte AES-128-GCM encryption update path, with the instruction count unchanged.

I reproduced the improvement across independent A/B measurements (~4.2–4.4%). I also ran a neighboring-size sweep; the improvement was localized to 48 bytes, while 16/32/64/80/96/128-byte measurements were within roughly ±0.2%.

After a clean rebuild, the targeted GCM/Wycheproof/NIST/ABI tests all pass (13/13).

I've isolated the optimization as a single signed-off commit directly on top of this PR's current `9bd1587` head:

[`545aec859`](https://github.com/namtran1812/aws-lc/commit/545aec859889295c680845c2dac3140f1a6247c3)

Happy to submit this in whatever form is easiest to review, or for you to cherry-pick the commit into this PR.

PR #2937, opened 2026-01-09
openssl-ca command implementation for self-sign certificates

skmcgrail:
### Description of changes: 
Adds a fairly limited `openssl ca` command that is targeted towards supporting the specific [efs-utils](https://github.com/aws/efs-utils) use case for generating and self-signing a certificate.

### Call-outs:
* This only supports the self-signed certificate use case
* Some code paths are retained to make it easier to add support for non-self-signed use cases in the future
* A limited set of command-line options were added to support efs-utils use-case, there are more that could be added in the future. There are analogous options in the configuration file that can be provided in lieu of the missing command line arguments.
* `openssl ca` has a series of "database" files that it record and tracks metadata in. Some of the behaviors like revocation checking is retained even though generation of revocations is not supported.

### DIFF
--- a/tool-openssl/ca.cc
+++ b/tool-openssl/ca.cc
@@ -15,9 +15,9 @@
 #include <openssl/digest.h>
 #include <openssl/ocsp.h>
 #include <openssl/pem.h>
-#include "txt_db/txt_db.h"
 #include "../crypto/fipsmodule/rsa/internal.h"
 #include "../crypto/rsa_extra/internal.h"
+#include "txt_db/txt_db.h"
 
 #include "ca_req_common.h"
 #include "internal.h"
@@ -99,7 +99,7 @@ static int WIN32_rename(const char *from, const char *to) {
   if (sizeof(TCHAR) == 1) {
     tfrom = (TCHAR *)from;
     tto = (TCHAR *)to;
-  } else { /* UNICODE path */
+  } else {  // UNICODE path
     size_t i, flen = strlen(from) + 1, tlen = strlen(to) + 1;
     tfrom = (TCHAR *)malloc(sizeof(TCHAR) * (flen + tlen));
     if (tfrom == NULL) {
@@ -129,7 +129,7 @@ static int WIN32_rename(const char *from, const char *to) {
   } else if (err == ERROR_ACCESS_DENIED) {
     errno = EACCES;
   } else {
-    errno = EINVAL; /* we could map more codes... */
+    errno = EINVAL; // we could map more codes...
   }
 err:
   ret = -1;
@@ -188,17 +188,15 @@ static EXT_COPY_TYPE ParseCopyExtensionValue(std::string value) {
 #define DB_type 0
 #define DB_exp_date 1
 #define DB_rev_date 2
-#define DB_serial 3 /* index - unique */
+#define DB_serial 3  // index - unique
 #define DB_file 4
-#define DB_name                           \
-  5 /* index - unique when active and not \
-     * disabled */
+#define DB_name 5  // index - unique when active and not disabled
 #define DB_NUMBER 6
 
-#define DB_TYPE_REV 'R'  /* Revoked  */
-#define DB_TYPE_EXP 'E'  /* Expired  */
-#define DB_TYPE_VAL 'V'  /* Valid ; inserted with: ca ... -valid */
-#define DB_TYPE_SUSP 'S' /* Suspended  */
+#define DB_TYPE_REV 'R'   // Revoked
+#define DB_TYPE_EXP 'E'   // Expired
+#define DB_TYPE_VAL 'V'   // Valid; inserted with: ca ... -valid
+#define DB_TYPE_SUSP 'S'  // Suspended
 
 static bssl::UniquePtr<CA_DB> LoadIndex(const std::string dbfile,
                                         DB_ATTR *db_attr) {
@@ -258,10 +256,10 @@ static bssl::UniquePtr<CA_DB> LoadIndex(const std::string dbfile,
 }
 
 static const char *crl_reasons[] = {
-    /* CRL reason strings */
+    // CRL reason strings
     "unspecified", "keyCompromise", "CACompromise", "affiliationChanged",
     "superseded", "cessationOfOperation", "certificateHold", "removeFromCRL",
-    /* Additional pseudo reasons */
+    // Additional pseudo reasons
     "holdInstruction", "keyTime", "CAkeyTime"};
 
 #define ARR_SIZE(x) (sizeof(x) / sizeof((x)[0]))
@@ -275,7 +273,6 @@ static int UnpackRevinfo(bssl::UniquePtr<ASN1_TIME> &prevtm, int *preason,
   ossl_char_ptr tmp = ossl_char_ptr(nullptr, OPENSSL_free);
   char *rtime_str = NULL, *reason_str = NULL, *arg_str = NULL, *p = NULL;
   int reason_code = -1;
-  int ret = 0;
   unsigned int i = 0;
   bssl::UniquePtr<ASN1_OBJECT> hold;
   bssl::UniquePtr<ASN1_GENERALIZEDTIME> comp_time;
@@ -283,7 +280,7 @@ static int UnpackRevinfo(bssl::UniquePtr<ASN1_TIME> &prevtm, int *preason,
   tmp.reset(OPENSSL_strdup(str.c_str()));
   if (!tmp) {
     fprintf(stderr, "memory allocation failure\n");
-    goto end;
+    return 0;
   }
 
   p = strchr(tmp.get(), ',');
@@ -304,11 +301,11 @@ static int UnpackRevinfo(bssl::UniquePtr<ASN1_TIME> &prevtm, int *preason,
   prevtm.reset(ASN1_UTCTIME_new());
   if (!prevtm) {
     fprintf(stderr, "memory allocation failure\n");
-    goto end;
+    return 0;
   }
   if (!ASN1_UTCTIME_set_string(prevtm.get(), rtime_str)) {
     fprintf(stderr, "invalid revocation date %s\n", rtime_str);
-    goto end;
+    return 0;
   }
 
   if (reason_str) {
@@ -320,22 +317,22 @@ static int UnpackRevinfo(bssl::UniquePtr<ASN1_TIME> &prevtm, int *preason,
     }
     if (reason_code == OCSP_REVOKED_STATUS_NOSTATUS) {
       fprintf(stderr, "invalid reason code %s\n", reason_str);
-      goto end;
+      return 0;
     }
 
     if (reason_code == 7) {
       reason_code = OCSP_REVOKED_STATUS_REMOVEFROMCRL;
-    } else if (reason_code == 8) { /* Hold instruction */
+    } else if (reason_code == 8) {  // Hold instruction
       if (!arg_str) {
         fprintf(stderr, "missing hold instruction\n");
-        goto end;
+        return 0;
       }
       reason_code = OCSP_REVOKED_STATUS_CERTIFICATEHOLD;
       hold.reset(OBJ_txt2obj(arg_str, 0));
 
       if (!hold) {
         fprintf(stderr, "invalid object identifier %s\n", arg_str);
-        goto end;
+        return 0;
       }
       if (phold) {
         phold = std::move(hold);
@@ -345,16 +342,16 @@ static int UnpackRevinfo(bssl::UniquePtr<ASN1_TIME> &prevtm, int *preason,
     } else if ((reason_code == 9) || (reason_code == 10)) {
       if (!arg_str) {
         fprintf(stderr, "missing compromised time\n");
-        goto end;
+        return 0;
       }
       comp_time.reset(ASN1_GENERALIZEDTIME_new());
       if (!comp_time) {
         fprintf(stderr, "memory allocation failure\n");
-        goto end;
+        return 0;
       }
       if (!ASN1_GENERALIZEDTIME_set_string(comp_time.get(), arg_str)) {
         fprintf(stderr, "invalid compromised time %s\n", arg_str);
-        goto end;
+        return 0;
       }
       if (reason_code == 9) {
         reason_code = OCSP_REVOKED_STATUS_KEYCOMPROMISE;
@@ -371,20 +368,14 @@ static int UnpackRevinfo(bssl::UniquePtr<ASN1_TIME> &prevtm, int *preason,
     pinvtm = std::move(comp_time);
   }
 
-  ret = 1;
-
-end:
-
-  return ret;
+  return 1;
 }
 
-/*-
- * Convert revocation field to X509_REVOKED entry
- * return code:
- * 0 error
- * 1 OK
- * 2 OK and some extensions added (i.e. V2 CRL)
- */
+// Convert revocation field to X509_REVOKED entry
+// return code:
+// 0 error
+// 1 OK
+// 2 OK and some extensions added (i.e. V2 CRL)
 static int MakeRevoked(X509_REVOKED *rev, std::string str) {
   int reason_code = -1;
   int i = 0, ret = 0;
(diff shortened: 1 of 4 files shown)


tool-openssl/internal.h
```diff
@@ -14,6 +14,19 @@
 #define O_BINARY 0
 #endif
 
+// Windows compatibility layer
+#ifdef _WIN32
```
justsmth:
```
#if defined(OPENSSL_WINDOWS)
```

tool-openssl/ca.cc
```diff
@@ -80,6 +80,66 @@ using ossl_string_ptr = std::unique_ptr<OPENSSL_STRING, ossl_free>;
 using ossl_uint8_ptr = std::unique_ptr<uint8_t, ossl_free>;
 using ossl_char_ptr = std::unique_ptr<char, ossl_free>;
 
+#ifdef _WIN32
```
justsmth:
```
#if defined(OPENSSL_WINDOWS)
```

tool-openssl/ca.cc
```diff
@@ -0,0 +1,1904 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR ISC
+
+#include "ca.h"
+
+#include <cctype>
+#include <cstdlib>
+#include <cstring>
+#include <memory>
+
+// TODO: figure out windows
```
justsmth: Comment still needed?

crypto/asn1/f_int.c
```diff
@@ -57,6 +57,100 @@
...
+    bufp = buf;
+    if (first) {
+      first = 0;
+      if ((bufp[0] == '0') && (bufp[1] == '0')) {
+        bufp += 2;
+        i -= 2;
+      }
+    }
+    k = 0;
+    i -= again;
```
justsmth: Why is there a `i -= again`? It's possible that the loop on lines 90-96 had stopped prior to reaching the '\\' character, so why decrease `i` based upon a character that exists potentially beyond the current position of `i`?

crypto/asn1/f_int.c
```diff
@@ -57,6 +57,100 @@
...
+    if (i % 2 != 0) {
+      OPENSSL_PUT_ERROR(ASN1, ASN1_R_ODD_NUMBER_OF_CHARS);
+      OPENSSL_free(s);
+      return 0;
+    }
+    i /= 2;
+    if (num + i > slen) {
+      sp = OPENSSL_realloc(s, num + i * 2);
```
justsmth: NP: Allocating `num + i * 2` bytes is unnecessary, since we only need `num + i` bytes to be allocated. This leads to 1/2 fewer allocations, but it also means we always over-allocate beyond what's needed.

crypto/asn1/f_int.c
```diff
@@ -57,6 +57,100 @@
...
+      s = sp;
+      slen = num + i * 2;
+    }
+    for (j = 0; j < i; j++, k += 2) {
+      for (n = 0; n < 2; n++) {
+        uint8_t hex = 0;
+        if (!OPENSSL_fromxdigit(&hex, bufp[k + n])) {
+          OPENSSL_PUT_ERROR(ASN1, ASN1_R_NON_HEX_CHARACTERS);
+          goto err;
+        }
```
justsmth, line 133: NP: This can (should?) never fail b/c of the loop on lines 90-96 truncates on any character that might fail. It's fine to leave, but I wonder where we should rethink the loop above.

crypto/asn1/f_int.c
```diff
@@ -57,6 +57,100 @@
 #include <openssl/asn1.h>
 
 #include <openssl/bio.h>
+#include <openssl/err.h>
+
+int a2i_ASN1_INTEGER(BIO *bp, ASN1_INTEGER *bs, char *buf, int size) {
```
justsmth, line 62: A comment above this function documenting what it expects from the input (and how it handles variance from expectations) would be helpful.

crypto/conf/conf.c
```diff
@@ -234,10 +237,87 @@ static int str_copy(CONF *conf, char *section, char **pto, char *from) {
...
+        OPENSSL_PUT_ERROR(CONF, ERR_R_MALLOC_FAILURE);
+        goto err;
+      }
+      while (*p) {
+        buf->data[to++] = *(p++);
+      }
+
+      // Since we change the pointer 'from', we also have to change the
+      // perceived length of the string it points at.  /RL
+      len -= e - from;
```
justsmth: `len` gets updated here, but it's never actually used for bounds checking.

crypto/conf/conf.c
justsmth, line 224: If this is not true, then `IS_EOF(conf, *from)` is true -- and we've reached the EOF w/o finding the matching closing quote. (Right?) Should we produce an error due to a malformed configuration like `"hello`?

crypto/conf/conf.c
```diff
@@ -234,10 +237,87 @@ static int str_copy(CONF *conf, char *section, char **pto, char *from) {
...
+      from = e;
+
+      // In case there were no braces or parenthesis around the
+      // variable reference, we have to put back the character that was
+      // replaced with a '\0'.  /RL
+      *rp = r;
```
justsmth: This was already done on line 294 above.

tool-openssl/txt_db/txt_db.cc
```diff
@@ -0,0 +1,338 @@
...
+#include <string.h>
+
+#include "../crypto/internal.h"
+#include "../crypto/lhash/internal.h"
+
+#undef BUFSIZE
+#define BUFSIZE 512
+
+
+TXT_DB *TXT_DB_read(BIO *in, int num) {
```
justsmth: The functions in this file get tested via "ca_test.cc", but I think it would be good to also have a "txt_db_test.cc" that covers many of the edge cases that couldn't easily be tested currently.

include/openssl/asn1.h
samuel40791765, line 1: Should we mark `a2i_ASN1_INTEGER` in our internal headers until it's actually needed? Same for the new APIs in `rsa.h`.

tool-openssl/txt_db/txt_db.cc
```diff
@@ -0,0 +1,338 @@
+/*
+ * Copyright 1995-2023 The OpenSSL Project Authors. All Rights Reserved.
+ *
+ * Licensed under the OpenSSL license (the "License").  You may not use
+ * this file except in compliance with the License.  You can obtain a copy
+ * in the file LICENSE in the source distribution or at
+ * https://www.openssl.org/source/license.html
+ */
```
samuel40791765: Nit: Comment format

tool-openssl/txt_db/txt_db.cc
```diff
@@ -0,0 +1,338 @@
...
+    }
+  }
+  return ret;
+err:
+  if (ret != NULL) {
+    sk_OPENSSL_PSTRING_free(ret->data);
+    OPENSSL_free(ret->index);
+    OPENSSL_free(ret->qual);
+    OPENSSL_free(ret);
+  }
```
samuel40791765: If we assign `ret` as a `bssl::UniquePtr`, will the C++ destructor automatically `TXT_DB_free` to free all the memory here?

I see that we've already mapped the necessary destructor in `txt_db.h`.

include/openssl/asn1.h
samuel40791765: Or is the cli tool only able to consume APIs from our external headers :(

tool-openssl/txt_db/txt_db.cc
```diff
@@ -0,0 +1,338 @@
...
+  ret = lh_OPENSSL_STRING_retrieve(lh, value);
+  db->error = DB_ERROR_OK;
+  return ret;
+}
+
+int TXT_DB_create_index(TXT_DB *db, int field, int (*qual)(OPENSSL_STRING *),
+                        lhash_OPENSSL_STRING_hash_func hash,
+                        lhash_OPENSSL_STRING_cmp_func cmp) {
+  LHASH_OF(OPENSSL_STRING) *idx = nullptr;
```
samuel40791765: Nit: could probably wrap this with a bssl::UniquePtr as well.

include/openssl/rsa.h
```diff
@@ -936,6 +936,8 @@ OPENSSL_EXPORT int RSA_print_fp(FILE *fp, const RSA *rsa, int indent);
 // the id-RSASSA-PSS key encoding.
 OPENSSL_EXPORT const RSA_PSS_PARAMS *RSA_get0_pss_params(const RSA *rsa);
 
+OPENSSL_EXPORT const RSASSA_PSS_PARAMS *RSA_get0_ssa_pss_params(const RSA *rsa);
```
samuel40791765: Nit: function documentation.

crypto/asn1/f_int.c
samuel40791765, line 62: Could we document a bit on what the function logic is doing is some places here? And a general function documentation for `a2i_ASN1_INTEGER` in `asn1.h` would be nice as well.

tool-openssl/test_util.cc
```diff
@@ -630,3 +630,117 @@ bool CheckKeyBoundaries(const std::string &content,
...
+      return false;
+    }
+    
+    if (EVP_PKEY_get_raw_public_key(private_key, priv_pubkey_raw, &priv_pubkey_len) != 1) {
+      std::cout << "Failed to extract ED25519 private key public key" << std::endl;
+      return false;
+    }
+    
+    // Compare the raw public key material
+    if (memcmp(cert_pubkey_raw, priv_pubkey_raw, 32) != 0) {
```
samuel40791765:
```suggestion
    if (OPENSSL_memcmp(cert_pubkey_raw, priv_pubkey_raw, 32) != 0) {
```

tool-openssl/ca_req_common.cc
```diff
@@ -0,0 +1,220 @@
...
+    return NULL;
+  }
+
+  // Remove newline character and carriage return if present
+  size_t len = OPENSSL_strnlen(buffer, buffer_size);
+  if (len > 0 && buffer[len - 1] == '\n') {
+    buffer[len - 1] = '\0';
+    len--;
+  }
+#if defined(_WIN32)
```
samuel40791765:
```suggestion
#if defined(OPENSSL_WINDOWS)
```

tool-openssl/ca.cc
```diff
@@ -0,0 +1,1904 @@
...
+    if (p) {
+      retdb->attributes.unique_subject = parse_bool(p, true);
+    }
+  }
+
+  retdb->dbfname = OPENSSL_strdup(dbfile.c_str());
+  retdb->dbst = dbst;
+
+err:
+  return retdb;
```
samuel40791765: Should we just `return retdb` on all of the places with `goto err`?

tool-openssl/ca.cc
```diff
@@ -0,0 +1,1904 @@
...
+      if (!hold) {
+        fprintf(stderr, "invalid object identifier %s\n", arg_str);
+        goto end;
+      }
+      if (phold) {
+        phold = std::move(hold);
+      } else {
+        hold.reset();
+      }
+    } else if ((reason_code == 9) || (reason_code == 10)) {
```
samuel40791765, line 342: Curious what the reason_code definitions are. Is there an RFC that we should reference? And should we abstract these numbers into macros or an enum?

tool-openssl/ca.cc
```diff
@@ -0,0 +1,1904 @@
...
+  GetString(&outfile, "-out", "", parsed_args);
+  GetString(&config_path, "-config", "", parsed_args);
+  GetString(&passin.get(), "-passin", "", parsed_args);
+  GetString(&start_date, "-startdate", "", parsed_args);
+  GetString(&end_date, "-enddate", "", parsed_args);
+  GetBoolArgument(&notext, "-notext", parsed_args);
+  GetBoolArgument(&self_sign, "-selfsign", parsed_args);
+  // GetBoolArgument(&batch, "-batch", parsed_args);
+  (void)self_sign;
+  (void)batch;
```
samuel40791765: Should we delete these?

tool-openssl/ca.cc
```diff
@@ -0,0 +1,1904 @@
...
+       preserveDN = false, rand_serial = false;
+  std::string ca_section, policy, keyfile, serialfile, extensions;
+  EXT_COPY_TYPE copy_extensions = EXT_COPY_NONE;
+  DB_ATTR dbattr = {0};
+  bssl::UniquePtr<CA_DB> db(nullptr);
+  const EVP_MD *md = nullptr;
+  DEF_DGST_USAGE md_usage = DEF_DGST_UNKNOWN;
+  bssl::UniquePtr<BIGNUM> serial;
+  const STACK_OF(CONF_VALUE) *attribs = nullptr;
+  auto cert_sk = sk_X509_new_null();
```
samuel40791765: Nit: could wrap this in a unique pointer

tool-openssl/ca.cc
```diff
@@ -0,0 +1,1904 @@
...
+  j = strlen(dbfile.c_str()) + strlen(suffix.c_str());
+  if (j + 6 >= BSIZE) {
+    fprintf(stderr, "file name too long\n");
+    goto err;
+  }
+  (void)BIO_snprintf(buf[2], sizeof(buf[2]), "%s.attr", dbfile.c_str());
+  (void)BIO_snprintf(buf[1], sizeof(buf[1]), "%s.attr.%s", dbfile.c_str(),
+                     suffix.c_str());
+  (void)BIO_snprintf(buf[0], sizeof(buf[0]), "%s.%s", dbfile.c_str(),
+                     suffix.c_str());
```
samuel40791765, line 498: I'm fine with ignoring the checked value, is there a particular reason for ignoring that we should document though?

crypto/asn1/f_int.c
skmcgrail: Shouldn't but it's just defensive check

include/openssl/asn1.h
skmcgrail: We could, I'd be more inclined to do that with the rsa functions since they don't fall under "compatibility" category. For `a2i_ASN1_INTEGER` it is function that existed in OpenSSL 1.1.1 that we just didn't have.

tool-openssl/ca.cc
skmcgrail: I need to dig into this and get back to you, but not sure if I will get to the bottom of it. The value names mostly come from the RFC, but since the indexing is off because value "7" is skipped there is this weird decode logic. But only values up to "removeFromCRL", after which they injected some pseudo-reasons not from the RFC, but then map them back to RFC values. i.e. `OCSP_REVOKED_STATUS_*` defines are inline with the RFC (well except for the -1 sentinel but). But the "crlReasons" numerical indexes are shifted by -1 from the RFC values starting at "removeFromCRL", and "privilegeWithdrawn" and "aACompromise" are not used.

crypto/asn1/f_int.c
skmcgrail: Added to the public header.

justsmth (approved):

tool-openssl/ca.cc
```diff
@@ -0,0 +1,1901 @@
...
+
+  if (reason_code != OCSP_REVOKED_STATUS_NOSTATUS) {
+    ret = 2;
+  } else {
+    ret = 1;
+  }
+
+end:
+
+  return ret;
```
samuel40791765: I think we could remove this `goto end` block and just return values everywhere

samuel40791765: Same comment for `SaveIndex`, `RotateIndex`, and potentially other functions that have `goto` statements in this file as well.

tool-openssl/ca.cc
samuel40791765, line 1: Nit: Should we clean up and align the commenting style in this file

tool-openssl/ca.cc
```diff
@@ -0,0 +1,1901 @@
...
+      OPENSSL_free(row[i]);
+    }
+  }
+
+  if (ok <= 0) {
+    ret.reset();
+  } else {
+    *xret = ret.release();
+  }
+  return ok;
```
samuel40791765, line 1276: We could probably wrap `row` with a type destructor

tool-openssl/txt_db/txt_db.cc
samuel40791765, line 1: Same C++ memory management style comment to this file as well.

tool-openssl/ca.cc
skmcgrail: I'll explore it in a follow-up

tool-openssl/txt_db/txt_db.cc
skmcgrail: same

Issue #1997, opened 2024-11-17
Windows 7 support? Missing `ProcessPrng`


0blu:
### Problem:

I actually came here from [aws-lc-rust](https://github.com/aws/aws-lc-rs). I tried to use the following minimal example with `x86_64-win7-windows-msvc` but as soon as the library tries to access `ProcessPrng` the program aborts.

```rust
use aws_lc_rs::rand::{SecureRandom, SystemRandom};

fn main() {
    let rng = SystemRandom::new();

    let mut buffer = [0u8; 32];
    println!("Going to fill bytes...");
    if rng.fill(&mut buffer).is_err() { // <-- Crash in `init_processprng`
        println!("Error filling random bytes");
    } else {
        println!("Everything is okay");
    }
}

```

Erroneous line:
https://github.com/aws/aws-lc/blob/745359e8569fdafa8897ac2fffdfd0fdcf620563/crypto/rand_extra/windows.c#L68

### Solution:
Rust recently switched from `BCryptGenRandom` to `ProcessPrng` but allowed a path for Windows 7.
https://github.com/rust-lang/rust/pull/121337

Is a similar solution also possible in this project?

justsmth:
Hello!

Thanks for letting us know about this problem, and for the reference to how Rust solved it.

I was trying to reproduce this with aws-lc-rs, but I'm having trouble with the build. I suspect my issue relates to not having the toolchain setup. (Since this is a tier 3 platform, are there any pre-built toolchains available?)

Here's what I tried (on a Windows host using the bash shell provided by Git), and the resulting error:
```
> rustup +nightly component add rust-src
...
> cargo +nightly build -Z build-std --target x86_64-win7-windows-msvc
...
...
  = note: LINK : fatal error LNK1181: cannot open input file 'windows.0.52.0.lib'␍


error: could not compile `aws-lc-rs` (example "digest") due to 1 previous error
```

If you can provide some guidance on how to setup a development environment for this, it would be appreciated. Thanks!

0blu:
Hey, I appreciate you being open to supporting a legacy system like Windows 7.
I am using a Windows 11 / Linux host for developing and a Windows 7 VM to test compatibility.

For my initial test project (shown above) I also used `cargo +nightly build --release --target x86_64-win7-windows-msvc -Zbuild-std` just like you, but with
```toml
[dependencies]
[...]

PR #2940, opened 2026-01-13
Windows 7 support

justsmth:
### Issues:
Addresses #1997

### Description of changes: 
`ProcessPrng` (from `bcryptprimitives.dll`) is only available on Windows 8+. So the current code fails at runtime on Windows 7 because `GetProcAddress` returns `NULL` and causes an `abort()`.

This change defines `AWSLC_WINDOWS_7_COMPAT` when `_WIN32_WINNT <= _WIN32_WINNT_WIN7`. Then uses the existing `BCryptGenRandom` logic when it's is defined.

* Related: aws/aws-lc-rs#1000

### DIFF
--- a/crypto/rand_extra/windows.c
+++ b/crypto/rand_extra/windows.c
@@ -26,7 +26,7 @@ OPENSSL_MSVC_PRAGMA(warning(push, 3))
 
 #include <windows.h>
 
-// ProcessPrng` (from `bcryptprimitives.dll`) is only available on Windows 8+.
+// ProcessPrng (from `bcryptprimitives.dll`) is only available on Windows 8+.
 #if !defined(__MINGW32__) && defined(_WIN32_WINNT) && _WIN32_WINNT <= _WIN32_WINNT_WIN7
 #define AWSLC_WINDOWS_7_COMPAT
 #endif


crypto/rand_extra/windows.c
```diff
@@ -26,16 +26,23 @@ OPENSSL_MSVC_PRAGMA(warning(push, 3))
 
 #include <windows.h>
 
-#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) && \
-    !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+// ProcessPrng (from `bcryptprimitives.dll`) is only available on Windows 8+.
+#if !defined(__MINGW32__) && defined(_WIN32_WINNT) && _WIN32_WINNT <= _WIN32_WINNT_WIN7
+#define AWSLC_WINDOWS_7_COMPAT
```
samuel40791765, line 31: Nit: Should we drop the `COMPAT`? seems like that's already implied like the rest of our compatibility macros like `OPENSSL_WINDOWS`. 
```suggestion
#define AWSLC_WINDOWS_7
```

PR #2941, opened 2026-01-13
Remove Kyber completely

torben-hansen:
### Issues:
* CryptoAlg-3052
* V2065732656

### Description of changes: 

Kyber has turned into ML-KEM and is no longer needed. Also remove some remnants of Dilithium and correct som inaccurate text in various files which wasn't updated when switching to ML-[KEM,DSA].

Did a bunch of `grep -ri "kyber" ./` and axed everything I could find.


tool/speed.cc
```diff
@@ -867,11 +867,8 @@ static bool SpeedKEM(std::string selected) {
 #if AWSLC_API_VERSION >= 30
          SpeedSingleKEM("ML-KEM-512", NID_MLKEM512, selected) &&
          SpeedSingleKEM("ML-KEM-768", NID_MLKEM768, selected) &&
-         SpeedSingleKEM("ML-KEM-1024", NID_MLKEM1024, selected) &&
+         SpeedSingleKEM("ML-KEM-1024", NID_MLKEM1024, selected);
```
justsmth: I think this needs an:
```
#else
    true;
#endif
```

torben-hansen: Right!

justsmth (approved):

PR #2946, opened 2026-01-17
Use existing session context if new is actually NULL

torben-hansen:
### Issues:

P369474299

### Description of changes: 

`SSL_set_SSL_CTX()` doesn't currently tolerate the `ctx` argument being `NULL` (it would just crash in some cases). I was pondering handling this by just explicitly error out if it was `NULL`. But I realised upstream OpenSSL has a different behaviour: falls back to existing session context from `ssl`.

This is unnecessary complexity, but in the name of interoperability I did that instead of flipping to an error state.


geedo0 (approved):
Confirmed that this is OpenSSL behavior.

https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/ssl/ssl_lib.c#L4111

PR #2948, opened 2026-01-17
Fix getentropy declaration for Android compilation

zsdsys:
### Issues:
Resolves aws-lc-rs/issues/1006

### Description of changes: 
added getentropy declarations for android to fix build error call to undeclared function 'getentropy'; ISO C99 and later do not support implicit function declarations

### Testing:
tested build via cargo android build against local updated repo


Kr328:
I think Android targets should use `/dev/urandom` (as Linux system do) instead of `getentropy`. `getentropy` functions in bionic libc requires Android 9.0 (SDK 28), which is too new to be available on more devices.

NOTE: `getentropy` definition in Android
```c
/**
 * [getentropy(3)](https://man7.org/linux/man-pages/man3/getentropy.3.html) fills the given buffer
 * with random bytes.
 *
 * Returns 0 on success, and returns -1 and sets `errno` on failure.
 *
 * Available since API level 28.
 *
 * See also arc4random_buf() which is available in all API levels.
 */

#if __BIONIC_AVAILABILITY_GUARD(28)
__nodiscard int getentropy(void* _Nonnull __buffer, size_t __buffer_size) __INTRODUCED_IN(28);
#endif /* __BIONIC_AVAILABILITY_GUARD(28) */
```

zsdsys:
I have reworked to use arc4random_buf

zsdsys:
this looks to be fixed better elsewhere, closing

PR #2964, opened 2026-01-27
Support GCC 4.8 for aarch64

justsmth:
### Issues:
Addresses: https://github.com/aws/aws-lc-rs/issues/1022

### Description of changes: 
* GCC v4.8 was the first to support aarch64, but it does not provide the `__ARM_ARCH` macro for this target.
* This change ensure that this macro is defined correctly when building for this target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1184,7 +1184,7 @@ endif()
 if(BUILD_TESTING)
   if(GO_EXECUTABLE)
     if(FIPS)
-      if(MSVC)
+      if(WIN32)
         set(ACVP_TOOL ${AWSLC_BINARY_DIR}/acvptool.exe)
         set(TEST_WRAPPER ${AWSLC_BINARY_DIR}/testmodulewrapper.exe)
       else()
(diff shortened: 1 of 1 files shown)


include/openssl/asm_base.h
```diff
@@ -70,6 +70,15 @@
 
 #if defined(OPENSSL_ARM) || defined(OPENSSL_AARCH64)
 
+#if defined(__aarch64__) && !defined(__ARM_ARCH)
+// Support for GCC v4.8+. GCC 4.8 was the first version to support aarch64
+// but it doesn't define __ARM_ARCH for aarch64 targets.
+#if defined(__GNUC__) && (__GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 8))
```
torben-hansen, line 76: `GNUC` doens't necessarily catch only GCC. But I guess it's good enough.

justsmth: Yeah. And any non-GCC toolchain that doesn't define `__ARM_ARCH` would've been failing this build previously as well.

include/openssl/asm_base.h
```diff
@@ -70,6 +70,15 @@
...
+  #error "GCC 4.8 or later is required for aarch64 support"
+#endif
```
WillChilds-Klein, line 78: nitpick: should we indent this inner conditional for clarity?

PR #2978, opened 2026-02-02
Miscellaneous CI improvements

skmcgrail:
### Description of changes: 
* Simplify a few CI duplicative jobs
* Fix abidiff so it doesn't hard-code the repository
* Add some additional GCC and Clang compiler to 24.04 images for more future simplification.

### DIFF
--- a/.github/docker_images/aws-lc/ubuntu/Dockerfile.2404
+++ b/.github/docker_images/aws-lc/ubuntu/Dockerfile.2404
@@ -21,7 +21,25 @@ apt-get -y --no-install-recommends install \
     ca-certificates \
     cmake \
     curl \
+    clang-14 \
+    clang-15 \
+    clang-16 \
+    clang-17 \
+    clang-18 \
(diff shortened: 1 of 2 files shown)


.github/docker_images/aws-lc/ubuntu/Dockerfile.2204
```diff
@@ -31,8 +31,10 @@ apt-get -y --no-install-recommends install \
     build-essential \
     ca-certificates \
     check \
+    clang-11 \
+    clang-12 \
+    clang-13 \
     clang-14 \
-    clang++-14 \
```
WillChilds-Klein, line 35: no longer needed? we install it in 24.04 below

skmcgrail: Turns out the clang package already installs what is needed for C and C++, and these lines were actually triggering a glob search pattern in apt that was installing unnecessary packages. :)

PR #2980, opened 2026-02-03
Add method to get type of ML-DSA instance configured under EVP PKEY

torben-hansen:
### Issues:

P375996324

### Description of changes: 

The actual ML-DSA instance configured under an EVP PKEY is two levels deep. And it's configured in an object whose definition is internal. There currently exist no accessor that consumers can use to access that. From the top-level the only information one can retrieve as a generic `EVP_PKEY_PQDSA` identifier.

This PR adds a function to get the more granular identifier. This information could be derived through e.g. the key size. But that's not a terribly good user experience.

I considered adding a getter for `PQDSA_KEY` that has a reference to the `PQDSA` object that configures the named identifier we are after. But none of those are referenced counted, so the implementation would look different from what the current pattern is for other PKEY key types (or one would need to implemented reference counting in the objects (no thanks)).

The concept "PQ DSA" is not a well-defined functional interface. But the abstraction has already escaped and no longer an internal implementation detail. So, I'm not sure we can revert back to a top-level `EVP_PKEY_foo` type anymore (e.g. similar to the 25519 pattern).

FWIW, same problem exist for the KEM key type. At least I don't immediately see a method to get the configured instance.

### Testing:

Simple test to check the correct NID is returned.

### DIFF
--- a/crypto/fipsmodule/evp/evp.c
+++ b/crypto/fipsmodule/evp/evp.c
@@ -300,6 +300,9 @@ int EVP_PKEY_pqdsa_get_type(const EVP_PKEY *pkey) {
     OPENSSL_PUT_ERROR(EVP, EVP_R_EXPECTING_A_PQDSA_KEY);
     return 0;
   }
+  if (!pkey->pkey.pqdsa_key || !pkey->pkey.pqdsa_key->pqdsa) {
+    return 0;
+  }
   return pkey->pkey.pqdsa_key->pqdsa->nid;
 }
 
--- a/crypto/fipsmodule/evp/internal.h
+++ b/crypto/fipsmodule/evp/internal.h
@@ -150,7 +150,7 @@ struct evp_pkey_st {
     DH *dh;
     EC_KEY *ec;
     KEM_KEY *kem_key;
-    PQDSA_KEY * pqdsa_key;
+    PQDSA_KEY *pqdsa_key;
   } pkey;
 
   // ameth contains a pointer to a method table that contains many ASN.1
--- a/include/openssl/evp.h
+++ b/include/openssl/evp.h
@@ -137,8 +137,9 @@ OPENSSL_EXPORT int EVP_PKEY_bits(const EVP_PKEY *pkey);
 // values.
 OPENSSL_EXPORT int EVP_PKEY_id(const EVP_PKEY *pkey);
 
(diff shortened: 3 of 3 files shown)


crypto/fipsmodule/evp/evp.c
```diff
@@ -293,6 +294,15 @@ int EVP_PKEY_id(const EVP_PKEY *pkey) {
   return pkey->type;
 }
 
+int EVP_PKEY_pqdsa_get_type(const EVP_PKEY *pkey) {
+  SET_DIT_AUTO_RESET;
+  if (pkey->type != EVP_PKEY_PQDSA) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_EXPECTING_A_PQDSA_KEY);
+    return 0;
+  }
+  return pkey->pkey.pqdsa_key->pqdsa->nid;
```
justsmth, line 306: Do we know that `pkey->pkey.pqdsa_key->pqdsa` is not NULL in all cases where `pkey->type == EVP_PKEY_PQDSA`?

torben-hansen: If used correctly, yeah. But fair, could be more careful.

crypto/fipsmodule/evp/evp.c
```diff
@@ -293,6 +294,15 @@ int EVP_PKEY_id(const EVP_PKEY *pkey) {
   return pkey->type;
 }
 
+int EVP_PKEY_pqdsa_get_type(const EVP_PKEY *pkey) {
+  SET_DIT_AUTO_RESET;
```
WillChilds-Klein, line 298: do we need this? i don't think we condition on any secret data/fields here.

dkostic (dismissed):

crypto/fipsmodule/evp/evp.c
torben-hansen: A bit more paranoid now. But I'm not sure a null-dereference would be reachable before unless you allocate an object and then immediately call this function.

crypto/fipsmodule/evp/evp.c
torben-hansen: AFAIR, the DIT is needed when you dereference a pointer to an object containing private data. I was following the pattern used throughout the source file

dkostic (approved):

PR #2988, opened 2026-02-12
Ensure no overflow in signed output length in do_buf

torben-hansen:
### Description of changes: 

In some cases, `outlen` would be able to overflow. But it would require the string to be very long (hundreds of mbs). This is partially avoided by `ASN1_STRING_set()` ensuring strings are less than `ASN1_STRING_MAX`. Yet, one could use `ASN1_STRING_set0()` which doesn't require that. Latter is a void function and I can't tell what might or might not be using it. Instead just check for overflow inline. Printing function anyway.

### Call-outs:

Reported by Petr Praus

### DIFF
--- a/crypto/asn1/a_strex.c
+++ b/crypto/asn1/a_strex.c
@@ -203,11 +203,21 @@ static int do_buf(const unsigned char *buf, int buflen, int encoding,
         if (len < 0) {
           return -1;
         }
+
+        // Ensure addition doesn't overflow and corrupt the signed output length
+        if (len > INT_MAX - outlen) {
+          OPENSSL_PUT_ERROR(ASN1, ASN1_R_STRING_TOO_LONG);
+          return -1;
+        }
         outlen += len;
       }
     } else {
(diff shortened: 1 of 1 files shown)


dkostic (approved):

crypto/asn1/a_strex.c
```diff
@@ -203,13 +203,23 @@ static int do_buf(const unsigned char *buf, int buflen, int encoding,
         if (len < 0) {
           return -1;
         }
+
+        // Ensure addition doesn't overflow and corrupt the signed output length
+        if (len > INT_MAX - outlen) {
```
WillChilds-Klein, line 208: does this hold up under our various endianness cocnerns?

torben-hansen: Yes, the order of the byte representation doesn't influence this kind of expression - it doesn't operate on the byte-representation.

WillChilds-Klein (approved):

PR #2990, opened 2026-02-12
Ensure public key is set before verifying through ML-DSA verify

torben-hansen:
### Description of changes: 

To verify, one would typically need the public key. It's an implicit contract. Ensure this is true for the ML-DSA verify code-path that passes through `pkey_pqdsa_verify_generic` 

### Call-outs:

Reported by Petr Praus

### DIFF
--- a/crypto/fipsmodule/evp/p_pqdsa.c
+++ b/crypto/fipsmodule/evp/p_pqdsa.c
@@ -167,7 +167,7 @@ static int pkey_pqdsa_verify_generic(EVP_PKEY_CTX *ctx, const uint8_t *sig,
 
   PQDSA_KEY *key = ctx->pkey->pkey.pqdsa_key;
   if (!key || !key->public_key) {
-    OPENSSL_PUT_ERROR(EVP, EVP_R_NO_PARAMETERS_SET);
+    OPENSSL_PUT_ERROR(EVP, EVP_R_NO_KEY_SET);
     return 0;
   }
 


crypto/fipsmodule/evp/p_pqdsa.c
```diff
@@ -166,6 +166,10 @@ static int pkey_pqdsa_verify_generic(EVP_PKEY_CTX *ctx, const uint8_t *sig,
   }
 
   PQDSA_KEY *key = ctx->pkey->pkey.pqdsa_key;
+  if (key->public_key == NULL) {
```
justsmth: Do we need one more defensive check?
```
if (key == NULL || key->public_key == NULL) {
```

torben-hansen: In practice no because of `EVP_PKEY_pqdsa_set_params`. But now I'm here I might as well. Did the same for the private key code path as well.

crypto/fipsmodule/evp/p_pqdsa.c
```diff
@@ -166,6 +166,10 @@ static int pkey_pqdsa_verify_generic(EVP_PKEY_CTX *ctx, const uint8_t *sig,
   }
 
   PQDSA_KEY *key = ctx->pkey->pkey.pqdsa_key;
+  if (!key || !key->public_key) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_NO_PARAMETERS_SET);
```
WillChilds-Klein: nit: no parameters or no key?

WillChilds-Klein (dismissed):

crypto/fipsmodule/evp/p_pqdsa.c
torben-hansen: True, it should be that error not and not the one I chose. Burning a bit more CI compute to change that :)

justsmth (approved):

WillChilds-Klein (approved):

PR #2994, opened 2026-02-13
Cache peer CA names on client side after handshake

WillChilds-Klein:
# Notes

OpenSSL [docs](https://docs.openssl.org/1.0.2/man3/SSL_get_client_CA_list/#description) for SSL_get_client_CA_list stipulate:

> In client mode, SSL_get_client_CA_list returns the list of client CAs sent from the server, if any.

AWS-LC’s `SSL_get_client_CA_list` [behavior](https://github.com/aws/aws-lc/blob/3d314dc79a07ca3cc43d1512f7b99d690c472bad/ssl/ssl_x509.cc#L1350-L1351) differs from OpenSSL’s [behavior](https://github.com/openssl/openssl/blob/bf39120c27047683c90f675ab1f1fbfa286cc026/ssl/ssl_cert.c#L677-L678)  on the client side in that AWS-LC keeps its client CA list on the `ssl->s3->hs` struct that gets [reset](https://github.com/aws/aws-lc/blob/3d314dc79a07ca3cc43d1512f7b99d690c472bad/ssl/ssl_lib.cc#L929) upon handshake completion whereas OpenSSL keeps its list on `sc->s3.tmp` which appears to retain state after the handshake completes.

This PR moves `cached_x509_ca_names` from the handshake (`ssl->hs`) struct to the `ssl->s3` struct to persist them after the handshake completes and `ssl->hs` is reset.

# Testing
- unit test
- pyopenssl CI job passes

---


ssl/ssl_test.cc
```diff
@@ -1396,6 +1396,87 @@ TEST(SSLTest, ClientCABuffers) {
   EXPECT_TRUE(cert_cb_called);
 }
 
+// Test that |SSL_get_client_CA_list| returns the server's CA list on the
+// client both during the cert callback and after the handshake completes.
+TEST(SSLTest, PeerCANamesX509DuringAndAfterHandshake) {
```
justsmth, line 1401: Add case verifying when the server does **not** request client authentication (no `CertificateRequest`)?

ssl/ssl_lib.cc
```diff
@@ -926,6 +926,16 @@ int SSL_do_handshake(SSL *ssl) {
...
+    // On the client, persist the CA names received in the CertificateRequest
+    // message so that |SSL_get_client_CA_list| can return them after the
+    // handshake. This eagerly converts to X509_NAMEs since the raw
+    // CRYPTO_BUFFERs in |hs->ca_names| will be destroyed with |hs|.
+    if (!ssl->server && ssl->s3->hs->ca_names &&
+        ssl->ctx->x509_method == &ssl_crypto_x509_method) {
+      // Failure is non-fatal: the handshake has already completed, so
+      // |SSL_get_client_CA_list| will simply return NULL post-handshake.
+      ssl_x509_persist_peer_ca_names(ssl);
+    }
```
justsmth, line 942: If a TLS 1.2 renegotiation occurs where the server does *not* send a `CertificateRequest`, `hs->ca_names` will be NULL, so the persist is skipped and `s3->cached_x509_peer_ca_names` retains stale data from the previous handshake:
```
  if (!early_return) {
    if (!ssl->server) {
      if (ssl->s3->hs->ca_names &&
          ssl->ctx->x509_method == &ssl_crypto_x509_method) {
        ssl_x509_persist_peer_ca_names(ssl);
      } else {
        // No CertificateRequest in this handshake — clear any stale peer
        // CA names left over from a previous handshake.
        ssl->ctx->x509_method->hs_flush_cached_ca_names(ssl->s3->hs.get());
      }
    }
    ssl->s3->hs.reset();
    ssl_maybe_shed_handshake_config(ssl);
  }
```

WillChilds-Klein: good catch, added.

PR #2999, opened 2026-02-16
Generate Rust Bindings

justsmth:
### Issues:
* Addresses: V2090958423

### Description of changes:

This adds CMake support for generating Rust FFI bindings from the AWS-LC public C headers using `bindgen-cli`. This is intended for use by downstream Rust consumers so that bindings can be generated directly from the CMake build system.

**New CMake option:** `-DGENERATE_RUST_BINDINGS=ON` (off by default)

When enabled, the build:
1. Validates that `bindgen-cli` (≥0.69.5) and `rustfmt` are available
2. Auto-discovers all public OpenSSL headers and generates a `rust_wrapper.h` aggregation header
3. Invokes `bindgen` to produce `aws_lc_bindings.rs` in the build directory
4. Supports the existing `BORINGSSL_PREFIX` mechanism — uses `--prefix-link-name` so Rust-side function names stay unprefixed while `#[link_name]` attributes carry the prefix for the linker
5. During install, the bindings are placed at `$INSTALL_DIR/share/rust/aws_lc_bindings.rs`

**Files added:**
- `cmake/rust_bindings.cmake` — Module containing the header discovery, prefix formatting, and bindgen invocation logic
- `cmake/rust_wrapper.h.in` — Template header that aggregates all public headers for bindgen input

### Call-outs:

- The prefix symbols header (`boringssl_prefix_symbols.h`) is intentionally excluded from bindgen's input. The rationale is documented in the module-level comment in `cmake/rust_bindings.cmake`.
- The `rust_bindings` target is not part of `ALL`. It is built on-demand when explicitly requested (`--target rust_bindings`) or automatically during `cmake --install`.
- Platform-specific symbol prefix formatting (Apple/Win32 leading underscore) is handled by `get_symbol_prefix_format()`.
 
### Testing:

CI workflow added in `.github/workflows/aws-lc-rs.yml` (`cmake-rust-bindings` job) covering:
- Linux (no prefix)
- Linux with `BORINGSSL_PREFIX=AWSLC`
- macOS (no prefix)
- macOS with `BORINGSSL_PREFIX=AWSLC`
- Windows (no prefix)

Each matrix entry verifies that bindings are generated, contain expected symbols, have correct prefix behavior, and compile as valid Rust.


.github/workflows/aws-lc-rs.yml
```diff
@@ -221,3 +221,126 @@ jobs:
...
           cargo test -p aws-lc-sys --target x86_64-pc-windows-msvc
           cargo test -p aws-lc-rs --target x86_64-pc-windows-msvc
+
+  # CMake Rust bindings generation tests
+  cmake-rust-bindings:
+    if: github.repository_owner == 'aws'
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
```
WillChilds-Klein, line 231: do we want to build/test the libssl bindings here as well? i don't think this PR covers them in CI.

.github/workflows/aws-lc-rs.yml
```diff
@@ -221,3 +221,126 @@ jobs:
...
+            echo "ERROR: Rust bindings file was not generated"
+            exit 1
+          fi
+          echo "Generated bindings file size: $(wc -c < build/rust/aws_lc_bindings.rs) bytes"
+          echo "Generated bindings line count: $(wc -l < build/rust/aws_lc_bindings.rs) lines"
+
+      - name: Verify bindings content
+        shell: bash
+        run: |
+          # All variants should generate SSL bindings
```
WillChilds-Klein: why? `ARG_INCLUDE_SSL` isn't specified by default, is it?

if it is specified by default, should we have a test case that excludes libssl bindings by setting `ARG_INCLUDE_SSL=0`?

.github/workflows/aws-lc-rs.yml
```diff
@@ -221,3 +221,126 @@ jobs:
...
+          EOF
+          cat > "${RUNNER_TEMP}/test-bindings/src/lib.rs" << 'EOF'
+          #![allow(clippy::all)]
+          #![allow(non_upper_case_globals)]
+          #![allow(non_camel_case_types)]
+          #![allow(non_snake_case)]
+          #![allow(dead_code)]
+          #![allow(improper_ctypes)]
+          #![allow(unpredictable_function_pointer_comparisons)]
+          include!(concat!(env!("BINDINGS_PATH"), "/aws_lc_bindings.rs"));
```
WillChilds-Klein, line 386: will these `/` path separators work on windows?

.github/workflows/aws-lc-rs.yml
```diff
@@ -221,3 +221,126 @@ jobs:
...
+            # Non-prefix builds should not have link_name attributes
+            if grep -q '#\[link_name' build/rust/aws_lc_bindings.rs; then
+              echo "ERROR: Unexpected link_name attributes found in no-prefix build"
+              exit 1
+            fi
+          fi
+          echo "Bindings content verification passed"
+
+
+      - name: Verify bindings compile
```
WillChilds-Klein: `cargo check` seems necessary, but is it sufficient? i wonder if we should also have a step that actually links (some subset of) the bindings against a test program and, if linking dynamically, executes the program.

.github/workflows/aws-lc-rs.yml
justsmth: `BUILD_LIBSSL` defaults to `ON`, so the bindings produced should also cover the libssl functions. We'll make some updates to the testing to provide better coverage for this.

.github/workflows/aws-lc-rs.yml
justsmth: Yeah, the `include!` macro supports forward slashes being used on Windows.

.github/workflows/aws-lc-rs.yml
justsmth: Upgrading to an actual "link test" requires creating a `build.rs` in the test project that tells cargo where to find `libcrypto`/`libssl`, then doing `cargo build` (or `cargo test` with a small smoke test).

I think we can make this happen. I'll update the test.

.github/workflows/aws-lc-rs.yml
```diff
@@ -221,3 +221,181 @@ jobs:
...
           cargo test -p aws-lc-rs --target x86_64-pc-windows-msvc
+
+  # CMake Rust bindings generation tests
+  cmake-rust-bindings:
+    if: github.repository_owner == 'aws'
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
```
samuel40791765, line 232: Nit: Seeing that this is part of a work to flip everything internally, do we need to test on AL2023 or AL2?

PR #3000, opened 2026-02-17
Address Reported Bug Findings

skmcgrail:
This PR contains 6 bug fixes for various issues in AWS-LC. We would like to thank Joshua Rogers (https://joshua.hu/) of AISLE Research Team (https://aisle.com/) for reporting these issues.

#### 1. evp: fix DSA keygen error-path UAF/double-free
`pkey_dsa_keygen()` assigns a freshly allocated DSA into `EVP_PKEY` before `EVP_PKEY_copy_parameters()`, then frees the DSA on failure, leaving `pkey` holding a dangling pointer. This fix avoids freeing after ownership transfer by nulling the local pointer and clearing/freeing the pkey-held key on error.

#### 2. 1-byte OOB read in `EVP_PKEY_asn1_find_str` length calculation
Fixes an off-by-one out-of-bounds read in the `EVP_PKEY_asn1_find_str` length calculation.

#### 3. pkcs8: cap ciphertext length before allocating in `pkcs8_pbe_decrypt`
`pkcs8_pbe_decrypt()` allocates `OPENSSL_malloc(in_len)` where `in_len` is influenced by ASN.1 OCTET STRING lengths. The existing `INT_MAX` check only guards `EVP_DecryptUpdate`'s int parameter and occurs after the allocation. This fix caps the ciphertext length before allocating.

#### 4. evp: disable `EVP_PKEY_derive` for KEM method
`EVP_PKEY_kem_pkey_meth` incorrectly assigned `out->derive` to `pkey_hkdf_derive`, causing `ctx->data` type confusion (`KEM_PKEY_CTX` vs `HKDF_PKEY_CTX`) when callers invoke `EVP_PKEY_derive` on a KEM context. This fix sets `out->derive` to `NULL` so the operation fails as unsupported.

#### 5. Reject zero-sized digests in HKDF EVP_PKEY
Rejects digests with `EVP_MD_size <= 0` when configuring HKDF and when deriving, preventing divide-by-zero in `HKDF_expand` if an XOF digest is selected.

#### 6. Reject XOF digests in `DH_compute_key_hashed`
`DH_compute_key_hashed` passes an uninitialized `out_len` to `EVP_Digest`. For XOF digests, `EVP_DigestFinalXOF` interprets `*out_size` as an input length, which can cause an out-of-bounds write to the caller-provided buffer. This fix rejects XOF digests and initializes `out_len` to the fixed digest size.

### DIFF
--- a/crypto/dh_extra/dh_test.cc
+++ b/crypto/dh_extra/dh_test.cc
@@ -67,6 +67,7 @@
 #include <openssl/bytestring.h>
 #include <openssl/crypto.h>
 #include <openssl/dh.h>
+#include <openssl/digest.h>
 #include <openssl/err.h>
 #include <openssl/mem.h>
 #include <openssl/nid.h>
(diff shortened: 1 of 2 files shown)


crypto/evp_extra/p_dsa.c
```diff
@@ -58,19 +58,20 @@ static void pkey_dsa_cleanup(EVP_PKEY_CTX *ctx) {
...
   }
-  ret = DSA_generate_key(pkey->pkey.dsa);
 
-err:
-  if (ret != 1) {
-    OPENSSL_free(dsa);
+  // |pkey| now has ownership of DSA, and EVP_PKEY_free will handle from this
+  // point on.
+
+  if (ctx->pkey == NULL || !EVP_PKEY_copy_parameters(pkey, ctx->pkey)) {
```
torben-hansen: `ctx->pkey == NULL` is actually checked at the start of the function `GUARD_PTR(ctx->pkey);`.

crypto/evp_extra/internal.h
```diff
@@ -12,6 +12,8 @@
 #define PKCS8_VERSION_ONE 0
 #define PKCS8_VERSION_TWO 1
 
+#define MAX_PEM_STR_LEN ((size_t)10) // "ED25519ph"
```
torben-hansen, line 15: `ED25519ph` is only 9 actually. `DILITHIUM3` is gone.

torben-hansen: Does it matter whether it's the exact size or just some reasonable upper bound?

skmcgrail: 10 is so that the null-terminator is included.

skmcgrail: The function that the pem_str is used with stops at the null-terminator value. So should be reasonable upper-bound, assuming all `ameth->pem_str` values are null-terminated. Each one varies in length given the algorithm name, but shorter strings that are null terminated (as they are in this case) would trigger `OPENSSL_strnlen` to stop.

torben-hansen (approved):

dkostic (approved):

PR #3007, opened 2026-02-18
Fix cmake compiler spec

Ra2-IFV:
### Description of changes: 

- cmakelists: if condition statements for CLANG and GCC specific flags
- github: issue templates

### Testing:

`cmake -B ...` succeed

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,14 +1,18 @@
 ### Issues:
-Resolves #ISSUE-NUMBER1
-Addresses #ISSUE-NUMBER2
+
+<!-- Resolves #ISSUE-NUMBER1 -->
+<!-- Addresses #ISSUE-NUMBER2 -->
 
 ### Description of changes: 
-Describe AWS-LC’s current behavior and how your code changes that behavior. If there are no issues this pr is resolving, explain why this change is necessary.
+
+<!-- Describe AWS-LC’s current behavior and how your code changes that behavior. If there are no issues this pr is resolving, explain why this change is necessary. -->
 
 ### Call-outs:
-Point out areas that need special attention or support during the review process. Discuss architecture or design changes.
+
+<!-- Point out areas that need special attention or support during the review process. Discuss architecture or design changes. -->
 
 ### Testing:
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,14 +1,18 @@
...
+
+<!-- Point out areas that need special attention or support during the review process. Discuss architecture or design changes. -->
 
 ### Testing:
-How is this change tested (unit tests, fuzz tests, etc.)? Are there any testing steps to be verified by the reviewer?
 
-By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.
+<!-- How is this change tested (unit tests, fuzz tests, etc.)? Are there any testing steps to be verified by the reviewer? -->
+
+- [] By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.
```
justsmth, line 18: Thanks for the PR!

I'd prefer to keep this line as a plain statement rather than adding a checkbox to it. Because this text is a positive indication that the contributor affirms the licensing, its presence in the PR description is currently sufficient and we have CI that verifies it's there.

Changing this to a checkbox introduces ambiguity: a contributor might leave the box unchecked intentionally (indicating a rejection of the statement) or incidentally (not realizing the box must be "checked" to affirm the statement). The current setup is simpler -- presence of the statement *is* the affirmation, with no opt-in/opt-out to parse.

CMakeLists.txt
```diff
@@ -505,9 +505,14 @@ if(GCC OR CLANG)
       endif()
       set(CMAKE_ASM_FLAGS "${CMAKE_ASM_FLAGS} -Wno-newline-eof")
     endif()
+    if (CLANG)
+      set(C_CXX_FLAGS "${C_CXX_FLAGS} -Wno-c11-extensions")
+    elseif (GCC)
+      set(C_CXX_FLAGS "${C_CXX_FLAGS} -Wno-c++11-extensions")
+    endif()
```
justsmth, line 512: * This is for `C_CXX_FLAGS`, which are the flags applied for both C and C++ source file compilation.
* Why differentiate the two compilers? Clang seems to also support [-Wc++11-extensions](https://clang.llvm.org/docs/DiagnosticsReference.html#wc-11-extensions).

PR #3008, opened 2026-02-18
Fix OPENSSL_memchr per C23

### LINKED ISSUE #2995: Fails to build on Arch Linux without -Wno-discarded-qualifiers
### Problem:

Fail to build on Arch Linux without `-Wno-discarded-qualifiers`.

With the extra cflag above it can build successfully. But I'm not sure if I should suppress the warning blindly.

A rough Googling shows that `-Wno-discarded-qualifiers` was introduced in GCC 5 which is released roughly 10 years ago. I believe that ruled out Arch Linux being too bleeding edge... Or maybe it is?

#### Relevant details
AWS-LC commit: https://github.com/aws/aws-lc/commit/728811eecec794802c78105be6dbfe9d79870ac7 tagged v1.66.2

System information:
 * CPU architecture: x86-64
 * CPU name: AMD Ryzen 7 5800X
 * OS: Arch Linux
 * C Compiler: gcc 15.2.1 but I believe Arch Linux ships commit=`0b99615a8aef011cff76c6caa8c09434f46598b3`, not a tagged release. See [gcc package in Arch](https://gitlab.archlinux.org/archlinux/packaging/packages/gcc/-/blob/main/PKGBUILD?ref_type=heads)

Build log: 
See the gist: https://gist.github.com/5long/be83f0b1ac35b88129ca2dd3228c0c1a

justsmth:
### Issues:
Resolves #2995

### Description of changes: 
C23 makes `memchr` const-correct, returning `const void *` when the input is `const void *`. Some C libraries (e.g. glibc 2.41+) apply this change even in C11 mode. This causes a `-Wdiscarded-qualifiers` error in the C code path of `OPENSSL_memchr`, which returns `void *`. Combined with `-Werror`, this breaks the build on distributions like Arch Linux that ship these newer toolchains.

Also guard the `-Wno-c11-extensions` flag behind a Clang check, as it is Clang-specific and produces a diagnostic note on GCC.


CMakeLists.txt
```diff
@@ -507,7 +507,10 @@ if(GCC OR CLANG)
     endif()
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wunused -Wcomment -Wchar-subscripts -Wuninitialized -Wshadow")
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wwrite-strings -Wformat-security -Wunused-result -Wno-overlength-strings")
-    set(C_CXX_FLAGS "${C_CXX_FLAGS} -Wno-c11-extensions -Wvla -Wtype-limits -Wno-unused-parameter")
+    set(C_CXX_FLAGS "${C_CXX_FLAGS} -Wvla -Wtype-limits -Wno-unused-parameter")
+    if(CLANG)
```
WillChilds-Klein, line 511: nit: do we also want to guard this by clang version?

justsmth: The `-Wno-c11-extensions` has been present unconditionally and hasn't caused any build failures w/ Clang compilers. I don't think it needs to be more restricted than this.

PR #3009, opened 2026-02-18
Add ACVP Support for KTS-IFC

nhatnghiho:
### Issues:
Addresses P355857148

### Description of changes: 
This PR adds ACVP support for `KTS-IFC KTS-OAEP-basic`

### Call-outs:
Due to the random nature of RSA-OAEP encryption, no expected vector was added. To review, please run the script locally to verify the result.

### Testing:
Run `check_expected.go` on the new KTS-IFC test vector.

### DIFF
--- a/util/fipstools/acvp/modulewrapper/modulewrapper.cc
+++ b/util/fipstools/acvp/modulewrapper/modulewrapper.cc
@@ -2810,11 +2810,23 @@ static bool RSAOAEPEncrypt(const Span<const uint8_t> args[],
 
   BIGNUM *n = BN_new();
   BIGNUM *e = BN_new();
-  bssl::UniquePtr<RSA> key(RSA_new());
+  bssl::UniquePtr<RSA> rsa(RSA_new());
 
   if (!BN_bin2bn(n_bytes.data(), n_bytes.size(), n) ||
       !BN_bin2bn(e_bytes.data(), e_bytes.size(), e) ||
-      !RSA_set0_key(key.get(), n, e, nullptr)) {
+      !RSA_set0_key(rsa.get(), n, e, nullptr)) {
+    return false;
+  }
+
+  bssl::UniquePtr<EVP_PKEY> pkey(EVP_PKEY_new());
+  if (!EVP_PKEY_set1_RSA(pkey.get(), rsa.get())) {
+    return false;
+  }
+
+  bssl::UniquePtr<EVP_PKEY_CTX> ctx(EVP_PKEY_CTX_new(pkey.get(), nullptr));
+  if (!ctx || !EVP_PKEY_encrypt_init(ctx.get()) ||
+      !EVP_PKEY_CTX_set_rsa_padding(ctx.get(), RSA_PKCS1_OAEP_PADDING) ||
+      !EVP_PKEY_CTX_set_rsa_oaep_md(ctx.get(), MDFunc())) {
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


util/fipstools/acvp/modulewrapper/modulewrapper.cc
```diff
@@ -2779,6 +2796,71 @@ static bool RSASigVer(const Span<const uint8_t> args[],
...
+
+  uint32_t dkm_len = 0;
+  memcpy(&dkm_len, dkm_len_bytes.data(), sizeof(dkm_len));
+
+  BIGNUM *n = BN_new();
+  BIGNUM *e = BN_new();
+
+  bssl::UniquePtr<RSA> key(RSA_new());
+
+  if (Encrypt) {
```
sgmenda: nit: Could we split this into two functions: one for Encrypt and one for Decrypt?

util/fipstools/acvp/modulewrapper/modulewrapper.cc
sgmenda, line 3755: re above nit: it is also confusing because the second bool template is used for different things. It seems to be usePSS here?

nhatnghiho: I think the template can be customized for each function, so the usage of the boolean in `RSASigVer` and `RSA_OAEP` can be as different as we want it to be. But yes, I agree splitting `RSA_OAEP` and remove the `Encrypt` bool altogether makes things cleaner :)

util/fipstools/acvp/modulewrapper/modulewrapper.cc
```diff
@@ -3674,6 +3765,8 @@ static struct {
     {"KDA/OneStep/HMAC-SHA2-512", 4, SSKDF_HMAC<EVP_sha512>},
     {"KDA/OneStep/HMAC-SHA2-512/224", 4, SSKDF_HMAC<EVP_sha512_224>},
     {"KDA/OneStep/HMAC-SHA2-512/256", 4, SSKDF_HMAC<EVP_sha512_256>},
+    {"KTS/OAEP/SHA2-256/transport", 3, RSAOAEPEncrypt<EVP_sha256>},
+    {"KTS/OAEP/SHA-1/receive", 6, RSAOAEPDecrypt<EVP_sha1>},
```
skmcgrail: You need to register more then just these variants right?

You request vectors for initiator and responder roles and list sha1, sha2-224, sha2-256, sha2-384, and sha2-512 when you register the capabilities with the ACVP server. But then you aren't registering in the modulewrapper support for all of those algorithms.

skmcgrail: I think it would also be less confusing if you align and use `initiator` and `responder` perhaps rather then translating them to transport and receive respectively? That way it maps 1-1 with the ACVP terminology.

util/fipstools/acvp/acvptool/subprocess/kts.go
```diff
@@ -0,0 +1,132 @@
+// Copyright (c) 2020, Google Inc.
```
skmcgrail: This should be our own copyright header since this is new development.

util/fipstools/acvp/modulewrapper/modulewrapper.cc
nhatnghiho: > You need to register more then just these variants right?

The KTS-IFC test vector that we received only need these 2, but I suppose there's no harm in adding more

util/fipstools/acvp/acvptool/subprocess/kts.go
```diff
@@ -0,0 +1,132 @@
...
+			ID: group.ID,
+		}
+
+		if group.Type != "AFT" {
+			return nil, fmt.Errorf("unknown test type %q", group.Scheme)
+		}
+
+		switch group.Role {
+		case "initiator", "responder":
+			break
```
skmcgrail: Same comment from the other PR no need to specify `break`

util/fipstools/acvp/acvptool/subprocess/kts.go
```diff
@@ -0,0 +1,132 @@
...
+func (k *kts) Process(vectorSet []byte, m Transactable) (interface{}, error) {
+	var parsed ktsVectorSet
+	if err := json.Unmarshal(vectorSet, &parsed); err != nil {
+		return nil, err
+	}
+
+	// See https://pages.nist.gov/ACVP/draft-hammett-acvp-kas-ifc.html
+	var ret []ktsTestGroupResponse
+	for _, group := range parsed.Groups {
+		group := group
```
skmcgrail:
```suggestion
```

You shouldn't need to do this.

PR #3011, opened 2026-02-19
Fix CI: gcc-4.8

justsmth:
### Description of changes:

The `gcc-4_8` CI job was failing because the `gcc:4.8.5` Docker Hub image can no longer be pulled (fails Docker image manifest validation). This updates the Dockerfile to use `ubuntu:18.04` as the base image and installs `gcc-4.8`/`g++-4.8` from apt. The manually downloaded CMake 3.6.3 binary is also replaced with the apt-provided CMake (3.10.2), as the pre-built binary had a runtime dependency (`libidn.so.11`) not present in the new base image.

### Call-outs:

Uses `mirrors.edge.kernel.org` instead of the default Ubuntu archive mirrors for reliable access to EOL Ubuntu 18.04 packages (same approach used in [aws-lc-rs](https://github.com/aws/aws-lc-rs/blob/main/.github/workflows/compilers.yml#L196-L197)).

### DIFF
--- a/.github/docker_images/gcc-4.8/Dockerfile
+++ b/.github/docker_images/gcc-4.8/Dockerfile
@@ -1,24 +1,37 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0 OR ISC
 
-FROM gcc:4.8.5
+FROM ubuntu:18.04
 
 VOLUME ["/awslc"]
+
+# Use kernel.org mirror for more reliable package access
+RUN sed -i 's|archive.ubuntu.com|mirrors.edge.kernel.org|g' /etc/apt/sources.list && \
+    sed -i 's|security.ubuntu.com|mirrors.edge.kernel.org|g' /etc/apt/sources.list
+
+RUN apt-get update && \
+    apt-get install -y \
+      ca-certificates \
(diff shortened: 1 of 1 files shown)


.github/docker_images/gcc-4.8/Dockerfile
```diff
@@ -1,24 +1,37 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0 OR ISC
 
-FROM gcc:4.8.5
+FROM ubuntu:18.04
```
WillChilds-Klein, line 4: looks like 18.04 is [EOL](https://en.wikipedia.org/wiki/Ubuntu_version_history)

> Ubuntu 18.04 LTS had normal LTS support for five years until May 2023

will GCC 4.8 work on a newer LTS?

WillChilds-Klein (approved):

.github/docker_images/gcc-4.8/Dockerfile
justsmth: > will GCC 4.8 work on a newer LTS?

I tried using a newer Ubuntu LTS when resolving a similar issue in aws-lc-rs, and had a lot of trouble finding a compatible gcc-4.8 package to install.

PR #3024, opened 2026-02-20
Various Small Additions to ACVP Tool

nhatnghiho:
### Issues:
Addresses P355857148

### Description of changes: 
This PR adds ACVP support for:
- `RSA signaturePrimitive`
- `RSA decryptionPrimitive`
- `KAS-ECC-SSC onePassDh`
- `ECDSA sigGen componentTest`

### Call-outs:
Due to the random nature of ECDH and ECDSA sigGen, no expected vector was added for these 2. To review, please run the script locally to verify the result.

### Testing:
Run `check_expected.go` on the new test vectors.


util/fipstools/acvp/modulewrapper/modulewrapper.cc
```diff
@@ -2779,6 +2836,133 @@ static bool RSASigVer(const Span<const uint8_t> args[],
...
+  }
+
+  std::vector<uint8_t> sig(RSA_size(key.get()));
+  size_t sig_len;
+  uint8_t success_flag[1] = {0};
+  if (!RSA_sign_raw(key.get(), &sig_len, sig.data(), sig.size(), msg.data(),
+                    msg.size(), RSA_NO_PADDING)) {
+    return write_reply(
+        {Span<const uint8_t>(success_flag), Span<const uint8_t>()});
+  }
```
justsmth, line 2984: `ERR_clear_error()` here as well?

util/fipstools/acvp/modulewrapper/modulewrapper.cc
```diff
@@ -2779,6 +2836,133 @@ static bool RSASigVer(const Span<const uint8_t> args[],
...
+static bool RSASignaturePrimitive(const Span<const uint8_t> args[],
+                                  ReplyCallback write_reply) {
+  const Span<const uint8_t> d_bytes = args[0];
+  const Span<const uint8_t> n_bytes = args[1];
+  const Span<const uint8_t> e_bytes = args[2];
+  const Span<const uint8_t> msg = args[3];
+
+  BIGNUM *d = BN_new();
+  BIGNUM *n = BN_new();
+  BIGNUM *e = BN_new();
```
justsmth, line 2967: This should use `bssl::UniquePtr<BIGNUM>` to avoid leaks and be consistent with others.

util/fipstools/acvp/modulewrapper/modulewrapper.cc
```diff
@@ -2779,6 +2836,133 @@ static bool RSASigVer(const Span<const uint8_t> args[],
...
+  bssl::UniquePtr<RSA> key(RSA_new());
+  if (!BN_bin2bn(n_bytes.data(), n_bytes.size(), n) ||
+      !BN_bin2bn(e_bytes.data(), e_bytes.size(), e) ||
+      !BN_bin2bn(d_bytes.data(), d_bytes.size(), d) ||
+      !RSA_set0_key(key.get(), n, e, d)) {
+    return false;
+  }
```
justsmth, line 2974: Should we add `|| !RSA_check_key(key.get())` to this condition, similar to what's in `RSADecryptionPrimitive`?

util/fipstools/acvp/modulewrapper/modulewrapper.cc
nhatnghiho: These values will be then used in `bssl::UniquePtr<RSA>` RSA key so they will be taken care there I believe.

util/fipstools/acvp/modulewrapper/modulewrapper.cc
nhatnghiho: It's not obvious but `RSA_new_private_key_large_e` runs `RSA_check_key` internally ([code](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/rsa/rsa.c#L203)). `RSADecryptionPrimitive` uses a public key instead. But we don't have a corresponding`RSA_new_public_key_large_e` so hence this manual check.

PR #3032, opened 2026-02-24
Fix issues in `pass_util.cc` password handling

justsmth:
### Description of changes:
Addresses multiple issues in `tool-openssl/pass_util.cc` and the `Password` class:
- Ensured all temporary buffers holding passwords are properly cleansed, including the `buf` stack buffer in `ExtractPasswordFromStream` and a `std::string` replaced with `Password` for secure clearing.
- Fixed `fd:` parsing to use `strtoul` instead of `atoi`, avoiding undefined behavior on overflow.
- Removed incorrect rejection of colons in `pass:` passwords, matching OpenSSL behavior.
- Fixed off-by-one in password length limits for `pass:` and `env:` sources.
- Minor hardening: fixed `same_file` initialization, replaced unsigned subtraction with overflow-safe comparison in `ExtractDirectPassword`.

### Call-outs:
- Behavioral change: `pass:a:b` was previously rejected and now succeeds with password `a:b`.

### Testing:
All existing `PassUtilTest` tests pass. Updated `DirectPasswordEdgeCases` to cover passwords containing colons.

### DIFF
--- a/tool-openssl/pass_util.cc
+++ b/tool-openssl/pass_util.cc
@@ -80,7 +80,7 @@ static bool ValidateSource(Password &passin, Password *passout = nullptr,
 
 static bool ExtractDirectPassword(Password &source) {
   // Check length before modification
-  if (source.get().length() - 5 >= PEM_BUFSIZE) {
+  if (source.get().length() >= 5 + PEM_BUFSIZE) {
     fprintf(stderr, "Password exceeds maximum allowed length (%d bytes)\n",
             PEM_BUFSIZE - 1);
     return false;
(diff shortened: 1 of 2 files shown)


tool-openssl/pass_util.cc
```diff
@@ -213,7 +210,6 @@ static bool ExtractPasswordFromStream(Password &source, Source source_type,
     }
   }
 
-  OPENSSL_cleanse(buf, sizeof(buf));
```
WillChilds-Klein, line 216: we don't still want to cleanse here?

justsmth: `buf` is only used within the `read_password_line` lambda, which I think cleanses it on all of its `return` paths.

PR #3034, opened 2026-02-24
Improve type safety and bounds checking in EVP cipher ctrl handlers

justsmth:
### Description of changes:
Fixes minor correctness and safety issues in the AES-GCM and AES-CCM EVP cipher ctrl handlers:

- **`e_aes.c` (`EVP_CTRL_GCM_SET_IV_INV`):** Added bounds validation for `arg` (`arg <= 0 || arg > gctx->ivlen`). The companion handler `EVP_CTRL_GCM_IV_GEN` already validates `arg` before the same `gctx->iv + gctx->ivlen - arg` pointer arithmetic, but `SET_IV_INV` did not, allowing potential out-of-bounds access from a misuse of the API.
- **`e_aesccm.c`** 
  - **`EVP_CTRL_GET_IVLEN`:** The handler wrote through a `uint32_t *` cast, but the caller in `EVP_CIPHER_CTX_iv_length` passes an `int *`. This is a strict aliasing violation. Changed to `*(int *)ptr` to match the caller and the GCM handler's convention.
  - **`CCM_L_TO_NONCE_LEN` macro:** Added parentheses around the macro parameter to follow standard C macro hygiene.

### Call-outs:
The `uint32_t *` vs `int *` mismatch is undefined behavior under C strict aliasing rules, though it is unlikely to cause issues in practice since both types are 32 bits on almost all platforms. The other two changes are defensive hardening.

### Testing:
Existing tests cover these code paths. No behavioral change is expected for valid inputs.


crypto/fipsmodule/cipher/e_aesccm.c
```diff
@@ -665,7 +665,7 @@ static int cipher_aes_ccm_ctrl(EVP_CIPHER_CTX *ctx, int type, int arg,
       cipher_ctx->message_len = 0;
       return 1;
     case EVP_CTRL_GET_IVLEN:
-      *(uint32_t *)ptr = CCM_L_TO_NONCE_LEN(cipher_ctx->L);
```
torben-hansen, line 668: This is generally safe AFAIK unless uint32_t has a weird definition(?). Anyway, I suppose most callers will actually use an int.

justsmth: Yeah, it's called from `EVP_CIPHER_CTX_iv_length` which provides an `int *`.

PR #3036, opened 2026-02-24
Fix PKCS7 verify content memleak

WillChilds-Klein:
# Notes

Previously we failed to free the whole BIO chain in PKCS7_verify, instead just popping the head. To remediate, we need to iterate through the chain until we hit `indata` (owned by the caller), freeing as we go.

# Testing
- new test triggering the memory leak, confirmed fix.

---

### DIFF
--- a/crypto/pkcs7/pkcs7.c
+++ b/crypto/pkcs7/pkcs7.c
@@ -832,8 +832,8 @@ BIO *PKCS7_dataInit(PKCS7 *p7, BIO *bio) {
     }
     BIO_set_mem_eof_return(bio, /*eof_value*/ 0);
     if (!PKCS7_is_detached(p7) && content && content->length > 0) {
-      // |bio |needs a copy of |os->data| instead of a pointer because the data
-      // will be used after |os |has been freed
+      // |bio| needs a copy of |content->data| instead of a pointer because the
+      // data may be used after |content| has been freed
       if (BIO_write(bio, content->data, content->length) != content->length) {
         goto err;
       }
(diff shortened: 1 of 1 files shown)


crypto/pkcs7/pkcs7_test.cc
```diff
@@ -2040,6 +2040,101 @@ TEST(PKCS7Test, TestSigned) {
...
+  bssl::UniquePtr<EVP_PKEY> root_pkey(EVP_PKEY_new());
+  ASSERT_TRUE(EVP_PKEY_set1_RSA(root_pkey.get(), root_rsa.get()));
+
+  bssl::UniquePtr<RSA> leaf1_rsa(RSA_new());
+  ASSERT_TRUE(RSA_generate_key_fips(leaf1_rsa.get(), 2048, nullptr));
+  bssl::UniquePtr<EVP_PKEY> leaf1_pkey(EVP_PKEY_new());
+  ASSERT_TRUE(EVP_PKEY_set1_RSA(leaf1_pkey.get(), leaf1_rsa.get()));
+
+  bssl::UniquePtr<RSA> leaf2_rsa(RSA_new());
+  ASSERT_TRUE(RSA_generate_key_fips(leaf2_rsa.get(), 2048, nullptr));
```
samuel40791765: Nit: Can we use a different key like ECDSA here instead? Just slightly wary about RSA key generation being slow and increasing test time. Or we could other existing certs in our tests that have different digest algorithm types and cut down on the amount of code to generate and sign here?

WillChilds-Klein: sure

samuel40791765 (approved):

PR #3037, opened 2026-02-24
Simplify `d2i_PKCS7` by removing redundant BER-to-DER conversion

justsmth:
### Description of changes:
`d2i_PKCS7` previously performed a manual BER-to-DER conversion using
`CBS_asn1_ber_to_der` before passing the result to `ASN1_item_d2i`. This
manual conversion had a bug: when BER input contained indefinite-length
elements, the converted DER encoding was shorter than the original BER, but
`*in` was advanced by the DER byte count rather than the original BER byte
count. This caused `*in` to be under-advanced, violating the d2i API contract.

Since `ASN1_item_d2i` already handles BER encoding natively (including
indefinite-length elements and constructed strings), the manual conversion
was redundant. This change removes it entirely and replaces
`IMPLEMENT_ASN1_ALLOC_FUNCTIONS(PKCS7)` plus the hand-written `d2i_PKCS7`
and `i2d_PKCS7` with a single `IMPLEMENT_ASN1_FUNCTIONS(PKCS7)` macro call,
consistent with every other type defined in this file.

The now-unused `#include "../bytestring/internal.h"` has also been removed.

### Call-outs:
The existing BER test in `TestCertReparse` was asserting the old (incorrect)
pointer-advancement behavior for BER input, expecting `*in` to advance by
only the DER-converted length. This has been updated to expect full
advancement past the original input, consistent with the non-BER path.

### Testing:
All 60 existing PKCS7 tests pass (`crypto_test --gtest_filter='*PKCS7*'`).

### DIFF
--- a/crypto/pkcs7/pkcs7.c
+++ b/crypto/pkcs7/pkcs7.c
@@ -791,6 +791,7 @@ BIO *PKCS7_dataInit(PKCS7 *p7, BIO *bio) {
     }
 
     if (EVP_CipherInit_ex(ctx, evp_cipher, NULL, key, iv, /*enc*/ 1) <= 0) {
+      OPENSSL_cleanse(key, keylen);
       goto err;
     }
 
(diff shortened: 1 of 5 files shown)


crypto/pkcs7/pkcs7_asn1.c
WillChilds-Klein: now that we've eliminated any special logic, we should be able to replace these definitions with a single `IMPLEMENT_ASN1_FUNCTIONS(PKCS7)` macro call

justsmth: Great idea! PR updated.

WillChilds-Klein (approved):

samuel40791765 (dismissed):

WillChilds-Klein (approved):

samuel40791765 (approved):

PR #3039, opened 2026-02-24
Fix PKCS8_decrypt to handle all negative pass_len values

geedo0:
### Issues:
N/A

### Description of changes: 
PKCS8_decrypt only checked for pass_len_in == -1 to trigger
strlen(pass), while PKCS8_encrypt checked pass_len_in < 0.
Passing other negative values (e.g. -2) caused pass_len to wrap
to a huge size_t, resulting in an out-of-bounds read and crash.

Change the condition to < 0 to match PKCS8_encrypt and document
the negative pass_len behavior in both functions' public headers.

### Call-outs:

N/A

### Testing:

CI

### DIFF
--- a/crypto/fipsmodule/pqdsa/pqdsa.c
+++ b/crypto/fipsmodule/pqdsa/pqdsa.c
@@ -48,9 +48,9 @@ int PQDSA_KEY_init(PQDSA_KEY *key, const PQDSA *pqdsa) {
   PQDSA_KEY_clear(key);
 
   key->pqdsa = pqdsa;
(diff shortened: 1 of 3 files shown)


nebeid (commented):
Would you like to add in `crypto/pkcs8/pkcs8_test.cc` a test like `PKCS8_decrypt(encrypted.get(), password, -2)` in addition to https://github.com/geedo0/aws-lc/blob/d44f1374c19d428c56ae7dafa84dadae82bcb815/crypto/pkcs8/pkcs8_test.cc#L204

nebeid (approved):

PR #3043, opened 2026-02-25
Add INT_MAX bounds check before EVP_CipherUpdate in PKCS8/PKCS12 encryption

geedo0:
### Issues:
N/A

### Description of changes: 
EVP_CipherUpdate takes int for in_len, but add_encrypted_data and
PKCS8_marshal_encrypted_private_key pass size_t values. Add a check
that in_len + block_size fits in int to prevent silent truncation on
64-bit systems and size_t wraparound on 32-bit systems. This replaces
the previous size_t-only overflow check with a unified check that
covers both concerns.

### Call-outs:

There are a couple overlapping concerns here and we should make sure the logic is sound.
- On a 64-bit system, `size_t` will be 64-bits wide and `int` will be 32-bits wide. This potentially results in a silent truncation due to `EVP_CipherUpdate`'s `int` input.
- The Cipher's block-size (128-bits for AES) is added to the input length to determine the output buffer to allocate. This has the potential to overflow the `size_t` used to allocate that buffer (more of a concern for 32-bit systems).

### Testing:

CI

### DIFF
--- a/crypto/fipsmodule/kem/kem.c
+++ b/crypto/fipsmodule/kem/kem.c
@@ -272,6 +272,7 @@ const KEM *KEM_KEY_get0_kem(KEM_KEY* key) {
 }
 
 int KEM_KEY_set_raw_public_key(KEM_KEY *key, const uint8_t *in) {
+  OPENSSL_free(key->public_key);
   key->public_key = OPENSSL_memdup(in, key->kem->public_key_len);
   if (key->public_key == NULL) {
     return 0;
@@ -281,6 +282,7 @@ int KEM_KEY_set_raw_public_key(KEM_KEY *key, const uint8_t *in) {
 }
 
 int KEM_KEY_set_raw_secret_key(KEM_KEY *key, const uint8_t *in) {
+  OPENSSL_free(key->secret_key);
   key->secret_key = OPENSSL_memdup(in, key->kem->secret_key_len);
   if (key->secret_key == NULL) {
     return 0;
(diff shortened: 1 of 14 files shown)


crypto/pkcs8/pkcs8.c
```diff
@@ -502,18 +502,23 @@ int PKCS8_marshal_encrypted_private_key(CBB *out, int pbe_nid,
...
-    OPENSSL_PUT_ERROR(PKCS8, PKCS8_R_TOO_LONG);
+  // |EVP_CipherUpdate| takes |int| for the input length and may output up to
+  // |plaintext_len + block_size| bytes. Ensure the total fits in |int| to avoid
+  // truncation and cannot overflow.
+  size_t block_size = EVP_CIPHER_CTX_block_size(&ctx);
+  if (plaintext_len + block_size < plaintext_len ||
+      plaintext_len + block_size > INT_MAX) {
+    OPENSSL_PUT_ERROR(PKCS8, ERR_R_OVERFLOW);
     goto err;
   }
```
justsmth, line 512: NP: I think it would be slightly easier to reason about as the following:
```
size_t block_size = EVP_CIPHER_CTX_block_size(&ctx);
if (plaintext_len > INT_MAX - block_size) {
    OPENSSL_PUT_ERROR(PKCS8, ERR_R_OVERFLOW);
    goto err;
}
```
This keeps all of the arithmetic on the "trusted" side of the inequality, only requires a single condition, and more clearly states the intent.

justsmth (dismissed):

crypto/pkcs8/pkcs8.c
geedo0: That logic is definitely a lot cleaner and undeniably correct. I spent a solid chunk of time drawing out a n-D truth table trying to reason about the logic I had down and convince myself that it worked for all the cases.

PR #3051, opened 2026-02-27
Fix PKCS12_verify_mac OOB read with invalid password_len

geedo0:
### Description of changes: 
Reject negative password_len values (other than -1) early in
PKCS12_verify_mac before they are used as an array index or cast
to size_t for OPENSSL_memchr. Update the public header comment to
clarify that -1 means the password is a NUL-terminated C string.

### Call-outs:
N/A

### Testing:

CI

### DIFF
--- a/include/openssl/pkcs8.h
+++ b/include/openssl/pkcs8.h
@@ -209,8 +209,10 @@ OPENSSL_EXPORT int PKCS12_set_mac(PKCS12 *p12, const char *password,
 // and zero otherwise. Since |PKCS12_parse| doesn't take a length parameter,
 // it's not actually possible to use a non-NUL-terminated password to actually
 // get anything from a |PKCS12|. Thus |password| and |password_len| may be
-// |NULL| and zero, respectively, or else |password_len| may be -1, or else
-// |password[password_len]| must be zero and no other NUL bytes may appear in
+// |NULL| and zero, respectively, or else |password_len| may be -1 to indicate
+// that |password| is a NUL-terminated C string whose length is determined via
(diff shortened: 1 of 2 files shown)


include/openssl/pkcs8.h
```diff
@@ -209,8 +209,10 @@ OPENSSL_EXPORT int PKCS12_set_mac(PKCS12 *p12, const char *password,
 // and zero otherwise. Since |PKCS12_parse| doesn't take a length parameter,
 // it's not actually possible to use a non-NUL-terminated password to actually
 // get anything from a |PKCS12|. Thus |password| and |password_len| may be
-// |NULL| and zero, respectively, or else |password_len| may be -1, or else
-// |password[password_len]| must be zero and no other NUL bytes may appear in
+// |NULL| and zero, respectively, or else |password_len| may be -1 to indicate
+// that |password| is a NUL-terminated C string whose length is determined via
+// |strlen|, or else |password_len| must be non-negative,
```
WillChilds-Klein, line 214: note for reviewers, looks like `strlen` is [called](https://github.com/aws/aws-lc/blob/37d86461a95782fd5d8b77873f9e1cb134ea2f95/crypto/pkcs8/pkcs8_x509.c#L658) a bit further down the callstack in `PKCS12_get_key_and_certs`

WillChilds-Klein: not your change, but this is such an odd API contract... it enforces that 1/ `password` must be NULL-terminated and 2/ `password_len == -1` indicates that it's NULL-terminated. then eventually calls `strlen` unconditionally.

PR #3052, opened 2026-02-27
Fix PKCS8_encrypt crash when pass is NULL with negative pass_len_in

geedo0:
### Description of changes: 
When pass is NULL and pass_len_in is negative, the cast to size_t wraps
to a huge value, causing a NULL pointer dereference downstream in HMAC
key setup. Handle pass==NULL first by setting pass_len=0, consistent
with the documented empty byte string semantics for NULL passwords.

### Call-outs:
N/A

### Testing:

CI

### DIFF
--- a/crypto/pkcs8/pkcs8_x509.c
+++ b/crypto/pkcs8/pkcs8_x509.c
@@ -188,7 +188,9 @@ X509_SIG *PKCS8_encrypt(int pbe_nid, const EVP_CIPHER *cipher, const char *pass,
                         int pass_len_in, const uint8_t *salt, size_t salt_len,
                         int iterations, PKCS8_PRIV_KEY_INFO *p8inf) {
   size_t pass_len;
-  if (pass_len_in < 0 && pass != NULL) {
+  if (pass == NULL) {
+    pass_len = 0;
+  } else if (pass_len_in < 0) {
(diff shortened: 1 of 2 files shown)


crypto/pkcs8/pkcs8_x509.c
```diff
@@ -188,7 +188,9 @@ X509_SIG *PKCS8_encrypt(int pbe_nid, const EVP_CIPHER *cipher, const char *pass,
                         int pass_len_in, const uint8_t *salt, size_t salt_len,
                         int iterations, PKCS8_PRIV_KEY_INFO *p8inf) {
   size_t pass_len;
-  if (pass_len_in < 0 && pass != NULL) {
+  if (pass == NULL) {
+    pass_len = 0;
+  } else if (pass_len_in < 0) {
     pass_len = strlen(pass);
```
justsmth, line 194: The documentation is silent about [this `pass_len` argument](https://github.com/aws/aws-lc/blob/37d86461a95782fd5d8b77873f9e1cb134ea2f95/include/openssl/pkcs8.h#L69-L85). We should add a sentence there about this logic.

justsmth (approved):

WillChilds-Klein (dismissed):

PR #3055, opened 2026-03-02
Add a test that arbitrary curves can be wrapped in EVP_PKEY

nebeid:
### Description of changes: 
BoringSSL commit https://github.com/google/boringssl/commit/4d48736c2acd6878caa93ac0a628b46bda1227b0: 
"Add a test that arbitrary curves can be wrapped in EVP_PKEY"

Changes made (6 files):

1. crypto/test/test_util.h / test_util.cc — Added shared `HexToBIGNUM(const char*)` helper (declaration + implementation + #include <openssl/bn.h>)

2. crypto/fipsmodule/ec/ec_test.cc — Removed duplicate local `HexToBIGNUM` (now uses the shared one from test_util.h, already included)

3. crypto/fipsmodule/ecdsa/ecdsa_test.cc — Same removal of duplicate

4. crypto/fipsmodule/bn/bn_test.cc — Renamed the different-signature `HexToBIGNUM(UniquePtr<BIGNUM>*, const char*)` → `HexToBIGNUMWithReturn` to avoid conflict (this one returns the 
BN_hex2bn return value, so it's a distinct function)

5. crypto/evp_extra/evp_extra_test.cc — Added the new `TEST(EVPExtraTest, CustomCurve)` test

### Adaptations for AWS-LC:
- Path: crypto/evp_extra/ instead of BoringSSL's crypto/evp/
- Replaced `bssl::StringAsBytes("hello")` (doesn't exist in AWS-LC) with a plain `static const uint8_t msg[] = "hello" + sizeof(msg) - 1`

### DIFF
--- a/crypto/evp_extra/evp_extra_test.cc
+++ b/crypto/evp_extra/evp_extra_test.cc
@@ -3509,3 +3509,83 @@ TEST(EVPExtraTest, SetNoneClearsKey) {
   // Calling operations on the |EVP_PKEY| should cleanly fail.
   EXPECT_EQ(EVP_PKEY_bits(pkey.get()), 0);
 }
+
+// Test that |EC_KEY|s using custom curves can be wrapped in |EVP_PKEY|. Callers
+// should not follow this code. This is maintained and tested only for
+// compatibility with one legacy application, and supported in only a limited
+// form. (E.g. such keys cannot be serialized.)
+TEST(EVPExtraTest, CustomCurve) {
+  // Coefficients for brainpoolp256r1, not supported by BoringSSL.
+  static const char kP[] =
+      "a9fb57dba1eea9bc3e660a909d838d726e3bf623d52620282013481d1f6e5377";
+  static const char kA[] =
+      "7d5a0975fc2c3057eef67530417affe7fb8055c126dc5c6ce94a4b44f330b5d9";
+  static const char kB[] =
+      "26dc5c6ce94a4b44f330b5d9bbd77cbf958416295cf7e1ce6bccdc18ff8c07b6";
+  static const char kX[] =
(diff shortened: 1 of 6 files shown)


crypto/evp_extra/evp_extra_test.cc
```diff
@@ -3509,3 +3509,83 @@ TEST(EVPExtraTest, SetNoneClearsKey) {
   // Calling operations on the |EVP_PKEY| should cleanly fail.
   EXPECT_EQ(EVP_PKEY_bits(pkey.get()), 0);
 }
+
+// Test that |EC_KEY|s using custom curves can be wrapped in |EVP_PKEY|. Callers
+// should not follow this code. This is maintained and tested only for
+// compatibility with one legacy application, and supported in only a limited
+// form. (E.g. such keys cannot be serialized.)
+TEST(EVPExtraTest, CustomCurve) {
+  // Coefficients for brainpoolp256r1, not supported by BoringSSL.
```
justsmth, line 3518: NP: "..., not supported by AWS-LC."?

crypto/test/test_util.h
```diff
@@ -152,6 +152,10 @@ void CustomDataFree(void *parent, void *ptr, CRYPTO_EX_DATA *ad,
 // |reason|.
 testing::AssertionResult ErrorEquals(uint32_t err, int lib, int reason);
 
+// HexToBIGNUM decodes |hex| as a hexadecimal, big-endian, unsigned integer and
+// returns it as a |BIGNUM|, or nullptr on error.
+bssl::UniquePtr<BIGNUM> HexToBIGNUM(const char *hex);
```
justsmth, line 157: NP: We should also add `#include <openssl/bn.h>` to this header since we're referencing `BIGNUM`.

justsmth (approved):

sgmenda (approved):

PR #3073, opened 2026-03-05
Allow zero-length PEM passwords in callback paths

geedo0:
### Description of changes: 
Four sites incorrectly rejected empty passwords by checking
pass_len <= 0 instead of pass_len < 0 after invoking the password
callback. The callback returns negative on error and 0 for a valid
empty password, so the check should be < 0.

### Testing:

CI

### DIFF
--- a/crypto/pkcs8/pkcs8.c
+++ b/crypto/pkcs8/pkcs8.c
@@ -502,18 +502,22 @@ int PKCS8_marshal_encrypted_private_key(CBB *out, int pbe_nid,
     goto err;
   }
 
-  size_t max_out = plaintext_len + EVP_CIPHER_CTX_block_size(&ctx);
-  if (max_out < plaintext_len) {
-    OPENSSL_PUT_ERROR(PKCS8, PKCS8_R_TOO_LONG);
+  // |EVP_CipherUpdate| takes |int| for the input length and may output up to
+  // |plaintext_len + block_size| bytes. Ensure the total fits in |int| to avoid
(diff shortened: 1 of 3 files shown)


crypto/pem/pem_test.cc
```diff
@@ -467,3 +467,61 @@ TEST(PEMTest, WriteReadTraditionalPem) {
...
+  ASSERT_TRUE(ec_key_read);
+  const BIGNUM* orig_priv_key = EC_KEY_get0_private_key(ec_key.get());
+  const BIGNUM* read_priv_key = EC_KEY_get0_private_key(ec_key_read.get());
+  ASSERT_EQ(0, BN_cmp(orig_priv_key, read_priv_key));
+}
+
+TEST(PEMTest, WriteReadPKCS8DerEmptyPassword) {
+  bssl::UniquePtr<EC_KEY> ec_key(EC_KEY_new_by_curve_name(NID_X9_62_prime256v1));
+  ASSERT_TRUE(ec_key);
+  ASSERT_TRUE(EC_KEY_generate_key(ec_key.get()));
```
justsmth, line 506: NP: Should this be conditioned as it is in `PEMTest.WriteReadECPemEmptyPassword`?
```
#if defined(BORINGSSL_FIPS)
  ASSERT_TRUE(EC_KEY_generate_key_fips(ec_key.get()));
#else
  ASSERT_TRUE(EC_KEY_generate_key(ec_key.get()));
#endif
```

PR #3075, opened 2026-03-05
Zeroize intermediate values for ed25519

justsmth:
### Description of changes: 
Zeroize intermediate values for ed25519

### DIFF
--- a/crypto/fipsmodule/curve25519/curve25519.c
+++ b/crypto/fipsmodule/curve25519/curve25519.c
@@ -110,6 +110,8 @@ void ED25519_keypair_from_seed(uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
   OPENSSL_memcpy(out_private_key, seed, ED25519_SEED_LEN);
   OPENSSL_memcpy(out_private_key + ED25519_SEED_LEN, out_public_key,
     ED25519_PUBLIC_KEY_LEN);
+
+  OPENSSL_cleanse(az, sizeof(az));
 }
 
 static int ed25519_keypair_pct(uint8_t public_key[ED25519_PUBLIC_KEY_LEN],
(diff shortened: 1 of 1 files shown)


crypto/fipsmodule/curve25519/curve25519.c
```diff
@@ -110,6 +110,8 @@ void ED25519_keypair_from_seed(uint8_t out_public_key[ED25519_PUBLIC_KEY_LEN],
   OPENSSL_memcpy(out_private_key, seed, ED25519_SEED_LEN);
   OPENSSL_memcpy(out_private_key + ED25519_SEED_LEN, out_public_key,
     ED25519_PUBLIC_KEY_LEN);
+
+  OPENSSL_cleanse(az, sizeof(az));
```
WillChilds-Klein, line 114: hm, are these 64-byte memsets worth it? `az` is a hash of the seed. assuming SHA2 can't be inverted, `az` isn't private material...

crypto/fipsmodule/curve25519/curve25519.c
```diff
@@ -304,6 +307,9 @@ int ed25519_sign_internal(
   // The signature is computed from the private key, but is public.
   CONSTTIME_DECLASSIFY(out_sig, 64);
 
+  OPENSSL_cleanse(az, sizeof(az));
+  OPENSSL_cleanse(r, sizeof(r));
```
WillChilds-Klein, line 311: r isn't private either...

crypto/fipsmodule/curve25519/curve25519.c
justsmth: The first 32 bytes of `az` are the private scalar used in the elliptic curve operations, and the last 32 bytes are the prefix used to derive the signing nonce. 

Knowing `az` is functionally equivalent to knowing the private key for the purpose of signing — an attacker wouldn't need to invert SHA-512, they'd just use `az` directly.

crypto/fipsmodule/curve25519/curve25519.c
justsmth: I believe that `r` is the nonce scalar. If it leaks for a signature, the private key can be recovered.

PR #3076, opened 2026-03-05
More NULL checks in bio_ssl.cc

justsmth:
### Description of changes: 
More NULL checks in bio_ssl.cc


sgmenda (approved):

ssl/bio_ssl.cc
```diff
@@ -182,6 +195,7 @@ static int ssl_free(BIO *bio) {
   if (bio->shutdown) {
     SSL_free(ssl);
   }
+  bio->ptr = NULL;
```
WillChilds-Klein, line 192: do we need to call `BIO_free(bio)`? perhaps it's a chain?

ssl/bio_ssl.cc
```diff
@@ -234,21 +253,20 @@ BIO *BIO_new_ssl_connect(SSL_CTX *ctx) {
...
-    SSL_set_connect_state(ssl);
+    SSL_set_connect_state(ssl.get());
   } else {
-    SSL_set_accept_state(ssl);
+    SSL_set_accept_state(ssl.get());
   }
 
-  if (BIO_set_ssl(ret.get(), ssl, BIO_CLOSE) <= 0) {
+  if (BIO_set_ssl(ret.get(), ssl.get(), BIO_CLOSE) <= 0) {
     return nullptr;
```
WillChilds-Klein, line 262: minor nitpick: looks like this is the only case where `ssl` would need to be free'd? nitpick: should we just free it here instead of relying on `bssl::UniquePtr`?

ssl/bio_ssl.cc
justsmth: No, I don't think so. `ssl_free` is the BIO method's free callback, so it's already being called as part of `BIO_free(bio)`.

ssl/bio_ssl.cc
justsmth: There are actually two failure points where `ssl` could leak:

1. In the `if (!ret || !ssl)` check -- if `ret` is null but `ssl` was successfully allocated (however unlikely), the original code leaked `ssl`.
2. In the `if (BIO_set_ssl(...) <= 0)` check.

`bssl::UniquePtr` handles both cases cleanly without needing `SSL_free` .

PR #3078, opened 2026-03-06
Abort on `RAND_bytes` failure

justsmth:
### Description of changes:

`RAND_bytes` in AWS-LC always returns 1 and cannot actually fail. However, call sites across the codebase handle its return value inconsistently:

1. **Ignored entirely** — bare `RAND_bytes(...)` calls with no check at all.
2. **Error propagated** — `if (!RAND_bytes(...)) { goto err; }` patterns that
   attempt graceful recovery.
3. **Abort on failure** — `if (RAND_bytes(...) != 1) { abort(); }` (only one
   call site).

For a cryptographic library, silently continuing with a failed RNG is never the right choice. This change introduces an `AWSLC_ABORT_IF_NOT_ONE` macro and applies it uniformly to all `RAND_bytes` call sites within `crypto/`. If `RAND_bytes` ever returns a non-1 value in the future, the process will abort immediately rather than operating with potentially uninitialized randomness.

### Call-outs:

- `RAND_bytes` calls in `ssl/` and test files are **not** converted in this change.
- The `ml_dsa.c` diff includes some incidental trailing-whitespace cleanup on blank lines.
- In `PKCS12_create` (`pkcs8_x509.c`), the `RAND_bytes` call is now evaluated before `CBB_flush` rather than being short-circuited behind it. This is benign since `RAND_bytes` has no dependency on CBB state.

### Testing:
Existing tests. No behavioral change under normal operation since `RAND_bytes` always returns 1 today.

### DIFF
--- a/crypto/fipsmodule/cipher/e_aes.c
+++ b/crypto/fipsmodule/cipher/e_aes.c
@@ -461,9 +461,8 @@ static int aes_gcm_ctrl(EVP_CIPHER_CTX *c, int type, int arg, void *ptr) {
       // lock functions to avoid updating the service indicator with the DRBG
       // functions.
       FIPS_service_indicator_lock_state();
-      if (c->encrypt && !RAND_bytes(gctx->iv + arg, gctx->ivlen - arg)) {
-        FIPS_service_indicator_unlock_state();
-        return 0;
+      if (c->encrypt) {
+        AWSLC_ABORT_IF_NOT_ONE(RAND_bytes(gctx->iv + arg, gctx->ivlen - arg));
       }
       FIPS_service_indicator_unlock_state();
       gctx->iv_gen = 1;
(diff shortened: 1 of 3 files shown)


crypto/fipsmodule/ml_dsa/mldsa_native_config.h
```diff
@@ -87,8 +87,8 @@ static MLD_INLINE void mld_zeroize(void *ptr, size_t len) {
 #include "mldsa/sys.h"
 #include <openssl/rand.h>
 static MLD_INLINE int mld_randombytes(uint8_t *ptr, size_t len) {
-    // RAND_bytes returns 1 on success, 0 or -1 on failure
-    return RAND_bytes(ptr, len) == 1 ? 0 : -1;
+    AWSLC_ABORT_IF_NOT_ONE(RAND_bytes(ptr, len));
```
nebeid, line 90: I wonder if these changes need to be communicated upstream so they're not overwritten by a re-import. Here and for mlkem.

justsmth: This is code that must be customized when imported:
https://github.com/pq-code-package/mldsa-native/blob/main/mldsa/mldsa_native_config.h#L400-L410

We would need to document this in our runbook for importing.

torben-hansen (approved):

PR #3080, opened 2026-03-06
Use explicit check for X509 path length

nhatnghiho:
### Description of changes: 
Add an explicit bounds check for X509 path length


crypto/x509/x509_vfy.c
```diff
@@ -729,7 +729,7 @@ static int check_chain_extensions(X509_STORE_CTX *ctx) {
     }
     // Check pathlen if not self issued
     if (i > 1 && !(x->ex_flags & EXFLAG_SI) && x->ex_pathlen != -1 &&
-        plen > x->ex_pathlen + 1) {
+        (x->ex_pathlen > LONG_MAX - 1 || plen > x->ex_pathlen + 1)) {
```
justsmth: `x->ex_pathlen` is a `long` value, so this condition only triggers when a CA certificate has a pathlen constraint of exactly `LONG_MAX` — an absurd value no real certificate would have.

We should instead just use the following comparison which is safe for all values:
```
plen - 1 > x->ex_pathlen
```

`plen` is a non-negative `int`, so `plen - 1` can't underflow past `-1`, and we've already filtered out `ex_pathlen == -1`

PR #3089, opened 2026-03-10
Correct purpose setting for OCSP_request_verify

samuel40791765:
Fixes a logic error from `&&` → `||` in `OCSP_request_verify` when initializing the store context and setting its purpose. With `&&`, a `set_purpose` failure is silently ignored when `init` succeeds, so `X509_verify_cert` runs without `X509_PURPOSE_OCSP_HELPER`.

`OCSP_HELPER` is already intentionally permissive (skips leaf EKU checks, uses basic self-signed root trust), and the default path without it behaves nearly the same way. The signature is already validated before this point.

### DIFF
--- a/crypto/ocsp/ocsp_verify.c
+++ b/crypto/ocsp/ocsp_verify.c
@@ -201,6 +201,8 @@ static int ocsp_check_ids(STACK_OF(OCSP_SINGLERESP) *sresp, OCSP_CERTID **ret) {
   return 1;
 }
 
+// Returns -1 on fatal error, 0 if there is no match and 1 if there is a
+// match.
 static int ocsp_match_issuerid(X509 *cert, OCSP_CERTID *cid,
                                STACK_OF(OCSP_SINGLERESP) *sresp) {
   if (cert == NULL) {
@@ -232,13 +234,14 @@ static int ocsp_match_issuerid(X509 *cert, OCSP_CERTID *cid,
         return 0;
       }
     }
-    if (OPENSSL_memcmp(md, cid->issuerNameHash->data, mdlen) != 0) {
+    if (0 != OPENSSL_memcmp(md, cid->issuerNameHash->data, mdlen)) {
       return 0;
     }
     if (1 != X509_pubkey_digest(cert, dgst, md, NULL)) {
-      if (OPENSSL_memcmp(md, cid->issuerKeyHash->data, mdlen) != 0) {
-        return 0;
-      }
+      return -1;
+    }
+    if (0 != OPENSSL_memcmp(md, cid->issuerKeyHash->data, mdlen)) {
+      return 0;
     }
     return 1;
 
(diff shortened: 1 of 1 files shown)


crypto/ocsp/ocsp_verify.c
justsmth: This condition appears to always be true.

crypto/ocsp/ocsp_verify.c
```diff
@@ -232,11 +232,11 @@ static int ocsp_match_issuerid(X509 *cert, OCSP_CERTID *cid,
         return 0;
       }
     }
-    if (memcmp(md, cid->issuerNameHash->data, mdlen) != 0) {
+    if (OPENSSL_memcmp(md, cid->issuerNameHash->data, mdlen) != 0) {
       return 0;
     }
-    if (0 <= X509_pubkey_digest(cert, dgst, md, NULL)) {
-      if (memcmp(md, cid->issuerKeyHash->data, mdlen) != 0) {
+    if (!X509_pubkey_digest(cert, dgst, md, NULL)) {
```
justsmth: NP: For safety/maintainability, I think it would be better to be explicit about the success value:
```
if (1 != X509_pubkey_digest(cert, dgst, md, NULL)) {
```

justsmth (approved):

crypto/ocsp/ocsp_verify.c
```diff
@@ -473,7 +473,7 @@ int OCSP_request_verify(OCSP_REQUEST *req, STACK_OF(X509) *certs,
   int ret = 0;
   if (!IS_OCSP_FLAG_SET(flags, OCSP_NOVERIFY)) {
     // Initialize and set purpose of |ctx| for verification.
-    if (!X509_STORE_CTX_init(ctx, store, signer, NULL) &&
+    if (!X509_STORE_CTX_init(ctx, store, signer, NULL) ||
         !X509_STORE_CTX_set_purpose(ctx, X509_PURPOSE_OCSP_HELPER)) {
```
justsmth: NP: We could also make these success values explicit:
```
    if (1 != X509_STORE_CTX_init(ctx, store, signer, NULL) ||
        1 != X509_STORE_CTX_set_purpose(ctx, X509_PURPOSE_OCSP_HELPER)) {
```

crypto/ocsp/ocsp_verify.c
```diff
@@ -232,11 +232,11 @@ static int ocsp_match_issuerid(X509 *cert, OCSP_CERTID *cid,
         return 0;
       }
     }
-    if (memcmp(md, cid->issuerNameHash->data, mdlen) != 0) {
+    if (OPENSSL_memcmp(md, cid->issuerNameHash->data, mdlen) != 0) {
       return 0;
     }
-    if (0 <= X509_pubkey_digest(cert, dgst, md, NULL)) {
-      if (memcmp(md, cid->issuerKeyHash->data, mdlen) != 0) {
+    if (1 != X509_pubkey_digest(cert, dgst, md, NULL)) {
```
WillChilds-Klein, line 240: these are logically distinct -- old logic would evaluate to true if `X509_pubkey_digest` returns either 0 or 1 while the new logic will only evaluate to true if `X509_pubkey_digest` return 0.

i think the new logic is correct, but to confirm -- that's what we want right?

samuel40791765: No sorry, you're right. I blindly took the recommendation here, should have adjusted accordingly.

crypto/ocsp/ocsp_verify.c
```diff
@@ -235,7 +235,7 @@ static int ocsp_match_issuerid(X509 *cert, OCSP_CERTID *cid,
     if (OPENSSL_memcmp(md, cid->issuerNameHash->data, mdlen) != 0) {
       return 0;
     }
-    if (1 != X509_pubkey_digest(cert, dgst, md, NULL)) {
+    if (1 == X509_pubkey_digest(cert, dgst, md, NULL)) {
```
WillChilds-Klein: should we error if this fails?

PR #3094, opened 2026-03-12
Fix modulewrapper memory leak

justsmth:
### Description of changes:
The `RSADecryptionPrimitiveCRT` function in the ACVP modulewrapper leaks 8 BIGNUMs on every call. `RSA_new_private_key_large_e` takes `const BIGNUM*` parameters and copies them internally, but the originals were never freed. `RSASignaturePrimitive` and `RSADecryptionPrimitive` had the same issue on their error paths (before `RSA_set0_key` takes ownership).

Under ASAN, LeakSanitizer reports these leaks to stderr when the modulewrapper exits. Since `check_expected.go` captures combined stdout+stderr, the leak output gets mixed into the ACVP test result, causing a mismatch against the expected output.

All three functions now use `bssl::UniquePtr<BIGNUM>` for automatic cleanup.

### Testing:
Existing ACVP tests in the FIPS ASAN CI job should now pass.

### DIFF
--- a/util/fipstools/acvp/modulewrapper/modulewrapper.cc
+++ b/util/fipstools/acvp/modulewrapper/modulewrapper.cc
@@ -2970,16 +2970,20 @@ static bool RSASignaturePrimitive(const Span<const uint8_t> args[],
   const Span<const uint8_t> e_bytes = args[2];
   const Span<const uint8_t> msg = args[3];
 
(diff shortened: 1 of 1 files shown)


nebeid (approved):

WillChilds-Klein (approved):

sgmenda (approved):
lgtm

PR #3095, opened 2026-03-12
Fix race condition in  `new_certs_dir` output path

justsmth:
### Description of changes:
`outdirlen` in `caTool` was never set after building the directory prefix, so it stayed at 0. This caused the cert filename (e.g. `01.pem`) to be written to the current working directory instead of the configured `new_certs_dir`. When parallel test processes shared a working directory, they'd stomp on each other's `01.pem`.

The fix sets `outdirlen = strlen(new_cert)` after constructing the directory prefix so certs land in the right place.

Also cleaned up `TearDown` to remove the files that now (correctly) end up inside `new_certs_dir`, and the auxiliary `.old` files the CA tool creates during serial/db rotation.

### Call-outs:
This was a bug in `ca.cc` itself, not the test. Every `openssl ca` invocation was writing the new cert to the CWD — the test flakiness just happened to surface it.

### Testing:
All 59 `CATest` tests pass. Verified via verbose output that certs now go to the temp directory (`/tmp/awslcTestDir.../01.pem`) instead of the CWD.

### DIFF
--- a/tool-openssl/ca_test.cc
+++ b/tool-openssl/ca_test.cc
@@ -115,7 +115,7 @@ class CATest : public ::testing::Test {
 #else
     DIR *d = opendir(dir_path);
     if (d) {
-      struct dirent *entry;
+      struct dirent *entry = nullptr;
       while ((entry = readdir(d)) != nullptr) {
         if (strcmp(entry->d_name, ".") != 0 &&
             strcmp(entry->d_name, "..") != 0) {


tool-openssl/ca_test.cc
```diff
@@ -73,19 +76,58 @@ class CATest : public ::testing::Test {
...
+        if (strcmp(entry->d_name, ".") != 0 &&
+            strcmp(entry->d_name, "..") != 0) {
+          std::string file_path =
+              std::string(dir_path) + "/" + entry->d_name;
+          unlink(file_path.c_str());
+        }
+      }
+      closedir(d);
     }
+    rmdir(dir_path);
```
WillChilds-Klein, line 129: do we need to account for failure conditions here? it's a test, so probably not...

WillChilds-Klein (approved):

nebeid (approved):

PR #3096, opened 2026-03-12
Add Symbol Versioning Support

skmcgrail:
## Summary

- Adds GNU ld symbol versioning (`.map` files) to `libcrypto-awslc` and `libssl-awslc` shared libraries, automatically enabled when building with `-DENABLE_DIST_PKG=ON`
- Introduces symbol registry files (`crypto/libcrypto.txt`, `ssl/libssl.txt`) as the source of truth, with Go-based tooling (`util/read_public_symbols/`, `util/generate_version_script/`) to generate and maintain version scripts
- Adds CI jobs for incremental and baseline symbol checks (in `abidiff.yml`) and multi-arch dist-pkg tests (Ubuntu 24.04, AL2, AL2023 × gcc/clang × x86_64/aarch64)
- Expands dist-pkg CI from a single x86 job to a full matrix across architectures and OS images, splitting install vs runtime test phases

## Details

**Build system changes:**
- New `cmake/GenerateVersionScript.cmake` provides `apply_version_script()` to attach a `.map` file to a target
- `CMakeLists.txt` probes for `--undefined-version` linker support (needed for lld compatibility) and sets `ENABLE_SYMBOL_VERSIONING` when `SET_LIB_SONAME` and `BUILD_SHARED_LIBS` are both true
- `crypto/CMakeLists.txt` and `ssl/CMakeLists.txt` call `apply_version_script()` on their respective targets

**Symbol management tooling:**
- `util/read_public_symbols/main.go` — parses headers to extract public API symbols and classify visibility (PUBLIC, PRIVATE, PRIVATE_CXX)
- `util/generate_version_script/main.go` — generates `.map` files from registry `.txt` files
- `util/generate_initial_version_scripts.sh` — bootstraps registries and maps from scratch
- `util/update_symbol_version.sh` — adds new symbols to a new version node with inheritance

**CI:**
- 4 new jobs in `abidiff.yml`: incremental and baseline symbol checks for both libraries
- Dist-pkg tests moved from `linux_x86_omnibus.yml` (single job) to `linux-multi-arch-omnibus.yml` (full matrix with install/run split)
- New Docker image (`.github/docker_images/symbol_check/`) and check script (`check_symbols.sh`)

**Testing:**
- `tests/ci/run_symbol_version_test.sh` — comprehensive validation that versioned symbols are applied, no unversioned symbols leak, libraries link correctly, and version definitions are present
- `tests/ci/run_dist_pkg_run_tests.sh` — runtime tests for installed dist-pkg libraries

**Documentation:**
- `docs/SymbolVersioning.md` — full developer guide covering version evolution, CI policy, and troubleshooting
- `BUILDING.md` and `README.md` updated with distribution packaging section

## Test plan

- [ ] Build with `-DENABLE_DIST_PKG=ON -DBUILD_SHARED_LIBS=ON` and verify `readelf --version-info` shows `AWS_LC_1.0` on both `.so` files
- [ ] Run `tests/ci/run_symbol_version_test.sh` locally
- [ ] Verify CI passes: abidiff symbol check jobs (incremental + baseline) and multi-arch dist-pkg matrix
- [ ] Confirm no unversioned dynamic symbols via `nm -D libcrypto-awslc.so | grep ' T ' | grep -v '@'`


tests/ci/run_symbol_version_test.sh
```diff
@@ -0,0 +1,326 @@
+#!/usr/bin/env bash
```
justsmth, line 1: - This tests indicates success when run, but returns an error exit status:
```
...
-- Build files have been written to: /home/justsmth/repos/aws-lc/build_symbol_test
ninja: Entering directory `/home/justsmth/repos/aws-lc/build_symbol_test'
[416/416] Creating library symlink ssl/libssl-awslc.so.0 ssl/libssl-awslc.so
✓ PASS: Build successful
```

- This test should be added to the CI.

tests/ci/run_symbol_version_test.sh
```diff
@@ -0,0 +1,326 @@
...
+if [[ $? -eq 0 ]]; then
+  print_pass "Build successful"
+else
+  print_fail "Build failed"
+  exit 1
+fi
+
+# Library paths
+LIBCRYPTO_SO="${BUILD_DIR}/crypto/libcrypto-awslc.so.0.0.0"
+LIBSSL_SO="${BUILD_DIR}/ssl/libssl-awslc.so.0.0.0"
```
justsmth: - `libcrypto-awslc.so.0.0.0` is hard-coded, but the build produces `libcrypto-awslc.so.1.70.0` (the `SOFTWARE_VERSION` from CMakeLists.txt)?

tests/ci/run_symbol_version_test.sh
```diff
@@ -0,0 +1,326 @@
...
+  echo ""
+  echo "=========================================="
+  echo "TEST: $1"
+  echo "=========================================="
+}
+
+print_pass() {
+  echo -e "${GREEN}✓ PASS${NC}: $1"
+  ((pass_count++))
+}
```
justsmth, line 53: The silent failure is due to `set -e` and `((pass_count++))`, which on first run evaluates to `((0++))` returning exit code 1 and terminating the script.

This should work:
```
pass_count=$((pass_count + 1))
```

ozbenh:
[IGNORE] Use different symbol versions for libcrypto and libssl. ie AWS_LC_SSL_1_0 and AWS_LC_CRYPTO_1_0. Otherwise the RPM provides/requires will get confused esp if one changes and not the other

[UPDATE] I am wrong here. It works fine to use the same version symbol, rpm namespaces it by soname anyways and that's how glibc etc... do it (all the sub libs have the same version symbol).

docs/SymbolVersioning.md
```diff
@@ -0,0 +1,344 @@
...
+} AWS_LC_1.0;  /* Inherits all AWS_LC_1.0 symbols */
+```
+
+The version inheritance (`} AWS_LC_1.0;`) means `AWS_LC_1.1` includes all `AWS_LC_1.0` symbols plus the new ones.
+
+#### Step 3: Update Baseline
+
+```bash
+# Generate new baseline
+./util/extract_symbols.sh build/crypto/libcrypto-awslc.so.0 \
```
justsmth: I don't see an `extract_symbols.sh` in this PR. Does it need to be added?

CMakeLists.txt
```diff
@@ -125,6 +126,34 @@ endif()
 
 enable_language(C)
 
+# Enable symbol versioning for shared libraries on UNIX platforms.
+# Version scripts are derived from the symbol registry files and have stable
+# filenames (crypto/libcrypto.map, ssl/libssl.map). The version node names are
+# recorded in the registry (crypto/libcrypto.txt, ssl/libssl.txt).
+if(SET_LIB_SONAME AND BUILD_SHARED_LIBS AND UNIX AND NOT APPLE)
```
justsmth: `SET_LIB_SONAME` is also set in the default `elseif(NOT ENABLE_PRE_SONAME_BUILD AND BUILD_SHARED_LIBS AND UNIX AND NOT APPLE)` branch (L105), so this enables symbol versioning for *every* standard Linux shared build, and not just `ENABLE_DIST_PKG`? 

That's broader than the PR description/docs state. If dist-pkg-only is intended:
```cmake
if(ENABLE_DIST_PKG AND BUILD_SHARED_LIBS AND UNIX AND NOT APPLE)
```
Otherwise, we should document that versioning now applies to all SONAME shared builds.

crypto/libcrypto.map
```diff
@@ -0,0 +1,3044 @@
+# GNU ld version script for AWS-LC
+# Auto-generated from symbol registry. Do not edit manually.
+# To add new symbols: util/update_symbol_version.sh <new_version>
+# To regenerate:      util/generate_initial_version_scripts.sh
+
```
justsmth, line 5: > `# Auto-generated from symbol registry. Do not edit manually.`

Should we add a CI step that runs `generate_version_script` on both registries and `git diff --exit-code`s the maps to avoid drift?

tests/ci/run_symbol_version_test.sh
```diff
@@ -0,0 +1,332 @@
...
+  nm -D "${LIBCRYPTO_SO}" | \
+    awk '$2 == "T" && $3 !~ /@/ && $3 !~ /^_/ && $3 !~ /^OPENSSL_memory/ { print "  " $3 }' | \
+    head -10
+fi
+
+SSL_UNVERSIONED=$(nm -D "${LIBSSL_SO}" | \
+  awk '$2 == "T" && $3 !~ /@/ && $3 !~ /^_/ { print $3 }' | \
+  wc -l)
+
+if [[ ${SSL_UNVERSIONED} -eq 0 ]]; then
```
justsmth, line 239: This validates that exported symbols *are* versioned, but not that all symbols exported before this change are still exported. (Since `local: *;` will hide any `OPENSSL_EXPORT` symbol the generator missed, a gap in the registry is a silent symbol drop.)

Are we relying on the `abidiff` job to catch this problem? Is it running against the versioned `.so`?

tests/ci/run_symbol_version_test.sh
```diff
@@ -0,0 +1,332 @@
...
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+# Test script to validate ELF symbol versioning in AWS-LC
+#
+# This script verifies that:
+# 1. Symbol version scripts are applied correctly
+# 2. All exported symbols have proper version tags
+# 3. No unversioned symbols leak
+# 4. Version definitions are present in shared libraries
+# 5. Libraries can be linked and used correctly
```
justsmth, line 12: The registry is generated from a non-FIPS build (no FIPS define in the target sets), but this applies it to a FIPS build? (FIPS-gated `OPENSSL_EXPORT` symbols won't be registered and will be localized.)

Can we confirm this job actually exercises those exports rather than just linking?

.github/docker_images/symbol_check/check_symbols.sh
```diff
@@ -0,0 +1,201 @@
...
+
+if [[ "${MODE}" != "incremental" && "${MODE}" != "baseline" ]]; then
+  echo "Error: second argument must be 'incremental' or 'baseline'"
+  exit 2
+fi
+
+REGISTRY_PATH="${CHECK_LIB}/lib${CHECK_LIB}.txt"
+
+if [[ "${CHECK_LIB}" == "crypto" ]]; then
+  READ_FLAGS=(-exclude ssl.h -internal-dirs crypto)
```
justsmth: The generation scripts use `-internal-dirs crypto,third_party/jitterentropy`, so the `JENT_*` symbols are registered but never scanned here and will permanently surface under the "registered but not in headers" warning. Aligning the flags avoids the standing noise:
```suggestion
  READ_FLAGS=(-exclude ssl.h -internal-dirs crypto,third_party/jitterentropy)
```

.github/docker_images/symbol_check/check_symbols.sh
```diff
@@ -0,0 +1,201 @@
...
+    exit 2
+  fi
+
+  # Extract current symbols from headers (public + internal, with visibility)
+  go run ./util/read_public_symbols \
+    -include-dir include \
+    -source-root . \
+    -emit-visibility \
+    "${READ_FLAGS[@]}" \
+    -out /tmp/header_syms.txt 2>/dev/null
```
justsmth: Suppressing stderr here means a broken extraction produces an empty `header_syms.txt`, which makes the baseline check pass trivially (no unregistered symbols, everything just "missing"). 

Suggest dropping `2>/dev/null` (or teeing to a log) so extraction failures are visible.

docs/SymbolVersioning.md
```diff
@@ -0,0 +1,344 @@
...
+  -DBUILD_SHARED_LIBS=ON \
+  -DENABLE_DIST_PKG=ON \
+  -DCMAKE_BUILD_TYPE=Release
+
+ninja -C build
+```
+
+This produces:
+- `build/crypto/libcrypto-awslc.so.0.0.0` - with AWS_LC_1.0 versioned symbols
+- `build/ssl/libssl-awslc.so.0.0.0` - with AWS_LC_1.0 versioned symbols
```
justsmth: Several stale references:

* The build output is now `.so.1.x.x` (ABI_VERSION bumped to 1).
* The version scripts are `crypto/libcrypto.map` / `ssl/libssl.map` (L93-94). 
* `util/extract_symbols.sh` doesn't exist (L145, L242, L257-260 -- the tooling is `read_public_symbols` / `generate_version_script` / `update_symbol_version.sh`). 
* `tests/ci/baselines/symbols/*.txt` has moved to `crypto/`+`ssl/`.

tests/ci/run_dist_pkg_run_tests.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR ISC
+
+set -exo pipefail
```
justsmth: Nit:
```suggestion
set -exuo pipefail
```

.github/workflows/abidiff.yml
```diff
@@ -108,3 +108,73 @@ jobs:
       - name: Perform libssl ABI Diff against release
         run: |
           docker run -v ${{ github.workspace }}/previous:/previous -v ${{ github.workspace }}/next:/next abidiff ssl
+
+  libcrypto-symbols-incremental:
+    if: github.repository_owner == 'aws'
+    name: libcrypto symbol check (incremental)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
justsmth: Nit: these new jobs use `actions/checkout@v3` while the rest of the repo is now on v5/v6.

justsmth (changes requested):

cmake/GenerateVersionScript.cmake
```diff
@@ -0,0 +1,57 @@
...
+
+# GenerateVersionScript.cmake
+#
+# Provides apply_version_script() to link a GNU ld version script into an ELF
+# shared library target. Version scripts are maintained as checked-in source
+# files (crypto/libcrypto.map, ssl/libssl.map) and generated from the symbol
+# registries (crypto/libcrypto.txt, ssl/libssl.txt) via:
+#
+#   util/generate_initial_version_scripts.sh   (initial population)
+#   util/update_symbol_version.sh <version>    (adding new symbols)
```
WillChilds-Klein, line 12: so we run `generate_initial_version_scripts.sh` for new ABI major versions and `update_symbol_version.sh` for minor/patch version updates within a major version? is that correct?

this documentation may get stale if we add new version script generator scripts. perhaps the symbol diff CI check suggested by justin would be a good place to document how to update the version scripts when new symbols are introduced?

cmake/GenerateVersionScript.cmake
```diff
@@ -0,0 +1,57 @@
...
+  endif()
+
+  if(NOT DEFINED ARG_VERSION_SCRIPT)
+    message(FATAL_ERROR "apply_version_script: VERSION_SCRIPT is required")
+  endif()
+
+  if(NOT UNIX OR APPLE)
+    message(STATUS "Symbol versioning not supported on this platform (requires GNU ld or compatible)")
+    return()
+  endif()
```
WillChilds-Klein, line 39: are we sure that all Linux linkers (lld, gold, mold) support symbol version maps? should we check the default linker to make sure it supports symbol versioning here?

crypto/CMakeLists.txt
```diff
@@ -632,6 +632,19 @@ function(build_libcrypto)
...
+  # Apply symbol versioning for shared libraries.
+  if(ENABLE_SYMBOL_VERSIONING)
+    set(VERSION_SCRIPT "${CMAKE_CURRENT_SOURCE_DIR}/libcrypto.map")
+    if(EXISTS "${VERSION_SCRIPT}")
+      apply_version_script(
+        TARGET ${arg_NAME}
+        VERSION_SCRIPT "${VERSION_SCRIPT}"
+      )
+    else()
+      message(WARNING "Symbol version script not found: ${VERSION_SCRIPT}")
```
WillChilds-Klein: here and for libssl, should this error? this execution path means `ENABLE_SYMBOL_VERSIONING`, so we should expect the version script to be present, right?

CMakeLists.txt
WillChilds-Klein: Also, should we add an `elseif(ENABLE_DIST_PKG)` clause that errors out if this is set on Apple/Windows?

tests/ci/run_symbol_version_test.sh
```diff
@@ -0,0 +1,332 @@
...
+  local dir="$1"
+  if [[ -d "${dir}/lib64" ]]; then
+    echo "lib64"
+  else
+    echo "lib"
+  fi
+}
+
+# Check if we're on a supported platform
+if [[ "$OSTYPE" != "linux-gnu"* ]] && [[ "$OSTYPE" != "freebsd"* ]]; then
```
WillChilds-Klein: nit: not supported on openbsd?

tests/ci/run_symbol_version_test.sh
```diff
@@ -0,0 +1,332 @@
...
+  print_test "Building and installing AWS-LC with symbol versioning"
+  rm -rf "${BUILD_DIR}"
+  mkdir -p "${BUILD_DIR}"
+
+  cd "${SOURCE_ROOT}"
+  cmake -GNinja -B "${BUILD_DIR}" \
+    -DBUILD_SHARED_LIBS=ON \
+    -DENABLE_DIST_PKG=ON \
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+    -DCMAKE_INSTALL_PREFIX="${INSTALL_DIR}"
```
WillChilds-Klein, line 114: similar to justin's comment above, are there cases where we need to pass `-DFIPS=ON`?

tests/ci/run_symbol_version_test.sh
```diff
@@ -0,0 +1,332 @@
...
+
+LIB_DIR=$(get_lib_dir "${INSTALL_DIR}")
+LIBCRYPTO_SO="${INSTALL_DIR}/${LIB_DIR}/libcrypto-awslc.so"
+LIBSSL_SO="${INSTALL_DIR}/${LIB_DIR}/libssl-awslc.so"
+INCLUDE_DIR="${INSTALL_DIR}/include/aws-lc"
+LINK_LIB_DIR="${INSTALL_DIR}/${LIB_DIR}"
+
+# Extract the symbol version tag from the linker map
+SYMBOL_VERSION=$(grep -m1 -oP '^AWS_LC_\S+(?= \{)' "${SOURCE_ROOT}/crypto/libcrypto.map")
+if [[ -z "${SYMBOL_VERSION}" ]]; then
```
WillChilds-Klein, line 136: beyond being non-empty, should we also test that it's a well-formed version string?

tests/ci/run_symbol_version_test.sh
```diff
@@ -0,0 +1,332 @@
...
+if readelf --version-info "${LIBSSL_SO}" | grep "${SYMBOL_VERSION}" > /dev/null; then
+  print_pass "libssl has ${SYMBOL_VERSION} version definition"
+else
+  print_fail "libssl missing ${SYMBOL_VERSION} version definition"
+fi
+
+# Test 3: Check versioned symbols
+print_test "Check for versioned symbols"
+
+CRYPTO_VERSIONED=$(nm -D "${LIBCRYPTO_SO}" | grep -c "@@${SYMBOL_VERSION}" || true)
```
WillChilds-Klein, line 191: `|| true` to avoid `-e` mysteriously failing? isn't there type confusion with the check on L164?

tests/ci/run_symbol_version_test.sh
```diff
@@ -0,0 +1,332 @@
...
+print_info "libcrypto: ${CRYPTO_VERSIONED} symbols with @@${SYMBOL_VERSION}"
+
+if [[ ${CRYPTO_VERSIONED} -gt 3000 ]]; then
```
WillChilds-Klein: where does 3k come from?

tests/ci/run_symbol_version_test.sh
```diff
@@ -0,0 +1,332 @@
...
+  print_pass "libssl has expected number of versioned symbols"
+else
+  print_fail "libssl has fewer versioned symbols than expected (${SSL_VERSIONED} < 500)"
+fi
+
+# Test 4: Check for unversioned exports
+print_test "Check for unversioned exported symbols"
+
+CRYPTO_UNVERSIONED=$(nm -D "${LIBCRYPTO_SO}" | \
+  awk '$2 == "T" && $3 !~ /@/ && $3 !~ /^_/ && $3 !~ /^OPENSSL_memory/ { print $3 }' | \
```
WillChilds-Klein, line 222: suggest doc comment explaining this awk expression

tests/ci/run_symbol_version_test.sh
```diff
@@ -0,0 +1,332 @@
...
+  else
+    print_fail "${symbol} missing or incorrectly versioned"
+  fi
+}
+
+check_symbol "${LIBCRYPTO_SO}" "EVP_MD_CTX_new" "${SYMBOL_VERSION}"
+check_symbol "${LIBCRYPTO_SO}" "AES_encrypt" "${SYMBOL_VERSION}"
+check_symbol "${LIBCRYPTO_SO}" "RSA_new" "${SYMBOL_VERSION}"
+check_symbol "${LIBSSL_SO}" "SSL_CTX_new" "${SYMBOL_VERSION}"
+check_symbol "${LIBSSL_SO}" "SSL_new" "${SYMBOL_VERSION}"
```
WillChilds-Klein, line 273: suggest comment explaining how these symbols were chosen and under what conditions new symbols should be added here.

tests/ci/run_symbol_version_test.sh
```diff
@@ -0,0 +1,332 @@
...
+    -I "${INCLUDE_DIR}" \
+    -L "${LINK_LIB_DIR}" \
+    -lcrypto-awslc -lssl-awslc \
+    -Wl,-rpath,"${LINK_LIB_DIR}"; then
+    print_pass "Test program compiled successfully"
+  else
+    print_fail "Test program compilation failed"
+  fi
+
+  if [[ -f "${TEST_PROG}" ]]; then
```
WillChilds-Klein: if compilation succeeds, how can `$TEST_PROG` not be present? and if it isn't, shouldn't we fail? there's currently no `else` clause to this conditional.

util/generate_initial_version_scripts.sh
```diff
@@ -0,0 +1,116 @@
...
+# of truth. Each line records a symbol, its version node, and its visibility:
+#
+#   AES_encrypt AWS_LC_1.0 PUBLIC
+#   CRYPTO_once AWS_LC_1.0 PRIVATE
+#   ssl_cert_check_key_usage AWS_LC_1.0 PRIVATE_CXX
+#
+# Visibility values:
+#   PUBLIC      - public API from include/openssl/*.h, can never be removed
+#   PRIVATE     - internal API from crypto/**/*.h (C linkage), can be removed
+#   PRIVATE_CXX - internal API from ssl/**/*.h (C++ linkage), can be removed
```
WillChilds-Klein, line 17: are these GNU-`ld` constants or defined/parsed by the go tooling? if former, are they standard across other Linux linkers?

util/generate_version_script/main.go
```diff
@@ -0,0 +1,234 @@
...
+
+func init() {
+	flag.StringVar(&inFile, "in", "", "Symbol registry file (one '<symbol> <version> [visibility]' per line)")
+	flag.StringVar(&outFile, "out", "", "Output GNU ld version script (.map)")
+}
+
+// symbolInfo holds a symbol name and its visibility classification.
+type symbolInfo struct {
+	name       string
+	visibility string // "PUBLIC", "PRIVATE", or "PRIVATE_CXX"
```
WillChilds-Klein: nit: not sure what the golang convention is, but should we tighten this with an  enum?

util/generate_initial_version_scripts.sh
```diff
@@ -0,0 +1,116 @@
...
+CC_FLAG="${CC:-cc}"
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --cc) CC_FLAG="$2"; shift 2 ;;
+    *) echo "Unknown argument: $1"; exit 1 ;;
+  esac
+done
+
+INITIAL_VERSION="AWS_LC_1.0"
```
WillChilds-Klein, line 44: will this need to be updated when we bump major/minor symbol version? can this be a config value set somewhere, perhaps in source headers like our FIPS validation version?

or is this that config?

util/generate_initial_version_scripts.sh
```diff
@@ -0,0 +1,116 @@
...
+echo ""
+echo "Step 3: Writing symbol registry files (${INITIAL_VERSION})..."
+
+# Write registry: "<symbol> AWS_LC_1.0 <visibility>" sorted by symbol name
+# Input format from read_public_symbols: "SYMBOL VISIBILITY"
+awk -v ver="${INITIAL_VERSION}" '{ print $1, ver, $2 }' /tmp/libcrypto_symbols.txt \
+  | sort > "${CRYPTO_REGISTRY}"
+
+awk -v ver="${INITIAL_VERSION}" '{ print $1, ver, $2 }' /tmp/libssl_symbols.txt \
+  | sort > "${SSL_REGISTRY}"
```
WillChilds-Klein: suggest giving the temp files a different basename than the registry. it looks like the temp files don't contain symbol version but the registry does?

util/read_public_symbols/main.go
```diff
@@ -0,0 +1,1116 @@
...
+		targeted:       makeSet(includeHeaders),
+		excluded:       makeSet(excludeHeaders),
+		targetSets:     parseTargetSets(targets),
+		emitVisibility: emitVisibility,
+		validatePaths:  splitTrimmed(validateAgainst, ","),
+		internalDirs:   splitTrimmed(internalDirsFlag, ","),
+		suppressDirs:   splitTrimmed(suppressDirsFlag, ","),
+		sourceRoot:     sourceRoot,
+	}
+
```
WillChilds-Klein, line 231: do we want to do some sort of assertion that this is running on a platform supported by `debug/elf`?

util/generate_initial_version_scripts.sh
```diff
@@ -0,0 +1,116 @@
...
+  -out /tmp/libssl_symbols.txt 2>&1
+
+echo ""
+echo "Step 3: Writing symbol registry files (${INITIAL_VERSION})..."
+
+# Write registry: "<symbol> AWS_LC_1.0 <visibility>" sorted by symbol name
+# Input format from read_public_symbols: "SYMBOL VISIBILITY"
+awk -v ver="${INITIAL_VERSION}" '{ print $1, ver, $2 }' /tmp/libcrypto_symbols.txt \
+  | sort > "${CRYPTO_REGISTRY}"
+
```
WillChilds-Klein:
```suggestion
  | sort > "${CRYPTO_REGISTRY}" \
  | sort -uc > "${CRYPTO_REGISTRY}"
```

fail if somehow duplicates got through?

util/read_public_symbols/main.go
```diff
@@ -0,0 +1,1116 @@
...
+		return nil, err
+	}
+	var filtered []string
+	for _, h := range all {
+		data, err := fs.ReadFile(fsys, h)
+		if err != nil {
+			continue
+		}
+		content := string(data)
+		if strings.Contains(content, "OPENSSL_EXPORT") {
```
WillChilds-Klein, line 612: nitpick: unlikely, but what if the macro name changes? can we figure out a symbol is exported from the ELF artifact rather than a macro invocation in a header?

util/update_symbol_version.sh
```diff
@@ -0,0 +1,149 @@
...
+#
+# When new OPENSSL_EXPORT symbols are added to public or internal headers,
+# this script:
+#   1. Extracts the current full symbol set from headers (public + internal)
+#   2. Identifies symbols not yet in the registry (new API)
+#   3. Appends them to the registry with the given version node and visibility
+#   4. Re-sorts the registry and regenerates the map files
+#
+# Usage: ./util/update_symbol_version.sh <version>
+# Example: ./util/update_symbol_version.sh AWS_LC_1.0
```
WillChilds-Klein, line 16: does this script need to update L39 in generate_initial_version_scripts.sh ?

.github/docker_images/symbol_check/check_symbols.sh
WillChilds-Klein: Agreed, dropped.

.github/workflows/abidiff.yml
WillChilds-Klein: replaced.

CMakeLists.txt
WillChilds-Klein: Implemented justin's suggestion. my concern is already taken care of on L92.

tests/ci/run_symbol_version_test.sh
WillChilds-Klein: I believe abidiff will ensure that all symbols previously exported remain exported after this PR. However, it does look like there's a gap in checking whether all exported symbols in unversioned (i.e. default) shared libs are exported with similar visibility in the versioned libs.

I'll add a check here that builds unversioned libs and diffs exported symbols with the versioned libs.

WillChilds-Klein: Added in [f34954e](https://github.com/aws/aws-lc/pull/3096/commits/f34954e389366e8757afcf5dae78cf98f7cb640e).

crypto/libcrypto.map
WillChilds-Klein: Added in [0157cb0](https://github.com/aws/aws-lc/pull/3096/commits/0157cb065c9d0b3e0e66a06a8d97277cc389e9dc) with subsequent commits to fix the jobs and add post-merge new symbols to registry + map.

tests/ci/run_symbol_version_test.sh
WillChilds-Klein: > FIPS-gated OPENSSL_EXPORT symbols won't be registered and will be localized.

do we have any such symbols, only exported in non-FIPS mode? Claude wasn't able to find any

> there are currently zero FIPS-gated OPENSSL_EXPORT symbols, so the feared class of symbol doesn't exist. I checked two ways:
>  - Ran the extractor with vs. without BORINGSSL_FIPS,AWSLC_FIPS for both libraries — identical sets (crypto 3035, ssl 638), zero difference.
>  - Scanned all public + internal headers for OPENSSL_EXPORT inside #if defined(*_FIPS) guards — none. The one FIPS guard in a public header (service_indicator.h:38) contains only
  macros, not exported symbols.

We do run tests against a versioned FIPS build (see [here](https://github.com/aws/aws-lc/pull/3096/changes#diff-5b2c9260916af13cb5c4da618da3200ac3e0170a0edd237b4157de7722cbcbb6R13)), but we don't do the artifact-only tests in `check_symbols.sh` against FIPS builds.

Shall I pull out a FIPS parameter and dimension the symbol check test matrix across that?

docs/SymbolVersioning.md
WillChilds-Klein: No, this is outdated. Removed in [0de7883](https://github.com/aws/aws-lc/pull/3096/commits/0de78834add914b7535d5c7be4ad5e926e3df52a).

docs/SymbolVersioning.md
WillChilds-Klein: I've updated the stale references in [0de7883](https://github.com/aws/aws-lc/pull/3096/commits/0de78834add914b7535d5c7be4ad5e926e3df52a), noting any concrete versions as examples ("e.g.")

cmake/GenerateVersionScript.cmake
WillChilds-Klein: **FROM CLAUDE:** Clarified the workflow in \`docs/SymbolVersioning.md\` (just rewritten): \`update_symbol_version.sh <version>\` is the normal path for adding new symbols (it takes the node as an argument, major or minor), while \`generate_initial_version_scripts.sh\` is a from-scratch bootstrap that only establishes the initial \`AWS_LC_1.0\` node. The doc now also documents the CI checks (incremental / baseline / map-drift) as the place that enforces registry updates when symbols change.

cmake/GenerateVersionScript.cmake
WillChilds-Klein: **FROM CLAUDE:** \`--version-script\` is supported by GNU ld, gold, lld, and mold (de-facto standard). The one known divergence — lld treating undefined version symbols as errors — is already handled by the \`LINKER_HAS_UNDEFINED_VERSION\` probe in \`CMakeLists.txt\`, which adds \`--undefined-version\` when supported. The build itself is the functional check: if the active linker could not apply the version script, the shared-lib link would fail.

crypto/CMakeLists.txt
WillChilds-Klein: **FROM CLAUDE:** Good catch — fixed. Since the \`.map\` files are checked-in source, a missing one under \`ENABLE_SYMBOL_VERSIONING\` should be a hard error, not a silently-unversioned library. Removed the \`if(EXISTS ...)\`/\`message(WARNING ...)\` wrapper in both \`crypto/CMakeLists.txt\` and \`ssl/CMakeLists.txt\` and now call \`apply_version_script()\` unconditionally; it already \`FATAL_ERROR\`s if the script is absent.

CMakeLists.txt
WillChilds-Klein: This is already handled: the \`if(ENABLE_DIST_PKG)\` block at \`CMakeLists.txt:92\` opens with \`if(NOT UNIX OR APPLE) message(FATAL_ERROR ...)\`, so setting \`ENABLE_DIST_PKG\` on macOS/Windows hard-fails at configure time, well before this symbol-versioning block. An \`elseif\` here would be unreachable for that case.

tests/ci/run_symbol_version_test.sh
WillChilds-Klein: **FROM CLAUDE:** Good nit — the guard was narrower than the docs claimed. Broadened the platform check to \`linux-gnu*|freebsd*|openbsd*|netbsd*\` (via a \`case\` statement) so it matches the BSD support stated in \`docs/SymbolVersioning.md\`.

tests/ci/run_symbol_version_test.sh
WillChilds-Klein: **FROM CLAUDE:** Deferring this one pending Justin's input on the FIPS question (same thread as his comment above). Investigation so far: extracting symbols with vs. without \`BORINGSSL_FIPS,AWSLC_FIPS\` yields identical sets for both libs, and no \`OPENSSL_EXPORT\` symbols are gated behind FIPS \`#if\`s today — so there is no live gap, but adding a FIPS leg would close the latent one. Will follow up based on his response.

tests/ci/run_symbol_version_test.sh
WillChilds-Klein: **FROM CLAUDE:** Added a format assertion: after extracting \`SYMBOL_VERSION\` from the map, the script now checks it matches \`^AWS_LC_[0-9]+\.[0-9]+$\` and fails otherwise, so a malformed \`.map\` is caught explicitly.

tests/ci/run_symbol_version_test.sh
WillChilds-Klein: **FROM CLAUDE:** Right on both counts. The \`|| true\` is load-bearing: \`grep -c\` exits 1 on a zero count, which under \`set -e\` would abort the script — the \`|| true\` lets a legitimate zero flow to the comparison (reported as a failure there). Added a comment explaining this. Re: L164 (Test 2) — that \`grep ... > /dev/null\` is inside an \`if\`, so \`set -e\` is suppressed for it; no \`|| true\` needed and no type confusion.

tests/ci/run_symbol_version_test.sh
WillChilds-Klein: **FROM CLAUDE:** These are sanity floors, not exact counts — they only catch a catastrophically empty/unversioned build. Registries currently hold ~3000 (libcrypto) / ~640 (libssl) symbols, and the thresholds sit well below so ordinary API churn never trips them. Pulled the magic numbers into named vars (\`CRYPTO_MIN_SYMBOLS\`/\`SSL_MIN_SYMBOLS\`) with a comment; exact accounting is left to the \`check_symbols.sh\` jobs.

tests/ci/run_symbol_version_test.sh
WillChilds-Klein: **FROM CLAUDE:** Added a doc comment breaking down the awk filter field-by-field (defined text symbol \`T\`, no \`@\` version tag, excluding \`_*\` toolchain internals and the intentionally-unversioned \`OPENSSL_memory_*\` hooks).

tests/ci/run_symbol_version_test.sh
WillChilds-Klein: **FROM CLAUDE:** Added a comment: these are ubiquitous, long-stable public symbols used as canaries to confirm concrete API is versioned (complementing the bulk count in Test 3). Any handful works and the list does not need to grow as new API is added — broad coverage comes from Test 3 and the \`check_symbols.sh\` jobs.

tests/ci/run_symbol_version_test.sh
WillChilds-Klein: **FROM CLAUDE:** Fixed. Added the missing \`else\` branch: if \`\$TEST_PROG\` is absent (compilation failed above, already recorded as a FAIL) the script now emits an explicit FAIL noting the dependent run/link/version checks could not execute, rather than silently skipping them.

util/generate_initial_version_scripts.sh
WillChilds-Klein: **FROM CLAUDE:** They are our classification, defined and parsed by the Go tooling — not GNU-ld keywords. \`ld\` only understands \`global:\`/\`local:\`/\`extern \"C++\"\`. \`generate_version_script\` maps our \`PRIVATE_CXX\`/\`*_CXX*\` values into an \`extern \"C++\"\` block and lists PUBLIC/PRIVATE under \`global:\`. The PUBLIC/PRIVATE distinction drives our removal policy (PUBLIC can never be removed), not linker behavior, so it is portable across linkers.

util/generate_version_script/main.go
WillChilds-Klein: **FROM CLAUDE:** Done. Introduced a defined \`visibility\` string type with typed constants (\`visPublic\`, \`visPrivate\`, \`visPrivateCxx\`, \`visPrivateCxxClas\`, \`visPublicCxx\`, \`visPublicCxxClass\`) — the Go idiom for enums, keeping parsing trivial since the on-disk value is the constant itself. Updated the switch and tests; regenerated maps are byte-identical.

util/generate_initial_version_scripts.sh
WillChilds-Klein: **FROM CLAUDE:** This is the bootstrap baseline and is intentionally fixed — \`generate_initial_version_scripts.sh\` only ever establishes the initial node. Subsequent bumps go through \`update_symbol_version.sh <version>\`, which takes the node as an argument rather than hardcoding it, so no shared config value is needed for the normal flow. Added a comment noting the major component corresponds to \`ABI_VERSION\` in \`CMakeLists.txt\`.

util/generate_initial_version_scripts.sh
WillChilds-Klein: **FROM CLAUDE:** Good eye on the format mismatch. Renamed the temp files to \`*_headersyms.txt\` and added a comment: the extractor output is two columns (\`SYMBOL VISIBILITY\`), and this step inserts the version node as the middle column to produce the three-column registry (\`SYMBOL VERSION VISIBILITY\`).

util/read_public_symbols/main.go
WillChilds-Klein: **FROM CLAUDE:** \`debug/elf\` is pure Go and parses ELF *files* regardless of host GOOS (you can read an ELF on macOS/Windows), so no host-platform assertion is needed. The \`-validate-against\` path only runs when a library is provided, and \`elf.Open\` returns a clean error on non-ELF input, which we already surface.

util/generate_initial_version_scripts.sh
WillChilds-Klein: **FROM CLAUDE:** Added a duplicate guard. (Note the literal suggestion pipes to the same file twice, and \`sort -uc -o\` cannot both check and write.) Implemented it as: sort into the registry, then re-verify with \`sort -c -u -k1,1\`; if a duplicate symbol slipped through, the script prints the offending names via \`uniq -d\` and aborts. Applied to both crypto and ssl.

util/read_public_symbols/main.go
WillChilds-Klein: **FROM CLAUDE:** Deliberate design tradeoff. Header scanning captures the *intended* API contract (what we promise to export), which is what the registry should track; the ELF only reflects what one particular build emitted (platform/FIPS-dependent), so keying off it would make the registry vary by build config. We do cross-check against the ELF via \`-validate-against\` (header→lib), and the new silent-drop test in \`run_symbol_version_test.sh\` compares versioned vs unversioned builds to catch a missed macro. If the macro name ever changed, \`extraExportMacros\` centralizes the set to update, and the silent-drop test would flag any regression.

tests/ci/run_symbol_version_test.sh
justsmth: > Shall I pull out a FIPS parameter and dimension the symbol check test matrix across that?

No, not worth it. No FIPS-gated exports exist today, and `run_dist_pkg_run_tests.sh` already exercises a FIPS build with the version script applied. The silent-drop test ("Test 7") would catch it if one ever appeared.

util/update_symbol_version.sh
justsmth: No -- `generate_initial_version_scripts.sh` hardcodes `INITIAL_VERSION="AWS_LC_1.0"` only as a fixed bootstrap constant and `update_symbol_version.sh` takes a version argument.

util/update_symbol_version.sh
```diff
@@ -0,0 +1,149 @@
...
+
+# Extract current symbol sets from headers (public + internal, with visibility)
+echo "Extracting current symbols from headers..."
+go run "${SOURCE_ROOT}/util/read_public_symbols" \
+  -include-dir "${SOURCE_ROOT}/include" \
+  -source-root "${SOURCE_ROOT}" \
+  -exclude ssl.h \
+  -internal-dirs crypto,third_party/jitterentropy \
+  -emit-visibility \
+  -out "${TMPDIR}/all_crypto.txt" 2>/dev/null
```
justsmth: Nit: we should drop the `2>/dev/null` (here and on line 80) -- we don't want to mask errors.

justsmth (dismissed):

util/update_symbol_version.sh
WillChilds-Klein: Good catch — dropped the `2>/dev/null` from both `read_public_symbols` invocations (lines 71 and 80) so extraction errors surface. Verified the extraction still succeeds afterward. Left the `2>&1` on the `generate_version_script` calls (L138/142) as-is since that merges stderr into stdout rather than masking it. Fixed in 6e263884b.

tests/ci/run_symbol_version_test.sh
WillChilds-Klein: Resolved: discussed with Justin and we agreed a separate FIPS symbol check isn't needed. Empirically, extracting symbols with vs. without `BORINGSSL_FIPS,AWSLC_FIPS` yields identical sets for both libraries, and no `OPENSSL_EXPORT` symbols are gated behind FIPS `#if`s today — so a FIPS build exports the same symbol set the registry already tracks. Not adding a FIPS leg to the test.

justsmth (approved):

WillChilds-Klein (approved):

util/generate_initial_version_scripts.sh
```diff
@@ -0,0 +1,145 @@
...
+# temp files are named *_headersyms to distinguish them (two columns, no
+# version) from the registry files (three columns, versioned).
+#
+# write_registry sorts the rows into the registry, then re-checks the result
+# with `sort -c` on the symbol column: if a duplicate symbol slipped through
+# (which would silently collapse an entry), the check aborts the script.
+write_registry() {
+  local src="$1" dest="$2"
+  awk -v ver="${INITIAL_VERSION}" '{ print $1, ver, $2 }' "${src}" | sort > "${dest}"
+  if ! sort -c -u -k1,1 "${dest}" 2>/dev/null; then
```
WillChilds-Klein, line 119: > do we really want to fail on duplicate symbol names, not duplicate (symbol, version) tuples? Don't we want to allow a symbol in multiple version nodes?

_(Analysis by Claude, at the author's request.)_

Deduping on symbol name (not `(symbol, version)`) is intentional and correct here. In a GNU ld version script, a **defined** symbol belongs to exactly one node — the version it was *introduced* in. Backward compatibility is provided by node **inheritance** (`AWS_LC_1.1 { ... } AWS_LC_1.0;`, meaning an app linked against 1.1 also resolves all 1.0 symbols), so an existing symbol is never re-listed in a newer node.

Listing the same symbol in two nodes is not just unnecessary — GNU ld silently mishandles it. Tested locally: a symbol `foo` placed in both `AWS_LC_1.0` and `AWS_LC_1.1` links with exit 0 and **no warning**, binding to `foo@@AWS_LC_1.0` and silently dropping the `AWS_LC_1.1` assignment. `generate_version_script` would emit the symbol in both `global:` blocks, producing exactly that mis-binding. So allowing `(symbol, version)` tuples would *permit* the one case the linker gets wrong.

The whole tooling keys on symbol name: `update_symbol_version.sh` only appends symbols **not already in the registry** (`comm -23` on names), so it structurally never adds an existing symbol to a new node. The one-row-per-symbol invariant is what the system relies on, and `sort -c -u -k1,1` enforces exactly that.

(Multiple versions of a single symbol *do* legitimately occur with `.symver` compatibility aliases in C source — but that's a separate mechanism, not expressed through these version-script `global:` lists or this registry.)

Net: no change recommended — the check is right as written.

WillChilds-Klein:
**Expected CI failures from the latest `aws/main` merge — working as intended**

_(Analysis by Claude, at the author's request.)_

The merge of `aws/main` (commit `d5d73edc3`) pulled in #3331, which added two new public **libssl** APIs — `SSL_get0_peer_certificate` and `SSL_get1_peer_certificate` (OpenSSL 3.0 peer-cert getters). These symbols are exported by the headers but are **not yet in the symbol registry** (`ssl/libssl.txt` / `ssl/libssl.map`), so the versioning CI correctly flags them. This is exactly what these jobs are designed to catch: new public API that hasn't been registered.

All four relevant failing jobs trace to these same two symbols:

- **`libssl symbol check (baseline)`** — present in the headers but missing from the registry:
  ```
  ❌ UNREGISTERED SYMBOLS (2):
  SSL_get0_peer_certificate
  SSL_get1_peer_certificate
  ```
- **`libssl release ABI check`** — abidiff vs. the last release reports them as added:
  ```
  Functions changes summary: 0 Removed, 0 Changed, 2 Added functions
    [A] 'function X509* SSL_get0_peer_certificate(const SSL*)'
    [A] 'function X509* SSL_get1_peer_certificate(const SSL*)'
  ```
- **`dist-pkg-install-tests`** — the silent-drop check flags them as exported by the compiler but hidden by `local: *;` (since they aren't in the `.map`):
  ```
  ✗ FAIL: libssl: 2 exported symbol(s) silently hidden by the version script
    SSL_get0_peer_certificate
    SSL_get1_peer_certificate
  ```
- **`dist-pkg-run-tests`** — `ssl_version_test.cc` fails to link with `undefined reference to 'SSL_get0_peer_certificate'` / `SSL_get1_peer_certificate`, because the versioned libssl doesn't export them (same root cause).

Every **libcrypto** check passes, and every **libssl** check that touches the exported symbol set fails — the expected signature of two new libssl-only symbols. The other CI failures are unrelated and non-blocking: `grpc-master` (informational, known upstream `cert_selection_offload_test` drift) and `execute` (internal security-review infra).

**Resolution:** register the two symbols by running `util/update_symbol_version.sh AWS_LC_1.0` (or regenerating) and committing the updated `ssl/libssl.txt` and `ssl/libssl.map`. The failures clear once the registry includes them.

WillChilds-Klein:
Merging now, will address any additional concerns in a follow-up.

PR #3098, opened 2026-03-12
Add -msg and -servername support to openssl s_client

geedo0:
### Description of changes: 
Add -msg flag with full implementation using SSL_set_msg_callback to
print TLS protocol messages in OpenSSL-compatible format.

Add -servername flag that maps to DoClient's existing -server-name
for SNI support.

### Call-outs:
I did not end up printing the hex dumps of each message like OpenSSL does. Should be easy enough to add if we wanted to, but the compatibility gap being addressed by this does not require it.

The `-msg` argument only applies to `s_client` but it had to be implemented inside `client.cc` because it impacts the `SSL_CTX`. This may raise the eyebrows of some purists.

### Testing:

```
# Example -msg output
➜  tool-openssl git:(msg-fix) ✗ ./openssl s_client -connect amazon.com:443 -msg -servername amazon.com | grep -E '>>>|<<<'
>>> TLS 1.2, RecordHeader [length 0005]
>>> TLS 1.2, Handshake [length 05c8], ClientHello
<<< TLS 1.2, RecordHeader [length 0005]
<<< TLS 1.2, Handshake [length 007a], ServerHello
>>> TLS 1.2, RecordHeader [length 0005]
>>> TLS 1.3, ChangeCipherSpec [length 0001]
<<< TLS 1.2, RecordHeader [length 0005]
<<< TLS 1.2, RecordHeader [length 0005]
<<< TLS 1.3, Handshake [length 000a], EncryptedExtensions
<<< TLS 1.2, RecordHeader [length 0005]
<<< TLS 1.3, Handshake [length 144c], Certificate
<<< TLS 1.2, RecordHeader [length 0005]
<<< TLS 1.3, Handshake [length 0108], CertificateVerify
<<< TLS 1.2, RecordHeader [length 0005]
<<< TLS 1.3, Handshake [length 0024], Finished
>>> TLS 1.3, Handshake [length 0024], Finished
>>> TLS 1.2, RecordHeader [length 0005]
<<< TLS 1.2, RecordHeader [length 0005]
<<< TLS 1.3, Handshake [length 0039], NewSessionTicket
<<< TLS 1.2, RecordHeader [length 0005]
<<< TLS 1.3, Handshake [length 0039], NewSessionTicket
<<< TLS 1.2, RecordHeader [length 0005]
<<< TLS 1.3, Alert [length 0002]
```


tool-openssl/s_client.cc
```diff
@@ -34,6 +34,10 @@ static const argument_t kArguments[] = {
                 "Use TLS version 1.2 only." },
         { "-tls1_3", kBooleanArgument,
                 "Use TLS version 1.3 only." },
+        { "-msg", kBooleanArgument,
+                "No-op, accepted for compatibility." },
```
justsmth: This claims it's a "No-op", but you actually implemented logic for it.

tool/client.cc
```diff
@@ -310,6 +310,94 @@ static void PrintOpenSSLConnectionInfo(SSL *ssl, bool show_certs) {
...
+      return "CertificateVerify";
+    case SSL3_MT_FINISHED:
+      return "Finished";
+    case SSL3_MT_ENCRYPTED_EXTENSIONS:
+      return "EncryptedExtensions";
+    case SSL3_MT_NEW_SESSION_TICKET:
+      return "NewSessionTicket";
+    default:
+      return nullptr;
+  }
```
justsmth, line 367: * What about `ServerKeyExchange`, `ServerHelloDone`, `ClientKeyExchange`, others?
* Should the default be something like "Unknown" or "Other" instead of `nullptr`?

tool-openssl/s_client.cc
geedo0: Good catch, it started out like that but I figured we could do a minimal implementation.

tool/client.cc
geedo0: `nullptr` ends up printing nothing after the length which I think is fine. Didn't feel like being exhaustive here unless we feel strongly about it. I implemented enough to avoid any unhandled cases for a typical endpoint.

`>>> TLS 1.2, Handshake [length 05c8]`

tool/client.cc
```diff
@@ -310,6 +310,94 @@ static void PrintOpenSSLConnectionInfo(SSL *ssl, bool show_certs) {
...
+    case SSL3_MT_CERTIFICATE_VERIFY:
+      return "CertificateVerify";
+    case SSL3_MT_FINISHED:
+      return "Finished";
+    case SSL3_MT_ENCRYPTED_EXTENSIONS:
+      return "EncryptedExtensions";
+    case SSL3_MT_NEW_SESSION_TICKET:
+      return "NewSessionTicket";
+    default:
+      return nullptr;
```
WillChilds-Klein, line 366: np: others return `"Unknown"` here, should this do same?

geedo0: See thread with @justsmth below

justsmth:
> I wanted to, but I couldn't find any patterns in the test framework for capturing stdout and asserting on that. Is there something I'm missing?  ...

We have a bunch of "comparison tests" that capture our command output and compare it to the output of openssl.  See the usage of [RunCommandsAndCompareOutput](https://github.com/aws/aws-lc/blob/34b4a026b4b33e19ccf9b623b7955da957df4fd2/tool-openssl/test_util.h#L61). Could that be used?

WillChilds-Klein (approved):

PR #3101, opened 2026-03-18
Reject point at infinity in EC_KEY_set_public_key

nebeid:
### Issues:
Addresses P398529487

### Description of changes: 
Add an explicit check in EC_KEY_set_public_key to reject the point at infinity. The point at infinity is not a valid public key.

This matches the hardening added in BoringSSL https://github.com/google/boringssl/commit/a135fe1fad70e4f87b38b0adebd44d93da7eb173.

### Call-outs:
Parsing (EC_POINT_oct2point) continues to accept the infinity encoding for OpenSSL compatibility, as added in
de33f5ecbe8171aed56ba542db2c7246d2ccaf3c, but it can no longer be installed as a public key on an EC_KEY.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ cmake_policy(SET CMP0091 NEW)
 endif()
 
 set(SOFTWARE_NAME "awslc")
-set(SOFTWARE_VERSION "1.70.0")
+set(SOFTWARE_VERSION "1.71.0")
 set(ABI_VERSION 0)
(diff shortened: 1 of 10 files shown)


WillChilds-Klein (dismissed):

alexw91:
You should probably also update the API documentation too like BoringSSL did:
https://github.com/google/boringssl/blob/a135fe1fad70e4f87b38b0adebd44d93da7eb173/include/openssl/ec_key.h#L90-L91

PR #3114, opened 2026-03-20
Add SSL_use_cert_and_key for per-connection cert/key setting

geedo0:
### Issues:
Addresses P400736589

### Description of changes: 

Postgres recently started calling SSL_use_cert_and_key (the per-SSL
variant) which AWS-LC did not implement. This broke the postgres
integration test on both x86_64 and aarch64 with an undefined reference
linker error.

Extract the body of SSL_CTX_use_cert_and_key into a static helper
cert_use_cert_and_key(CERT *) and add SSL_use_cert_and_key as a thin
wrapper, following the established pattern used by SSL_use_PrivateKey,
ssl_set_cert, and cert_set_chain_and_key.

### Testing:

CI

### DIFF
--- a/crypto/ec_extra/ec_asn1.c
+++ b/crypto/ec_extra/ec_asn1.c
@@ -562,6 +562,10 @@ EC_KEY *o2i_ECPublicKey(EC_KEY **keyp, const uint8_t **inp, long len) {
     OPENSSL_PUT_ERROR(EC, ERR_R_EC_LIB);
     return NULL;
   }
+  if (EC_POINT_is_at_infinity(ret->group, ret->pub_key)) {
+    OPENSSL_PUT_ERROR(EC, EC_R_POINT_AT_INFINITY);
+    return NULL;
+  }
   // save the point conversion form
   ret->conv_form = (point_conversion_form_t)(*inp[0] & ~0x01);
   *inp += len;
--- a/crypto/fipsmodule/ec/ec_key.c
+++ b/crypto/fipsmodule/ec/ec_key.c
@@ -206,6 +206,11 @@ int EC_KEY_set_public_key(EC_KEY *key, const EC_POINT *pub_key) {
     return 0;
   }
 
+  if (pub_key != NULL && EC_POINT_is_at_infinity(key->group, pub_key)) {
+    OPENSSL_PUT_ERROR(EC, EC_R_POINT_AT_INFINITY);
+    return 0;
+  }
+
   if (pub_key != NULL && EC_GROUP_cmp(key->group, pub_key->group, NULL) != 0) {
     OPENSSL_PUT_ERROR(EC, EC_R_GROUP_MISMATCH);
     return 0;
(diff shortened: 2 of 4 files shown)


include/openssl/ssl.h
```diff
@@ -1224,6 +1224,25 @@ OPENSSL_EXPORT int SSL_CTX_use_cert_and_key(SSL_CTX *ctx, X509 *x509,
...
+// SSL_use_cert_and_key sets |x509|, |privatekey|, and |chain| on |ssl|.
+// The |privatekey| argument must be the private key of the certificate |x509|.
+// If the override argument is 0, then |x509|, |privatekey|, and |chain| are
+// set only if all were not previously set. If override is non-0, then the
+// certificate, private key and chain certs are always set. |privatekey| and
+// |x509| are not copied or duplicated, their reference counts are incremented.
+// In OpenSSL, a shallow copy of |chain| is stored with a reference count
+// increment for all |X509| objects in the chain. In AWS-LC,
+// we represent X509 chains as a CRYPTO_BUFFER stack. Therefore, we create a
+// an internal copy and leave the |chain| parameter untouched. This means,
```
justsmth, line 1236: NP: "... We create **a an** internal copy ..."

ssl/ssl_cert.cc
```diff
@@ -1001,6 +999,21 @@ int SSL_CTX_use_cert_and_key(SSL_CTX *ctx, X509 *x509, EVP_PKEY *privatekey,
...
+                             STACK_OF(X509) *chain, int override) {
+  return cert_use_cert_and_key(ctx->cert.get(), x509, privatekey, chain,
+                               override);
+}
+
+int SSL_use_cert_and_key(SSL *ssl, X509 *x509, EVP_PKEY *privatekey,
+                         STACK_OF(X509) *chain, int override) {
+  if (!ssl->config) {
+    return 0;
+  }
```
justsmth, line 1012: NP: should we set an error here?
```
OPENSSL_PUT_ERROR(SSL, ERR_R_PASSED_NULL_PARAMETER);
```

ssl/ssl_cert.cc
justsmth, line 256: NP: Maybe for a future PR -- since `cert_pkey->chain` is being replaced here, shouldn't the cache be invalidated? Otherwise a caller that previously triggered the cache population would see stale data after this call.
```
  if (cert_pkey->chain) {
    cert_pkey->chain.reset();
  }
  cert_pkey->chain = std::move(certs);

  // Invalidate the parsed X509 chain cache for this slot since the backing
  // CRYPTO_BUFFER chain was just replaced.
  sk_X509_pop_free(cert_pkey->x509_chain, X509_free);
  cert_pkey->x509_chain = nullptr;

  cert->cert_private_key_idx = slot_idx;
```

justsmth: See: https://github.com/aws/aws-lc/pull/3117

PR #3117, opened 2026-03-23
ssl: invalidate X509 leaf/chain caches in cert_set_chain_and_key and …

justsmth:
### Description of changes: 
`CERT_PKEY` lazily caches parsed `X509` objects in `x509_leaf` and `x509_chain`, which are populated on first access via `SSL_CTX_get0_certificate` and `SSL_CTX_get0_chain_certs`. Neither cache was being invalidated when the backing `CRYPTO_BUFFER` chain was replaced, which means a caller who reads the certificate back after replacing it would silently receive the old cached value.

This affects two code paths:
- `cert_set_chain_and_key` (`SSL_CTX_set_chain_and_key` / `SSL_set_chain_and_key`): replaces the entire chain via `std::move`, leaving both caches stale.
- `ssl_set_cert` (`SSL_CTX_use_certificate_ASN1` / `SSL_use_certificate_ASN1`): swaps the leaf `CRYPTO_BUFFER` in-place within an existing chain, leaving `x509_leaf` stale. Note that the `SSL_CTX_use_certificate` (X509-based) path was not affected because `ssl_use_certificate` explicitly reassigns `x509_leaf` after calling `ssl_set_cert`.

The fix calls `cert_flush_leaf` and `cert_flush_cached_chain` (as appropriate) through `x509_method` after each replacement. This correctly handles both the full X509 case (`ssl_crypto_x509_method`) and the no-op case (`ssl_noop_x509_method` / `TLS_with_buffers_method`).

Two regression tests are added that warm the caches before replacing the certificate and assert that subsequent getter calls reflect the new certificate rather than the old cached one.

### DIFF
--- a/ssl/ssl_test.cc
+++ b/ssl/ssl_test.cc
@@ -1210,6 +1210,111 @@ TEST(SSLTest, SetChainAndKey) {
                                      server_ctx.get()));
 }
 
+// Verify that SSL_CTX_set_chain_and_key invalidates the X509 leaf and chain
+// caches so that SSL_CTX_get0_certificate and SSL_CTX_get0_chain_certs reflect
+// the newly-configured certificate chain rather than stale cached values.
+TEST(SSLTest, SetChainAndKeyCacheInvalidation) {
+  bssl::UniquePtr<SSL_CTX> ctx(SSL_CTX_new(TLS_method()));
+  ASSERT_TRUE(ctx);
+
+  // Configure an RSA leaf + intermediate and warm up both the x509_leaf and
+  // x509_chain caches.
+  bssl::UniquePtr<EVP_PKEY> key1 = GetChainTestKey();
+  ASSERT_TRUE(key1);
+  bssl::UniquePtr<CRYPTO_BUFFER> leaf1 = GetChainTestCertificateBuffer();
+  ASSERT_TRUE(leaf1);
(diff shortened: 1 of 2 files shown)


ssl/ssl_test.cc
samuel40791765, line 1: Are there any existing tests we can just use to insert a few functions and validate the new behavior?

justsmth: The existing `SetChainAndKey` test uses `TLS_with_buffers_method()` which has the noop x509_method -- there's no X509 cache layer to exercise there, so we'd need a separate test with `TLS_method()` regardless.

For the ASN1 path, `GetCertificateASN1` already sets up the same initial state, so I could append the replacement + re-read steps there instead of a separate test. Is that what you would prefer?

geedo0 (approved):

ssl/ssl_test.cc
samuel40791765: Yeah appending it was my preference, not a blocker though.

justsmth: Looking at the existing tests in `ssl_test.cc`, I don't think appending to an existing test is the right approach for the ASN1 path. The current approach of dedicated cache invalidation tests is cleaner and makes it obvious what behavior is being verified.

samuel40791765 (approved):

PR #3120, opened 2026-03-24
openssl-tool CLI: CA cleanup

justsmth:
### Issues:
Resolves AWS-LC-1036

### Description of changes:
- Fix unchecked `EVP_get_digestbyname()` return in `req.cc` — an unrecognized digest name from a config file would pass NULL into `X509_sign`.
- Change `std::string` pass-by-value parameters to `const std::string &` across 15 static functions in `ca.cc` that only read the string.
- Add defensive NULL checks after `sk_OPENSSL_PSTRING_value()` in `txt_db.cc`.
- Check `rename()` return values in error-recovery paths and log secondary failures.
- Fix `sizeof(OPENSSL_STRING *)` → `sizeof(OPENSSL_STRING)` in the `irow` allocation (same size in practice, but matches intent).

### Call-outs:
- `ParseSubjectName` signature changed from `std::string &` to `const std::string &` in both `ca_req_common.h` and `internal.h`. The function only called `.c_str()` on the parameter.
- All changes are scoped to `tool-openssl/`. No crypto library code is affected.

### Testing:
Existing `CA*`, `Req*`, and `SubjectName*` tests pass.

### DIFF
--- a/tool-openssl/ca.cc
+++ b/tool-openssl/ca.cc
@@ -156,8 +156,8 @@ ok:
 #endif
 
 static std::unique_ptr<std::string> GetSectionValue(bssl::UniquePtr<CONF> &conf,
-                                                    std::string section,
-                                                    std::string key) {
+                                                    const std::string &section,
+                                                    const std::string &key) {
   auto *value = NCONF_get_string(conf.get(), section.c_str(), key.c_str());
   if (value == NULL) {
     return std::unique_ptr<std::string>(nullptr);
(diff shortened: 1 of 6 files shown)


tool-openssl/ca.cc
```diff
@@ -1257,7 +1271,7 @@ static int DoBody(bssl::UniquePtr<BIO> &bio_err, X509 **xret,
     goto end;
   }
 
-  irow.reset((OPENSSL_STRING *)OPENSSL_malloc(sizeof(OPENSSL_STRING *) *
+  irow.reset((OPENSSL_STRING *)OPENSSL_malloc(sizeof(OPENSSL_STRING) *
```
geedo0, line 1274: Wow, nice catch from the robots.

PR #3124, opened 2026-03-26
`WIN32_rename`: fix errno mapping and increase retry budget for transient failures

justsmth:
### Issues:
Follow-up to #3100. Resolves intermittent `CATest.CopyExtensionsNone` failure on `windows-ltsc2022_sde-x86_64` CI job.

### Context:
PR #3100 added a retry loop in `WIN32_rename` for transient file lock errors, but the CI is still producing intermittent rename failures under SDE. The root cause is that `ERROR_SHARING_VIOLATION` is correctly retried but, if retries are exhausted, is misreported as `EINVAL` ("Invalid argument") due to a gap in the post-loop errno mapping. The 1-second retry window (5 × 200ms) is also too short for the SDE environment under parallel load.

### Description of Changes:
- Increase `kMaxRetries` from 5 to 10 (~2s total retry window)
- Add `ERROR_LOCK_VIOLATION` to the set of retried error codes
- Fix errno mapping so `ERROR_SHARING_VIOLATION` / `ERROR_LOCK_VIOLATION` map to `EACCES` instead of falling through to `EINVAL`
- Log the actual Windows error code on unexpected failures for future diagnostics

Scoped entirely to `tool-openssl/ca.cc`. No crypto library code is affected.

### DIFF
--- a/tool-openssl/ca.cc
+++ b/tool-openssl/ca.cc
@@ -97,9 +97,11 @@ static int WIN32_rename(const char *from, const char *to) {
   int ret = 0;
   // On Windows, antivirus software, file indexing services, or other
   // background processes can briefly lock files, causing transient
-  // ERROR_ACCESS_DENIED or ERROR_SHARING_VIOLATION failures. Retry with a
-  // short delay to handle these.
-  static const int kMaxRetries = 5;
+  // ERROR_ACCESS_DENIED, ERROR_SHARING_VIOLATION, or ERROR_LOCK_VIOLATION
+  // failures. Retry with a short delay to handle these. Under heavy load
+  // (e.g. running under SDE with parallel tests), locks can persist for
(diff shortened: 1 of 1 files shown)


tool-openssl/ca.cc
```diff
@@ -97,9 +97,11 @@ static int WIN32_rename(const char *from, const char *to) {
   int ret = 0;
   // On Windows, antivirus software, file indexing services, or other
```
geedo0, line 98: Dumb question, but can we just not have antivirus, file indexing, etc, software running on these hosts? Or is the call being made here that figuring that out is just as much work and we'll probably still need these retries?

justsmth: I'm not sure we even have easy access to reconfigure the OS on these hosts, at least not in a maintainable way.

And if it happens in our tests, it could happen to consumers.

PR #3128, opened 2026-03-26
Exclude OCSPIntegrationTest from normal CI test runs

justsmth:
### Desccription of Changes:
`OCSPIntegrationTest` connects to live Amazon Trust Services OCSP responder endpoints, which are occasionally unavailable and cause spurious CI failures. This PR removes the `integration_test` binary from all normal CI test paths and moves it to a dedicated workflow that runs on every PR, on a daily schedule, and via manual dispatch. As a separate workflow, failures due to external endpoint issues will not block other CI checks.

### Call-outs:
There were four independent places where `integration_test` was invoked, all of which are removed:
- `util/all_tests.json` — consumed by the `run_tests` CMake target (the primary test path)
- `tests/ci/run_bsd_tests.sh` — direct invocation for BSD CI (this also had a NetBSD-specific `GTEST_FILTER` workaround for the same problem, which is now removed)
- `tests/ci/run_cross_tests.sh` — direct invocation for QEMU cross-compilation CI
- `tests/ci/run_cross_mingw_tests.sh` — direct invocation for MinGW cross-compilation CI

The `integration_test` binary also contains `OCSPRoundTripTest`, which uses a local mock and doesn't depend on external endpoints. It will now only run in the dedicated workflow as well.

### Testing:
The new workflow (`.github/workflows/ocsp-external-tests.yml`) can be verified via `workflow_dispatch`.

### DIFF
--- a/.github/workflows/ocsp-external-tests.yml
+++ b/.github/workflows/ocsp-external-tests.yml
@@ -1,8 +1,7 @@
 name: OCSP integration tests
 on:
-  schedule:
-    # Run daily at 8 AM UTC to detect external endpoint issues early.
-    - cron: "0 8 * * *"
+  pull_request:
+    branches: ["*"]
   workflow_dispatch:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name }}


tests/ci/run_bsd_tests.sh
```diff
@@ -28,7 +23,7 @@ if [ "$PLATFORM" != "amd64" ] && [ "$PLATFORM" != "x86_64" ]; then
     shard_gtest ${BUILD_ROOT}/crypto/tree_drbg_jitter_entropy_isolated_test
 
     shard_gtest ${BUILD_ROOT}/ssl/ssl_test
-    shard_gtest ${BUILD_ROOT}/ssl/integration_test
+    shard_gtest "${BUILD_ROOT}/ssl/integration_test --gtest_filter=-All/OCSPIntegrationTest.*"
```
samuel40791765: This probably isn't doing anything, so we can probably just remove this.

util/all_tests.json
```diff
@@ -124,7 +124,8 @@
     "shard": true
   },
   {
-    "cmd": ["ssl/integration_test"],
+    "comment": "OCSPIntegrationTest depends on external OCSP responder endpoints and is excluded here. It runs in a dedicated CI job.",
+    "cmd": ["ssl/integration_test", "--gtest_filter=-All/OCSPIntegrationTest.*"],
```
samuel40791765: Can't we just remove this all together? This probably isn't doing anything, I believe all the tests under the executable have the same tag.

util/all_tests.json
```diff
@@ -124,7 +124,8 @@
     "shard": true
   },
   {
-    "cmd": ["ssl/integration_test"],
+    "comment": "OCSPIntegrationTest depends on external OCSP responder endpoints and is excluded here. It runs in a dedicated CI job.",
```
nebeid: Maybe add this comment in the other locations as well.

tests/ci/run_cross_mingw_tests.sh
samuel40791765, line 1: We can probably just remove the line references to the executable and use the new CI run to test.

PR #3129, opened 2026-03-27
Hardening fixes for ML-DSA digest mode, XTS key comparison, and urandom fd

justsmth:
### Issues
* Addresses: P398529487

### Description of changes:

Several small hardening fixes:

- **ML-DSA digest length validation**: Added `digest_len` to the `PQDSA` struct and enforce that the `mu` input to ML-DSA sign/verify digest mode is exactly `MLDSA_CRHBYTES` (64 bytes). Previously the digest-mode paths accepted any `message_len` and passed it directly to the underlying implementation without validation.
- **PQDSA signature buffer check**: Changed the `sig_len` check from `!=` to `<` so callers providing a larger-than-needed buffer aren't rejected. This is consistent with standard OpenSSL buffer conventions.
- **XTS key comparison**: Switched the XTS duplicate-key check from `OPENSSL_memcmp` to `CRYPTO_memcmp` to avoid timing leakage on key material.
- **urandom fd initialization**: Changed `urandom_fd` from `0` to `-1` so that any use before initialization fails with `EBADF` rather than silently reading from stdin.

### Call-outs:

The `digest_len` field in `pqdsa.c` is set to the literal `64` rather than `MLDSA_CRHBYTES` because the macro is not available in that translation unit (due to `bcm.c` undef behavior). Comments in the code explain this.

### Testing:

Existing test coverage for ML-DSA sign/verify, XTS, and RAND should exercise these code paths. The ML-DSA digest length checks are new validation — negative test cases that pass an incorrect `mu` length could be a useful follow-up.

### DIFF
--- a/crypto/fipsmodule/cipher/e_aes.c
+++ b/crypto/fipsmodule/cipher/e_aes.c
@@ -565,7 +565,7 @@ static int aes_xts_init_key(EVP_CIPHER_CTX *ctx, const uint8_t *key,
     //
     // key_len is two AES keys
 
-    if (OPENSSL_memcmp(key, key + ctx->key_len / 2, ctx->key_len / 2) == 0) {
+    if (CRYPTO_memcmp(key, key + ctx->key_len / 2, ctx->key_len / 2) == 0) {
       OPENSSL_PUT_ERROR(CIPHER, CIPHER_R_XTS_DUPLICATED_KEYS);
       return 0;
     }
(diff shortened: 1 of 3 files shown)


crypto/fipsmodule/pqdsa/pqdsa.c
```diff
@@ -301,6 +301,7 @@ DEFINE_LOCAL_DATA(PQDSA, sig_ml_dsa_44) {
   out->signature_len = MLDSA44_SIGNATURE_BYTES;
   out->keygen_seed_len = MLDSA44_KEYGEN_SEED_BYTES;
   out->sign_seed_len = MLDSA44_SIGNATURE_SEED_BYTES;
+  out->digest_len = 64; // MLDSA_CRHBYTES, unavailable here due to bcm.c undef
```
jakemas, line 304: Looks good to me. I see no need to statically assert `MLDSA_CRHBYTES == 64` anywhere as this has to match the out put length of fixed shake256.

PR #3130, opened 2026-03-27
Small fixes for RSA_METHOD and EVP_PKEY_derive_set_peer

dkostic:
### Issues:
N/A

### Description of changes: 
Fix RSA_METHOD parameter mismatch and use-after-free in EVP_PKEY_derive_set_peer

Custom RSA_METHOD callbacks (sign_raw, verify_raw, encrypt, decrypt) were receiving max_out (output buffer size) as their first parameter instead of in_len (input data length). This can cause callbacks to read past the input buffer by up to (RSA_size - actual_input_length) bytes. Fixed all 4 dispatch sites to pass in_len, matching OpenSSL's flen semantics for RSA_private_encrypt/RSA_public_decrypt/etc.

EVP_PKEY_derive_set_peer had a use-after-free when called with the same peer key already set and a refcount of 1: EVP_PKEY_free freed the key before EVP_PKEY_up_ref took a new reference. Reordered to up_ref before free, and added proper cleanup in the ctrl failure path.

### Call-outs:

### Testing:

### DIFF
--- a/crypto/fipsmodule/evp/evp_ctx.c
+++ b/crypto/fipsmodule/evp/evp_ctx.c
@@ -366,17 +366,20 @@ int EVP_PKEY_derive_set_peer(EVP_PKEY_CTX *ctx, EVP_PKEY *peer) {
     return 0;
   }
 
+  // Take a reference to the new peer before freeing the old one, in case
+  // peer == ctx->peerkey and the caller holds the sole reference.
+  EVP_PKEY_up_ref(peer);
   EVP_PKEY_free(ctx->peerkey);
   ctx->peerkey = peer;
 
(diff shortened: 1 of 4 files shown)


crypto/fipsmodule/evp/evp_ctx.c
```diff
@@ -366,17 +366,20 @@ int EVP_PKEY_derive_set_peer(EVP_PKEY_CTX *ctx, EVP_PKEY *peer) {
     return 0;
   }
 
+  // Take a reference to the new peer before freeing the old one, in case
+  // peer == ctx->peerkey and the caller holds the sole reference.
```
WillChilds-Klein, line 370: nit: can we add a test case exercising the previous UAF here?

WillChilds-Klein (approved):

justsmth (approved):

PR #3131, opened 2026-03-27
Upgrade custom libc++ to LLVM 19 and add sanitizer support to `build_and_test.sh`

justsmth:
### Description of changes:
Currently, the custom libc++ used for MSAN/TSAN-instrumented builds is pinned to LLVM 15, and running sanitizer builds requires manually assembling the correct CMake flags. Additionally, the no-ASM code paths are not tested under sanitizers in CI.

This change upgrades the custom libc++ from LLVM 15 to LLVM 19.1.7, adds first-class `--sanitizer` support to `util/build_and_test.sh`, and introduces new CI jobs that test the no-ASM code paths under ASan and MSan.

**LLVM 19 upgrade:**
- Update `Dockerfile.al2023` to clone LLVM 19.1.7 (from 15.0.6)
- Bump `CXX_STANDARD` from 20 to 23 for the custom libc++ build (LLVM 19 added `expected.cpp` which requires C++23)
- Expand `util/bot/libcxx-config/__config_site` with required LLVM 19+ configuration macros (`_LIBCPP_HARDENING_MODE`, positive-sense feature flags, thread API selection, ASan instrumentation flag)
- Add `util/bot/libcxx-config/__assertion_handler` (required by LLVM 19+ which no longer ships a default)
- Update `linux-multi-arch-omnibus.yml` to use the unversioned `setup-clang.sh`

**Sanitizer support in `build_and_test.sh`:**
- Add `--sanitizer <name>` flag supporting `asan`, `msan`, `tsan`, `ubsan`, and `cfi`
- Add `--test <binary>` flag to build and run a single test binary directly
- Auto-clone LLVM (sparse, cached) when MSAN/TSAN need a custom libc++ and `LLVM_PROJECT_HOME` is not set
- Clean the build directory on sanitizer runs to avoid stale configuration artifacts

**Bug fix:**
- Zero-initialize `res` and `tmp` arrays in `ec_nistp_scalar_mul` to eliminate MSAN false positives on unused limbs

### Call-outs:
The new `sanitizer-tests` job in `actions-ci.yml` runs ASan and MSan with `-DOPENSSL_NO_ASM=1` on x86-64 and aarch64. These jobs do **not** overlap with the existing `sanitizers` job in `linux-multi-arch-omnibus.yml`, which runs all five sanitizers (ASan, MSan, TSan, UBSan, CFI) with ASM enabled inside Docker containers on CodeBuild via `run_posix_sanitizers.sh`. The new jobs specifically cover the no-ASM code paths under sanitizers, which were previously untested in CI.

### Testing:
- New `sanitizer-tests` CI matrix covers ASan + MSan (no-ASM) × {x86-64, aarch64}
- Existing Docker-based omnibus sanitizer jobs updated to use LLVM 19

### DIFF
--- a/util/build_and_test.sh
+++ b/util/build_and_test.sh
@@ -123,7 +123,7 @@ cmake --build "${AWS_LC_BUILD}" -j --target ${BUILD_TARGET}
 if [[ -n "${TEST_BINARY}" ]]; then
   # --test was specified: find and run the test binary directly instead of
   # going through the heavyweight run_tests CMake target.
-  TEST_PATH=$(find "${AWS_LC_BUILD}" -name "${TEST_BINARY}" -type f | head -1)
+  TEST_PATH=$(find "${AWS_LC_BUILD}" -name "${TEST_BINARY}" -type f -executable | head -1)
   if [[ -z "${TEST_PATH}" ]]; then
     echo "Error: Could not find test binary '${TEST_BINARY}' in ${AWS_LC_BUILD}"
     exit 1


crypto/fips_callback_test.cc
```diff
@@ -106,8 +106,8 @@ TEST(FIPSCallback, PowerOnSelfTests) {
   uint8_t private_key[ED25519_PRIVATE_KEY_LEN];
   ED25519_keypair(public_key, private_key);
 
-  uint8_t message[2];
-  uint8_t context[2];
+  uint8_t message[2] = {0};
+  uint8_t context[2] = {0};
   uint8_t signature[ED25519_SIGNATURE_LEN];
```
WillChilds-Klein, line 111: should we also zeroize `signature`?

CMakeLists.txt
```diff
@@ -1174,8 +1174,9 @@ if(USE_CUSTOM_LIBCXX)
   set_target_properties(
     libcxx libcxxabi PROPERTIES
     COMPILE_FLAGS "-Wno-missing-prototypes -Wno-implicit-fallthrough"
-    # libc++ and libc++abi must be built in C++20 mode.
-    CXX_STANDARD 20
+    # libc++ and libc++abi must be built in C++23 mode.
+    # LLVM 19+ added expected.cpp which requires C++23.
+    CXX_STANDARD 23
```
WillChilds-Klein, line 1179: does this mean we can no longer build AWS-LC with older versions of clang that don't support C++23? if so, is there relevant documentation we need to update?

util/build_and_test.sh
```diff
@@ -13,8 +13,38 @@ DEFAULT_CMAKE_FLAGS=("-DCMAKE_BUILD_TYPE=Debug" "-GNinja")
...
+    echo "Using cached LLVM project at ${llvm_cache_dir}"
+  else
+    echo "LLVM_PROJECT_HOME not set. Cloning llvm-project to ${llvm_cache_dir}..."
+    rm -rf "${llvm_cache_dir}"
+    git clone https://github.com/llvm/llvm-project.git \
+      --branch llvmorg-19.1.7 --depth 1 \
+      --filter=blob:none --sparse \
+      "${llvm_cache_dir}"
+    pushd "${llvm_cache_dir}"
+    git sparse-checkout set libcxx libcxxabi
```
WillChilds-Klein, line 42: you don't need to actually build LLVM?

util/build_and_test.sh
```diff
@@ -37,13 +76,60 @@ while [[ $# -gt 0 ]]; do
...
 cmake "${BASE_DIR}" -B "${AWS_LC_BUILD}" "${DEFAULT_CMAKE_FLAGS[@]}" "${CMAKE_ARGS[@]}"
 
 cmake --build "${AWS_LC_BUILD}" -j --target ${BUILD_TARGET}
 
-# Only run tests if --build-only was not specified
-if [ "$BUILD_ONLY" = false ]; then
+if [[ -n "${TEST_BINARY}" ]]; then
+  # --test was specified: find and run the test binary directly instead of
+  # going through the heavyweight run_tests CMake target.
+  TEST_PATH=$(find "${AWS_LC_BUILD}" -name "${TEST_BINARY}" -type f | head -1)
```
WillChilds-Klein:
```suggestion
  TEST_PATH=$(find "${AWS_LC_BUILD}" -name "${TEST_BINARY}" -type f -executable | head -1)
```

crypto/fips_callback_test.cc
justsmth: `signature` is solely an output buffer. Its value is only written to, so out of scope for MSAN which is concerned about values read prior to having been written.

CMakeLists.txt
justsmth: This is inside the `if(USE_CUSTOM_LIBCXX)` block, so the C++23 requirement only applies when building the custom libc++ (i.e., sanitizer builds). AWS-LC is otherwise unaffected.

util/build_and_test.sh
justsmth: Correct -- the `libc++` and `libc++abi` sources are compiled directly as part of our CMake build (the `libcxx` and `libcxxabi` targets in `CMakeLists.txt` under `USE_CUSTOM_LIBCXX`). 

We just need the source tree, not a pre-built LLVM. The sparse clone grabs only `libcxx/` and `libcxxabi/`, which is all that CMake needs for the build.

util/build_and_test.sh
justsmth: Good call! Will update.

PR #3133, opened 2026-03-29
use /map: linker flag to avoid running a binary to capture the hash

### LINKED ISSUE #3021: Building fips compliant dll on Windows targets from an Ubuntu machine
I've been playing with my team's CI, currently we cross-build windows from Ubuntu with a few exceptions running on a Windows VM. One blocker I've had is transferring aws-lc-fips.dll build from being in the windows side to the ubuntu side. There were a few approaches but i think - supporting cross compilation is the best one so far.

Clearly it is not supported atm - you have to run the .exe in order to get the correct hash via capture_hash.go.

I have a solution that doesn't require running anything, it is actually quite elegant!
Could use:
https://learn.microsoft.com/en-us/cpp/build/reference/map-generate-mapfile?view=msvc-170
This generates a mapping between symbols names to .dll file offsets as an output of the build, then you can just use this map to get the offsets of:
BORINGSSL_bcm_text_start, BORINGSSL_bcm_text_end
BORINGSSL_bcm_rodata_start, BORINGSSL_bcm_rodata_end
and calculate the hashes without ever having to run fips_empty_main against the fips dll!

The solution is both elegant and more widely supported - it will require adding .dll parsing code in inject_hash.go like other platforms do!

I was able to vibe a solution pretty quickly, i think after going over dead code i'd be able to wipe deleting more code than the newly added code - but before continuing to a pull request with a full solution i'd like to get your approval - is that a step in the right direction for you? Do you accept such pull requests at a basic level?

yoavwizstein:
### Issues:
Resolves #3021

### Description of changes:
Replaces the Windows FIPS hash-injection mechanism. Previously, the build linked and *ran* `fips_empty_main.exe` to trigger the integrity self-test, captured the expected hash from its stderr, then patched it into `crypto.dll`. This required a Windows (or Wine) execution environment at build time, blocking cross-compilation.

Now Windows uses the same approach as Linux/macOS: the linker produces a map file (`/MAP:`), and `inject_hash.go -windows` parses it alongside the PE to locate the FIPS module boundaries, compute the hash, and patch the DLL directly. No binary execution needed.

### Call-outs:
- `fips_empty_main.c` and `capture_hash/capture_hash.go` are deleted entirely.
- `break-hash.go` is extended to handle PE via the new `fipscommon` package (shared with `inject_hash.go`).
- The `/Fo:` → `/Fo` change in `fipsmodule/CMakeLists.txt` drops the colon separator for compatibility with `clang-cl`.
- `lib.exe` invocation now uses a response file (`bcm_objects.rsp`) to avoid command-line length limits.
- `lib.exe` discovery uses `find_program` instead of assuming `CMAKE_AR` or a hardcoded path relative to the linker, improving robustness with clang-cl toolchains.

### Testing:
The Windows CI (`run_windows_tests.bat`) now runs a dedicated `fips_build_and_test` routine for FIPS dimensions that performs both a positive test (`test_fips.exe` loads `crypto.dll` and passes the integrity check + KATs) and a negative test (corrupts the FIPS text region via `break-hash.go -map`, then confirms `test_fips.exe` aborts). The cross-compilation path is exercised in the `aws-lc-rs.yml` workflow which builds with `cargo xwin` and runs the FIPS sanity test under Wine.


justsmth (commented):
Thanks for this PR! The approach of using the `/MAP:` linker flag to avoid running a binary during build aligns with how Linux and macOS already work.

The previous `capture_hash` approach had a nice side-effect: by running `fips_empty_main.exe` and observing the integrity check fail (with the placeholder hash), it implicitly proved the integrity check machinery was wired up and functional. With the new static-analysis approach, that implicit verification goes away.

To compensate, I've pushed a commit on top of yours that adds FIPS integrity validation to the Windows CI. It does a couple of things:

1. **Extends `break-hash.go` with PE support** -- this existing tool could only corrupt ELF binaries. It now accepts a `-map` flag and uses the same map file + `debug/pe` approach your PR adds to `inject_hash.go` to locate and corrupt the FIPS module in a Windows DLL.

2. **Added a `fips_build_and_test` subroutine to `run_windows_tests.bat`** -- after the normal build and test pass, it runs a negative test: corrupts `crypto.dll` via `break-hash.go`, runs `test_fips.exe`, and verifies the process aborts. This proves the integrity check actually runs on DLL load and can detect tampering.

torben-hansen:
Haven't looked at the full implementation yet. When inspecting some of the failing CI dimensions, removal of `capture_hash.go` breaks some assumptions about its presence. I tried re-running them just in case, but I think this PR needs a few more changes to compensate/remove those assumptions.

yoavwizstein:
@justsmth yea i've noticed your changes, thanks for helping here!
once current round of ci finishes i'll take a look too

justsmth (approved):

torben-hansen (approved):
nice!

torben-hansen:
Bypassed the omnibus jobs that couldn't pass through the authz check (proxied through the fake deployment). Those doesn't touch Windows machinery though.

yoavwizstein:
Thanks for helping with that one @justsmth 
Do you guys think we could cherry-pick this one into `fips-2025-09-12-lts`?

PR #3134, opened 2026-03-30
Add OPENSSL_INIT_ATFORK compatibility stub

geedo0:
### Description of changes: 
OPENSSL_INIT_ATFORK is an OpenSSL 1.1.x flag that registers
pthread_atfork() handlers for PRNG fork safety. AWS-LC handles fork
detection automatically via kernel-level mechanisms, so this flag is a
safe no-op. Define it as 0 alongside the other stubbed OPENSSL_INIT_*
flags to unblock consumers that reference it (e.g. Valkey).

### Testing:

CI

### DIFF
--- a/include/openssl/crypto.h
+++ b/include/openssl/crypto.h
@@ -179,6 +179,7 @@ OPENSSL_EXPORT void OPENSSL_load_builtin_modules(void);
 #define OPENSSL_INIT_LOAD_CONFIG 0
 #define OPENSSL_INIT_NO_LOAD_CONFIG 0
 #define OPENSSL_INIT_ENGINE_ALL_BUILTIN 0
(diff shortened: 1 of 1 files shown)


WillChilds-Klein (approved):

justsmth (approved):

justsmth:
> > AWS-LC handles fork detection automatically via kernel-level mechanisms, so this flag is a safe no-op.
> 
> just to confirm -- is this true for non-Linux platforms?

Yes, the PRNG is fork-safe on all platforms. The BSDs use `minherit(MAP_INHERIT_ZERO)`. On macOS, it doesn't detect forks but falls back to re-seeding every call. Windows doesn't "fork", so entropy is always generated for a new process.

Issue #3077, opened 2026-03-06
MLDSA Context String


DarkaMaul:
### Problem:

The `EVP_DigestSign`/`EVP_DigestVerify` code path for ML-DSA hardcodes the FIPS 204 context string to `(NULL, 0)`, making it impossible to sign or verify with a non-empty context string through the standard EVP interface.

https://github.com/aws/aws-lc/blob/092453ff03f2c87cac9494cf87e0113fae59af49/crypto/fipsmodule/evp/p_pqdsa.c#L109-L115

And

https://github.com/aws/aws-lc/blob/092453ff03f2c87cac9494cf87e0113fae59af49/crypto/fipsmodule/evp/p_pqdsa.c#L184-L190

The underlying `ml_dsa_{44,65,87}_sign` and `ml_dsa_{44,65,87}_verify` functions already accept `ctx_string` and `ctx_string_len` parameters, and the `PQDSA_METHOD` function pointer table declares them.

In my efforts into adding support for MLDSA using AWS-LC in `pyca/cryptography`, I've resorted to use the lower level API functions ([here](https://github.com/pyca/cryptography/pull/14404/changes/BASE..ef4345fd618535dc00f4062cda7bcc03eb2a22af#diff-383c86b728adb2c75cd13508c56c15d08b55cdabf0a92e20b8f66ae788da0431R18-R20) - ⚠️ code not merged, in PR only).

### Solution:

Add a control to `EVP_PKEY_CTX` that lets callers set the context string before signing or verifying. 

1. Extend [`PQDSA_PKEY_CTX`](https://github.com/aws/aws-lc/blob/092453ff03f2c87cac9494cf87e0113fae59af49/crypto/fipsmodule/evp/p_pqdsa.c#L16-L18)to store a context string:
   ```c
   typedef struct {
     const PQDSA *pqdsa;
     uint8_t ctx_string[MLDSA_MAX_CONTEXT_STRING_LEN]; // 255
     size_t ctx_string_len;
   } PQDSA_PKEY_CTX;
   ```

2. Add an helper function to set the context
3. Update `pkey_pqdsa_sign_message` and `pkey_pqdsa_verify_message` to pass `dctx->ctx_string` and `dctx->ctx_string_len` instead of `NULL, 0`.

* **Does this change any public APIs?** Yes - adds one new exported function (to set the context) and one new constant (for the context max size). No existing APIs change.
[...]

PR #3135, opened 2026-03-31
Support non-empty context strings in ML-DSA EVP sign/verify

jakemas:
### Issues:
Addresses:
- https://github.com/aws/aws-lc-rs/issues/1079
- https://github.com/aws/aws-lc/issues/3077

### Description of changes: 
Plumb FIPS 204 context strings through the `EVP_DigestSign/EVP_DigestVerify` code path for ML-DSA. The lower-level `ml_dsa_*_sign/verify` functions already accept `ctx_string` parameters but they were hardcoded to `(NULL, 0)` in the EVP layer.

### Call-outs:
- Extend `PQDSA_PKEY_CTX` with a `context[255]` buffer and `context_len`
- Add `pkey_pqdsa_ctrl` handling `EVP_PKEY_CTRL_SIGNING_CONTEXT` and `EVP_PKEY_CTRL_GET_SIGNING_CONTEXT `(reusing and renaming the existing generic `EVP_PKEY_CTX_set_signature_context` API from Ed25519ph to `EVP_PKEY_CTX_set1_signature_context_string` to be consistent with bssl)
- Add `pkey_pqdsa_copy` to support `EVP_PKEY_CTX_dup` with context state
- Pass `dctx->context/context_len` to `pqdsa_sign_message` and `pqdsa_verify_message` instead of `NULL, 0`
- Update Wycheproof test helpers to use `EVP_PKEY_CTX_set_signature_context` via EVP_DigestSign/Verify instead of manually computing ExternalMu
- Add `ContextString` unit test covering: round-trip sign+verify with context, mismatched context failure, empty-context backward compatibility, >255 byte rejection, and max-length (255 byte) acceptance

All three ML-DSA variants (44, 65, 87) are covered. Default behavior (empty context string) is unchanged.


crypto/fipsmodule/evp/p_pqdsa.c
```diff
@@ -33,6 +35,58 @@ static void pkey_pqdsa_cleanup(EVP_PKEY_CTX *ctx) {
...
+  if (!pkey_pqdsa_init(dst)) {
+    return 0;
+  }
+
+  PQDSA_PKEY_CTX *dctx = dst->data;
+  PQDSA_PKEY_CTX *sctx = src->data;
+  GUARD_PTR(dctx);
+  GUARD_PTR(sctx);
+
+  dctx->pqdsa = sctx->pqdsa;
```
torben-hansen, line 50: This does a shallow copy. It works in this case because the object has static-storage (returned by `PQDSA_find_dsa_by_nid()`). But in general this pattern doesn't work.

Can leave as-is, but it's worth making a code comment about it - which would also help guide agents and silence bad tooling.

jakemas: Added in [384cc82](https://github.com/aws/aws-lc/pull/3135/commits/384cc82cccb4c180d649a3c764c784be8fb6565a)

crypto/fipsmodule/evp/p_pqdsa.c
```diff
@@ -33,6 +35,60 @@ static void pkey_pqdsa_cleanup(EVP_PKEY_CTX *ctx) {
...
+  return 1;
+}
+
+static int pkey_pqdsa_ctrl(EVP_PKEY_CTX *ctx, int type, int p1, void *p2) {
+  GUARD_PTR(ctx);
+  PQDSA_PKEY_CTX *dctx = (PQDSA_PKEY_CTX *)ctx->data;
+  switch (type) {
+    case EVP_PKEY_CTRL_SIGNING_CONTEXT: {
+      EVP_PKEY_CTX_SIGNATURE_CONTEXT_PARAMS *params = p2;
+      if (!params || !dctx || params->context_len > sizeof(dctx->context)) {
```
justsmth: Also require `params->context != NULL`?
```c
if (!params || !dctx || !params->context || params->context_len > sizeof(dctx->context)) {
```

crypto/evp_extra/p_pqdsa_test.cc
```diff
@@ -2061,6 +2061,92 @@ TEST_P(PQDSAParameterTest, SIGOperations) {
   md_ctx_verify.Reset();
 }
 
+TEST_P(PQDSAParameterTest, ContextString) {
```
justsmth, line 2064: This test should also exercise `EVP_PKEY_CTX_get0_signature_context` to ensure the context can be read back correctly.

include/openssl/evp.h
```diff
@@ -833,8 +833,10 @@ OPENSSL_EXPORT int EVP_PKEY_CTX_get_signature_md(EVP_PKEY_CTX *ctx,
...
 //
-// EVP_PKEY_ED25519PH is the only key type that currently supports setting a
-// a signature context that is used in computing the HashEdDSA signature.
+// EVP_PKEY_ED25519PH and EVP_PKEY_PQDSA are the key types that currently
+// support setting a signature context. For Ed25519ph, the context is used in
+// computing the HashEdDSA signature. For ML-DSA (PQDSA), the context string is
+// used per FIPS 204 sections 5.2-5.3. The maximum context length is 255 bytes.
 //
 // It returns one on success or zero on error.
 OPENSSL_EXPORT int EVP_PKEY_CTX_set_signature_context(EVP_PKEY_CTX *ctx,
```
justsmth, line 854: We should also mention that the context string is ignored when using `EVP_PKEY_sign`/`EVP_PKEY_verify`.

alex:
FWIW in boringssl this is named `EVP_PKEY_CTX_set1_signature_context_string`. Sharing that name would be (moderately) useful to us :-)

crypto/fipsmodule/evp/p_pqdsa.c
jakemas: Added !params->context to the guard in pkey_pqdsa_ctrl. The set path now rejects null context pointers.

crypto/evp_extra/p_pqdsa_test.cc
jakemas: Added two round-trip checks: (1) after setting a context, read it back and verify the content matches and the pointer is an internal copy, (2) verify a freshly initialized context returns NULL with length 0.

include/openssl/evp.h
jakemas: Added a note to the evp.h doc comment clarifying the context string is only used with EVP_DigestSign/EVP_DigestVerify and is ignored for EVP_PKEY_sign/EVP_PKEY_verify.

jakemas:
> FWIW in boringssl this is named `EVP_PKEY_CTX_set1_signature_context_string`. Sharing that name would be (moderately) useful to us :-)

Thanks Alex! I've renamed it. Left `EVP_PKEY_CTX_get0_signature_context` as-is since BoringSSL doesn't have a getter equivalent.

include/openssl/evp.h
```diff
@@ -828,25 +828,30 @@ OPENSSL_EXPORT int EVP_PKEY_CTX_get_signature_md(EVP_PKEY_CTX *ctx,
...
+// support setting a signature context. For Ed25519ph, the context is used in
+// computing the HashEdDSA signature. For ML-DSA (PQDSA), the context string is
+// used per FIPS 204 sections 5.2-5.3. The maximum context length is 255 bytes.
+// Note: the context string is only used with |EVP_DigestSign|/|EVP_DigestVerify|
+// (message signing). It is ignored when using |EVP_PKEY_sign|/|EVP_PKEY_verify|
+// (digest signing).
 //
 // It returns one on success or zero on error.
-OPENSSL_EXPORT int EVP_PKEY_CTX_set_signature_context(EVP_PKEY_CTX *ctx,
+OPENSSL_EXPORT int EVP_PKEY_CTX_set1_signature_context_string(EVP_PKEY_CTX *ctx,
```
justsmth, line 847: This would break any consumers already using this function for `EVP_PKEY_ED25519PH`. I think we will have to support both names.

crypto/fipsmodule/evp/p_pqdsa.c
```diff
@@ -33,6 +35,60 @@ static void pkey_pqdsa_cleanup(EVP_PKEY_CTX *ctx) {
...
+  return 1;
+}
+
+static int pkey_pqdsa_ctrl(EVP_PKEY_CTX *ctx, int type, int p1, void *p2) {
+  GUARD_PTR(ctx);
+  PQDSA_PKEY_CTX *dctx = (PQDSA_PKEY_CTX *)ctx->data;
+  switch (type) {
+    case EVP_PKEY_CTRL_SIGNING_CONTEXT: {
+      EVP_PKEY_CTX_SIGNATURE_CONTEXT_PARAMS *params = p2;
+      if (!params || !dctx || !params->context || params->context_len > sizeof(dctx->context)) {
```
justsmth: Should we also have [Ed25519ph](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/evp/p_ed25519ph.c#L120) do a check for `params->context == NULL`? At least, we should avoid the UB of calling `OPENSSL_memcpy` with a NULL param.


In addition, we could properly support a "reset semantic" here and apply the same pattern to Ed25519ph for consistency:
```c
case EVP_PKEY_CTRL_SIGNING_CONTEXT: {
  EVP_PKEY_CTX_SIGNATURE_CONTEXT_PARAMS *params = p2;
  if (!params || !dctx ||
      params->context_len > sizeof(dctx->context) ||
      (params->context_len > 0 && !params->context)) {
    OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_BUFFER_SIZE);
    return 0;
  }
  OPENSSL_cleanse(dctx->context, sizeof(dctx->context));
  if (params->context_len > 0) {
    OPENSSL_memcpy(dctx->context, params->context, params->context_len);
  }

  dctx->context_len = params->context_len;
  break;
```

crypto/fipsmodule/evp/p_pqdsa.c
```diff
@@ -33,6 +35,60 @@ static void pkey_pqdsa_cleanup(EVP_PKEY_CTX *ctx) {
...
+        OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_BUFFER_SIZE);
```
justsmth: `EVP_R_INVALID_BUFFER_SIZE` might mislead a consumer about the problem. Perhaps, `EVP_R_INVALID_PARAMETERS` would be more clear?

crypto/evp_extra/p_pqdsa_test.cc
```diff
@@ -2061,6 +2061,110 @@ TEST_P(PQDSAParameterTest, SIGOperations) {
...
+  // ---- 7. Max length context (255 bytes) is accepted ----
+  md_ctx.Reset();
+  pkey_ctx = nullptr;
+  ASSERT_TRUE(EVP_DigestSignInit(md_ctx.get(), &pkey_ctx, nullptr, nullptr,
+                                 pkey.get()));
+  uint8_t max_ctx[255];
+  OPENSSL_memset(max_ctx, 0x42, sizeof(max_ctx));
+  ASSERT_TRUE(EVP_PKEY_CTX_set1_signature_context_string(pkey_ctx, max_ctx,
+                                                  sizeof(max_ctx)));
+}
```
justsmth, line 2233: One more test for `EVP_PKEY_CTX_dup`?
```c
  // ---- 8. EVP_PKEY_CTX_dup preserves context ----
  bssl::UniquePtr<EVP_PKEY_CTX> src_ctx(EVP_PKEY_CTX_new(pkey.get(), nullptr));
  ASSERT_TRUE(src_ctx);
  ASSERT_TRUE(EVP_PKEY_sign_init(src_ctx.get()));
  ASSERT_TRUE(EVP_PKEY_CTX_set1_signature_context_string(
      src_ctx.get(), ctx_bytes, sizeof(ctx_bytes)));
  bssl::UniquePtr<EVP_PKEY_CTX> dup_ctx(EVP_PKEY_CTX_dup(src_ctx.get()));
  ASSERT_TRUE(dup_ctx);
  const uint8_t *dup_ctx_bytes = nullptr;
  size_t dup_ctx_len = 0;
  ASSERT_TRUE(EVP_PKEY_CTX_get0_signature_context(
      dup_ctx.get(), &dup_ctx_bytes, &dup_ctx_len));
  ASSERT_EQ(Bytes(ctx_bytes, sizeof(ctx_bytes)),
            Bytes(dup_ctx_bytes, dup_ctx_len));
```

crypto/evp_extra/p_pqdsa_test.cc
```diff
@@ -2061,6 +2061,110 @@ TEST_P(PQDSAParameterTest, SIGOperations) {
...
+
+  // ---- 7. Max length context (255 bytes) is accepted ----
+  md_ctx.Reset();
+  pkey_ctx = nullptr;
+  ASSERT_TRUE(EVP_DigestSignInit(md_ctx.get(), &pkey_ctx, nullptr, nullptr,
+                                 pkey.get()));
+  uint8_t max_ctx[255];
+  OPENSSL_memset(max_ctx, 0x42, sizeof(max_ctx));
+  ASSERT_TRUE(EVP_PKEY_CTX_set1_signature_context_string(pkey_ctx, max_ctx,
+                                                  sizeof(max_ctx)));
```
justsmth: NP: It would be better to round-trip this so that we know all level properly support a max-length context string.

crypto/fipsmodule/evp/p_pqdsa.c
```diff
@@ -108,7 +164,7 @@ static int pkey_pqdsa_sign_generic(EVP_PKEY_CTX *ctx, uint8_t *sig,
 
   // RAW sign mode
   if (!sign_digest) {
-    if (!pqdsa->method->pqdsa_sign_message(key->private_key, sig, sig_len, message, message_len, NULL, 0)) {
+    if (!pqdsa->method->pqdsa_sign_message(key->private_key, sig, sig_len, message, message_len, dctx->context_len > 0 ? dctx->context : NULL, dctx->context_len)) {
```
justsmth: NP: line wrap?

include/openssl/evp.h
jakemas: Added `EVP_PKEY_CTX_set_signature_context` back as a wrapper that delegates to `EVP_PKEY_CTX_set1_signature_context_string`, so both names work.

crypto/fipsmodule/evp/p_pqdsa.c
jakemas: Applied the suggested pattern to both p_pqdsa.c and p_ed25519ph.c: NULL context allowed when context_len == 0, OPENSSL_cleanse before copy, guarded OPENSSL_memcpy.

crypto/fipsmodule/evp/p_pqdsa.c
jakemas: Changed to `EVP_R_INVALID_PARAMETERS`.

crypto/evp_extra/p_pqdsa_test.cc
jakemas: Added.

crypto/evp_extra/p_pqdsa_test.cc
jakemas: Converted to a full sign + verify round-trip with the 255-byte context.

crypto/fipsmodule/evp/p_pqdsa.c
jakemas: Fixed.

include/openssl/evp.h
```diff
@@ -828,15 +828,27 @@ OPENSSL_EXPORT int EVP_PKEY_CTX_get_signature_md(EVP_PKEY_CTX *ctx,
...
 //
-// EVP_PKEY_ED25519PH is the only key type that currently supports setting a
-// a signature context that is used in computing the HashEdDSA signature.
+// EVP_PKEY_ED25519PH and EVP_PKEY_PQDSA are the key types that currently
+// support setting a signature context. For Ed25519ph, the context is used in
+// computing the HashEdDSA signature. For ML-DSA (PQDSA), the context string is
+// used per FIPS 204 sections 5.2-5.3. The maximum context length is 255 bytes.
+// Note: the context string is only used with |EVP_DigestSign|/|EVP_DigestVerify|
+// (message signing). It is ignored when using |EVP_PKEY_sign|/|EVP_PKEY_verify|
+// (digest signing).
```
justsmth: NP: This is accurate for ML-DSA, but for Ed25519ph the context *is* used in the `EVP_PKEY_sign`/`EVP_PKEY_verify` path. Should the note be narrowed to ML-DSA?
```c
// Note: for ML-DSA, the context string is only used with
// |EVP_DigestSign|/|EVP_DigestVerify| (message signing). It is ignored when
// using |EVP_PKEY_sign|/|EVP_PKEY_verify| (digest signing).
```

crypto/evp_extra/p_pqdsa_test.cc
```diff
@@ -2551,125 +2687,47 @@ INSTANTIATE_TEST_SUITE_P(
...
 
 // SignMLDSAWithContext produces a signature |sig| for message |msg| with
-// context |msg_ctx|. We need this wrapper because |EVP_DigestSign| does not
-// support signing with contexts.
+// context |msg_ctx| using the EVP_DigestSign path with
+// EVP_PKEY_CTX_set1_signature_context_string.
 //
 // It returns one on success and zero on error.
 static int SignMLDSAWithContext(EVP_PKEY *pkey, std::vector<uint8_t> &sig,
                                 const std::vector<uint8_t> &pk,
```
justsmth: NP: I don't think `pk` is used now.

crypto/evp_extra/p_pqdsa_test.cc
```diff
@@ -2551,125 +2687,47 @@ INSTANTIATE_TEST_SUITE_P(
...
-
 // VerifyMLDSAWithContext verifies that |sig| is a valid signature for |msg|
-// with context |msg_ctx|. We need this wrapper because |EVP_DigestVerify| does
-// not support verification with contexts.
+// with context |msg_ctx| using the EVP_DigestVerify path with
+// EVP_PKEY_CTX_set1_signature_context_string.
 //
 // It returns one on success and zero on error.
 static int VerifyMLDSAWithContext(EVP_PKEY *pkey,
                                   const std::vector<uint8_t> &pk,
```
justsmth: NP: I don't think `pk` is used any more.

crypto/fipsmodule/evp/evp.c
```diff
@@ -659,14 +659,20 @@ int EVP_PKEY_CTX_get_signature_md(EVP_PKEY_CTX *ctx, const EVP_MD **out_md) {
...
   EVP_PKEY_CTX_SIGNATURE_CONTEXT_PARAMS params = {context, context_len};
   return EVP_PKEY_CTX_ctrl(ctx, -1, EVP_PKEY_OP_TYPE_SIG,
                            EVP_PKEY_CTRL_SIGNING_CONTEXT, 0, &params);
 }
 
+int EVP_PKEY_CTX_set_signature_context(EVP_PKEY_CTX *ctx,
+                                       const uint8_t *context,
+                                       size_t context_len) {
+  return EVP_PKEY_CTX_set1_signature_context_string(ctx, context, context_len);
+}
```
justsmth, line 688: NP: Could one of the tests call `EVP_PKEY_CTX_set_signature_context` instead of `EVP_PKEY_CTX_set1_signature_context_string` just to ensure coverage for it?

crypto/fipsmodule/evp/p_pqdsa.c
justsmth, line 191: NP: Should we insert another error condition here?
```
    // For ML-DSA, the digest-sign path (EVP_PKEY_sign) takes a pre-hashed
    // |mu| input which already encodes the context string per FIPS 204
    // section 5.3. Applying a separately-configured context here would be
    // silently ignored and produce a signature inconsistent with the
    // caller's intent, so reject the combination explicitly.
    if (dctx->context_len > 0) {
      OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_OPERATION);
      return 0;
    }
```

justsmth, line 282: NP: Similar error condition here:
```
    if (dctx->context_len > 0) {
      OPENSSL_PUT_ERROR(EVP, EVP_R_INVALID_OPERATION);
      return 0;
    }
```

justsmth (dismissed):

include/openssl/evp.h
jakemas: Narrowed the note to ML-DSA in [848ed20](https://github.com/aws/aws-lc/pull/3135/commits/848ed20d2de58124b92e9519e65b1c0a6044c371). The comment now explicitly calls out that the `EVP_PKEY_sign`/`EVP_PKEY_verify` exclusion applies to ML-DSA because the pre-hashed `mu` input already encodes the context per FIPS 204 §5.3.

crypto/evp_extra/p_pqdsa_test.cc
jakemas: Dropped the unused `pk` parameter from `SignMLDSAWithContext` (and updated all callers) in [848ed20](https://github.com/aws/aws-lc/pull/3135/commits/848ed20d2de58124b92e9519e65b1c0a6044c371).

crypto/evp_extra/p_pqdsa_test.cc
jakemas: Dropped the unused `pk` parameter from `VerifyMLDSAWithContext` (and updated all callers) in [848ed20](https://github.com/aws/aws-lc/pull/3135/commits/848ed20d2de58124b92e9519e65b1c0a6044c371).

crypto/fipsmodule/evp/evp.c
jakemas: Switched the verify-side setup in the `ContextString` test to call `EVP_PKEY_CTX_set_signature_context` in [848ed20](https://github.com/aws/aws-lc/pull/3135/commits/848ed20d2de58124b92e9519e65b1c0a6044c371), so both setter names are covered in the round-trip.

crypto/fipsmodule/evp/p_pqdsa.c
jakemas: Added the guard on the digest-sign path in [848ed20](https://github.com/aws/aws-lc/pull/3135/commits/848ed20d2de58124b92e9519e65b1c0a6044c371). A non-empty `dctx->context_len` now returns `EVP_R_INVALID_OPERATION` with a comment explaining that the pre-hashed `mu` input already encodes the context per FIPS 204 §5.3, so a separately-configured context would otherwise be silently ignored.

jakemas: Added the symmetric guard on the digest-verify path in [848ed20](https://github.com/aws/aws-lc/pull/3135/commits/848ed20d2de58124b92e9519e65b1c0a6044c371), mirroring the sign-side check.

include/openssl/evp.h
```diff
@@ -828,15 +828,28 @@ OPENSSL_EXPORT int EVP_PKEY_CTX_get_signature_md(EVP_PKEY_CTX *ctx,
...
 //
-// EVP_PKEY_ED25519PH is the only key type that currently supports setting a
-// a signature context that is used in computing the HashEdDSA signature.
+// EVP_PKEY_ED25519PH and EVP_PKEY_PQDSA are the key types that currently
+// support setting a signature context. For Ed25519ph, the context is used in
+// computing the HashEdDSA signature. For ML-DSA (PQDSA), the context string is
+// used per FIPS 204 sections 5.2-5.3. The maximum context length is 255 bytes.
+// Note: for ML-DSA, the context string is only used with
+// |EVP_DigestSign|/|EVP_DigestVerify| (message signing). It is ignored when
+// using |EVP_PKEY_sign|/|EVP_PKEY_verify| (digest signing), because the pre-
```
justsmth: NP: Suggested tweak: `s/is ignored/is not permitted and will return an error/` (or similar).

crypto/evp_extra/p_pqdsa_test.cc
```diff
@@ -2061,6 +2061,144 @@ TEST_P(PQDSAParameterTest, SIGOperations) {
...
+  bssl::UniquePtr<EVP_PKEY_CTX> dup_ctx(EVP_PKEY_CTX_dup(pkey_ctx));
+  ASSERT_TRUE(dup_ctx);
+
+  read_ctx = nullptr;
+  read_ctx_len = 0;
+  ASSERT_TRUE(EVP_PKEY_CTX_get0_signature_context(dup_ctx.get(), &read_ctx,
+                                                   &read_ctx_len));
+  ASSERT_TRUE(read_ctx);
+  ASSERT_EQ(Bytes(ctx_bytes, sizeof(ctx_bytes)), Bytes(read_ctx, read_ctx_len));
+}
```
justsmth, line 2233: Also coverage for a couple more negative cases:
```c++
  // ---- 9. Context is rejected on the EVP_PKEY_sign/verify digest path ----
  // For ML-DSA, the pre-hashed |mu| input already encodes the context per
  // FIPS 204 section 5.3, so combining a configured context with the digest
  // path must fail rather than be silently ignored.
  bssl::UniquePtr<EVP_PKEY_CTX> raw_sign_ctx(
      EVP_PKEY_CTX_new(pkey.get(), nullptr));
  ASSERT_TRUE(raw_sign_ctx);
  ASSERT_TRUE(EVP_PKEY_sign_init(raw_sign_ctx.get()));
  ASSERT_TRUE(EVP_PKEY_CTX_set1_signature_context_string(
      raw_sign_ctx.get(), ctx_bytes, sizeof(ctx_bytes)));

  // |mu| is the 64-byte SHAKE256 output expected by the digest path; the
  // context guard fires before any cryptographic check, so the contents
  // don't need to be a valid mu for this test.
  uint8_t mu[64] = {0};
  size_t raw_sig_len = 0;
  // Query the signature size (|sig == NULL| short-circuits before the guard).
  ASSERT_TRUE(EVP_PKEY_sign(raw_sign_ctx.get(), nullptr, &raw_sig_len, mu,
                            sizeof(mu)));
  std::vector<uint8_t> raw_sig(raw_sig_len);
  ASSERT_FALSE(EVP_PKEY_sign(raw_sign_ctx.get(), raw_sig.data(), &raw_sig_len,
                             mu, sizeof(mu)));

  bssl::UniquePtr<EVP_PKEY_CTX> raw_verify_ctx(
      EVP_PKEY_CTX_new(pkey.get(), nullptr));
  ASSERT_TRUE(raw_verify_ctx);
  ASSERT_TRUE(EVP_PKEY_verify_init(raw_verify_ctx.get()));
  ASSERT_TRUE(EVP_PKEY_CTX_set1_signature_context_string(
      raw_verify_ctx.get(), ctx_bytes, sizeof(ctx_bytes)));
  std::vector<uint8_t> dummy_sig(raw_sig_len, 0);
  ASSERT_FALSE(EVP_PKEY_verify(raw_verify_ctx.get(), dummy_sig.data(),
                               dummy_sig.size(), mu, sizeof(mu)));
```

justsmth (commented):
Looks good. Just a couple of minor concerns...

include/openssl/evp.h
jakemas: Applied in [df71998](https://github.com/aws/aws-lc/pull/3135/commits/df7199885). The note on |EVP_PKEY_CTX_set1_signature_context_string| now reads "is not permitted and will return an error" to match the actual digest-path guard behavior.

crypto/evp_extra/p_pqdsa_test.cc
jakemas: Added the suggested negative coverage in [df71998](https://github.com/aws/aws-lc/pull/3135/commits/df7199885). The `ContextString` test now exercises the digest-path guard on both `EVP_PKEY_sign` and `EVP_PKEY_verify` (section 9 of the test), so a configured context combined with the pre-hashed `mu` input fails loudly rather than being silently dropped.

justsmth (approved):

torben-hansen (approved):

PR #3136, opened 2026-04-02
Make randomness generation deterministic under BORINGSSL_UNSAFE_FUZZER_MODE

torben-hansen:
### Description of changes: 

When `BORINGSSL_UNSAFE_DETERMINISTIC_MODE` is defined (e.g. via `-DFUZZ=1`), the entropy source layer now routes seed and extra entropy requests  through `CRYPTO_sysrand`, which is already deterministic in this mode (ChaCha20 with counter-based nonce). This bypasses jitter and hardware entropy sources, ensuring fully deterministic randomness generation end-to-end to support fuzzing.

### Testing:

Added small assertion test.


WillChilds-Klein (approved):

crypto/fipsmodule/rand/entropy/entropy_sources.c
```diff
@@ -122,11 +122,54 @@ DEFINE_LOCAL_DATA(struct entropy_source_methods, opt_out_cpu_jitter_entropy_sour
...
+//   |OPENSSL_RAND_DETERMINISTIC| mode.
+DEFINE_LOCAL_DATA(struct entropy_source_methods, deterministic_entropy_source_methods) {
+  out->initialize = entropy_deterministic_initialize;
+  out->zeroize_thread = entropy_deterministic_zeroize_thread;
+  out->free_thread = entropy_deterministic_free_thread;
+  out->get_seed = entropy_deterministic_get;
+  out->get_extra_entropy = entropy_deterministic_get;
+  out->get_prediction_resistance = NULL;
+  out->id = DETERMINISTIC_ENTROPY_SOURCE;
+}
```
justsmth, line 162: For further safety, we should avoid even compiling a deterministic option unless the guard enables it:
```
#if defined(OPENSSL_RAND_DETERMINISTIC)
```

crypto/fipsmodule/rand/entropy/entropy_sources.c
justsmth, line 178: If the guard is set, why even bother with the other options.
```c
static const struct entropy_source_methods * get_entropy_source_methods(void) {
#if defined(OPENSSL_RAND_DETERMINISTIC)
    return deterministic_entropy_source_methods();
#else
  if (*allow_entropy_source_methods_override_bss_get() == 1) {
    return *entropy_source_methods_override_bss_get();
  }

  if (use_opt_out_cpu_jitter_entropy()) {
    return opt_out_cpu_jitter_entropy_source_methods();
  }

  return tree_jitter_entropy_source_methods();
#endif
}
```

crypto/fipsmodule/rand/entropy/entropy_sources.c
torben-hansen: The idea of organising it like this, is to reduce the need for synchronised if/defs. So, against implementing more.

crypto/fipsmodule/rand/entropy/entropy_sources.c
torben-hansen: See comment above.

torben-hansen:
Implementation doesn't quite yet meet requirements I had. Right now, one cannot effectively "reset" generators in a multi-threaded environment. For that, I would need some thread-local controls.

PR #3138, opened 2026-04-06
Check RSA-PSS digest algorithms for X509

skmcgrail:
### Description of changes: 
MD4 and MD5 digest algorithms are not allowed for signature algorithms for signing X.509 certificates with RSA keys. This check however is not enforced for RSA-PSS signature algorithm. This pull request modifies this behavior to also enforce the constraint for the RSA-PSS signature algorithm when verifying X.509 certificates.

### DIFF
--- a/crypto/x509/rsa_pss.c
+++ b/crypto/x509/rsa_pss.c
@@ -230,6 +230,12 @@ int x509_rsa_pss_to_ctx(EVP_MD_CTX *ctx, const X509_ALGOR *sigalg,
     goto err;
   }
 
+  // Check that both signature algorithms are not weak
+  if(!x509_digest_nid_ok(EVP_MD_type(md)) || !x509_digest_nid_ok(EVP_MD_type(mgf1md))) {
+    OPENSSL_PUT_ERROR(X509, ASN1_R_DIGEST_AND_KEY_TYPE_NOT_SUPPORTED);
+    goto err;
+  }
+
   int saltlen = 20;
   if (pss->saltLength != NULL) {
     saltlen = ASN1_INTEGER_get(pss->saltLength);
--- a/crypto/x509/x509_test.cc
+++ b/crypto/x509/x509_test.cc
@@ -333,6 +333,56 @@ lC9+9hPHIoc9UMmAQNo1vGIW3NWVoeGbaJ8=
 -----END CERTIFICATE-----
 )";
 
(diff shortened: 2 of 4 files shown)


crypto/x509/rsa_pss.c
justsmth, line 161: There's an analogous check [here in `x509_digest_sign_algorithm`](https://github.com/aws/aws-lc/blob/main/crypto/x509/algorithm.c#L60-L66) that's bypassed for PSS.

Should we add a check here too?
```c
  // Check that the digest algorithms are not weak
  if (!x509_digest_nid_ok(EVP_MD_type(sigmd)) ||
      !x509_digest_nid_ok(EVP_MD_type(mgf1md))) {
    OPENSSL_PUT_ERROR(X509, ASN1_R_DIGEST_AND_KEY_TYPE_NOT_SUPPORTED);
    return 0;
  }
```

justsmth: Ok, the scenario that would trigger this is narrow -- a caller would need to be using a regular `EVP_PKEY_RSA` key, explicitly configuring PSS padding mode on the context, then sign the X.509 certificate via `ASN1_item_sign_ctx`.

justsmth (approved):

crypto/x509/x509_test.cc
```diff
@@ -333,6 +333,56 @@ lC9+9hPHIoc9UMmAQNo1vGIW3NWVoeGbaJ8=
 -----END CERTIFICATE-----
 )";
 
+// kPSSWithMD5CertPEM is a self-signed RSA-PSS certificate where both the
+// signature hash and MGF1 hash are MD5. Verification should be rejected because
+// MD5 is not an allowed digest for X509.
+static const char kPSSWithMD5CertPEM[] = R"(
```
torben-hansen, line 339: By the order of validation (RSA-PSS blob parsed before validating cert attributes), below time-bomb is not currently reachable.

One technicality is that `X509_verify()` is called on these certs which could end up validating the validity period: that is already expired
```
validity Validity SEQUENCE (2 elem)
  notBefore utcTime Time UTCTime 2024-01-01 00:00:00 UTC
  notAfter utcTime Time UTCTime 2025-12-31 00:00:00 UTC
```
Causing all builds to fail.

torben-hansen (approved):

PR #3139, opened 2026-04-06
Prevent possible use after free in add_string and possible memory leak

skmcgrail:
### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 
In the rare case `lh_CONF_VALUE_insert` fails its possible to both leak memory by not freeing the duplicated section string name on `CONF_VALUE`, and then cause a potential for a use after free by retaining the `CONF_VALUE` in the section stack of values.

This failure scenario is only possible if memory allocation fails during `OPENSSL_lh_insert`.

### DIFF
--- a/crypto/conf/conf.c
+++ b/crypto/conf/conf.c
@@ -331,6 +331,9 @@ static int add_string(const CONF *conf, CONF_VALUE *section,
   }
 
   if (!lh_CONF_VALUE_insert(conf->data, &old_value, value)) {
+    (void)sk_CONF_VALUE_delete_ptr(section_stack, value);
+    OPENSSL_free(value->section);
+    value->section = NULL;
     return 0;
   }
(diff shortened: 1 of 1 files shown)


WillChilds-Klein (approved):

crypto/conf/conf.c
```diff
@@ -331,6 +331,9 @@ static int add_string(const CONF *conf, CONF_VALUE *section,
   }
```
justsmth, line 331: Also here?
```c
  value->section = OPENSSL_strdup(section->section);
  if (value->section == NULL || !sk_CONF_VALUE_push(section_stack, value)) {
    OPENSSL_free(value->section);
    value->section = NULL;
    return 0;
  }
```

justsmth:
Following up in:
* #3187

PR #3142, opened 2026-04-07
Add CI for Zig compiler support

justsmth:
### Description of changes:

Adds CI to build and test AWS-LC using the [Zig](https://ziglang.org/) compiler toolchain (`zig cc` / `zig c++`). The Zig compiler is Clang-based and is increasingly popular for cross-compilation and as an alternative C/C++ toolchain. This PR validates that AWS-LC builds and passes tests when compiled with Zig across five platform configurations:

| Runner | Arch | OS |
|---|---|---|
| `ubuntu-latest` | x86_64 | Linux |
| `ubuntu-24.04-arm` | aarch64 | Linux |
| `windows-latest` | x86_64 | Windows |
| `windows-11-arm` | aarch64 | Windows |
| `macos-latest` | aarch64 | Darwin |

Zig doesn't accept a few GCC/Clang flags that our build emits, so this PR includes thin wrapper scripts (`util/zig-cc`, `util/zig-c++`) that translate them before forwarding to `zig cc`/`zig c++`. The Windows variants (`.cmd`) are shims that call the bash scripts through Git Bash.

### Call-outs:

There are three platform-specific workarounds, all documented inline:

1. **macOS linking**: Zig's self-hosted MachO linker segfaults ([ziglang/zig#25451](https://github.com/ziglang/zig/issues/25451), [ziglang/zig#23704](https://github.com/ziglang/zig/issues/23704)), so the toolchain falls back to the system compiler (`/usr/bin/cc`) for linking only. Zig still handles all compilation.

2. **Windows `SSL_peek` tests**: The `Peek-*-Server` SSL runner tests hang when `bssl_shim` is compiled with Zig. These are disabled via a shim config file (`zig_windows_config.json`). A small CMake change plumbs the `-DSSL_RUNNER_SHIM_CONFIG` option through to the runner's `-shim-config` flag.

3. **Windows SEH unwind tests**: Zig's LLD linker produces `.pdata`/`.xdata` (SEH unwind metadata) that is incompatible with the `RtlVirtualUnwind`-based unwind tester, causing `STATUS_ACCESS_VIOLATION` in `ABITest.SanityCheck`. The non-unwind ABI register-preservation tests pass fine — only the single-step unwind verification crashes. A `-no-unwind-tests` flag was added to `all_tests.go` (following the existing SDE pattern) and is used for the Windows Zig targets.

### Testing:

CI across the five matrix configurations listed above. All tests pass.

### DIFF
--- a/.github/workflows/zig.yml
+++ b/.github/workflows/zig.yml
@@ -67,11 +67,31 @@ jobs:
           set(CMAKE_VERBOSE_MAKEFILE ON)
           set(CMAKE_MESSAGE_LOG_LEVEL DEBUG)
           EOF
+          # Workaround: zig's self-hosted MachO linker segfaults on macOS.
+          # Override CMake's link rules to use the system compiler for linking.
+          # Zig is still used for all compilation steps.
+          # Tracked upstream:
+          #   https://github.com/ziglang/zig/issues/25451
+          #   https://github.com/ziglang/zig/issues/23704
+          if [[ "${{ matrix.config.target_system }}" == "Darwin" ]]; then
+            cat <<EOF >> ./toolchain.cmake
+          set(CMAKE_C_LINK_EXECUTABLE "/usr/bin/cc <CMAKE_C_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
+          set(CMAKE_CXX_LINK_EXECUTABLE "/usr/bin/c++ <CMAKE_CXX_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
+          EOF
+          fi
       - name: Setup CMake
         shell: bash
         run: |
           cat ./toolchain.cmake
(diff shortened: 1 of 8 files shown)


.github/workflows/zig.yml
```diff
@@ -0,0 +1,114 @@
...
+            echo "ZIGCXX=$(cygpath -m "${PWD}/util/zig-c++.cmd")" >> $GITHUB_ENV
+          else
+            echo "ZIGCC=${PWD}/util/zig-cc" >> $GITHUB_ENV
+            echo "ZIGCXX=${PWD}/util/zig-c++" >> $GITHUB_ENV
+          fi
+      - name: Create toolchain
+        shell: bash
+        run: |
+          cat <<EOF > ./toolchain.cmake
+          set(CMAKE_C_COMPILER ${ZIGCC})
```
dougch, line 62: Curious why these didn't go into `CMakeLists.txt` or a `toolchain.${target_system}.cmake` with a flag to enable them?

dougch (approved):

.github/workflows/zig.yml
justsmth: The toolchain file needs absolute paths that are specific to the environment, so unfortunately it has to be generated dynamically.

nhatnghiho (approved):

PR #3146, opened 2026-04-08
Improve test portability for OPENSSL_NO_SOCK, OPENSSL_THREADS, and OPENSSL_NO_TTY

justsmth:
### Description of changes:

Several test files don't compile on platforms where `OPENSSL_NO_SOCK` or `OPENSSL_NO_THREADS_CORRUPT_MEMORY_AND_LEAK_SECRETS_IF_THREADED` is defined. These flags are already set for nanolibc, Android baremetal, CROS_EC, and CROS_ZEPHYR, so the missing guards are latent bugs — they just haven't been hit because we don't build tests for those targets.

This PR adds the missing compile-time guards and introduces a new `OPENSSL_NO_TTY` flag for platforms without terminal I/O:

- Guard `BIO_new_ssl_connect` (declaration and implementation) behind `OPENSSL_NO_SOCK`, matching the existing pattern for `SSL_set_fd`, `SSL_set_rfd`, and `SSL_set_wfd` in the same header.
- Add `OPENSSL_NO_TTY` stub implementations in `crypto/console/console.c` for headless platforms.
- Guard socket-dependent tests (`bio_socket_test`, `ssl_test`, etc.) with `OPENSSL_NO_SOCK`.
- Guard thread-dependent tests (`x509_test`, `ssl_handshake_test`, `ssl_version_test`) with `OPENSSL_THREADS`.

All changes are no-ops for currently-tested platforms.

### Call-outs:
- `BIO_new_ssl_connect` is now conditionally compiled out when `OPENSSL_NO_SOCK` is defined. This is a public API visibility change but is consistent with how `SSL_set_fd` et al. are already handled in the same header.
- The `GTEST_HAS_DEATH_TEST` checks are changed from `#if defined(GTEST_HAS_DEATH_TEST)` to `#if GTEST_HAS_DEATH_TEST`. This is a bug fix: when a toolchain defines it to `0`, `defined()` still evaluates to true.

### Testing:
All changes are guarded by preprocessor flags that are not active on any currently-tested platform. Existing CI should confirm no regressions.


crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
```diff
@@ -14,7 +14,7 @@
 
 #include <cstdio>
 
-#if defined(GTEST_HAS_DEATH_TEST)
+#if GTEST_HAS_DEATH_TEST
```
samuel40791765, line 18: I'm assuming this adjustment was needed, would it be worth documenting why?

ssl/ssl_test.cc
```diff
@@ -443,7 +444,9 @@ static void ExpectFDs(const SSL *ssl, int rfd, int wfd) {
     EXPECT_EQ(SSL_get_rbio(ssl), SSL_get_wbio(ssl));
   }
 }
+#endif  // !OPENSSL_NO_SOCK
 
+#if !defined(OPENSSL_NO_SOCK)
```
samuel40791765: Nit: could probably consolidate these

crypto/console/console.c
```diff
@@ -1,6 +1,43 @@
...
+  return 0;
+}
+
+int openssl_console_read(char *buf, int minsize, int maxsize, int echo) {
+  (void)buf;
+  (void)minsize;
+  (void)maxsize;
+  (void)echo;
+  return -1;  // Error: no console available
+}
```
samuel40791765, line 44: Just confirming, these stub values are identical to OpenSSLs? It seems like some functions using like `EVP_read_pw_string_min` would start failing too, would it be worth documenting?

crypto/fipsmodule/rand/entropy/tree_drbg_jitter_entropy_isolated_test.cc
justsmth: Good call — added a comment above both sites: 
```c
// Use #if GTEST_HAS_DEATH_TEST (not defined()); some toolchains define it to 0.
```

ssl/ssl_test.cc
justsmth: Done, merged them into a single guard block.

crypto/console/console.c
justsmth: Not identical -- OpenSSL uses `OPENSSL_NO_UI_CONSOLE` to gate out the entire `UI_OpenSSL()` method at a higher level. Our approach stubs at the lower `openssl_console_*` layer so the rest of the code compiles unconditionally and fails gracefully at runtime. I expanded the comment block to explain this distinction and call out that `EVP_read_pw_string`/`EVP_read_pw_string_min` will fail on these platforms.

Also noticed that `OPENSSL_NO_TTY` wasn't actually being defined for any of the embedded platforms — added it to the nanolibc, Android baremetal, CROS_EC, and CROS_ZEPHYR blocks in `target.h`.

samuel40791765 (approved):

dougch (approved):

Issue #3140, opened 2026-04-07
ML-KEM keys seed storage


DarkaMaul:
### Problem:

AWS-LC can **decode** ML-KEM private keys from the compact 64-byte seed format but cannot **encode** to that format. After key generation the seed is discarded and `kem_priv_encode` always emits the expanded secret key.

In the encoding part, a comment acknowledges this:

https://github.com/aws/aws-lc/blob/863ed59b19feac3f0da174b9ef69301962080bb3/crypto/evp_extra/p_kem_asn1.c#L238

Compare the two structs (MLDSA and MLKEM):

https://github.com/aws/aws-lc/blob/863ed59b19feac3f0da174b9ef69301962080bb3/crypto/fipsmodule/pqdsa/internal.h#L70-L76

https://github.com/aws/aws-lc/blob/863ed59b19feac3f0da174b9ef69301962080bb3/crypto/fipsmodule/kem/internal.h#L62-L66

### Solution:

1. **Add `uint8_t *seed` to `kem_key_st`**, mirroring `pqdsa_key_st`:

   ```c
   struct kem_key_st {
       const KEM *kem;
       uint8_t *public_key;
       uint8_t *secret_key;
       uint8_t *seed;
   };
2. **Populate the seed during keygen**, the same way ML-DSA does in [`pkey_pqdsa_keygen`](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/evp/p_pqdsa.c#L52).
3. **Update `kem_priv_encode` to serialize the seed**, mirroring [`pqdsa_priv_encode`](https://github.com/aws/aws-lc/blob/main/crypto/evp_extra/p_pqdsa_asn1.c#L211-L224) which emits `seed [0] OCTET STRING` when the seed is available.

* **Does this change any public APIs?** No
* **Which algorithm(s) will this impact?** ML-KEM-512, ML-KEM-768, ML-KEM-1024.


### Requirements / Acceptance Criteria:

What must a solution address in order to solve the problem? How do we know the solution is complete?

* **RFC links:** [FIPS 203](https://doi.org/10.6028/NIST.FIPS.203)
* **Related Issues:** None
* **Will the Usage Guide or other documentation need to be updated?** I'm unsure
* **Testing:** Unit tests in the KEM test suite covering the cases listed above for all three ML-KEM variants.
  * **Will this change trigger AWS LibCrypto Formal Verification changes?** No
  * **Should this change be fuzz tested?** No 

### Out of scope:

[...]

PR #3149, opened 2026-04-08
Change ML-KEM PKCS#8 encoding from expanded to seed form

geedo0:
### Issues:
Addresses P409194150, #3140 

### Description of changes: 
Add seed field to KEM_KEY and make seed format the default PKCS#8
encoding for all ML-KEM private keys, mirroring the existing ML-DSA
pattern. This implements RFC 9935 seed format encoding.

Changes:
- Add uint8_t *seed to kem_key_st, preserved during all keygen paths
  and seed-format PKCS#8 import
- Rewrite pkey_kem_keygen to generate seed via RAND_bytes and call
  keygen_deterministic, so all generated keys retain their seed
- Save seed in pkey_kem_keygen_deterministic and
  KEM_KEY_set_raw_keypair_from_seed (decode path)
- Update kem_priv_encode to emit seed format ([0] IMPLICIT OCTET
  STRING) when seed is available, expanded format otherwise

Keys imported via raw expanded bytes (EVP_PKEY_kem_new_raw_secret_key)
continue to encode as expanded format since no seed is available.
EVP_PKEY_get_raw_private_key continues to return the expanded secret
key regardless of encoding format.

### Call-outs:
This changes the default output format of from the expanded key format to the seed format. There is a remote possibility that this is breaking for some customer that relies on the expanded format being output by AWS-LC and passing that key into some software stack that does not support seed format. Of all the major implementers (AWS-LC, BoringSSL, OpenSSL, BouncyCastle...) of RFC 9935, the consensus is to support importing from either format so we don't expect there to be any large scale regressions.

### Testing:

CI

### DIFF
--- a/crypto/evp_extra/p_kem_asn1.c
+++ b/crypto/evp_extra/p_kem_asn1.c
@@ -104,7 +104,7 @@ static int kem_cmp_parameters(const EVP_PKEY *a, const EVP_PKEY *b) {
 
 
 static int kem_pub_decode(EVP_PKEY *out, CBS *oid, CBS *params, CBS *key) {
-  // See https://datatracker.ietf.org/doc/draft-ietf-lamps-kyber-certificates/
+  // See https://datatracker.ietf.org/doc/rfc9935/
   // section 4. There should be no parameters
   if (CBS_len(params) > 0) {
     OPENSSL_PUT_ERROR(EVP, EVP_R_DECODE_ERROR);
@@ -132,7 +132,7 @@ static int kem_pub_encode(CBB *out, const EVP_PKEY *pkey) {
     return 0;
   }
 
-  // See https://datatracker.ietf.org/doc/draft-ietf-lamps-kyber-certificates/
+  // See https://datatracker.ietf.org/doc/rfc9935/
   // section 4.
   CBB spki, algorithm, oid, key_bitstring;
   if (!CBB_add_asn1(out, &spki, CBS_ASN1_SEQUENCE) || // SubjectPublicKeyInfo SEQUENCE
@@ -165,7 +165,7 @@ static int kem_pub_cmp(const EVP_PKEY *a, const EVP_PKEY *b) {
 
 static int kem_priv_decode(EVP_PKEY *out, CBS *oid, CBS *params, CBS *key,
                            CBS *pubkey) {
-  // See https://datatracker.ietf.org/doc/draft-ietf-lamps-kyber-certificates/
+  // See https://datatracker.ietf.org/doc/rfc9935/
   // section 6. There should be no parameters.
   if (CBS_len(params) > 0) {
     OPENSSL_PUT_ERROR(EVP, EVP_R_DECODE_ERROR);
@@ -179,7 +179,7 @@ static int kem_priv_decode(EVP_PKEY *out, CBS *oid, CBS *params, CBS *key,
   }
 
   // Support multiple ML-KEM private key formats from
-  // https://datatracker.ietf.org/doc/draft-ietf-lamps-kyber-certificates/
+  // https://datatracker.ietf.org/doc/rfc9935/
   // Case 1: seed [0] OCTET STRING
   // Case 2: expandedKey OCTET STRING
   // Case 3: TODO: both SEQUENCE {seed, expandedKey}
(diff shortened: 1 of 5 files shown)


crypto/fipsmodule/evp/p_kem.c
```diff
@@ -97,9 +104,24 @@ static int pkey_kem_keygen(EVP_PKEY_CTX *ctx, EVP_PKEY *pkey) {
   KEM_KEY *key = KEM_KEY_new();
   size_t pubkey_len = kem->public_key_len;
   size_t secret_len = kem->secret_key_len;
+
+  // Generate random seed, use deterministic keygen, and save the seed.
+  // All ML-KEM variants use 64-byte seeds (d || z).
+  uint8_t seed[64];
+  RAND_bytes(seed, kem->keygen_seed_len);
```
justsmth, line 112: * Since this assumes 64 for all ML-KEM, might as well double down in the RAND_bytes call:
```c
uint8_t seed[64];
RAND_bytes(seed, 64);
```
* Can we also add a debug assertion just for sanity?
```c
// Assertion for sanity
assert(kem->keygen_seed_len == 64);
```

crypto/evp_extra/p_kem_asn1.c
```diff
@@ -226,22 +226,39 @@ static int kem_priv_encode(CBB *out, const EVP_PKEY *pkey) {
...
-      !CBB_add_asn1(&pkcs8, &private_key, CBS_ASN1_OCTETSTRING) || // // privateKey: OCTET STRING (outer container)
-      !CBB_add_asn1(&private_key, &expanded_key, CBS_ASN1_OCTETSTRING) || // expandedKey CHOICE variant, AWS-LC uses expandedKey for the moment
-      !CBB_add_bytes(&expanded_key, key->secret_key, kem->secret_key_len) || // raw private key 
-      !CBB_flush(out)) {
+      !CBB_add_asn1(&pkcs8, &private_key, CBS_ASN1_OCTETSTRING)) { // privateKey: OCTET STRING (outer container)
     OPENSSL_PUT_ERROR(EVP, EVP_R_ENCODE_ERROR);
     return 0;
   }
 
+  if (key->seed != NULL) {
```
justsmth, line 242: It's plausible that a consumer could want to serialize the older format to remain compatible with older ML-KEM implementations, but the API doesn't provide a mechanism like that.  Should we have a `EVP_PKEY_kem_clear_seed` or a `EVP_PKEY_kem_set_encoding_format` to facilitate this use-case?

crypto/fipsmodule/evp/p_kem.c
justsmth: For a check that's present even in release builds, we could use `BSSL_CHECK`:
```c
BSSL_CHECK(kem->keygen_seed_len == 64);
```

crypto/evp_extra/p_kem_asn1.c
geedo0: My thinking around this is that I agree that this is plausible, but we also don't have to add this functionality until we know for sure that this use case exists. In the description, I mentioned that all major implementers of this spec happily accept the seed format already. Adding new APIs now would lock us into supporting it forever.

crypto/fipsmodule/evp/p_kem.c
geedo0: Mixed about it. I think the assertion is good, but the rest of the code base standardizes on `kem->keygen_seed_len` by convention. I don't see a strong reason to deviate locally. What we're seeing here is just the conflict between needing the value at compile time for the stack allocation vs using the KEM struct as the source of truth at run time.

crypto/evp_extra/p_kem_asn1.c
justsmth: Worst case, there is already a workaround hack:
```c
size_t sk_len = 0;
EVP_PKEY_get_raw_private_key(pkey, NULL, &sk_len);
uint8_t *sk = malloc(sk_len);
// Extract private key
EVP_PKEY_get_raw_private_key(pkey, sk, &sk_len);

// Create key without seed
EVP_PKEY *expanded_pkey = EVP_PKEY_kem_new_raw_secret_key(nid, sk, sk_len);
```

dkostic (approved):

justsmth (approved):

PR #3157, opened 2026-04-09
Fix Clang 19 C++ headers and LLVM tool version mismatches on AL2023

justsmth:
### Issues:
Follow-up to #3152

### Description of changes:
The `setup-clang-19.sh` script sets `CC` and `CXX` but doesn't update `PATH`, so CMake still picks up `/usr/bin/llvm-ar` and `/usr/bin/llvm-ranlib` from the dnf-installed LLVM 15. These can't read object files produced by Clang 19 (`Unknown attribute kind (86) — Producer: 'LLVM19.1.7' Reader: 'LLVM 15.0.7'`). This appends `export PATH=/opt/clang-19/bin:$PATH` to the generated setup script so the matching LLVM 19 tools are found first.

### Testing:
Verified locally in the aarch64 Docker image that `which llvm-ar` resolves to `/opt/clang-19/bin/llvm-ar` after sourcing the updated script, and the ASAN build compiles and links successfully.

### DIFF
--- a/.github/docker_images/aws-lc/amazonlinux/Dockerfile.al2023
+++ b/.github/docker_images/aws-lc/amazonlinux/Dockerfile.al2023
@@ -89,6 +89,10 @@ COPY --from=scripts create-compiler-env.sh /tmp
 RUN <<EOF
 set -ex
 /tmp/create-compiler-env.sh
+# Ensure setup-clang-19.sh also prepends /opt/clang-19/bin to PATH so that
+# matching versions of llvm-ar, llvm-ranlib, llvm-symbolizer, etc. are used
+# instead of the system LLVM 15 versions.
+echo 'export PATH=/opt/clang-19/bin:$PATH' >> /opt/compiler-env/setup-clang-19.sh
 EOF
 
 # Install SSM Agent Config


geedo0 (approved):

.github/docker_images/aws-lc/amazonlinux/Dockerfile.al2023
```diff
@@ -89,6 +89,10 @@ COPY --from=scripts create-compiler-env.sh /tmp
 RUN <<EOF
 set -ex
 /tmp/create-compiler-env.sh
+# Ensure setup-clang-19.sh also prepends /opt/clang-19/bin to PATH so that
+# matching versions of llvm-ar, llvm-ranlib, llvm-symbolizer, etc. are used
+# instead of the system LLVM 15 versions.
+echo 'export PATH=/opt/clang-19/bin:$PATH' >> /opt/compiler-env/setup-clang-19.sh
```
WillChilds-Klein, line 95: do we need to do this for other OS's or just AL?

https://github.com/search?q=repo%3Aaws%2Faws-lc%20setup-clang-19.sh&type=code

WillChilds-Klein (approved):

.github/docker_images/aws-lc/amazonlinux/Dockerfile.al2023
justsmth: Yes, this is specific to AL2023. It's the only image where we installed a pre-built Clang 19 alongside a system LLVM 15 from `dnf`. The two LLVM versions provide conflicting tools at different paths (`/opt/clang-19/bin/llvm-ar` vs `/usr/bin/llvm-ar`).

On Ubuntu 24.04, for example, Clang 14–19 are all installed via `apt` and the LLVM tools come in versioned variants (`llvm-ar-14`, `llvm-ar-19`, etc.) that don't conflict. The `create-compiler-env.sh` pattern of just setting `CC`/`CXX` works fine there.

PR #3158, opened 2026-04-10
Shard valgrind CI job to avoid GitHub Actions timeout

justsmth:
### Description of changes:

The `valgrind-x86_64` job in `linux-multi-arch-omnibus` is frequently timing out. When successful it barely squeaks by at [2:53:48](https://github.com/aws/aws-lc/actions/runs/24212207410/job/70683671713), and it recently [timed out at 3:09:21](https://github.com/aws/aws-lc/actions/runs/24211552799/job/70681415264) (GitHub's default job timeout is 6 hours, but CodeBuild has a 3-hour limit).

This PR adds job-level sharding to `util/all_tests.go` so the valgrind work can be split across multiple parallel CI jobs. Two new environment variables control this: `AWS_LC_VALGRIND_SHARD_INDEX` and `AWS_LC_VALGRIND_TOTAL_SHARDS`. The workflow now runs 6 shards per architecture instead of 1.

### Call-outs:

The sharding works by multiplying the GTest shard count by the number of job shards. With 8 cores and 6 job shards, `crypto/crypto_test` gets 48 GTest shards total instead of 8. Each CI job runs a contiguous block of 8 shards on its local cores, so each shard covers 1/48th of the test cases instead of 1/8th. This brings individual job runtime from ~3 hours down to ~30 minutes.

When the env vars are unset, `all_tests.go` behaves identically to before — no existing usage is affected.

The shard count is easy to tune: just change the `shard` matrix list and `AWS_LC_VALGRIND_TOTAL_SHARDS` value in the workflow.

### Testing:

The `all_tests.go` change is exercised by the valgrind CI jobs themselves. The sharding logic can also be verified locally by setting the environment variables and confirming that `all_tests.go` partitions the test list correctly.

### DIFF
--- a/.github/workflows/linux-multi-arch-omnibus.yml
+++ b/.github/workflows/linux-multi-arch-omnibus.yml
@@ -396,17 +396,20 @@ jobs:
             ./tests/ci/run_prefix_tests.sh
 
   valgrind:
-    name: valgrind-${{ matrix.arch }}
+    name: valgrind-${{ matrix.arch }}-${{ matrix.shard }}
     runs-on:
       - codebuild-aws-lc-ci-github-actions-${{ github.run_id }}-${{ github.run_attempt }}
         image:${{ matrix.arch == 'x86_64' && 'linux-5.0' || matrix.arch == 'aarch64' && 'arm-3.0' }}
         instance-size:large
     env:
       AWS_LC_GO_TEST_TIMEOUT: 60m
+      AWS_LC_VALGRIND_SHARD_INDEX: ${{ matrix.shard }}
+      AWS_LC_VALGRIND_TOTAL_SHARDS: 6
     strategy:
       fail-fast: false
       matrix:
         arch: [x86_64, aarch64]
+        shard: [0, 1, 2, 3, 4, 5]
     steps:
       - uses: actions/checkout@v5
       - name: Login to Amazon ECR
@@ -418,6 +421,8 @@ jobs:
           image: ${{ steps.login-ecr.outputs.registry }}/aws-lc/amazonlinux:2023
           env: |
             AWS_LC_GO_TEST_TIMEOUT
+            AWS_LC_VALGRIND_SHARD_INDEX
+            AWS_LC_VALGRIND_TOTAL_SHARDS
           run: |
             source /opt/compiler-env/setup-gcc-11.sh
             ./tests/ci/run_valgrind_tests.sh
--- a/.github/workflows/linux-multi-arch-omnibus.yml
+++ b/.github/workflows/linux-multi-arch-omnibus.yml
@@ -404,6 +404,8 @@ jobs:
     env:
       AWS_LC_GO_TEST_TIMEOUT: 60m
       AWS_LC_VALGRIND_SHARD_INDEX: ${{ matrix.shard }}
+      # NOTE: TOTAL_SHARDS must equal the number of values in matrix.shard below.
+      # If they drift, some tests will silently never run.
       AWS_LC_VALGRIND_TOTAL_SHARDS: 6
     strategy:
       fail-fast: false
--- a/util/all_tests.go
+++ b/util/all_tests.go
@@ -92,9 +92,10 @@ type test struct {
 }
 
 type result struct {
-	Test   test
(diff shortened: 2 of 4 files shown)


.github/workflows/linux-multi-arch-omnibus.yml
```diff
@@ -396,17 +396,20 @@ jobs:
...
         instance-size:large
     env:
       AWS_LC_GO_TEST_TIMEOUT: 60m
+      AWS_LC_VALGRIND_SHARD_INDEX: ${{ matrix.shard }}
+      AWS_LC_VALGRIND_TOTAL_SHARDS: 6
     strategy:
       fail-fast: false
       matrix:
         arch: [x86_64, aarch64]
+        shard: [0, 1, 2, 3, 4, 5]
```
WillChilds-Klein, line 414: is there some way to ensure that these numbers stay in sync? i.e. descriptive error if we remove a `shard` value without changing TOTAL_SHARDS? perhaps something like `${{ len(matrix.shard) }}` if such a syntax exists?

in that scenario, we'd drop a shard and lose 15% of testing coverage.

or perhaps just a loud ALL-CAPS comment?

util/all_tests.go
```diff
@@ -43,6 +43,44 @@ var (
...
+	v, err := strconv.Atoi(s)
+	if err != nil {
+		return defaultVal
+	}
+	return v
+}
+
+func initJobSharding() {
+	jobShardIndex = getEnvInt("AWS_LC_VALGRIND_SHARD_INDEX", 0)
+	jobTotalShards = getEnvInt("AWS_LC_VALGRIND_TOTAL_SHARDS", 0)
```
WillChilds-Klein: don't we want to default to the number of cores on the current machine?

justsmth: The names here are misleading. `jobTotalShards` is the number of CI jobs (separate machines) splitting the work, which is separate from `numWorkers` (which already defaults to `runtime.NumCPU()`).

I'll change the names to be more clear.

.github/workflows/linux-multi-arch-omnibus.yml
justsmth: I don't think GitHub provides a way to get a `len()` for an array inside an expression, but I can add a more clear comment about it.

WillChilds-Klein (dismissed):

dougch (dismissed):
Couple not-blocking questions... 
- How deterministic are valgrinds shards? If shard 3/6 fails in CI will running locally with shard 3/6 also fail?
- Do you have a failure case? What does it look like when a shard fails?

justsmth:
Great questions! This motivated a couple of improvements in the latest push.

> * How deterministic are valgrinds shards? If shard 3/6 fails in CI will running locally with shard 3/6 also fail?

The sharding is deterministic given the same values of `AWS_LC_NUM_TEST_WORKERS`, `AWS_LC_VALGRIND_SHARD_INDEX`, and `AWS_LC_VALGRIND_TOTAL_SHARDS`. The total GTest shard count is `numWorkers × ciJobCount`, and test cases are assigned to shards via `test_index % total_shards`. Since `numWorkers` defaults to `runtime.NumCPU()`, the partition will differ between machines with different core counts. I added `AWS_LC_NUM_TEST_WORKERS` as an env var override so you can match the CI runner's core count without needing to pass CLI flags through CMake.

> * Do you have a failure case? What does it look like when a shard fails?

On failure, `all_tests.go` now prints the exact env vars and command needed to reproduce. For example:

```
1 of 89 tests failed:
    crypto/crypto_test [shard 9/48]

To reproduce this failure locally, run with:
  AWS_LC_NUM_TEST_WORKERS=8 AWS_LC_VALGRIND_SHARD_INDEX=1 AWS_LC_VALGRIND_TOTAL_SHARDS=6 go run util/all_tests.go -build-dir <build-dir> -valgrind

To run the failing shard directly:
  GTEST_TOTAL_SHARDS=48 GTEST_SHARD_INDEX=9 valgrind ./<build-dir>/crypto/crypto_test

dougch (approved):

WillChilds-Klein (approved):

PR #3159, opened 2026-04-10
Bump minimum Go version to 1.20 and update Go dependencies

justsmth:
### Issues:
Addresses 
* #2856

### Description of changes:
The minimum Go version was pinned at 1.17, which forced `golang.org/x/crypto` to stay at v0.14.0 (the last version supporting Go 1.17). This bumps the minimum to Go 1.20 and updates the Go dependencies to current versions:

- `golang.org/x/crypto` v0.14.0 → v0.31.0
- `golang.org/x/sys` v0.13.0 → v0.28.0
- `golang.org/x/term` v0.13.0 → v0.27.0

The minimum version is updated consistently across `go.mod`, `BUILDING.md`, and `cmake/go.cmake`. The Go compatibility CI job is also updated to test against 1.20 instead of 1.17.13.

### Call-outs:
Dependabot opened #2856 which bumps `golang.org/x/crypto` to v0.45.0, requiring Go 1.24. This is a more conservative alternative that raises the floor to 1.20 — enough to unpin the `golang.org/x` dependencies without requiring a very recent Go toolchain.

### Testing:
Existing CI covers this. The `go.yml` workflow has been updated to test against the new minimum version.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -14,11 +14,11 @@ permissions:
   contents: read
 
 jobs:
-  go-version-1_17_13:
+  go-version-1_20:
     if: github.repository_owner == 'aws'
     env:
       GOROOT: "/usr/local/go"
-      GO_ARCHIVE: "go1.17.13.linux-amd64.tar.gz"
+      GO_ARCHIVE: "go1.20.linux-amd64.tar.gz"
     runs-on: ubuntu-latest
     steps:
       - name: Install OS Dependencies
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -20,7 +20,7 @@ If in doubt, use the most recent stable version of each build tool.
     `PERL_EXECUTABLE`.
     * To build without Perl (not recommended) see [this section.](#using-pre-generated-build-files)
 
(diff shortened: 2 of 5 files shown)


samuel40791765:
I think we broke some other team that was still pinning to the specific version a while back. Not sure which team, but we might have to sync up with them before doing this. 

@skmcgrail or @andrewhop might remember more details.

WillChilds-Klein (approved):

justsmth:
> I think we broke some other team that was still pinning to the specific version a while back. Not sure which team, but we might have to sync up with them before doing this.
> 
> @skmcgrail or @andrewhop might remember more details.

I checked with Sean and Andrew.  There have been issues on our FIPS branches ([for example](https://github.com/aws/aws-lc/pull/1725)) where we've needed to rollback to support an older Go version.  But I'm not aware of anyone consuming our main branch that would be unable to easily update the Go runtime in their build environment.

dougch (approved):
Nit; there are a few min. versions still on `1.18` (CI is pulling down much newer versions in most cases): actions-ci and cross-test and clang-tidy-review

PR #3161, opened 2026-04-13
Fix intermittent CA test failure on Windows CI when TEMP is unset

justsmth:
### Description of changes:

CA tests (`CATest.DatabaseWithRevokedEntryBasicTimestamp`, `CATest.CustomEndDate`, etc.) fail intermittently on the `windows-ltsc2022_sde-x86_64` CI job due to two issues with `WIN32_rename`:

1. **Temp files landing in `C:\Windows\`**: `GetTempPathA` falls back to the Windows directory when `TMP`, `TEMP`, and `USERPROFILE` are all unset — which is the case when running as SYSTEM in our Docker-based CI. This introduces a `GetSafeTempPathA` wrapper that detects the fallback and redirects to `C:\Windows\Temp\` instead.

2. **NTFS pending-delete race in `WIN32_rename`**: The `DeleteFile` + `MoveFile` workaround for `ERROR_ALREADY_EXISTS` is racy. NTFS defers file removal until all handles are closed, so `DeleteFile` can return success while the file is still visible. The subsequent `MoveFile` fails with `ERROR_ALREADY_EXISTS` (183), which wasn't in the retryable error set, so the retry budget was never used. Replaced with `MoveFileEx` using `MOVEFILE_REPLACE_EXISTING`, which atomically replaces the target — matching POSIX `rename()` semantics.

### Call-outs:

`CATest::TearDown` was missing cleanup of `.new` and `.attr.new` files. These get left behind when a test fails between `SaveIndex` and `RotateIndex`. Added while in here.

See failure [here](https://github.com/aws/aws-lc/actions/runs/24354082397/job/71116230692?pr=2934#step:4:3018).

### Testing:

Existing CA tests cover this path. The fix should be verified by the `windows-ltsc2022_sde-x86_64` job no longer failing intermittently.

### DIFF
--- a/crypto/test/test_util.cc
+++ b/crypto/test/test_util.cc
@@ -182,11 +182,42 @@ bool PEM_to_DER(const char *pem_str, uint8_t **out_der, long *out_der_len) {
 }
 
 #if defined(OPENSSL_WINDOWS)
+// GetTempPathA falls back to the Windows directory (e.g. C:\Windows\) when the
+// TMP, TEMP, and USERPROFILE environment variables are all unset. This commonly
+// happens when running as SYSTEM in Docker containers or CI agents. The Windows
+// directory has special protections that cause file rename operations to fail
+// intermittently. Detect this case and redirect to C:\Windows\Temp\ instead.
+static DWORD GetSafeTempPathA(DWORD nBufferLength, LPSTR lpBuffer) {
+  DWORD ret = GetTempPathA(nBufferLength, lpBuffer);
+  if (ret == 0 || ret >= nBufferLength) {
+    return ret;
+  }
+  char win_dir[PATH_MAX];
+  UINT win_len = GetWindowsDirectoryA(win_dir, sizeof(win_dir));
+  if (win_len == 0 || win_len >= sizeof(win_dir)) {
+    return ret;
+  }
+  // Append trailing backslash to match GetTempPathA's format for comparison.
+  if (win_len + 1 >= sizeof(win_dir)) {
+    return ret;
+  }
+  win_dir[win_len] = '\\';
+  win_dir[win_len + 1] = '\0';
+  if (_stricmp(lpBuffer, win_dir) == 0) {
+    int written = snprintf(lpBuffer, nBufferLength, "%sTemp\\", win_dir);
+    if (written < 0 || (DWORD)written >= nBufferLength) {
+      return 0;
+    }
+    ret = (DWORD)written;
+  }
+  return ret;
+}
+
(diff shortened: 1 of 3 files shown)


crypto/test/test_util.cc
```diff
@@ -182,11 +182,42 @@ bool PEM_to_DER(const char *pem_str, uint8_t **out_der, long *out_der_len) {
 }
 
 #if defined(OPENSSL_WINDOWS)
+// GetTempPathA falls back to the Windows directory (e.g. C:\Windows\) when the
+// TMP, TEMP, and USERPROFILE environment variables are all unset. This commonly
+// happens when running as SYSTEM in Docker containers or CI agents. The Windows
+// directory has special protections that cause file rename operations to fail
+// intermittently. Detect this case and redirect to C:\Windows\Temp\ instead.
```
WillChilds-Klein, line 189: why don't we just always use `Temp`? or perhaps a random directory name under `Temp`?

WillChilds-Klein (approved):

crypto/test/test_util.cc
justsmth: I think only `C:\Windows\Temp` is guaranteed to exist. Any other assumption might not hold.

justsmth: I believe the directory must exist for this use case.

crypto/test/test_util.cc
```diff
@@ -182,11 +182,42 @@ bool PEM_to_DER(const char *pem_str, uint8_t **out_der, long *out_der_len) {
...
+    return ret;
+  }
+  win_dir[win_len] = '\\';
+  win_dir[win_len + 1] = '\0';
+  if (_stricmp(lpBuffer, win_dir) == 0) {
+    int written = snprintf(lpBuffer, nBufferLength, "%sTemp\\", win_dir);
+    if (written < 0 || (DWORD)written >= nBufferLength) {
+      return 0;
+    }
+    ret = (DWORD)written;
```
nebeid, line 211: Suggestion: the current code assumes C:\Windows\Temp\ exists; it almost always does on Windows. But if you want to ensure it does, the following call doesn't harm, and no need to check for return value if it returns false because the directory exists.
```suggestion
    // Ensure the directory exists                                                                                                                                                          
    CreateDirectoryA(lpBuffer, NULL); 
    ret = (DWORD)written;
```

justsmth: I don't think we'd have permissions to create the `C:\Windows\Temp` directory as `C:\Windows\` is a system directory.

nebeid (approved):

PR #3167, opened 2026-04-15
Fix FIPS build under MSAN by broadening integrity-test guards

justsmth:
### Issues:
* In `aws-lc-rs`: https://github.com/aws/aws-lc-rs/issues/1077
  * Related PR: https://github.com/aws/aws-lc-rs/pull/1100

### Description of changes:

CMake already skips the delocator (and linker script) for both ASAN and MSAN FIPS builds, but `bcm.c` and `fips.c` only guarded the integrity-test symbol references with `#if !defined(OPENSSL_ASAN)`. This means an MSAN+FIPS build still `extern`-declares `BORINGSSL_bcm_text_start/end/hash` and tries to call `BORINGSSL_integrity_test()`, even though nothing provides those symbols — resulting in link failures.

This broadens all three guards in `bcm.c` and the `FIPS_mode()` check in `fips.c` to `!defined(OPENSSL_ASAN) && !defined(OPENSSL_MSAN)`, matching the existing guard in `delocate.h`.

### Call-outs:

`FIPS_mode()` now returns 0 under MSAN, just as it does under ASAN. `is_fips_build()` is unaffected — it still returns 1 for any `BORINGSSL_FIPS` build regardless of sanitizer, which is the intended distinction between the two.

### Testing:

Added a targeted MSAN+FIPS CI job (`ubuntu:22.04` / `clang-14` / `x86_64`) in `linux-multi-arch-omnibus.yml`, gated behind `AWSLC_ENABLE_FIPS_MSAN`. The corresponding `run_fips_tests.sh` block builds with `-DFIPS=1 -DMSAN=1 -DUSE_CUSTOM_LIBCXX=1`, verifies `bssl isfips` returns 0, and runs the test suite.

### DIFF
--- a/tests/ci/run_fips_tests.sh
+++ b/tests/ci/run_fips_tests.sh
@@ -81,6 +81,16 @@ if [[ "${AWSLC_ENABLE_FIPS_ASAN:-0}" == "1" ]]; then
   fi
 fi
 
+if [[ "${AWSLC_ENABLE_FIPS_MSAN:-0}" == "1" ]]; then
+  # The ACVP modulewrapper subprocess crashes under MSAN, so we use
+  # run_build + all_tests.go instead of fips_build_and_test to skip the
+  # ACVP tests. On ARM, MSAN also gets stuck on PoolTest.Threads
+  # (https://github.com/aws/aws-lc/issues/13).
+  echo "Building with Clang and testing AWS-LC in FIPS Release mode with memory sanitizer."
+  run_build -DFIPS=1 -DMSAN=1 -DUSE_CUSTOM_LIBCXX=1 -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=1
+  go run util/all_tests.go -build-dir "$BUILD_ROOT"
+fi
+
 echo "Testing shared AWS-LC in FIPS Debug mode in a different folder."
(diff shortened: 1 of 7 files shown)


tests/ci/run_fips_tests.sh
```diff
@@ -81,6 +81,18 @@ if [[ "${AWSLC_ENABLE_FIPS_ASAN:-0}" == "1" ]]; then
   fi
 fi
 
+if [[ "${AWSLC_ENABLE_FIPS_MSAN:-0}" == "1" ]]; then
```
WillChilds-Klein, line 84: for this and ASAN, do we really want to do the other redundant test runs on lines 25, 28, 33, 62, 67, and 98?

also, should `AWSLC_ENABLE_FIPS_ASAN` and `AWSLC_ENABLE_FIPS_ASAN` be mutually exclusive?

justsmth: This is a script that I'd like to refactor up so that jobs can complete in under an hour -- possibly have tests setup to provide coverage for more specific build configurations rather than "check everything every time".

Let me think about how to split this up. I'll likely defer this to a subsequent PR.

WillChilds-Klein (approved):

nhatnghiho (approved):

PR #3169, opened 2026-04-16
Handle id-pkix-ocsp-nocheck in OCSP responder verification

dkostic:
### Issues:
D431426583

### Description of changes: 
When verifying a delegated OCSP responder certificate, check for the id-pkix-ocsp-nocheck extension (RFC 6960 section 4.2.2.2.1). If present, clear X509_V_FLAG_CRL_CHECK from the verification parameters so that CRL checking is not required for the responder certificate.

This brings AWS-LC to parity with OpenSSL's handling of this extension.

### Call-outs:

### Testing:

### DIFF
--- a/crypto/ocsp/ocsp_test.cc
+++ b/crypto/ocsp/ocsp_test.cc
@@ -2054,3 +2054,149 @@ TEST(OCSPTest, OCSP_SINGLERESP) {
                                   X509_EXTENSION_get_data(ext.get())),
             0);
 }
+
+// Helper to create an OCSP responder cert issued by |ca_cert|/|ca_key|, with
+// EKU=OCSP Signing. If |add_nocheck| is true, the id-pkix-ocsp-nocheck
+// extension is added.
+static bssl::UniquePtr<X509> MakeOCSPResponderCert(EVP_PKEY *responder_key,
+                                                   X509 *ca_cert,
+                                                   EVP_PKEY *ca_key,
+                                                   bool add_nocheck) {
+  bssl::UniquePtr<X509> cert(X509_new());
+  if (!cert ||
+      !X509_set_version(cert.get(), X509_VERSION_3) ||
+      !ASN1_INTEGER_set(X509_get_serialNumber(cert.get()), 1) ||
+      !X509_set_issuer_name(cert.get(), X509_get_subject_name(ca_cert)) ||
+      !X509_NAME_add_entry_by_txt(
+          X509_get_subject_name(cert.get()), "CN", MBSTRING_UTF8,
+          reinterpret_cast<const uint8_t *>("OCSP Responder"), -1, -1, 0) ||
+      !X509_set_pubkey(cert.get(), responder_key) ||
+      !ASN1_TIME_adj(X509_getm_notBefore(cert.get()), kReferenceTime, -1, 0) ||
+      !ASN1_TIME_adj(X509_getm_notAfter(cert.get()), kReferenceTime, 1, 0)) {
+    return nullptr;
+  }
+
+  // Add EKU: OCSP Signing (required for delegated responder).
+  bssl::UniquePtr<STACK_OF(ASN1_OBJECT)> eku(sk_ASN1_OBJECT_new_null());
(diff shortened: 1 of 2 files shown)


crypto/ocsp/ocsp_verify.c
```diff
@@ -123,7 +123,7 @@ static int ocsp_setup_untrusted(OCSP_BASICRESP *bs, STACK_OF(X509) *certs,
 }
 
 
-static int ocsp_verify_signer(X509 *signer, X509_STORE *st,
+static int ocsp_verify_signer(X509 *signer, int response, X509_STORE *st,
                               STACK_OF(X509) *untrusted,
                               STACK_OF(X509) **chain) {
```
justsmth: This static function is only called from line #377, which hard-code `response=1`. There's no need to add a `response` parameter -- just assume it's true below and simplify the one place it's used.

crypto/ocsp/ocsp_verify.c
```diff
@@ -142,6 +143,17 @@ static int ocsp_verify_signer(X509 *signer, X509_STORE *st,
     OPENSSL_PUT_ERROR(OCSP, ERR_R_X509_LIB);
     goto end;
   }
+  if ((vp = X509_STORE_CTX_get0_param(ctx)) == NULL) {
+    goto end;
```
justsmth: Add a `OPENSSL_PUT_ERROR`?

justsmth (commented):
I think we should also add tests around this.

justsmth (dismissed):

crypto/ocsp/ocsp_verify.c
```diff
@@ -142,6 +143,17 @@ static int ocsp_verify_signer(X509 *signer, X509_STORE *st,
     OPENSSL_PUT_ERROR(OCSP, ERR_R_X509_LIB);
     goto end;
   }
+  if ((vp = X509_STORE_CTX_get0_param(ctx)) == NULL) {
+    OPENSSL_PUT_ERROR(OCSP, ERR_R_X509_LIB);
+    goto end;
```
justsmth: NP: Since `vp` is only used inside the next `if` block, could we move this down?
```c
  // RFC 6960 section 4.2.2.2.1: if the responder certificate has the
  // id-pkix-ocsp-nocheck extension, the CA has indicated that the client
  // should trust the responder for its lifetime without revocation checking.
  // Locally disable CRL-based revocation checking in this case.
  if (X509_get_ext_by_NID(signer, NID_id_pkix_OCSP_noCheck, -1) >= 0) {
    X509_VERIFY_PARAM *vp = X509_STORE_CTX_get0_param(ctx);
    if (vp == NULL) {
      OPENSSL_PUT_ERROR(OCSP, ERR_R_X509_LIB);
      goto end;
    }
    X509_VERIFY_PARAM_clear_flags(vp, X509_V_FLAG_CRL_CHECK);
  }
```

torben-hansen: Second this.

crypto/ocsp/ocsp_test.cc
```diff
@@ -2054,3 +2054,149 @@ TEST(OCSPTest, OCSP_SINGLERESP) {
...
+                                       this_update.get(), next_update.get()));
+
+    ASSERT_TRUE(OCSP_basic_sign(bs.get(), responder_cert.get(), resp_key.get(),
+                                EVP_sha256(),
+                                CertsToStack({ca_cert.get()}).get(), 0));
+
+    int ret = OCSP_basic_verify(
+        bs.get(), CertsToStack({responder_cert.get()}).get(), trust_store.get(),
+        0);
+    EXPECT_NE(ret, OCSP_VERIFYSTATUS_SUCCESS)
```
torben-hansen, line 2198: negative test error code could be a bit more specific. But nitty.

crypto/ocsp/ocsp_verify.c
dkostic: done

torben-hansen (approved):

justsmth (approved):

PR #3171, opened 2026-04-16
Mitigate intermittent SSL runner timeouts on FreeBSD CI

justsmth:
### Description of changes:

The FreeBSD 15.0 (x86-64) CI job fails intermittently on CBC record splitting tests (`CBCRecordSplitting-AES128`, `CBCRecordSplittingPartialWrite-AES128`) with runner-side TCP timeouts like `read tcp [::1]:...->[::1]:...: i/o timeout`. The QEMU-emulated FreeBSD VM is occasionally too slow to respond within the runner's default 30-second idle timeout.

The runner already has retry logic for shim-side idle timeouts and transient signals, but runner-side TCP timeouts (`net.Error` with `Timeout() == true`) fall through without a retry — the test just fails.

This PR does two things:
1. Increases the idle timeout to 120s and enables retry-on-timeout for the FreeBSD CI job.
2. Adds a new `-retry-on-timeout` flag to the SSL test runner that retries tests on runner-side TCP timeouts. A real bug will fail again on the retry.

### Call-outs:

The `-retry-on-timeout` flag is off by default. It's gated behind `AWS_LC_SSL_RUNNER_RETRY_ON_TIMEOUT` at the CMake level, currently only enabled for FreeBSD. If similar flakiness shows up on NetBSD or OpenBSD, we can add the same env vars to those jobs.

This PR also fixes a pre-existing variable shadowing bug in the retry block: `shim, err := newShimProcess(...)` was creating a new `shim` scoped to the `if` block, so after a successful retry the code would still read `stdout`/`stderr` from the original failed shim. This caused retried tests to fail with `unexpected error output` even when the retry passed. Changed to `shim, err = ...` to reassign the outer variable.

### Testing:

This addresses a CI flakiness issue — the fix will be validated by observing the FreeBSD job over subsequent runs. The retry logic is structurally identical to the existing `shim.idled` retry path.

### DIFF
--- a/ssl/test/runner/runner.go
+++ b/ssl/test/runner/runner.go
@@ -69,6 +69,7 @@ var (
 	includeDisabled    = flag.Bool("include-disabled", false, "If true, also runs disabled tests.")
 	repeatUntilFailure = flag.Bool("repeat-until-failure", false, "If true, the first selected test will be run repeatedly until failure.")
 	// Added by aws-lc
+	retryOnTimeout     = flag.Bool("retry-on-timeout", false, "If true, retry tests that fail due to a runner-side TCP timeout. Useful for resource-constrained environments (e.g. emulated VMs).")
 	sslTransferConfig  = flag.String("ssl-transfer-test-file", "", "A path to file which includes the test names that can be converted for SSL transfer.")
 	sslFuzzSeedDir     = flag.String("ssl-fuzz-seed-dir", "", "The directory in which to write the output of |SSL_to_bytes|.")
 	testCaseStartIndex = flag.Int("test-case-start-index", -1, "If non-negative, test case is filtered in if the index in |testCases| >= test-case-start-index.")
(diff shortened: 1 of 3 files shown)


ssl/test/runner/runner.go
```diff
@@ -69,6 +69,7 @@ var (
 	includeDisabled    = flag.Bool("include-disabled", false, "If true, also runs disabled tests.")
 	repeatUntilFailure = flag.Bool("repeat-until-failure", false, "If true, the first selected test will be run repeatedly until failure.")
 	// Added by aws-lc
+	retryOnTimeout     = flag.Bool("retry-on-timeout", false, "If true, retry tests that fail due to a runner-side TCP timeout. Useful for resource-constrained environments (e.g. emulated VMs).")
```
geedo0, line 72: Is there any downsides to enabling this by default? Seems like it'll improve stability generally at the cost of having to retry a few more times in the event of a "true failure".

geedo0 (approved):

ssl/test/runner/runner.go
justsmth: My "fear" is that it could potentially hide a problem. I feel more confident by only enabling it for jobs that frequently have this failure.


If the flakiness gets more painful, we can consider making it default.

sgmenda (approved):

PR #3172, opened 2026-04-16
Add WASIp2 build and test support

justsmth:
### Description of changes:

Add support for building and testing AWS-LC on the `wasm32-wasip2` target using the WASI SDK. WASI Preview 2 does not support pthreads, BSD sockets, or terminal I/O, so the corresponding capability flags (`OPENSSL_NO_SOCK`, `OPENSSL_NO_TTY`, `OPENSSL_NO_THREADS_CORRUPT_MEMORY_AND_LEAK_SECRETS_IF_THREADED`) are set when the new `OPENSSL_WASM_WASI` macro is defined. WASI does provide filesystem access and `getentropy()`, so those capabilities are left intact.

New files:
- `util/wasi-sdk-toolchain.cmake` — CMake toolchain file for cross-compiling to `wasm32-wasip2`.
- `tests/ci/run_wasi_tests.sh` — CI script that builds AWS-LC and runs `crypto_test`, `ssl_test`, and `urandom_test` under wasmtime.
- `.github/workflows/wasi.yml` — CI workflow for WASI builds.

### Call-outs:
- `boringssl_gtest_main` is built as an `OBJECT` library (instead of `STATIC`) on WASI to work around wasm-ld not pulling `main` from static archives.
- The `iovec` redefinition guard in `ssl/bio_ssl.cc` is scoped to `OPENSSL_WASM_WASI`; WASI's libc provides `iovec` through standard headers so the redefinition would be a conflict.
- Tests requiring fork, sockets, or threads are excluded via gtest filter at runtime rather than additional compile-time guards — this keeps the source changes minimal.

### Testing:
The `wasi.yml` CI workflow builds with WASI SDK 25 and runs tests under wasmtime 29.0.1. `crypto_test`, `ssl_test`, and `urandom_test` are executed with a gtest filter excluding tests that require fork, sockets, threads, or unsupported file operations.


util/wasi-sdk-toolchain.cmake
```diff
@@ -0,0 +1,122 @@
...
+set(WASI_COMPILE_FLAGS "-D_WASI_EMULATED_SIGNAL -D_WASI_EMULATED_PROCESS_CLOCKS")
+
+# WASI-specific linker flags
+# -lwasi-emulated-signal: Link signal emulation library
+# -lwasi-emulated-process-clocks: Link process clocks emulation
+# Memory configuration flags:
+# -Wl,-z,stack-size=1048576: Set stack size to 1MB (default is typically 64KB).
+#   Some crypto tests allocate large buffers on the stack (e.g., 128KB for digest tests).
+# -Wl,--initial-memory=134217728: Set initial linear memory to 128MB.
+#   Large test binaries need sufficient memory for test data and operations.
```
dougch, line 102: Does this warrant a WASI_TEST_MODE? Does a downstream consumer need this much for a minimum or are constrained WASM environments not a need ?

.github/workflows/wasi.yml
```diff
@@ -0,0 +1,52 @@
...
+      - uses: actions/checkout@v4
+
+      - name: Install build dependencies
+        run: |
+          sudo apt-get update -o Acquire::Languages=none -o Acquire::Translation=none
+          sudo apt-get -y --no-install-recommends install cmake ninja-build
+
+      - name: Setup WASI SDK
+        run: |
+          curl -L https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-${WASI_SDK_VERSION}/wasi-sdk-${WASI_SDK_VERSION}.0-x86_64-linux.tar.gz | tar xz
```
dougch: having the arch as a variable would be nice; bit of a nit though (I'm validating this on arm).

dougch (dismissed):
Passed locally on my aarch64 instance.

util/wasi-sdk-toolchain.cmake
justsmth: I don't think a formal `WASI_TEST_MODE` is warranted here. 

The toolchain file is intended purely as our CI harness, not as a supported artifact for downstream consumers. Of course, they're welcome to draw inspiration from it, but we're not committing to it being a minimal/consumable build config.

.github/workflows/wasi.yml
justsmth: Addressed in latest commit.

dougch (approved):

nhatnghiho (approved):

PR #3177, opened 2026-04-16
ci: add gh-pages workflow for API documentation

dougch:
### Description of changes: 
Generate HTML API docs from header comments using the existing
doc.go tool (inherited from BoringSSL) and publish to gh-pages, on pushes to `main` where specific files have been touched, or a user with write manually kicks off the workflow.

There were some missing annotations that kiro filled in for me. 

### Call-outs:
I've pushed a sample to gh-pages to show how this would look at https://dougch.github.io/aws-lc/headers.html

The doxygen annotations weren't playing nice with the go doc tool, so the changes to [base64.h](https://github.com/aws/aws-lc/pull/2908) were reverted.

The repo settings will need updating post merge to allow an `app` vs. a `branch` to write to gh-pages; see the [docs](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site).

The first party action for publishing allows for fine grained publish controls, but does a complete replacement of the content (it's sending a tarball to an API vs. committing to a branch); this is a 2 way door...

### Testing:
Because of the `workflow_dispatch`, you can run these actions in a fork.... the new actions are working fine, example at https://dougch.github.io/aws-lc/headers.html

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -1,5 +1,6 @@
 name: API Documentation
 on:
+  workflow_dispatch:
   push:
     branches: [main]
     paths:
@@ -10,11 +11,20 @@ on:
       - '.github/workflows/api-docs.yml'
 
 permissions:
-  contents: write
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: pages
+  cancel-in-progress: false
 
 jobs:
   publish-api-docs:
     runs-on: ubuntu-latest
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v4
@@ -24,10 +34,11 @@ jobs:
         run: |
           mkdir -p api-output
           cd util
-          go run doc.go --config doc.config --out ../api-output
-      - name: Deploy to gh-pages
-        uses: peaceiris/actions-gh-pages@v4
+          go run doc.go --config doc.config --out ../pages-output
+      - name: Upload pages artifact
+        uses: actions/upload-pages-artifact@v3
         with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_dir: ./api-output
-          destination_dir: api
+          path: ./pages-output
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v4
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -1,0 +1,33 @@
+name: API Documentation
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'include/openssl/**'
+      - 'util/doc.go'
+      - 'util/doc.config'
+      - 'util/doc.css'
+      - '.github/workflows/api-docs.yml'
+
+permissions:
+  contents: write
+
+jobs:
+  publish-api-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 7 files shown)


.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ">=1.22"
+      - name: Generate API documentation
+        run: |
+          mkdir -p api-output
+          cd util
+          go run doc.go --config doc.config --out ../api-output
+      - name: Deploy to gh-pages
+        uses: peaceiris/actions-gh-pages@v4
```
WillChilds-Klein: given that this workflow uses our `GITHUB_TOKEN` w/ write permissions per L13, should we pin this 3p workflow to a particular commit hash?

.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,33 @@
...
+    branches: [main]
+    paths:
+      - 'include/openssl/**'
+      - 'util/doc.go'
+      - 'util/doc.config'
+      - 'util/doc.css'
+      - '.github/workflows/api-docs.yml'
+
+permissions:
+  contents: write
```
WillChilds-Klein: why do we need repo write permissions to publish to GH pages?

WillChilds-Klein (commented):
very nice!

.github/workflows/api-docs.yml
justsmth: Switching to GitHub's first-party Pages deployment ([`actions/upload-pages-artifact`](https://github.com/actions/upload-pages-artifact) + [`actions/deploy-pages`](https://github.com/actions/deploy-pages)) might also address Will's concern about `contents: write` since it uses the narrower `pages: write` scope.

.github/workflows/api-docs.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: |
+          mkdir -p api-output
+          cd util
+          go run doc.go --config doc.config --out ../api-output
+      - name: Deploy to gh-pages
+        uses: peaceiris/actions-gh-pages@v4
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./api-output
+          destination_dir: api
```
justsmth: The PR description mentions deploying to `/api/` to allow "other doc generators to coexist in future PRs".  It's worth adding `keep_files: true` to preserve content outside `/api/` to prevent it from wiping out other content.

justsmth (commented):
You mention "other doc generators to coexist in future PRs", but I'm not sure Doxygen can co-exist peacefully with the current documentation setup -- the syntax of one interferes with the other. (Perhaps this can be mitigated with some modifications to `doc.go`?)

For example: the output for `base64.h` has some issues: https://aws.github.io/aws-lc/api/base64.h.html

The Doxygen-style `/** ... */` comment blocks in that header aren't recognized by `doc.go`. The `/**` lines fall through to the declaration parser and get rendered as raw code blocks. The other modified headers don't have this problem since they only use `// ` style comments.

Should we also cleanup the `base64.h` as part of this?

.github/workflows/api-docs.yml
dougch: Yes, just pointing out that `gh-pages` is just a branch that can (co)-host more than one set of content, if we need to.

.github/workflows/api-docs.yml
dougch: I'll try converting to the other/newer actions, but we're essentially writing content directly to a branch without a PR, branch protection, etc...

justsmth (approved):

WillChilds-Klein (approved):

Issue #3165, opened 2026-04-15
Public accessor for ML-KEM algorithm NID


DarkaMaul:
### Problem:

AWS-LC provides `EVP_PKEY_pqdsa_get_type` (declared in [evp.h](https://github.com/aws/aws-lc/blob/be77a8ad3d8c016950e411bcf3b6a56baa097024/include/openssl/evp.h#L89)) to retrieve the specific algorithm NID from an `EVP_PKEY` of type `EVP_PKEY_PQDSA`. 

There is no equivalent function for `EVP_PKEY_KEM` keys.

Callers that receive an `EVP_PKEY` wrapping an ML-KEM key currently have no public API to determine which ML-KEM variant it holds. The NID is available internally, but it is not exposed through the public EVP interface.

### Solution:

Add `EVP_PKEY_kem_get_type`, following the same pattern:

```c
// EVP_PKEY_kem_get_type returns the |nid| of the configured KEM key.
// |pkey| must not be NULL.
OPENSSL_EXPORT int EVP_PKEY_kem_get_type(const EVP_PKEY *pkey);
```

* **Does this change any public APIs?**  Yes — adds one new exported function`EVP_PKEY_kem_get_type`. No existing APIs change.
* **Which algorithm(s) will this impact?** ML-KEM-*

### Requirements / Acceptance Criteria:

- [ ] Unit tests in `crypto/evp_extra/p_kem_test.cc` covering:
  - Returns correct NID for each ML-KEM variant (512, 768, 1024)
  - Returns 0 and sets error when called on a non-KEM `EVP_PKEY` (e.g. RSA or PQDSA key)
  - Returns 0 when called with an uninitialized KEM key

What must a solution address in order to solve the problem? How do we know the solution is complete?

* **RFC links:** N/A (API parity issue, not a new algorithm)
* **Related Issues:** None known
* **Will the Usage Guide or other documentation need to be updated?** The function should be documented in the header.
* **Testing:** How will this change be tested? Call out new integration tests, functional tests, or particularly
  interesting/important unit tests.
  * **Will this change trigger AWS LibCrypto Formal Verification changes?** No
  * **Should this change be fuzz tested?** No

Note: In my current PR in `pyca/cryptography`, I currently use the public key size to determine the variant.

/cc @alex

justsmth:
[...]

PR #3179, opened 2026-04-17
Add `EVP_PKEY_kem_get_type` public accessor for KEM key NID

justsmth:
### Issues:
* #3165

### Description of changes:
We expose `EVP_PKEY_pqdsa_get_type` for PQDSA keys but have no equivalent for KEM keys. This adds `EVP_PKEY_kem_get_type` to close that gap, following the same pattern. Without this, callers receiving an `EVP_PKEY` wrapping an ML-KEM key have no public API to determine which variant (512, 768, 1024) it holds.

### Call-outs:
The implementation in `evp.c` required adding `#include "../kem/internal.h"` — the PQDSA counterpart already had its internal header included but the KEM one was missing.

Existing tests in `ParsePublicKey` and `ParseExamplePrivateKey` were reaching into `kem_key->kem->nid` directly; those assertions now use the new public API instead.

### Testing:
- `All/KEMTest.GetType` — parameterized across all ML-KEM variants, verifies the correct NID is returned from a generated key pair.
- `KEMTest.GetTypeWrongKeyType` — verifies the function returns 0 and sets `EVP_R_EXPECTING_A_KEM_KEY` for non-KEM keys, covering both a classical (EC) key and an adjacent post-quantum (PQDSA) key so neighbouring types cannot slip past the type check.
- `KEMTest.GetTypeUninitializedKey` — verifies the function returns 0 without queueing an error when `pkey` is of type `EVP_PKEY_KEM` but has no `KEM_KEY` attached.


crypto/fipsmodule/evp/evp.c
```diff
@@ -257,6 +258,18 @@ int EVP_PKEY_pqdsa_get_type(const EVP_PKEY *pkey) {
...
 }
 
+int EVP_PKEY_kem_get_type(const EVP_PKEY *pkey) {
+  SET_DIT_AUTO_RESET;
+  if (pkey->type != EVP_PKEY_KEM) {
+    OPENSSL_PUT_ERROR(EVP, EVP_R_EXPECTING_A_KEM_KEY);
+    return 0;
+  }
+  if (!pkey->pkey.kem_key || !pkey->pkey.kem_key->kem) {
+    return 0;
```
torben-hansen, line 269: could set error to `EVP_R_NO_PARAMETERS_SET`?

justsmth: I agree. PR updated.

torben-hansen (approved):

dkostic (approved):

PR #3181, opened 2026-04-22
Map rsaesOaep SPKI to RSA in parse_key_type

crlorentzen:
Map rsaesOaep (OID 1.2.840.113549.1.1.7) SubjectPublicKeyInfo to RSA in parse_key_type(), enabling X509_get_pubkey() and X509_verify_cert() for TPM 1.2 EK certificates.

  ### Issues:
  P418478063

  ### Description of changes:

  AWS-LC's `parse_key_type()` in `crypto/evp_extra/evp_asn1.c` does not recognize `NID_rsaesOaep` (OID 1.2.840.113549.1.1.7) as a valid SubjectPublicKeyInfo algorithm. When `X509_get_pubkey()` encounters a certificate using this OID, the OID lookup fails and returns `PUBLIC_KEY_DECODE_ERROR`, which in turn causes `X509_verify_cert()` to fail with error code 1.

  TPM 1.2 Endorsement Key certificates use `rsaesOaep` as their SPKI algorithm per the [TCG Credential Profiles Specification V1.2](https://trustedcomputinggroup.org/wp-content/uploads/Credential_Profiles_V1.2_Level2_Revision8.pdf) (see section 3.2.7 for the RSAES-OAEP OID definition). The underlying key is a standard RSA key, only the algorithm identifier differs from `rsaEncryption`.

This change adds `NID_rsaesOaep` alongside the existing `NID_rsa` special case in `parse_key_type()`, returning `rsa_asn1_meth` for both. This is the same approach used for the `NID_rsa` (ITU-T X.509) alias added in #1754.

### Call-outs:

- The fix is intentionally minimal — a single additional NID check appended to the existing `NID_rsa` condition. No new ASN.1 methods or OID registrations are needed since the underlying key type is identical.
- The test certificates are real-world public certs: a GlobalSign TPM Root CA, STMicroelectronics intermediate/root CAs (distributed via Microsoft's TrustedTpm.cab), and an STM TPM 1.2 EK leaf certificate. None contain private or internal data.

### Testing:

One new test in `crypto/x509/x509_test.cc`:
- `RsaesOaepSPKI`: verifies `X509_get_pubkey()` succeeds and returns an RSA key from a cert with `rsaesOaep` SPKI. Uses an inline PEM constant matching the `ITUT_X509_nid_rsa` test pattern.

Test was confirmed to fail without the fix (`X509_get_pubkey()` returns null) and pass with it.


samuel40791765 (commented):
The test files are a bit overkill imo. We probably just need a brief test that contains the OID and verifies we map to our RSA implementation.

Similar to what we had in https://github.com/aws/aws-lc/commit/4916fe8641c5c220e6e6921defc026a8c564f855 should suffice. We could also inline the single leaf PEM as a string constant in `x509_test.cc`, so that `sources.cmake` doesn't need an update.

crlorentzen:
Updated to single test and inline PEM.

crypto/evp_extra/evp_asn1.c
justsmth: NP: Since you've now hoisted the result of `OBJ_cbs2nid(&oid)` into a local `nid`, we can pass `nid` through to `PQDSA_find_asn1_by_nid` and `KEM_find_asn1_by_nid` as well.

crypto/evp_extra/evp_asn1.c
```diff
@@ -46,9 +46,12 @@ static const EVP_PKEY_ASN1_METHOD *parse_key_type(CBS *cbs, CBS *out_oid) {
     }
   }
 
-  // Special logic to handle the rarer |NID_rsa|.
+  // Special logic to handle the rarer |NID_rsa| and |NID_rsaesOaep|.
+  // NID_rsa:
   // https://www.itu.int/ITU-T/formal-language/itu-t/x/x509/2008/AlgorithmObjectIdentifiers.html
-  if (OBJ_cbs2nid(&oid) == NID_rsa) {
+  // NID_rsaesOaep: underlying key is the same as |NID_rsa|.
```
justsmth: NP: You mention TCG Credential Profiles Specification in the PR description.  You could also include that in the comment here:
```
// NID_rsaesOaep: underlying key is the same as |NID_rsa|. Used by
// TPM 1.2 Endorsement Key certificates per TCG Credential Profiles
// V1.2, section 3.2.7:
// https://trustedcomputinggroup.org/wp-content/uploads/Credential_Profiles_V1.2_Level2_Revision8.pdf
```

justsmth (dismissed):

samuel40791765 (dismissed):

crypto/evp_extra/evp_asn1.c
crlorentzen: Sorry I didn't understand this comment until now, will take a pass and saving extra calls to `OBJ_cbs2nid(&oid)`

crypto/evp_extra/evp_asn1.c
crlorentzen: I'll update the comments but avoid the pdf link IMO it's better to allow someone to websearch for the information in case the pdf goes missing. I followed the same plan on the OpenSSL PR.

justsmth (approved):

samuel40791765 (approved):

PR #3183, opened 2026-04-23
Make OPENSSL_memcmp constant time

WillChilds-Klein:
## Summary

Reimplements `OPENSSL_memcmp` in `crypto/internal.h` to be **constant-time while preserving `memcmp`'s ordering semantics** (negative / zero / positive). The loop scans the full length and uses `constant_time_select_int` to capture the lowest-index differing byte pair, so it can be used safely with secret data and at sites that need ordering for sort comparators.

## Performance Benchmark Report

Benchmarks were run on the same machine (macOS / ARM64, Release build) across **three** versions, with iterations interleaved (round-robin per iteration) to even out thermal and system noise.

### Methodology

- **`main`:** PR's parent commit on `main` (libc `memcmp` everywhere) — `de73c3a51`
- **`prior`:** Tip of the PR before this revision — `OPENSSL_memcmp` delegated to `CRYPTO_memcmp` (constant-time, no ordering); ordering call sites used `OPENSSL_memcmp_ordered` → libc `memcmp`. Commit `3e81966ba`.
- **`head`:** Current PR — single constant-time-with-ordering `OPENSSL_memcmp`. Commit `c8eb92a9f`.
- **Iterations:** 5 runs per benchmark per version, fully interleaved.
- **Tool:** `bssl speed -filter <suite> -timeout 2`.

### Headline Benchmarks (5 interleaved iterations each)

| Benchmark | main | prior | head | prior vs main | head vs main | head vs prior |
|---|---:|---:|---:|---:|---:|---:|
| ECDSA P-224 signing | 55280.4 | 54997.8 | 54861.4 | -0.51% | -0.76% | -0.25% |
| ECDSA P-224 verify | 30114.1 | 30245.5 | 30145.7 | +0.44% | +0.11% | -0.33% |
| ECDSA P-256 signing | 77973.3 | 79323.6 | 79384.7 | +1.73% | +1.81% | +0.08% |
| ECDSA P-256 verify | 32061.2 | 31570.5 | 31516.7 | -1.53% | -1.70% | -0.17% |
| ECDSA P-384 signing | 18429.7 | 18171.5 | 18378.6 | -1.40% | -0.28% | +1.14% |
| ECDSA P-384 verify | 7634.3 | 7688.3 | 7805.7 | +0.71% | +2.25% | +1.53% |
| ECDSA P-521 signing | 9610.2 | 9640.6 | 9647.0 | +0.32% | +0.38% | +0.07% |
| ECDSA P-521 verify | 4567.3 | 4531.2 | 4633.8 | -0.79% | +1.46% | +2.27% |
| ECDSA secp256k1 signing | 9542.5 | 9458.8 | 9641.7 | -0.88% | +1.04% | +1.93% |
| ECDSA secp256k1 verify | 10352.3 | 10163.2 | 10358.5 | -1.83% | +0.06% | +1.92% |
| RSA 2048 verify (same key) | 97862.5 | 99069.6 | 99615.8 | +1.23% | +1.79% | +0.55% |
| RSA 4096 verify (same key) | 26549.8 | 27070.8 | 26254.4 | +1.96% | -1.11% | -3.02% |
| RSA 2048 verify (fresh key) | 83157.7 | 81100.9 | 83595.4 | -2.47% | +0.53% | +3.08% |
| RSA 4096 verify (fresh key) | 23952.1 | 24178.3 | 23977.4 | +0.94% | +0.11% | -0.83% |
| RSA 2048 signing | 2455.1 | 2464.5 | 2456.0 | +0.38% | +0.04% | -0.34% |
| RSA 3072 signing | 814.1 | 812.0 | 832.1 | -0.26% | +2.21% | +2.48% |
| RSA 4096 signing | 382.7 | 391.2 | 377.9 | +2.22% | -1.25% | -3.39% |
| RSA 8192 signing | 50.8 | 51.2 | 51.2 | +0.83% | +0.83% | +0.00% |
| Ed25519 signing | 201957.0 | 201858.3 | 201340.3 | -0.05% | -0.31% | -0.26% |
| Ed25519 verify | 43793.1 | 43355.5 | 43750.8 | -1.00% | -0.10% | +0.91% |
| ECDH X25519 | 48984.6 | 48890.1 | 49445.0 | -0.19% | +0.94% | +1.13% |

### Aggregate Summary (33 common benchmarks)

| Comparison | Mean abs diff | Median abs diff | Within ±1% | Within ±3% | Within ±5% |
|---|---:|---:|---:|---:|---:|
| **head vs main** | 0.85% | 0.69% | 20/33 | 33/33 | 33/33 |
| **head vs prior** | 1.08% | 0.83% | 19/33 | 30/33 | 33/33 |

### Interpretation

**No statistically significant performance impact.** The new constant-time-with-ordering `OPENSSL_memcmp` performs essentially identically to both the original libc `memcmp` (`main`) and the prior `CRYPTO_memcmp`-based version (`prior`):

- 100% of the 33 common benchmarks fall within ±5% of `main`, and 33/33 fall within ±3% of `main`.
- Differences are bidirectional (some up, some down) and within the run-to-run variance shown by interleaving.
- The hot RSA signing path (which calls `OPENSSL_memcmp` on 256–512 byte buffers in the verify-after-sign self-check) shows no consistent regression: RSA 2048 signing is +0.04% vs `main`, RSA 4096 signing is -1.25%, RSA 8192 signing is +0.83%.
- Per-byte work in the new loop is one XOR, one mask, and one `constant_time_select_int`. For the buffer sizes we compare (32 bytes for hashes, up to ~512 bytes for RSA keys), this remains negligible relative to the surrounding crypto operations.

The earlier revision of this PR showed a noticeable RSA 2048 signing regression in non-interleaved runs; with proper interleaving across all three versions on the same machine, that signal disappears, confirming it was thermal/scheduler noise rather than a real cost.

## Ordering Semantics

`OPENSSL_memcmp` now matches `memcmp`'s tri-state return contract:
- Negative if the first differing byte in `s1` is less than the corresponding byte in `s2`,
- Zero if all `n` bytes are equal,
- Positive if the first differing byte in `s1` is greater than the corresponding byte in `s2`.

The implementation walks the buffer from the end so that, when the same-position bytes differ, the **earliest** (lowest index) difference is the one whose signed difference is left in `result` — preserving `memcmp` ordering — while the loop always runs `n` iterations regardless of where the difference occurs.

---


crypto/internal.h
```diff
@@ -842,6 +843,15 @@ static inline void *OPENSSL_memchr(const void *s, int c, size_t n) {
...
 
 static inline int OPENSSL_memcmp(const void *s1, const void *s2, size_t n) {
+  return CRYPTO_memcmp(s1, s2, n);
+}
+
+// OPENSSL_memcmp_ordered wraps memcmp with NULL-safe behavior for n == 0.
+// Unlike |OPENSSL_memcmp|, it preserves ordering semantics (negative, zero,
+// positive return values) and is NOT constant-time. Use this only for
+// comparisons of non-secret data that require ordering.
+static inline int OPENSSL_memcmp_ordered(const void *s1, const void *s2,
```
justsmth: Since this is a `static` function that we can name whatever, why not make the name more clear, like `OPENSSL_memcmp_not_const_time` or `OPENSSL_memcmp_data_dependent_timing` or similar?

justsmth: * Is the "ordered" property of this function needed by any of the callers? Or is it just a side-effect of the implementation. 
* If needed we could implement an "ordered" memory comparison in a constant-time way as well.

WillChilds-Klein: sure, we can name it something noisier.

> Is the "ordered" property of this function needed by any of the callers? Or is it just a side-effect of the implementation.

yes. those are the callers identified as needing ordering property from memcmp

> If needed we could implement an "ordered" memory comparison in a constant-time way as well.

yeah, might as well just do that and drop the separate function.

crypto/internal.h
```diff
@@ -803,7 +803,10 @@ static inline uint64_t CRYPTO_bswap8(uint64_t x) {
 // These wrapper functions behave the same as the corresponding C standard
 // functions, but behave as expected when passed NULL if the length is zero.
 //
-// Note |OPENSSL_memcmp| is a different function from |CRYPTO_memcmp|.
+// |OPENSSL_memcmp| performs a constant-time comparison that preserves the
+// ordering semantics of |memcmp| (negative, zero, or positive). It iterates
```
dkostic, line 807: I have to ask -- do we actually have any usage within aws-lc that depends on the negative/postive distinction?

WillChilds-Klein: surprisingly, yes. both crypto and ssl tests segfault if we just use `CRYPTO_memcmp` here, which returns only 0 or 1. additionally, some crypto tests fail w/o segfaulting

```
$ git diff
diff --git a/crypto/internal.h b/crypto/internal.h
index 1f7ad6781..036ffb24f 100644
--- a/crypto/internal.h
+++ b/crypto/internal.h
@@ -845,21 +845,7 @@ static inline void *OPENSSL_memchr(const void *s, int c, size_t n) {
 #endif  // __cplusplus

 static inline int OPENSSL_memcmp(const void *s1, const void *s2, size_t n) {
-  const uint8_t *a = (const uint8_t *)s1;
-  const uint8_t *b = (const uint8_t *)s2;
-  // Walk from the end so the lowest-index differing byte (the one that decides
-  // memcmp's sign) is the last write to |result|. Equal-byte iterations leave
-  // |result| unchanged. The loop always runs |n| iterations, so timing depends
-  // only on |n|.
-  int result = 0;
-  for (size_t i = n; i > 0; i--) {
-    uint8_t ai = a[i - 1];
-    uint8_t bi = b[i - 1];
-    crypto_word_t diff_mask =
-        ~constant_time_is_zero_w((crypto_word_t)(ai ^ bi));
-    result = constant_time_select_int(diff_mask, (int)ai - (int)bi, result);
-  }
-  return result;
+    return CRYPTO_memcmp(s1, s2, n);
 }

 static inline void *OPENSSL_memcpy(void *dst, const void *src, size_t n) {


$ ./build/crypto/crypto_test | grep '\[  FAILED'
[  FAILED  ] ASN1Test.Integer (2 ms)
[  FAILED  ] ASN1Test.ParseASN1Object (0 ms)
[  FAILED  ] ASN1Test.StringCmp (0 ms)
[  FAILED  ] ASN1Test.PrintASN1Object (0 ms)
[  FAILED  ] CBBTest.FlushASN1SetOf (0 ms)
[  FAILED  ] EVPExtraTest.d2i_PrivateKey (0 ms)
[  FAILED  ] KEMTest.ParsePublicKeyInvalidLength (0 ms)
[  FAILED  ] KEMTest.ParsePrivateKeyInvalidLength (0 ms)
[  FAILED  ] KEMTest.InvalidSeedLength (0 ms)
[  FAILED  ] ObjTest.TestBasic (0 ms)
[  FAILED  ] ObjTest.TestObj2Txt (0 ms)
[  FAILED  ] OCSPTest.TestGoodOCSP (0 ms)
[  FAILED  ] OCSPTest.TestUntrustedDataOCSP (0 ms)
[  FAILED  ] OCSPTest.TestGoodOCSP_SHA256 (0 ms)
[  FAILED  ] OCSPTest.BasicAddCert (0 ms)
[  FAILED  ] OCSPTest.OCSPResponseRecreate (0 ms)
[  FAILED  ] OCSPTest.OCSPNonce (0 ms)
[  FAILED  ] OCSPTest.OCSPResponsePrint (0 ms)
[  FAILED  ] OCSPTest.OCSPRequestPrint (0 ms)
[  FAILED  ] OCSPTest.OCSPNoCheckExtension (61 ms)
[  FAILED  ] PKCS7Test.CertReparseNSS (0 ms)
[  FAILED  ] PKCS7Test.CertReparseWindows (0 ms)
[  FAILED  ] PKCS7Test.CertSignedWithSignerInfos (0 ms)
[  FAILED  ] PKCS7Test.CrlReparse (0 ms)
[  FAILED  ] PKCS7Test.SortCerts (0 ms)
[  FAILED  ] PKCS7Test.SortCertsRaw (1 ms)
[  FAILED  ] PKCS7Test.GettersSetters (0 ms)
[  FAILED  ] PKCS7Test.DataInitFinal (0 ms)
[  FAILED  ] PKCS7Test.PKCS7SignedAttributes (0 ms)
[  FAILED  ] PKCS7Test.PKCS7SignedAttributesRuby (0 ms)
[  FAILED  ] X509Test.TestVerify (0 ms)
[  FAILED  ] X509Test.PartialChain (0 ms)
[  FAILED  ] X509Test.VerifyThreads (1 ms)
[  FAILED  ] X509Test.CRLThreads (0 ms)
[  FAILED  ] X509Test.StoreThreads (0 ms)
[  FAILED  ] X509Test.ZeroLengthsWithX509PARAM (0 ms)
[  FAILED  ] X509Test.ZeroLengthsWithCheckFunctions (0 ms)
[  FAILED  ] X509Test.TestCRL (1 ms)
[  FAILED  ] X509Test.CRLDistributionPointScope (72 ms)
[  FAILED  ] X509Test.CRLSpecificIDPPreferredOverBroadCRL (40 ms)
[  FAILED  ] X509Test.ManyNamesAndConstraints (2 ms)
[  FAILED  ] X509Test.TestPSS (0 ms)
[  FAILED  ] X509Test.TestRsaSsaPss (0 ms)
[  FAILED  ] X509Test.TestPSSMD5Digest (0 ms)
[  FAILED  ] X509Test.TestPSSMD5MGF1Digest (1 ms)
[  FAILED  ] X509Test.TestEd25519 (0 ms)
[  FAILED  ] X509Test.TestEd25519BadParameters (0 ms)
[  FAILED  ] X509Test.RSASign (0 ms)
[  FAILED  ] X509Test.SignCertificate (0 ms)
[  FAILED  ] X509Test.SignCRL (1 ms)
[  FAILED  ] X509Test.SignCSR (0 ms)
[  FAILED  ] X509Test.PqdsaCSR (0 ms)
[  FAILED  ] X509Test.Ed25519Sign (0 ms)
[  FAILED  ] X509Test.TestMLDSA65 (0 ms)
[  FAILED  ] X509Test.TestBadSigAlgMLDSA65 (0 ms)
[  FAILED  ] X509Test.TestBadParamsMLDSA65 (0 ms)
[  FAILED  ] X509Test.NoBasicConstraintsCertSign (0 ms)
[  FAILED  ] X509Test.NoBasicConstraintsNetscapeCA (0 ms)
[  FAILED  ] X509Test.CommonNameFallback (0 ms)
[  FAILED  ] X509Test.ServerGatedCryptoEKUs (0 ms)
[  FAILED  ] X509Test.InvalidExtensions (1 ms)
[  FAILED  ] X509Test.BasicConstraints (0 ms)
[  FAILED  ] X509Test.AlgorithmParameters (0 ms)
[  FAILED  ] X509Test.SignatureVerification (1 ms)
[  FAILED  ] X509Test.NamePrint (5 ms)
```

justsmth (approved):

crypto/internal.h
```diff
@@ -803,7 +803,10 @@ static inline uint64_t CRYPTO_bswap8(uint64_t x) {
 // These wrapper functions behave the same as the corresponding C standard
 // functions, but behave as expected when passed NULL if the length is zero.
 //
-// Note |OPENSSL_memcmp| is a different function from |CRYPTO_memcmp|.
+// |OPENSSL_memcmp| performs a constant-time comparison that preserves the
```
sgmenda, line 806: nit: could we add a short comment here to avoid confusion with CRYPTO_memcmp https://github.com/aws/aws-lc/blob/53695e3ae0dfaa8363d9028758486af030cf72d3/include/openssl/mem.h#L58-L63

sgmenda (approved):

crypto/internal.h
WillChilds-Klein: `OPENSSL_memcmp` is a strictly internal function, so I'm not sure we want to mention it in our public headers.

PR #3186, opened 2026-04-23
chore: simplify the api docs workflow

dougch:
### Issues:
Resolves: none

### Description of changes:
The `doc.go` tool requires being run from the `util/` directory with explicit `--config` and `--out` flags. This makes the GitHub Pages workflow brittle and the tool unusable from the repo root.

as a reminder, the final doc URL: https://aws.github.io/aws-lc/api/headers.html

Changes:
- `doc.go` auto-locates its config file via `runtime.Caller(0)` — no more CWD dependency
- Add `OutputDirectory` field to `doc.config` so output path lives in config, not the CLI
- Resolve `BaseDirectory` and `OutputDirectory` relative to config file location
- Auto-create output directory
- Fix `doc.css` copy path to be config-relative

### Call-outs:

- This workflow only triggers publish on push to main, but does a validation of comments on PRs.
- `workflow_dispatch` can be used to test manually after merge
- Backwards compatible: `--config` and `--out` flags still work as overrides

### Testing:
Run locally: `go run util/doc.go` from repo root — generates docs in `pages-output/`, with the main entrypoint `pages-output/headers.html`.


util/doc.go
```diff
@@ -938,6 +946,31 @@ func main() {
...
+	if !filepath.IsAbs(config.BaseDirectory) {
+		config.BaseDirectory = filepath.Join(configDir, config.BaseDirectory)
+	}
+
+	// Use --out if provided, otherwise fall back to config, resolved
+	// relative to the config file's location.
+	if len(*outputDir) == 0 {
+		*outputDir = config.OutputDirectory
+		if !filepath.IsAbs(*outputDir) {
+			*outputDir = filepath.Join(configDir, *outputDir)
```
justsmth, line 966: If `OutputDirectory` is omitted in the config and `--out` is not provided, this resolution path turns `""` into `configDir` via `filepath.Join`, so the check below never fires and the tool writes generated files next to `doc.config` instead of failing. Can we validate that `config.OutputDirectory` is non-empty before resolving it relative to `configDir`?

util/doc.go
justsmth: This says `--config` is required and HTML is written to the current directory. Since this PR changes both defaults, can we update that comment too so the source matches the new behavior?

util/doc.go
dougch: Good catch; updated.

justsmth (dismissed):
Nit: Add `pages-output/` to `.gitignore`?

util/doc.go
```diff
@@ -938,6 +947,31 @@ func main() {
...
+			fmt.Printf("No output directory given by --out or OutputDirectory in config\n")
+			os.Exit(1)
+		}
+		*outputDir = config.OutputDirectory
+		if !filepath.IsAbs(*outputDir) {
+			*outputDir = filepath.Join(configDir, *outputDir)
+		}
+	}
+
+	if err := os.MkdirAll(*outputDir, 0777); err != nil {
```
WillChilds-Klein: do we need to give perms beyond user?

WillChilds-Klein (commented):
> This workflow only triggers on push to main, not on PRs — cannot be validated in CI before merge

what if a malformed comment in a PR causes this to fail? do we have some alarm to know main is broken? can this be caught before merge with a per-PR CI job?

dougch:
> > This workflow only triggers on push to main, not on PRs — cannot be validated in CI before merge
> 
> what if a malformed comment in a PR causes this to fail? do we have some alarm to know main is broken? can this be caught before merge with a per-PR CI job?

Good call; separated out the check from the publish with an if condition.

util/doc.go
```diff
@@ -908,23 +913,27 @@ func copyFile(outPath string, inFilePath string) error {
...
 
 	if len(*configFlag) == 0 {
-		fmt.Printf("No config file given by --config\n")
-		os.Exit(1)
-	}
-
-	if len(*outputDir) == 0 {
-		fmt.Printf("No output directory given by --out\n")
-		os.Exit(1)
+		*configFlag = filepath.Join(sourceDir(), "doc.config")
```
WillChilds-Klein, line 936: should we check that this file actually exists and error out if not?

WillChilds-Klein (dismissed):

.github/workflows/api-docs.yml
```diff
@@ -20,7 +27,19 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
+  validate-api-docs:
```
justsmth, line 30: The `pages` concurrency group is shared, and this sets `cancel-in-progress: false`. So PRs touching headers will queue behind each other, and even behind a push-to-main publish run, but the job never touches `pages`. 

Why not give it its own group?
```
  validate-api-docs:
    concurrency:
      group: api-docs-validate-${{ github.ref }}
      cancel-in-progress: true
    runs-on: ubuntu-latest
```

.github/workflows/api-docs.yml
```diff
@@ -20,7 +27,19 @@ concurrency:
...
+    runs-on: ubuntu-latest
+    steps:
```
justsmth, line 37: This now runs on PRs and inherits `pages: write` and `id-token: write` from the workflow-level permissions -- but it never uses them. We should lock it down here:
```
    runs-on: ubuntu-latest
    permissions:
      contents: read
```

util/doc.go
justsmth: Tested it -- the `os.ReadFile` below already errors clearly if the config's missing:
```
Failed to open config file: open /nonexistent/doc.config: no such file or directory
```

I think that covers it. Feel free to resolve unless you want a more specific message for the auto-detected case.

util/doc.go
```diff
@@ -949,7 +983,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := copyFile(*outputDir, "doc.css"); err != nil {
+	if err := copyFile(*outputDir, filepath.Join(filepath.Dir(*configFlag), "doc.css")); err != nil {
```
justsmth: NP: `configDir` already has this (line 952), no need to recompute it.

```suggestion
	if err := copyFile(*outputDir, filepath.Join(configDir, "doc.css")); err != nil {
```

justsmth (dismissed):

justsmth (dismissed):

WillChilds-Klein (approved):

justsmth (approved):

PR #3190, opened 2026-04-27
ci: harden zig wrappers against libc++ ABI drift and opaque failures

justsmth:
### Description of changes:

Three small, independent improvements to the zig CI wrappers and workflow. None of them fix a current outage; all three prevent a class of problem we saw during a recent test using Zig 0.16.0 experiment.

1. **Darwin libc++ ABI fix in `util/zig-c++`.** Our toolchain compiles with `zig c++` but links with `/usr/bin/c++` (MachO-linker workaround). That split requires the libc++ ABI to match. Force `_LIBCPP_HAS_VENDOR_AVAILABILITY_ANNOTATIONS=1` on Darwin so zig's libc++ inlines helpers the system dylib doesn't yet export, instead of emitting references to them. No-op on zig 0.15.2, prevents a linker failure on future zig versions whose libc++ runs ahead of the runner's.

2. **Real exit-code reporting in both wrappers.** The existing error path reported `status 0` on actual zig failures (including SIGSEGV) because of a shell-scripting bug. Fix the capture, merge stderr, and log the invocation on failure.

3. **Self-test workflow step.** Progressively probes `zig version → zig env → -c → link → full CMake-style link line` so future regressions surface with a clean reproducer rather than nested inside CMake's TryCompile error envelope.

### Call-outs:

The Darwin change is a no-op against the currently-pinned zig 0.15.2 — it's deliberately landing now so the next zig bump doesn't eat another CI outage. Commit message has the gory details on the availability-macro mechanism if you want them.

### Testing:

Existing CI matrix. Locally verified the libc++ fix with a `std::unordered_set<std::string>` reproducer (undefined symbol disappears, binary links and runs), and the exit-code fix with a fake zig that SIGSEGVs (wrapper now reports 139 instead of 0).


torben-hansen (approved):

util/zig-cc
```diff
@@ -38,4 +38,17 @@ for arg in "$@"; do
...
+# Merge stderr into stdout and print the invocation on failure so CI logs
+# always surface zig's diagnostic even when something in the cmd.exe ->
+# bash -> zig chain drops stderr (observed on windows-11-arm previously).
+#
+# Capture rc via `|| rc=$?` rather than `if ! ...; then rc=$?` because the
+# latter makes $? reflect the negated if-test result (0/1) instead of zig's
+# real exit status (e.g. 139 for SIGSEGV).
+rc=0
+zig cc -fno-sanitize=undefined -fno-sanitize=address -fno-sanitize=memory "${filtered_args[@]}" 2>&1 || rc=$?
+if [[ "${rc}" -ne 0 ]]; then
```
WillChilds-Klein, line 50: trivial nit: here and elsewhere, `rc` is guaranteed to be non-zero by `||` -- you can just check `$?` here

.github/workflows/zig.yml
```diff
@@ -55,6 +55,45 @@ jobs:
...
+          # here rather than being nested inside CMake's TryCompile harness:
+          #   1) zig runs at all on this host?          (`zig version`)
+          #   2) zig can report its environment?        (`zig env`)
+          #   3) zig can compile without linking?       (`-c hello.c`)
+          #   4) zig can compile+link trivially?        (`hello.c -o hello`)
+          #   5) zig can handle CMake's Windows link line? (full TryCompile args)
+          # Each step runs unconditionally (no `set -e`) so we see results from
+          # every stage even when an earlier one fails. The wrapper reports
+          # zig's real exit status (including 139 for SIGSEGV) on failure.
+          set -x
```
WillChilds-Klein, line 71: should we `set -e` so failed probes fail the CI job?

or perhaps conditionally exit a failed probe on 76 so we print the exit code?

.github/workflows/zig.yml
```diff
@@ -55,6 +55,45 @@ jobs:
...
+          probe() {
+            echo "=== probe: $* ==="
+            "$@"
+            echo "=== probe exit: $? ==="
+          }
+
+          probe zig version
+          probe zig env
+
+          echo 'int main(void){return 0;}' > hello.c
```
WillChilds-Klein, line 81: not really relevant to your  change, but do we assert that zigc links AWS-LC (instead of e.g. system OpenSSL) into executables? such a check has been helpful for python/ruby.

WillChilds-Klein (dismissed):

.github/workflows/zig.yml
WillChilds-Klein: @justsmth clarified offline that this step is never meant to fail -- it's just printing diagnostic information.

WillChilds-Klein (approved):

PR #2986, opened 2026-02-11
Import ML-DSA x86_64 NTT/INTT assembly optimizations

jakemas:
### Issues:
Resolves #ISSUE-NUMBER1
Addresses #ISSUE-NUMBER2

### Description of changes: 
- Does a fresh import of mldsa-native from upstream
- Modified importer.sh to import only ntt.S, intt.S, nttunpack.S from x86_64 backend, as well as zeta tables
- Rewrites meta.h to only enable NTT/INTT functions, so we can continue to add one at a time
- Processes assembly files with s2n-bignum macros, follows same methodology as mlkem-native native import

### Call-outs:
Point out areas that need special attention or support during the review process. Discuss architecture or design changes.

### Testing:
Functions have been formally verified within s2n-bignum at:
- https://github.com/awslabs/s2n-bignum/commit/b8670b221f9860e21b441ca44c4f782c64b8ee41
- https://github.com/awslabs/s2n-bignum/commit/002fdb0174a0e1b4f82911e01566d02c70284d58
- https://github.com/awslabs/s2n-bignum/pull/348

### Benchmarks

- CPU: Intel Xeon Platinum 8175M @ 2.50GHz (8 cores, 16 threads)
- OS: Amazon Linux 2 (kernel 5.10.247)
- Compiler: GCC 7.3.1
- Build: Release mode (`cmake3 -GNinja -DCMAKE_BUILD_TYPE=Release`)

```
Throughput ops/sec from bssl speed:
┌──────────────────────┬──────────┬──────────┬─────────────┐
│ Operation            │  Before  │  After   │ Improvement │
├──────────────────────┼──────────┼──────────┼─────────────┤
│ MLDSA44 keygen       │  10,229  │  10,343  │    +1.1%    │
│ MLDSA44 signing      │   2,429  │   3,955  │   +62.8%    │
│ MLDSA44 verify       │   9,676  │  11,573  │   +19.6%    │
│ MLDSA65 keygen       │   5,569  │   5,618  │    +0.9%    │
│ MLDSA65 signing      │   1,591  │   2,520  │   +58.3%    │
│ MLDSA65 verify       │   5,819  │   6,903  │   +18.6%    │
│ MLDSA87 keygen       │   3,754  │   3,781  │    +0.7%    │
│ MLDSA87 signing      │   1,281  │   1,914  │   +49.4%    │
│ MLDSA87 verify       │   3,629  │   3,988  │    +9.9%    │
└──────────────────────┴──────────┴──────────┴─────────────┘

[...]

PR #2631, opened 2025-08-20
ML-KEM: import and enable x86_64 backend from mlkem-native

dkostic:
### Issues:
N/A

### Description of changes: 
Import and enable x86_64 backend from mlkem-native.

Several small changes were needed to do this:
 - `importer.sh` script was extended to handle the x86_64 backend,
 - `mlkem_native_backend.h` was modified to include the x86_64
    `meta.h` file,
 - `mlkem_native_config.h` was updated with a custom CPU capability
   function that checks if AVX2 is available.

Significant performance improvement across all ML-KEM operations.
For example, benchmark (ops/s) on Intel(R) Xeon(R) Platinum 8488C:
```
┌─────────────────────┬──────────┬──────────┬─────────────┐
│ Operation           │  Before  │  After   │ Improvement │
├─────────────────────┼──────────┼──────────┼─────────────┤
│ ML-KEM-512 keygen   │  60,568  │  97,450  │    +60.9%   │
│ ML-KEM-512 encaps   │  54,471  │  86,422  │    +58.7%   │
│ ML-KEM-512 decaps   │  45,414  │  71,586  │    +57.6%   │
│ ML-KEM-768 keygen   │  35,840  │  61,248  │    +70.9%   │
│ ML-KEM-768 encaps   │  35,326  │  54,329  │    +53.8%   │
│ ML-KEM-768 decaps   │  30,022  │  45,912  │    +52.9%   │
│ ML-KEM-1024 keygen  │  23,727  │  41,377  │    +74.4%   │
│ ML-KEM-1024 encaps  │  24,516  │  37,087  │    +51.3%   │
│ ML-KEM-1024 decaps  │  20,957  │  31,768  │    +51.6%   │
└─────────────────────┴──────────┴──────────┴─────────────┘
```
### Call-outs:
Point out areas that need special attention or support during the review process. Discuss architecture or design changes.

### Testing:
How is this change tested (unit tests, fuzz tests, etc.)? Are there any testing steps to be verified by the reviewer?

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and the ISC license.

PR #3195, opened 2026-04-27
ML-DSA: import and enable x86_64 assembly backend from mldsa-native

jakemas:
## Summary
- Imports the x86_64 native arithmetic backend from mldsa-native into ML-DSA, providing AVX2-accelerated assembly for core polynomial operations
- Only pure assembly (`.S`) files with completed HOL-Light formal verification proofs are imported; C intrinsics and unverified assembly are excluded
- Follows the same integration pattern as ML-KEM's x86_64 backend (PR #2631 / commit 3b1e95e4), using s2n-bignum macros for symbol visibility

## Benchmark

Measured on Intel Xeon Platinum 8175M @ 2.50GHz (EC2 c5.2xlarge), average of 4 runs (ops/sec, higher is better). "Before" is `bssl speed -filter MLDSA` built from `main`; "After" is the same with this PR applied:

```
┌──────────────────────┬──────────┬──────────┬─────────────┐
│ Operation            │  Before  │  After   │   Speedup   │
├──────────────────────┼──────────┼──────────┼─────────────┤
│ MLDSA44 keygen       │  14,472  │  18,151  │    1.25x    │
│ MLDSA44 signing      │   2,343  │   4,794  │    2.05x    │
│ MLDSA44 verify       │  11,537  │  19,064  │    1.65x    │
│ MLDSA65 keygen       │   6,607  │   8,599  │    1.30x    │
│ MLDSA65 signing      │   1,489  │   3,265  │    2.19x    │
│ MLDSA65 verify       │   8,302  │  10,894  │    1.31x    │
│ MLDSA87 keygen       │   5,599  │   6,932  │    1.24x    │
│ MLDSA87 signing      │   1,310  │   2,862  │    2.18x    │
│ MLDSA87 verify       │   5,705  │   7,723  │    1.35x    │
└──────────────────────┴──────────┴──────────┴─────────────┘
```

## Changes

**Commit 1 — Manual/preparatory changes:**
- `CMakeLists.txt` — adds ML-DSA x86_64 assembly sources to BCM build via `file(GLOB ... CONFIGURE_DEPENDS "*.S")`
- `mldsa_native_config.h` — enables native backend with `MLD_CONFIG_USE_NATIVE_BACKEND_ARITH`
- `mldsa_native_backend.h` — platform dispatcher (x86_64 + AVX2 only for now)
- `mldsa_x86_64_meta.h` — hand-maintained backend header outside the imported `mldsa/` tree, declaring only the assembly-backed operations we use
- `importer.sh` — imports x86_64 backend and processes assembly with s2n-bignum macros

**Commit 2 — Clean import output:**
- Solely the output of `GITHUB_SHA=08d40f9403a9ca80f160118bc63e96bf36627866 ./importer.sh --force`
- Contains no manual changes — reproducible by running the above command

## Functions accelerated
All imported functions have completed HOL-Light formal verification proofs:
- NTT (forward) — `mldsa_ntt.ml`
- INTT (inverse) — `mldsa_intt.ml`
- NTT unpack (custom coefficient order) — `mldsa_nttunpack.ml`
- Pointwise Montgomery multiplication — `mldsa_pointwise.ml`
- Polyvec pointwise accumulate for L=4/5/7 — `mldsa_pointwise_acc_l{4,5,7}.ml`
- Conditional add Q (caddq) — `mldsa_caddq.ml`

## Importer simplifications (addressing review feedback)
The importer is now significantly simpler compared to the previous revision:
- **Removed** `MLD_INTERNAL_API` sed — upstream now uses `MLD_INTERNAL_DATA_DECLARATION/DEFINITION`
- **Removed** ELF section move — upstream already places it correctly
- **Removed** file renaming — upstream now uses unique `_avx2_asm.S` names (no collision with mlkem)
- **Glob copy** for all `.S` files — no need to enumerate each assembly file individually
- **Single sed pattern** to strip C-intrinsic includes from BCM (keeps only `consts.c`)
- **Simple substitution** for `MLD_ASM_FN_SIZE` → `S2N_BN_SIZE_DIRECTIVE` (upstream now defines it)

## Call-outs
- Supersedes #2986 which only imported NTT/INTT — this PR imports all formally verified assembly-backed operations
- C intrinsics (rej_uniform, decompose, use_hint, chknorm, polyz_unpack) are intentionally excluded
- Runtime dispatch via `CRYPTO_is_AVX2_capable()` — falls back to C on non-AVX2 systems
- AArch64 native backend is not included in this PR

## Testing
- All ML-DSA tests pass (KAT, Wycheproof, expanded key validation, PQDSA parameter tests)


jakemas:
**Note on `aws-lc-rs` CI failure:** The `aws-lc-rs-linux` CI failure is expected. The `aws-lc-rs` CC builder scripts (`aws-lc-sys/scripts/cc_builder/linux_x86_64.sh`, `apple_x86_64.sh`) currently only discover ML-KEM native assembly via `find crypto/fipsmodule/ml_kem/mlkem/native/x86_64/src -name "*.S"`. A follow-up PR to `aws-lc-rs` will be needed to add the equivalent line for ML-DSA:

```bash
mapfile -O ${#SOURCE_FILES[@]} -t SOURCE_FILES < <(find crypto/fipsmodule/ml_dsa/mldsa/native/x86_64/src -name "*.S" -type f | sort -f)
```

This is the same pattern as when ML-KEM's x86_64 backend was first added.

WillChilds-Klein (commented):
Is the second commit solely the result of running the import script? i.e. does it contain **any** manual changes?

geedo0 (dismissed):

jakemas:
Good question. The previous structure had manual edits mixed into the import commit. I've restructured into two clean commits:

1. **`c085723e` — Manual/preparatory changes**: `CMakeLists.txt`, `mldsa_native_config.h`, `mldsa_native_backend.h`, and `importer.sh` (extended to import x86_64 backend, process assembly with s2n-bignum macros, strip C-intrinsic operations, and rename files with `mldsa_` prefix).

2. **`3f03ce6f` — Clean import output**: Solely the result of running `GITHUB_SHA=b61e84f0c73d4ed612ffcaea4282a9d682de3f46 ./importer.sh --force`. Contains no manual changes — anyone can reproduce the same output by running the script.

The importer now handles stripping C-intrinsic content from `meta.h` and `arith_native_x86_64.h`, adding `MLD_INTERNAL_API` to `consts.c`/`consts.h`, removing AArch64 references from the BCM file, and appending `S2N_BN_SIZE_DIRECTIVE` to assembly files (upstream mldsa-native doesn't define `MLD_ASM_FN_SIZE` unlike mlkem-native's `MLK_ASM_FN_SIZE`).

crypto/fipsmodule/CMakeLists.txt
```diff
@@ -397,6 +397,28 @@ if((ARCH STREQUAL "x86_64") AND UNIX AND NOT MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
...
 
+# mldsa-native assembly files can be compiled on Unix platforms for x86_64 only.
+if((ARCH STREQUAL "x86_64") AND UNIX AND NOT MY_ASSEMBLER_IS_TOO_OLD_FOR_AVX)
+
+  set(MLDSA_NATIVE_DIR "${AWSLC_SOURCE_DIR}/crypto/fipsmodule/ml_dsa")
+
+  set(MLDSA_NATIVE_X86_64_ASM_SOURCES
+
+    ${MLDSA_NATIVE_DIR}/mldsa/native/x86_64/src/mldsa_intt.S
+    ${MLDSA_NATIVE_DIR}/mldsa/native/x86_64/src/mldsa_ntt.S
```
WillChilds-Klein: nit: if we're including all files in this directory, you might want to just list the files at build time to compose this list

https://cmake.org/cmake/help/latest/command/file.html

WillChilds-Klein (dismissed):
I reviewed the importer script changes in de9597, but not any of the auto-generate changes in the second commit.

script changes LGTM, but might want to have @dkostic or someone else familiar with mlkem-native's import script have a look.

crypto/fipsmodule/CMakeLists.txt
jakemas: good call, done in c085723e - switched to `file(GLOB ... CONFIGURE_DEPENDS "*.S")` so refreshes that add/remove files don't need a matching edit here

WillChilds-Klein (dismissed):

crypto/fipsmodule/ml_dsa/importer.sh
```diff
@@ -72,31 +72,129 @@ popd
...
+echo "Strip C-intrinsic content from arith_native_x86_64.h"
+ARITH_H=$SRC/native/x86_64/src/arith_native_x86_64.h
+# Remove #include <stdint.h>
+sed "${SED_I[@]}" '/#include <stdint.h>/d' "$ARITH_H"
+# Remove everything between #include "consts.h" and #define mld_ntt_avx2
+# (buffer length macros, comments, rej_uniform_table extern),
+# but preserve one blank line separator
+sed "${SED_I[@]}" '/^#include "consts.h"$/,/^#define mld_ntt_avx2/{/^#include "consts.h"$/b; /^#define mld_ntt_avx2/b; d;}' "$ARITH_H"
+# Replace __contract__ annotated mld_ntt_avx2 declaration with plain declaration
+sed "${SED_I[@]}" '/^void mld_ntt_avx2/,/^);$/c\void mld_ntt_avx2(int32_t *r, const int32_t *qdata);' "$ARITH_H"
```
hanno-becker: I'm not convinced of this approach, it seems error prone and difficult to understand. If think it's cleaner to have a custom `meta.h` that outside of the mldsa-native source tree that defines what is being used.

hanno-becker (changes requested):
The changes to the `importer.sh` accommodating the removal of AVX2 intrinsics seem too complicated. Are they really needed? Compare @dkostic's original PR for mlkem-native https://github.com/aws/aws-lc/pull/2631, specifically https://github.com/aws/aws-lc/pull/2631/changes#diff-4d3f13a41a2db9217769beee4b9f8bec4c390a02856cbb033ec4930d158a5b99R117.

* Is there something we can do upstream so all you need to do here is, analogously to the mlkem-native case, just #undef' the `USE_NATIVE_XXX` lines? If so, we should do that.
* If the modifications required to the `meta.h` are inevitably more complex than for mlkem-native, then the restricted `meta.h` should just be maintained in `ml_kem/*` outside of the mldsa-native copy. It is expected that consuming libraries may need to customize and maintain their own backend headers.

crypto/fipsmodule/ml_dsa/importer.sh
jakemas: Ok replaced the sed-stripping with a custom `meta.h` maintained outside the mldsa-native source tree (see `crypto/fipsmodule/ml_dsa/mldsa_x86_64_meta.h` in commit 40d50eb3). The importer no longer touches `meta.h` or `arith_native_x86_64.h` at all.

jakemas:
@hanno-becker thanks for the pointer to #2631. Went with your second suggestion: a custom `meta.h` (`crypto/fipsmodule/ml_dsa/mldsa_x86_64_meta.h`) maintained outside the imported `mldsa/` tree, declaring only the assembly-backed ops we actually use (NTT, INTT, nttunpack, pointwise, polyvecl_pointwise_acc).

What this buys us:
* `importer.sh` no longer modifies `meta.h` or `arith_native_x86_64.h` -- both the upstream meta.h is simply not copied, and `arith_native_x86_64.h` is imported verbatim (unused C-intrinsic declarations are harmless).
* The only sed in the importer now is BCM `#include` pruning for the C-intrinsic `.c` files we don't compile, `MLD_INTERNAL_API` on `mld_qdata`, and the existing s2n-bignum macro fixup for `.S` files.
* Commit 2 (`d968d42b`) is now purely the output of `./importer.sh --force` with no manual touch-ups.

The upstream-flag route would be cleaner long-term but requires an mldsa-native PR first; happy to pursue if you'd like, given that the C-intrinsic/assembly split in mldsa-native is less tidy than mlkem-native's.

All ML-DSA tests pass on x86_64 Linux.

crypto/fipsmodule/ml_dsa/importer.sh
```diff
@@ -110,6 +157,71 @@ cp $TMP/mldsa/mldsa_native.h $SRC
...
+  s2n_header="_internal_s2n_bignum_x86_att.h"
+  sed "${SED_I[@]}" "s/#include \"\.\.\/\.\.\/\.\.\/common\.h\"/#include \"$s2n_header\"/" "$file"
+
+  func_name=$(grep -o '\.global MLD_ASM_NAMESPACE(\([^)]*\))' "$file" | sed 's/\.global MLD_ASM_NAMESPACE(\([^)]*\))/\1/')
+  if [ -n "$func_name" ]; then
+    sed "${SED_I[@]}" "s/\.global MLD_ASM_NAMESPACE($func_name)/        S2N_BN_SYM_VISIBILITY_DIRECTIVE(mldsa_$func_name)\n        S2N_BN_SYM_PRIVACY_DIRECTIVE(mldsa_$func_name)/" "$file"
+    sed "${SED_I[@]}" "s/MLD_ASM_FN_SYMBOL($func_name)/S2N_BN_SYMBOL(mldsa_$func_name):/" "$file"
+
+    # Append S2N_BN_SIZE_DIRECTIVE after .cfi_endproc. Upstream mldsa-native
+    # does not define MLD_ASM_FN_SIZE (unlike mlkem-native's MLK_ASM_FN_SIZE),
```
hanno-becker: Is this something we should fix upstream first instead of patching the importer?

crypto/fipsmodule/ml_dsa/importer.sh
```diff
@@ -72,31 +72,78 @@ popd
...
 # Copy formatting file
 cp $TMP/.clang-format $SRC
 
+if [[ "$(uname)" == "Darwin" ]]; then
+  SED_I=(-i "")
+else
+  SED_I=(-i)
+fi
+
+# Add MLD_INTERNAL_API to the mld_qdata declaration/definition so the
```
hanno-becker: Can you check if this is still needed? Those are now decorated with `MLD_INTERNAL_DATA_DECLARATION/DEFINITION` in mldsa-native precisely to allow giving them internal linkage.

crypto/fipsmodule/ml_dsa/importer.sh
```diff
@@ -72,31 +72,78 @@ popd
 
 echo "Pull source code from remote repository..."
 
-# Copy mldsa-native source tree -- C source only (no native backends for now)
+# Copy mldsa-native source tree -- C source
 mkdir $SRC
-cp $TMP/mldsa/src/* $SRC
+# Copy only files (not subdirectories like native/ and fips202/)
+find $TMP/mldsa/src -maxdepth 1 -type f -exec cp {} $SRC \;
```
hanno-becker, line 78: For my benefit, what's the practical difference between the before/after here?

crypto/fipsmodule/ml_dsa/importer.sh
```diff
@@ -72,31 +72,78 @@ popd
...
+# Shared constants (zetas table); needed by the assembly kernels
+cp $TMP/mldsa/src/native/x86_64/src/consts.h $SRC/native/x86_64/src
+cp $TMP/mldsa/src/native/x86_64/src/consts.c $SRC/native/x86_64/src
+# Assembly source files (the only backend sources we actually compile)
+cp $TMP/mldsa/src/native/x86_64/src/*.S $SRC/native/x86_64/src
+
+# Rename assembly files with mldsa_ prefix to avoid basename collisions
+# with ML-KEM assembly files (e.g., ntt.S exists in both). This matters
+# because s2n-tls's libcrypto interning uses `ar x` which extracts into
+# a flat directory and would overwrite files with the same basename.
```
hanno-becker: This is an interesting note. Could you please bring this up on mldsa-native so we can discuss whether to add the explicit `mldsa_` prefix to the ASM files?

Would the same issue exist with source files, but somehow doesn't because we use the BCM build?

crypto/fipsmodule/ml_dsa/importer.sh
```diff
@@ -72,31 +72,78 @@ popd
...
+# ================================================================
+
 # Copy and statically simplify BCM file
 # The static simplification is not necessary, but improves readability
 # by removing directives related to the FIPS-202 backend that we provide
 # via our own glue layer.
+# Also strip AArch64 backend since we only import x86_64.
 unifdef -DMLD_CONFIG_FIPS202_CUSTOM_HEADER                             \
         -UMLD_CONFIG_USE_NATIVE_BACKEND_FIPS202                        \
+        -UMLD_SYS_AARCH64                                              \
```
hanno-becker, line 114: Hm, I don't think this is right -- this is a derived option, not a configuration option. We should not be meddling with the system detection macros.

crypto/fipsmodule/ml_dsa/importer.sh
```diff
@@ -110,6 +157,71 @@ cp $TMP/mldsa/mldsa_native.h $SRC
...
+S2N_BN_SIZE_DIRECTIVE(mldsa_$func_name)" "$file"
+  fi
+
+  # Move ELF section before .text (match s2n-bignum convention)
+  sed "${SED_I[@]}" '/#if defined(__ELF__)/,/#endif/d' "$file"
+  sed "${SED_I[@]}" '/^\.text$/i\
+#if defined(__ELF__)\
+.section .note.GNU-stack,"",@progbits\
+#endif\
+' "$file"
```
hanno-becker: This is not needed, I think

crypto/fipsmodule/ml_dsa/importer.sh
```diff
@@ -110,6 +157,71 @@ cp $TMP/mldsa/mldsa_native.h $SRC
...
+
+  # Clean up: strip leading whitespace from lines left indented after unifdef
+  sed "${SED_I[@]}" 's/^ \(\/\*\)/\1/' "$file"
+  sed "${SED_I[@]}" 's/^ \(#include\)/#include/' "$file"
+  # Remove consecutive blank lines
+  sed "${SED_I[@]}" '/^$/N;/^\n$/d' "$file"
```
hanno-becker: Do we _need_ this? Should we apply a formatter instead of `sed`?

crypto/fipsmodule/ml_dsa/importer.sh
hanno-becker: Hold on, mldsa-native _does_ have `MLD_ASM_FN_SIZE` and all the ASM files use it

crypto/fipsmodule/ml_dsa/importer.sh
hanno-becker: Hmm ok, I see we do this for RISCV64 as well in mlkem-native.

hanno-becker (changes requested):
Thanks @jakemas for the changes!

The importer.sh can be simplified further, I believe -- see comments.

crypto/fipsmodule/ml_dsa/importer.sh
jakemas: Done — updated to mldsa-native HEAD (1b47ba6) which now uses `MLD_INTERNAL_DATA_DECLARATION/DEFINITION`. The sed is removed from the importer; `consts.h`/`consts.c` are imported verbatim and the internal linkage is handled by our existing `MLD_CONFIG_INTERNAL_API_QUALIFIER` in `mldsa_native_config.h`.

crypto/fipsmodule/ml_dsa/importer.sh
jakemas: The old `cp $TMP/mldsa/src/* $SRC` would also copy subdirectories (native/, fips202/) via shell glob expansion. Switching to `find -maxdepth 1 -type f` ensures only top-level source files are copied, while the x86_64 backend files are selectively copied below.

crypto/fipsmodule/ml_dsa/importer.sh
jakemas: Good point. Upstream files are now named `ntt_avx2_asm.S`, `intt_avx2_asm.S`, etc. which no longer collide with mlkem-native's `ntt.S`, `intt.S`. Removed the renaming from the importer. Will bring up the prefix discussion on mldsa-native for any future collision concerns (e.g., if mlkem-native also adopts `_avx2_asm` naming).

crypto/fipsmodule/ml_dsa/importer.sh
jakemas: Removed `-UMLD_SYS_AARCH64` from the unifdef call. The AArch64 includes are guarded by `#if defined(MLD_SYS_AARCH64)` which won't be true on x86_64, so they're harmless dead code in the BCM. (Note: the mlkem-native importer does the same with `-UMLK_SYS_RISCV64` — might be worth removing there too for consistency.)

crypto/fipsmodule/ml_dsa/importer.sh
jakemas: Fixed upstream mldsa-native HEAD (1b47ba6) — now defines `MLD_ASM_FN_SIZE`. The importer now does a simple substitution (`MLD_ASM_FN_SIZE(func) -> S2N_BN_SIZE_DIRECTIVE(mldsa_func)`) matching the mlkem-native pattern. The append logic is removed.

crypto/fipsmodule/ml_dsa/importer.sh
jakemas: Confirmed — upstream now places the `#if defined(__ELF__)` section at the end of each .S file. Removed the ELF section move from the importer.

jakemas:
Addressed all review feedback. Updated to mldsa-native HEAD (`1b47ba6`) and significantly simplified the importer:

**Changes to `importer.sh`:**
- Removed `MLD_INTERNAL_API` sed — upstream now uses `MLD_INTERNAL_DATA_DECLARATION/DEFINITION` which works with our `MLD_CONFIG_INTERNAL_API_QUALIFIER`
- Removed `-UMLD_SYS_AARCH64` from unifdef — system-derived macro, not a config option
- Replaced `S2N_BN_SIZE_DIRECTIVE` append with `MLD_ASM_FN_SIZE` substitution — upstream now defines this macro
- Removed ELF section move — upstream already places it correctly
- Removed file renaming — upstream now uses unique `_avx2_asm.S` names (no collision with mlkem's `ntt.S`)
- Selectively copy only assembly files with verified proofs (excludes `poly_caddq_avx2_asm.S`)

**Changes to `mldsa_x86_64_meta.h`:**
- Updated function names for new upstream API (`_asm` suffix)
- Updated `mld_poly_pointwise_montgomery_native` to new destructive-in-first-arg signature
- Added `MLD_MUST_CHECK_RETURN_VALUE` annotations

All 82 ML-DSA tests pass on x86_64 Linux.

crypto/fipsmodule/ml_dsa/importer.sh
jakemas: Removed — those cosmetic sed cleanups (leading whitespace stripping, blank line removal) are no longer in the importer. The upstream assembly files are now clean after unifdef processing without needing post-fixup.

crypto/fipsmodule/ml_dsa/importer.sh
hanno-becker: No, I don't think `cp` copies subdirectories -- only `cp -r` would do.

crypto/fipsmodule/ml_dsa/importer.sh
```diff
@@ -72,16 +72,56 @@ popd
...
 
 # Copy formatting file
 cp $TMP/.clang-format $SRC
 
+if [[ "$(uname)" == "Darwin" ]]; then
+  SED_I=(-i "")
+else
+  SED_I=(-i)
+fi
+
```
hanno-becker: Unnecessary code movement

crypto/fipsmodule/ml_dsa/importer.sh
jakemas: You're right — `cp` without `-r` won't copy subdirectories, it'll just error/skip them. The `find -maxdepth 1 -type f` avoids the error that `cp src/*` would produce when it encounters subdirectories (native/, fips202/) in the glob expansion. Updated the comment to reflect this.

crypto/fipsmodule/ml_dsa/importer.sh
hanno-becker: I was actually wrong here. This only concerns the BCM file, where the unifdef avoids us refrencing files we don't actually import. Yes, it's harmless, but still confusing. The importer now does the file removal via sed, which is arguably worse. So, apologies for the change of mind, but I think the original approach was better.

crypto/fipsmodule/ml_dsa/importer.sh
jakemas: Fixed — moved `SED_I` back to its original position (after the unifdef/BCM copy).

crypto/fipsmodule/ml_dsa/importer.sh
jakemas: No worries, reverted — put `-UMLD_SYS_AARCH64` back in the unifdef and dropped the sed approach.

jakemas:
@hanno-becker anything outstanding here?

crypto/fipsmodule/ml_dsa/mldsa/.clang-format
```diff
@@ -26,4 +26,4 @@ Macros:
  - __contract__(x)={ void a; void b; void c; void d; void e; void f; } void abcdefghijklmnopqrstuvw()
  - __loop__(x)={} do
  # Make this artifically long to force line break
- - MLK_INTERNAL_API=void abcdefghijklmnopqrstuvwabcdefghijklmnopqrstuvwabcdefg();
+ - MLD_INTERNAL_API=void abcdefghijklmnopqrstuvwabcdefghijklmnopqrstuvwabcdefg();
```
WillChilds-Klein, line 29: what's the significance of this change?

WillChilds-Klein (dismissed):

crypto/fipsmodule/ml_dsa/mldsa/.clang-format
jakemas: Just a typo fix from upstream (ML-KEM uses MLK, ML-DSA uses MLD). when we first imported mldsa-native at [b61e84f0](https://github.com/pq-code-package/mldsa-native/commit/b61e84f0) it still had `MLK_INTERNAL_API` in the .clang-format (leftover from when mldsa-native was forked off mlkem-native). they fixed it in https://github.com/pq-code-package/mldsa-native/commit/d4554bee, and the re-import here picks that up. no manual change on our side.

crypto/fipsmodule/ml_dsa/importer.sh
```diff
@@ -72,22 +72,57 @@ popd
...
+# intentionally excluded.
+#
+# The upstream meta.h advertises both assembly and C-intrinsic operations.
+# Rather than modify it, we keep a hand-maintained replacement in
+# ../mldsa_x86_64_meta.h (referenced via MLD_CONFIG_ARITH_BACKEND_FILE) that
+# declares only the assembly-backed subset. Upstream meta.h is not copied.
+mkdir -p $SRC/native/x86_64/src
+# Backend API and specification assumed by mldsa-native frontend
+cp $TMP/mldsa/src/native/api.h $SRC/native
+# Backend header -- unused C-intrinsic declarations are harmless and left intact
```
hanno-becker: I've seen old compilers (specifically GCC 4.8) complain about unresolved external function references even in `static inline` functions that were never used. I believe this could still happen here if we don't remove the inline definitions for the intrinsics backend. But your CI does not seem to flag it, so up to you if this risk is tolerable.

crypto/fipsmodule/ml_dsa/importer.sh
```diff
@@ -110,6 +145,51 @@ cp $TMP/mldsa/mldsa_native.h $SRC
...
+  s2n_header="_internal_s2n_bignum_x86_att.h"
+  sed "${SED_I[@]}" "s/#include \"\.\.\/\.\.\/\.\.\/common\.h\"/#include \"$s2n_header\"/" "$file"
+
+  func_name=$(grep -o '\.global MLD_ASM_NAMESPACE(\([^)]*\))' "$file" | sed 's/\.global MLD_ASM_NAMESPACE(\([^)]*\))/\1/')
+  if [ -n "$func_name" ]; then
+    sed "${SED_I[@]}" "s/\.global MLD_ASM_NAMESPACE($func_name)/        S2N_BN_SYM_VISIBILITY_DIRECTIVE(mldsa_$func_name)\n        S2N_BN_SYM_PRIVACY_DIRECTIVE(mldsa_$func_name)/" "$file"
+    sed "${SED_I[@]}" "s/MLD_ASM_FN_SYMBOL($func_name)/S2N_BN_SYMBOL(mldsa_$func_name):/" "$file"
+    sed "${SED_I[@]}" "s/MLD_ASM_FN_SIZE($func_name)/S2N_BN_SIZE_DIRECTIVE(mldsa_$func_name)/" "$file"
+  fi
+done
```
hanno-becker, line 169: We should avoid this complexity if we can. I'll do some experiments on the x-native side to see if we can get rid of it.

hanno-becker: It's more difficult than I thought. Let's stick with this for now.

crypto/fipsmodule/ml_dsa/importer.sh
```diff
@@ -110,6 +145,51 @@ cp $TMP/mldsa/mldsa_native.h $SRC
 echo "Fixup include paths"
 sed "${SED_I[@]}" 's/#include "src\/\([^"]*\)"/#include "\1"/' $SRC/mldsa_native_bcm.c
 
+# Drop #include directives for the C-intrinsic .c files we did not import.
+# Only consts.c (shared with the assembly backend) needs to be compiled.
+echo "Strip C-intrinsic includes from mldsa_native_bcm.c"
+BCM=$SRC/mldsa_native_bcm.c
+sed "${SED_I[@]}" '/^#include "native\/x86_64\/src\/poly_caddq_avx2\.c"/d'      "$BCM"
```
hanno-becker: This file does not exist anymore upstream I think.

hanno-becker (commented):
I have reviewed the importer and it looks mostly good, except for one sed applying to a file that doesn't exist.

I'm not sure though if this is the right time to import: We're excluding a single *.S file from the backend (caddq) for which we don't yet have the proof upstream, and we're importing a non-tagged commit.

I'd prefer for the PR to wait until caddq is proved upstream as well, and we have a tagged release.

jakemas:
> I have reviewed the importer and it looks mostly good, except for one sed applying to a file that doesn't exist.
> 
> I'm not sure though if this is the right time to import: We're excluding a single *.S file from the backend (caddq) for which we don't yet have the proof upstream, and we're importing a non-tagged commit.
> 
> I'd prefer for the PR to wait until caddq is proved upstream as well, and we have a tagged release.

I agree, unfortunately, we are actively in the process of converting the AVX2 from c to S, there is another function nearing completing in PR now. I opened https://github.com/pq-code-package/mldsa-native/pull/1068 to try to make the import logic even cleaner, if we can merge that upstream, version mldsa-native, not merge any more AVX2 conversions while we import, then yes, I am good with waiting to include cadddq.

crypto/fipsmodule/ml_dsa/importer.sh
jakemas: Correct, it's now `poly_caddq_avx2_asm.S` (assembly with a verified proof). Removed this sed line and imported the assembly file directly in 3f72ba07.

crypto/fipsmodule/ml_dsa/importer.sh
```diff
@@ -72,22 +72,46 @@ popd
...
+# The upstream meta.h advertises both assembly and C-intrinsic operations.
+# Rather than modify it, we keep a hand-maintained replacement in
+# ../mldsa_x86_64_meta.h (referenced via MLD_CONFIG_ARITH_BACKEND_FILE) that
+# declares only the assembly-backed subset. Upstream meta.h is not copied.
+mkdir -p $SRC/native/x86_64/src
+cp $TMP/mldsa/src/native/api.h $SRC/native
+cp $TMP/mldsa/src/native/x86_64/src/arith_native_x86_64.h $SRC/native/x86_64/src
+cp $TMP/mldsa/src/native/x86_64/src/consts.h $SRC/native/x86_64/src
+cp $TMP/mldsa/src/native/x86_64/src/consts.c $SRC/native/x86_64/src
+# NOTE: all imported .S files must have verified proofs in s2n-bignum.
```
jakemas, line 95: If new .S functions are converted upstream without a hol-light proof at the same time, this import script will pick them up. This could be awkward if we run imports before the proofs land for new functions. We could add a manual check, but its clunky. anything we add will be temporary, as over time we will add all .S functions, as all will have proofs. 

Upstream I'm writing the proof at the same time as the conversion, so this  won't be an issue for upcoming uniform_rej for example.

crypto/fipsmodule/ml_dsa/META.yml
```diff
@@ -1,5 +1,5 @@
 name: mldsa-native
 source: pq-code-package/mldsa-native.git
-branch: main
-commit: b61e84f0c73d4ed612ffcaea4282a9d682de3f46
-imported-at: 2026-01-16T13:12:01-0800
+branch: 1b47ba602b3220fb06380840fd516dde4243122e
+commit: 1b47ba602b3220fb06380840fd516dde4243122e
+imported-at: 2026-05-14T03:52:22+0000
```
WillChilds-Klein: are we merging this without a more recent import?

WillChilds-Klein (dismissed):

crypto/fipsmodule/ml_dsa/META.yml
jakemas: Thanks, I was avoiding the churn of making you review a new version this close to your approval, but, this message tells me you're up for it! Done, updated to mldsa-native 08d40f9

WillChilds-Klein (approved):

geedo0 (approved):

PR #3202, opened 2026-04-28
Reject URIs containing '@' in name constraint checking

nhatnghiho:
### Issues:
Addresses P422947832

### Description of changes: 
 `nc_uri()` extracts the host from a URI by scanning for `:` (port) then `/` (path start). It does not account for the `@` delimiter that separates  optional userinfo from the host in RFC 3986 §3.2 (authority = `[userinfo "@"] host [":" port]`). Since `:` is permitted inside userinfo, a URI like `spiffe://x.team-a.corp:x@team-b.corp/admin` causes nc_uri() to extract x`.team-a.corp` as the host instead of the actual host `team-b.corp`                                                                                                                  
                                                                                                                                                  
This change makes three improvements to nc_uri():                                                                                               
1. Reject `@` in authority — URIs containing userinfo are rejected with `X509_V_ERR_UNSUPPORTED_NAME_SYNTAX`. Certificate SAN URIs have no standards-defined use for userinfo, so rejecting is the correct fail-closed behavior — consistent with the IPv6 literal rejection added in `fed51c342`. The @ scan stops at /, ?, or # (the characters that terminate the authority per RFC 3986), so @ appearing in the path (e.g., `foo://example.com/@user`) is not affected.                                                                                                       
2. Trailing dot normalization — Strip trailing dots from both the host and constraint before comparison, so `team-a.corp.` and `team-a.corp` are treated as the same FQDN (RFC 1034 §3.1). This matches the existing normalization in nc_dns().                                                  
3. FQDN validation — Validate that the extracted host contains only characters valid in a DNS name (`a-zA-Z0-9-.`) per RFC 1034 §3.5. RFC 5280 §4.2.1.10 requires the host to be a fully qualified domain name, so any host containing percent-encoding, `?`, `#`, or other non-FQDN characters is rejected. This prevents equivalence bypasses (e.g., `b%61d.com` evading an exclusion for .bad.com).

### Call-outs:
This intentionally rejects rather than normalizes percent-encoded hosts. RFC 5280 requires the host to be a FQDN, and `%` is not a valid FQDN  character — a conformant CA should never issue such a cert. Rejecting is stricter than the RFC 3986 §6.2.2.2 SHOULD-level normalization guidance, but avoids being more permissive than the standard.     

### Testing:
Nine new cases in `X509Test.NameConstraints`:
- Basic `user@host` rejection
- Colon-in-userinfo bypass (`spiffe://x.team-a.corp:x@team-b.corp/admin`) tested against both the fake and real host constraints
- Userinfo with path, query, and fragment variants
- `@` in path (after `/`) is correctly not rejected
  
All 102 existing X509 tests pass.


manastasova (dismissed):

justsmth (dismissed):

crypto/x509/v3_ncons.c
```diff
@@ -698,6 +713,35 @@ static int nc_uri(const ASN1_IA5STRING *uri, const ASN1_IA5STRING *base) {
...
+  }
+  if (ends_with_byte(&base_cbs, '.')) {
+    uint8_t unused;
+    CBS_get_last_u8(&base_cbs, &unused);
+  }
+
+  // RFC 5280 §4.2.1.10 requires the host to be a fully qualified domain name.
+  // Validate that the host contains only characters valid in a DNS name
+  // (RFC 1034 §3.5): letters, digits, hyphens, and dots. This rejects
+  // percent-encoded hosts, IPv4 literals, and any other non-FQDN syntax,
```
justsmth: NP: This doesn't reject "IPv4 literals": `192.168.0.1` would not be rejected by this.

justsmth (dismissed):

torben-hansen (approved):

justsmth (approved):

PR #3204, opened 2026-05-01
Ensure correct bio memory buffer type is assigned

torben-hansen:
### Issues:

V2186717843

### Description of changes: 

Stash correct type in `mem_ctrl()`. A `BUF_MEM` needs to be wrapped by `BIO_BUF_MEM`. 

### DIFF
--- a/crypto/bio/bio_mem.c
+++ b/crypto/bio/bio_mem.c
@@ -261,11 +261,25 @@ static long mem_ctrl(BIO *bio, int cmd, long num, void *ptr) {
         *pptr = (b->data != NULL) ? &b->data[bbm->read_off] : NULL;
       }
       break;
-    case BIO_C_SET_BUF_MEM:
+    case BIO_C_SET_BUF_MEM: {
+      // |bio->ptr| stores a |BIO_BUF_MEM| wrapper, not a raw |BUF_MEM*|.
+      // Allocate the wrapper before releasing the old state so a failure here
+      // leaves |bio| usable and does not take ownership of |ptr|.
+      BIO_BUF_MEM *new_bbm = OPENSSL_zalloc(sizeof(*new_bbm));
+      if (new_bbm == NULL) {
+        ret = 0;
+        break;
+      }
+      new_bbm->buf = ptr;
+      new_bbm->read_off = 0;
       mem_free(bio);
       bio->shutdown = (int)num;
-      bio->ptr = ptr;
+      bio->init = 1;
+      bio->ptr = new_bbm;
+      bio->num = -1;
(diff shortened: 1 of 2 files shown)


crypto/bio/bio_mem.c
```diff
@@ -261,11 +261,25 @@ static long mem_ctrl(BIO *bio, int cmd, long num, void *ptr) {
         *pptr = (b->data != NULL) ? &b->data[bbm->read_off] : NULL;
       }
       break;
-    case BIO_C_SET_BUF_MEM:
+    case BIO_C_SET_BUF_MEM: {
+      // |bio->ptr| stores a |BIO_BUF_MEM| wrapper, not a raw |BUF_MEM*|.
+      // Allocate the wrapper before releasing the old state so a failure here
+      // leaves |bio| usable and does not take ownership of |ptr|.
+      BIO_BUF_MEM *new_bbm = OPENSSL_zalloc(sizeof(*new_bbm));
```
WillChilds-Klein, line 268: does this allocation change the API contract? how would existing callers know to free this?

torben-hansen: I believe that should be handled by `mem_free()`, which actually expects this behaviour, not the prior behaviour if I'm not mistaken.

WillChilds-Klein: `mem_free` is called on `bio` before `new_bmm` is assigned to `bio` on L278.

or do you mean the call to `mem_free` if/when the caller eventually calls `BIO_free` later on?

WillChilds-Klein (approved):

crypto/bio/bio_mem.c
WillChilds-Klein: chatted offline. agreed that caller in my hypothetical is misusing API; `BIO_free` should always be called on a heap-allocated BIO and that is the caller's responsibility.

crypto/bio/bio_test.cc
```diff
@@ -447,6 +447,63 @@ TEST(BIOTest, BioGetMemLeak) {
...
+
+  ASSERT_TRUE(BIO_set_mem_buf(bio.get(), m, BIO_CLOSE));
+
+  char out[8] = {0};
+  EXPECT_EQ(5, BIO_read(bio.get(), out, sizeof(out)));
+  EXPECT_EQ(Bytes(out, 5), Bytes("world", 5));
+
+  // Exercise to surface memory violations. 
+  bio.reset();
+}
```
samuel40791765, line 489: Nit: could we consolidate these two tests? Seems to just be a slight logic difference depending on whether `BIO_NOCLOSE` or `BIO_CLOSE` was set.

samuel40791765 (approved):

PR #3209, opened 2026-05-04
tool-openssl/s_client: default SNI to -connect host to match OpenSSL

alexw91:
### Description of changes:

aws-lc's `openssl s_client` tool does not send the SNI extension unless the user explicitly passes `-servername`. OpenSSL's `s_client` has defaulted SNI to the `-connect` hostname since OpenSSL 1.1, with a DNS-name check that suppresses SNI for IP literals (RFC 6066 §3 forbids SNI for IPs), and exposes `-noservername` as an explicit opt-out.

This PR brings aws-lc's `is_openssl_s_client` code path in line with that behavior:

- Default SNI to the hostname parsed from `-connect` when neither `-servername` / `-server-name` nor `-noservername` was provided.
- Skip the default for IP literals and single-label hostnames via a verbatim port of OpenSSL's `is_dNS_name` helper.
- Add `-noservername` as an opt-out flag, with the same mutual-exclusion error message as OpenSSL when combined with `-servername`.

The `bssl client` tool (`is_openssl_s_client == false`) is unaffected.

### Call-outs:

The `is_dNS_name` helper in `tool/client.cc` is copied verbatim from OpenSSL 3.x `apps/s_client.c` (commit `653f437c2a`, file lines 3847–3910) and carries an Apache-2.0 attribution header matching the style used elsewhere in aws-lc (e.g. `tool-openssl/txt_db/txt_db.cc`, `generated-src/*`). Please flag any aws-lc conventions I should match more closely for this kind of verbatim third-party copy.

**Behavior change visible to CLI users:**

| Command | Before | After |
|---|---|---|
| `./openssl s_client -connect foo.com:443` | no SNI sent | sends SNI `foo.com` |
| `./openssl s_client -connect foo.com:443 -servername bar.com` | sends SNI `bar.com` | unchanged |
| `./openssl s_client -connect foo.com:443 -noservername` | unknown argument | suppresses SNI (new flag) |
| `./openssl s_client -connect 1.2.3.4:443` | no SNI sent | no SNI sent (IP literal, matches OpenSSL) |
| `./openssl s_client -connect [::1]:443` | no SNI sent | no SNI sent (IPv6 literal, matches OpenSSL) |
| `./openssl s_client -connect localhost:443` | no SNI sent | no SNI sent (single-label, matches OpenSSL) |

---


tool/client.cc
```diff
@@ -26,6 +26,95 @@ OPENSSL_MSVC_PRAGMA(warning(pop))
...
+/*
+ * Host dNS Name verifier: used for checking that the hostname is in dNS format
+ * before setting it as SNI
+ */
+static int is_dNS_name(const char *host)
+{
+    const size_t MAX_LABEL_LENGTH = 63;
+    size_t i = 0;
+    int isdnsname = 0;
+    size_t length = strlen(host);
```
WillChilds-Klein, line 41: unless we need to keep this port verbatim, suggest `strnlen(host, MAX_LABEL_LENGTH)` and moving declaration of `i` inside the `for` statement

WillChilds-Klein (dismissed):

tool/client.cc
alexw91: `strnlen(host, 63)` would cap the measured length at 63 characters, but 63 is the per-label limit for DNS names, not the per-hostname limit — real hostnames can be much longer (e.g. `subdomain.reasonably-long-second-label-that-pushes-us-past-the-limit.com`). Swapping in `strnlen` would silently reject those as invalid. The existing `strlen` is safe because the only caller passes a `std::string`, which is guaranteed to be properly terminated.

WillChilds-Klein (approved):

samuel40791765 (approved):

justsmth (approved):

PR #3211, opened 2026-05-04
Decouple the FIPS version number from the AWS-LC version number

samuel40791765:
### Description of changes:
Historically, the FIPS version number has been tied to AWS-LC's own
version number — consumers inferred which FIPS submission a build
corresponded to from the AWS-LC version. This change decouples the two
so the FIPS version number can evolve independently on a separate
cadence from AWS-LC releases.

Adds `FIPS_version`, a new public API declared in `openssl/crypto.h`,
that returns the FIPS version number of the current build. Returns 0
when AWS-LC is not built in FIPS mode. `AWSLC_FIPS_VERSION_NUMBER` in
`openssl/base.h` is the single source of truth — bump it there when
cutting a new FIPS branch. `FIPS_version` mirrors `FIPS_mode`'s gating
so the two agree on what counts as a FIPS build, including under
Address and Memory sanitizers.

### DIFF
--- a/crypto/fipsmodule/self_check/fips.c
+++ b/crypto/fipsmodule/self_check/fips.c
@@ -26,4 +26,12 @@ int FIPS_is_entropy_cpu_jitter(void) {
   return 1;
 }
 
+uint32_t FIPS_version(void) {
+#if defined(BORINGSSL_FIPS) && !defined(OPENSSL_ASAN) && !defined(OPENSSL_MSAN)
+  return AWSLC_FIPS_VERSION_NUMBER;
+#else
+  return 0;
+#endif
+}
+
 int FIPS_mode_set(int on) { return on == FIPS_mode(); }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ endif()
 
 set(SOFTWARE_NAME "awslc")
 set(SOFTWARE_VERSION "5.0.0")
+set(AWSLC_FIPS_VERSION 4)
 set(ABI_VERSION 0)
 set(CRYPTO_LIB_NAME "crypto")
 set(SSL_LIB_NAME "ssl")
--- a/crypto/crypto_test.cc
+++ b/crypto/crypto_test.cc
@@ -69,6 +69,15 @@ TEST(CryptoTest, aws_lc_assert_entropy_cpu_jitter) {
   }
 }
 
+// FIPS_version returns the FIPS version number in FIPS builds and 0 otherwise.
+TEST(CryptoTest, FIPSVersion) {
+  if (FIPS_mode() == 1) {
+    EXPECT_EQ(FIPS_version(), (uint32_t)AWSLC_FIPS_VERSION_NUMBER);
+  } else {
+    EXPECT_EQ(FIPS_version(), 0u);
+  }
+}
+
 TEST(CryptoTest, OPENSSL_hexstr2buf) {
   const char *test_cases[][2] = {{"a2", "\xa2"},
                                  {"a213", "\xa2\x13"},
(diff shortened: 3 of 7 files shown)


crypto/fipsmodule/self_check/fips.c
```diff
@@ -26,4 +26,12 @@ int FIPS_is_entropy_cpu_jitter(void) {
   return 1;
 }
 
+uint32_t FIPS_version(void) {
+#if defined(BORINGSSL_FIPS) && !defined(OPENSSL_ASAN) && !defined(OPENSSL_MSAN)
+  return AWSLC_FIPS_VERSION_NUMBER;
```
justsmth, line 31: Have we considered encoding this with room for minor/patch granularity from the start? Especially for LTS FIPS branch that will hang around many years, it might receive backported features, CVE patches and bug fixes. Consumers calling this API (who tend to be compliance-oriented) will likely want to answer "does my FIPS module include the fix for CVE-X?" That's a patch-level question.

Since the return type is already `uint32_t`, we could pack major/minor/patch without any API change — e.g. `major * 10000 + minor * 100 + patch`, so FIPS v4 starts at `40000`. Comparisons like `FIPS_version() >= 40001` still work intuitively, and you could provide decomposition macros:

```
#define AWSLC_FIPS_VERSION_MAJOR(v) ((v) / 10000)
#define AWSLC_FIPS_VERSION_MINOR(v) (((v) / 100) % 100)
#define AWSLC_FIPS_VERSION_PATCH(v) ((v) % 100)
```

If we ship this as a bare `4` now and later realize LTS consumers need the granularity, bumping from `4` to `40000` would be a semantic break for anyone already comparing against `4`. Cheaper to bake the encoding in from day one.

samuel40791765: The FIPS version number intentionally has no minor/patch granularity — it's just the submission identifier, one integer per FIPS submission. That's the whole point  of decoupling it from the AWS-LC version number: the FIPS version number stays stable while the AWS-LC version number evolves with patches, bug fixes, and backports.

Patch-level questions ("does my build include the fix for CVE-X?") are answered by the AWS-LC version number, which is already semver and is queryable at runtime via `awslc_version_string()`. On an LTS branch, a CVE fix bumps the AWS-LC patch version; the FIPS version number stays put.

Encoding semver semantics into FIPS_version would re-couple the two identifiers, which is exactly what this change is trying to avoid.

justsmth: Makes sense -- I agree `FIPS_version()` should stay as a pure submission identifier and doesn't need granularity. My concern isn't with this PR then, it's with making sure the versioning scheme in #3212 explicitly supports the claim that "patch-level questions are answered by the AWS-LC version number."

For that to hold without collisions, the following invariant needs to be true: 
* **mainline only ever bumps minor** (releases are always `X.Y.0`), so that the patch space (`X.Y.Z` where `Z > 0`) is exclusively available to FIPS branches when they receive security fixes. 
* Otherwise, if mainline issues a `5.6.1` and a non-LTS FIPS snapshot was cut at `5.6.0`, you get two different releases with the same AWS-LC version string -- which breaks the "just check the AWS-LC version" answer.

I think `VERSIONING.md` in #3212 needs to make this explicit. Right now the Patch definition says "bug fixes that do not change public API or ABI" without scoping it to branches, and the non-LTS snapshot section doesn't specify what version a patched snapshot gets. 

If we nail that down, then I think the model works well.

justsmth (dismissed):

justsmth (commented):
Approved assuming we can resolve the concerns raised with clarifications in the versioning policy in #3212.

crypto/fipsmodule/self_check/fips.c
WillChilds-Klein: > mainline only ever bumps minor (releases are always X.Y.0), so that the patch space (X.Y.Z where Z > 0) is exclusively available to FIPS branches when they receive security fixes.

i think this makes sense. FIPS branches are essentially just snapshots-with-patches while mainline is active development. mainline owning minor version increments and FIPS branches owning patch version increments aligns with that.

Also agreed that it's time for us to make our versioning semantics (they're not semver!) explicit and codify them in VERSIONING.md

WillChilds-Klein (dismissed):

justsmth (approved):

WillChilds-Klein (approved):

PR #3212, opened 2026-05-04
Document new versioning scheme and bump mainline to v5.0.0

samuel40791765:
### Issues:                                                                                                                                                         
 Resolves P425126007                                                                                                                                                 
                                                                                                                                                                      
### Description of changes:                                                                                                                                         
AWS-LC does not currently document its versioning scheme or release                                                                                                 
types for external consumers. This change adds `VERSIONING.md` at the                                                                                             
repo root describing the AWS-LC version number scheme, the FIPS version                                                                                             
number, the distinction between rolling mainline and LTS releases, the                                                                                              
LTS release policy, non-LTS FIPS branches, branch naming conventions,                                                                                               
and deprecation timelines for legacy FIPS branches.                                                                                                                 
                                                                                                                                                                    
Also bumps mainline from v1.72.1 to v5.0.0 to reflect that the 4.x line                                                                                             
now belongs to the first LTS branch (`fips-2025-09-12-lts`) under the                                                                                             
new versioning scheme.                                                                                                                                              
                                                                                                                                                                                          
                                                                                                                                                                    

VERSIONING.md
```diff
@@ -0,0 +1,134 @@
...
+
+Each LTS branch is supported for 5 years from the date the branch is cut. End of support (EOS) means security fixes and other changes are no longer backported to the branch. EOS applies regardless of the status of any FIPS certificate associated with the branch; once an LTS reaches EOS, consumers should migrate to mainline or to a newer LTS.
+
+At any given time, at least one LTS branch is within its support window. Consecutive LTS branches overlap so that consumers always have a supported migration target.
+
+## Non-LTS FIPS branches
+
+FIPS validation requires a fixed snapshot of the cryptographic module's source code. Each time AWS-LC is submitted for FIPS validation, a branch is cut from mainline that preserves the exact code submitted. Most of these branches are not designated as LTS.
+
+Non-LTS FIPS branches exist solely to preserve the validated snapshot. They are not published as GitHub tagged releases, and consumers should not depend on them.
```
dougch:
```suggestion
Non-LTS FIPS branches exist solely to preserve the validated snapshot. They do not receive release tags, and consumers should not depend on them.
```
bit of a nit; I read this as "the branch is a secret", but the branch names are laid out in the next section.

dougch (dismissed):

VERSIONING.md
samuel40791765: I think your suggestion reads better.

dougch (dismissed):

VERSIONING.md
```diff
@@ -0,0 +1,134 @@
+# AWS-LC Versioning
```
justsmth, line 1: See the discussion [here](https://github.com/aws/aws-lc/pull/3211#discussion_r3242829172). 

I think what's missing in the policy is a mechanism that ensures AWS-LC "mainline" version numbers are distinct from version numbers on security-patched non-LTS FIPS branches (hopefully a rare occurrence) -- the non-LTS branch would need to be able to bump its version number in a way that doesn't conflict with a version number already present on the main branch.

One possible mechanism is simply to have the main branch avoid ever bumping the "patch number", only bump minor versions on main. Reserve patch-version bumps for the FIPS branches to use for security-patch releases.

samuel40791765: Pushed this commit as an update according to our internal discussions: https://github.com/aws/aws-lc/pull/3212/changes/af58f4f3497d9f174a554e7738a66d9f6845060f.

LTS FIPS branches receive minor and patch (same strategy as before). Non-LTS FIPS branches only receive patch number bumps. Patch number bumps are used to indicate potential security patches for both types of FIPS branches.

VERSIONING.md
```diff
@@ -0,0 +1,133 @@
...
+## Overview
+
+AWS-LC offers two release types:
+
+* **Rolling mainline releases** are the primary release for most consumers. Rolling mainline receives the latest features, performance improvements, and security fixes as they are developed.
+* **Long-term support (LTS) releases** are intended for consumers that require a stable ABI over a multi-year support window.
+
+Rolling mainline is periodically submitted for FIPS 140-3 validation, though changes may land between submissions. LTS branches are fixed FIPS submissions; they are not modified after they are cut, except for the backports permitted by the [LTS release policy](#lts-release-policy). The FIPS version number a given build corresponds to is tracked independently of the AWS-LC version number (see [FIPS version number](#fips-version-number)).
+
+AWS-LC is committed to providing a stable public API across both release types. LTS releases additionally guarantee ABI stability for the duration of the support window.
```
justsmth, line 14: * Should we point consumers to our [`AWSLC_API_VERSION`](https://github.com/aws/aws-lc/blob/main/include/openssl/base.h#L60) macro or [`ABI_VERSION`](https://github.com/aws/aws-lc/blob/main/CMakeLists.txt#L9) as better indicators of compatibility breaking changes than the version number? Should this policy also cover our use of these values? (Should `ABI_VERSION` be exposed in our header files as well?)
* I think we should state explicitly that our versioning is NOT SemVer:
```
Mainline does not follow SemVer with respect to major-version bumps; ABI may
change between minor releases on mainline. Consumers requiring ABI stability
should pin to an LTS branch.
```

VERSIONING.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Support window
+
+Each LTS branch is supported for 5 years from the date the branch is cut. End of support (EOS) means security fixes and other changes are no longer backported to the branch. EOS applies regardless of the status of any FIPS certificate associated with the branch; once an LTS reaches EOS, consumers should migrate to mainline or to a newer LTS.
+
+At any given time, at least one LTS branch is within its support window. Consecutive LTS branches overlap so that consumers always have a supported migration target.
+
+## Non-LTS FIPS branches
+
+FIPS validation requires a fixed snapshot of the cryptographic module's source code. Each time AWS-LC is submitted for FIPS validation, a branch is cut from mainline that preserves the exact code submitted. Most of these branches are not designated as LTS.
```
justsmth: The "Most of these branches are not designated as LTS" introduces some ambiguity. Instead, should we clarify that a branch is marked "LTS", or not, at time of creation -- a branch is never "promoted"?

```
An LTS branch is always cut fresh from mainline at the time of designation;
existing non-LTS branches are never promoted to LTS.
```

VERSIONING.md
```diff
@@ -0,0 +1,133 @@
...
+For example, when mainline is at `4.13.0` and an LTS branch is cut, the branch takes ownership of the `4.x` line. From that point on, the `4.x` line receives only the changes permitted below, as patch (`4.13.1`, `4.13.2`, ...) or minor (`4.14.0`, `4.15.0`, ...) increments. Mainline advances to `5.0.0` and continues normal development with minor increments only (`5.1.0`, `5.2.0`, ...). Because mainline never returns to the `4.x` line, version numbers on the LTS cannot collide with mainline. When the next LTS is cut approximately two years later, it takes ownership of whatever major version line mainline is on at that moment, and mainline bumps again.
+
+### Permitted changes on LTS branches
+
+The following are permitted on LTS branches:
+
+* Security fixes for CVEs and critical vulnerabilities. These may alter the FIPS module integrity hash when necessary.
+* Bug fixes that do not alter public API behavior, ABI compatibility, or the FIPS module integrity hash.
+* Platform compatibility fixes for supported operating environments that do not alter the FIPS module integrity hash.
+* Additive changes that preserve existing function signatures.
```
justsmth: It seems to be implied, but could be made explicit. I think the intent is:

- Feature Additive (signature-preserving) changes --> minor bump
- Security / Bug / Platform (possibly not signature-preserving) fixes --> patch bump

VERSIONING.md
```diff
@@ -0,0 +1,133 @@
...
+
+At any given time, at least one LTS branch is within its support window. Consecutive LTS branches overlap so that consumers always have a supported migration target.
+
+## Non-LTS FIPS branches
+
+FIPS validation requires a fixed snapshot of the cryptographic module's source code. Each time AWS-LC is submitted for FIPS validation, a branch is cut from mainline that preserves the exact code submitted. Most of these branches are not designated as LTS.
+
+Non-LTS FIPS branches exist solely to preserve the validated snapshot. They do not receive release tags, and consumers should not depend on them.
+
+We may apply critical security fixes to a non-LTS FIPS branch while it is the most recently NIST-certified FIPS branch. This is a maintenance concession, not a supported consumption model. Once a newer FIPS branch receives certification, the previous non-LTS branch is frozen and receives no further updates.
```
justsmth: I think we could spell this out even more:
```
We may apply critical security fixes to a non-LTS FIPS branch only while 
it is in review and until it is the most recently NIST-certified FIPS branch.
This is a maintenance concession; these branches are NOT supported for
consumption. Once a newer FIPS branch (whether LTS or non-LTS) receives NIST
certification, the previous non-LTS branch is frozen and will receive no
further updates.
```

VERSIONING.md
```diff
@@ -0,0 +1,133 @@
...
+
+A non-LTS FIPS branch inherits its version from mainline at cut time and only ever issues patch-level increments (e.g., a branch cut at `5.6.0` becomes `5.6.1` after a security fix). Because mainline only produces minor increments (`5.6.0` → `5.7.0`), patch versions on a non-LTS branch cannot collide with mainline.
```
justsmth, line 118: We could add another sentence at the end of this paragraph:
```
Note that at the branch point, both mainline and the non-LTS branch share the
same AWS-LC version string; they are distinguished by `FIPS_version()` and
`AWSLC_FIPS_VERSION_NUMBER`.
```

VERSIONING.md
justsmth, line 14: NP: I don't know that we could do anything about it -- but what if CVE fix fundamentally requires us to change a function signature or add a field to a stack-allocatable struct? We'd be force to choose between patching the CVE or not breaking ABI?

`¯\_(ツ)_/¯`

VERSIONING.md
```diff
@@ -0,0 +1,133 @@
...
+This document describes how AWS-LC is versioned and released, and how consumers should choose between the release types we offer.
+
+## Overview
+
+AWS-LC offers two release types:
+
+* **Rolling mainline releases** are the primary release for most consumers. Rolling mainline receives the latest features, performance improvements, and security fixes as they are developed.
+* **Long-term support (LTS) releases** are intended for consumers that require a stable ABI over a multi-year support window.
+
+Rolling mainline is periodically submitted for FIPS 140-3 validation, though changes may land between submissions. LTS branches are fixed FIPS submissions; they are not modified after they are cut, except for the backports permitted by the [LTS release policy](#lts-release-policy). The FIPS version number a given build corresponds to is tracked independently of the AWS-LC version number (see [FIPS version number](#fips-version-number)).
```
justsmth, line 12: NP: Should there be an API or Macro (or both) to indicate at build or runtime whether this version of AWS-LC is an LTS?

Of course, at build or runtime a consumer can compare `AWSLC_FIPS_VERSION_NUMBER` with `AWSLC_VERSION_NUMBER_STRING` -- if they align, then we're on an LTS. Should we call this out?  Or provide a function that does this?

VERSIONING.md
```diff
@@ -0,0 +1,133 @@
...
+LTS release characteristics:
+
+* Cut from mainline every 2 years and supported for 5 years from the date the branch is cut.
+* Each LTS branch is submitted for FIPS validation.
+
+## LTS release policy
+
+### LTS version evolution
+
+Each LTS branch inherits mainline's major version at the time it is cut. Mainline then immediately bumps to the next major version. This guarantees that mainline and every LTS branch have distinct major version numbers.
```
justsmth: NP: Instead of "immediately", should we say: "In the next commit, mainline must bump to the next major version."

VERSIONING.md
```diff
@@ -0,0 +1,133 @@
...
+
+Each LTS branch is supported for 5 years from the date the branch is cut. End of support (EOS) means security fixes and other changes are no longer backported to the branch. EOS applies regardless of the status of any FIPS certificate associated with the branch; once an LTS reaches EOS, consumers should migrate to mainline or to a newer LTS.
+
+At any given time, at least one LTS branch is within its support window. Consecutive LTS branches overlap so that consumers always have a supported migration target.
+
+## Non-LTS FIPS branches
+
+FIPS validation requires a fixed snapshot of the cryptographic module's source code. Each time AWS-LC is submitted for FIPS validation, a branch is cut from mainline that preserves the exact code submitted. Most of these branches are not designated as LTS.
+
+Non-LTS FIPS branches exist solely to preserve the validated snapshot. They do not receive release tags, and consumers should not depend on them.
```
justsmth, line 114: This indicates that non-LTS branches will not receive tags, but do we mention anywhether whether LTS branches will?

VERSIONING.md
```diff
@@ -0,0 +1,133 @@
...
+* Cut from mainline every 2 years and supported for 5 years from the date the branch is cut.
+* Each LTS branch is submitted for FIPS validation.
+
+## LTS release policy
+
+### LTS version evolution
+
+Each LTS branch inherits mainline's major version at the time it is cut. Mainline then immediately bumps to the next major version. This guarantees that mainline and every LTS branch have distinct major version numbers.
+
+For example, when mainline is at `4.13.0` and an LTS branch is cut, the branch takes ownership of the `4.x` line. From that point on, the `4.x` line receives only the changes permitted below, as patch (`4.13.1`, `4.13.2`, ...) or minor (`4.14.0`, `4.15.0`, ...) increments. Mainline advances to `5.0.0` and continues normal development with minor increments only (`5.1.0`, `5.2.0`, ...). Because mainline never returns to the `4.x` line, version numbers on the LTS cannot collide with mainline. When the next LTS is cut approximately two years later, it takes ownership of whatever major version line mainline is on at that moment, and mainline bumps again.
```
justsmth: NP: The second sentence might be more clear as 
```
... From this point forward, the `4.x` version prefix is reserved exclusively
for the LTS branch. The LTS `4.x` line receives only the changes permitted below,
...
```

VERSIONING.md
samuel40791765: Great point, I had briefly thought about this and there were mentions that there were possible workarounds around this. In most cases, the LTS branch can ship a custom patch that fixes the same security issue while preserving ABI compatibility (even when mainline's version doesn't) by using a more minimal patch, side state, or wrapper variant rather than restructuring existing types. 

Of course, there could be edge cases where none of those approaches fit and we'd be forced into the trade-off scenario you described. If it comes to that, we could treat them as one-offs and decide on the appropriate approach based on the specific circumstance.

VERSIONING.md
samuel40791765: > Should there be an API or Macro (or both) to indicate at build or runtime whether this version of AWS-LC is an LTS? 

I think an LTS designation API may be worth doing if the need arises. Happy to take up explicit LTS detection in a separate PR if we want it.


I do want to call out that It would have to be its own value that's set and returned however. The comparison only works today because LTS-1 was coincidentally cut at major 4 with FIPS submission 4. As non-LTS submissions accumulate, the heuristic breaks down:
  
  | Event | Mainline (major / FIPS) | LTS-1 (major / FIPS) | LTS-2 (major / FIPS) |
  |---|---|---|---|
  | Today | `5` / `4` | `4` / `4` | — |
  | Non-LTS #5 cut | `5` / `5` | `4` / `4` | — |
  | Non-LTS #6 cut | `5` / `6` | `4` / `4` | — |
  | Non-LTS #7 cut | `5` / `7` | `4` / `4` | — |
  | LTS-2 cut at FIPS 8 | `6` / `8` | `4` / `4` | `5` / `8` |

When the second LTS is cut, its major (5) and FIPS submission (8) don't align, so the comparison would break.

VERSIONING.md
samuel40791765: Applied most wording updates.

Good point on ABI_VERSION. Although, I'm slightly unsure on how we've committed to updating this and exposing would require better knowledge of how we're planning to document and maintain the API.

VERSIONING.md
samuel40791765: Added as a sentence under "Release types"!

VERSIONING.md
samuel40791765: Restructured the above a bit to reflect that clearly

VERSIONING.md
samuel40791765: I reworded it slightly, but added a sentence that should reflect what we're looking for here.

justsmth (approved):

dougch (approved):

PR #3213, opened 2026-05-04
Fixes several issues across X509 and EVP parsing/comparison code

torben-hansen:
### Issues:

P425780141

### Description
                              
- The return value of `i2d_name_canon` was not being checked. Now propagates the error via `goto err` when encoding fails.
- When `X509_REVOKED_get_ext_d2i` failed on a critical extension, the function set `EXFLAG_INVALID` and returned success. Consumers that don't inspect `EXFLAG_INVALID` would silently skip the error, so now we return `0` directly.
- `EVP_parse_private_key` was calling `CBS_get_asn1` on the outer `cbs` instead of the inner `pkcs8` CBS, which had already advanced past the attributes. Fixed to operate on `&pkcs8`.
- `kAttributesTag` was defined as context-specific `[0]` primitive, but `attributes` is `[0] IMPLICIT SET OF Attribute` and `SET OF` is always constructed. Added `CBS_ASN1_CONSTRUCTED` so the field is detected correctly.
- `OPENSSL_memcmp` was called without checking whether either `KEM_KEY`'s `public_key` was `NULL`. Added a guard returning `-2` when either side is missing.
  
### Tests
  - `evp_extra_test.cc`: Added tests catching the two PKCS8 parsing bugs above.
  - `p_kem_test.cc`: Added test for `kem_pub_cmp`.
  - `evp_extra_test.cc`: Ensured the new KEM comparison test defines the public key before comparing.


skmcgrail (approved):

crypto/evp_extra/evp_asn1.c
justsmth, line 174: NP: also reject trailing data:
```c
  // Reject trailing data within the SEQUENCE, matching the strictness of
  // EVP_parse_public_key.
  if (CBS_len(&pkcs8) != 0) {
    OPENSSL_PUT_ERROR(EVP, EVP_R_DECODE_ERROR);
    return NULL;
  }
```

justsmth (approved):

PR #3220, opened 2026-05-06
Fix thread-local DRBG cleanup deadlock at process exit

### LINKED ISSUE #3197: Thread-local cleanup can cause deadlocks on Windows
AWS-LC registers an `atexit` handler that intentionally acquires a lock without releasing it: https://github.com/aws/aws-lc/blob/a7106f8abb39c468c8e0cd2720656147af2fa833/crypto/fipsmodule/rand/rand.c#L50-L93 

AWS-LC also acquires this lock during cleanup of thread-local objects:
https://github.com/aws/aws-lc/blob/a7106f8abb39c468c8e0cd2720656147af2fa833/crypto/fipsmodule/rand/rand.c#L165-L174
https://github.com/aws/aws-lc/blob/a7106f8abb39c468c8e0cd2720656147af2fa833/crypto/fipsmodule/rand/rand.c#L95-L99

We have been experiencing a race condition where, if a thread exits after `atexit` handlers have run, this thread-local cleanup will block indefinitely waiting to acquire the lock which is 'held' by the `atexit` handler. Since this thread-local cleanup runs under the [loader lock](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-best-practices#general-best-practices), the loader lock will also be held indefinitely. This breaks many Windows APIs that require the loader lock. 

For example, in our application, we frequently see that our process doesn't exit after shutting down the program. During shutdown, various worker threads start exiting and can hit the described race condition. For example, the thread which is blocked on the thread-local cleanup has the following stack:
```
	ntdll.dll!NtWaitForAlertByThreadId()
 	ntdll.dll!RtlpAcquireSRWLockExclusiveContended()
 	ntdll.dll!RtlAcquireSRWLockExclusive()
 	[Inline Frame] Zed.exe!thread_local_list_delete_node(rand_thread_local_state *)
 	Zed.exe!rand_thread_local_state_free(void * state_in)
 	Zed.exe!thread_local_destructor(void * module, unsigned long reason, void * reserved)
 	ntdll.dll!ImageTlsCallbackCaller()
 	ntdll.dll!LdrpCallInitRoutineInternal()
 	ntdll.dll!LdrpCallInitRoutine()
 	ntdll.dll!LdrpCallTlsInitializers()
 	ntdll.dll!LdrShutdownThread()
 	ntdll.dll!RtlExitUserThread()
 	kernel32.dll!BaseThreadInitThunk()
 	ntdll.dll!RtlUserThreadStart()
```

Since the loader lock is being held by this thread, APIs like `ExitProcess` block indefinitely waiting to acquire it, preventing the process from exiting.

```
 	ntdll.dll!NtWaitForSingleObject()
 	ntdll.dll!LdrpDrainWorkQueue()
 	ntdll.dll!RtlExitUserProcess()
 	kernel32.dll!ExitProcessImplementation()
 	Zed.exe!exit_or_terminate_process(const unsigned int return_code, bool should_call_terminate_proc)
 	Zed.exe!common_exit(const int return_code, const _crt_exit_cleanup_mode cleanup_mode, const _crt_exit_return_mode return_mode)
	Zed.exe!__scrt_common_main_seh()
 	kernel32.dll!BaseThreadInitThunk()
 	ntdll.dll!RtlUserThreadStart()
```

A different cleanup mechanism that isn't susceptible to this race condition should be used.

justsmth:
### Issues:
Resolves #3197

### Description of changes:
The `atexit`-registered shutdown routine (`rand_thread_local_state_clear_all`) previously acquired `thread_local_states_list_lock` and never released it. Any thread whose TLS destructor fired after that point would deadlock waiting for the same lock. On Windows, TLS destructors run under the loader lock, so this hangs `ExitProcess`.

This change releases the list lock at the end of `clear_all` and instead uses a `thread_local_drbg_shutdown_started` flag (checked under the lock) to coordinate post-shutdown behavior. The per-state `state_clear_lock` — still write-locked forever — continues to enforce the FIPS requirement that no output is returned from a zeroized DRBG.

### Call-outs:
- Post-shutdown, TLS destructors intentionally leak their state node rather than freeing a struct with a permanently-held mutex. This is process-exit-only and the OS reclaims the memory. The DRBG and entropy source have already been zeroized by rand_thread_local_state_clear_all before this path is reached, so no secret material remains in the leaked allocation.
- `clear_all` has an idempotency guard because the new `_FOR_TESTING` entry point means it could theoretically run twice; re-locking already-held per-state rwlocks would be UB.

### Testing:
Adds `randTest.ShutdownDoesNotDeadlockExitingThread` — forks a child that triggers shutdown zeroization on the main thread, then lets a worker thread exit. Parent enforces a 30s timeout so a regression fails the test instead of hanging the suite. Forking is necessary because shutdown zeroization permanently disables the RNG. Test is POSIX-only (the underlying lock cycle is platform-agnostic; the loader-lock symptom is Windows-specific).

### DIFF
--- a/crypto/fipsmodule/rand/internal.h
+++ b/crypto/fipsmodule/rand/internal.h
@@ -28,6 +28,15 @@ OPENSSL_EXPORT uint64_t get_private_thread_reseed_calls_since_initialization(voi
 OPENSSL_EXPORT uint64_t get_public_thread_generate_calls_since_seed(void);
 OPENSSL_EXPORT uint64_t get_public_thread_reseed_calls_since_initialization(void);
 
+// rand_thread_local_state_clear_all_FOR_TESTING runs the same shutdown
+// zeroization sequence that is registered via |atexit|. It is exposed for tests
+// that need to verify shutdown behavior (e.g. that a thread exiting after
+// shutdown does not deadlock). After this is called, no further |RAND_bytes|
+// calls in the test process will return output, so the caller is responsible
+// for arranging that no other code paths in the same process require
+// randomness afterwards.
+OPENSSL_EXPORT void rand_thread_local_state_clear_all_FOR_TESTING(void);
+
 // CTR_DRBG_STATE contains the state of a CTR_DRBG based on AES-256. See SP
 // 800-90Ar1.
 struct ctr_drbg_state_st {
--- a/crypto/fipsmodule/rand/rand.c
+++ b/crypto/fipsmodule/rand/rand.c
@@ -47,6 +47,16 @@ OPENSSL_STATIC_ASSERT((sizeof((struct rand_thread_local_state*)0)->generate_call
 DEFINE_BSS_GET(struct rand_thread_local_state *, thread_states_list_head)
 DEFINE_STATIC_MUTEX(thread_local_states_list_lock)
 
+// thread_local_drbg_shutdown_started is set to a non-zero value during process
+// exit by |rand_thread_local_state_clear_all|, while the linked-list lock is
+// held. All reads and writes occur under |thread_local_states_list_lock|, so
+// no atomic accessors are needed; doing the check under the lock also avoids
+// the otherwise-racy window where a TLS destructor could observe the flag as
+// unset, then block on the lock while shutdown zeroization runs, and then
+// free a state whose |state_clear_lock| has been intentionally write-locked
+// forever.
+DEFINE_BSS_GET(int, thread_local_drbg_shutdown_started)
+
 #if defined(_MSC_VER)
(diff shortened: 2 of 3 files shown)


torben-hansen (approved):

nebeid (approved):
Nit: maybe add to the PR description (so it's included in the commit message) in the first Callout bullet something along "the DRBG and entropy source have already been zeroized by `rand_thread_local_state_clear_all` before this path is reached, so no secret material remains in the leaked allocation." because I wondered about that when I reached this bullet.

I assume you tested with and without the fix.

PR #3229, opened 2026-05-08
Check parameters before comparing pqdsa public keys

samuel40791765:
### Issues:
Addresses `V2196133741`

### Description of changes:
`pqdsa_pub_cmp` memcmp'd two public keys using the length from `a`'s variant, with no check that `a` and `b` were the same ML-DSA parameter set.  Adds a `pqdsa_cmp_parameters` helper (mirroring `kem_cmp_parameters` in `p_kem_asn1.c`) that validates both keys are populated and share the same NID, and makes `pqdsa_pub_cmp` return early when parameters don't match.
  
### DIFF
--- a/crypto/evp_extra/p_pqdsa_asn1.c
+++ b/crypto/evp_extra/p_pqdsa_asn1.c
@@ -137,13 +137,42 @@ static int pqdsa_pub_encode(CBB *out, const EVP_PKEY *pkey) {
   return 1;
 }
 
-static int pqdsa_pub_cmp(const EVP_PKEY *a, const EVP_PKEY *b) {
-  PQDSA_KEY *a_key = a->pkey.pqdsa_key;
-  PQDSA_KEY *b_key = b->pkey.pqdsa_key;
+// pqdsa_cmp_parameters returns 1 if |a| and |b| hold populated PQDSA keys
+// with the same ML-DSA NID, 0 if their NIDs differ, or -2 if either operand
+// is missing its key or parameters. The tri-state return aligns with the
+// |EVP_PKEY_cmp| convention (1 = equal, 0 = not equal, negative = error).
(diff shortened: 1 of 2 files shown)


crypto/evp_extra/p_pqdsa_asn1.c
```diff
@@ -137,13 +131,35 @@ static int pqdsa_pub_encode(CBB *out, const EVP_PKEY *pkey) {
   return 1;
 }
 
+static int pqdsa_cmp_parameters(const EVP_PKEY *a, const EVP_PKEY *b) {
+  const PQDSA_KEY *a_key = a->pkey.pqdsa_key;
```
WillChilds-Klein, line 148: do we need NULL checks on `a` and `b` here before the dereference?

also, can we document this functions return values in a comment above? why return -2 instead of 0? callers might inadvertently do `!pqdsa_cmp_parameters(...)` and have unexpected behavior

samuel40791765: Added NULL checks and a doc comment. 

I was confused about this at first too, but the tri-state return is intentional to align with `EVP_PKEY_cmp`'s convention (1 = equal, 0 = not equal, negative = error). Also mirrored the same `NULL` check and docstring onto `kem_cmp_parameters` since it's applicable there as well.
https://github.com/aws/aws-lc/blob/96514809cf22281828e8e8ce993f8e89099acc66/include/openssl/evp.h#L56-L61

WillChilds-Klein: thanks. if we need to return negative on error, -2 still "feels" like an odd return value vs. -1 but if "negative" is the documented contract it doesn't really matter.

WillChilds-Klein (approved):

geedo0 (approved):

PR #3231, opened 2026-05-08
Free existing union arm by current type in PKCS7_set_type

samuel40791765:
### Issues:
Addresses `V2196133741`, `V2186717843`

### Description of changes:
`PKCS7_set_type` installed the new `p7->type` and then called the free function matching the *new* type on whatever was in the `p7->d` union (e.g. `PKCS7_DIGEST_free(p7->d.digest)` on memory that was actually a `PKCS7_SIGNED*` from an earlier call). When the second call's type differs from the first, this frees memory through the wrong destructor.

### Testing:
New test that tests correct freeing

### DIFF
--- a/crypto/pkcs7/pkcs7.c
+++ b/crypto/pkcs7/pkcs7.c
@@ -198,10 +198,39 @@ int PKCS7_set_type(PKCS7 *p7, int type) {
     return 0;
   }
 
+  // Free any previously-installed content using the current |p7->type| so the
+  // correct destructor runs on the existing union member, instead of the new
+  // type's destructor running on type-confused memory.
+  if (p7->d.ptr != NULL) {
+    switch (OBJ_obj2nid(p7->type)) {
+      case NID_pkcs7_signed:
+        PKCS7_SIGNED_free(p7->d.sign);
+        break;
+      case NID_pkcs7_digest:
+        PKCS7_DIGEST_free(p7->d.digest);
+        break;
+      case NID_pkcs7_data:
+        ASN1_OCTET_STRING_free(p7->d.data);
+        break;
+      case NID_pkcs7_signedAndEnveloped:
+        PKCS7_SIGN_ENVELOPE_free(p7->d.signed_and_enveloped);
+        break;
+      case NID_pkcs7_enveloped:
+        PKCS7_ENVELOPE_free(p7->d.enveloped);
+        break;
+      case NID_pkcs7_encrypted:
+        PKCS7_ENCRYPT_free(p7->d.encrypted);
+        break;
+      default:
+        ASN1_TYPE_free(p7->d.other);
+        break;
+    }
+    p7->d.ptr = NULL;
+  }
+
   switch (type) {
     case NID_pkcs7_signed:
       p7->type = obj;
-      PKCS7_SIGNED_free(p7->d.sign);
       p7->d.sign = PKCS7_SIGNED_new();
       if (p7->d.sign == NULL) {
         return 0;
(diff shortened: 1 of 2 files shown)


justsmth (dismissed):

crypto/pkcs7/pkcs7_test.cc
```diff
@@ -1565,6 +1565,22 @@ TEST(PKCS7Test, GettersSetters) {
...
+  bssl::UniquePtr<PKCS7> p7(PKCS7_new());
+  ASSERT_TRUE(p7);
+  ASSERT_TRUE(PKCS7_set_type(p7.get(), NID_pkcs7_signed));
+  ASSERT_TRUE(PKCS7_set_type(p7.get(), NID_pkcs7_digest));
+  ASSERT_TRUE(PKCS7_set_type(p7.get(), NID_pkcs7_data));
+  ASSERT_TRUE(PKCS7_set_type(p7.get(), NID_pkcs7_signedAndEnveloped));
+  ASSERT_TRUE(PKCS7_set_type(p7.get(), NID_pkcs7_enveloped));
+  ASSERT_TRUE(PKCS7_set_type(p7.get(), NID_pkcs7_encrypted));
+  EXPECT_TRUE(PKCS7_type_is_encrypted(p7.get()));
+}
```
justsmth, line 1592: NP: more testing:
```c++
  // Also exercise the |default:| branch: a PKCS7 whose current arm is
  // |d.other| (e.g. one parsed with an unrecognized content OID) must be
  // freed via |ASN1_TYPE_free| before installing a known type.
  static const unsigned char other_data[] = {0x30, 0x02, 0x01, 0x02};
  p7.reset(PKCS7_new());
  ASSERT_TRUE(p7);
  ASN1_STRING *seq = ASN1_STRING_new();
  ASSERT_TRUE(seq);
  ASSERT_TRUE(ASN1_STRING_set(seq, other_data, sizeof(other_data)));
  p7->d.other = ASN1_TYPE_new();
  ASSERT_TRUE(p7->d.other);
  ASN1_TYPE_set(p7->d.other, V_ASN1_OCTET_STRING, seq);
  ASSERT_TRUE(PKCS7_set_type(p7.get(), NID_pkcs7_signed));
  EXPECT_TRUE(PKCS7_type_is_signed(p7.get()));
```

justsmth (approved):

crypto/pkcs7/pkcs7.c
```diff
@@ -198,10 +198,39 @@ int PKCS7_set_type(PKCS7 *p7, int type) {
...
+        PKCS7_SIGN_ENVELOPE_free(p7->d.signed_and_enveloped);
+        break;
+      case NID_pkcs7_enveloped:
+        PKCS7_ENVELOPE_free(p7->d.enveloped);
+        break;
+      case NID_pkcs7_encrypted:
+        PKCS7_ENCRYPT_free(p7->d.encrypted);
+        break;
+      default:
+        ASN1_TYPE_free(p7->d.other);
```
WillChilds-Klein, line 225: what if the `ptr` field is the one used in the union?

https://github.com/aws/aws-lc/blob/96514809cf22281828e8e8ce993f8e89099acc66/include/openssl/pkcs7.h#L143

is this function safe to call on a `char *`?

samuel40791765: `d.ptr` is just a `NULL`-check alias for the same storage as d.other,  no code in the tree ever actually populates or frees it as a raw `char *.` The ASN.1 template in `pkcs7_asn1.c` populates unknown content types as `d.other = ASN1_TYPE*` (the `p7default_tt`  fallback) and the d.ptr references elsewhere only read it as a `NULL`-check handle. 
https://github.com/aws/aws-lc/blob/96514809cf22281828e8e8ce993f8e89099acc66/crypto/pkcs7/pkcs7_asn1.c#L12-L27

So the default arm here always sees memory that was written through `d.other`.

WillChilds-Klein (approved):

PR #3232, opened 2026-05-08
Reject len < -1 in ASN1_mbstring_ncopy

samuel40791765:
### Issues:
Addresses `V2196133741`

### Description of changes:
`ASN1_mbstring_ncopy` treats `len == -1` as "call strlen on in". Any other negative value fell through unchanged and was cast to `size_t` by `CBS_init`, producing a huge length. Adds an explicit early reject for `len < -1` with `ASN1_R_ILLEGAL_FORMAT`; `-1` and non-negative values continue to behave as before.

### Call-outs:
None.

### Testing:
Existing ASN.1 string tests continue to pass. 

### DIFF
--- a/crypto/asn1/a_mbstr.c
+++ b/crypto/asn1/a_mbstr.c
@@ -33,6 +33,10 @@ OPENSSL_DECLARE_ERROR_REASON(ASN1, INVALID_UTF8STRING)
 int ASN1_mbstring_ncopy(ASN1_STRING **out, const unsigned char *in,
                         ossl_ssize_t len, int inform, unsigned long mask,
                         ossl_ssize_t minsize, ossl_ssize_t maxsize) {
+  if (len < -1) {
+    OPENSSL_PUT_ERROR(ASN1, ASN1_R_ILLEGAL_FORMAT);
+    return -1;
+  }
   if (len == -1) {
     len = strlen((const char *)in);
   }
--- a/crypto/asn1/asn1_test.cc
+++ b/crypto/asn1/asn1_test.cc
@@ -1711,6 +1711,25 @@ TEST(ASN1Test, MBString) {
     ERR_clear_error();
     EXPECT_EQ(nullptr, str);
   }
+
+  // |len| values below -1 must be rejected; only -1 is special-cased to mean
+  // "call strlen on |in|". Any other negative value would otherwise be cast
+  // to a huge |size_t| by |CBS_init|.
+  static const uint8_t kDummy[] = {'a'};
+  ASN1_STRING *str = nullptr;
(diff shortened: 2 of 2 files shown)


justsmth (dismissed):
NP: We could add testing to [the end of `TEST(ASN1Test, MBString)`](https://github.com/aws/aws-lc/blob/main/crypto/asn1/asn1_test.cc#L1713)
```c++
  // |len| values below -1 must be rejected. Only -1 is special-cased to mean
  // "call strlen on |in|"; any other negative value would otherwise be cast to
  // a huge |size_t| by |CBS_init|.
  static const uint8_t kDummy[] = {'a'};
  for (ossl_ssize_t bad_len : {(ossl_ssize_t)-2, (ossl_ssize_t)INT_MIN,
                               std::numeric_limits<ossl_ssize_t>::min()}) {
    SCOPED_TRACE(bad_len);
    EXPECT_EQ(-1, ASN1_mbstring_copy(nullptr, kDummy, bad_len, MBSTRING_UTF8,
                                     B_ASN1_UTF8STRING));
    ERR_clear_error();
    ASN1_STRING *str = nullptr;
    EXPECT_EQ(-1, ASN1_mbstring_ncopy(&str, kDummy, bad_len, MBSTRING_UTF8,
                                      B_ASN1_UTF8STRING, /*minsize=*/0,
                                      /*maxsize=*/0));
    EXPECT_EQ(nullptr, str);
    ERR_clear_error();
  }
```

justsmth (dismissed):

crypto/asn1/a_mbstr.c
```diff
@@ -33,6 +33,10 @@ OPENSSL_DECLARE_ERROR_REASON(ASN1, INVALID_UTF8STRING)
 int ASN1_mbstring_ncopy(ASN1_STRING **out, const unsigned char *in,
                         ossl_ssize_t len, int inform, unsigned long mask,
                         ossl_ssize_t minsize, ossl_ssize_t maxsize) {
+  if (len < -1) {
+    OPENSSL_PUT_ERROR(ASN1, ASN1_R_ILLEGAL_FORMAT);
+    return -1;
+  }
   if (len == -1) {
```
WillChilds-Klein, line 40: not your change, but do we have a test case for this?

WillChilds-Klein (dismissed):

crypto/asn1/a_mbstr.c
samuel40791765: Added

justsmth (approved):

crypto/asn1/a_mbstr.c
WillChilds-Klein: thanks sam!

WillChilds-Klein (approved):

PR #3234, opened 2026-05-08
Free existing responderId union arm in OCSP_RESPID setters

samuel40791765:
### Issues:
Addresses V2196133741 (F9 — OCSP_RESPID_set_by_{key,name}: type-confused free on responderId union)
  
### Description of changes:
`OCSP_RESPID` stores either an `X509_NAME*` (byName) or an `ASN1_OCTET_STRING*` (byKey) in a union. Re-setting the responderId leaked the prior arm or freed it through the wrong destructor, depending on the transition. Both setters now call a small `ocsp_respid_clear` helper that frees the arm matching the current `respid->type` before installing the new value, so every sequence (name↔name, name↔key, key↔key, key↔name) is correct.
  
### Call-outs:
None.

### Testing:
Extends `OCSPResponseSignTestExtended.OCSPResponseSign` with four sequential `OCSP_basic_sign` calls on one `OCSP_BASICRESP`, covering all four responderId transitions. ASAN in CI will surface any leak or misuse.
  
### DIFF
--- a/crypto/ocsp/ocsp_server.c
+++ b/crypto/ocsp/ocsp_server.c
@@ -65,14 +65,33 @@ int OCSP_basic_add1_cert(OCSP_BASICRESP *resp, X509 *cert) {
   return 1;
 }
 
+// ocsp_respid_clear frees any previously-installed content of |respid| using
+// the destructor matching the current |respid->type|, so callers can install
+// a new value without leaking or triggering a type-confused free.
+static void ocsp_respid_clear(OCSP_RESPID *respid) {
+  switch (respid->type) {
+    case V_OCSP_RESPID_NAME:
+      X509_NAME_free(respid->value.byName);
+      respid->value.byName = NULL;
+      break;
+    case V_OCSP_RESPID_KEY:
+      ASN1_OCTET_STRING_free(respid->value.byKey);
+      respid->value.byKey = NULL;
+      break;
+  }
+}
+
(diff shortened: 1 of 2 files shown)


crypto/ocsp/ocsp_server.c
justsmth, line 96: NP: `ocsp_respid_clear` is called before `X509_NAME_set`, so if `X509_NAME_dup` (inside `X509_NAME_set`) fails, the prior union arm has already been freed while `respid->type` still names it.  `set_by_key` (below) avoids this by building the replacement first.
```c++
static int OCSP_RESPID_set_by_name(OCSP_RESPID *respid, X509 *cert) {
  GUARD_PTR(respid);
  GUARD_PTR(cert);

  // Duplicate before clearing so an allocation failure leaves |respid|'s
  // prior union arm intact.
  X509_NAME *byName = X509_NAME_dup(X509_get_subject_name(cert));
  if (byName == NULL) {
    return 0;
  }
  ocsp_respid_clear(respid);
  respid->type = V_OCSP_RESPID_NAME;
  respid->value.byName = byName;
  return 1;
}
```

justsmth (dismissed):

crypto/ocsp/ocsp_server.c
samuel40791765: Reordered `OCSP_RESPID_set_by_name` to dup first so an allocation failure leaves respid unchanged. Both setters now share the same allocate → clear → install shape.

justsmth (approved):

geedo0 (approved):

PR #3235, opened 2026-05-08
Fix correctness findings from penpal testing

prasden:
### Issues:
Addresses: P409219367

#### Description of changes: 
This PR adds a small series of correctness fixes brought up by Penpal testing: 
 - `BIO_write_ex` returned early on NULL bio without initializing `*written_bytes` in `bio.c.` We now initialize `written_bytes` to 0 before the NULL check so the caller receives that 0 bytes were written if any checks fail. 
 
 - `NCONF_load_bio` had a signed overflow in bufnum in `conf.c`. We now guard against overflow before the addition. 
 -  `BN_le2bn` did not reset the sign flag when reusing an existing BIGNUM in `bytes.c`. We now clear it to avoid staleness.
 - `PEM_write_bio` previously  used an `int` loop counter in pem_lib.c. We now use `long` to match the input length type and prevent overflow.
 - `crl_cb` in `x_crl.c` did not free `idp/akid` on error paths. We now free and NULL them before returning to prevent leaks


samuel40791765 (dismissed):

crypto/pem/pem_lib.c
```diff
@@ -483,7 +483,8 @@ int PEM_write(FILE *fp, const char *name, const char *header,
 
 int PEM_write_bio(BIO *bp, const char *name, const char *header,
                   const unsigned char *data, long len) {
-  int nlen, n, i, j, outl;
+  int nlen, n, i, outl;
+  long j;
```
justsmth: `j` was promoted to `long` to avoid overflow, but `i` has the same risk -- it accumulates `outl` each iteration (`i += outl`) and is used in the return value. `i` should also be `long`.

Additionally, the return statement `return i + outl` narrows back to the `int` return type -- if the total exceeds `INT_MAX`, the behavior is implementation-defined. A bounds check before returning (or erroring out) would make this more correct.

prasden: Changed `i` to a long as well and added an overflow check before return in  [98a3f76](https://github.com/aws/aws-lc/commit/98a3f76a20d58c9998dd5ac419c542a5be897d6d)

justsmth (dismissed):

crypto/bio/bio.c
samuel40791765, line 328: Is this still needed if we call it earlier in the function?

prasden: It was redundant, removed.

geedo0 (approved):

samuel40791765 (approved):

PR #3236, opened 2026-05-08
Include trailing NUL in BIO_ADDR_rawaddress AF_UNIX length

samuel40791765:
### Issues:
Addresses V2196133741 (F4 — BIO_ADDR_rawaddress writes len+1 after reporting len)

### Description of changes:
For `AF_UNIX`, `BIO_ADDR_rawaddress` wrote a NUL terminator to `p[len]` while reporting `*l = len` (NUL excluded). A caller sizing `p` from the returned `*l` would take a 1-byte out-of-bounds write.

Changes `*l` to include the trailing NUL for AF_UNIX so that the reported length matches the bytes written and reflects the  NULL-terminated path convention. The public contract in `include/openssl/bio.h` is updated to match. Non-UNIX families (`AF_INET`, `AF_INET6`) are unchanged.

### Call-outs:
 callers who previously read `*l` for AF_UNIX now receive `strlen(path) + 1` instead of `strlen(path)`. The existing `bio_socket_test` already allocates `len + 1` buffers and is unaffected; out-of-tree callers that feed `*l` back into `BIO_ADDR_rawmake` will now pass the NUL-inclusive length, which `BIO_ADDR_rawmake` already tolerates.

### Testing:
Existing `bio_socket_test` continues to pass.
  
### DIFF
--- a/crypto/bio/bio_addr.c
+++ b/crypto/bio/bio_addr.c
@@ -151,7 +151,10 @@ int BIO_ADDR_rawaddress(const BIO_ADDR *ap, void *p, size_t *l) {
     return 0;
   }
   if (l != NULL) {
-    *l = len;
+    // AF_UNIX includes the trailing NUL written below so |*l| matches the
+    // bytes written to |p| and reflects sockaddr_un's NUL-terminated path.
+    // OpenSSL does not write this NUL; aws-lc intentionally does.
+    *l = (ap->sa.sa_family == AF_UNIX) ? len + 1 : len;
   }
 
   if (p != NULL) {
--- a/crypto/bio/bio_socket_test.cc
+++ b/crypto/bio/bio_socket_test.cc
@@ -130,7 +130,7 @@ struct SockaddrStorage {
         if (1 !=
             BIO_ADDR_rawmake(
                 bap, AF_UNIX, &sock->sun_path,
-                OPENSSL_strnlen(sock->sun_path, sizeof(sock->sun_path)) + 1,
+                OPENSSL_strnlen(sock->sun_path, sizeof(sock->sun_path)),
                 0)) {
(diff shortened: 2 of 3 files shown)


justsmth (commented):
* I noticed on the other PR -- the `+ 1` here is wrong: https://github.com/aws/aws-lc/blob/main/crypto/bio/bio_socket_test.cc#L133
* Add test:
```c++
// |BIO_ADDR_rawaddress| writes a NUL terminator after the AF_UNIX path. The
// reported |*l| must include that NUL so that callers using the standard
// probe-then-allocate-exactly-|*l| pattern allocate a buffer large enough to
// hold both the path and the terminator.
TEST_F(BIOAddrTest, RawAddressUnixLengthIncludesNul) {
  const char *path = "/tmp/test.sock";
  const size_t path_len = strlen(path);
  ASSERT_EQ(BIO_ADDR_rawmake(addr1, AF_UNIX, path, path_len + 1, 0), 1);

  size_t probed_len = 0;
  ASSERT_EQ(BIO_ADDR_rawaddress(addr1, nullptr, &probed_len), 1);
  EXPECT_EQ(probed_len, path_len + 1);

  std::vector<char> buf(probed_len);
  size_t written_len = buf.size();
  ASSERT_EQ(BIO_ADDR_rawaddress(addr1, buf.data(), &written_len), 1);
  EXPECT_EQ(written_len, probed_len);
  EXPECT_EQ(buf[written_len - 1], '\0');
  EXPECT_EQ(std::string(buf.data()), std::string(path));
}
```

samuel40791765:
good catch! fixed it and added a test

justsmth (approved):

WillChilds-Klein (approved):

PR #3240, opened 2026-05-11
Add SHRT_MAX caps to bound iteration and input lengths

prasden:
### Issues:
Addresses `P409219367`

### Description of changes: 
This PR adds iteration/input length bound checks across PKCS7, OCSP, SSKDF, X509 SPKI, and ASN.1 OID parsing to address the Penpal findings covering unbounded iteration counts and input lengths.

### Call-outs:
-  The relevant RFCs don't specify upper bounds, and choosing a tight limit would require knowing sizes typically seen in real data, which isn't easy to determine. We apply `SHRT_MAX` as a conservative cap to protect against bad input. 

### DIFF
--- a/crypto/bytestring/cbb.c
+++ b/crypto/bytestring/cbb.c
@@ -614,6 +614,11 @@ int CBB_add_asn1_oid_from_text(CBB *cbb, const char *text, size_t len) {
     return 0;
   }
 
+  if (len > (size_t)SHRT_MAX) {
+    OPENSSL_PUT_ERROR(CRYPTO, ERR_R_OVERFLOW);
+    return 0;
+  }
+
   CBS cbs;
   CBS_init(&cbs, (const uint8_t *)text, len);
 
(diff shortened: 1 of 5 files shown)


WillChilds-Klein (commented):
`SHRT_MAX`, ~32k on most platforms, seems a bit arbitrary. i'm not saying we shouldn't bound input lengths, but is this the right bound for all of these mostly unrelated pieces of functionality? i'm trying to understand the reasoning for each bound.

prasden:
> `SHRT_MAX`, ~32k on most platforms, seems a bit arbitrary. i'm not saying we shouldn't bound input lengths, but is this the right bound for all of these mostly unrelated pieces of functionality? i'm trying to understand the reasoning for each bound.

Yes, part of the goal with this PR was to open that discussion. The RFCs don't define any explicit upper bounds, and deriving a specific cap for each operation would require real-world data to estimate against, so a single consistent cap was chosen to resolve Penpal findings. That said, going unbounded is also an option if we believe this cap is too arbitrary and doesn't provide a clear benefit. Open to discussion.

WillChilds-Klein (approved):
Well, looking through each case, I guess I don't have better suggestions for a limit.

geedo0 (approved):
Some of these are such overly paranoid checks, but fine

PR #3243, opened 2026-05-12
Release cipher_data on error path too for EVP_CTRL_INIT and EVP_CTRL_COPY

torben-hansen:
### Description of changes: 

On the error-path, which I don't think is actually reachable right now, `cipher_data` should be released, otherwise it would/can leak. It's not exactly likely a context would ever be re-used in the error case, but who knows. It doesn't cost anything to gate it.

### Testing:

Two tests, exercising the error path and test new assertion.

### DIFF
--- a/crypto/cipher_extra/cipher_test.cc
+++ b/crypto/cipher_extra/cipher_test.cc
@@ -20,6 +20,7 @@
 #include <openssl/sha.h>
 #include <openssl/span.h>
 
+#include "../fipsmodule/cipher/internal.h"
 #include "../internal.h"
 #include "../test/file_test.h"
 #include "../test/test_util.h"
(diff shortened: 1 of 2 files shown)


crypto/cipher_extra/cipher_test.cc
```diff
@@ -1406,6 +1407,101 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+                               size_t) {
+  return 1;
+}
+
+static int mock_failing_ctrl(EVP_CIPHER_CTX *, int type, int, void *) {
+  if (type == EVP_CTRL_INIT) {
+    return 0;
+  }
+  return -1;
+}
```
justsmth, line 1430: Returning `1` means SUCCESS, not failure:
```c++
// Mock ciphers used to exercise the error paths in |EVP_CipherInit_ex| and
// |EVP_CIPHER_CTX_copy|. The mocks need |ctx_size| != 0, otherwise
// |cipher_data| is not allocated before the ctrl callback runs, which is
// what |InitErrorPathReleasesCipherData| and |CopyErrorPathReleasesCipherData|
// test is free'd on the error path.
static int mock_noop_init(EVP_CIPHER_CTX *, const uint8_t *, const uint8_t *,
                          int) {
  return 1;
}

static int mock_noop_cipher(EVP_CIPHER_CTX *, uint8_t *, const uint8_t *,
                            size_t) {
  return 1;
}

static int mock_failing_init_ctrl(EVP_CIPHER_CTX *, int type, int, void *) {
  if (type == EVP_CTRL_INIT) {
    return 0;
  }
  return -1;
}
```

crypto/cipher_extra/cipher_test.cc
```diff
@@ -1406,6 +1407,101 @@ TEST(CipherTest, Empty_EVP_CIPHER_CTX_V1187459157) {
...
+  std::vector<uint8_t> key(16, 0);
+  ASSERT_TRUE(EVP_CipherInit_ex(src.get(), &kFailingCopyCipher, nullptr,
+                                key.data(), nullptr, 1));
+
+  bssl::UniquePtr<EVP_CIPHER_CTX> dst(EVP_CIPHER_CTX_new());
+  ASSERT_TRUE(dst);
+  EXPECT_FALSE(EVP_CIPHER_CTX_copy(dst.get(), src.get()));
+  EXPECT_EQ(dst->cipher, nullptr);
+  EXPECT_EQ(dst->cipher_data, nullptr);
+}
```
justsmth, line 1508: One more thing to test:
```c++
  // A follow-up init with a real cipher on |dst| must still work and must not
  // depend on leaked state from the failed copy.
  EXPECT_TRUE(EVP_CipherInit_ex(dst.get(), EVP_aes_128_ecb(), nullptr,
                                key.data(), nullptr, 1));
```

crypto/cipher_extra/cipher_test.cc
torben-hansen: Right. The namespace is used for mocking a failing cipher operation not for the each individual cipher. So, on the side of not changing name.

crypto/cipher_extra/cipher_test.cc
torben-hansen: Right, as the other test does. took this.

samuel40791765 (approved):

prasden (approved):

PR #3245, opened 2026-05-12
Add Keccak-256 (Ethereum-style, original 0x01 padding)

fabrice102:
### Issues:
N/A

### Description of changes:

**[edit]** Following review, Keccak-256 was moved **out of the FIPS module** into `crypto/keccak/` (mirroring blake2), reusing the module's FIPS 202 primitives via external linkage; `FIPS202_Init` is unchanged. The public API is now `EVP_keccak256()` (no underscore), tests live in `crypto/keccak/keccak256_test.cc`, and `AWSLC_API_VERSION` is bumped to 36. See [the comment below](https://github.com/aws/aws-lc/pull/3245#issuecomment-5024179491) for the full changelog — the "Approach" section above describes the original in-module design and is superseded.

---

Adds Keccak-256 with the original Keccak submission's padding byte (0x01) — the variant used by Ethereum, Solidity's `keccak256()`, and other blockchain ecosystems. This is **not** FIPS 202 SHA3-256 (which uses 0x06 padding); rate (1088 bits) and digest length (256 bits) match, but the padding byte differs, so outputs differ for every input.

### Approach:

The implementation lives inside the FIPS module
(`crypto/fipsmodule/sha/sha3.c`) so it shares the audited Keccak-f[1600] permutation, FIPS202_Init/Update/Finalize buffering, and assembly backends with the SHA-3 family. The only delta in the module's algorithmic code path is one extra accepted padding byte:

    if (pad != SHA3_PAD_CHAR &&
        pad != SHAKE_PAD_CHAR &&
        pad != KECCAK_PAD_CHAR) {  // 0x01
      return 0;
    }

The new public surface is exactly `EVP_keccak_256()` (plus the existing `EVP_Digest*` machinery). This matches the BoringSSL/AWS-LC convention established by SHA-3: the `EVP_MD` is the public API; direct C entry points (`KECCAK_256_Init/Update/Final`, one-shot `KECCAK_256()`) live in `crypto/fipsmodule/sha/internal.h` and are exported only for in-tree consumers, mirroring `SHA3_256_*`.

### Call-outs:

- **NOT FIPS-approved.** `KECCAK_256_Final` and the `KECCAK_256()` one-shot intentionally omit `FIPS_service_indicator_update_state()`, so calling Keccak-256 leaves the per-thread approved-services counter unchanged. Verified by a new row in `service_indicator_test.cc::kDigestTestVectors` with `AWSLC_NOT_APPROVED`. The negative invariant is enforced both at runtime (the parameterized `EVPMDServiceIndicatorTest.EVP_Digests` catches accidental `_update_state` calls) and structurally (a future reviewer can grep the file and see only "Intentionally no" comments).

- **No NID, no OID.** The Ethereum-style 0x01-padding Keccak is not standardised and OpenSSL has not assigned an OID for it either. The EVP_MD's `type` is `NID_undef`, and the digest is registered in `nid_to_digest_mapping` by name only. As a consequence, `EVP_keccak_256()` cannot be used with `EVP_DigestSign*` / `EVP_DigestVerify*` for RSA-PKCS#1-v1_5 or ECDSA, and `EVP_marshal_digest_algorithm` will fail with `DIGEST_R_UNKNOWN_HASH`. `objects.txt`, `obj_dat.h`, and `nid.h` are intentionally untouched.

- **OpenSSL compatibility.** The registered EVP_MD name is `KECCAK-256` (uppercase, hyphenated), matching OpenSSL 3.2+'s default-provider registration in `providers/implementations/include/prov/names.h`. Portable code that does `EVP_MD_fetch(NULL, "KECCAK-256", NULL)` on OpenSSL works against AWS-LC via `EVP_get_digestbyname("KECCAK-256")`. A `keccak-256` lowercase alias is also registered.

- **FIPS module impact.** The audited Keccak-f[1600] permutation, absorb/squeeze code, and existing SHA-3 entry points are unchanged byte-for-byte. The only edit inside the module boundary is the 3-line `FIPS202_Init` pad-byte check above plus the new `KECCAK_256_*` wrappers.

### Testing:

All Keccak-256 tests live in `crypto/digest_extra/digest_test.cc` alongside the rest of the digest tests. New coverage:

- `kTestVectors[]` — 3 well-known Ethereum-ecosystem vectors (empty / "abc" / fox-pangram). These exercise the EVP one-shot, byte-at-a-time streaming, copy/move-on-context, and unaligned-input paths in `TestDigest()`.
- `DigestTest.Keccak256DiffersFromSHA3_256` — invariant: for the same input, Keccak-256 and SHA3-256 outputs must always differ. Guards against a future maintainer "fixing" the padding byte.
- `DigestTest.Keccak256KAT` — file-driven tests using 137 short-message vectors (0..1088 bits) and 100 long-message vectors (2184..110688 bits), in the same `Len/Msg/MD` format used by SHA-3 KATs. Mirrors the `FileTestGTest` pattern from `sha3_test.cc::NISTTestVectors`.
- `DigestTest.Getters` — extended with `EVP_get_digestbyname("KECCAK-256")`, the lowercase alias, and an `EVP_MD_type == NID_undef` assertion.
- `service_indicator_test.cc::kDigestTestVectors` — one new row, parameterized as `EVP_Digests/13`, asserts `AWSLC_NOT_APPROVED` through Init / Update / Final.

KAT vectors are generated by `util/gen_keccak256_kat.py`, which uses pycryptodome (an independent Keccak-256 implementation) as the reference. The script sanity-checks pycryptodome against the well-known Ethereum vectors before generating, so a misconfigured environment fails loudly. Reference and implementation under test come from completely different code, so the test asserts genuine cross-implementation agreement.

`bssl speed -filter KECCAK-256` is wired up next to the SHA-3 benchmarks for performance regression tracking.

Verified end-to-end:
- Debug build: 2707/2707 active `crypto_test` cases pass, including `DigestTest.Keccak256*`, `SHA3Test.*` (no regression), and the full digest suite.
- FIPS build (`-DFIPS=1 -DBUILD_SHARED_LIBS=1`): 3719/3719 active cases pass, including all 14 rows of `EVPMDServiceIndicatorTest.EVP_Digests` (was 13).
- `bssl speed -filter KECCAK-256` produces sensible numbers across all default chunk sizes.


crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -403,6 +403,12 @@ static const uint8_t kOutput_sha3_512[SHA3_512_DIGEST_LENGTH] = {
     0x85, 0x3c, 0x64, 0xa1, 0x56, 0x6f, 0xeb, 0x76, 0x25, 0x9a, 0x4a, 0x44,
     0x23, 0xf7, 0xcf, 0x46};
 
+// Keccak-256 of |kPlaintext|, computed by util/keccak_ref.py.
```
manastasova: `util/gen_keccak256_kat.py` ?

crypto/digest_extra/digest_extra.c
```diff
@@ -262,6 +267,40 @@ static const EVP_MD evp_md_blake2b256 = {
...
+static int keccak_256_update(EVP_MD_CTX *ctx, const void *data, size_t len) {
+  return KECCAK_256_Update(ctx->md_data, data, len);
+}
+
+static void keccak_256_final(EVP_MD_CTX *ctx, uint8_t *md) {
+  AWSLC_ASSERT(KECCAK_256_Final(md, ctx->md_data));
+}
+
+// |EVP_MD_FLAG_DIGALGID_ABSENT| matches OpenSSL 3.2+'s
+// |PROV_DIGEST_FLAG_ALGID_ABSENT| on KECCAK-{224,256,384,512}. With
```
manastasova, line 286: Good point. Are there other algorithms that should update the flag `EVP_MD_FLAG_DIGALGID_ABSENT`?

crypto/digest_extra/digest_extra.c
```diff
@@ -41,6 +42,10 @@ static const struct nid_to_digest nid_to_digest_mapping[] = {
     {NID_sha3_512, EVP_sha3_512, SN_sha3_512, LN_sha3_512},
     {NID_shake128, EVP_shake128, SN_shake128, LN_shake128},
     {NID_shake256, EVP_shake256, SN_shake256, LN_shake256},
+    // Keccak-256 has no NID/OID (the Ethereum-style 0x01-padding variant is
+    // not standardised). Registered by name only; matches OpenSSL 3.2+'s
+    // "KECCAK-256" provider name for cross-library lookups.
+    {NID_undef, EVP_keccak_256, "KECCAK-256", "keccak-256"},
```
manastasova: Do we need the `keccak-256` entry? `blake2` has no entry in the table. Should we add it as well?

crypto/fipsmodule/sha/internal.h
```diff
@@ -506,6 +514,28 @@ int SHA3_512_Update(KECCAK1600_CTX *sha, const void *data, size_t len);
 // digest to |out|. It returns one on success and zero on programmer error.
 int SHA3_512_Final(uint8_t out[SHA3_512_DIGEST_LENGTH], KECCAK1600_CTX *sha);
 
+// Keccak-256 (Ethereum-style, padding 0x01). Reuses the same context type and
+// permutation as the SHA-3 family but is NOT FIPS-approved: it does not
+// update the FIPS service indicator. Streaming functions are exposed only
+// internally; public consumers reach Keccak-256 via |EVP_keccak_256|.
+//
```
manastasova: nit: extra comment line here

tool/speed.cc
```diff
@@ -2962,6 +2962,9 @@ bool Speed(const std::vector<std::string> &args) {
        !SpeedHash(EVP_sha3_384(), "SHA3-384", selected) ||
        !SpeedHash(EVP_sha3_512(), "SHA3-512", selected) ||
 #endif
+#if defined(OPENSSL_IS_AWSLC)
```
manastasova: Is it not compatible with openssl?

util/gen_keccak256_kat.py
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env python3
```
manastasova, line 1: Missing license header.

util/gen_keccak256_kat.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def main():
+    out_dir = sys.argv[1] if len(sys.argv) > 1 else (
+        os.path.join(os.path.dirname(__file__), '..', 'crypto', 'keccak',
+                     'testvectors'))
+    out_dir = os.path.normpath(out_dir)
+    os.makedirs(out_dir, exist_ok=True)
+
+    # Sanity check pycryptodome's output against the well-known Ethereum
```
manastasova, line 78: nice :)

sources.cmake
```diff
@@ -116,6 +116,8 @@ set(
   crypto/fipsmodule/sha/testvectors/SHA3_384ShortMsg.txt
   crypto/fipsmodule/sha/testvectors/SHA3_512LongMsg.txt
   crypto/fipsmodule/sha/testvectors/SHA3_512ShortMsg.txt
+  crypto/keccak/testvectors/KECCAK_256ShortMsg.txt
```
manastasova: Alphabetical order would be better

manastasova (commented):
Really clean PR! My only major comment is around the decision to add Keccak256 into the FIPS module. Would it be possible to integrate it outside of the module instead, similar to how blake2 is handled, while reusing the FIPS 202 functions from the FIPS module? This would also mean relocating the test vectors.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
fabrice102: fixed in second commit

crypto/digest_extra/digest_extra.c
fabrice102: We need this entry to be able to use `EVP_get_digestbyname("KECCAK-256")`, which is necessary to make it compatible with OpenSSL 3. OpenSSL 3 does not support `EVP_keccak_256()`.

crypto/fipsmodule/sha/internal.h
fabrice102: fixed in second commit

tool/speed.cc
fabrice102: OpenSSL 3 does not support `EVP_keccak_256()` which this code uses.

util/gen_keccak256_kat.py
fabrice102: fixed in second commit

sources.cmake
fabrice102: fixed in second commit

fabrice102:
Thank you for the detailed review @manastasova! I've added a new commit to address your comments.

Re:

> My only major comment is around the decision to add Keccak256 into the FIPS module. Would it be possible to integrate it outside of the module instead, similar to how blake2 is handled, while reusing the FIPS 202 functions from the FIPS module?

We would still need to make 2 changes to the FIPS module:
1. Export the FIPS 202 functions (they are currently not exported).
2. Modify the `FIPS202_Init` function to allow for the Keccak padding: https://github.com/fabrice102/aws-lc/blob/e1c7d24fc99a986b5881212ced67da70ad9c92ff/crypto/fipsmodule/sha/sha3.c#L119-L124

Let me know if you prefer doing it that way.

manastasova:
Re:

> We would still need to make 2 changes to the FIPS module:
> 
>     Export the FIPS 202 functions (they are currently not exported).
>     Modify the FIPS202_Init function to allow for the Keccak padding: https://github.com/fabrice102/aws-lc/blob/e1c7d24fc99a986b5881212ced67da70ad9c92ff/crypto/fipsmodule/sha/sha3.c#L119-L124
> 
> Let me know if you prefer doing it that way.

These changes sound reasonable.

fabrice102:
Pushed the refactor implementing the approved direction (move outside the FIPS module, mirror blake2). Summary of what changed since the last review:

**Moved Keccak-256 out of the FIPS module** — it now lives in a new `crypto/keccak/` (`keccak.c` + `internal.h`), reusing the module's audited Keccak-f[1600] permutation and FIPS 202 buffering rather than duplicating them, exactly as discussed.

**The two FIPS-module changes we agreed on:**
- `FIPS202_Reset/Update/Finalize` are given external linkage (de-`static`'d, declared in `sha/internal.h`) so the out-of-module code can reuse them. They are **not** `OPENSSL_EXPORT`ed, so nothing new leaves `libcrypto`.
- `FIPS202_Init` is **reverted** — it no longer accepts the Keccak pad and is unchanged from mainline. Instead, `Keccak256_Init` sets up the 0x01 padding itself. So the module's algorithmic code is now byte-for-byte identical to mainline.

**Public API renamed `EVP_keccak_256` → `EVP_keccak256`** (no underscore) for consistency with `EVP_shake256` / `EVP_blake2b256` and the `keccak256` spelling used by Ethereum.

**Tests moved** to a dedicated `crypto/keccak/keccak256_test.cc` (like `blake2_test.cc`); `digest_test.cc` keeps only the generic vector row + getter assertions, matching how blake2 is handled. The service-indicator test row was removed since Keccak-256 is no longer a module function.

**Bumped `AWSLC_API_VERSION` to 36** so consumers can feature-detect `EVP_keccak256` (v5.2.0 shipped 35 without it); the `bssl speed` gate uses `>= 36`.

**Name registration:** since Keccak-256 has no OID/NID, `"KECCAK-256"`/`"keccak-256"` are now string literals in the digest table rather than `nid.h` macros (`nid.h` is generated from `objects.txt` and can't carry an OID-less entry) — so `objects.txt`/`nid.h` stay untouched and the generated-file CI check passes.

crypto/keccak/keccak.c
```diff
@@ -0,0 +1,79 @@
...
+
+#include <assert.h>
+
+#include "../internal.h"
+
+// Keccak-256 (Ethereum-style, original 0x01 padding). NOT FIPS-approved.
+//
+// This reuses the FIPS module's FIPS 202 buffering primitives and Keccak-f[1600]
+// permutation but initialises the context with the original Keccak padding byte
+// (|KECCAK256_PAD_CHAR|) instead of a FIPS 202 one. Because the module's
```
manastasova: nit: 
```suggestion
// (|KECCAK256_PAD_CHAR|) instead of the FIPS 202 ones. Because the module's
```

crypto/keccak/keccak.c
```diff
@@ -0,0 +1,79 @@
...
+// |FIPS202_Init| deliberately rejects non-FIPS-202 padding, we set up the
+// context here rather than calling it.
+
+int Keccak256_Init(KECCAK1600_CTX *ctx) {
+  if (ctx == NULL) {
+    return 0;
+  }
+
+  const size_t block_size = KECCAK256_CBLOCK;
```
manastasova: Do we need `const size_t block_size`?

crypto/keccak/keccak.c
```diff
@@ -0,0 +1,79 @@
...
+  }
+  return FIPS202_Update(ctx, data, len);
+}
+
+int Keccak256_Final(uint8_t out[KECCAK256_DIGEST_LENGTH], KECCAK1600_CTX *ctx) {
+  if (out == NULL || ctx == NULL) {
+    return 0;
+  }
+  // This function must be paired with |Keccak256_Init|, which sets |md_size| to
+  // |KECCAK256_DIGEST_LENGTH|. Unlike SHA-3/SHAKE, Keccak-256 is a fixed-length
```
manastasova:
```suggestion
  // |KECCAK256_DIGEST_LENGTH|. Unlike SHAKE, Keccak-256 is a fixed-length
```

crypto/keccak/keccak.c
```diff
@@ -0,0 +1,79 @@
...
+  // hash, so |md_size| is never legitimately zero here.
```
manastasova: What if `Keccak256_Init` was (accidentally) skipped -- assert will only catch it in debug more. I would rather keep an `if` instead of `assert`.

crypto/keccak/keccak256_test.cc
```diff
@@ -0,0 +1,85 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
manastasova, line 1: As these are now new functions, it'd be worth adding negative tests to match the coverage SHA-3 already has -- DoubleFinalFails, UpdateAfterFinalFail, etc.

crypto/keccak/keccak.c
fabrice102: fixed

crypto/keccak/keccak.c
fabrice102: removed, and used a static assert

crypto/keccak/keccak.c
fabrice102: fixed as part of a larger change, see below

crypto/keccak/keccak.c
fabrice102: Done. Now, it follows what SHA-3 does and return success if md_size=0. Ideally, we should return failure, but this would diverge from SHA-3.

The same issue occurs in the Update function. Here, we return failure. SHA-3 currently just hangs (infinite loop). We can make a separate PR to fix that in SHA-3, but I did not want to add non-Keccak changes to this PR.

crypto/keccak/keccak256_test.cc
fabrice102: Done

manastasova (dismissed):

crypto/fipsmodule/sha/internal.h
```diff
@@ -440,6 +440,32 @@ OPENSSL_EXPORT uint8_t *SHAKE128(const uint8_t *data, const size_t in_len,
...
+
+/*
+ * FIPS202 APIs manage the internal input/output buffer on top of the Keccak1600
+ * API layer. They are the shared padding-and-buffering primitives underneath the
+ * SHA3 and SHAKE APIs below. They are also reused, outside the FIPS module, by
+ * the (non-FIPS) Keccak-256 implementation in crypto/keccak/keccak.c, which
+ * sets up the context with the original Keccak padding byte rather than a FIPS
+ * 202 one. FIPS202_Init deliberately accepts only FIPS 202 padding characters and
+ * is therefore kept private to this module.
+ */
```
samuel40791765: Could we also give better function names to the `FIPS202_*` family? It's understandable within the `SHA3/Keccak` context, but my concern is that this becomes a bit of tribal knowledge, now that these functions are being shared and called within other places in the library.

Minor Nit: would prefer this piece to match our coding style `//`.

crypto/fipsmodule/sha/internal.h
```diff
@@ -440,6 +440,32 @@ OPENSSL_EXPORT uint8_t *SHAKE128(const uint8_t *data, const size_t in_len,
...
 OPENSSL_EXPORT uint8_t *SHAKE256(const uint8_t *data, const size_t in_len,
                                  uint8_t *out, size_t out_len);
+
+/*
+ * FIPS202 APIs manage the internal input/output buffer on top of the Keccak1600
+ * API layer. They are the shared padding-and-buffering primitives underneath the
+ * SHA3 and SHAKE APIs below. They are also reused, outside the FIPS module, by
+ * the (non-FIPS) Keccak-256 implementation in crypto/keccak/keccak.c, which
+ * sets up the context with the original Keccak padding byte rather than a FIPS
+ * 202 one. FIPS202_Init deliberately accepts only FIPS 202 padding characters and
```
samuel40791765:
```suggestion
 * 202 one. |FIPS202_Init| deliberately accepts only FIPS 202 padding characters and
```

samuel40791765 (commented):
The 237 vectors in `KECCAK256ShortMsg.txt` and `KECCAK256LongMsg.txt` all come from the same source (`util/gen_keccak256_kat.py`). Is there another source of one more independently published, citable vector set that we can layer on top?

It doesn't have to be much, just looking for more possible coverage for a new algorithm being introduced.

fabrice102:
Thanks for the review! Regarding test vectors, we're also explicitly testing on the Ethereum test vectors:
https://github.com/aws/aws-lc/pull/3245/changes#diff-a486879d22cdbac91336850b85c6f1c88b0725a1115b3d2b7674326925c17225R229-R237

crypto/fipsmodule/sha/internal.h
fabrice102: fixed in third commit

crypto/fipsmodule/sha/internal.h
fabrice102: fixed in third commit

samuel40791765 (approved):

manastasova (approved):

Issue #2921, opened 2025-12-24
Is TLSv1.2 and TLSv1.3 approved in any versions of AWS-LC-FIPS?


xnox:
### Security issue notifications

This is not a security issue, but a compliance one.

### Problem:

ACVP testing has 3 TLS related algorithms that one can test:
- KDF TLS (for TLSv1.0/v1.1/v1.2 without Extended Master Secret Support)
- TLS v1.2 KDF RFC7627 (TLSv1.2 with Extended Master Secret Support)
- TLS v1.3 KDF

Searching all Amazon vendor modules with algorithms set to "TLS v1.2 KDF RFC7627" and "TLS v1.3 KDF" brings up that only Amazon Linux OpenSSL has approved mode for modern TLS.

https://csrc.nist.gov/projects/cryptographic-algorithm-validation-program/validation-search?searchMode=implementation&vendor=Amazon&productType=-1&algorithm=164%2C165&ipp=25

Noticeably no tests for AWS-LC.

Separately if one instead searches for KDF TLS, lots of Amazon cryptographic modules show up, however all of them are likely capped at deprecated TLS v1.2 without Extended Master Secret support - or simply use TLSv1.2 with EMS / TLSv1.3 in unapproved mode and ignore service indicators.

Also security policies either do not mention TLSv1.3 at all in neither approved or unapproved service tables; or do make TLSv1.3 scenario references.

### Solution:

Is TLSv1.3 vendor affirmed in AWS-LC? If yes, can the future security policies be updated to state so?

Can you please add "TLS v1.2 KDF RFC7627" and "TLS v1.3 KDF" in future ACVP tests for AWS-LC?

### Requirements / Acceptance Criteria:

Please see FIPS 140-3 IG 2.4.B, Resolution, Item 7 w.r.t. TLS v1.3
- https://csrc.nist.gov/Projects/cryptographic-module-validation-program/fips-140-3-ig-announcements
- https://csrc.nist.gov/csrc/media/Projects/cryptographic-module-validation-program/documents/fips%20140-3/FIPS%20140-3%20IG.pdf 

### Out of scope:

I actually don't know if anybody actually cares. As clearly lots of people are using TLSv1.3 and don't check service indicators and likely no auditor ever looks at this.

samuel40791765:
Thanks for reaching out! Apologies for the delayed response due to the holidays.

[...]

PR #3247, opened 2026-05-12
Move TLS 1.3 KDF into the FIPS module and wire up ACVP

justsmth:
### Issues
Addresses:
* #2921 

### Description of changes:

Moves TLS 1.3 HKDF-Expand-Label (RFC 8446 §7.1) into the FIPS module and wires up the ACVP modulewrapper so we advertise `TLS-v1.3` as a validated composite algorithm.

1. **`CRYPTO_tls13_hkdf_expand_label`** in `crypto/fipsmodule/tls/kdf.c` — builds the `HkdfLabel` structure, calls `HKDF_expand`, handles the FIPS service indicator (approved for SHA2-256 / SHA2-384).

2. **Delegates both consumers** — `ssl/tls13_enc.cc::hkdf_expand_label` becomes a thin wrapper; the modulewrapper's `TLS13_HKDFExpandLabel` handler (wire command `HKDFExpandLabel/<HASH>`) routes through the same code path.

3. **ACVP wiring** — registers `TLS-v1.3` / `RFC8446` in `getConfig`, implements `HKDFExtract/` and `HKDFExpandLabel/` commands, adds 250 NIST test vectors from the ACVP-Server repo.

4. **FIPS self-test** — TLS13-KDF KAT in `self_check.c` plus matching `break-kat.go` entry.

### Call-outs:

- **Public API surface.** `CRYPTO_tls13_hkdf_expand_label` is declared in `include/openssl/kdf.h` alongside `CRYPTO_tls1_prf` for consistency. BoringSSL keeps it private; happy to switch if reviewers prefer tighter scope.

- **Ported from BoringSSL.** The `HkdfLabel`/`CBB` construction in `CRYPTO_tls13_hkdf_expand_label`, the ACVP handlers, the `getConfig` block, and the KAT vectors (commit `480344d4fa`) are all ported from BoringSSL. The FIPS service-indicator wiring is AWS-LC-specific. Attribution comments are in-line in the diff.

- **Helper naming.** The modulewrapper handlers are `TLS13_HKDFExtract` / `TLS13_HKDFExpandLabel` (diverging from BoringSSL's names) to disambiguate from the pre-existing generic `HKDF` / `HKDF_expand` helpers in the same TU.

### Testing:

All existing `crypto_test`, `ssl_test`, and `acvp_tests` pass. New coverage: five `TLS13KDF_ServiceIndicatorTest` cases and the TLS13-KDF self-test KAT.


include/openssl/kdf.h
```diff
@@ -25,6 +25,22 @@ OPENSSL_EXPORT int CRYPTO_tls1_prf(const EVP_MD *digest,
                                    const uint8_t *seed1, size_t seed1_len,
                                    const uint8_t *seed2, size_t seed2_len);
 
+// CRYPTO_tls13_hkdf_expand_label computes the TLS 1.3 HKDF-Expand-Label
+// function as defined in RFC 8446, Section 7.1. It derives |out_len| bytes of
+// output into |out| using |digest| as the underlying HMAC hash, |secret| as
```
nebeid, line 30: The doc says `|label_len| must be between 1 and 249 inclusive` but the implementation intentionally allows `label_len == 0` (commit 3 explains why: `SSL_export_keying_material` callers pass empty labels). The header should match the actual contract — either drop the lower bound from the doc or document that 0 is accepted as a special case.

crypto/fipsmodule/tls/kdf.c
```diff
@@ -127,3 +129,66 @@ int CRYPTO_tls1_prf(const EVP_MD *digest,
...
+  // Construct the HkdfLabel structure per RFC 8446 Section 7.1:
+  //   struct {
+  //       uint16 length = Length;
+  //       opaque label<7..255> = "tls13 " + Label;
+  //       opaque context<0..255> = Context;
+  //   };
+  // The CBB_add_u16 / CBB_add_u8_length_prefixed calls implicitly enforce the
+  // RFC-mandated upper bounds: CBB_add_u16 fails if |out_len| does not fit in a
+  // uint16_t, and the u8-length-prefixed children fail if the "tls13 "-prefixed
+  // label or |hash| exceed 255 bytes. The RFC's lower bound on the label field
```
nebeid: `out_len` is `size_t` but `CBB_add_u16` takes `uint16_t` — if `out_len > 65535`, the value is silently truncated by the implicit narrowing conversion before `CBB_add_u16` sees it. The comment above states "CBB_add_u16 fails if |out_len| does not fit in a uint16_t" but the truncation occurs at the call site, not inside CBB. Consider adding an explicit bounds check (like the modulewrapper does with `if (out_len > 0xFFFF)`) or at minimum updating the comment.

include/openssl/kdf.h
justsmth: Agreed. Addressed in latest commit.

crypto/fipsmodule/tls/kdf.c
justsmth: Agreed. Addressed in latest commit.

crypto/fipsmodule/service_indicator/service_indicator_test.cc
```diff
@@ -3397,6 +3397,42 @@ TEST_P(KDF_ServiceIndicatorTest, TLSKDF) {
...
+
+INSTANTIATE_TEST_SUITE_P(All, TLS13KDF_ServiceIndicatorTest,
+                         testing::ValuesIn(kTLS13KDFTestVectors));
+
+TEST_P(TLS13KDF_ServiceIndicatorTest, HKDFExpandLabel) {
+  const TLS13KDFTestVector &test = GetParam();
+
+  static const uint8_t kLabel[] = "c e traffic";
+  static const uint8_t kHash[32] = {0};
+  FIPSStatus approved = AWSLC_NOT_APPROVED;
```
samuel40791765: Minor nit, but `kHash[32] = {0}` is fixed at 32 bytes even when the test exercises SHA-384 (output is 48 bytes). It doesn't really effect the correctness of the implementation though, just something that felt off.

include/openssl/kdf.h
```diff
@@ -25,6 +25,24 @@ OPENSSL_EXPORT int CRYPTO_tls1_prf(const EVP_MD *digest,
...
+// the HKDF pseudorandom key, |label| (which is prefixed with "tls13 "
+// internally) and |hash| (the context/transcript hash). |out_len| must be at
+// most 65535, |label_len| must be at most 249 so that the "tls13 "-prefixed
+// label fits in the RFC 8446 |opaque label<...255>| bound, and |hash_len|
+// must be at most 255. An empty |label| (|label_len == 0|) is permitted to
```
samuel40791765: Should we introduce a check to bail early and enforce this?
```
 if (label_len > 249 || hash_len > 255) {
   OPENSSL_PUT_ERROR(CRYPTO, ERR_R_OVERFLOW);
   return 0;
}
```

nebeid (approved):

samuel40791765 (approved):

Issue #3188, opened 2026-04-24
`cpu_arm_linux.c` and `cpu_aarch64_linux.c` unconditionally require `getauxval` / `<sys/auxv.h>`


justsmth:
### Problem:

`crypto/fipsmodule/cpucap/cpu_arm_linux.c` and `crypto/fipsmodule/cpucap/cpu_aarch64_linux.c` unconditionally `#include <sys/auxv.h>` and call `getauxval`. These are glibc extensions also provided by musl and recent uclibc-ng, but not by older uclibc/uclibc-ng, which causes a hard build failure:

```text
.../cpu_arm_linux.c:10:10: fatal error: sys/auxv.h: No such file or directory
   10 | #include <sys/auxv.h>
      |          ^~~~~~~~~~~~
```

Reported downstream for a uclibc-based `armv7-unknown-linux-uclibceabihf` target: https://github.com/aws/aws-lc-rs/issues/1108

Users can work around this by defining `OPENSSL_STATIC_ARMCAP`, but that isn't easily discoverable and users hit the build failure first.

There's already precedent for gating `getauxval` behind a feature macro in `crypto/rand_extra/urandom.c` (`OPENSSL_HAS_GETAUXVAL`), but it's keyed only off `__GLIBC_PREREQ` and so doesn't help non-glibc libcs that might lack `getauxval`. Generalizing that pattern (e.g. via a CMake build-time probe) and applying it to the ARM/AArch64 cpucap sources with a fallback to `/proc/self/auxv` or `/proc/cpuinfo` when `getauxval` is unavailable would let these targets build out of the box while preserving runtime detection where possible.

### Related Issues:

* https://github.com/aws/aws-lc-rs/issues/1108

PR #3250, opened 2026-05-13
Add `getauxval` availability detection with `/proc/self/auxv` fallback for uclibc targets

justsmth:
### Issues:
Addresses:
* #3188
* aws/aws-lc-rs#1108

### Description of changes:

On Linux targets using older uclibc/uclibc-ng (pre-1.0.43), `<sys/auxv.h>` and `getauxval` are not available, causing hard build failures in the ARM, AArch64, and PPC64LE CPU capability detection code and in the urandom entropy path.

This introduces a shared header (`cpu_getauxval_linux.h`) that detects `getauxval` availability at the preprocessor level using `__has_include`, with library-specific heuristics as a fallback for compilers that lack it. When the header is absent, a fallback reads `/proc/self/auxv` directly. After including the header, `OPENSSL_HAS_GETAUXVAL` is unconditionally defined so call sites don't need to care which path was taken.

The shared helper replaces the previous `<sys/auxv.h>` includes in `cpu_arm_linux.c`, `cpu_aarch64_linux.c`, and `cpu_ppc64le.c`, and supersedes the old glibc-only detection in `urandom.c`.

### Call-outs:

The fallback `getauxval` is a `static` function defined in a header. This means each translation unit that includes it (when the fallback is active) gets its own copy. This is intentional — it avoids needing to wire a new `.c` file into the build, and the fallback path only compiles on uclibc targets where `<sys/auxv.h>` is genuinely missing.

The `#if defined(AT_HWCAP2)` guard in `cpu_ppc64le.c` is removed because the header now unconditionally defines `AT_HWCAP2` in the fallback path, and it was always defined via `<sys/auxv.h>` in the non-fallback path.

### Testing:

New CI job (`armv7-uclibc-build-test`) builds and runs tests against Bootlin's armv7-eabihf uclibc-ng 1.0.42 toolchain under QEMU. This toolchain predates uclibc-ng's `<sys/auxv.h>` (added in 1.0.43), so `__has_include` evaluates false and the `/proc/self/auxv` fallback is exercised end-to-end without any artificial force-flags.

### DIFF
--- a/.github/workflows/cross-test.yml
+++ b/.github/workflows/cross-test.yml
@@ -54,7 +54,7 @@ jobs:
           sudo apt-get -y install qemu-user qemu-user-binfmt
       - uses: actions/checkout@v6
       - name: PPC64LE Build/Test
-        run: tests/ci/run_cross_tests.sh ppc64le powerpc64le-unknown-linux-gnu "-DCMAKE_BUILD_TYPE=Release" "-DCMAKE_BUILD_TYPE=Release -DFIPS=1 -DBUILD_SHARED_LIBS=1"
+        run: tests/ci/run_cross_tests.sh ppc64le powerpc64le-unknown-linux-gnu "-DCMAKE_BUILD_TYPE=Release" "-DCMAKE_BUILD_TYPE=Release -DFIPS=1 -DBUILD_SHARED_LIBS=1" "-DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS=-DOPENSSL_GETAUXVAL_FORCE_PROC_FALLBACK"
   riscv64-non-fips-build-test:
     if: github.repository_owner == 'aws'
     runs-on: ubuntu-24.04
--- a/.github/workflows/cross-test.yml
+++ b/.github/workflows/cross-test.yml
@@ -84,6 +84,35 @@ jobs:
       - uses: actions/checkout@v6
       - name: armv6 Build/Test
         run: tests/ci/run_cross_tests.sh armv6 armv6-unknown-linux-gnueabi "-DCMAKE_BUILD_TYPE=Release"
+  armv7-uclibc-build-test:
+    # Bootlin 2022.08-1 ships uclibc-ng 1.0.42, which predates uclibc-ng's
+    # own <sys/auxv.h> (added in 1.0.43). On this sysroot,
+    # __has_include(<sys/auxv.h>) in cpu_getauxval_linux.h evaluates false,
+    # so natural detection lands on the /proc/self/auxv fallback -- exactly
+    # the scenario from https://github.com/aws/aws-lc/issues/3188. A single
+    # job thus covers both the non-glibc detection path and the fallback
+    # code path end-to-end. If this toolchain is ever bumped to a release
+    # that does ship <sys/auxv.h>, the "INFO:" line in
+    # run_cross_tests_bootlin.sh will make the coverage loss visible in the
+    # CI log.
+    if: github.repository_owner == 'aws'
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Install qemu
+        run: |
+          sudo apt-get update -o Acquire::Languages=none -o Acquire::Translation=none
+          sudo apt-get -y install qemu-user qemu-user-binfmt
+      - uses: actions/checkout@v6
+      - name: armv7 uclibc Build/Test
+        run: |
+          tests/ci/run_cross_tests_bootlin.sh \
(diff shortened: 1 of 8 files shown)


crypto/fipsmodule/cpucap/cpu_getauxval_linux.h
```diff
@@ -0,0 +1,137 @@
...
+// |type|, or 0 if no such entry exists or /proc/self/auxv is unavailable.
+//
+// Note: unlike glibc >= 2.19, this implementation does not set errno to
+// ENOENT when |type| is not present. No caller in AWS-LC inspects errno
+// after getauxval, so this difference is not observable.
+//
+// OPENSSL_UNUSED guards against -Wunused-function in translation units that
+// include this header but do not call getauxval (none today, but cheap
+// insurance against future refactors).
+static OPENSSL_UNUSED unsigned long getauxval(unsigned long type) {
```
geedo0, line 79: > This is intentional — it avoids needing to wire a new .c file into the build...

For my understanding, why is this something we care to avoid? Just seems odd to me to have a function defined in this header.

.github/workflows/cross-test.yml
```diff
@@ -84,6 +84,35 @@ jobs:
...
+    # that does ship <sys/auxv.h>, the "INFO:" line in
+    # run_cross_tests_bootlin.sh will make the coverage loss visible in the
+    # CI log.
+    if: github.repository_owner == 'aws'
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Install qemu
+        run: |
+          sudo apt-get update -o Acquire::Languages=none -o Acquire::Translation=none
+          sudo apt-get -y install qemu-user qemu-user-binfmt
```
geedo0, line 104: Went down a rabbit hole looking into how `/proc/self/auxv` gets handled by qemu. Looks like it traps and emulates a bunch of files in procfs to make things look like the guest ISA. Really cool.

geedo0 (dismissed):

crypto/fipsmodule/cpucap/cpu_getauxval_linux.h
justsmth: Adding a "*.c" source file is just a minor inconvenience -- we'd also have to modify CMakeLists.txt and then remember to include the source file for aws-lc-sys's `cc` builder. Since this function will only be defined in the very rare case that a Linux host build environment doesn't have a `sys/auxv.h` header, it seems reasonable to define it as a `static` function only for the compilation units (object files) its needed in.

nebeid (commented):
Is `OPENSSL_GETAUXVAL_FORCE_PROC_FALLBACK` exercised anywhere in CI? I see it documented but not used in the new workflow. If it's purely for local developer testing that's fine, but it could also be a cheap way to regression-test the fallback on glibc runners without needing a separate toolchain.

crypto/fipsmodule/cpucap/cpu_ppc64le.c
```diff
@@ -57,14 +65,10 @@ static void handle_cpu_env(unsigned long *out, const char *in) {
 extern uint8_t OPENSSL_cpucap_initialized;
 
 void OPENSSL_cpuid_setup(void) {
-#if defined(AT_HWCAP2)
 #if defined(OPENSSL_LINUX)
```
nebeid, line 68: For the FreeBSD path (which still uses `#include <sys/auxv.h>` directly without going through the new header): does FreeBSD's `<sys/auxv.h>` define `AT_HWCAP2` on PPC64LE? If not, removing this guard would be a compile error on FreeBSD PPC64LE. Might be worth a `#if !defined(AT_HWCAP2)` fallback define here, or confirming FreeBSD coverage.

justsmth: Good catch — added `&& defined(AT_HWCAP2)` to the FreeBSD `#elif`, consistent with `cpu_arm_freebsd.c`.

justsmth:
> Is `OPENSSL_GETAUXVAL_FORCE_PROC_FALLBACK` exercised anywhere in CI? I see it documented but not used in the new workflow. If it's purely for local developer testing that's fine, but it could also be a cheap way to regression-test the fallback on glibc runners without needing a separate toolchain.

Added it as a third build pass on the `ppc64le-build-test` job. Gives us 64-bit fallback coverage on glibc alongside the natural 32-bit coverage from the uclibc armv7 job.

nebeid (approved):

geedo0 (approved):

Issue #3207, opened 2026-05-01
Adding MinGW support for the FIPS build


Will-Low:
We are looking to use the FIPS feature of the `aws-lc-rs` Rust crate. However, our project cross compiles to the the `x86_64-pc-windows-gnu` Rust target, which utilizes MinGW. The compatibility matrix shows that FIPS is not currently supported for this build target and I confirmed that trying to perform this build does fail.

Is there an appetite to support FIPS MinGW builds? If so, I may be able to submit a PR for the relevant `aws-lc` changes.

justsmth:
Thanks for reaching out. You're correct that FIPS is not currently supported for the `x86_64-pc-windows-gnu` target.

For Windows FIPS builds, only the following targets are supported today:
- `x86_64-pc-windows-msvc`
- `aarch64-pc-windows-msvc`

The primary technical challenge with MinGW is that our FIPS integrity check mechanism relies on MSVC-specific pragmas (`code_seg`, `data_seg`, `const_seg`) to place code and data into custom PE sections with specific ordering. MinGW GCC does not support these pragmas, so supporting this target would require an alternative approach to section placement.

We don't currently have plans to add FIPS support for the `*-windows-gnu` targets, but we would be happy to review a PR that adds this support if you're interested in working through the technical constraints.

PR #3263, opened 2026-05-21
Add FIPS shared library support for MinGW

Will-Low:
### Issues:
Addressed: #3207, aws/aws-lc-rs#534

### Description of changes:
AWS-LC's FIPS build does not currently support cross-compilation targeting Windows using MinGW. This PR adds that support: MinGW-specific BCM boundary markers, PE/COFF support in the integrity hash injection, statically linked libwinpthread in the FIPS DLL, Win32 threading on MinGW, and a `cross-mingw-fips` CI job (positive + negative integrity tests).

### Call-outs:
The most important thing to review is how the runtime integrity check handles relocations. MinGW is the only supported FIPS target whose hashed region contains base relocations: GCC emits `.refptr` indirection cells and genuine imports (e.g. the CRT's `free`) into `.rdata` inside the integrity boundary. Every other toolchain keeps the hashed window relocation-free by construction (ELF/Mach-O segregate relocated pointers into separate sections via linker scripts + delocate; MSVC emits no `.refptr` cells), so they hash the region directly. PE linkers don't support linker scripts, so we can't use that approach here.

Instead, `hmac_update_module_region` reconstructs the preferred-base image at runtime: it walks the PE base relocation table and subtracts the ASLR load delta from each DIR64 relocation inside the hashed window before hashing, recovering the bytes `inject_hash.go` measured on disk. The load delta is `actual_base − preferred_base`, where `preferred_base` is read from the injected `BORINGSSL_bcm_preferred_base` constant — not the in-memory PE header, since the loader overwrites `OptionalHeader.ImageBase` with the actual load address. The delta is a single module-wide constant derived independently of the hashed bytes, so coverage stays complete and any genuine modification (including to a relocated pointer) is still detected; the check fails closed on anything unexpected (non-DIR64 reloc in the window, out-of-order relocs, malformed blocks).

This is a deliberate departure from AWS-LC's usual "relocation-free hashed region" paradigm and is worth a FIPS/lab conversation before we commit to it — flagging so reviewers with validation context can weigh in on whether runtime normalization is acceptable versus pushing the `.refptr`/import cells out of the boundary.

Two supporting pieces are needed to make this correct and safe: GCC hot/cold partitioning is disabled for FIPS_SHARED MinGW builds (so split-out `.text.unlikely`/`.text.startup` sections don't land outside the markers after the plain `ld -r` partial link), and `check_fips_mingw_sections.cmake` fails the build if any split-out section remains — the MinGW analogue of the ELF `/DISCARD/` net.

### Testing:
`cross-mingw-fips` CI job runs the FIPS tests under Wine. The negative test (`break-hash.go -mingw`) corrupts a byte inside the module and asserts the integrity check fails, since the positive test alone can't distinguish a working check from a no-op (injection and the runtime check measure the same window). Also cross-compiled a FIPS-enabled Rust app for `x86_64-pc-windows-gnu` from a Linux/aarch64 host and confirmed it starts on Windows Server 2022.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1333,8 +1333,6 @@ endif()
 if(NOT CMAKE_SYSTEM_NAME MATCHES "^(Generic|Android|Emscripten)$")
   find_package(Threads REQUIRED)
   if(MINGW)
-    # Statically link libwinpthread so the FIPS DLL has no runtime dependency
-    # on libwinpthread-1.dll, which is not present on Windows without MinGW.
     set_property(TARGET Threads::Threads PROPERTY
       INTERFACE_LINK_LIBRARIES "-Wl,-Bstatic,-lwinpthread;-Wl,-Bdynamic")
   endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1034,9 +1034,11 @@ if(FIPS)
 endif()
 
 # -ffunction-sections / -fdata-sections are incompatible with FIPS_SHARED.
-# The Android CMake files set these flags. Although the MinGW build does not
-# currently enable them, defensively disabling them here to keep the FIPS
-# integrity check robust if added in the future.
+# On ELF, gcc_fips_shared.lds already handles this: it gathers known split
+# sections into .text and /DISCARD/s anything unexpected. Android and MinGW
+# lack that linker-script safety net (Android's NDK toolchain injects these
+# flags externally; MinGW/PE has no linker script support), so we override
+# at the compiler level.
 if(FIPS_SHARED AND (ANDROID OR MINGW))
   set(CMAKE_C_FLAGS
       "${CMAKE_C_FLAGS} -fno-function-sections -fno-data-sections")
@@ -1347,17 +1349,10 @@ endif()
 if(NOT CMAKE_SYSTEM_NAME MATCHES "^(Generic|Android|Emscripten)$")
   find_package(Threads REQUIRED)
   if(MINGW AND FIPS)
-    # If CMake's MinGW Threads::Threads target resolves to winpthread, keep it
-    # statically linked so the FIPS build products do not acquire an external
-    # libwinpthread DLL dependency at runtime.
-    #
-    # NOTE: crypto links Threads::Threads PUBLIC, so mutating the imported
-    # target's interface here statically links winpthread into *every* artifact
-    # in this build that pulls in threads (libssl.dll, bssl.exe, and the test
-    # binaries), not just the crypto DLL. That is intentional for the integrity
-    # boundary's sake and otherwise harmless (just some duplicated code), but it
-    # is broader than "the FIPS DLL only". This mutation is build-local and is
-    # not propagated to downstream consumers of the installed crypto target.
+    # Statically link winpthread so the FIPS module has no DLL dependency on it.
+    # This overwrites (not appends) INTERFACE_LINK_LIBRARIES because
+    # find_package(Threads) already set it to -lpthread (dynamic); appending
+    # would leave the dynamic entry on the link line.
     set_property(TARGET Threads::Threads PROPERTY
       INTERFACE_LINK_LIBRARIES "-Wl,-Bstatic,-lwinpthread;-Wl,-Bdynamic")
   endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1347,9 +1347,9 @@ endif()
 if(NOT CMAKE_SYSTEM_NAME MATCHES "^(Generic|Android|Emscripten)$")
   find_package(Threads REQUIRED)
   if(MINGW AND FIPS)
-    # The FIPS shared module on MinGW must contain its own copy of the
-    # winpthread implementation so that the integrity-checked DLL does not
-    # depend on an external libwinpthread DLL at runtime.
+    # If CMake's MinGW Threads::Threads target resolves to winpthread, keep it
+    # statically linked so the FIPS build products do not acquire an external
+    # libwinpthread DLL dependency at runtime.
     #
     # NOTE: crypto links Threads::Threads PUBLIC, so mutating the imported
     # target's interface here statically links winpthread into *every* artifact
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1332,6 +1332,12 @@ endif()
 # via Web Workers and SharedArrayBuffer, configured through compiler flags.
 if(NOT CMAKE_SYSTEM_NAME MATCHES "^(Generic|Android|Emscripten)$")
   find_package(Threads REQUIRED)
+  if(MINGW)
+    # Statically link libwinpthread so the FIPS DLL has no runtime dependency
+    # on libwinpthread-1.dll, which is not present on Windows without MinGW.
+    set_property(TARGET Threads::Threads PROPERTY
+      INTERFACE_LINK_LIBRARIES "-Wl,-Bstatic,-lwinpthread;-Wl,-Bdynamic")
+  endif()
   set(AWSLC_LINK_THREADS TRUE)
   # CMAKE_THREAD_LIBS_INIT contains the actual linker flags (e.g., -lpthread)
   # set by find_package(Threads). Use this for pkgconfig instead of the
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1033,9 +1033,11 @@ if(FIPS)
   add_definitions(-DBORINGSSL_FIPS)
 endif()
 
-# The Android CMake files set -ffunction-sections and -fdata-sections, which
-# is incompatible with FIPS_SHARED.
-if(FIPS_SHARED AND ANDROID)
+# -ffunction-sections / -fdata-sections are incompatible with FIPS_SHARED.
+# The Android CMake files set these flags. Although the MinGW build does not
+# currently enable them, defensively disabling them here to keep the FIPS
+# integrity check robust if added in the future.
+if(FIPS_SHARED AND (ANDROID OR MINGW))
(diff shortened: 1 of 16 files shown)


CMakeLists.txt
```diff
@@ -1311,6 +1311,10 @@ endif()
 # via Web Workers and SharedArrayBuffer, configured through compiler flags.
 if(NOT CMAKE_SYSTEM_NAME MATCHES "^(Generic|Android|Emscripten)$")
   find_package(Threads REQUIRED)
+  if(MINGW)
+    set_property(TARGET Threads::Threads PROPERTY
+      INTERFACE_LINK_LIBRARIES "-Wl,-Bstatic,-lwinpthread;-Wl,-Bdynamic")
```
justsmth: Limit this to FIPS builds:
```
  if(MINGW AND FIPS)
    # The FIPS shared module on MinGW must contain its own copy of the
    # winpthread implementation
``

File: CMakeLists.txt
Comment (justsmth) [inline CMakeLists.txt]: This should also be used for MINGW:
```
# -ffunction-sections / -fdata-sections are incompatible with FIPS_SHARED.
# The Android CMake files set these flags. Although the MinGW build does not
# currently enable them, defensively disabling them here to keep the FIPS 
# integrity check robust if added in the future.
if(FIPS_SHARED AND (ANDROID OR MINGW))
```

File: crypto/fipsmodule/CMakeLists.txt [hunk 2]
```diff
@@ -642,6 +642,28 @@ elseif(FIPS_SHARED)
...
+    # BORINGSSL_bcm_text_start/end bracket all BCM code.
+    set(BCM_NAME bcm.o)
+    add_custom_command(
+      OUTPUT fips_gnu_start.o
+      COMMAND ${CMAKE_C_COMPILER} -c ${CMAKE_CURRENT_SOURCE_DIR}/fips_shared_library_marker.c -DAWSLC_FIPS_SHARED_START -o fips_gnu_start.o
+      DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/fips_shared_library_marker.c
+    )
+    add_custom_command(
+      OUTPUT fips_gnu_end.o
+      COMMAND ${CMAKE_C_COMPILER} -c ${CMAKE_CURRENT_SOURCE_DIR}/fips_shared_library_marker.c -DAWSLC_FIPS_SHARED_END -o fips_gnu_end.o
```
justsmth: It would be more clear to name these `fips_mingw_start.o` and `fips_mingw_end.o`.

util/fipstools/inject_hash/inject_hash.go
```diff
@@ -297,7 +298,120 @@ func doAppleOS(objectBytes []byte) ([]byte, []byte, error) {
...
+	}
+
+	if textStart == nil || textEnd == nil {
+		return nil, nil, errors.New("could not find .text module boundaries in object")
+	}
+
+	if rodataStart != nil && rodataSection == nil {
+		return nil, nil, errors.New("rodata start marker found but no .rdata section present")
+	}
+
```
justsmth: NP: check for inconsistency like the others:
```
	if (rodataStart == nil) != (rodataSection == nil) {
		return nil, nil, errors.New("rodata start marker inconsistent with .rdata section presence")
	}
```

util/fipstools/inject_hash/inject_hash.go
```diff
@@ -297,7 +298,120 @@ func doAppleOS(objectBytes []byte) ([]byte, []byte, error) {
...
+		case textSectionIndex:
+		case rodataSectionIndex:
+			// rodataSectionIndex is 0 if no .rdata section was found,
+			// which would match undefined symbols (COFF section number 0) — skip those.
+			if rodataSection == nil {
+				continue
+			}
+		default:
+			continue
+		}
```
justsmth: NP:  
* Clarify comment.
* This looks a lot like a "fallthrough", but it's not. It could be more clear:
```
		sn := int(symbol.SectionNumber)
		if sn != textSectionIndex && (rodataSection == nil || sn != rodataSectionIndex) {
			continue
		}
```

crypto/fipsmodule/CMakeLists.txt
```diff
@@ -642,6 +642,28 @@ elseif(FIPS_SHARED)
...
+      DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/fips_shared_library_marker.c
+    )
+    add_custom_command(
+      OUTPUT ${BCM_NAME}
+      COMMAND ${CMAKE_LINKER} -r fips_gnu_start.o --whole-archive $<TARGET_FILE:bcm_library> --no-whole-archive fips_gnu_end.o -o ${BCM_NAME}
+      DEPENDS fips_gnu_start.o fips_gnu_end.o bcm_library
+      WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+    )
```
justsmth, line 858: We need to forward the `CMAKE_C_FLAGS` here:
```
    # These custom commands invoke the compiler directly (outside CMake's normal
    # target compilation), so we must forward CMAKE_C_FLAGS manually. This is
    # necessary for cross-compilation where the flags contain --target and
    # system include paths (e.g. clang targeting MinGW).
    separate_arguments(FIPS_MARKER_C_FLAGS NATIVE_COMMAND "${CMAKE_C_FLAGS}")
    add_custom_command(
      OUTPUT fips_gnu_start.o
      COMMAND ${CMAKE_C_COMPILER} ${FIPS_MARKER_C_FLAGS} -c ${CMAKE_CURRENT_SOURCE_DIR}/fips_shared_library_marker.c -DAWSLC_FIPS_SHARED_START -o fips_gnu_start.o
      DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/fips_shared_library_marker.c
    )
    add_custom_command(
      OUTPUT fips_gnu_end.o
      COMMAND ${CMAKE_C_COMPILER} ${FIPS_MARKER_C_FLAGS} -c ${CMAKE_CURRENT_SOURCE_DIR}/fips_shared_library_marker.c -DAWSLC_FIPS_SHARED_END -o fips_gnu_end.o
      DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/fips_shared_library_marker.c
    )
```

justsmth (commented):
We will also need to add CI coverage for this.

Will-Low:
@justsmth sorry for the delay. I was away from my computer for a few weeks. Is there anything outstanding that I can do here? Thank you for all of your efforts.

CMakeLists.txt
```diff
@@ -1033,15 +1033,29 @@ if(FIPS)
...
 endif()
 
-# The Android CMake files set -ffunction-sections and -fdata-sections, which
-# is incompatible with FIPS_SHARED.
-if(FIPS_SHARED AND ANDROID)
+# -ffunction-sections / -fdata-sections are incompatible with FIPS_SHARED.
+# The Android CMake files set these flags. Although the MinGW build does not
+# currently enable them, defensively disabling them here to keep the FIPS
+# integrity check robust if added in the future.
+if(FIPS_SHARED AND (ANDROID OR MINGW))
```
samuel40791765, line 1091: > Although the MinGW build does not currently enable them, defensively disabling them here to keep the FIPS integrity check robust if added in the future.

Wouldn't this logic apply to all platforms? I'm fine with the logic, I just think we need stronger reasoning behind it.

CMakeLists.txt
```diff
@@ -1332,6 +1346,21 @@ endif()
...
+    #
+    # NOTE: crypto links Threads::Threads PUBLIC, so mutating the imported
+    # target's interface here statically links winpthread into *every* artifact
+    # in this build that pulls in threads (libssl.dll, bssl.exe, and the test
+    # binaries), not just the crypto DLL. That is intentional for the integrity
+    # boundary's sake and otherwise harmless (just some duplicated code), but it
+    # is broader than "the FIPS DLL only". This mutation is build-local and is
+    # not propagated to downstream consumers of the installed crypto target.
+    set_property(TARGET Threads::Threads PROPERTY
+      INTERFACE_LINK_LIBRARIES "-Wl,-Bstatic,-lwinpthread;-Wl,-Bdynamic")
```
samuel40791765, line 1416: Can we use `APPEND` here? This seems to directly overwrite the existing properties. 
```suggestion
    set_property(TARGET Threads::Threads APPEND PROPERTY INTERFACE_LINK_LIBRARIES ."-Wl,-Bstatic,-lwinpthread;-Wl,-Bdynamic")
```

CMakeLists.txt
justsmth: On ELF targets, [`gcc_fips_shared.lds`](https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/gcc_fips_shared.lds) already handles this. It gathers known split sections (`.text.unlikely`, `.text.exit`, `.text.startup`) into `.text` between the integrity markers, and the `/DISCARD/` rule catches anything else (`*.text.*`, `*.data.*`). So the linker script is the protection there.

MinGW (PE/COFF) doesn't have linker script support. We use a plain `ld -r` partial link, so the compiler-level override is the only defense. (Android is included because the NDK toolchain files inject these flags externally and we can't rely on a linker script to clean up after them.)

I'll update the comment here.

CMakeLists.txt
justsmth: The overwrite is intentional here. `find_package(Threads)` on MinGW sets `INTERFACE_LINK_LIBRARIES` to `-lpthread` (linking winpthread dynamically). We want to *replace* that with the static-link recipe, not append alongside it. Otherwise the original dynamic `-lpthread` stays on the link line and the linker may still pull in the DLL (or produce duplicate symbol conflicts).

I'll update the comment.

PR #3265, opened 2026-05-22
Gate Linux specific code to fix compilation on AIX

pgimalac:
### Issues:
None, but I can open one if needed.

I'm trying to build a Rust binary which depends on the `aws-lc-sys` crate on IBM AIX (a Unix OS), this PR gates linux-specific symbols to fix compiler errors.

### Description of changes:
- Guard `getrandom` logic with `OPENSSL_LINUX` (on top of `OPENSSL_RAND_URANDOM`), I believe this logic is only reachable on Linux
- Guard `RNDGETENTCNT` with `OPENSSL_LINUX ` as the symbol is defined in `linux/random.h`

### Call-outs:
I'm aware AIX is not a supported platform, but the changes required to make it compile are minimal, and I think it makes the code technically more correct. Let me know what you think.

### Testing:
Ran tests locally, built the library on an AIX host.


crypto/rand_extra/urandom.c
```diff
@@ -354,6 +354,9 @@ static void ensure_dev_urandom_is_initialized(void) {
 
   // On platforms where urandom doesn't block at startup, we ensure that the
   // kernel has sufficient entropy before continuing.
+  //
+  // RNDGETENTCNT is a Linux kernel ioctl (from <linux/random.h>)
+#if defined(HAVE_LINUX_RANDOM_H)
```
justsmth: NP: We could improve the comment here further:
```
  // On platforms where urandom doesn't block at startup, we ensure that the
  // kernel has sufficient entropy before continuing. We do this via the
  // RNDGETENTCNT ioctl from <linux/random.h>, which is Linux-specific.
  //
  // On other URANDOM-path platforms (e.g. AIX) we have no portable way to
  // query the kernel entropy pool, so we skip this pre-check and proceed.
```

justsmth (dismissed):

crypto/rand_extra/urandom.c
torben-hansen:
```suggestion
#if defined(OPENSSL_LINUX)
```

torben-hansen:
If IBM AIX could just get `getentropy` such as IBM z/OS does it, that would be great.

crypto/rand_extra/urandom.c
pgimalac: Done !

pgimalac: Done 👍

pgimalac:
@torben-hansen agreed 😄 

Thanks for the reviews !

torben-hansen (approved):

crypto/rand_extra/urandom.c
justsmth: Shouldn't we guard here on `HAVE_LINUX_RANDOM_H` b/c the `RNDGETENTCNT` macro below wouldn't be defined otherwise?

https://www.man7.org/linux/man-pages/man4/urandom.4.html

justsmth: Ok, this file would not even be compiled if `defined(OPENSSL_LINUX)` and not `defined(HAVE_LINUX_RANDOM_H)` -- so guarding w/ `defined(OPENSSL_LINUX)` more clearly indicates the intent.

justsmth (approved):

PR #3266, opened 2026-05-22
BoringSSL: Harden nc_email name constraint checking

nebeid:
### Issues
Addresses: P432271605

### Description of changes:
(Supersedes #3156)
BoringSSL commits:
- [google/boringssl@1616d692b0](https://boringssl-review.googlesource.com/c/boringssl/+/90907): "crypto/x509 nc_email: sync to libpki's logic"
- [google/boringssl@9708047e14](https://boringssl-review.googlesource.com/c/boringssl/+/92067): "crypto/x509: remove OpenSSL exception for name constraints @example.com"

This hardens `nc_email()` to reject malformed email addresses and constraints
that could bypass RFC 5280 name constraint enforcement:

1. Reject subject emails with multiple `@` (catches `a@b@evil.example`)
2. Reject non-atext characters in local-part (catches `"a@b"@evil.example`)
3. Validate domain portions contain only DNS-valid characters
4. Reject `@example.com` as a constraint (empty local-part)
5. Use case-insensitive local-part matching for excluded subtrees

### Adaptations for AWS-LC:
- File remains `.c` (not `.cc` as in BoringSSL)
- Uses C-style `int` return instead of C++ `bool`
- Helper functions named `is_valid_rfc822_local_part`/`is_valid_rfc822_domain`
- No `std::string_view` or `Span<>` usage
- Both CLs combined into a single commit

### Testing:

`crypto_test --gtest_filter=X509Test.NameConstraint*` passes (103 X509 tests total).
New test cases added for:
- Quoted local-part bypass (`"a@b"@evil.example`)
- Multiple `@` bypass (`a@b@evil.example`)
- Multiple `@` in constraint (`a@b@example.com`)
- Case-insensitive excluded subtree matching (`FOO@example.com` excludes `foo@example.com`)


crypto/x509/v3_ncons.c
```diff
@@ -614,48 +615,120 @@ static int nc_dns(const ASN1_IA5STRING *dns, const ASN1_IA5STRING *base,
   return X509_V_OK;
 }
 
-static int nc_email(const ASN1_IA5STRING *eml, const ASN1_IA5STRING *base) {
+// Returns 1 if |cbs| contains only characters valid in an RFC 5321 Sec.4.1.2
+// local-part atom (atext per RFC 5322 Sec.3.2.3). Quoted local-parts are
+// intentionally not supported -- they are vanishingly rare in practice and
```
crlorentzen: Nit: Suggestion to clarify (a quick read confused me) I think it would be clearer to say:

RFC allows quoted local-parts which may contain '@' characters. However, Rather than parsing quoted-strings, we reject local-parts containing non-atext characters.

crlorentzen (dismissed):

crypto/x509/x509_test.cc
```diff
@@ -2773,16 +2773,39 @@ TEST(X509Test, NameConstraints) {
...
+       X509_V_ERR_UNSUPPORTED_NAME_SYNTAX},
+      // Reject constraints with multiple '@' signs.
+      {GEN_EMAIL, "foo@example.com", "a@b@example.com",
+       X509_V_ERR_UNSUPPORTED_NAME_SYNTAX,
+       X509_V_ERR_UNSUPPORTED_NAME_SYNTAX},
+
+      // Excluded subtrees use case-insensitive local-part matching.
+      // Permitted subtrees use case-sensitive local-part matching.
+      {GEN_EMAIL, "foo@example.com", "FOO@example.com",
+       X509_V_ERR_PERMITTED_VIOLATION, X509_V_ERR_EXCLUDED_VIOLATION},
```
justsmth: NP: Isn't this a duplicate of what's on lines 2775-2776?
```
      {GEN_EMAIL, "foo@example.com", "FOO@example.com",
       X509_V_ERR_PERMITTED_VIOLATION, X509_V_ERR_EXCLUDED_VIOLATION},
```

crypto/x509/v3_ncons.c
```diff
@@ -614,48 +615,120 @@ static int nc_dns(const ASN1_IA5STRING *dns, const ASN1_IA5STRING *base,
...
+    if (CBS_get_until_first(&base_cbs, &unused, '@')) {
+      return X509_V_ERR_UNSUPPORTED_NAME_SYNTAX;
+    }
+    if (!is_valid_rfc822_local_part(&base_local) ||
+        !is_valid_rfc822_domain(&base_domain)) {
+      return X509_V_ERR_UNSUPPORTED_NAME_SYNTAX;
+    }
+
+    // For excluded subtrees, compare the local-part case-insensitively.
+    // RFC 2821 and RFC 2985 conflict on case sensitivity; for security we
```
justsmth: NP: Should we cite the more recent RFC 5321 (for SMTP) instead of RFC 2821?

justsmth (dismissed):

crypto/x509/v3_ncons.c
nebeid: Updated to RFC 5321 and RFC 5322.

crypto/x509/x509_test.cc
nebeid: Removed. Already covered at line 2775.

crypto/x509/v3_ncons.c
nebeid: Updated. Now explains that RFC 5322 allows quoted local-parts which may contain '@', and that we reject non-atext rather than parsing quoted-strings.

justsmth (approved):

crlorentzen (approved):

PR #3267, opened 2026-05-22
Add inline documentation for API contracts

nebeid:
### Issues:

Addresses P409219367

### Description of changes:

Adds clarifying comments to document existing API contracts that were reviewed and determined to be working as designed:

- SSL session deserialization: document that session bytes are assumed non-malleable
- BIO pair: document that concurrent free requires caller synchronization
- BIO callbacks: document that callbacks must not free the BIO
- X509 thread safety: add general thread-safety note to the public header

### DIFF
--- a/crypto/bio/bio.c
+++ b/crypto/bio/bio.c
@@ -194,9 +194,6 @@ void BIO_free_all(BIO *bio) {
   BIO_free(bio);
 }
 
-// Note: BIO callbacks are designed for debugging, logging, and auditing. The
-// callback is invoked without incrementing the BIO's reference count, so a
-// callback must not call BIO_free on the BIO it is invoked on.
 int BIO_read(BIO *bio, void *buf, int len) {
 
   if (bio == NULL || bio->method == NULL || bio->method->bread == NULL) {
--- a/crypto/bio/bio.c
+++ b/crypto/bio/bio.c
@@ -194,6 +194,9 @@ void BIO_free_all(BIO *bio) {
   BIO_free(bio);
 }
 
+// Note: BIO callbacks are designed for debugging, logging, and auditing. The
+// callback is invoked without incrementing the BIO's reference count, so a
+// callback must not call BIO_free on the BIO it is invoked on.
 int BIO_read(BIO *bio, void *buf, int len) {
 
   if (bio == NULL || bio->method == NULL || bio->method->bread == NULL) {
--- a/crypto/bio/pair.c
+++ b/crypto/bio/pair.c
@@ -42,6 +42,10 @@ static int bio_new(BIO *bio) {
   return 1;
 }
 
+// bio_destroy_pair unlinks the two halves of a BIO pair. Concurrent calls to
+// BIO_free on both halves from separate threads is the caller's responsibility
+// to synchronize, matching the OpenSSL contract. The BIO pair API does not
+// provide internal locking for teardown.
 static void bio_destroy_pair(BIO *bio) {
   struct bio_bio_st *b = bio->ptr;
   BIO *peer_bio;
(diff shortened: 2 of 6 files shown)


samuel40791765 (dismissed):

crypto/bio/bio.c
```diff
@@ -194,6 +194,9 @@ void BIO_free_all(BIO *bio) {
   BIO_free(bio);
 }
 
+// Note: BIO callbacks are designed for debugging, logging, and auditing. The
```
torben-hansen: Should be in header file?

nebeid: Moved to `bio.h`. Thanks

include/openssl/bio.h
```diff
@@ -286,7 +286,9 @@ OPENSSL_EXPORT uint64_t BIO_number_read(const BIO *bio);
 // |bio|.
 OPENSSL_EXPORT uint64_t BIO_number_written(const BIO *bio);
 
-// BIO_set_callback_ex sets the |callback_ex| for |bio|.
+// BIO_set_callback_ex sets the |callback_ex| for |bio|. The callback is
+// invoked without incrementing the BIO's reference count, so a callback must
+// not call BIO_free on the BIO it is invoked on.
```
justsmth, line 291: This is also true for `BIO_set_callback` -- can we add this comment there as well?

crypto/bio/pair.c
```diff
@@ -42,6 +42,10 @@ static int bio_new(BIO *bio) {
   return 1;
 }
 
+// bio_destroy_pair unlinks the two halves of a BIO pair. Concurrent calls to
+// BIO_free on both halves from separate threads is the caller's responsibility
```
justsmth: NP: "Concurrent calls ... ARE the callers's responsibility ..."

crypto/bio/pair.c
```diff
@@ -42,6 +42,10 @@ static int bio_new(BIO *bio) {
...
+// to synchronize, matching the OpenSSL contract. The BIO pair API does not
```
justsmth: NP: I'm not 100% sure we're "matching the OpenSSL contract", but I did find similar claims in [their documentation](https://docs.openssl.org/3.3/man3/BIO_s_bio):

> Both halves must usually by handled by the same application thread since no locking is done on the internal data structures.
> ...
> Both halves of a BIO pair should be freed. That is even if one half is implicit freed due to a BIO_free_all() or SSL_free() call the other half needs to be freed.

ssl/ssl_asn1.cc
```diff
@@ -497,6 +497,10 @@ UniquePtr<SSL_SESSION> SSL_SESSION_parse(CBS *cbs,
     return nullptr;
   }
 
+  // Note: The secret field is validated only with an upper-bound check. A
+  // minimum-length check is intentionally omitted because the serialized
+  // session bytes are assumed to be non-malleable: the calling application is
+  // responsible for protecting the integrity of session data.
```
justsmth, line 503: Can this also be stated in the public header for `SSL_SESSION_from_bytes`?

nebeid: Done. Added integrity/validation note to `SSL_SESSION_from_bytes` in `ssl.h`.

include/openssl/bio.h
nebeid: Done. Added a reference to the `BIO_set_callback_ex` note.

crypto/bio/pair.c
nebeid: Rephrased to "As in OpenSSL, the BIO pair API does not provide internal locking." — states the fact without claiming contract equivalence.

crypto/bio/pair.c
nebeid: Fixed.

justsmth (approved):

torben-hansen (approved):

Issue #3261, opened 2026-05-20
FIPS build fails with `zig cc`


RafalSumislawski:
### Problem:

An invalid `libbcm_c_generated_asm.a` is generated when `bcm.c` is compiled with Zig, leading to `delocate` failing with a parse error. This only happens when `-DFIPS=1` is set.

```
error while parsing "/work/build/linux-arm64-fips/cmake-build/crypto/fipsmodule/libbcm_c_generated_asm.a":
parse error near Unknown (line 1 symbol 1 - line 1 symbol 1):
""
```

I'm working with `83d85826595641bb2f23c6108f2e031fcd5cd505` commit, as that's what `aws-lc-fips-sys` `0.13.14` mentions in [`Cargo.toml`](https://docs.rs/crate/aws-lc-fips-sys/0.13.14/source/Cargo.toml).

I'm sharing a minimal reproduction of the issue: https://github.com/RafalSumislawski/aws-lc-zig-fips-repro

#### Why it fails?

AWS-LC's FIPS CMakeList defines a normal static library target named [`bcm_c_generated_asm`](https://github.com/aws/aws-lc/blob/83d85826595641bb2f23c6108f2e031fcd5cd505/crypto/fipsmodule/CMakeLists.txt#L338-L344) from [`crypto/fipsmodule/bcm.c`](https://github.com/aws/aws-lc/blob/83d85826595641bb2f23c6108f2e031fcd5cd505/crypto/fipsmodule/bcm.c), adds the regular include/definition setup for that target, and sets [`COMPILE_OPTIONS "-S"`](https://github.com/aws/aws-lc/blob/83d85826595641bb2f23c6108f2e031fcd5cd505/crypto/fipsmodule/CMakeLists.txt#L345-L352) so the compiler emits assembly instead of an object file. As a result the `bcm.c` compile command includes both assembly-output and object/dependency flags:

```sh
zig-cc ... -S -MD -MT ... -MF ... -o bcm.c.o -c bcm.c
```

GCC/Clang tolerate this command shape and emit assembly. Zig does not produce an archive member that `delocate` can parse, so the archive starts with object-like data instead of assembly text.

[...]

PR #3269, opened 2026-05-26
Make FIPS compiler wrapper unconditional

justsmth:
### Issues:
Addresses #3261

### Description of changes:

The `bcm_c_generated_asm` target uses `COMPILE_OPTIONS "-S"` on a CMake static library, which causes CMake to emit both `-S` and `-c` in the same compile command. These flags are semantically contradictory (`-S` = emit assembly, `-c` = emit object code). GCC and older Clang silently let `-S` win, but Zig produces an object file instead of textual assembly, causing `delocate` to fail with a parse error.

We already had a `CompilerWrapper` that strips `-c` when `-S` is present, but it was gated to only activate for Clang ≥ 20. This change removes the version check so the wrapper applies unconditionally, making the build correct for all compilers.

### Call-outs:

This fixes the immediate build failure reported in #3261 but does not guarantee full end-to-end FIPS support with Zig. Other parts of the delocate pipeline (e.g., `-E -dI` preprocessing, assembly parsing) haven't been validated with Zig yet.

### Testing:

Verified that CMake configure succeeds for the `FIPS_DELOCATE` path and that the wrapper is generated and applied to `bcm_c_generated_asm`. Existing CI (including the Zig workflow and FIPS builds with GCC/Clang) will exercise both sides.

### DIFF
--- a/util/compiler_wrapper/CompilerWrapper.cmake
+++ b/util/compiler_wrapper/CompilerWrapper.cmake
@@ -24,6 +24,10 @@
 # wrapper unconditionally strips -c when -S is present, making the build
 # correct for all compilers.
 
+# Save the directory of this file at include-time. Inside functions,
+# CMAKE_CURRENT_LIST_DIR reflects the *caller's* directory, not ours.
+set(_COMPILER_WRAPPER_DIR "${CMAKE_CURRENT_LIST_DIR}")
+
 # Generate wrapper scripts with hardcoded compiler paths
 function(generate_compiler_wrapper)
   # Get current timestamp for generated file header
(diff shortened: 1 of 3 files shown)


util/compiler_wrapper/CompilerWrapper.cmake
prasden: The changes in `compiler_wrapper_needed()` always set  `NEED_WRAPPER` to true now. Do we still need this?

justsmth: Good point.  In the latest commit, I've removed it.

prasden (dismissed):

WillChilds-Klein (dismissed):

WillChilds-Klein (approved):

prasden (approved):

PR #3275, opened 2026-05-28
Add SSL_CTX_set_security_callback and related APIs for OpenSSL compat…

WillChilds-Klein:
### Issues:
Resolves V2219781696

### Description of changes: 

OpenSSL provides security callback APIs that allow applications to customize security decisions based on security levels. AWS-LC does not support security levels, but some consumers depend on the API existing. This adds the callback and ex_data getter/setter functions for both SSL_CTX and SSL objects as no-op shims that store and retrieve values without ever invoking the callback.

Nmap introduced a dependency on this callback API in https://github.com/nmap/nmap/commit/082894d .

Also pin nmap integration test to v7.99 by commit.

### Call-outs:

n/a

### Testing:

- nmap CI job

---


ssl/ssl_test.cc
```diff
@@ -2665,5 +2665,64 @@ TEST(SSLTest, ErrorSyscallAfterCloseNotify) {
...
+  SSL_CTX_set0_security_ex_data(ctx.get(), &dummy_data);
+  EXPECT_EQ(SSL_CTX_get0_security_ex_data(ctx.get()), &dummy_data);
+
+  // Per-SSL object tests.
+  bssl::UniquePtr<SSL> ssl(SSL_new(ctx.get()));
+  ASSERT_TRUE(ssl);
+
+  // Initially no callback is set on the SSL object.
+  EXPECT_EQ(SSL_get_security_callback(ssl.get()), nullptr);
+  EXPECT_EQ(SSL_get0_security_ex_data(ssl.get()), nullptr);
```
justsmth: Since we're implementing the *introspection* part of this mechanism, I think we should at least make it "correct" (i.e., consistent with OpenSSL) --  `SSL_new` should propagate the callback to the `SSL` struct, even though it's not actually used.

include/openssl/ssl.h
```diff
@@ -5989,6 +5989,50 @@ OPENSSL_EXPORT OPENSSL_DEPRECATED int SSL_CTX_get_security_level(
 OPENSSL_EXPORT OPENSSL_DEPRECATED void SSL_CTX_set_security_level(
     const SSL_CTX *ctx, int level);
 
+// SSL_security_callback is the type of a security callback function set via
+// |SSL_CTX_set_security_callback| or |SSL_set_security_callback|. In OpenSSL,
+// this callback is invoked to make security decisions based on the configured
+// security level. In AWS-LC, security levels are not supported and the callback
+// is never invoked. These functions are provided only for API compatibility.
+typedef int (*SSL_security_callback)(const SSL *ssl, const SSL_CTX *ctx,
+                                     int op, int bits, int nid, void *other,
```
justsmth, line 6008: We should mark these new functions as `OPENSSL_DEPRECATED`, just as `SSL_CTX_set_security_level` is. Since consumers might be utilizing these callbacks to enforce security features, we should provide as much visibility as we can about it.

We should also make the comment about callbacks not being used even more prominent:
```
// AWS-LC never invokes the callback, so any restrictions the callback would
// have enforced are silently bypassed. AWS-LC's defaults already disallow the
// algorithms/parameters that OpenSSL's default level-1 callback rejects, but
// a callback that is *more* restrictive than OpenSSL's default will not take
// effect.
```

include/openssl/ssl.h
```diff
@@ -5989,6 +5989,50 @@ OPENSSL_EXPORT OPENSSL_DEPRECATED int SSL_CTX_get_security_level(
 OPENSSL_EXPORT OPENSSL_DEPRECATED void SSL_CTX_set_security_level(
```
justsmth, line 5989: We currently only have `SSL_CTX_set_security_level`, but OpenSSL also has [`SSL_set_security_level`](
https://docs.openssl.org/3.0/man3/SSL_set_security_level/). 

Since we're adding the functions for both `SSL_CTX` and `SSL` below, should we also add `SSL_set_security_level`?

WillChilds-Klein: yes, good call.

include/openssl/ssl.h
WillChilds-Klein: yeah, agreed we should deprecate them. will adjust the comment as well.

justsmth (dismissed):

.github/workflows/integration_omnibus.yml
```diff
@@ -295,7 +295,7 @@ jobs:
             run: ./tests/ci/integration/run_accp_integration.sh
 
           # NMAP Integration Tests
-          - name: nmap
+          - name: nmap-7.99
```
justsmth, line 298: I think locking to a specific release/commit is important for any "required" job, but we lose visibility into changes our library might need to maintain compatibility with `nmap` long term.

tests/ci/integration/run_nmap_integration.sh
```diff
@@ -62,7 +62,11 @@ function nmap_run_tests() {
   make check
 }
 
-git clone --depth 1 https://github.com/nmap/nmap.git ${NMAP_SRC_FOLDER} 
+# Pin to nmap 7.99 release (2026-03-27)
+NMAP_RELEASE_COMMIT="deb076224e9f138ea29fa4823bcce0030301dc54"
```
justsmth: Could we make the commit/tag be an *optional* parameter to the script? That way we could have separate jobs using this script -- one (or more) building against known supported releases (required) and the other tracking the tip of main (not-required)?

.github/workflows/integration_omnibus.yml
WillChilds-Klein: agreed.

justsmth (dismissed):

justsmth (dismissed):

.github/workflows/integration_omnibus.yml
```diff
@@ -295,12 +295,20 @@ jobs:
...
             arch: x86_64
             size: medium
             image: ubuntu:22.04
             compiler: gcc-12
-            run: ./tests/ci/integration/run_nmap_integration.sh
+            run: ./tests/ci/integration/run_nmap_integration.sh deb076224e9f138ea29fa4823bcce0030301dc54
```
samuel40791765, line 303: If we're pinning to a version, we should add an alarm similar to what we're doing here: https://github.com/aws/aws-lc/blob/9426cbfbd5e4056d7fc1feab62b3c15a081d8f66/tests/ci/integration/run_mysql_integration.sh#L40-L47.

Just so we don't forget about this.

samuel40791765 (approved):

justsmth (approved):

PR #3276, opened 2026-05-28
ci: move BSD actions to scheduled

dougch:
### Description of changes: 
Cross platform emulation significantly slows down the *BSD actions. 

By moving these runs to a nightly scheduled run and only running faster tests on FreeBSD-x86 for PRs, we can remove the longest delay in the PR check cycle.

### Call-outs:
This lowers the wall-clock runtime of the BSD jobs to ~17 minutes.
We'll need to monitor for the nightly failures of the scheduled jobs.
Skipping the FIPS build on x86 might not be required to run in under 60 minutes.

### Testing:
How is this change tested (unit tests, fuzz tests, etc.)?  CI


justsmth (approved):

crypto-alg (approved):

.github/workflows/bsd-ci.yml
```diff
@@ -0,0 +1,160 @@
...
+    name: FreeBSD ${{ matrix.version }} (${{ matrix.arch }}) test
+    strategy:
+      fail-fast: false
+      matrix:
+        arch:
+          - "x86-64"
+        version:
+          - "14.2"
+          - "15.0"
+        include: ${{ github.event_name == 'schedule' && fromJSON('[{"arch":"arm64","version":"14.2"},{"arch":"arm64","version":"15.0"}]') || fromJSON('[]') }}
```
nhatnghiho, line 39: nit: The FreeBSD arm64 tests are invisible on PRs (via conditional include), but OpenBSD and NetBSD still show up as "Skipped" checks. Is there a way to make this more consistent?

nhatnghiho (approved):

.github/workflows/bsd-ci.yml
dougch: Unfortunately GitHub Actions doesn't give us a clean way to unify this, without duplicating the FreeBSD block into two; one for scheduled and one for PR.

PR #3280, opened 2026-06-01
Null-guard EVP_AEAD_CTX_cleanup for fork safety with shared memory

WillChilds-Klein:
### Issues:
Resolves #P425555771

#### Summary

  Applications that use `fork()` with `MAP_SHARED` memory for TLS connection state can trigger a SIGSEGV in `EVP_AEAD_CTX_cleanup`. This occurs because `DEFINE_METHOD_FUNCTION` lazily initializes AEAD vtables via `CRYPTO_once` into `.bss` storage. After `fork()`, each process has an independent `.bss` copy. If a child process initializes the vtable (by performing TLS) and stores an `EVP_AEAD_CTX` in shared memory, the parent process's vtable remains zeroed. When the parent calls `EVP_AEAD_CTX_cleanup`, it dereferences `ctx->aead->cleanup`
  which is NULL, crashing the process.

### Call-outs:
n/a

### Testing:

Adds `aead_fork_test`, an isolated test binary that verifies `EVP_AEAD_CTX_cleanup` tolerates an uninitialized vtable after `fork()`. The test is a separate executable (like `rand_isolated_test`) because it requires that AEAD vtables are NOT initialized before the test runs. The FIPS power-on self-test and other tests in `crypto_test` trigger vtable initialization as a side effect.

---


WillChilds-Klein:
Closing in favor of  PR #3336

PR #3286, opened 2026-06-03
Add brainpoolP224r1, brainpoolP256r1, brainpoolP320r1, brainpoolP384r1, brainpoolP512r1 EC group support

### LINKED ISSUE #2939: Feature Request: Support for Brainpool Curves (RFC 5639) in TLS 1.2 and TLS 1.3
### Problem:

**Regulatory Compliance Gap for European Markets (DACH Region)** Currently, `aws-lc` does not support Brainpool Elliptic Curves (RFC 5639). While these curves are less common in the public Web PKI, they are a mandatory requirement for critical infrastructure (KRITIS), government applications, and the energy sector (Smart Meter Gateways) in Germany, Austria, and Switzerland.

Specifically, the German Federal Office for Information Security (BSI) Technical Guideline BSI **TR-03116** mandates the use of Brainpool curves (primarily brainpoolP256r1, brainpoolP384r1, brainpoolP512r1) for secure connections in these sectors.

**Impact on AWS Ecosystem**: Developers building applications on AWS (e.g., using aws-lc-rs for Rust Lambda functions or s2n-tls with aws-lc backend) are currently blocked from entering these regulated markets if they rely on the native AWS crypto stack. They are forced to switch to alternative libraries (like OpenSSL or Bouncy Castle) via Foreign Function Interfaces (FFI), which negates the performance optimizations aws-lc provides for AWS Graviton (ARM) processors and complicates the supply chain.

**Industry Context:** Other major cryptographic libraries have recognized this enterprise need and added full TLS 1.3 support, diverging from the strict "NIST-only" web focus:

- OpenSSL: Added full Brainpool support for TLS 1.3 in Version 3.2.0 (Nov 2023) specifically to address these interoperability needs.
- WolfSSL: Fully supports these curves, often cited as a competitive advantage for embedded/automotive security.
- Bouncy Castle: Widely used in the Java ecosystem for this specific capability.

### Solution:
The proposed solution is to integrate support for Brainpool curves into `aws-lc`'s elliptic curve implementation and expose them for TLS negotiation.

1. Mathematical Primitives: Implement the curve parameters defined in RFC 5639 for P256r1, P384r1, and P512r1.
2. TLS 1.2 Support: Register the legacy NamedCurve codepoints as defined in RFC 7027.
3. TLS 1.3 Support: Implement RFC 8734, which defines the new NamedGroup (e.g., brainpoolP256r1tls13 / 0x001f) and SignatureScheme (e.g., ecdsa_brainpoolP256r1tls13_sha256 / 0x081A) identifiers required for TLS 1.3.

Architecture & Design Decisions:

- Does this change any public APIs? Yes. It would require adding new public constants/NIDs for these curves (e.g., NID_brainpoolP256r1) to allow applications to explicitly configure them.
- Which algorithm(s) will this impact? ECDHE (Key Exchange) and ECDSA (Signatures).

### Requirements / Acceptance Criteria:

The solution must enable a client/server using aws-lc to successfully negotiate a TLS handshake using a Brainpool certificate or ephemeral key share.

RFC links:
* (https://datatracker.ietf.org/doc/html/rfc5639)
* (https://datatracker.ietf.org/doc/html/rfc7027) (TLS 1.2)
* (https://datatracker.ietf.org/doc/html/rfc8734) (TLS 1.3)

Related Issues: N/A (New request for aws-lc)
Will the Usage Guide or other documentation need to be updated? Yes, the list of supported curves/groups should be updated.

Testing:
- **Interoperability**: Integration tests ensuring successful handshakes with OpenSSL 3.2+ (which supports RFC 8734) and WolfSSL.
- **Vectors**: Validation against the test vectors provided in RFC 5639.
- **Will this change trigger AWS LibCrypto Formal Verification changes?** Ideally, yes. While a generic implementation is a good first step, formally verified assembly optimizations (similar to P-256) would be highly desirable for the long-term roadmap given the high-security context of these curves.
- **Should this change be fuzz tested?** Yes. Handling of non-NIST curve parameters should be fuzzed to prevent side-channel leaks or validation bypasses.

### Out of scope:

- **Small Curves**: Support for Brainpool curves smaller than 256 bits (e.g., brainpoolP160r1) is not requested, as they do not meet modern security standards (BSI TR-03116) and are not required for the targeted enterprise use cases.
- **Twisted Brainpool Curves**: The "twisted" variants (e.g., brainpoolP256t1) are generally not required for TLS/X.509 usage, only the "random" (r1) variants are standard for certificates.

sfarestam-iproov:
### Issues:
Resolves #2939

### Description of changes:
AWS-LC currently does not support Brainpool elliptic curves. This PR adds EC group support for the five Brainpool r1 prime curves defined in [RFC 5639](https://datatracker.ietf.org/doc/html/rfc5639): brainpoolP224r1, brainpoolP256r1, brainpoolP320r1, brainpoolP384r1, and brainpoolP512r1.

The implementation follows the same pattern as secp256k1 — generic Montgomery arithmetic via `EC_GFp_mont_method()`, with no hand-optimized assembly. The only structural addition is `ec_group_set_a_mont()` for setting an arbitrary Montgomery-form `a` coefficient, since Brainpool curves have `a ≠ -3` (unlike NIST curves) and `a ≠ 0` (unlike secp256k1). The existing `ec_GFp_mont_dbl()` already handles this case correctly via its `a_is_minus3 == 0` code path.

NIDs (925, 927, 929, 931, 933) and OIDs were already registered in `nid.h` and `obj_dat.h`. All domain parameters are sourced from RFC 5639 Sections 3.3–3.7, with OIDs from Section 4.1.

This PR covers EC primitive support (key generation, ECDSA, ECDH). TLS negotiation support (RFC 7027 / RFC 8734) can follow in a separate PR.

**Motivation:** These curves are required by [ICAO Doc 9303](https://www.icao.int/sites/default/files/publications/DocSeries/9303_p12_cons_en.pdf) (ePassport standard, 30+ countries), [BSI TR-03116-4](https://www.bsi.bund.de/SharedDocs/Downloads/DE/BSI/Publikationen/TechnischeRichtlinien/TR03116/BSI-TR-03116-4.html) (German federal regulation), and are blocking the [pyca/cryptography](https://github.com/pyca/cryptography) project from accepting Brainpool improvements ([pyca/cryptography#14905](https://github.com/pyca/cryptography/pull/14905)).

### Call-outs:
- **FIPS boundary**: The code is in `crypto/fipsmodule/ec/`. If adding curves inside the FIPS module is a concern for recertification, the implementation can be moved to `crypto/ec_extra/`. Please advise on the preferred placement.
- **Arbitrary `a` coefficient**: Added a small helper `ec_group_set_a_mont()` (6 lines) to copy a Montgomery-form `a` value. This is the only new function beyond the existing curve template pattern.
- **`make_tables.go` changes**: Added a `curveWithA` wrapper type and `writeCurveDataWithA()` function to emit `MontA` constants for curves where `a` is not -3 or 0. The standard `elliptic.CurveParams` in Go doesn't carry an `A` field (it assumes `a = -3`).

### Testing:
- `BrainpoolKeygenSignVerify`: generates a key, signs a digest with ECDSA, verifies the signature, and checks that a corrupted signature is rejected — for all 5 curves.
- `ECPKParmatersBio`: round-trip `i2d_ECPKParameters_bio` / `d2i_ECPKParameters_bio` for all 5 curves.
- `GetNamedCurve`: dispatches Brainpool curve names for file-based test vectors.
- Scalar base multiplication test vectors are not included in this PR (generating them requires a Go Brainpool library); they can be added as a follow-up.


crypto/fipsmodule/ec/ec_test.cc
```diff
@@ -2777,6 +2792,53 @@ TEST(ECTest, ECPKParmatersBio) {
...
+  for (int nid : kBrainpoolNIDs) {
+    SCOPED_TRACE(nid);
+    bssl::UniquePtr<EC_KEY> key(EC_KEY_new_by_curve_name(nid));
+    ASSERT_TRUE(key);
+    ASSERT_TRUE(EC_KEY_generate_key(key.get()));
+
+    // Sign a digest.
+    std::vector<uint8_t> sig(ECDSA_size(key.get()));
+    unsigned sig_len;
+    ASSERT_TRUE(
```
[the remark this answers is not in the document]
sfarestam-iproov: Fixed — initialized to 0. Thanks.

nebeid:
Thank you @sfarestam-iproov for this contribution.
1. FIPS boundary placement — I'd lean toward moving this to crypto/ec_extra/ as you offered. 

2. Known-answer test vectors — Unless I'm missing them, I don't see any KATs. The current tests (BrainpoolKeygenSignVerify, ECPKParmatersBio) are round-trip/self-consistency checks, which would pass even if the domain parameters were subtly off as long as they're internally consistent. Since hardcoded curve parameters are the core of this change, it'd be good to validate `p`/`a`/`b`/`G`/`n` and the generated tables against independent reference vectors (RFC 5639, Wycheproof, or another library). I'd suggest including these in this PR rather than as a follow-up.

sfarestam-iproov:
Thank you for the review @nebeid — both changes are now addressed:

**1. FIPS boundary placement** (commit `4ee8115`): Moved all 5 Brainpool `DEFINE_METHOD_FUNCTION` blocks from `crypto/fipsmodule/ec/ec.c` to `crypto/ec_extra/ec_brainpool.c`. The small helper functions (`ec_group_init_static_mont`, `ec_group_set_a_mont`) are duplicated in the new file rather than exposed via `internal.h`, to avoid widening the FIPS module's internal API surface.

**2. Known-answer test vectors** (commit `193a2a2`): Added two independent validation layers:

- **`BrainpoolDomainParamsRFC5639`** (`ec_test.cc`): Extracts `p`/`a`/`b`/`Gx`/`Gy`/`n` from each curve's `EC_GROUP` (converting out of Montgomery form) and compares against the hex reference values from RFC 5639 Sections 3.2–3.7. This catches subtly wrong precomputed constants that self-consistency tests would miss.

- **`WycheproofECDSABrainpool`** (`evp_test.cc`): Runs Wycheproof ECDSA verify vectors for all 5 curves (~22k lines of test vectors sourced from [C2SP/wycheproof](https://github.com/C2SP/wycheproof)), validating that the precomputed multiplication tables produce correct point arithmetic. Follows the existing `WycheproofECDSAsecp256k1` pattern.

sfarestam-iproov:
CI fix in `dc21069`: `DEFINE_METHOD_FUNCTION` can't be used outside `crypto/fipsmodule/` — in FIPS builds it expands to external BSS symbol references that only exist within the delocated `bcm.o`. Replaced with an equivalent `CRYPTO_once`-based lazy-init macro defined locally in `ec_brainpool.c`.

nebeid:
Thanks for the fast turnaround on the FIPS link failure — and for adding the KAT + Wycheproof coverage earlier. Your `dc210696b` fix is correct: replacing `DEFINE_METHOD_FUNCTION` with the
  local `DEFINE_CURVE_DATA` macro (plain `static` storage + `CRYPTO_once`). `DEFINE_METHOD_FUNCTION` relies on delocated BSS storage that only exists inside `bcm.o`, so it can't be used from `ec_extra` as you found out — your replacement reproduces the non-FIPS path (the same thing [`delocate.h`'s `#else` branch](https://github.com/aws/aws-lc/blob/cf95ba04fd934450b6747e90109ed2a2ffdc1740/crypto/fipsmodule/delocate.h#L25) does) and the FIPS builds now pass. 

 **Recommended change: avoid per-curve growth in `bcm.o`**

 One thing still grows the FIPS module a little: the 10 hardcoded `case NID_brainpool*` arms added to  [`EC_GROUP_new_by_curve_name`](https://github.com/aws/aws-lc/blob/cf95ba04fd934450b6747e90109ed2a2ffdc1740/crypto/fipsmodule/ec/ec.c#L329) and [`_mutable`](https://github.com/aws/aws-lc/blob/cf95ba04fd934450b6747e90109ed2a2ffdc1740/crypto/fipsmodule/ec/ec.c#L347) (5 curves × 2 functions). These live inside `bcm.o`, and the cost scales linearly with every future non-FIPS curve. I'd recommend collapsing them into a single generic fallthrough that defers to a resolver in `ec_extra`, e.g.:

  ```c
      default: {
        const EC_GROUP *g = ec_group_new_by_curve_name_nonfips(nid);  // defined in ec_extra
        if (g != NULL) {
          return (EC_GROUP *)g;
        }
        OPENSSL_PUT_ERROR(EC, EC_R_UNKNOWN_GROUP);
        return NULL;
      }
  ```
  
  with a small NID→accessor table outside the boundary:

  ```c
  // crypto/ec_extra/ec_brainpool.c (or a small ec_extra registry)
  const EC_GROUP *ec_group_new_by_curve_name_nonfips(int nid) {
    switch (nid) {
      case NID_brainpoolP224r1: return EC_group_brainpoolP224r1();
      case NID_brainpoolP256r1: return EC_group_brainpoolP256r1();
      case NID_brainpoolP320r1: return EC_group_brainpoolP320r1();
      case NID_brainpoolP384r1: return EC_group_brainpoolP384r1();
      case NID_brainpoolP512r1: return EC_group_brainpoolP512r1();
      default: return NULL;
    }
  }
  ```
  
  Why this is worth doing:
  - `bcm.o` then grows by a **fixed, small amount** (one external call site per function) instead of per-curve.
  - It keeps all Brainpool dispatch logic, like the curve data, outside the FIPS boundary.
  - It mirrors the established pattern AWS-LC already uses for non-FIPS items — the in-module selector that searches FIPS entries first and then falls through to a non-FIPS list reached across the boundary: [`evp_pkey_meth_find`](https://github.com/aws/aws-lc/blob/cf95ba04fd934450b6747e90109ed2a2ffdc1740/crypto/fipsmodule/evp/evp_ctx.c#L29) + [`non_fips_pkey_evp_methods`](https://github.com/aws/aws-lc/blob/cf95ba04fd934450b6747e90109ed2a2ffdc1740/crypto/evp_extra/p_methods.c#L10). (Note that's keyed by `pkey_id` at the algorithm layer, so it can't be reused directly — Brainpool curves all share the single `EVP_PKEY_EC` method and are selected by NID — but the *shape* is what we want, applied one layer down at the`EC_GROUP` NID dispatch.)

(Aside: once the dispatch is a single fallthrough, you may find the `DEFINE_CURVE_DATA` macro reads more simply as a plain per-function `CRYPTO_once` + accessor, but that's up to you if you'd like to change it.)

nebeid:
The `minimal-*`, `legacy-avx-*`, `wasm`/`wasi`, `macOS-ARM-FIPS`, and `pre_push_checker` jobs all fail for one reason that's unrelated to the curve/FIPS code:

  ```
  File not embedded: .../ecdsa_brainpoolP224r1_sha224_test.txt.
  ```

  You added the vectors to `CRYPTO_TEST_DATA`, but the no-Go toolchains embed test data from the committed `generated-src/crypto_test_data.cc.tar.bz2` (see [INCORPORATING.md](https://github.com/aws/aws-lc/blob/main/INCORPORATING.md)), which wasn't regenerated.

  Fix: **regenerate and commit `generated-src/`**:
  ```sh
  python3 util/generate_build_files.py
  ```
  Verify with the same guard CI uses: `./tests/check_generated_src.sh`. This is the documented step for adding vectors (per [`.github/ISSUE_TEMPLATE/update-vectors.md`](https://github.com/aws/aws-lc/blob/main/.github/ISSUE_TEMPLATE/update-vectors.md)).

(The 3 ABI checks are unrelated/expected — just the 5 new public `EC_group_brainpool*` symbols.)

sfarestam-iproov:
Both items addressed:

**1. NID dispatch out of bcm.o** (`bce99c1`): Replaced the 10 per-curve `case NID_brainpoolP*:` arms in `EC_GROUP_new_by_curve_name` and `_mutable` with a single `default:` fallthrough to `ec_group_new_by_curve_name_nonfips()`, declared in `ec_extra/internal.h` and defined in `ec_extra/ec_brainpool.c`. Follows the `evp_pkey_meth_find` → `AWSLC_non_fips_pkey_evp_methods` pattern.

**2. Regenerated `generated-src/`** (`af4d4cf`): Ran `generate_build_files.py` to include the Brainpool Wycheproof vectors in `crypto_test_data.cc.tar.bz2`. Verified with `check_generated_src.sh`.

sfarestam-iproov:
The `strongswan-x86_64` CI failure is caused by our changes and is now fixed in 43d698ef7.

**Root cause**: `evp_pkey_tls_encodedpoint_ec_curve_supported()` in `crypto/fipsmodule/evp/evp.c` had a hardcoded allowlist of only four NIST curves (P-224, P-256, P-384, P-521). strongswan's openssl plugin uses `EVP_PKEY_get1_tls_encodedpoint()` / `EVP_PKEY_set1_tls_encodedpoint()` for its DH key exchange implementation. Before this PR, `EC_GROUP_new_by_curve_name(NID_brainpoolP256r1)` returned NULL so strongswan skipped the test vector. With Brainpool support now present, strongswan detects the curve and attempts an ECDH exchange, but the public key export/import fails at the allowlist.

**Fix**: Added `NID_secp256k1` and all five Brainpool NIDs to the allowlist. The underlying point encoding (`EC_POINT_oct2point`, `EC_KEY_key2buf`) is curve-agnostic — the restriction was just overly conservative. The actual ECDH computation path (`EVP_PKEY_derive`) has no curve restrictions.

Note: the `clang` (Windows) failure is unrelated — it's a flaky `ALPS-Decline-Server-Old-DTLS-TLS12` test failing due to Windows socket exhaustion (`system lacked sufficient buffer space`).

nebeid:
Thank you for widening the curve allowlist in [`evp_pkey_tls_encodedpoint_ec_curve_supported`](https://github.com/aws/aws-lc/blob/43d698ef70d0b25f5aebe557ec02421e583ab46e/crypto/fipsmodule/evp/evp.c#L739) 

  Suggestion: add a direct unit test so this gate is guarded explicitly rather than incidentally. The existing [`ECTLSEncodedPoint` test](https://github.com/aws/aws-lc/blob/43d698ef70d0b25f5aebe557ec02421e583ab46e/crypto/evp_extra/evp_test.cc#L1334) already does hardcoded per-curve `set1`/`get1` round-trips for the NIST curves; extending it to cover `secp256k1` and at least one Brainpool curve would pin the allowlist and prevent future drift.

sfarestam-iproov:
Done in 18fe06703 — extended the `ECTLSEncodedPoint` test to cover `secp256k1` (Wycheproof `ecdh_secp256k1_test.json` tcId 1) and `brainpoolP256r1` (RFC 7027 Section A.1). Both entries exercise the full `set1` → `EVP_PKEY_derive` → `get1` round-trip, pinning the allowlist.

nebeid (dismissed):

crypto/ec_extra/ec_asn1.c
```diff
@@ -29,8 +29,11 @@ static const CBS_ASN1_TAG kPublicKeyTag =
 // acceptable groups, so parsers don't have to pull in all four.
```
manastasova: nit:
```suggestion
// acceptable groups, so parsers don't have to pull in all ten.
```

crypto/ec_extra/ec_brainpool.c
```diff
@@ -0,0 +1,254 @@
...
+#include "../internal.h"
+#include "internal.h"
+
+#include "../fipsmodule/ec/builtin_curves.h"
+
+// DEFINE_METHOD_FUNCTION (from delocate.h) cannot be used outside the FIPS
+// module because it relies on delocated BSS storage within bcm.o. Since
+// Brainpool curves are not FIPS-approved, we use an equivalent lazy-init
+// pattern directly.
+#define DEFINE_CURVE_DATA(type, name)                                         \
```
manastasova, line 26: nit: alignment is off

crypto/ec_extra/ec_asn1.c
sfarestam-iproov: Fixed in e82b181 — updated to "all ten".

crypto/ec_extra/ec_brainpool.c
sfarestam-iproov: Fixed in e82b181 — all continuation lines now aligned to column 79, matching `DEFINE_DATA` in `delocate.h`.

sfarestam-iproov:
@manastasova Both nits addressed in e82b181.

The PR currently has a merge conflict against `main` — it's in `generated-src/crypto_test_data.cc.tar.bz2` (binary file, both sides modified). Resolving it requires a rebase onto `main` followed by regenerating `generated-src`.

Since @justsmth already rebased the branch once, happy to defer to you on whether you'd like to handle this rebase or if I should do it on my end.

crypto/ec_extra/ec_brainpool.c
manastasova: Thanks

crypto/ec_extra/ec_asn1.c
manastasova: Thank you

manastasova (dismissed):

sfarestam-iproov:
Resolved the merge conflict — merged upstream main and regenerated `generated-src/crypto_test_data.cc.tar.bz2`. PR is mergeable again. Both approvals were dismissed by the push, so re-review is needed when you get a chance.

manastasova (approved):

nebeid (approved):

PR #3287, opened 2026-06-04
Switch integration tests to nightly and raise alarm on failures

prasden:
### Issues:
Resolves #V2139430768

### Description of changes: 
Some integration tests are mandatory in the CI, so unrelated PRs are occasionally blocked by changes downstream in the integration provider's repo. This PR converts the integration test suite workflow to run nightly, and raise an alarm to CloudWatch whenever one or more integration tests are failing, so failing tests can be fixed and contributor PRs can be unblocked quicker. 

### Call-outs:
The CloudWatch Alarm was added manually to the AWS-LC namespace. 

### Testing:
Tested with AWS CLI tool and raised a dummy alarm and verified ticket was created in SIM.


.github/workflows/integration_omnibus.yml
```diff
@@ -496,3 +531,37 @@ jobs:
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -exo pipefail
+          failure_count=$(gh run view "$GITHUB_RUN_ID" --json jobs \
+            --jq '[.jobs[] | select(.conclusion == "failure" and .name != "report-failures")] | length')
+
+          aws cloudwatch put-metric-data \
+            --namespace AWS-LC \
+            --metric-name IntegrationOmnibusFailure \
+            --value "$failure_count"
```
justsmth: This produces the metric, but do we have an alarm configured for it in this PR as well? Or will this be manually created?

.github/workflows/integration_omnibus.yml
```diff
@@ -496,3 +531,37 @@ jobs:
...
+        uses: ./.github/actions/emit-autofix-target
+        with:
+          integration: ruby
+          version: ${{ matrix.version }}
+          failed-integration: ruby-${{ matrix.version }}-${{ matrix.buildFIPS }}-${{ matrix.arch }}
+
+  report-failures:
+    name: report-failures
+    needs: [integrations, python, ruby]
+    if: ${{ always() && github.event_name == 'schedule' && (needs.integrations.result == 'failure' || needs.python.result == 'failure' || needs.ruby.result == 'failure') }}
```
justsmth: This reports failure, but it's best-practice to also emit metrics on success so you can distinguish between "this is not running" from "this is always succeeding".

.github/workflows/autofix_integration_omnibus.yml
```diff
@@ -0,0 +1,162 @@
...
+      - codebuild-aws-lc-ci-github-actions-${{ github.run_id }}-${{ github.run_attempt }}
+        image:linux-5.0
+        instance-size:small
+    env:
+      CLAUDE_CODE_USE_BEDROCK: "1"
+      CLAUDE_CODE_SUBPROCESS_ENV_SCRUB: "1" # Strip secret creds from the env Claude's Bash subprocesses inherit to guard against prompt injection
+      AWS_REGION: us-west-2
+      ANTHROPIC_DEFAULT_OPUS_MODEL: us.anthropic.claude-opus-4-8
+      ANTHROPIC_DEFAULT_SONNET_MODEL: us.anthropic.claude-sonnet-4-6
+      ANTHROPIC_DEFAULT_HAIKU_MODEL: us.anthropic.claude-haiku-4-5-20251001-v1:0
```
justsmth: How often will these need bumping? Bedrock doesn't seem to offer a `-latest` alias for these, so is there a plan/reminder to keep them current (or pin them somewhere more discoverable)?

tests/ci/cdk/cdk/aws_lc_base_ci_stack.py
```diff
@@ -27,7 +27,6 @@ def __init__(
         super().__init__(scope, id, env=env, **kwargs)
         self.ignore_failure = ignore_failure
         self.timeout = timeout
-        self.env = env
```
justsmth, line 30: Is this change needed?

tests/ci/integration/auto_fix_integrations/claude-settings.json
```diff
@@ -0,0 +1,59 @@
...
+      "Bash(docker *)"
+    ]
+  },
+  "sandbox": {
+    "enabled": true,
+    "failIfUnavailable": true,
+    "enableWeakerNestedSandbox": true,
+    "network": {
+      "allowedDomains": [
+        "downloads.nwtime.org"
```
justsmth: This domain allowlist and the `Bash(wget ...)` allow on line 20 have to be kept in sync by hand, and a new wget-based integration on a different host would need both updated. Is there a way to enforce/centralize that?

tests/ci/integration/auto_fix_integrations/auto_fix.sh
```diff
@@ -0,0 +1,307 @@
...
+  if LC_ALL=C grep -E -i -q "${patterns}" "$@"; then
+    echo "::error::Potential secret detected in patch; refusing to open PR."
+    exit 1
+  fi
+}
+
+open_pr() {
+  local target="$1"
+  local branch_name="$2"
+  local push_url="https://x-access-token:${GH_TOKEN}@github.com/${REPO}.git"
```
justsmth: We need to be cautious about exposing the `GH_TOKEN`. Can we use `git -c http.extraheader=...` or some credential helper instead?

.github/workflows/integration_omnibus.yml
```diff
@@ -496,3 +531,37 @@ jobs:
...
+      - uses: ./.github/actions/configure-aws-credentials
+        with:
+          roleName: AwsLcGitHubActionStandardRole
+      - name: Emit failure metric
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -exo pipefail
+          failure_count=$(gh run view "$GITHUB_RUN_ID" --json jobs \
+            --jq '[.jobs[] | select(.conclusion == "failure" and .name != "report-failures")] | length')
```
justsmth: Might we be able to filter out failures due to operational flakiness?

tests/ci/integration/auto_fix_integrations/auto_fix.sh
```diff
@@ -0,0 +1,307 @@
...
+                    --paginate \
+                    --jq ".jobs[]
+                          | select(.conclusion == \"failure\" and (.name | startswith(\"${prefix}\")))
+                          | .id")
+  do
+    gh api "/repos/${REPO}/actions/jobs/${job_id}/logs" \
+      | tail -n 200 | sanitize_log > "${logs_dir}/${job_id}.log" || true
+  done
+}
+
```
justsmth: This matches failed jobs via `startswith("${prefix}")` where `prefix` is the runner-script-derived integration name. Will this assumption always hold? Can we enforce it?

.github/workflows/integration_omnibus.yml
prasden: Yep, I created this manually in CloudWatch before this PR

.github/workflows/autofix_integration_omnibus.yml
```diff
@@ -0,0 +1,162 @@
...
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - name: Recognize failed integration targets
+        id: recognize
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -eo pipefail
+          chmod +x "$AUTOFIX_SCRIPT"
```
nhatnghiho: These `chmod`'s are little redundant. If you add the correct permissions to `auto_fix.sh` locally and commit the file to Github, CodeBuild will inherit the same file permissions and we can remove these lines.

.github/workflows/integration_omnibus.yml
```diff
@@ -496,3 +531,37 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/configure-aws-credentials
+        with:
+          roleName: AwsLcGitHubActionStandardRole
+      - name: Emit failure metric
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
nhatnghiho: nit: I think this is the same as `{{ github.token }}`, which is used throughout the other workflow? Change this to make it more consistent.

.github/workflows/autofix_integration_omnibus.yml
```diff
@@ -0,0 +1,162 @@
...
+        env:
+          TARGET: ${{ matrix.target }}
+      - name: Claude's thinking traces
+        run: |
+          set -exo pipefail
+          IFS='|' read -r integration version <<< "$TARGET"
+          "$AUTOFIX_SCRIPT" reason "$integration" "$version" "$TARGET_RUN_ID"
+        env:
+          TARGET: ${{ matrix.target }}
+          GIT_ALLOW_PROTOCOL: file:git:http:https:ssh
```
nhatnghiho: Do we need ssh here if we deny ssh in claude-settings.json?

.github/workflows/autofix_integration_omnibus.yml
prasden: How often we need to update depends on Anthropic. For now, I think we can pin the model/names to this file because nothing else uses Claude Code for any automation. When a new feature in AWS-LC wants to use Claude, I think we can then move these model pins into a more centralized file we can iteratively change. But maybe after Mythos drops we won't need to ever change it again..

tests/ci/cdk/cdk/aws_lc_base_ci_stack.py
prasden: Yes it is, I forgot to include this in the call outs, my mistake. CDK was failing to deploy on my local account with the new IAM roles and I traced to this variable. We remove this as CDK already exposes self.env as a built-in read-only property: https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk/Stack.html#aws_cdk.Stack.env.

.github/workflows/autofix_integration_omnibus.yml
prasden: Good call, no we don't

.github/workflows/integration_omnibus.yml
prasden: Ack, I modified it so it runs on every scheduled run and emits 0 if no failures are reported, and no artifact is produced.

tests/ci/integration/auto_fix_integrations/claude-settings.json
prasden: Good call, removed the `Bash(wget)` from the allowlist and centralized which external domains are allowed in the sandbox in `allowedDomains.`

tests/ci/integration/auto_fix_integrations/auto_fix.sh
prasden: Dropped it from the URL and switched to `gh auth setup-git`, we now read GH_TOKEN from the env

.github/workflows/integration_omnibus.yml
prasden: Yes, added an id on each integration test to catch failure only on steps which contain the id for integration tests

tests/ci/integration/auto_fix_integrations/auto_fix.sh
prasden: It's not currently enforced but I think we should for new tests, maybe somewhere like `CONTRIBUTING.md.`  The integration suite has many jobs and each job can have different build flags like ACCP where it has 4 rows across architectures/FIPS and some patches like `python_patch` have different versions like `python_patch/3.13.` We should lock down the naming across job, script, and the patch directory.

justsmth (dismissed):

.github/workflows/integration_omnibus.yml
```diff
@@ -418,6 +422,40 @@ jobs:
           run: |
             source /opt/compiler-env/setup-${{ matrix.compiler }}.sh
             ${{ matrix.run }}
+      # matrix.name (e.g. "postgresql") doesn't always match the patch dir
+      # name ("postgres_patch"), we use the runner script path as the source of truth.
+      - name: Get failed integration name and version
+        id: autofix-target
+        if: steps.integration-test.outcome == 'failure' && github.event_name == 'schedule'
```
WillChilds-Klein: nit: here and elsewhere, suggest adding a LOUD comment here reminding the reader that if the latter part of the clause gets deleted, this will get run on every PR

.github/actions/emit-autofix-target/action.yml
```diff
@@ -0,0 +1,25 @@
...
+description: "Upload the (integration, version) pair of a failed integration job for autofix to consume."
+inputs:
+  integration:
+    description: "Integration name"
+    required: true
+  version:
+    description: "Version/branch the runner ran with (empty for flat patch dirs)."
+    required: false
+    default: ""
+  failed-integration:
```
WillChilds-Klein: these don't need to be unique across runs, right? artifact upload is scoped per-run?

.github/actions/fetch-github-token/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    default: "us-west-2"
+runs:
+  using: "composite"
+  steps:
+    - shell: bash
+      run: |
+        set -eo pipefail
+        token=$(aws secretsmanager get-secret-value --region "${{ inputs.aws-region }}" \
+          --secret-id "${{ inputs.secret-id }}" --query SecretString --output text)
+        echo "::add-mask::$token"
```
WillChilds-Klein: where is this `echo` statement emitting the token? i see that [add-mask hides the token in public logs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#example-masking-a-generated-output-within-a-single-job), is it needed if we're already adding the token to $GITHUB_ENV?

if it's needed, suggest adding a comment about `add-mask`. removing that directive in the future would be pretty bad.

.github/workflows/autofix_integration_omnibus.yml
```diff
@@ -0,0 +1,159 @@
...
+      ANTHROPIC_DEFAULT_HAIKU_MODEL: us.anthropic.claude-haiku-4-5-20251001-v1:0
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/configure-aws-credentials
+        with:
+          roleName: AwsLcGitHubActionIntegrationFailureReasoningRole
+      - name: Install Claude Code
+        run: npm install -g @anthropic-ai/claude-code@2.1.161
```
WillChilds-Klein: what's the thought on pinning this version -- avoid unexpected breakages? if so, maybe we can pin it at minor version rather than build?

.github/workflows/autofix_integration_omnibus.yml
```diff
@@ -0,0 +1,159 @@
...
+          slug=${TARGET//[|\/]/-}
+          slug=${slug%-}
+          echo "dir=$slug" >> "$GITHUB_OUTPUT"
+        env:
+          TARGET: ${{ matrix.target }}
+      - name: Claude's thinking traces
+        run: |
+          set -exo pipefail
+          IFS='|' read -r integration version <<< "$TARGET"
+          "$AUTOFIX_SCRIPT" reason "$integration" "$version" "$TARGET_RUN_ID"
```
WillChilds-Klein: our various creds scrubbers will run over this output too, right?

do you have sample logs showing a reasoning  trace? i can't find one in the [synthetic job run failure](https://github.com/prasden/pd-aws-lc/actions/runs/26936087533) that created your test PR

.github/workflows/autofix_integration_omnibus.yml
```diff
@@ -0,0 +1,159 @@
...
+        env:
+          TARGET: ${{ matrix.target }}
+          GIT_ALLOW_PROTOCOL: file:git:http:https
+      - uses: actions/upload-artifact@v4
+        with:
+          name: resolve-patch-${{ steps.target.outputs.dir }}
+          path: .resolve-integration-failures/${{ steps.target.outputs.dir }}/out/
+          if-no-files-found: ignore
+      - uses: actions/upload-artifact@v4
+        if: always()
```
WillChilds-Klein: do we need this condition?

.github/workflows/integration_omnibus.yml
```diff
@@ -418,6 +422,40 @@ jobs:
...
+
+          integration=${script_path##*/}  # Strip everything before the last / so we have just the filename.
+          integration=${integration%.sh}  # Drop the .sh extension.
+          integration=${integration#run_} # Drop the run_ prefix shared by every runner.
+          integration=${integration%_integration} # Drop the _integration suffix to leave the bare integration name.
+          version=${version//[\'\"]/} # Strip surrounding quotes (some matrix entries quote the version arg).
+
+          # openssh doesn't pass its branch on the command line as it reads it
+          # from the $OPENSSH_BRANCH env var, so grab the version from there.
+          [[ -z "$version" && "$integration" == "openssh" ]] && version="$OPENSSH_BRANCH"
```
WillChilds-Klein: is openssh the only integration test target like this?

.github/workflows/integration_omnibus.yml
```diff
@@ -462,6 +501,12 @@ jobs:
           run: |
             source /opt/compiler-env/setup-gcc-12.sh
             ./tests/ci/integration/run_python_integration.sh ${{ matrix.version }}
+      - if: steps.integration-test.outcome == 'failure' && github.event_name == 'schedule'
+        uses: ./.github/actions/emit-autofix-target
+        with:
+          integration: python
+          version: ${{ matrix.version }}
+          failed-integration: python-${{ matrix.version }}-${{ matrix.buildFIPS }}-${{ matrix.crtUseSystemCrypto }}-${{ matrix.arch }}
```
WillChilds-Klein: why do we have special cases here and for ruby below? shouldn't failures be detected automatically like the other tests?

.github/workflows/integration_omnibus.yml
```diff
@@ -496,3 +542,39 @@ jobs:
...
+        uses: ./.github/actions/emit-autofix-target
+        with:
+          integration: ruby
+          version: ${{ matrix.version }}
+          failed-integration: ruby-${{ matrix.version }}-${{ matrix.buildFIPS }}-${{ matrix.arch }}
+
+  report-failures:
+    name: report-failures
+    needs: [integrations, python, ruby]
+    if: ${{ always() && github.event_name == 'schedule' }}
```
WillChilds-Klein, line 537: why `always() &&` here?

tests/ci/integration/auto_fix_integrations/claude-settings.json
```diff
@@ -0,0 +1,59 @@
...
+      ]
+    },
+    "filesystem": {
+      "denyRead": ["~/.aws", "~/.ssh"]
```
WillChilds-Klein: can we go further? like only work in some `~/workdir` and default-deny access to everything else?

tests/ci/integration/auto_fix_integrations/auto_fix.sh
```diff
@@ -0,0 +1,320 @@
...
+    124) echo "::warning::Claude timed out after ${CLAUDE_TIMEOUT}s." ;;
+    *)   echo "::warning::Claude exited with code ${rc}." ;;
+  esac
+  return "${rc}"
+}
+
+
+scan_secrets() {
+  local patterns='(AKIA|ASIA|AGPA|AIDA|AROA|AIPA|ANPA|ANVA)[A-Z0-9]{16}|aws_secret_access_key|gh[pousr]_[A-Za-z0-9]{36}|github_pat_[A-Za-z0-9_]{20,}|-----BEGIN[A-Z ]*PRIVATE KEY-----|[Bb]earer[[:space:]]+[A-Za-z0-9._-]{16,}'
+  if LC_ALL=C grep -E -i -q "${patterns}" "$@"; then
```
WillChilds-Klein: should we pipe through `scan_log` as well? what if there's some `beare<unprintable>r` string?

tests/ci/integration/auto_fix_integrations/prompt.md
```diff
@@ -0,0 +1,45 @@
...
+   ### How I fixed it
+   <2-4 sentences; include a small fenced diff or `inline code` for hunk
+   headers, file paths, or symbols where it makes the change concrete>
+   ```
+
+   Use `` `inline code` `` for symbols, file paths, and line refs (e.g.
+   `` `vio/viosslfactories.c:488` ``). Use fenced code blocks for diffs,
+   error excerpts, or commands. Keep prose tight.
+
+Do NOT push or open a PR. If you cannot fix it, print 'AUTOFIX_FAILED: INTEGRATION_PLACEHOLDER' and exit.
```
WillChilds-Klein: exit with non-zero exit code?

WillChilds-Klein (commented):
I really like the role separation of GH PAT access and bedrock access across resolve/reason jobs!

.github/workflows/autofix_integration_omnibus.yml
prasden: Yes, under the [`autofix-integration-omnibus`](https://github.com/prasden/pd-aws-lc/actions/runs/26981339274/job/79621171182) workflow, in the `reason` job, you can see the logs in the `Claude's thinking traces` step.

.github/actions/emit-autofix-target/action.yml
prasden: Yes, correct

.github/actions/fetch-github-token/action.yml
prasden: Yes, it is needed. It's not actually emitting the token through `echo,` it's a command that tells the runner to hide the value from subsequent steps. Adding the variable to `GITHUB_ENV` doesn't hide the variable, if any later steps log it, it will be shown in plaintext.

.github/workflows/autofix_integration_omnibus.yml
prasden: It's as per the AWS' wiki on `Secure Usage of Agents in Github Actions.` It's to keep failures reproducible and traceable, and avoid breakages.  Yes, we can pin it to version 2.1

.github/workflows/autofix_integration_omnibus.yml
prasden: Yes, so even if the reasoning step fails halfway (Claude goes down midway), we still get the debug artifact it was producing.

.github/workflows/integration_omnibus.yml
prasden: Failures are detected the same way by checking if the outcome == failure. They're separate jobs because they test certain versions * FIPS * other dimensions and it would be verbose to check each one.  Since they're already separate jobs, we pass `integration` and `version` directly from the matrix fields instead of parsing them

.github/workflows/integration_omnibus.yml
prasden: If any of the `needs` jobs fail, it would [skip this step](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idneeds)

tests/ci/integration/auto_fix_integrations/claude-settings.json
prasden: By default, sandboxes can only write to the [current working directory and its subdirectories](https://code.claude.com/docs/en/sandboxing#configure-sandboxing). Claude's working directory is the repo checkout, and everything it [needs (patches, scripts, clones, logs) ](https://github.com/aws/aws-lc/pull/3287/files#diff-3d4356aca7514caf7b7a0c527f2663f817e862544a2afb01e5d5774e13d85db4R10-R12)lives inside that same directory. So it already can't write anywhere else.

tests/ci/integration/auto_fix_integrations/prompt.md
prasden: When we exit with non-zero code, [we retry](https://github.com/aws/aws-lc/pull/3287/changes#diff-3d4356aca7514caf7b7a0c527f2663f817e862544a2afb01e5d5774e13d85db4R269-R270) until MAX_ATTEMPTS

tests/ci/integration/auto_fix_integrations/auto_fix.sh
prasden: Yes, I added a `scrub_logs` function and called it in the main loop to remove secrets from the logs. For the second point, if we want to fully hide and not worry about secrets, the alternative is potentially an s3 bucket.

.github/workflows/integration_omnibus.yml
prasden: Openssh has no `_patch` dir, so recognize filters it out before the version is ever used, I removed it

WillChilds-Klein: fair. this may come up again with other patchsets, so we should make the patchset directory structure uniform to simplify things.

I'd suggest a follow-up PR adjusting openssh's pathset directory structure, deleting this special case (as well as others), and adding a check to the integration workflow enforcing that structure moving forward.

tests/ci/integration/auto_fix_integrations/prompt.md
```diff
@@ -0,0 +1,45 @@
...
+3. In the matching subdirectory under SRC_CLONE_DIR_PLACEHOLDER, `git checkout` the exact ref the runner script uses.
+4. Run `patch --dry-run -p1` with the existing patches against that checkout to identify failing hunks. Only fix patches for the failing ref — do not modify patches for other refs.
+5. Read the failing source context in the checkout to understand what changed upstream.
+6. Author corrected patches, replacing broken ones. If a hunk is redundant (upstream already does what the patch did), delete it. To remove a whole patch file, use `git rm` (bare `rm` is not permitted in this environment).
+7. Validate: re-run `patch --dry-run -p1` against a clean checkout at the correct ref (use `git checkout -- .` / `git clean` is unavailable, so reset the checkout with `git checkout .` and re-checkout the ref). Every patch must apply with zero fuzz and zero rejects. If validation fails, go back to step 6 and fix the patches until they pass.
```
WillChilds-Klein: if we notice a lot of cleanly applied patches that still produce integration test failures, we should consider adding a step 7a. here that actually runs the integration test script with the new patchset.

tests/ci/integration/auto_fix_integrations/auto_fix.sh
```diff
@@ -0,0 +1,340 @@
...
+        else empty end' \
+      "${work_dir}/claude-${attempt}.log" || true
+    echo "::endgroup::"
+
+    # Detect changes to see if Claude committed a fix, and if so create the resolving patch.
+    if [[ -n "$(git -C "${SRC_ROOT}" rev-list "${base_ref}..HEAD")" ]]; then
+      git -C "${SRC_ROOT}" format-patch "${base_ref}..HEAD" -o "${work_dir}/out"
+      return
+    fi
+    [[ "${rc}" -eq 0 ]] && { echo "::warning::${target}: Claude declined; not retrying."; return; }
```
WillChilds-Klein: hm, perhaps "Claude ran but declined to create a patch"?

will that failure result in the CI job failing?

tests/ci/integration/auto_fix_integrations/prompt.md
WillChilds-Klein: gotcha. but when retries are exhausted, we just return from the `reason_integration_failure` function and the script exits. shouldn't we fail the CI job in the case where claude fails to produce a patch (for any reason)?

tests/ci/integration/auto_fix_integrations/claude-settings.json
WillChilds-Klein: i see. so by default, claude already shouldn't be able to read `~/.aws` or `~/.ssh`, we're just adding them here as a fail-safe?

tests/ci/integration/auto_fix_integrations/auto_fix.sh
```diff
@@ -261,6 +273,14 @@ reason_integration_failure() {
   echo "::warning::${target}: ${MAX_ATTEMPTS} transient Claude failures; giving up."
 }
 
+scrub_logs() {
+  local patterns='(AKIA|ASIA|AGPA|AIDA|AROA|AIPA|ANPA|ANVA)[A-Z0-9]{16}|gh[pousr]_[A-Za-z0-9]{36}|github_pat_[A-Za-z0-9_]{20,}|-----BEGIN[A-Z ]*PRIVATE KEY-----|[Bb]earer[[:space:]]+[A-Za-z0-9._-]{16,}'
+  for f in "$@"; do
+    [[ -f "$f" ]] || continue
+    LC_ALL=C sed -E -i "s/${patterns}/[REDACTED]/g" "$f"
+  done
+}
```
WillChilds-Klein: don't you already declare this on L109?

tests/ci/integration/auto_fix_integrations/auto_fix.sh
WillChilds-Klein: > For the second point, if we want to fully hide and not worry about secrets, the alternative is potentially an s3 bucket.

I'm somewhat inclined towards this, as it would eliminate the chance of leakage. However, it would mean more infrastructure we have to maintain... My main concern is -- what if some newly formatted secret is introduced later? What's the mechanism for updating this regex?

Curious to hear your thoughts @prasden as well as @nhatnghiho and @justsmth .

.github/workflows/autofix_integration_omnibus.yml
```diff
@@ -0,0 +1,159 @@
...
+        with:
+          name: resolve-audit-log-${{ steps.target.outputs.dir }}
+          path: .resolve-integration-failures/${{ steps.target.outputs.dir }}/claude-*.log
```
WillChilds-Klein: I prompted claude to consider the possibility of a malicious patch re-encoding env secrets to exfiltrate them in a format our scanning mechanisms can't detect (but an attacker can decode)

---

The audit-log artifact (`claude-*.log`) records Claude's full transcript including raw tool outputs, and uploads it with `if: always()`. In this repo that artifact is downloadable by anyone with `actions: read` — i.e., world-readable.

The credential-exfil channels you've defended against (env scrub, sandbox network allowlist, deny-list) all assume the *secrets* are what we're protecting. They don't protect against an injection that gets Claude to print attacker-chosen text into its tool outputs, which then lands in the public transcript verbatim. That's enough for a steganographic signal back to whoever planted the injection upstream — even if no actual secret leaves the box.

Two options:

1. Drop raw tool outputs from the transcript before upload (keep only Claude's own assistant-role text and a list of tool *invocations*, not their stdout/stderr). The current `jq` filter in `auto_fix.sh:707-711` already only renders `text` and `tool_use` — the upload still ships the unfiltered `claude-${attempt}.log` though. Upload only the rendered transcript, or scrub the raw log before upload.
2. Move the audit log out of public artifacts entirely (push to a private S3 bucket the resolve role can read; surface a link in the PR for maintainers).

Option 1 is probably enough for now — it's defense in depth against the threat model where Claude is tricked into emitting a covert channel.

tests/ci/integration/auto_fix_integrations/auto_fix.sh
```diff
@@ -0,0 +1,340 @@
...
+
+# Take the patch the reason step produced, scan it for leaked secrets, apply it
+# to a new branch, and open a draft PR. Skips if no patch was produced.
```
WillChilds-Klein: I prompted Claude to consider the fact that by applying (untrusted?) patches and executing the resulting binary, we're effectively running untrusted code in our CI environment (COdeBuild or GHA). does that pose any risk for credential exfil? or are our CodeBuild/GHA environments sufficiently locked down?

---

The path check (`grep -vE "^tests/ci/integration/${integration}_patch/"`) prevents Claude from writing *outside* the patch dir, but doesn't constrain the *content* of patches inside that dir. By design, the patch dir holds diffs that get applied to downstream source and built into the next nightly's binaries. A diff in there is shaped like a legitimate fix — there's no syntactic difference between "fix MD5/SHAKE128 usage" and "add a malicious gadget under a runtime condition."

Concrete attack: an upstream contributor (or someone who compromised an upstream maintainer) lands an upstream commit whose message body contains a prompt injection plus a "preferred fix snippet" for AWS-LC. The autofix flow reads `git log` per the prompt, ingests the injection, writes the suggested diff into `<integration>_patch/`. Reviewer skims the draft PR, sees a small diff that "matches the upstream change," merges. Next nightly applies it and runs the binary in CodeBuild.

.github/workflows/autofix_integration_omnibus.yml
WillChilds-Klein: This is similar to concern/mitigation discussed [here](https://github.com/aws/aws-lc/pull/3287#discussion_r3389081694).

tests/ci/integration/auto_fix_integrations/auto_fix.sh
prasden: Duplicated unintentionally, removed

tests/ci/integration/auto_fix_integrations/prompt.md
prasden: The first pass is checking the CI after the draft PR is created by Claude, and if see we failures consistently, agreed this is worth doing. But the caveat is that it would need to be in its own step as it we would need to build aws-lc and downstream with the toolchain preinstalled.

tests/ci/integration/auto_fix_integrations/prompt.md
prasden: Yes, we should. Modified so that the CI job fails when Claude is unable to produce a patch.  This is done in the script itself, no need to tell Claude in the system prompt.

tests/ci/integration/auto_fix_integrations/auto_fix.sh
prasden: See https://github.com/aws/aws-lc/pull/3287#discussion_r3390419062

.github/workflows/autofix_integration_omnibus.yml
prasden: Responding to this and https://github.com/aws/aws-lc/pull/3287#discussion_r3389081694 here  as they're similar: 

Going off this, I think Option 1 is a bit against the listed policy as we emit the output generated, which is where the secrets can potentially leak. You're right that they could use some regex pattern or different secrets can emerge which we don't scan for, and for that worry, we would need to create the S3 bucket and store logs there instead, which might be better because Github workflows artifacts [only last 90 days under their workflows](https://docs.github.com/en/organizations/managing-organization-settings/configuring-the-retention-period-for-github-actions-artifacts-and-logs-in-your-organization), which might be better in line with the policy. But I'm not sure how strict that is, considering our autofix bot is scoped to small patches, so option 1 could still be viable. 

S3 also seems pragmatic, but also makes us deal with infrastructure overhead. But as a defense in depth, we set [CLAUDE_CODE_SUBPROCESS_ENV_SCRUB](https://github.com/aws/aws-lc/pull/3287/changes#diff-27e0a28fb3fc24758e6e11f59dcdf5ba9555c9ba906c2423268ddba2c62d4128R57-R58) when Claude runs. As per [Anthropic's docs](https://code.claude.com/docs/en/env-vars), with this flag on: 

> The parent Claude process keeps these credentials for API calls, but child processes cannot read them, reducing exposure to prompt injection attacks that attempt to exfiltrate secrets via shell expansion. On Linux, this also runs Bash subprocesses in an isolated PID namespace so they cannot read host process environments via /proc; as a side effect, ps, pgrep, and kill cannot see or signal host processes. claude-code-action sets this automatically when allowed_non_write_users is configured. 

I also researched more into the docs and am also adding [`allowUnsandboxedCommands: false`](https://code.claude.com/docs/en/settings#sandbox-settings) to the settings. Without setting this, a sandbox blocked command (e.g wget to the AL instance) could retry outside the sandbox and bypass the network allowlist to fetch role credentials through a command using a prompt injection, which is another defense in depth we add. Not sure why Anthropic even allows out of sandbox commands when it fails in the first place. 

With this, I'm curious if option 1 seems viable or S3 seems more pragmatic from others based on these defenses and the policy listed. I think option 1 could be viable but s3 might be the defense in depth answer, in which I would need to refactor.   

cc redirect to this comment @justsmth @nhatnghiho.

tests/ci/integration/auto_fix_integrations/auto_fix.sh
prasden: see https://github.com/aws/aws-lc/pull/3287#discussion_r3391192639

tests/ci/integration/auto_fix_integrations/auto_fix.sh
prasden: For security in terms of merging to the repo, I think we have one defense layer with the [security-review](https://d28bfvmis1skm5.cloudfront.net/aws-lc/pr-3287/8bedc28a9c100e75839870bf582e721dcf0ca916.html) CI job, but that is also dependent on no prompt injections as it also runs on Claude.

cc @nhatnghiho -- does the bot have any safeguards around prompt injections in the PR content?

To prevent any leakages or trouble, one mitigation is not running `integration_omnibus` on draft PRs, which is doable by editing the workflow file.

We could also have the autofix flow open the PR with the workflow's GITHUB_TOKEN instead of the PAT. Events created by GITHUB_TOKEN don't [trigger new workflow runs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#triggering-a-workflow-from-a-workflow), so the integration CI wouldn't fire automatically when the draft is opened.

However, GITHUB_TOKEN is read only  and `Allow GitHub Actions to create and approve pull requests` is disabled under the aws-lc repo wide settings, which is why we use a PAT. This change would let workflows with `pull-request:write` permissions to create PRs, but also approve PRs. There is no separate `approve` scope,  so there's no way to enable write without also enabling "approve."   

The PAT needs no setting change and still runs CI on the draft since it [triggers workflows](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#triggering-a-workflow-from-a-workflow), but it's a broader token because its `public_repo` scope can write to any public repo, not just aws-lc. We would need to make a fine-grained token if we want just aws-lc and store it in Secrets Manager. The `GITHUB_TOKEN` route is workflow scoped and won't auto-run the CI triggered by the draft PR, but we'd need to turn on [Allow GitHub Actions to create and approve pull requests](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/disable-and-enable-workflows#preventing-github-actions-from-creating-or-approving-pull-requests) in the repo settings(Settings > Actions > General > Workflow permissions) and it would potentially mean approval perms.

However, Claude only runs in the `reason` step due to our separated steps, so I think this approach could be better if we don't want any malicious patches triggered on the CodeBuild machines, but we accept the risk of approval perms. With this, we could remove the PAT token retrieval role from the CDK as well. Curious on your thoughts.

prasden: I also pointed Claude at how we can harden our CI and Codebuild infra to defend against malicious attacks, I would need to research into these fixes:

-------------------

Since the malicious patch executes in the integration CodeBuild environmen the durable mitigations are hardening that environment itself — these are pre-existing posture items.

  **1. Egress restriction** — the integration CodeBuild runner is created with no VPC
  ([`aws_lc_github_actions_stack.py#L127`](https://github.com/aws/aws-lc/blob/main/tests/ci/cdk/cdk/aws_lc_github_actions_stack.py#L127)), so putting it in a VPC with an egress allowlist (as the [fuzz stack](https://github.com/aws/aws-lc/blob/main/tests/ci/cdk/cdk/aws_lc_github_fuzz_ci_stack.py#L66-L123) already does) would stop a malicious binary from exfiltrating to arbitrary hosts.

**2. Cut the container's path to creds** — the inner container always mounts the host Docker socket
  ([`codebuild-docker-run.sh#L65`](https://github.com/aws/aws-lc/blob/main/.github/actions/codebuild-docker-run/codebuild-docker-run.sh#L65)), so defaulting `AWS_EC2_METADATA_DISABLED=true` and only mounting the socket for jobs that need Docker-in-Docker would remove the breakout path to the runner's role.
  
  **3. Re-examine `credentials: true`** — only `mysql`/`monit` forward AWS creds into the untrusted container ([`codebuild-docker-run.sh#L55-L56`](https://github.com/aws/aws-lc/blob/main/.github/actions/codebuild-docker-run/codebuild-docker-run.sh#L55-L56)), so if those only need an ECR pull we can do it on the outer runner instead of handing creds to downstream code.
  
**4. Drop `--privileged`** — it's auto-added for every ipv6 job  ([`codebuild-docker-run.sh#L47-L48`](https://github.com/aws/aws-lc/blob/main/.github/actions/codebuild-docker-run/codebuild-docker-run.sh#L47-L48)), and privileged + the Docker socket is a trivial host takeover, so scoping it to only jobs that truly need it shrinks the escape surface.
  
**5. Scope the runner's ECR push** — integration tests don't push images but the task role can ([`aws_lc_github_actions_stack.py#L81-L90`](https://github.com/aws/aws-lc/blob/main/tests/ci/cdk/cdk/aws_lc_github_actions_stack.py#L81-L90)), so making the integration project pull-only removes the main payoff of a breakout (CI supply-chain poisoning).

tests/ci/integration/auto_fix_integrations/claude-settings.json
prasden: I think I misunderstood what you were asking initially. I now updated it so read access is denied to all but the src folder we created in the script as you were saying: [8d0dc75](https://github.com/aws/aws-lc/pull/3287/commits/8d0dc75b7bf9f65196ae36bf1bbb4539b6273679). The sandbox statements still hold true -- now we scope it to only the directories we create, which is what you were suggesting.

.github/workflows/autofix_integration_omnibus.yml
justsmth: My vote is to use S3. I think the policy decides it for us. Option 1 drops raw tool responses, which the policy says to keep and store immutably -- so scrubbing seems like a violation.

Two things that should be part of the S3 setup:
- write from "reason" with `s3:PutObject` only (not `GetObject`, so a compromised job can't read back other transcripts), and
- Use "Object Lock" or "versioning + deny-delete" for immutability.

We're already touching CDK here, so a put-only bucket feels in-scope.

@nhatnghiho objections?

tests/ci/integration/auto_fix_integrations/auto_fix.sh
justsmth: I'd split this in two.

**This PR:** keep untrusted patches off the public repo until a human looks. Gating on draft alone won't do it -- the branch lands in `aws/aws-lc` and `push: ['*']` builds it before anyone closes it. Cleaner is to open the draft in a *private* staging repo and have a human promote it to a real PR against `aws/aws-lc` once reviewed. That keeps the branch off the public runners (no `push` build, no auto-CI) and also keeps the body/diff out of public view during the review window -- which matters since a nightly failure could otherwise sit exposed over a weekend. Cost is a private mirror + promotion step.


**Follow-up:** the CodeBuild posture items (VPC egress, `--privileged`, IMDS, ECR push) are real but pre-existing infrastructure hardening with a large blast radius -- it deserves its own design + testing. With cross-repo draft PRs, the marginal risk is ~"same as any contributor PR." Can you capture these as tracked issues?

cc @nhatnghiho @WillChilds-Klein

prasden: Do you mean the contributor's fork by a private mirror? Could you elaborate on that? Private mirrors/forks get tricky because you would have to deploy the CDK and set up all the credentials with the Github PAT for that repo/account again.

justsmth: > Do you mean the contributor's fork by a private mirror?

The goal is simply that the PR that gets posted is not publicly visible. Is there a path to getting there that wouldn't be painful?

prasden:
Had an offline discussion regarding running Claude on third party logs and AI generated code -- https://github.com/aws/aws-lc/pull/3287#discussion_r3389213305, https://github.com/aws/aws-lc/pull/3287#discussion_r3389212302,  decided security risks outweigh the pros for now. Removed the "autofix" part from the PR and now we only raise an alarm when integration tests fail.

WillChilds-Klein (dismissed):

.github/workflows/integration_omnibus.yml
```diff
@@ -4,13 +4,16 @@ on:
...
+  schedule:
+    - cron: "0 0 * * *"
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name }}
   cancel-in-progress: true
 env:
   GOPROXY: https://proxy.golang.org,direct
 permissions:
   contents: read
+  actions: read
```
justsmth, line 18: This also needs `id-token: write`?

.github/workflows/integration_omnibus.yml
```diff
@@ -418,6 +422,16 @@ jobs:
           run: |
             source /opt/compiler-env/setup-${{ matrix.compiler }}.sh
             ${{ matrix.run }}
+      # Emit a marker only when the integration test itself fails, not on
+      # operational failures, so the report-failures alarm counts real downstream breakages only.
+      - name: Mark integration failure
+        if: steps.integration-test.outcome == 'failure' && github.event_name == 'schedule'
```
justsmth: This `if` has no status-check function, so GitHub implicitly evaluates it as `success() && (...)`. But `integration-test` has just failed at this point, so the job is already in a failed state, `success()` is `false`, and this step — plus the `upload-artifact` step below — gets **skipped**. The result is that no marker artifacts are ever uploaded on a real failure, `report-failures` always counts `0`, and the alarm never fires?

This needs [`always()`](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#always).
```yaml
      # Emit a marker only when the integration test itself fails, not on
      # operational failures, so the report-failures alarm counts real downstream breakages only.
      # always() is required so the step still runs after the test step has failed; allow_failure
      # entries (informational, subject to upstream breakage) are excluded to avoid false alarms.
      - name: Mark integration failure
        if: always() && steps.integration-test.outcome == 'failure' && github.event_name == 'schedule' && !matrix.allow_failure
        run: echo "${{ matrix.name }}-${{ matrix.arch }}" > integration-failure.txt
      - uses: actions/upload-artifact@v4
        if: always() && steps.integration-test.outcome == 'failure' && github.event_name == 'schedule' && !matrix.allow_failure
```

.github/workflows/integration_omnibus.yml
justsmth, line 314: `nmap-master` is explicitly informational ("subject to upstream breakage"), but the new marker logic fires on *any* entry whose test step fails -- including this one. 


We should append `&& !matrix.allow_failure` to the marker `if`s below -- unless we want it to fire for this?

.github/workflows/integration_omnibus.yml
```diff
@@ -496,3 +520,42 @@ jobs:
...
+      - name: Emit failure metric
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -exo pipefail
+          # Each failed integration test uploads one artifact, so count them to
+          # get the number of breakages.
+          failure_count=$(gh api --paginate \
+            "/repos/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}/artifacts" \
+            --jq '[.artifacts[] | select(.name | startswith("integration-failure-"))] | length')
```
justsmth: With `--paginate`, the `--jq` filter runs once per page, so if artifacts ever exceed one page (100) you'd get multiple numbers on separate lines and `--value "$failure_count"` would break.

How close might we get to exceeding a page?

.github/workflows/integration_omnibus.yml
```diff
@@ -462,6 +477,14 @@ jobs:
           run: |
             source /opt/compiler-env/setup-gcc-12.sh
             ./tests/ci/integration/run_python_integration.sh ${{ matrix.version }}
+      - name: Mark integration failure
+        if: steps.integration-test.outcome == 'failure' && github.event_name == 'schedule'
+        run: echo "python-${{ matrix.version }}-${{ matrix.buildFIPS }}-${{ matrix.crtUseSystemCrypto }}-${{ matrix.arch }}" > integration-failure.txt
+      - uses: actions/upload-artifact@v4
+        if: steps.integration-test.outcome == 'failure' && github.event_name == 'schedule'
```
justsmth: `always()`

.github/workflows/integration_omnibus.yml
prasden: We don't want to fire for this if it's only informational. Added a gate for this, thank you

.github/workflows/integration_omnibus.yml
prasden: Yes, to assume the credentials. Thank you

.github/workflows/integration_omnibus.yml
prasden: Good catch, added.

.github/workflows/integration_omnibus.yml
prasden: Made the count work correctly across pages so it stays accurate even if we ever go past 100

.github/workflows/integration_omnibus.yml
WillChilds-Klein: How was this able to post metrics previously?

WillChilds-Klein (dismissed):

.github/workflows/integration_omnibus.yml
prasden: If you're talking about the example ticket, I used the CLI federated into AWS-LC's namespace and triggered the threshold to go over

.github/workflows/integration_omnibus.yml
```diff
@@ -496,3 +522,44 @@ jobs:
...
+      - codebuild-aws-lc-ci-github-actions-${{ github.run_id }}-${{ github.run_attempt }}
+        image:linux-5.0
+        instance-size:small
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/configure-aws-credentials
+        with:
+          roleName: AwsLcGitHubActionStandardRole
```
nhatnghiho: nit: this can be removed since `configure-aws-credentials` use `AwsLcGitHubActionStandardRole` by default.

.github/workflows/integration_omnibus.yml
```diff
@@ -496,3 +522,44 @@ jobs:
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -exo pipefail
+          # Count the failure artifacts to get the number of broken integration tests.
+          # We keep only the integration-failure-* markers and count lines with wc -l instead of jq length,
+          # since --paginate would make jq emit a separate count for each page.
+          failure_count=$(gh api --paginate \
+            "/repos/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}/artifacts" \
+            --jq '.artifacts[] | select(.name | startswith("integration-failure-")) | .name' \
+            | wc -l | tr -d ' ')
```
nhatnghiho, line 558: nit: CloudWatch restrictions made it so that we can't have a more detailed description to use on our tickets. To reduce friction, consider updating the alarm description to contain a URL that points directly to scheduled integration runs. Something like [this](https://github.com/aws/private-aws-lc-staging/actions/workflows/integration_omnibus.yml?query=event%3Aschedule)

.github/workflows/integration_omnibus.yml
```diff
@@ -496,3 +522,44 @@ jobs:
           run: |
             source /opt/compiler-env/setup-gcc-13.sh
             ./tests/ci/integration/run_ruby_integration.sh ${{ matrix.version }}
+      - name: Mark integration failure
+        if: always() && steps.integration-test.outcome == 'failure' && github.event_name == 'schedule'
+        run: echo "ruby-${{ matrix.version }}-${{ matrix.buildFIPS }}-${{ matrix.arch }}" > integration-failure.txt
```
nhatnghiho: For this and all similar commands, if we only care about the existence of `integration-failure.txt` per failed job, the `echo` is a bit redundant? Consider using `touch integration-failure.txt` instead for better clarity

nhatnghiho (commented):
Sorry for the late reply. I have just had time to sit down and review all the conversations. 

To summarize my conversations with @prasden here:

Removing `auto-fix` seems to be a most practical approach right now, but if we ever want to revisit this idea in the future, we can consider:
1. Running this workflow from a dedicated branch in the private staging repo. We already have all the CodeBuild infrastructure set up for that.
2. To reduce friction, consider adding another workflow or bot that can tag the approved patch PRs and open a new PR in the public repo.

.github/workflows/integration_omnibus.yml
prasden: Done, removed the redundant `roleName` since it defaults to `AwsLcGitHubActionStandardRole`.

.github/workflows/integration_omnibus.yml
prasden: Done

.github/workflows/integration_omnibus.yml
prasden: The alarm was created manually after federating into AWS-LC's namespace using the CloudWatch UI, and it was set to trigger when this failure count threshold exceeds 0. It is not defined in CDK, so unfortunately the limitation is that we cannot link the specific integration that is failing. See `P441267965` in ticketing to view an alarm that I manually raised to test before I created the PR. I did link to the actual `integration_omnibus.yml` in the description though.

nhatnghiho: The current link points to all integration tests. I meant to update the description so that it'll show only scheduled integration tests.

prasden: Updated the alarm in CloudWatch to link the scheduled runs.

justsmth (approved):
Thanks for re-scoping this down to just the nightly + alarm.

Nit: Would you update the description to be consistent with the current changes?

prasden:
Yes, I've updated the description and the title.

WillChilds-Klein (approved):

nhatnghiho (approved):

PR #3288, opened 2026-06-04
ci:ctest timing summary for MacOS

dougch:
### Description of changes:

Our macOS CI jobs run several build configurations sequentially, making it hard to identify which tests are the long poles. This PR adds per-test-case timing by capturing gtest's native JSON reports and surfaces the top 20 slowest tests in the GitHub Actions step summary for both the macOS-x86 (FIPS) and macOS-ARM jobs.

#### Changes:
  - In `util/all_tests.go`, when `GTEST_REPORT_DIR` is set, pass `GTEST_OUTPUT=json:<file>` to each test binary so gtest emits a per-test-case timing report. Report filenames include the binary name, shard index, and PID so sequential/sharded runs don't overwrite each other.
  - Add `tests/ci/test_results_summary.py`, which aggregates all gtest JSON reports in the directory and emits a markdown table of the slowest test cases to the step summary.
  - Wire `GTEST_REPORT_DIR` and a `Test timing summary` step (`if: always()`) into the macOS-x86 (FIPS) and macOS-ARM jobs in `actions-ci.yml`.

### Call-outs:

1. This is just a sample; not all jobs that could emit slow test data were set up.
2. With the size of these workflows, this table format might not scale/be readable — 20 entries that are ~1 sec apart might not be that useful.
3. Does this data belong in CloudWatch instead/also? Are these metrics actionable ?

### Testing:

This PR; see an example [Summary](https://github.com/aws/aws-lc/actions/runs/27224117549?pr=3288).
For each arch the python script creates a top-20 slowest-tests table; an in-line example:
Screenshot 2026-06-09 at 12 10 28


util/all_tests.go
```diff
@@ -285,6 +285,17 @@ func runTestOnce(test test, mallocNumToFail int64) (passed bool, err error) {
 		cmd.Env = append(cmd.Env, fmt.Sprintf("GTEST_SHARD_INDEX=%d", test.shard))
 		cmd.Env = append(cmd.Env, fmt.Sprintf("GTEST_TOTAL_SHARDS=%d", test.numShards))
 	}
+	// Write per-test-case gtest JSON report if report dir is set.
+	if reportDir := os.Getenv("GTEST_REPORT_DIR"); reportDir != "" {
+		os.MkdirAll(reportDir, 0777)
+		binName := filepath.Base(prog)
+		reportFile := filepath.Join(reportDir, fmt.Sprintf("%s_shard_%d_%d.json", binName, test.shard, os.Getpid()))
```
justsmth: The pid will be the *parent* process's pid, so the same across across all of the children. You could use an atomic counter to ensure unique values: 
```
// import sync
import (
    ...
    "sync"
    "sync/atomic"
    ...
)
...
/// package-level counter
var reportSeq int64
...
// Then here:
reportFile := filepath.Join(reportDir,
    fmt.Sprintf("%s_shard_%d_%d.json", binName, test.shard, atomic.AddInt64(&reportSeq, 1)))
```

tests/ci/test_results_summary.py
```diff
@@ -0,0 +1,73 @@
...
+        except (json.JSONDecodeError, IOError):
+            continue
+
+        for suite in data.get("testsuites", []):
+            suite_name = suite.get("name", "")
+            for test in suite.get("testsuite", []):
+                name = f"{suite_name}.{test.get('name', '')}"
+                time_str = test.get("time", "0s")
+                secs = float(time_str.rstrip("s")) if time_str.endswith("s") else float(time_str)
+                status = "PASS" if test.get("failures") is None and test.get("result") != "SUPPRESSED" else "FAIL"
```
justsmth: Why treat `SUPPRESSED` as as a failure? Shouldn't these just be skipped?
```
if result in ("SUPPRESSED",) or test.get("status") == "NOTRUN":
    continue  # not run; skip
status = "FAIL" if test.get("failures") is not None else "PASS"
```

util/all_tests.go
```diff
@@ -285,6 +285,17 @@ func runTestOnce(test test, mallocNumToFail int64) (passed bool, err error) {
 		cmd.Env = append(cmd.Env, fmt.Sprintf("GTEST_SHARD_INDEX=%d", test.shard))
 		cmd.Env = append(cmd.Env, fmt.Sprintf("GTEST_TOTAL_SHARDS=%d", test.numShards))
 	}
+	// Write per-test-case gtest JSON report if report dir is set.
+	if reportDir := os.Getenv("GTEST_REPORT_DIR"); reportDir != "" {
+		os.MkdirAll(reportDir, 0777)
```
justsmth: * This ignores any failures.
* Only `0755` is needed -- only owner needs write permission.

.github/workflows/actions-ci.yml
```diff
@@ -87,8 +92,13 @@ jobs:
         with:
           go-version: ">=1.18"
       - name: Build ${{ env.PACKAGE_NAME }}
+        env:
+          GTEST_REPORT_DIR: ${{ github.workspace }}/gtest_reports
```
WillChilds-Klein: why just MacOS? is there cost to measuring other platforms?

WillChilds-Klein (commented):
> This is just a sample; not all jobs that could emit slow test data were set up.

why not just time all tests and only show the 20 slowest in the summary?

> Does this data belong in CloudWatch instead/also? Are these metrics actionable ?

i would vote just keeping it here. if you're annoyed with slow runtime on a test/PR you care about you can check the summary right from the PR. timing might fluctuate by run and over time, so adequately tuning some alarm threshold seems brittle.

.github/workflows/actions-ci.yml
dougch: MacOS is the next slowest job after BSD. Cost is added PR lines...but I'm happy to expand it, if it proves useful.

util/all_tests.go
dougch: Thanks, I misunderstood how this would play out.

util/all_tests.go
```diff
@@ -285,6 +289,22 @@ func runTestOnce(test test, mallocNumToFail int64) (passed bool, err error) {
...
 	}
+	// Write per-test-case gtest JSON report if report dir is set. Report
+	// generation is a diagnostic aid, so on error we warn and skip the report
+	// rather than failing the test run.
+	if reportDir := os.Getenv("GTEST_REPORT_DIR"); reportDir != "" {
+		if err := os.MkdirAll(reportDir, 0755); err != nil {
+			fmt.Fprintf(os.Stderr, "Warning: could not create GTEST_REPORT_DIR %q: %v; skipping timing report\n", reportDir, err)
+		} else {
+			binName := filepath.Base(prog)
+			reportFile := filepath.Join(reportDir, fmt.Sprintf("%s_shard_%d_%d.json", binName, test.shard, atomic.AddInt64(&reportSeq, 1)))
```
justsmth: Sorry -- I think the PID would still be needed for uniqueness across iterations:

```
reportFile := filepath.Join(reportDir, fmt.Sprintf("%s_shard_%d_%d_%d.json",
    binName, test.shard, os.Getpid(), atomic.AddInt64(&reportSeq, 1)))
```

justsmth (commented):
Looks good -- I have just one lingering "uniqueness" concern.

justsmth (approved):

WillChilds-Klein (approved):

WillChilds-Klein (approved):

justsmth (approved):

